PR #37527, opened 2022-11-11
[Impeller] reuse texture if size and type matches

jonahwilliams:
Might help slightly, still need to pay the cost of reupload though

PR #37727, opened 2022-11-17
Revert "[Impeller] reuse texture if size and type matches"

jonahwilliams:
Reverts flutter/engine#37527

Fixes https://github.com/flutter/flutter/issues/115580


zanderso (approved):

jonahwilliams:
This caused visible issues on the segmented control page in flutter/gallery. Reverting made those issues go away. I believe this is the cause of the jumbled text, not anything to do with runtime effect or snapshot widget

PR #37737, opened 2022-11-18
Make NotifyIdle reject close and past deadlines.

dnfield:
This patch also eliminates some extraneous tracing that is happening every frame. It is possible to get the same trace calls by enabling the API stream if needed.

Also refactors the NotifyIdle callsites to just always work in TimeDeltas rather than converting back and forth between them and TimePoints, which I think reads more clearly.

Also refactors the way the delayed task gets posted to hopefully be a little more safe/clear.

This patch does not "solve" the problem of when we call notify idle and then immediately start pumping frames, e.g. when we get a super cheap frame and then follow it up with a transition.

### DIFF
--- a/shell/common/animator.cc
+++ b/shell/common/animator.cc
@@ -57,12 +57,6 @@ void Animator::EnqueueTraceFlowId(uint64_t trace_flow_id) {
       });
 }
 
-static fml::TimePoint FxlToDartOrEarlier(fml::TimePoint time) {
-  auto dart_now = fml::TimeDelta::FromMicroseconds(Dart_TimelineGetMicros());
-  fml::TimePoint fxl_now = fml::TimePoint::Now();
-  return fml::TimePoint::FromEpochDelta(time - fxl_now + dart_now);
-}
-
 void Animator::BeginFrame(
     std::unique_ptr<FrameTimingsRecorder> frame_timings_recorder) {
   TRACE_EVENT_ASYNC_END0("flutter", "Frame Request Pending",
@@ -81,7 +75,6 @@ void Animator::BeginFrame(
   }
 
   frame_scheduled_ = false;
-  notify_idle_task_id_++;
   regenerate_layer_tree_ = false;
   pending_frame_semaphore_.Signal();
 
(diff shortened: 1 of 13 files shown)


dnfield:
@a-siva @rmacnak-google fyi

shell/common/animator.cc
```diff
@@ -106,34 +99,29 @@ void Animator::BeginFrame(
...
 
   if (!frame_scheduled_ && has_rendered_) {
-    // Under certain workloads (such as our parent view resizing us, which is
-    // communicated to us by repeat viewport metrics events), we won't
-    // actually have a frame scheduled yet, despite the fact that we *will* be
-    // producing a frame next vsync (it will be scheduled once we receive the
-    // viewport event).  Because of this, we hold off on calling
-    // |OnAnimatorNotifyIdle| for a little bit, as that could cause garbage
-    // collection to trigger at a highly undesirable time.
+    // Wait a tad more than 3 60hz frames before reporting a big idle period.
```
zanderso, line 107: Do we have any logic in place that prevents having multiple outstanding NotifyIdle calls scheduled? Said differently, is there logic to ensure that there is a maximum of one call at a time scheduled to run? If that property is a side-effect of something else, would it be worth making it explicit?

zanderso (approved):
lgtm w/ some questions.

shell/common/animator.cc
dnfield: The old logic said if `RequestFrame` and `BeginFrame` have not been called again for 51ms, fire a big idle notification. The new logic says if `RequestFrame` has not been called in the last 51ms and the last known vsync target time is not in the future, fire.

We could still easily end up in a situation where `NotifyIdle` is called at the end of a frame and then called again ~51ms later if there are no more frames.

The situation we are still not avoiding is that we could call `NotifyIdle` at the end of a very cheap frame only to immediately start pumping an animation/route transition in the very next frame.

dnfield: And to answer your specific question: we could schedule multiple tasks, but the guards in those tasks would prevent notify idle from being called repeatedly (i.e. more than once per frame or every 51ms when no frames are pumped).

PR #37738, opened 2022-11-18
[web] Render in custom target

ditman:
This PR is the first iteration of the embedding of a Flutter web app in a custom host element.

It touches the following bits of the engine:

* view embedding (to render the flutter app in the correct location)
* pointer binding (so pointer events are relative to the app's glassPane, and not the browser viewport)
* window sizing (to detect resizing also in DIVs)

This PR is backwards compatible, and by default, Flutter web apps render in "full-screen" mode (the same as what flutter has been doing until this change). If an app is configured correctly, it can run in "custom-element" mode, and then, it'll be restricted to that single element in the DOM.

This introduces some bits of [the new design for Flutter views](https://github.com/flutter/flutter/issues/117098):

* `DimensionsProvider` provides an api to observe/retrieve measurements from flutter's host element.
* `EmbeddingStrategy` provides an api to render flutter in its host element.

### Issues

* Fixes https://github.com/flutter/flutter/issues/32329
* Fixes https://github.com/flutter/flutter/issues/114702
* Fixes https://github.com/flutter/flutter/issues/114704
* Fixes https://github.com/flutter/flutter/issues/115216 (opportunistic refactor)

### Known issues

In custom-element mode:

* https://github.com/flutter/flutter/issues/117091
* https://github.com/flutter/flutter/issues/116561

In full-screen mode:

* n/a

In general:

* A bunch of technical debt was uncovered while working on this PR. See individual issues mentioning this PR in the message threads below.

### Testing

* New files are unit-tested.
* Old tests pass with minimal modifications.
* Test app deployed to Firebase hosting.

## Pre-launch Checklist

- [x] I signed the [CLA].


lib/web_ui/lib/src/engine/embedder.dart
```diff
@@ -17,7 +17,7 @@ import 'pointer_binding.dart';
 import 'safe_browser_api.dart';
 import 'semantics.dart';
 import 'text_editing/text_editing.dart';
-import 'util.dart';
+import 'view_embedder/application_dom/application_dom.dart';
```
ditman, line 20: Please, **do** look at this file, even if github doesn't want to show it to you by default!

lib/web_ui/lib/src/engine/host_node.dart
```diff
@@ -152,6 +155,18 @@ class ShadowDomHostNode implements HostNode {
 class ElementHostNode implements HostNode {
```
ditman, line 155: This only applies in the case where ShadowDOM is not supported.

lib/web_ui/lib/src/engine/host_node.dart
```diff
@@ -187,3 +202,105 @@ class ElementHostNode implements HostNode {
   @override
   void appendAll(Iterable<DomNode> nodes) => nodes.forEach(append);
 }
+
+// Applies the required global CSS to an incoming [DomCSSStyleSheet] `sheet`.
+void applyGlobalCssRulesToSheet(
```
ditman, line 207: We should apply the `color: red` rule here, so it only affects to the contents of the `flt-glass-pane`, and not the outer slots!

lib/web_ui/lib/src/engine/view_embedder/application_dom/application_dom.dart
```diff
@@ -0,0 +1,95 @@
...
+///
+/// This class is specialized to handle different types of DOM embeddings:
+///
+/// * [FullPageApplicationDom] - The default behavior, where flutter takes
+///   control of the whole web page. This is how Flutter Web used to operate.
+/// * [CustomElementApplicationDom] - Flutter is rendered inside a custom host
+///   element, provided by the web app programmer through the engine
+///   initialization.
+abstract class ApplicationDom {
+  @mustCallSuper
```
ditman, line 26: This annotation is probably not needed in the constructor, as the default constructor will be called by default unless another one is called in the subclass?

lib/web_ui/lib/src/engine/view_embedder/application_dom/full_page_application_dom.dart
```diff
@@ -0,0 +1,128 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:js/js.dart';
+
+import '../../dom.dart';
+import '../../util.dart' show assertionsEnabled, setElementStyle;
+import 'application_dom.dart';
+
+
```
ditman, line 11: Lots of DartDoc missing, why isn't analyzer shouting at me?

lib/web_ui/lib/src/engine/view_embedder/dimensions_provider/custom_element_dimensions_provider.dart
```diff
@@ -0,0 +1,45 @@
...
+
+import 'package:ui/src/engine/window.dart';
+import 'package:ui/ui.dart' as ui show Size;
+
+import '../../dom.dart';
+import 'dimensions_provider.dart';
+
+// import 'custom_element_application_dom.dart';
+// import 'full_page_application_dom.dart';
+// import 'hot_restart_cache_handler.dart';
```
ditman, line 13: // Delete

lib/web_ui/lib/src/engine/view_embedder/dimensions_provider/dimensions_provider.dart
```diff
@@ -0,0 +1,49 @@
...
+import 'package:ui/ui.dart' as ui show Size;
+
+import '../../dom.dart';
+import '../../platform_dispatcher.dart';
+import 'custom_element_dimensions_provider.dart';
+import 'full_page_dimensions_provider.dart';
+
+// import 'custom_element_application_dom.dart';
+// import 'full_page_application_dom.dart';
+// import 'hot_restart_cache_handler.dart';
```
ditman, line 15: // Delete

ditman:
The failing test looks like a CIPD issue (or maybe that my branch is missing some fix from master):

```
Built 167 tests in 685 seconds using 8 concurrent compile processes.
Unhandled exception:
Bad state: Rejecting attempt to install Chromium on LUCI. LUCI is expected to provide Chromium as a CIPD dependency managed using .ci.yaml.
#0      ChromeInstaller.install (file:///C:/b/s/w/ir/cache/builder/src/flutter/lib/web_ui/dev/chrome_installer.dart:143:7)
#1      getOrInstallChrome (file:///C:/b/s/w/ir/cache/builder/src/flutter/lib/web_ui/dev/chrome_installer.dart:54:23)
```

lib/web_ui/lib/src/engine/view_embedder/application_dom/custom_element_application_dom.dart
```diff
@@ -0,0 +1,104 @@
...
+import 'dart:async';
+
+import '../../dom.dart';
+import 'application_dom.dart';
+
+class CustomElementApplicationDom extends ApplicationDom {
+
+  CustomElementApplicationDom(this._hostElement) {
+    // Clear children...
+    while(_hostElement.firstChild != null) {
```
yjbanov, line 14: Should we instead assert that the host element is empty?

lib/web_ui/lib/src/engine/view_embedder/application_dom/custom_element_application_dom.dart
```diff
@@ -0,0 +1,104 @@
...
+      return true;
+    }());
+
+    _resizeObserver?.observe(_hostElement);
+  }
+
+  final DomElement _hostElement;
+  late DomResizeObserver? _resizeObserver;
+
+  final StreamController<DomResizeObserverEntry> _streamController =
```
yjbanov, line 40: Let's pick a name that reflects this stream controller's role. Otherwise, it just repeast the type name (`String _string` 😛).

lib/web_ui/lib/src/engine/view_embedder/application_dom/application_dom.dart
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  /// Keeps a list of elements to be cleaned up at hot-restart.
+  HotRestartCacheHandler? _hotRestartCache;
+
+  /// Returns the 'type' (a String for debugging) of the ApplicationDom instance.
+  String get type;
+
+  /// Sets-up the viewport Meta-Tag for the app.
+  void applyViewportMeta();
```
yjbanov, line 50: Viewport meta is a page-global property. I don't think this class should contain methods that are inherently page-global. Instead, can we have something more generic, such as `void initialize();`, and if an implementation needs to do global things, e.g. `FullPageApplicationDom`, it would do it privately. Others, such as `CustomElementApplicationDom`, wouldn't. For exampe, there would no need for the `// NOOP` implementation of this method.

lib/web_ui/lib/src/engine/view_embedder/application_dom/application_dom.dart
yjbanov: Yeah, I'm not sure this annotation has any effect on constructors.

lib/web_ui/lib/src/engine/embedder.dart
```diff
@@ -180,102 +133,32 @@ class FlutterViewEmbedder {
...
-      hasAutofillOverlay: browserHasAutofillOverlay(),
-    );
-
-    final DomHTMLBodyElement bodyElement = domDocument.body!;
-
-    bodyElement.setAttribute(
+    _applicationDom.setHostAttribute(
       'flt-renderer',
       '${renderer.rendererTag} (${FlutterConfiguration.flutterWebAutoDetect ? 'auto-selected' : 'requested explicitly'})',
     );
```
yjbanov, line 139: I think `flt-renderer` and `flt-build-mode` are inherently global, so we can just stick them in the `<body>` tag. `ApplicationDom` shouldn't be concerned with them.

lib/web_ui/lib/src/engine/embedder.dart
```diff
@@ -180,102 +133,32 @@ class FlutterViewEmbedder {
...
-    // events properly.
-    setElementStyle(bodyElement, 'touch-action', 'none');
-
-    // These are intentionally outrageous font parameters to make sure that the
-    // apps fully specify their text styles.
-    setElementStyle(bodyElement, 'font', defaultCssFont);
-    setElementStyle(bodyElement, 'color', 'red');
-
+    _applicationDom.setHostAttribute('flt-build-mode', buildMode);
+    _applicationDom.setHostAttribute('flt-application-dom', _applicationDom.type);
```
yjbanov, line 141: I think each implementation can set this attribute privately. I'm not sure there's value in having them be set from outside. As a bonus, it will also obviate the need for the `type` getter.

lib/web_ui/lib/src/engine/embedder.dart
```diff
@@ -180,102 +133,32 @@ class FlutterViewEmbedder {
...
-          print(
-            'WARNING: found an existing <meta name="viewport"> tag. Flutter '
-            'Web uses its own viewport configuration for better compatibility '
-            'with Flutter. This tag will be replaced.',
-          );
-        }
-      }
-      viewportMeta.remove();
-    }
+    _applicationDom.setHostAttribute('spellcheck', 'false');
```
yjbanov, line 144: This can also be set internally by the implementation. Together with the suggestions above, I think we can remove `setHostAttribute` too. I think it would be best for `ApplicationDom` to fully own its element, and not let random mutations of its style and other attributes from outside.

lib/web_ui/lib/src/engine/embedder.dart
```diff
@@ -180,102 +133,32 @@ class FlutterViewEmbedder {
...
 
-    // This must be appended to the body, so the engine can create a host node
-    // properly.
-    bodyElement.append(glassPaneElement);
+    // This must be appended to the applicationDom now, so the engine can create
+    // a host node (ShadowDOM or a fallback) next.
+    //
+    // The applicationDom will take care of cleaning up the glassPane on hot
+    // restart.
+    _applicationDom.attachGlassPane(glassPaneElement);
```
yjbanov, line 161: Here too, I think `ApplicationDom` can fully own the glass pane element, instead of having something from outside create and set it. Similarly, it should own the initialization of `PointerBinding` and metrics change listener.

I'm OK with making this transition gradually, but I think the final design should consolidate ownership of view resources into the "application DOM" abstraction. Splitting it between the view and the embedder will make things more complex.

lib/web_ui/lib/src/engine/view_embedder/application_dom/application_dom.dart
```diff
@@ -0,0 +1,95 @@
...
+/// like handling hot-restart cleanup.
+///
+/// This class is specialized to handle different types of DOM embeddings:
+///
+/// * [FullPageApplicationDom] - The default behavior, where flutter takes
+///   control of the whole web page. This is how Flutter Web used to operate.
+/// * [CustomElementApplicationDom] - Flutter is rendered inside a custom host
+///   element, provided by the web app programmer through the engine
+///   initialization.
+abstract class ApplicationDom {
```
yjbanov, line 25: Bikeshed time! I'm not sure `ApplicationDom` is the best choice because both "application" and "DOM" are global concepts. I think the name should link more directly to the [FlutterView](https://api.flutter.dev/flutter/dart-ui/FlutterView-class.html) class. Ideas:

* `FlutterWebView`
* `WebEngineView`
* `EngineView` (just because we have a lot of `EngineBlah` classes already)

Implementations can be `FullPageView` and `CustomElementView`.

lib/web_ui/lib/src/engine/view_embedder/application_dom/custom_element_application_dom.dart
```diff
@@ -0,0 +1,104 @@
...
+    StreamController<DomResizeObserverEntry>.broadcast();
+
+  @override
+  final String type = 'custom-element';
```
yjbanov, line 44: From the comment above, I don't think we need this as this class can just set the attribute here. However, if we absolutely need it for something, let's call it `debugType` to stress that this is used for debugging only and has no behavioral effect.

lib/web_ui/lib/src/engine/view_embedder/application_dom/full_page_application_dom.dart
yjbanov: `/engine/**/*.dart` might be exempt? Or maybe the whole web engine is 🤔

lib/web_ui/lib/src/engine/view_embedder/application_dom/application_dom.dart
```diff
@@ -0,0 +1,95 @@
...
+  void attachGlassPane(DomElement glassPaneElement);
+
+  /// Attaches the resourceHost element into the hostElement.
+  void attachResourcesHost(DomElement resourceHost, { DomElement? nextTo });
+
+  /// Register a listener for window resize events
+  void setMetricsChangeHandler(void Function(DomEvent? event) handler);
+
+  /// Register a listener for locale change events.
+  void setLanguageChangeHandler(void Function(DomEvent event) handler);
```
yjbanov, line 74: I think this is way too many methods. Instead, I would just have one: `initialize`. Then each implementation can start with a copy&paste of the original `ViewEmbedder.reset` and customize it to its needs.

lib/web_ui/lib/src/engine/view_embedder/dimensions_provider/custom_element_dimensions_provider.dart
```diff
@@ -0,0 +1,45 @@
...
+
+/// This class provides the real-time dimensions of a "hostElement".
+///
+/// Note: all the measurements returned from this class are potentially
+/// *expensive*, and should be cached as needed. Every call to every method on
```
yjbanov, line 18: A pattern we use for expensive methods is use prefix `compute` instead of `get`.

Do you think we can remove the need for these by replacing with `ResizeObserver` and such?

ditman:
(Force-pushed after rebasing with some conflicts)

lib/web_ui/lib/src/engine/view_embedder/dimensions_provider/custom_element_dimensions_provider.dart
ditman: Right now we're only listening to the "resize" events to THEN go and `compute` the physical size of the host element (this is inherited from the visualViewport approach).

ResizeObserver, however, returns the dimensions rect of the observer, which we could cache and use the next time that computePhysicalSize happens. Even when there's no resize, we get an immediate event with the size of the element that we're observing as the observer connects.

So for custom elements: yes, we could use only a ResizeObserver and cache the last metric we've seen for each observed element as its "physicalSize".

I suspect that for the `visualViewport`, we can't use this pattern, because the VisualViewport object is an EventTarget (and not an HtmlElement).

lib/web_ui/lib/src/engine/view_embedder/application_dom/application_dom.dart
ditman: What about `GlassPaneHost`? I was thinking that this class would only own the injection of the GlassPane element into the DOM (and whatever config is required for it to work, like the CSS tweaks and all that jazz).

lib/web_ui/lib/src/engine/embedder.dart
ditman: For now, this is the line that I wanted to keep:

* ApplicationDOM knows *where* the GlassPane ends up in the page. This needs to know the hostElement and behaves differently.
* ViewEmbedder knows *what* the GlassPane made of, and how to make it. This doesn't care about the hostElement.

If your idea is to keep moving logic from the ViewEmbedder to the ApplicationDOM, shouldn't we just create a `FullPageViewEmbedder` and a `CustomElementViewEmbedder` and take the plunge now?

lib/web_ui/lib/src/engine/embedder.dart
ditman: I'll take a look at the API surface and simplify the public bits. We may only need a "prepareHostElement" method and some incoming configuration for it to work.

lib/web_ui/lib/src/engine/embedder.dart
ditman: > I'm not sure there's value in having them be set from outside.

I was trying to rely less in global variables (like the buildMode or the type of renderer), and more in incoming parameters.

The ViewEmbedder already uses these globals, but there's no need for those globals to spread to more places than strictly needed :P

lib/web_ui/lib/src/engine/embedder.dart
ditman: They're global to the flutter application, but not global to the users' web application in which we're being embedded.

Our goal should be to *not* modify anything outside of the hostElement, because we don't know how the users' app is going to react to that.

lib/web_ui/lib/src/engine/view_embedder/application_dom/application_dom.dart
ditman: Yeah, at first I extracted a "semantic" API for my own mental mapping, but this can totally be a private method that is used from the public "initialize". I'll simplify the public API surface of the ApplicationDOM classes.

lib/web_ui/lib/src/engine/view_embedder/application_dom/custom_element_application_dom.dart
ditman: Damn, my energy to find a good name for a variable must have ran out after writing the ResizeObserver js-interop :P

lib/web_ui/lib/src/engine/view_embedder/application_dom/custom_element_application_dom.dart
ditman: I think I'm going to do both, actually :P

lib/web_ui/lib/src/engine/embedder.dart
```diff
@@ -35,33 +35,36 @@ import 'window.dart';
...
-    }());
+  /// Creates a FlutterViewEmbedder.
+  ///
+  /// The incoming [hostElement] parameter specifies the root element in the DOM
+  /// into which Flutter will be rendered.
+  ///
+  /// The hostElement is abstracted by an [ApplicationDom] instance, which has
+  /// different behavior depending on the `hostElement` value:
+  ///
+  /// - A `null` `hostElement` will preserve Flutter web's original behavior, where
```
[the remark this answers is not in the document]
ditman: That makes a lot of sense, also we can come up with a better "full screen" hosting in the future, that looks the same as the old one but has a completely different implementation.

lib/web_ui/lib/src/engine/view_embedder/application_dom/custom_element_application_dom.dart
ditman: I also moved this to the DimensionsProvider, it makes sense that the same object that helps me compute the size is the one that also tells me that the Size has changed!

lib/web_ui/lib/src/engine/embedder.dart
ditman: Removed this, I'm still passing values in from the embedder, but as a "metadata" map.

lib/web_ui/lib/src/engine/view_embedder/application_dom/custom_element_application_dom.dart
ditman: Removed!

lib/web_ui/lib/src/engine/view_embedder/application_dom/application_dom.dart
ditman: Removed the metricsChangeHandler to a Stream coming from the DimensionsProvider, exposed through the EngineWindow!

ditman: I still don't know what to do with the languageChangeHandler. I could listen to mutations to the attribute "lang" of the hostElement, BUUUT I'm not sure anybody will want that behavior :)

lib/web_ui/lib/src/engine/view_embedder/application_dom/custom_element_application_dom.dart
ditman: TL;DR: Asserting _hostElement.children.isEmpty here breaks hot restart :)

lib/web_ui/lib/src/engine/embedder.dart
yjbanov: True, but for things that do not interact with anything else (like specifically these two attributes), I wouldn't be too worried. We already ask the developer to add `<script src="flutter.js"></>` globally, so we can't avoid altering global scope completely.

Having said that, my main concern is about these expanding the API surface of `ApplicationDom`. We can still keep these attributes local to the glasspane by having `ApplicationDom` set them on itself.

lib/web_ui/lib/src/engine/embedder.dart
yjbanov: Ah, I see where you are going with this. I guess in my mind `ViewEmbedder` is still a global singleton, which manages multiple views. In contrast, `ApplicationDom` is not a singleton and is agnostic of embedding strategy (full-page, custom element, multiple custom elements). So all my suggestions revolve around moving any view-specific initialization logic from single scope to per-view scope.

If the idea is that `ViewEmbedder` ceases to be singleton, then it might scale. However:

> shouldn't we just create a FullPageViewEmbedder and a CustomElementViewEmbedder and take the plunge now?

If this is not boiling too many oceans, then it looks like the cleanest solutions. I still think we need a root-level object for managing all the views to avoid top-level statics. To me `ViewEmbedder` looks like a good candidate for that, with `EngineView`, or some such, being "managed objects".

lib/web_ui/lib/src/engine/host_node.dart
yjbanov: I think with Stadia gone, we no longer have customers for shadow-DOMless mode.

lib/web_ui/lib/src/engine/view_embedder/application_dom/application_dom.dart
yjbanov: I know this ship may have sailed by now, but "glasspane" is kind of a misnomer. I would avoid propagating this name. It used to refer to the element sitting on top of the canvas that captures pointer events (like a glasspane on a picture frame), but now it has a different role entirely even though it still bears the old name. I would still go with "view" somewhere in the name because it's exactly what each element is for.

lib/web_ui/lib/src/engine/view_embedder/application_dom/application_dom.dart
yjbanov: Language is expose through `SingleFlutterWindow` only. `FlutterView` doesn't have it. So I think this is an inherently global property and should live outside `ApplicationDom`.

https://api.flutter.dev/flutter/dart-ui/SingletonFlutterWindow-class.html
https://api.flutter.dev/flutter/dart-ui/FlutterView-class.html

lib/web_ui/lib/src/engine/view_embedder/application_dom/application_dom.dart
yjbanov: If/when we get the chance, we should rename `<flt-glass-pane>` to `<flutter-view>`.

lib/web_ui/lib/src/engine/host_node.dart
ditman: I thought the same too, but I wasn't sure that the number of customers for this would be exactly zero.

lib/web_ui/lib/src/engine/view_embedder/application_dom/application_dom.dart
ditman: It makes sense, I removed some of the comments about the position of the glasspane in the sandwich of the flutter app, which were no longer true. I just think that the word "View" is almost meaningless (too many uses) in our codebase :P

lib/web_ui/lib/src/engine/view_embedder/application_dom/application_dom.dart
ditman: I shoehorned it here because I was attempting to reuse the same `onHotRestart` handler, but since we can have a bunch of those in the engine, it can be self contained and go elsewhere. I'll refactor it elsewhere; and that'll let me simplify the code of the Cache class so it only handles the DOM elements that we want to delete after hot restart.

lib/web_ui/lib/src/engine/embedder.dart
ditman: > my main concern is about these expanding the API surface of `ApplicationDom`.

I figured this was the case, and moved them to be private for each implementation and into their init methods, without being part of the core API. The *values* that are applied to the DOM are passed from the outside, however.

lib/web_ui/lib/src/engine/embedder.dart
ditman: "If this is not boiling too many oceans" -> I took a look at the code, and this is *definitely* boiling too many oceans.

Form what I've seen (and I might be wrong, because I've only read code and not really *traced* most of it running:

#### Rendering:

* The EnginePlatformDispatcher knows how to handle [multiple views](https://github.com/flutter/engine/blob/7e03ebcbee0dfb4d361618deeeb3d4595156ca65/lib/web_ui/lib/src/engine/window.dart#L48-L51) and their configurations; however that is hardcoded to a single `engine.window`. Also, the view configuration is never really set to anything other than the default values.
    * Views have access to the PlatformDispatcher. The PlatformDispatcher can [`render` a `Scene`](https://github.com/flutter/engine/blob/7e03ebcbee0dfb4d361618deeeb3d4595156ca65/lib/web_ui/lib/src/engine/platform_dispatcher.dart#L661-L664) into its `View` through a global `renderer` object, but we never use the incoming FlutterView to render the scene.
        * The renderer knows about the global `embedder` and its [`addSceneToSceneHost` method](https://github.com/flutter/engine/blob/7e03ebcbee0dfb4d361618deeeb3d4595156ca65/lib/web_ui/lib/src/engine/canvaskit/renderer.dart#L75) that glues this all together, however this bit is definitely not multi-view compliant, because the embedder controls the embedding strategy.
* The Views have a "peculiar" inheritance. For all intents and purposes, on the web, we only care about the `EngineFlutterWindow` class (there's a `EngineSingletonFlutterWindow` that exposes the DPI of the device, but that should probably end up disappearing)
    * The default `engine.view` (of type `EngineSingletonFlutterWindow`) is returned from`ui.view`.

If there's interest in the multi-view project, I think I'd start by fixing how the renderers control the targets of their scenes. Instead, the renderer should paint through `view.viewConfiguration.embeddingStrategy.renderScene(scene)` (or `getEmbeddingStrategy(view).renderScene(scene);`).

> I think the final design should consolidate ownership of view resources into the "application DOM" abstraction. Splitting it between the view and the embedder will make things more complex.

I think the centralization of things needs to happen in the web specializations of the "View" classes, that already exist (the URL Strategy is already there; the Embedding Strategy should probably live inside each View too).

The current EmbeddingStrategy objects can be maintained as they are (they're small and pretty much single-purpose), but you're right that the global `view_embedder` has its days numbered :P

PS: It **does** makes sense in the above world that each EmbeddingStrategy owns its own glassPane element. I'll add it there, and for now, have the Singleton Embedder consult its Singleton EmbeddingStrategy for the glassPaneElement to maintain its API.

Takeaway: I'm renaming `ApplicationDom` to `EmbeddingStrategy` :)

lib/web_ui/lib/src/engine/view_embedder/application_dom/application_dom.dart
ditman: > We should rename `<flt-glass-pane>` to `<flutter-view>`

It's a [trivial change](https://github.com/flutter/engine/blob/7e03ebcbee0dfb4d361618deeeb3d4595156ca65/lib/web_ui/lib/src/engine/embedder.dart#L50-L51), however it has the potential of [breaking the internet](https://github.com/search?q=%22flt-glass-pane%22&type=code)?

lib/web_ui/lib/src/engine/host_node.dart
ditman: I created this issue: https://github.com/flutter/flutter/issues/116204 so this can be tackled later.

ditman:
The code is breaking ~18 tests mostly because uninitialized `late` variables now. I'll take a look too.

lib/web_ui/lib/src/engine/embedder.dart
ditman: I backpedaled on the EmbeddingStrategy owning the glassPane element. I think there's still value in having a standard way of creating a Flutter App regardless of where it is injected. For now I left it in the View Embedder, but we should have a class that "knows how to create and assemble a flutter app" in a unified way.

lib/web_ui/lib/src/engine/view_embedder/application_dom/application_dom.dart
ditman: I moved this to the EnginePlatformDispatcher, near other global configuration coming from the browser, like the brightness, high contrast, and font size.

It is also now self contained within the dispatcher. The old implementation was only calling platformDispatcher methods in a very roundabout way.

ditman:
I think I've resolved all the issues that were discussed (here and on a quick meeting), and added unit tests for the new code. PTAL @yjbanov!

ditman:
This currently has a bug while computing coordinates on top of a platform view, they get reset to 0,0 over the platform view, rather than continuing on the app.

This is easy to solve in a "flat" app, but not trivial when the app has had a transform applied to it. Working on a fix.

lib/web_ui/lib/src/engine/view_embedder/embedding_strategy/embedding_strategy.dart
```diff
@@ -0,0 +1,68 @@
...
+    Map<String, String>? embedderMetadata,
+  });
+
+  /// Attaches the glassPane element into the hostElement.
+  void attachGlassPane(DomElement glassPaneElement);
+
+  /// Attaches the resourceHost element into the hostElement.
+  void attachResourcesHost(DomElement resourceHost, {DomElement? nextTo});
+
+  /// A callback that runs when hot restart is triggered.
```
yjbanov, line 55: nit: s/when/just before/

lib/web_ui/lib/src/engine/view_embedder/embedding_strategy/full_page_embedding_strategy.dart
```diff
@@ -0,0 +1,106 @@
...
+
+  @override
+  void attachGlassPane(DomElement glassPaneElement) {
+    /// Tweaks style so the glassPane works well with the hostElement.
+    glassPaneElement.style
+      ..position = 'absolute'
+      ..top = '0'
+      ..right = '0'
+      ..bottom = '0'
+      ..left = '0';
```
yjbanov, line 33: Ah, this is where I got that intuition from 😆

lib/web_ui/lib/src/engine/view_embedder/hot_restart_cache_handler.dart
```diff
@@ -0,0 +1,52 @@
...
+/// Handles elements that need to be cleared after a hot-restart.
+class HotRestartCacheHandler {
+  HotRestartCacheHandler() {
+    if (_elements.isNotEmpty) {
+      // We are in a post hot-restart world, clear the elements now.
+      _clearAllElements();
+    }
+  }
+
+  /// This is state persistent across hot restarts that indicates what
```
yjbanov, line 19: nit: s/is state/state is/

lib/web_ui/lib/src/engine/view_embedder/hot_restart_cache_handler.dart
```diff
@@ -0,0 +1,52 @@
...
+  /// to clear.  Delay removal of old visible state to make the
+  /// transition appear smooth.
```
yjbanov, line 21: I feel like the second sentence belongs somewhere else, since a "name" can't possibly delay anything.

Or perhaps we can reword this something like: "This variable stores state clean-up callbacks that will be called after the hot restart is complete. This makes the restart appear smooth because the old UI continues to be visible during the restart."

lib/web_ui/lib/src/engine/window.dart
```diff
@@ -305,26 +273,16 @@ class EngineFlutterWindow extends ui.SingletonFlutterWindow {
...
-      width = domWindow.visualViewport!.width! * devicePixelRatio;
-    } else {
-      height = domWindow.innerHeight! * devicePixelRatio;
-      width = domWindow.innerWidth! * devicePixelRatio;
-    }
-
     // This method compares the new dimensions with the previous ones.
     // Return false if the previous dimensions are not set.
     if (_physicalSize != null) {
+      final ui.Size current = _dimensionsProvider.computePhysicalSize();
```
yjbanov, line 279: Can you please leave a TODO linking to https://github.com/flutter/flutter/issues/117036. I think we can make this better.

lib/web_ui/lib/src/engine/embedder.dart
```diff
@@ -34,34 +33,45 @@ import 'window.dart';
...
+  ///
+  /// - A `null` `hostElement` will allow Flutter to take over the whole screen.
+  /// - A non-`null` `hostElement` will render flutter inside that element.
+  FlutterViewEmbedder({DomElement? hostElement}) {
+    // Create an appropriate EmbeddingStrategy using its factory...
+    _embeddingStrategy = EmbeddingStrategy.create(hostElement: hostElement);
+
+    // Configure the EngineWindow that this embedder uses, so it knows how to
+    // measure itself.
+    window.configureDimensionsProvider(DimensionsProvider.create(
```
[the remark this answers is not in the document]
ditman: I think this will eventually morph in the embedding strategy having a getter for an appropriate dimensions provider.

lib/web_ui/lib/src/engine/view_embedder/embedding_strategy/custom_element_embedding_strategy.dart
```diff
@@ -0,0 +1,54 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:ui/src/engine/dom.dart';
+
+import 'embedding_strategy.dart';
+
+class CustomElementEmbeddingStrategy extends EmbeddingStrategy {
+  CustomElementEmbeddingStrategy(this._hostElement) {
+    // Clear children...
```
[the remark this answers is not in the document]
ditman: This is the type of quality content I'm here for :P

lib/web_ui/lib/src/engine/view_embedder/embedding_strategy/custom_element_embedding_strategy.dart
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  @override
+  void attachGlassPane(DomElement glassPaneElement) {
+    glassPaneElement
+      ..style.width = '100%'
+      ..style.height = '100%'
+      ..style.display = 'block'
+      ..style.overflow = 'hidden'
+      ..style.position = 'relative';
```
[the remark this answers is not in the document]
ditman: > Could there be a difference?

Yes, if we set `position: absolute` here, we're forcing users to make the container they provide `position: relative`, which they may (or may not) want to do.

If we make this `position: relative`, everything inside will Work As Expected, and the user shouldn't need to do anything extra for their flutter app to work.

lib/web_ui/lib/src/engine/view_embedder/application_dom/application_dom.dart
ditman: Will be followed up here: https://github.com/flutter/flutter/issues/117034

lib/web_ui/lib/src/engine/dom.dart
```diff
@@ -1426,6 +1436,77 @@ extension DomCSSRuleListExtension on DomCSSRuleList {
   external double get length;
 }
 
+/// ResizeObserver constructor.
```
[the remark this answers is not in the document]
ditman: it's a JS binding :P

lib/web_ui/lib/src/engine/embedder.dart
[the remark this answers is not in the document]
ditman: Created this: https://github.com/flutter/flutter/issues/117098

lib/web_ui/lib/src/engine/embedder.dart
```diff
@@ -34,34 +33,45 @@ import 'window.dart';
...
+      hostElement: hostElement,
+    ));
+
     reset();
+
     assert(() {
-      _registerHotRestartCleanUp();
+      // _embeddingStrategy needs to clean-up stuff in the page on hot restart.
+      registerHotRestartListener(_embeddingStrategy.onHotRestart);
```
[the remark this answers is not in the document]
ditman: Let me revisit. This started life as a direct "port" of what the old code did, but you're right, the embedding strategy should register its own cleanup.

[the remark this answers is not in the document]
ditman: Moved to the embedding strategy.

lib/web_ui/lib/src/engine/view_embedder/embedding_strategy/full_page_embedding_strategy.dart
```diff
@@ -0,0 +1,106 @@
...
+    setElementStyle(bodyElement, 'top', '0');
+    setElementStyle(bodyElement, 'right', '0');
+    setElementStyle(bodyElement, 'bottom', '0');
+    setElementStyle(bodyElement, 'left', '0');
+    setElementStyle(bodyElement, 'overflow', 'hidden');
+    setElementStyle(bodyElement, 'padding', '0');
+    setElementStyle(bodyElement, 'margin', '0');
+
+    // TODO(yjbanov): fix this when KVM I/O support is added. Currently scroll
+    //                using drag, and text selection interferes.
```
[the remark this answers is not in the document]
ditman: Removed.

lib/web_ui/lib/src/engine/view_embedder/dimensions_provider/custom_element_dimensions_provider.dart
```diff
@@ -0,0 +1,86 @@
...
+
+import 'package:ui/src/engine/dom.dart';
+import 'package:ui/src/engine/window.dart';
+import 'package:ui/ui.dart' as ui show Size;
+
+import 'dimensions_provider.dart';
+
+/// This class provides the real-time dimensions of a "hostElement".
+///
+/// Note: all the measurements returned from this class are potentially
```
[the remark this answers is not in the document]
ditman: Removed.

lib/web_ui/lib/src/engine/view_embedder/dimensions_provider/custom_element_dimensions_provider.dart
```diff
@@ -0,0 +1,86 @@
...
+
+    _hostElementResizeObserver?.observe(_hostElement);
+  }
+
+  final DomElement _hostElement;
+
+  // Handle resize events
+  late DomResizeObserver? _hostElementResizeObserver;
+  final StreamController<ui.Size> _onResizeStreamController =
+      StreamController<ui.Size>.broadcast();
```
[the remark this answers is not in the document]
ditman: We don't use the streamController to size the app initially for now, so the answer is "no". (I think?)

lib/web_ui/lib/src/engine/view_embedder/dimensions_provider/full_page_dimensions_provider.dart
```diff
@@ -0,0 +1,112 @@
...
+import 'package:ui/src/engine/browser_detection.dart';
+import 'package:ui/src/engine/dom.dart';
+import 'package:ui/src/engine/window.dart';
+import 'package:ui/ui.dart' as ui show Size;
+
+import 'dimensions_provider.dart';
+
+/// This class provides the real-time dimensions of a "full page" viewport.
+///
+/// Note: all the measurements returned from this class are potentially
```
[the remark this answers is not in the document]
ditman: Gone.

lib/web_ui/lib/src/engine/view_embedder/dimensions_provider/full_page_dimensions_provider.dart
```diff
@@ -0,0 +1,112 @@
...
+/// *expensive*, and should be cached as needed. Every call to every method on
+/// this class WILL perform actual DOM measurements.
+class FullPageDimensionsProvider extends DimensionsProvider {
+  FullPageDimensionsProvider() {
+    // Determine what 'resize' event we'll be listening to.
+    // This is needed for older browsers (Firefox < 91, Safari < 13)
+    final DomEventTarget resizeEventTarget =
+        domWindow.visualViewport ?? domWindow;
```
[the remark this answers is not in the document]
ditman: Good point, I'll create a cleanup task to remove the "visualViewport is null" case.

lib/web_ui/lib/src/engine/view_embedder/dimensions_provider/dimensions_provider.dart
```diff
@@ -0,0 +1,56 @@
...
+    }
+  }
+
+  /// Returns the DPI reported by the browser.
+  double getDevicePixelRatio() {
+    // This is overridable in tests.
+    return window.devicePixelRatio;
+  }
+
+  /// Returns the [ui.Size] of the "viewport".
```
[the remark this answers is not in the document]
ditman: Added, alongside a note of how very expensive this class in general is at the top (same as for the specific implementations)

lib/web_ui/lib/src/engine/view_embedder/embedding_strategy/custom_element_embedding_strategy.dart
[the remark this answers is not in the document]
ditman: I even removed the comment, `_hostElement.clearChildren()` looks self-documenting enough to me.

lib/web_ui/lib/src/engine/view_embedder/embedding_strategy/custom_element_embedding_strategy.dart
```diff
@@ -0,0 +1,54 @@
...
+    while (_hostElement.firstChild != null) {
+      _hostElement.removeChild(_hostElement.lastChild!);
+    }
+  }
+
+  final DomElement _hostElement;
+
+  @override
+  void initialize({
+    Map<String, String>? embedderMetadata,
```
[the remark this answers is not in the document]
ditman: I'm going to call this "hostElementAttributes", which is what they are. Not all of them are debug info though, we have the one to disable auto-correct, for example :)

lib/web_ui/lib/src/engine/view_embedder/embedding_strategy/custom_element_embedding_strategy.dart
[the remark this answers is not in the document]
ditman: I'm curious about setting this through the inset rather than width/height. It might affect the div not having browser layout or something, but I'm giving it a shot anyways!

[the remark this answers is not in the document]
ditman: Ah, `display: block` is required to ensure the glass-pane covers the whole app (and mouse events work), otherwise by default it's `display: inline`:

Screen Shot 2022-12-14 at 4 31 12 PM

`inset: 0px` needs `position: absolute` to work as intended, otherwise it's not enough to completely size the element  (but it has the problem I described above).

lib/web_ui/lib/src/engine/view_embedder/embedding_strategy/embedding_strategy.dart
```diff
@@ -0,0 +1,68 @@
...
+/// This class is specialized to handle different types of DOM embeddings:
+///
+/// * [FullPageEmbeddingStrategy] - The default behavior, where flutter takes
+///   control of the whole web page. This is how Flutter Web used to operate.
+/// * [CustomElementEmbeddingStrategy] - Flutter is rendered inside a custom host
+///   element, provided by the web app programmer through the engine
+///   initialization.
+abstract class EmbeddingStrategy {
+  EmbeddingStrategy() {
+    // Prepare some global stuff...
```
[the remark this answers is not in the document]
ditman: "Initialize code to handle hot-restart (debug only)."

lib/web_ui/lib/src/engine/view_embedder/embedding_strategy/full_page_embedding_strategy.dart
ditman: This only works with `position: absolute`.

Eventually we should make both implementations converge, and make the difference between full screen and single element just a little bit of extra CSS :)

lib/web_ui/lib/src/engine/view_embedder/hot_restart_cache_handler.dart
ditman: Yes, I rewrote the docs of this class so they make more sense.

lib/web_ui/lib/src/engine/embedder.dart
ditman: Here's the future: https://github.com/flutter/flutter/issues/117098

lib/web_ui/test/engine/host_node_test.dart
```diff
@@ -33,30 +33,90 @@ void testMain() {
...
+      // Check that style.sheet! contains input::-ms-reveal rule
+      final bool hidesRevealIcons = hasCssRule(style,
+          selector: 'input::-ms-reveal', declaration: 'display: none');
+
+      final bool codeRanInFakeyBrowser = hasCssRule(style,
+          selector: 'input.fallback-for-fakey-browser-in-ci',
+          declaration: 'display: none');
+
+      if (codeRanInFakeyBrowser) {
+        print('Please, fix https://github.com/flutter/flutter/issues/116302');
```
htoor3, line 96: Nit: I think this issue relates to running Edge tests locally, not necessarily adding Edge support to our CI tests.  Should I create a new issue to track that?

ditman: Yes please, if you feel that this needs a new issue, do it. If you end up creating a new issue, let me know, and I'll update the error message :)

lib/web_ui/lib/src/engine/pointer_binding.dart
```diff
@@ -1080,7 +1123,7 @@ class _MouseAdapter extends _BaseAdapter with _WheelEventListenerMixin {
       _callback(pointerData);
     });
 
-    _addMouseEventListener(domWindow, 'mousemove', (DomMouseEvent event) {
+    _addMouseEventListener(glassPaneElement, 'mousemove', (DomMouseEvent event) {
```
mdebbar, line 1126: It's explained here why this needs to be attached to `window`:
https://github.com/flutter/engine/blob/13ae6eb75ab97947b2bcd507bf45ee33a1d2033d/lib/web_ui/lib/src/engine/pointer_binding.dart#L283-L286

Changing it to listen on `glassPaneElement` will break some use cases.

lib/web_ui/lib/src/engine/view_embedder/dimensions_provider/dimensions_provider.dart
```diff
@@ -0,0 +1,62 @@
...
+  WindowPadding computeKeyboardInsets(
+    double physicalHeight,
+    bool isEditingOnMobile,
+  );
+
+  /// Returns a Stream with the changes to [ui.Size] (when cheap to get).
+  Stream<ui.Size?> get onResize;
+
+  /// Clears all the resources grabbed by the DimensionsProvider instance.
+  void onHotRestart();
```
mdebbar, line 61: nit: this could be useful for tests too, so a more general name might be better? e.g. `reset()`.

mdebbar (commented):
This looks great! So many tests 😍

Most of my comments are nit picks. The only serious one is about the target of the mousemove listener.

lib/web_ui/lib/src/engine/embedder.dart
```diff
@@ -180,106 +133,42 @@ class FlutterViewEmbedder {
...
+        ? 'auto-selected'
+        : 'requested explicitly';
+
+    // Initializes the embeddingStrategy so it can host a single-view Flutter app.
+    _embeddingStrategy.initialize(
+      hostElementAttributes: <String, String>{
+        'flt-renderer': '${renderer.rendererTag} ($rendererSelection)',
+        'flt-build-mode': buildMode,
+        // TODO(mdebbar): Disable spellcheck until changes in the framework and
+        // engine are complete.
```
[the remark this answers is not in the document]
ditman: It's staying then :P

lib/web_ui/lib/src/engine/host_node.dart
ditman: Yeah, I think this ended up fixing that issue. Added to the description, thanks for the link!!

lib/web_ui/lib/src/engine/pointer_binding.dart
```diff
@@ -834,6 +872,10 @@ class _PointerAdapter extends _BaseAdapter with _WheelEventListenerMixin {
         return coalescedEvents;
       }
     }
+    // Important: coalesced events lack the `eventTarget` property (because they're
+    // being handled in a deferred way).
+    //
+    // See the "Note" here: https://developer.mozilla.org/en-US/docs/Web/API/Event/currentTarget
```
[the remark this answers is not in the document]
ditman: It does, but it doesn't matter much, we pretty much only care about the `glassPaneElement` :)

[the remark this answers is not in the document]
ditman: (Also unclear about the benefits of `getCoalescedEvents`  :P)

lib/web_ui/lib/src/engine/pointer_binding.dart
ditman: AFAIK the MouseAdapter is only being used by Safari 12 and older. Also I'm struggling to understand the comment here.

Is this comment saying that the browser is FIRING `mousemove` events when the cursor is OUTSIDE of `window`? I don't believe this comment one bit :P

Anyway, can revert no problem :)

lib/web_ui/lib/src/engine/pointer_binding.dart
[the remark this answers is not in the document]
mdebbar: The browser batches multiple events and fires them together. `getCoalescedEvents` gives you the list of batched events.

I'm assuming browsers do this for performance reasons.

lib/web_ui/lib/src/engine/embedder.dart
```diff
@@ -34,34 +33,38 @@ import 'window.dart';
...
-  DomSubscription? _resizeSubscription;
-
-  /// Listens to window locale events.
-  DomSubscription? _localeSubscription;
-
-  /// Contains Flutter-specific CSS rules, such as default margins and
-  /// paddings.
-  DomHTMLStyleElement? _styleElement;
+  /// Abstracts all the DOM manipulations required to embed a Flutter app in an user-supplied `hostElement`.
+  late EmbeddingStrategy _embeddingStrategy;
```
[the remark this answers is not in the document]
ditman: Good call, I prefer this to being too precious with `late` afterwards.

lib/web_ui/lib/src/engine/pointer_binding.dart
mdebbar: I think when you listen on window, you receive some events happening outside the window (e.g. when you start dragging inside the window, then move the mouse outside).

You lose that when you listen on an element.

But my memory is vague here, so I could be wrong.

mdebbar: Since this only affects the mouse adapter (that's going away soon), I'm not worried about the effects of this change anymore. So feel free to leave it attached to glassPaneElement.

lib/web_ui/lib/src/engine/dom.dart
```diff
@@ -1313,6 +1314,10 @@ class DomCSSStyleSheet extends DomStyleSheet {}
 
 extension DomCSSStyleSheetExtension on DomCSSStyleSheet {
   external DomCSSRuleList get cssRules;
+  Iterable<DomCSSRule> get rules =>
+      createDomListWrapper<DomCSSRule>(js_util
+          .getProperty<_DomList>(this, 'cssRules'));
```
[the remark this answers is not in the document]
ditman: I went back and forth on this one, and I don't know why I didn't do it (I guess there's a method that calls .length on this repeatedly). I don't think this will negatively impact performance though, so done.

lib/web_ui/lib/src/engine/host_node.dart
```diff
@@ -13,7 +14,25 @@ import 'text_editing/text_editing.dart';
...
+/// HostNodes; for ShadowDOM there's not much to do, but for ElementNodes, the
+/// stylesheet is "namespaced" by the `flt-glass-pane` prefix, so it "only"
+/// affects things that Flutter web owns.
 abstract class HostNode {
+  /// Returns an appropriate HostNode for the given [root].
+  ///
+  /// If `attachShadow` is supported, this returns a [ShadowDomHostNode], else
+  /// this will fall-back to an [ElementHostNode].
+  factory HostNode.create(DomElement root, String defaultFont) {
+    if (getJsProperty<Object?>(root, 'attachShadow') != null) {
```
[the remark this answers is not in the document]
ditman: I _think_ that syntax is Bad For Dart2Wasm, [look at this change](https://github.com/flutter/engine/commit/3aa3d2a8f23da96274939b7e494d29be4776d2df#diff-85b3bd302d38fbd6b511f9d2636fee56d38b4969e9b8ced1f144db161e720112L26-R34)

Not exactly sure if it's because we're calling an external function something that we "get" (just to check if it exists) or what. This code already exists in the engine as is, I'm just moving it from the embedder to the host_node file :/

lib/web_ui/lib/src/engine/pointer_binding.dart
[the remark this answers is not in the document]
ditman: I know that, I just haven't seen any coalescedEvent with > 1 event, I need a worse computer :P

lib/web_ui/lib/src/engine/pointer_binding.dart
ditman: You're right, this is shocking!
https://jsfiddle.net/ditman/7towxaqp

ditman: I've reverted the code so the event is attached to the window so we keep seeing those events outside of the window. This might be weird if we want to drag-and-drop across two flutter views at some point :/

lib/web_ui/lib/src/engine/view_embedder/dimensions_provider/dimensions_provider.dart
ditman: I called it `close()`, because this is very similar to the Stream behavior of "no more events are going to be dispatched after close". To me `reset` implies "close, then reopen". I also improved the blurb of documentation a little bit explaining this behavior. Good call!

lib/web_ui/lib/src/engine/host_node.dart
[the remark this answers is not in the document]
ditman: Confirmed with the dart2wasm folk that this might be a problematic change, so skipping this one.

Sorry Mouad, but thanks for taking the time to point this improvement out!

mdebbar (approved):
Thanks for addressing all my nit picks!

htoor3 (approved):

yjbanov:
nit: this PR should probably refer to the quintessential https://github.com/flutter/flutter/issues/32329 :)

ditman:
> nit: this PR should probably refer to the quintessential [flutter/flutter#32329](https://github.com/flutter/flutter/issues/32329) :)

@yjbanov totally forgot about that one. Added to the description so it gets closed by this.

ditman:
@yjbanov discovered an issue with TalkBack that *seems* to not be reporting the correct offsetX/offsetY (always 0,0) to the framework.

ditman:
The issue with TalkBack seems to be resolved. Rebasing branch.

lib/web_ui/lib/src/engine/pointer_binding/event_position_helper.dart
```diff
@@ -0,0 +1,113 @@
...
+/// to what the DOM would return if we had currentTarget readily available.
+///
+/// This needs an `actualTarget`, because the `event.currentTarget` (which is what
+/// this would really need to use) gets lost when the `event` comes from a "coalesced"
+/// event.
+///
+/// It also takes into account semantics being enabled to fix the case where
+/// offsetX, offsetY == 0 (TalkBack events).
+//
+// TODO(dit): Make this understand 3D transforms in the platform view case, https://github.com/flutter/flutter/issues/117091
```
yjbanov, line 22: Should we also teach it to understand 3D transforms in the semantics tree? (not that people use this often, but for completeness)

lib/web_ui/lib/src/engine/pointer_binding/event_position_helper.dart
```diff
@@ -0,0 +1,113 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
```
yjbanov, line 1: Do logics in this file work if the host page has a non-zero scroll offset? I think scroll offset doesn't affect clientX/Y, but it will affect the position of the cursor relative to the elements.

yjbanov (approved):
lgtm

lib/web_ui/lib/src/engine/pointer_binding/event_position_helper.dart
ditman: Ah yes, I should have removed the "in the platform view case" bit of text when I moved this from the other file. I think we need to solve the 3D case for all the use cases where we cannot use directly offsetX/offsetY (that comes pre-computed by the browser)

lib/web_ui/lib/src/engine/pointer_binding/event_position_helper.dart
ditman: Yes, methods that only use offsetX/Y are relative to the element, so the scroll position does not matter.

The one that requires scroll handling is the pageX/Y one, because that gives us positions from the top left position of the page (with all its scroll position added)

lib/web_ui/lib/src/engine/pointer_binding/event_position_helper.dart
ditman: I've moved the TODOs to a better spot.

lib/web_ui/lib/src/engine/html/scene.dart
```diff
@@ -45,12 +46,10 @@ class PersistedScene extends PersistedContainerSurface {
...
     // The scene clip is the size of the entire window.
-    // TODO(yjbanov): in the add2app scenario where we might be hosted inside
-    //                a custom element, this will be different. We will need to
-    //                update this code when we add add2app support.
-    final double screenWidth = domWindow.innerWidth!;
-    final double screenHeight = domWindow.innerHeight!;
-    localClipBounds = ui.Rect.fromLTRB(0, 0, screenWidth, screenHeight);
+    final ui.Size screen = window.physicalSize / window.devicePixelRatio;
+    // Question: why is the above a logical size, rather than a physical size
+    // like everywhere else in the metrics?
```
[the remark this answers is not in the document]
ditman: This broke our golden tests. Scary!

PR #37750, opened 2022-11-18
[Impeller] Add fenced command buffer that waits on-submit

iskakaushik:
this isn't ideal for performance long term but aligns well with what the other backends do. This also lets us remove the dependence on frame numbers for renderpasses setting up for implementing MSAA.

fixes: https://github.com/flutter/flutter/issues/112387


impeller/renderer/backend/vulkan/fenced_command_buffer_vk.cc
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  auto wait = device_.waitForFences(fence.get(), true, UINT64_MAX);
+  if (wait != vk::Result::eSuccess) {
+    VALIDATION_LOG << "Failed to wait for fence: " << vk::to_string(wait);
+    return false;
+  }
+
+  // cleanup all the resources held by the command buffer and its children.
+  deletion_queue_->Flush();
```
iskakaushik, line 90: the above fence can be awaited on a different thread similar to https://github.com/chinmaygarde/pixelpusher/blob/master/source/machine/fence_waiter.h, but that's an optimization for another day :-)

iskakaushik:
cc: @chinmaygarde friendly ping

chinmaygarde (approved):

PR #37753, opened 2022-11-18
Ignore keydown event for tab during IME composition

htoor3:
During IME composition, Tab gets fired twice: once as part of the composition of the character and once without. This leads to unexpected behavior such as skipping a textfield during tab and input mutation. The solution is to ignore the keydown event that's a part of composition.

Fixes https://github.com/flutter/flutter/issues/74813

## Pre-launch Checklist

- [x] I signed the [CLA].


lib/web_ui/lib/src/engine/raw_keyboard.dart
```diff
@@ -96,6 +96,14 @@ class RawKeyboard {
     final FlutterHtmlKeyboardEvent event = FlutterHtmlKeyboardEvent(domEvent as DomKeyboardEvent);
     final String timerKey = event.code!;
 
+    // During IME composition, Tab fires twice (once for composition and once
+    // for regular tabbing behavior), which causes issues. Intercepting the
+    // tab keydown event during composition prevents these issues from occurring.
+    // https://developer.mozilla.org/en-US/docs/Web/API/Element/keydown_event#ignoring_keydown_during_ime_composition
+    if (event.type == 'keydown' && event.key == 'Tab' && event.isComposing) {
+      return;
+    }
```
mdebbar, line 105: I suggest putting this logic in a function:
```suggestion
    if (_shouldIgnore(event)) {
      return;
    }
```
Then add:
```dart
bool _shouldIgnore(FlutterHtmlKeyboardEvent event) {
  // During IME composition, Tab fires twice (once for composition and once
  // for regular tabbing behavior), which causes issues. Intercepting the
  // tab keydown event during composition prevents these issues from occurring.
  // https://developer.mozilla.org/en-US/docs/Web/API/Element/keydown_event#ignoring_keydown_during_ime_composition
  return event.type == 'keydown' && event.key == 'Tab' && event.isComposing;
}
```

lib/web_ui/lib/src/engine/raw_keyboard.dart
```diff
@@ -96,6 +96,14 @@ class RawKeyboard {
     final FlutterHtmlKeyboardEvent event = FlutterHtmlKeyboardEvent(domEvent as DomKeyboardEvent);
     final String timerKey = event.code!;
 
+    // During IME composition, Tab fires twice (once for composition and once
+    // for regular tabbing behavior), which causes issues. Intercepting the
+    // tab keydown event during composition prevents these issues from occurring.
+    // https://developer.mozilla.org/en-US/docs/Web/API/Element/keydown_event#ignoring_keydown_during_ime_composition
+    if (event.type == 'keydown' && event.key == 'Tab' && event.isComposing) {
```
mdebbar, line 103: Do we also need to ignore the corresponding keyup?

mdebbar (commented):
Thanks @htoor3 for fixing this issue!

cc @dkwingsmt in case we need to do something similar for the hardware keyboard.

lib/web_ui/lib/src/engine/raw_keyboard.dart
htoor3: It doesn't seem to affect the behavior whether I ignore it or not.  All of the conditional logic in in the keydown/keyup callback `_handleHtmlEvent` just seems to be handling for when the event type is `keydown`

Should I add the logic to ignore it regardless?

mdebbar: If you tried it with an app, and the framework doesn't complain about it, then it should be fine.

lib/web_ui/lib/src/engine/raw_keyboard.dart
```diff
@@ -88,6 +88,14 @@ class RawKeyboard {
     return _onMacOs;
   }
 
+  // During IME composition, Tab fires twice (once for composition and once
+  // for regular tabbing behavior), which causes issues. Intercepting the
+  // tab keydown event during composition prevents these issues from occurring.
+  // https://developer.mozilla.org/en-US/docs/Web/API/Element/keydown_event#ignoring_keydown_during_ime_composition
```
mdebbar, line 94: This comment describes an implementation detail of the function. I think it should go inside the function. What do you think?

mdebbar (approved):
LGTM!

lib/web_ui/lib/src/engine/raw_keyboard.dart
htoor3: That definitely makes sense, I was just following the pattern of the `_shouldDoKeyGuard()` function where the comment is included right above the function.  I don't have a strong preference either way

PR #37754, opened 2022-11-18
Preliminary implementation of UIA for A11y on Windows

yaakovschectman:
We are now able to incorporate the implementation for UI Automation for accessibility in addition to/in place of MSAA. By default, the UIA object will not be returned to a screen reader in the `WM_GETOBJECT` message. For this, the engine must be built with the `FLUTTER_ENGINE_USE_UIA` macro defined.
- In addition to the change to the window message handler, the Flutter platform node delegate has been modified to return its `HWND` in `GetTargetForNativeAccessibilityEvent` so that the root of the element can be obtained for `IRawElementProviderFragment::get_FragmentRoot`.
- A typo in `AXActivePopup` has been fixed.
- Accessibility event dispatch procedure on Windows has been modified to defer to `AXPlatformNodeWin::NotifyAccessibilityEvent` so that either MSAA and/or UIA events may be fired in response.
- Unit tests have been added/modified/re-enabled.

We will still need to implement `ITextProvider` and `ITextRangeProvider` for their appropriate UIA patterns in order to have text edit carat navigation narrated correctly.

Part of https://github.com/flutter/flutter/issues/114547

## Pre-launch Checklist

- [ ] I signed the [CLA].

### DIFF
--- a/shell/platform/windows/window.cc
+++ b/shell/platform/windows/window.cc
@@ -203,6 +203,15 @@ LRESULT Window::OnGetObject(UINT const message,
   if (is_uia_request && root_view) {
     // TODO(cbracken): https://github.com/flutter/flutter/issues/94782
     // Implement when we adopt UIA support.
+    // Retrieve UIA object for the root view.
+    Microsoft::WRL::ComPtr<IRawElementProviderSimple> root;
+    root_view->QueryInterface(
+        IID_PPV_ARGS(&root));
+
+    // Return the UIA object via UiaReturnRawElementProvider(). See:
+    // https://docs.microsoft.com/en-us/windows/win32/winauto/wm-getobject
+    reference_result =
+        UiaReturnRawElementProvider(window_handle_, wparam, lparam, root.Get());
   } else if (is_msaa_request && root_view) {
     // Create the accessibility root if it does not already exist.
     if (!accessibility_root_) {
@@ -212,10 +221,9 @@ LRESULT Window::OnGetObject(UINT const message,
     // Microsoft::WRL::ComPtr<IAccessible> root(root_view);
     accessibility_root_->SetWindow(root_view);
     Microsoft::WRL::ComPtr<IAccessible> root(accessibility_root_);
-    LRESULT lresult = LresultFromObject(IID_IAccessible, wparam, root.Get());
-    return lresult;
+    reference_result = LresultFromObject(IID_IAccessible, wparam, root.Get());
   }
-  return 0;
+  return reference_result;
 }
 
 void Window::OnImeSetContext(UINT const message,
--- a/shell/platform/windows/window.cc
+++ b/shell/platform/windows/window.cc
@@ -201,6 +201,9 @@ LRESULT Window::OnGetObject(UINT const message,
   }
 
   gfx::NativeViewAccessible root_view = GetNativeViewAccessible();
+  // TODO(schectman): UIA is currently disabled by default.
+  // https://github.com/flutter/flutter/issues/114547
+#ifdef FLUTTER_ENGINE_USE_UIA
   if (is_uia_request && root_view) {
     if (!ax_fragment_root_) {
       if (!ax_fragment_delegate_) {
@@ -222,7 +225,10 @@ LRESULT Window::OnGetObject(UINT const message,
     } else {
       FML_LOG(ERROR) << "Failed to query AX fragment root.";
     }
-  } else if (is_msaa_request && root_view) { // TODO(schectman) revert this FALSE check; it's just for testing
+  } else if (is_msaa_request && root_view) {
+#else
+  if (is_msaa_request && root_view) {
+#endif  // FLUTTER_ENGINE_USE_UIA
     // Create the accessibility root if it does not already exist.
     if (!accessibility_root_) {
       CreateAccessibilityRootNode();
--- a/shell/platform/windows/window.cc
+++ b/shell/platform/windows/window.cc
@@ -203,7 +203,10 @@ LRESULT Window::OnGetObject(UINT const message,
   gfx::NativeViewAccessible root_view = GetNativeViewAccessible();
   if (is_uia_request && root_view) {
     if (!ax_fragment_root_) {
-      ax_fragment_root_ = std::make_unique<ui::AXFragmentRootWin>(window_handle_, this);
+      if (!ax_fragment_delegate_) {
+        ax_fragment_delegate_ = std::make_unique<WindowAXFragmentRootDelegate>(*this);
+      }
+      ax_fragment_root_ = std::make_unique<ui::AXFragmentRootWin>(window_handle_, ax_fragment_delegate_.get());
     }
 
     // TODO(cbracken): https://github.com/flutter/flutter/issues/94782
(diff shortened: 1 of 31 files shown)


shell/platform/windows/window.cc
```diff
@@ -200,10 +201,36 @@ LRESULT Window::OnGetObject(UINT const message,
...
+        ax_fragment_delegate_ =
+            std::make_unique<WindowAXFragmentRootDelegate>(*this);
+      }
+      ax_fragment_root_ = std::make_unique<ui::AXFragmentRootWin>(
+          window_handle_, ax_fragment_delegate_.get());
+    }
+
     // TODO(cbracken): https://github.com/flutter/flutter/issues/94782
     // Implement when we adopt UIA support.
+    // Retrieve UIA object for the root view.
```
loic-sharma, line 219: Comments above can be deleted

shell/platform/windows/window.cc
```diff
@@ -200,10 +201,36 @@ LRESULT Window::OnGetObject(UINT const message,
...
+    Microsoft::WRL::ComPtr<IRawElementProviderSimple> root;
+    if (SUCCEEDED(ax_fragment_root_->GetNativeViewAccessible()->QueryInterface(
+            IID_PPV_ARGS(&root)))) {
```
loic-sharma, line 222: I see that Chrome doesn't have this `SUCCEEDED` check. Just curious, why did you add this? Feel free to keep as is, just asking for knowledge :)

shell/platform/windows/window.cc
```diff
@@ -200,10 +201,36 @@ LRESULT Window::OnGetObject(UINT const message,
   }
 
   gfx::NativeViewAccessible root_view = GetNativeViewAccessible();
+  // TODO(schectman): UIA is currently disabled by default.
+  // https://github.com/flutter/flutter/issues/114547
+#ifdef FLUTTER_ENGINE_USE_UIA
```
loic-sharma, line 206: Could we move this into the `if (is_uia_request && root_view) {` and then keep the `else if (is_msaa_request ... )`?

shell/platform/windows/window.h
```diff
@@ -299,6 +303,33 @@ class Window : public KeyboardManager::WindowDelegate {
...
+
+  // Delegate for Fragment.
+  std::unique_ptr<WindowAXFragmentRootDelegate> ax_fragment_delegate_;
+
+  // Allow WindowAXFragmentRootDelegate to access protected method.
+  friend class WindowAXFragmentRootDelegate;
+};
+
+// A delegate class to the window.
+class WindowAXFragmentRootDelegate : public ui::AXFragmentRootDelegateWin {
```
loic-sharma, line 318: Could we make the Windows accessibility bridge implement `ui::AXFragmentRootDelegateWin` instead? That seems like a better location for this logic.

shell/platform/windows/window.cc
yaakovschectman: I do not know for sure why they do not use the check, but `QueryInterface` can return failure

shell/platform/windows/window.h
yaakovschectman: I will test that. I think `UiaReturnRawElementProvider` performs some copying, so I would just need to check that non-idempotent setup would not be repeated if we do that.

shell/platform/windows/window.cc
loic-sharma: Can we not guarantee that `QueryInterface` will succeed here since we entirely own the logic?

shell/platform/windows/window.h
loic-sharma: > ... but FlutterWindowsEngine stores `AccessibilityBridge` rather than `AccessibilityBridgeWindows`.

This was necessary until a recent refactoring (`AccessibilityBridgeWindows` is new, previously there was only the `AccessibilityBridge`). We can make the engine store a `AccessibilityBridgeWindows` and skip casting magic.

shell/platform/windows/window.cc
yaakovschectman: I don't know how accurate it is to say we entirely own the logic. ATL provides a default `QueryInterface` which calls `InternalQueryInterface`, which we own, which then calls, `CComObjectRootBase::InternalQueryInterface`, which is also a provided ATL implementation.

skia-gold:
Gold has detected about 1 new digest(s) on patchset 16.
View them at https://flutter-engine-gold.skia.org/cl/github/37754

shell/platform/windows/windowsx_shim.h
```diff
@@ -0,0 +1,28 @@
...
+// found in the LICENSE file.
+
+#ifndef BASE_WIN_WINDOWSX_SHIM_H_
+#define BASE_WIN_WINDOWSX_SHIM_H_
+
+// The Win32 platform header <windowsx.h> contains some macros for
+// common function names. To work around that, windowsx.h is not to be
+// included directly, and instead this file should be included. If one
+// of the removed Win32 macros is wanted, use the expanded form
+// manually instead.
```
cbracken, line 12: Thank you for cleaning this up. Much nicer.

shell/platform/windows/window.h
```diff
@@ -20,6 +19,9 @@
 #include "flutter/shell/platform/windows/sequential_id_generator.h"
 #include "flutter/shell/platform/windows/text_input_manager.h"
 #include "flutter/shell/platform/windows/windows_proc_table.h"
+#include "flutter/shell/platform/windows/windowsx_shim.h"
```
loic-sharma, line 22: Can this be moved to the implementation `.cc` file?

loic-sharma (approved):
LGTM, nice work!

shell/platform/windows/window_binding_handler_delegate.h
```diff
@@ -135,6 +136,9 @@ class WindowBindingHandlerDelegate {
 
   // Update the status of the high contrast feature
   virtual void UpdateHighContrastEnabled(bool enabled) = 0;
+
+  // Obtain a pointer to the fragment root delegate.
```
cbracken, line 140: It may be useful to add a little detail on how this is used by UIA, since it's not something that was required in MSAA.

third_party/accessibility/ax/platform/ax_platform_node_win.cc
```diff
@@ -5205,6 +5218,10 @@ std::optional<DWORD> AXPlatformNodeWin::MojoEventToMSAAEvent(
       return EVENT_OBJECT_SHOW;
     case ax::mojom::Event::kValueChanged:
       return EVENT_OBJECT_VALUECHANGE;
+#ifdef EVENT_OBJECT_TEXTSELECTIONCHANGED
```
cbracken, line 5221: I'm assuming this was added in Win 8 or 10 maybe? Consider adding a doc comment pointing out why it's `ifdef`ed.

third_party/accessibility/ax/platform/test_ax_node_wrapper.cc
```diff
@@ -54,9 +54,14 @@ std::map<AXNode::AXID, AXNode::AXID> g_hit_test_result;
 class TestAXTreeObserver : public AXTreeObserver {
  private:
   void OnNodeDeleted(AXTree* tree, int32_t node_id) override {
-    const auto iter = g_node_id_to_wrapper_map.find(node_id);
+    const auto& iter = g_node_id_to_wrapper_map.find(node_id);
```
cbracken, line 57: Nice.

cbracken (approved):

third_party/accessibility/ax/platform/ax_platform_node_win.cc
yaakovschectman: Actually on second look we probably do not need it. This macro should be defined from Windows Server 2008 onwards.

PR #37757, opened 2022-11-18
[Impeller] register modern shaders on Vulkan too

jonahwilliams:
Otherwise looking these up fails at runtime.


iskakaushik (approved):
LGTM

PR #37761, opened 2022-11-18
Add strong qualifier to backgroundColor.

chingjun:
This is causing issue when `-Wobjc-property-no-attribute` is turned on.

Since LLVM 3.1, strong has been the default value so this change is effectively a no-op, but improves readability.

Similar to #36929, this is needed internally.

Also, can we consider enabling `-Wobjc-property-no-attribute` in the engine tree? I can do that if someone points me to where this is specified. Thanks

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/shell/platform/darwin/macos/framework/Headers/FlutterViewController.h
+++ b/shell/platform/darwin/macos/framework/Headers/FlutterViewController.h
@@ -115,6 +115,6 @@ FLUTTER_DARWIN_EXPORT
  * }
  * ```
  */
-@property(readwrite, nonatomic, nullable) NSColor* backgroundColor;
+@property(readwrite, nonatomic, nullable, strong) NSColor* backgroundColor;
 
 @end


jmagman:
Re https://github.com/flutter/engine/pull/36929 generally properties with mutable variants, like `NSMutableArray`, should be `copy` not `strong.


> Also, can we consider enabling `-Wobjc-property-no-attribute` in the engine tree? I can do that if someone points me to where this is specified. Thanks

@cbracken is this the right spot?
https://github.com/flutter/engine/blob/965f87d314b87c2ab029c59f066e3dbe85d7e88a/common/config.gni#L68
```
flutter_cflags_objc_arc = flutter_cflags_objc + [
  "-fobjc-arc",
  "-Wobjc-property-no-attribute",
] 
```

shell/platform/darwin/macos/framework/Headers/FlutterViewController.h
```diff
@@ -115,6 +115,6 @@ FLUTTER_DARWIN_EXPORT
  * }
  * ```
  */
-@property(readwrite, nonatomic, nullable) NSColor* backgroundColor;
+@property(readwrite, nonatomic, nullable, strong) NSColor* backgroundColor;
```
cbracken, line 118: Use `copy` here rather than `strong`. NSColor is immutable but there's no guarantee someone hasn't subclassed it. (IIRC NSColor is actually a tagged pointer these days so yeah, this is probably inconsequential, but really copy is the right semantics to express).

cbracken:
> @cbracken is this the right spot?

@jmagman that is indeed the right spot! Go for it!

cbracken:
> This is causing issue when -Wobjc-property-no-attribute is turned on.

Did some playing around with this and this diagnostic is [enabled by default](https://releases.llvm.org/14.0.0/tools/clang/docs/DiagnosticsReference.html#wobjc-property-no-attribute) in clang. We don't disable it in our build.

I experimented with turning it on (and making it an error), but a bit of digging suggests that warning is disabled when ARC is enabled (we enable it in the macOS embedder): https://github.com/flutter/engine/blob/f81ac3b1996cad041d558fb503d165fb686d6520/shell/platform/darwin/macos/BUILD.gn#L136

Are we compiling with ARC enabled internally? I'd imagine we must be since otherwise `__weak` annotations should upset the compiler.

cbracken:
I've sent out https://github.com/flutter/engine/pull/37768 which fixes all the cases of NSObject properties that weren't explicitly annotated in the macOS embedder and fixes the annotation introduced in https://github.com/flutter/engine/pull/36929.

I'm happy to enable the warning explicitly in that patch, but as I say, it should be enabled by default -- we don't disable it externally, but enabling ARC via `-fobjc-arc` (which we do) does appear to disable it and I'm not sure how to force it back on. I see a few mentions of this elsewhere on the web.

PR #37764, opened 2022-11-18
[Impeller] Fix glyph sampling overlap

bdero:
Fixes https://github.com/flutter/flutter/issues/114563.

* Rename overloaded terminology in the shaders (glyph, atlas glyph, glyph atlas, etc.. are now "source" and "destination").
* Add 2 pixels of padding for every glyph drawn to the atlas.
* Don't Ceil the source glyph size.
* Don't shift the source rect right/down by `1/atlas size`.
* Remove 0.5 from the start and end of the source and destination rects -- this gets nearest sampling to nearly work as expected. Remaining problems are detailed below.

There are two problems remaining:
* The reason we need linear sampling right now because we don't snap the source rect to the pixel grid of the destination framebuffer when drawing glyphs. This means the sampling range of each for each texel from the source image is `+/- 0.5 +/- epsilon^n`, where `epsilon^n` is the accumulated floating point error from the screen space->UV conversion of the source rect. I have a nearly working solution locally, but putting it on the back burner for the moment because there are bigger fish to fry before the branch cut.
* There is still 1 pixel of unexplained overlap when linear sampling is turned off (overlaps at most 1.5 pixels). I suspect there is legitimately an issue with the deprecated min/max metrics we're pulling from Skia and we need to switch to variable sized/per glyph metrics instead, but I haven't looked deeply into this aspect yet.

Before
Screen Shot 2022-11-18 at 3 37 26 PM

After
Screen Shot 2022-11-18 at 3 33 27 PM

### DIFF
--- a/impeller/compiler/shader_lib/impeller/transform.glsl
+++ b/impeller/compiler/shader_lib/impeller/transform.glsl
@@ -13,22 +13,16 @@ vec2 IPVec2TransformPosition(mat4 matrix, vec2 point) {
 
 // Returns the transformed gl_Position for a given glyph position in a glyph
 // atlas.
-vec4 IPPositionForGlyphPosition(mat4 mvp, vec2 unit_vertex, vec2 glyph_position, vec2 glyph_size) {
-  vec4 translate = mvp[0] * glyph_position.x
-                 + mvp[1] * glyph_position.y
-                 + mvp[3];
-  mat4 translated_mvp = mat4(
-    mvp[0],
-    mvp[1],
-    mvp[2],
-    vec4(
-      translate.xyz,
-      mvp[3].w
-    )
-  );
-  return translated_mvp *
-         vec4(unit_vertex.x * glyph_size.x,
-              unit_vertex.y * glyph_size.y, 0.0, 1.0);
+vec4 IPPositionForGlyphPosition(mat4 mvp,
+                                vec2 unit_position,
+                                vec2 glyph_position,
+                                vec2 glyph_size) {
+  vec4 translate =
+      mvp[0] * glyph_position.x + mvp[1] * glyph_position.y + mvp[3];
+  mat4 translated_mvp =
+      mat4(mvp[0], mvp[1], mvp[2], vec4(translate.xyz, mvp[3].w));
+  return translated_mvp * vec4(unit_position.x * glyph_size.x,
+                               unit_position.y * glyph_size.y, 0.0, 1.0);
 }
 
 #endif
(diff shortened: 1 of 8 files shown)


impeller/compiler/shader_lib/impeller/transform.glsl
```diff
@@ -13,22 +13,16 @@ vec2 IPVec2TransformPosition(mat4 matrix, vec2 point) {
...
+vec4 IPPositionForGlyphPosition(mat4 mvp,
+                                vec2 unit_position,
+                                vec2 glyph_position,
+                                vec2 glyph_size) {
+  vec4 translate =
+      mvp[0] * glyph_position.x + mvp[1] * glyph_position.y + mvp[3];
+  mat4 translated_mvp =
+      mat4(mvp[0], mvp[1], mvp[2], vec4(translate.xyz, mvp[3].w));
+  return translated_mvp * vec4(unit_position.x * glyph_size.x,
+                               unit_position.y * glyph_size.y, 0.0, 1.0);
```
dnfield, line 25: Was this done by hand or with a formatter?

bdero: Formatter!

dnfield (commented):
Can you add a test that manually provides some glyphs with metrics that would cause the overlap to happen without this patch ?That way if we refactor or figure out a way to remove the padding in the future we'll have it. Even if it's just a playground test that isn't verified on CI yet that would help I think.

impeller/compiler/shader_lib/impeller/transform.glsl
dnfield: We should see if we can add this to `format.sh` somehow.

bdero:
I unfortunately haven't been able to reproduce this in a playground. I think part of the reason we've only seen it in one app is that a few stars need to align to make it happen -- typeface, fallback behavior (platform specific?), size variation between text frames, fractional pixel offset of the source rect, etc. We might need to pull in another typeface to reproduce it. I agree it would be good if we had a convenient playground, but mind if I punt this investigation down the road to https://github.com/flutter/flutter/issues/115041?

dnfield (approved):

impeller/compiler/shader_lib/impeller/transform.glsl
bdero: Ah, we've discussed this in a couple of weekly syncs, but I couldn't find a bug for it. I went ahead and investigated how my formatter works in VSCode and filed a bug: https://github.com/flutter/flutter/issues/115677

Turns out it just runs everything through the C++ formatter!

PR #37789, opened 2022-11-20
[macOS] Refactor rendering infrastructure

knopp:
Notable changes:

- Removed `FlutterIOSurfaceHolder`, `FlutterResizableBackingStore`
- Replaced `FlutterResizeSynchronizer` with `FlutterThreadSynchronizer`
- Changed the role of `FlutterSurfaceManager`
- Unified most of the rendering codepath with and without compositor

### High level overview

- `FlutterView` owns `FlutterSurfaceManager`
- `FlutterCompositor` and `FlutterRenderer` interact with `FlutterSurfaceManager` in order to 
    - Obtain render surfaces, which are then represented as metal textures and provided to engine.
    - Present render surfaces, which `FlutterSurfaceManager` sets as contents of CALayers that it manages.
- `FlutterSurfaceManager` caches surfaces so that during normal (no resizing) repaint IOSurfaces do not need to be recreated.
- `FlutterThreadSynchronizer` is in charge of synchronising raster and platform thread:
    - During resizing blocks platform thread until content of requested size is available.
    - When rasterizing is done blocks raster thread while layers contents are updated and platform views are mutated (on platform thread).

### Fixes

- Simplifies rendering flow.
- Fixes threading issues with platform views (`FlutterEngine.Compositor` test now works)
- Fixes issue with IOSurfaces constantly being recreated during regular (non resizing) repaints when using platform views.

Fixes https://github.com/flutter/flutter/issues/96668
Fixes https://github.com/flutter/flutter/issues/113785

## Pre-launch Checklist

- [X] I signed the [CLA].


shell/platform/darwin/macos/framework/Source/FlutterCompositor.mm
```diff
@@ -29,53 +25,16 @@
...
-      CFRelease(user_data);
-    }
-  };
-
+  backing_store_out->metal.struct_size = sizeof(FlutterMetalBackingStore);
+  backing_store_out->metal.texture = surface.asFlutterMetalTexture;
   return true;
 }
 
 bool FlutterCompositor::CollectBackingStore(const FlutterBackingStore* backing_store) {
```
dkwingsmt, line 37: Do we still need `CollectBackingStore`? Can we remove it?

dkwingsmt: I find the only "override" of this method being `EmbedderTestCompositor`, which is not a subclass of `FlutterCompositor` despite being of a similar name.

shell/platform/darwin/macos/framework/Source/FlutterCompositor.mm
```diff
@@ -128,60 +87,18 @@
...
-  SetFrameStatus(FrameStatus::kStarted);
-}
-
-bool FlutterCompositor::EndFrame(bool has_flutter_content) {
-  bool status = present_callback_(has_flutter_content);
-  SetFrameStatus(FrameStatus::kEnded);
-  return status;
-}
-
 FlutterView* FlutterCompositor::GetView(uint64_t view_id) {
```
dkwingsmt, line 100: We can actually remove this function (replace it with its body) now that `FlutterCompositor` doesn't have subclasses any more.

shell/platform/darwin/macos/framework/Source/FlutterCompositor.mm
knopp: I removed the method. Empty callback still needs to be registered in the engine though (it's mandatory).

shell/platform/darwin/macos/framework/Source/FlutterSurface.mm
```diff
@@ -0,0 +1,94 @@
...
+
+- (CGSize)size {
+  return _size;
+}
+
+- (int64_t)textureId {
+  return reinterpret_cast<int64_t>(_texture);
+}
+
+- (instancetype)initWithSize:(CGSize)size device:(id<MTLDevice>)device {
```
dkwingsmt, line 30: Is it the Objective-C style to hide these methods in the `.mm` although they're used by other classes? I don't think it's necessary to make `FlutterSurface` so opaque since it's only used in the embedding anyway. Rather, it's much clearer to show which methods are intended to be used by other internal classes.

knopp:
I updated @iskakaushik's platform view example if you want to test platform views:

https://github.com/knopp/flutter_macos_platform_view_example

https://user-images.githubusercontent.com/96958/203168601-b16caf5f-d512-471c-ac0d-4400f151afbc.mov

shell/platform/darwin/macos/framework/Source/FlutterView.h
```diff
@@ -81,3 +72,13 @@ constexpr uint64_t kFlutterDefaultViewId = 0;
...
+
+@interface FlutterView (FlutterViewPrivate)
+
+/**
+ * Returns FlutterThreadSynchronizer for this view.
+ * Used for FlutterEngineTest.
+ */
+- (nonnull FlutterThreadSynchronizer*)threadSynchronizer;
+
+@end
```
dkwingsmt, line 84:
```suggestion
@end

```

shell/platform/darwin/macos/framework/Source/FlutterCompositor.mm
```diff
@@ -23,103 +19,62 @@
...
-  return EndFrame(has_flutter_content);
+  [view.surfaceManager present:surfaces
+                        notify:^{
+                          for (size_t i = 0; i < layers_count; i++) {
+                            FlutterLayer* layer = (FlutterLayer*)layers[i];
+                            if (layer->type == kFlutterLayerContentTypePlatformView) {
+                              PresentPlatformView(view, layer, i);
+                            }
+                          }
+                          [platform_view_controller_ disposePlatformViews];
```
dkwingsmt, line 69: I searched and found nobody calling this method at all right now. Is it because the presenting is not properly implemented anyway?

dkwingsmt (commented):
The change looks OK to me, but I don't have enough knowledge to make the final approval.

shell/platform/darwin/macos/framework/Source/FlutterCompositor.mm
knopp: Do you mean calling `disposePlatformViews`? I'm not sure what you mean by nobody, `FlutterPlatformViewController` is only used on macOS.

On iOS for example the vies are disposed [here](https://github.com/flutter/engine/blob/7b4291548841f91dd2796803fc27e707f84ebf22/shell/platform/darwin/ios/framework/Source/FlutterPlatformViews.mm#L634).

shell/platform/darwin/macos/framework/Source/FlutterCompositor.h
```diff
@@ -9,11 +9,12 @@
...
 #include "flutter/fml/macros.h"
-#import "flutter/shell/platform/darwin/macos/framework/Source/FlutterPlatformViewController.h"
-#include "flutter/shell/platform/darwin/macos/framework/Source/FlutterView.h"
-#include "flutter/shell/platform/darwin/macos/framework/Source/FlutterViewProvider.h"
 #include "flutter/shell/platform/embedder/embedder.h"
 
+#import "flutter/shell/platform/darwin/macos/framework/Source/FlutterPlatformViewController.h"
+#import "flutter/shell/platform/darwin/macos/framework/Source/FlutterView.h"
+#import "flutter/shell/platform/darwin/macos/framework/Source/FlutterViewProvider.h"
+
```
cbracken, line 17: These guys (unfortunately) are meant to be interleaved with includes in a single group according to the style guide

knopp: Looks a bit weird, but sure :) Fixed.

shell/platform/darwin/macos/framework/Source/FlutterRenderer.mm
```diff
@@ -95,30 +95,34 @@ - (FlutterMetalTexture)createTextureForView:(uint64_t)viewId size:(CGSize)size {
...
-  FlutterView* view = [_viewProvider getView:viewId];
-  if (view == nil) {
-    return;
+  FlutterSurface* surface = [view.surfaceManager lookupSurface:texture];
+  if (surface == nil) {
+    return NO;
   }
-  [view presentWithoutContent];
+  FlutterSurfacePresentInfo* info = [[FlutterSurfacePresentInfo alloc] init];
+  info.offset = CGPointMake(0, 0);
```
cbracken, line 121:
```suggestion
  info.offset = CGPointZero;
```

shell/platform/darwin/macos/framework/Source/FlutterRenderer.mm
```diff
@@ -95,30 +95,34 @@ - (FlutterMetalTexture)createTextureForView:(uint64_t)viewId size:(CGSize)size {
...
+  info.zIndex = 0;
```
cbracken, line 122: Consider moving defaulting to the `init` method. `alloc` already zero-initialises, but I have no idea whether CGPointZero is just zeros (I assume so), and the cost is negligible anyway.

cbracken (commented):
Overall, this is a really nice cleanup. Almost all of my comments are nitpicks.

Sending first round of comments. Still looking at FlutterSurfaceManager and FlutterThreadManager.

shell/platform/darwin/macos/framework/Source/FlutterCompositor.mm
cbracken: @dkwingsmt it's because platform views aren't yet supported on macOS, so the lack of the dispose call comes down to the current state of platform views not being finished and/or an oversight.

shell/platform/darwin/macos/framework/Source/FlutterCompositor.mm
```diff
@@ -23,103 +19,62 @@
...
       }
-    };
+    }
   }
 
-  return EndFrame(has_flutter_content);
+  [view.surfaceManager present:surfaces
+                        notify:^{
+                          for (size_t i = 0; i < layers_count; i++) {
+                            FlutterLayer* layer = (FlutterLayer*)layers[i];
```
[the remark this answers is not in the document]
knopp: I'm not sure why the cast was there. It stripped the const from layer, but that was quite unnecessary. I removed it.

shell/platform/darwin/macos/framework/Source/FlutterSurfaceManager_Internal.h
```diff
@@ -0,0 +1,41 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
```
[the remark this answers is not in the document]
knopp: I removed the _Internal headers.

shell/platform/darwin/macos/framework/Source/FlutterRenderer.mm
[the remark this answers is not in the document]
knopp: We are in FlutterCompositor.mm.

shell/platform/darwin/macos/framework/Source/FlutterSurface.mm
```diff
@@ -0,0 +1,94 @@
...
+  if (self = [super init]) {
+    self->_size = size;
+    self->_ioSurface = [FlutterSurface createIOSurfaceWithSize:size];
+    self->_texture = [FlutterSurface createTextureForIOSurface:_ioSurface size:size device:device];
+  }
+  return self;
+}
+
+static void ReleaseSurface(void* surface) {
```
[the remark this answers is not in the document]
knopp: It gets little bit ugly because of ARC. I think it'd need to be something like this. Every cast needs a __bridge.

```objc
static void ReleaseSurface(void* user_data) {
  FlutterSurface* surface = (__bridge id)user_data;
  if (surface != nullptr) {
    CFBridgingRelease((__bridge void*)surface);
  }
}
```

or maybe 

```objc
static void ReleaseSurface(void* user_data) {
  if (user_data != nullptr) {
    FlutterSurface* surface = CFBridgingRelease(user_data);
    (void)surface;
  }
}
```

Which one do you prefer?

skia-gold:
Gold has detected about 2 new digest(s) on patchset 8.
View them at https://flutter-engine-gold.skia.org/cl/github/37789

shell/platform/darwin/macos/framework/Source/FlutterRenderer.mm
```diff
@@ -95,30 +95,34 @@ - (FlutterMetalTexture)createTextureForView:(uint64_t)viewId size:(CGSize)size {
...
-  embedderTexture.texture_id = reinterpret_cast<int64_t>(texture);
-  return embedderTexture;
+  FlutterSurface* surface = [view.surfaceManager surfaceForSize:size];
+  FlutterMetalTexture texture = surface.asFlutterMetalTexture;
+  // This is here because the non-compositor path completely ignores
+  // destruction and callback. It is ugly, but the surface manager
+  // will keep the borrowed surface retained until present: is called.
+  if (texture.destruction_callback != nullptr) {
+    texture.destruction_callback(texture.user_data);
+  }
```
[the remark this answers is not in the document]
knopp: The root problem here is that the non compositor path of embedder throws away the `user_data` and `destruction_callback` fields in `FlutterMetalTexture`. If we could fix that and ensure that when `FlutterMetalPresentCallback` is called, the user_data is intact, we could remove the "borrowed surfaces" bookkeeping from surface manager completely.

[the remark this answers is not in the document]
knopp: In [this commit](https://github.com/flutter/engine/pull/37789/commits/27f0891bf7f571f0143b88cbb28561744e2f283f), I made sure that non compositing path properly remember user data and calls the destruction callback. I'm not sure that's the best implementation, but it needs to be done. I'm not too happy with that shared_ptr there. I tried to make `GPUMTLTextureInfo` non copyable, but unfortunately it is captured by `std::function`, which is copyable so this doesn't work.

Still, this needs to be fixed one way or another, otherwise the user_data in passed to the present callback is just garbage, which is really bad.

With that in place, I could remove any bookkeeping of borrowed surfaces. So the `FlutterSurface` looks like this now:

```objc
@interface FlutterSurface : NSObject

- (FlutterMetalTexture)asFlutterMetalTexture;
+ (nullable FlutterSurface*)fromFlutterMetalTexture:(nonnull const FlutterMetalTexture*)texture;

@end
```

And the general flow is
```objc

FlutterMetalTexture GetTexture() {
   FlutterSurface *surface = [surfaceManager surfaceForSize: size];
   return [surface asFlutterMetalTexture];
}

void PresentMetalTexture(const FlutterMetalTexture *texture) {
   FlutterSurface *surface = [FlutterSurface fromFlutterMetalTexture:texture];
  [surfaceManager present: ...];
}
```

skia-gold:
Gold has detected about 66 new digest(s) on patchset 19.
View them at https://flutter-engine-gold.skia.org/cl/github/37789

shell/platform/darwin/macos/framework/Source/FlutterSurface.mm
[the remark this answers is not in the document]
cbracken: Ugh yeah, the ARC implication hadn't crossed my mind. 

Of the two, the first reads a little better to me, mostly since people will be less likely to simplify it, the go through the whole thing of hitting the unused var warning, then removing the local that was only there to show the type, at which point we're left with the same thing with a different variable name.

To be honest, after all that, I'd be tempted to just leave as-is. I don't think the very minor hinting we'd get is worth the tradeoff.

lukasz-lukasz-lukasz:
> I updated @iskakaushik's platform view example if anyone wants to test platform views:
> 
> https://github.com/knopp/flutter_macos_platform_view_example
> 
>  platform.view.example.mov

I have tested the example app today and works well 🎉 , but after extending it a bit to have something more than displaying the text in it, I met a small issue.

If we show the dialogue e.g AlertDialog on top of Platform View, the Platform View is under the dialogue (what is correct) but the dialogue's barrier is "transparent" for the events. I mean that I can interact with Platform View content even though there is AlertDialog displayed on top of it. I can't interact with other flutter components, I can only with Platform View content.

See the attached video.

https://user-images.githubusercontent.com/119860089/205910375-a298e747-5292-40c1-b36e-b071ad78a724.mp4

Is that a known issue? or maybe something we can address easily here?

P.S. I am sorry if this comment shouldn't be added here and the issue should be reported in another way. If so, and if possible please guide me on how it should be so I will know in the future.

knopp:
This PR is not meant to address event handling in any way. That will come later.

shell/platform/darwin/macos/framework/Source/FlutterRenderer.mm
```diff
@@ -95,30 +95,24 @@ - (FlutterMetalTexture)createTextureForView:(uint64_t)viewId size:(CGSize)size {
...
-  FlutterRenderBackingStore* backingStore = [view backingStoreForSize:size];
-  id<MTLTexture> texture = backingStore.texture;
-  FlutterMetalTexture embedderTexture;
-  embedderTexture.struct_size = sizeof(FlutterMetalTexture);
-  embedderTexture.texture = (__bridge void*)texture;
-  embedderTexture.texture_id = reinterpret_cast<int64_t>(texture);
-  return embedderTexture;
+  FlutterSurface* surface = [view.surfaceManager surfaceForSize:size];
+  FlutterMetalTexture texture = surface.asFlutterMetalTexture;
+  return texture;
```
cbracken, line 100: pedantic nit: I'd be tempted to just squash this into one line. It might not be too bad to squash all three into one line.

cbracken, line 100: pedantic nit: I'd be tempted to squash the last two lines into one line, or maybe all three:
```suggestion
  return [view.surfaceManager surfaceForSize:size].asFlutterMetalTexture;
```

shell/platform/darwin/macos/framework/Source/FlutterSurfaceManager.mm
```diff
@@ -110,52 +67,128 @@ - (void)reschedule {
...
+    [_layers.lastObject removeFromSuperlayer];
+    [_layers removeLastObject];
+  }
+
+  // Add missing layers.
+  while (_layers.count < _frontSurfaces.count) {
+    CALayer* layer = [CALayer layer];
+    [_containingLayer addSublayer:layer];
+    [_layers addObject:layer];
+  }
```
cbracken, line 159: Consider just one comment followed by the two while blocks one after the other.
```
// Add or remove layers to match the count of surfaces to present.
```

cbracken (commented):
Alright - last round of comments I think!

shell/platform/darwin/macos/framework/Source/FlutterRenderer.mm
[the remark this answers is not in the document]
cbracken: To update this thread, the embedder API release bug was tracked in this bug and is now fixed: https://github.com/flutter/flutter/issues/116381

shell/platform/darwin/macos/framework/Source/FlutterThreadSynchronizer.mm
```diff
@@ -0,0 +1,132 @@
...
+
+  // Used to block [beginResize:].
+  std::condition_variable _condBlockBeginResize;
+}
+
+@end
+
+namespace {
+/// Single use event (can only be signalled once)
+class Event {
```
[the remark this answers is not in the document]
knopp: The event is only ever signalled once. It doesn't matter if it resets. I think it can be safely replaced with `AutoResetWaitableEvent`. I didn't do it originally because I didn't want to add another depenency on FML, but that's not a problem anymore so I'll go ahead and replace it.

shell/platform/darwin/macos/framework/Source/FlutterThreadSynchronizer.mm
```diff
@@ -0,0 +1,132 @@
...
+@interface FlutterThreadSynchronizer () {
+  std::mutex _mutex;
+  BOOL _shuttingDown;
+  CGSize _contentSize;
+  std::vector<dispatch_block_t> _scheduledBlocks;
+
+  BOOL _beginResizeWaiting;
+
+  // Used to block [beginResize:].
+  std::condition_variable _condBlockBeginResize;
```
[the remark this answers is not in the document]
knopp: The mutex is used to synchronize access to `_scheduledBlocks` and `_contentSize`. I don't think this instance can be replaced with `AutoResetWaitableEvent`. The original code had actually too condition variables (one for blocking platform thread and one for blocking raster thread, since we need to do both). The refactor simplified this significantly replacing one of them with event, but I don't think we can safely get rid of both of them without introducing subtle synchronisation issues.

shell/platform/darwin/macos/framework/Source/FlutterThreadSynchronizer.mm
```diff
@@ -0,0 +1,132 @@
...
+  std::unique_lock<std::mutex> lock(_mutex);
+
+  if (CGSizeEqualToSize(_contentSize, CGSizeZero) || _shuttingDown) {
+    // No blocking until framework produces at least one frame
+    notify();
+    return;
+  }
+
+  [CATransaction begin];
+  [CATransaction setDisableActions:YES];
```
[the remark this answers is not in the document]
knopp: I think the transaction in beginResize is not necessary. We might end up in multiple transactions from drain during beginResize:, but that's all in same run-loop turn so it shouldn't matter.

I'd leave the one in drain, since I think it is better to do it there than on two completely separate places (updating CALayers and mutating platform views).

shell/platform/darwin/macos/framework/Source/FlutterThreadSynchronizer.mm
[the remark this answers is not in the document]
cbracken: Ugh yeah, too much code review in one day for me, apparently. Agreed between the need for wait and for memory fences for synchronisation we're probably stuck with it here.

cbracken (approved):

🎉🎉🎉

PR #37821, opened 2022-11-21
Add setImageSampler (for replacing setSampler)

bdero:
`setImageSampler` guarantees we don't need an additional texture/render pass per-sampler. It's possible for users to emulate the functionality of `setSampler` in-shader.

The intention is to do the following before the stable branch cut:
1. Land this and roll into the framework.
1. Migrate stuff in the framework from `setSampler` to `setImageSampler`.
1. Remove `setSampler`.

I filed https://github.com/flutter/flutter/issues/115794 along the way.

### DIFF
--- a/lib/ui/painting.dart
+++ b/lib/ui/painting.dart
@@ -4323,7 +4323,7 @@ class FragmentShader extends Shader {
   /// results will be undefined.
   void setImageSampler(int index, Image image) {
     assert(!debugDisposed, 'Tried to access uniforms on a disposed Shader: $this');
-    _setImageSampler(index, image);
+    _setImageSampler(index, image._image);
   }
 
   /// Sets the sampler uniform at [index] to [sampler].
@@ -4354,7 +4354,7 @@ class FragmentShader extends Shader {
   external Float32List _constructor(FragmentProgram program, int floatUniforms, int samplerUniforms);
 
   @FfiNative<Void Function(Pointer<Void>, Handle, Handle)>('ReusableFragmentShader::SetImageSampler')
-  external void _setImageSampler(int index, Image sampler);
+  external void _setImageSampler(int index, _Image sampler);
 
   @FfiNative<Void Function(Pointer<Void>, Handle, Handle)>('ReusableFragmentShader::SetSampler')
   external void _setSampler(int index, ImageShader sampler);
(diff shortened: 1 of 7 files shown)


lib/ui/painting.dart
```diff
@@ -4314,13 +4314,27 @@ class FragmentShader extends Shader {
     _floats[index] = value;
   }
 
+  /// Sets the sampler uniform at [index] to [image].
+  ///
+  /// The index provided to setSampler is the index of the sampler uniform defined
```
zanderso, line 4319:
```suggestion
  /// The index provided to setImageSampler is the index of the sampler uniform defined
```

zanderso (approved):
lgtm w/ nit.

lib/ui/painting.dart
```diff
@@ -4314,13 +4314,27 @@ class FragmentShader extends Shader {
...
+
   /// Sets the sampler uniform at [index] to [sampler].
   ///
   /// The index provided to setSampler is the index of the sampler uniform defined
   /// in the fragment program, excluding all non-sampler uniforms.
   ///
   /// All the sampler uniforms that a shader expects must be provided or the
   /// results will be undefined.
+  ///
+  /// Use of this method may result in poor performance with the Impeller backend.
```
jonahwilliams, line 4337: I would be careful about describing something as having poor performance with no refernce point. I've seen users go to extensive lengths to avoid invoking apis that have warnings like this, sometimes exceeding the original performance cost.

lib/ui/painting.dart
bdero: Done.

lib/ui/painting.dart
bdero: SGTM, removed this.

jonahwilliams (approved):
LGTM!

bdero:
Added implementation for canvaskit

bdero:
Smoke tested this on iOS and fixed an issue where the engine handle wasn't being passed in and it's working as expected now. Attempting to smoke test canvaskit.

jonahwilliams:
I wouldn't worry about canvaskit right now, the required Skia API isn't available in most engines and I'll probably disable shader compilation for the next stable

bdero:
Alrighty, I'll go ahead and merge after this CI run then.

PR #37823, opened 2022-11-21
[Impeller] use explicit half types

jonahwilliams:
Fixes https://github.com/flutter/flutter/issues/115044

WIP because of syntax issues and generated headers being broken.

See 
* https://registry.khronos.org/OpenGL/extensions/AMD/AMD_gpu_shader_half_float.txt
* https://github.com/KhronosGroup/Vulkan-Samples/tree/master/samples/performance/16bit_arithmetic#explicit-16-bit-arithmetic-vs-mediump--relaxedprecision


jonahwilliams:
Current problems: generated headers are broken on iOS. Can't use f16mat4 in uniforms or we hit a SPIRV_CROSS_THROW error because it doesn't seem to recognize the f16mat4 as a matrix type.

jonahwilliams:
I'm closing this for now until we figure out how to fix codegen

PR #37829, opened 2022-11-22
[impeller] Remove declare_undefined_values

CaseyHillers:
This is a temporary fix to unblock the internal roll.

* This call has been deprecated in the upstream dep
* https://github.com/KhronosGroup/SPIRV-Cross/pull/2062

https://github.com/flutter/flutter/issues/115785

## Pre-launch Checklist

- [X] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [X] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [X] I read and followed the [Flutter Style Guide] and the [C++, Objective-C, Java style guides].
- [X] I listed at least one issue that this PR fixes in the description above.
- [ ] I added new tests to check the change I am making or feature I am adding, or Hixie said the PR is test-exempt. See [testing the engine] for instructions on writing and running engine tests.
- [X] I updated/added relevant documentation (doc comments with `///`).
- [X] I signed the [CLA].
- [ ] All existing and new tests are passing.

PR #37831, opened 2022-11-22
Revert "[impeller] Remove declare_undefined_values"

zanderso:
Reverts flutter/engine#37829

This breaks impellerc unless spirv-cross is also rolled forward.

### DIFF
--- a/impeller/compiler/spirv_sksl.cc
+++ b/impeller/compiler/spirv_sksl.cc
@@ -349,6 +349,8 @@ void CompilerSkSL::emit_resources() {
   if (emit_global_variable_resources()) {
     statement("");
   }
+
+  declare_undefined_values();
 }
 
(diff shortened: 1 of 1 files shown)


CaseyHillers (approved):

CaseyHillers:
> This breaks impellerc unless spirv-cross is also rolled forward.

Can you expand on this? This seems to remove an assertion check, which I assumed was fine to temporarily remove.

zanderso:
> > This breaks impellerc unless spirv-cross is also rolled forward.
> 
> Can you expand on this? This seems to remove an assertion check, which I assumed was fine to temporarily remove.

No, the PR being reverted removes a line of code that emits declarations into the output SkSL. In a future version of spirv-cross, those declarations are emitted elsewhere, and this line of code will no longer be needed. But without rolling spirv-cross forward, those declarations would just be omitted and the emitted code would be incorrect.

PR #37835, opened 2022-11-22
[Impeller] make VerticesGeometry delegate to the DL class

jonahwilliams:
Simplify the implementation of VerticesGeometry by delegating directly to the DlVertices object. Additionally reduces copying.

TriangleFan is already tested via CanConvertTriangleFanToTriangles


jonahwilliams:
I wanted to add texture coordinates support, but the vertices implementation had gotten a bit over complicated - so this PR simplifies it first

bdero (approved):
LGTM!

PR #37839, opened 2022-11-22
Remove setSampler from FragmentShader

bdero:
Followup to https://github.com/flutter/engine/pull/37821.

I actually didn't find any setSampler usage in the framework, so just removing it before the engine roll finishes. 🤔 Am I missing anything?


zanderso (approved):
I think you're probably not missing anything. Unless @jonahwilliams knows of any other tests in the framework repo, I think the only usage of this API might have been for the ink sparkle shader, which doesn't use samplers.

PR #37850, opened 2022-11-22
[fuchsia] Set nested clip nodes for Flatland

uysalere:
This CL processes the clip nodes that might be a part of MutatorStack. We use the same strategy as Gfx code by adding parent transforms.

Bug: fxb/115348


naudzghebre (approved):
LGTM

PR #37872, opened 2022-11-23
[web] Use generic variable name for trusted url

mdebbar:
I was exploring around this code when I found canvaskit variable names where they shouldn't be.

Rename `trustedCanvasKitUrl` => `trustedUrl`.


ditman (approved):
Good catch! This probably started life inline somewhere, and when I extracted it to a method I forgot to rename that var. Sssship it!

PR #37876, opened 2022-11-23
[impeller] use half precision for all advanced blends

jonahwilliams:
Example of differences in shader: https://gist.github.com/jonahwilliams/2194d45d671991f3eb0b030bf27a9068

Advanced blends seem like a good usage of half precision, since they generaly only have two inputs to convert and mostly operate on colors. TBD whether or not this is actually faster on iOS.

https://github.com/flutter/flutter/issues/115044

### DIFF
--- a/impeller/entity/shaders/blending/advanced_blend.glsl
+++ b/impeller/entity/shaders/blending/advanced_blend.glsl
@@ -43,8 +43,7 @@ void main() {
                        kTileModeDecal                 // tile mode
                        ));
 
-  vec4 blended = vec4(Blend(dst.rgb, src.rgb), 1) * dst.a;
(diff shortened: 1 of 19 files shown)


impeller/entity/shaders/blending/advanced_blend.glsl
```diff
@@ -43,8 +43,7 @@ void main() {
                        kTileModeDecal                 // tile mode
                        ));
 
-  vec4 blended = vec4(Blend(dst.rgb, src.rgb), 1) * dst.a;
+  vec4 blended = vec4(Blend(f16vec3(dst.rgb), f16vec3(src.rgb)), 1) * dst.a;
```
jonahwilliams, line 46: This is the only place we need to do a conversion from float to half, so assuming the half-based logic is more performant it seems like a reasonable trade-off.

jonahwilliams:
We're already using mediump everywhere so this should make no difference for gles

PR #37939, opened 2022-11-28
[impellerc] speculative fix for include errors with impellerc shader lib

jonahwilliams:
Part of a speculative fix I'm making for https://github.com/flutter/flutter/issues/115433. If the path is already absolute, don't convert again. probably does nothing. 🤷


bdero (approved):
I'm game for throwing spaghetti.

PR #37951, opened 2022-11-29
[gn] embed mac codesign metadata in android artifacts

XilaiZhang:
Embeds mac codesign metadata in android artifacts. TLDR: metadata is embedded at top level as opposed to attached to leaf nodes.

The metadata isn't attached to the leaf nodes of build rules, because of the different gen_snapshot_(host tool chain) variations. As pointed out by Zach, in engine v1, not all andriod snapshots are generated through android build rules. For example, windows android snapshots are copied to a common path `"$root_build_dir/gen_snapshot/gen_snapshot.exe"`, and used out of the android gn file. In engine v2, all android snapshots are generated via the shell/platform/android build rules, and therefore it is good to embed a top level metadata. Since we only care about mac codesign metadata, windows gen snapshot entry points might not be relevant.

### DIFF
--- a/shell/platform/android/BUILD.gn
+++ b/shell/platform/android/BUILD.gn
@@ -594,9 +594,7 @@ zip_bundle("android_javadoc") {
 
 generated_file("android_entitlement_config") {
   outputs = [ "$target_gen_dir/android_entitlements.txt" ]
-  contents = [
-    "gen_snapshot"
-  ]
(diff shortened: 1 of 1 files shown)


XilaiZhang:
impeller unit tests were failing. I triggered a re-run for mac unopt.

godofredoc (approved):

PR #37976, opened 2022-11-30
[macOS] Make FlutterEngine support multiple views

dkwingsmt:
This PR adds support for multiview to `FlutterEngine`. Objective-C user code can add more than one view controllers to`FlutterEngine` as well as removing them.

However, through discussion with goderbauer@ and loic-sharma@, we've decided that publicly `FlutterEngine` will only handle one (unmanaged) view controller, so that it looks exactly the same to the native code by having a `setViewController` that always points to the same view ID. `FlutterEngine` will still support multi-view internally, and a future class will expose this functionality.

## Pre-launch Checklist

- [ ] I signed the [CLA].


dkwingsmt:
This PR is ready for review.

shell/platform/darwin/macos/framework/Headers/FlutterEngine.h
```diff
@@ -6,6 +6,7 @@
 #define FLUTTER_FLUTTERENGINE_H_
 
 #import <Foundation/Foundation.h>
+#include <stdint.h>
 
 #include <stdint.h>
```
loic-sharma, line 11:
```suggestion

#include <stdint.h>
```

shell/platform/darwin/macos/framework/Headers/FlutterEngine.h
```diff
@@ -76,10 +77,11 @@ FLUTTER_DARWIN_EXPORT
...
 
 /**
- * The default `FlutterViewController` associated with this engine, if any.
+ * The default `FlutterViewController` of this engine, if any.
  *
- * The default view always has ID kFlutterDefaultViewId, and is the view
- * operated by the APIs that do not have a view ID specified.
+ * The default view is the first view added to the engine, always has ID
+ * kFlutterDefaultViewId, and is operated by the legacy APIs that do not specify
+ * view IDs.
```
loic-sharma, line 84: Consider something like:

```suggestion
 * The default view is the first view added to the engine, always has ID
 * kFlutterDefaultViewId, and is used by legacy APIs that assume a single view.
```

shell/platform/darwin/macos/framework/Headers/FlutterEngine.h
```diff
@@ -89,9 +91,40 @@ FLUTTER_DARWIN_EXPORT
  *
  * Setting this field from non-nil to a different non-nil FlutterViewController
  * is prohibited and will throw an assertion error.
+ *
+ * This method is deprecated. Querying view controllers should use
```
loic-sharma, line 95: Are properties referred as methods in Objective-C?

```suggestion
 * This property is deprecated. Querying view controllers should use
```

shell/platform/darwin/macos/framework/Headers/FlutterEngine.h
```diff
@@ -89,9 +91,40 @@ FLUTTER_DARWIN_EXPORT
...
+ * The first added view controller (either with this method or the
+ * viewController property) will always have ID kFlutterDefaultViewId.
+ *
+ * If the given view controller is already attached to an engine, this call
+ * throws an assertion.
+ */
+- (void)addViewController:(nonnull FlutterViewController*)viewController;
+
+/**
+ * Deassociate the given view controller from this engine.
```
loic-sharma, line 116:
```suggestion
 * Dissociate the given view controller from this engine.
```

shell/platform/darwin/macos/framework/Headers/FlutterViewController.h
```diff
@@ -89,7 +89,9 @@ FLUTTER_DARWIN_EXPORT
     NS_DESIGNATED_INITIALIZER;
 - (nonnull instancetype)initWithCoder:(nonnull NSCoder*)nibNameOrNil NS_DESIGNATED_INITIALIZER;
 /**
- * Initializes this FlutterViewController with the specified `FlutterEngine`.
+ * Initializes this FlutterViewController with an existing `FlutterEngine`.
+ *
+ * The initialized viewcontroller will add itself to the engine as part of this process.
```
loic-sharma, line 94:
```suggestion
 * The initialized view controller will add itself to the engine as part of this process.
```

shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
```diff
@@ -595,15 +717,10 @@ - (FlutterPlatformViewController*)platformViewController {
...
-  // this method should be able to return the view for any IDs.
-  NSAssert(viewId == kFlutterDefaultViewId, @"Unexpected view ID %llu", viewId);
-  if (viewId == kFlutterDefaultViewId) {
-    return _viewController;
-  }
-  return nil;
+- (uint64_t)generateViewId {
+  uint64_t result = _nextViewId;
+  _nextViewId += 1;
+  return result;
```
loic-sharma, line 723: What happens if my app creates a view, transitions back to headless mode, then creates another view? Should we guarantee that at least one view has `kFlutterDefaultViewId` if the engine is not in headless mode?

For example, imagine my app is a popup from the menu bar. It starts in headless mode until the user opens the popup. The app returns to headless mode once the user closes the popup. (This is the Google One macOS app scenario)

shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
```diff
@@ -84,6 +84,29 @@ @interface FlutterEngine () <FlutterBinaryMessenger>
 
 - (nullable FlutterViewController*)viewControllerForId:(uint64_t)viewId;
 
+/**
+ * An internal method that adds view controller with the given ID.
+ *
+ * This method assigns the controller with the ID, puts the controller into the
+ * map, and does assertions related to the default view ID.
+ */
+- (void)registerViewController:(FlutterViewController*)controller forId:(uint64_t)viewId;
```
loic-sharma, line 93: Is `withId:` more idiomatic? Same comment for `deregisterViewControllerForId` below

```suggestion
- (void)registerViewController:(FlutterViewController*)controller withId:(uint64_t)viewId;
```

shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
```diff
@@ -408,39 +461,77 @@ - (void)loadAOTData:(NSString*)assetsDir {
...
+           @"given %lld.",
+           viewId);
+  [controller attachToEngine:self withId:viewId];
+  NSAssert(controller.id == viewId, @"Failed to assign view ID.");
+  [_viewControllers setObject:controller forKey:@(viewId)];
+}
+
+- (void)deregisterViewControllerForId:(uint64_t)viewId {
+  FlutterViewController* oldController = [self viewControllerForId:viewId];
+  if (oldController != nil) {
```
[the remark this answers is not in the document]
loic-sharma: Same feedback for `viewControllerForId` below

shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
```diff
@@ -579,7 +694,14 @@ - (void)setSemanticsEnabled:(BOOL)enabled {
...
   }
   _semanticsEnabled = enabled;
-  [_viewController notifySemanticsEnabledChanged];
+
+  // Update all view controllers' bridges.
+  NSEnumerator* viewControllerEnumerator = [_viewControllers objectEnumerator];
+  FlutterViewController* nextViewController;
+  while ((nextViewController = [viewControllerEnumerator nextObject])) {
+    [nextViewController notifySemanticsEnabledChanged];
+  }
```
loic-sharma, line 703: Should this change also update the embedder API's `update_semantics_callback` callback?

loic-sharma:
@dkwingsmt Could you make sure to include me as a reviewer to any multi-view related engine pull requests? I need to do similar changes in the Windows platform, it's useful for me to track how you implement things on macOS.

dkwingsmt:
Absolutely! Thank you for your careful reviews, they're very helpful.

shell/platform/darwin/macos/framework/Headers/FlutterEngine.h
goderbauer, line 95: Consider moving this snippet further up in the doc? I have to read about all the things this can do before I learn that I shouldn't use it.

goderbauer: Also, there is no official annotation to mark something as deprecated that tools will pick up on?

shell/platform/darwin/macos/framework/Headers/FlutterEngine.h
```diff
@@ -89,9 +91,40 @@ FLUTTER_DARWIN_EXPORT
...
+ *
+ * The engine holds a weak reference to each attached view controller.
+ *
+ * The first added view controller (either with this method or the
+ * viewController property) will always have ID kFlutterDefaultViewId.
+ *
+ * If the given view controller is already attached to an engine, this call
+ * throws an assertion.
+ */
+- (void)addViewController:(nonnull FlutterViewController*)viewController;
```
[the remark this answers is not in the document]
goderbauer: +1 to this. We should also document that limitation in the doc comment.

shell/platform/darwin/macos/framework/Headers/FlutterPluginRegistrarMacOS.h
```diff
@@ -36,12 +36,20 @@ FLUTTER_DARWIN_EXPORT
...
 /**
- * The view displaying Flutter content. May return |nil|, for instance in a headless environment.
+ * The default view displaying Flutter content.
  *
- * WARNING: If/when multiple Flutter views within the same application are supported (#30701), this
- * API will change.
+ * This method may return |nil|, for instance in a headless environment.
+ *
+ * The default view is a special view operated by the legacy single-view APIs.
+ * This property is deprecated. Its calls should be replaced with viewForId:
```
goderbauer, line 44: Same comment as in the other file about the deprecation notice.

shell/platform/darwin/macos/framework/Headers/FlutterEngine.h
```diff
@@ -89,9 +91,40 @@ FLUTTER_DARWIN_EXPORT
...
 @property(nonatomic, nullable, weak) FlutterViewController* viewController;
 
+/**
+ * Attach a view controller to the engine and associate it with a newly
+ * generated ID.
+ *
+ * The engine holds a weak reference to each attached view controller.
+ *
+ * The first added view controller (either with this method or the
+ * viewController property) will always have ID kFlutterDefaultViewId.
```
goderbauer, line 108: I would consider this an implementation detail and not document this. People shouldn't assume that a view will get a certain ID. They should always read the ID from the view controller.

shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
goderbauer: Overall, I think we should make fewer assumptions about what ID a view will have or get. Ideally, we make no assumptions at all about this. Once consumers of these APIs have migrated to the multi-view APIs they should really check what ID they got instead of relying on some default behavior.

shell/platform/darwin/macos/framework/Source/FlutterViewProvider.h
```diff
@@ -15,11 +15,18 @@ extern const uint64_t kFlutterDefaultViewId;
  */
 @protocol FlutterViewProvider
 
+/**
+ * Get the FlutterView for the default view.
+ *
+ * Returns nil if it doesn't exist.
+ */
+- (nullable FlutterView*)getDefaultView;
```
goderbauer, line 23: Do we have to add this? Ideally, we should no longer make the assumption that there is a default view. If we have this, it bags the question: What is the default view and why is it special?

goderbauer:
Overall it seems a bit strange that we add these public-facing APIs without having them actually be backed by an implementation. See also this comment: https://github.com/flutter/engine/pull/37976#discussion_r1093595076.

shell/platform/darwin/macos/framework/Headers/FlutterEngine.h
```diff
@@ -6,6 +6,7 @@
 #define FLUTTER_FLUTTERENGINE_H_
 
 #import <Foundation/Foundation.h>
+#include <stdint.h>
```
cbracken, line 9: This is already present on line 11

shell/platform/darwin/macos/framework/Headers/FlutterEngine.h
cbracken: A getter and for mutable properties, a setter method are synthesised, but properties should be referred to as properties in docs.

shell/platform/darwin/macos/framework/Headers/FlutterEngine.h
```diff
@@ -89,9 +91,40 @@ FLUTTER_DARWIN_EXPORT
  *
  * Setting this field from non-nil to a different non-nil FlutterViewController
  * is prohibited and will throw an assertion error.
+ *
+ * This method is deprecated. Querying view controllers should use
+ * viewControllerForId: instead. Assigning or replacing view controllers do not
+ * have a replacement. Consider addViewController: and removeViewController:.
  */
 @property(nonatomic, nullable, weak) FlutterViewController* viewController;
```
cbracken, line 99:
```suggestion
@property(nonatomic, nullable, weak) FlutterViewController* viewController FLUTTER_DEPRECATED("Use viewControllerForId: instead");
```

shell/platform/darwin/macos/framework/Headers/FlutterEngine.h
cbracken: We have a macro for this. Added a suggestion below.

shell/platform/darwin/macos/framework/Headers/FlutterPluginRegistrarMacOS.h
```diff
@@ -36,12 +36,20 @@ FLUTTER_DARWIN_EXPORT
...
+ * kFlutterDefaultViewId for viewId.
+ */
+- (nullable NSView*)view;
```
cbracken, line 47:
```suggestion
- (nullable NSView*)view FLUTTER_DEPRECATED("Use viewForId: instead");
```

shell/platform/darwin/macos/framework/Source/FlutterCompositorTest.mm
```diff
@@ -11,6 +11,8 @@
 #import "flutter/shell/platform/darwin/macos/framework/Source/FlutterViewProvider.h"
 #import "flutter/testing/testing.h"
 
+extern const uint64_t kFlutterDefaultViewId;
```
cbracken, line 14: I'm assuming the reason this is a const global rather than constexpr is that it's part of our public headers and therefore we need it to be pure Obj-C and not Obj-C++?

shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
cbracken: `For` is better here. In Obj-C, typically you'll see "withFoo" in constructors and class factory methods (`initWithView:` etc.). For getters, you'll see a lot of forFoo: (e.g. `getFirstRectForRange:`, `getObjectValue:forString:range:error:`).

shell/platform/darwin/macos/framework/Source/FlutterCompositorTest.mm
```diff
@@ -30,7 +32,11 @@ - (nonnull instancetype)initWithDefaultView:(nonnull FlutterView*)view {
   return self;
 }
 
-- (nullable FlutterView*)getView:(uint64_t)viewId {
+- (nullable FlutterView*)getDefaultView {
+  return [self getViewForId:kFlutterDefaultViewId];
+}
+
+- (nullable FlutterView*)getViewForId:(uint64_t)viewId {
```
cbracken, line 39: Don't prefix getters with 'get' in Obj-C. `viewForId:...`

shell/platform/darwin/macos/framework/Source/FlutterViewProvider.h
```diff
@@ -15,11 +15,18 @@ extern const uint64_t kFlutterDefaultViewId;
...
+
 /**
  * Get the FlutterView with the given view ID.
  *
  * Returns nil if the ID is invalid.
  */
-- (nullable FlutterView*)getView:(uint64_t)id;
+- (nullable FlutterView*)getViewForId:(uint64_t)id;
```
cbracken, line 30: Getters don't take a 'get' prefix in Obj-C. Use `viewForId:...`

shell/platform/darwin/macos/framework/Source/FlutterViewProvider.h
cbracken: Agreed - I'd avoid this. It feels like adding new API for a legacy concept. Since this isn't public API, it's probably actually advantageous to force the longhand version of looking it up by default ID where needed, just to make the code more explicit and to avoid people accidentally adding more use of single-view assumptions/conveniences.

If, for whatever reason, we somehow absolutely can't live without this method, it should be called `defaultView` not `getDefaultView`. Getter's don't take a leading 'get' in Obj-C.

shell/platform/darwin/macos/framework/Source/FlutterViewEngineProvider.mm
```diff
@@ -22,14 +22,12 @@ - (instancetype)initWithEngine:(FlutterEngine*)engine {
...
-  if (viewId == kFlutterDefaultViewId) {
-    return _engine.viewController.flutterView;
-  }
-  return nil;
+- (nullable FlutterView*)getDefaultView {
+  return _engine.viewController.flutterView;
+}
+
+- (nullable FlutterView*)getViewForId:(uint64_t)viewId {
+  return [_engine viewControllerForId:viewId].flutterView;
```
cbracken, line 30: See notes below on method naming/existence.

shell/platform/darwin/macos/framework/Source/FlutterKeyboardManager.mm
```diff
@@ -57,7 +57,7 @@ @interface FlutterKeyboardManager ()
 /**
  * The text input plugin set by initialization.
  */
-@property(nonatomic) id<FlutterKeyboardViewDelegate> viewDelegate;
+@property(nonatomic, weak) id<FlutterKeyboardViewDelegate> viewDelegate;
```
cbracken, line 60: Why weak? It might be worth documenting who's guaranteeing ownership in this relationship.

shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
```diff
@@ -161,10 +190,18 @@ - (instancetype)initWithPlugin:(NSString*)pluginKey flutterEngine:(FlutterEngine
...
-  if (!_flutterEngine.viewController.viewLoaded) {
-    [_flutterEngine.viewController loadView];
+  return [self viewForId:kFlutterDefaultViewId];
+}
+
+- (NSView*)viewForId:(uint64_t)viewId {
+  FlutterViewController* controller = [_flutterEngine viewControllerForId:viewId];
+  if (controller == nil) {
+    return nil;
   }
```
cbracken, line 200: Technically this isn't needed since `controller.viewLoaded` will return nil, `[controller loadView]` will be a no-op, and `controller.flutterView` will return nil... but I'm kind of tempted to say leave it to make the code easier to reason about anyway.

cbracken (commented):
Overall idea-wise, looks good. Mostly style/etc. nitpicks. Agree with @goderbauer we should avoid adding more support/convenience for the idea of 'default' view.

shell/platform/darwin/macos/framework/Source/FlutterKeyboardManager.mm
dkwingsmt: This `viewDelegate` is actually the FVC, and it must be weak because the FVC owns this `FlutterKeyboardManager`, and this cyclic ownership causes memory leak.

As I've documented in [FlutterViewController](https://github.com/flutter/engine/blob/main/shell/platform/darwin/macos/framework/Headers/FlutterViewController.h#L42), FVCs strongly reference FE, and FE weakly references FVCs. The user should hold the only strong reference to FVC (in practice, that strong reference is then moved to the NSWindow). If that  strong reference is no more, then the FVC is deallocated by the system, when removes itself from the FE. Therefore, this memory leak even has behavioral significance.

shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
dkwingsmt: It depends on which API you're using. 

If you're using `setViewController`, then the resulting FVC will always have the default ID, since `setViewController` always operates the default view.

If you're using `addViewController`, then the following FVCs will never have the default ID again. This doesn't hurt, because since you're using a new API anyway, it should be safe to assume you're not relying on the default API suite.

loic-sharma (changes requested):
Please see multi-window chat. The embedder needs to notify the framework whether it can make single view assumptions or not. The current API does not let the embedder distinguish between these scenarios:

1. Current default app - The framework can make single view assumptions
2. Add-to-app with multiple views - For example, a native app where each item in a list is a Flutter view. The framework should not make single view assumptions.

I don't think the embedder can determine whether the framework can make single view assumptions with the current API.

shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
dkwingsmt: Sorry. Are you suggesting me to keep the nil check of `controller` or not?

shell/platform/darwin/macos/framework/Headers/FlutterEngine.h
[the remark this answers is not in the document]
dkwingsmt: I'm moving adding and removing to the internal header for now. I can't assert it because there are test cases that test adding a 2nd FVC.

shell/platform/darwin/macos/framework/Headers/FlutterEngine.h
dkwingsmt: I think this detail is actually pretty important because it means that the first `addViewController` is equivalent to `setViewController`, which matters if people are migrating from `setViewController` to `addViewController`. To put it in another way, if you do a `addViewController` first for your engine, then `setViewController` always operates that FVC.

shell/platform/darwin/macos/framework/Headers/FlutterPluginRegistrarMacOS.h
dkwingsmt: I'm postponing all deprecation to another day.

shell/platform/darwin/macos/framework/Source/FlutterCompositorTest.mm
dkwingsmt: I'm not sure which header to put the constexpr to, because the files that rely on this value might not all import the same header. In this way I can put the declaration in any header file and just use it. Also, you can say in this way we're hiding the actual value of this magic value.

Do you think I should try to change it?

shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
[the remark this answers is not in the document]
dkwingsmt: `deregisterViewControllerForId` is an internal method, and it really boils down to where to put that assertion. Doesn't matter much.

As of `viewControllerForId`, it's a public method, and I think it should really be allowed to return nil. It's just like querying a hashtable.

shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
dkwingsmt: Great find! I've added a TODO to it.

shell/platform/darwin/macos/framework/Headers/FlutterEngine.h
dkwingsmt: Postponing deprecation.

shell/platform/darwin/macos/framework/Headers/FlutterPluginRegistrarMacOS.h
dkwingsmt: Postponing deprecation.

dkwingsmt:
@loic-sharma @goderbauer @cbracken Can you review again?

shell/platform/darwin/macos/framework/Headers/FlutterViewController.h
```diff
@@ -89,7 +89,9 @@ FLUTTER_DARWIN_EXPORT
     NS_DESIGNATED_INITIALIZER;
 - (nonnull instancetype)initWithCoder:(nonnull NSCoder*)nibNameOrNil NS_DESIGNATED_INITIALIZER;
 /**
- * Initializes this FlutterViewController with the specified `FlutterEngine`.
+ * Initializes this FlutterViewController with an existing `FlutterEngine`.
```
loic-sharma, line 92: Should there be an empty line between this comment and the previous method?

shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
```diff
@@ -84,6 +84,29 @@ @interface FlutterEngine () <FlutterBinaryMessenger>
 
 - (nullable FlutterViewController*)viewControllerForId:(uint64_t)viewId;
 
+/**
+ * An internal method that adds view controller with the given ID.
```
loic-sharma, line 88:
```suggestion
 * An internal method that adds the view controller with the given ID.
```

shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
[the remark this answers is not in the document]
loic-sharma: Sounds good, marking as resolved

shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
```diff
@@ -408,39 +464,77 @@ - (void)loadAOTData:(NSString*)assetsDir {
...
+- (void)shutDownIfNeeded {
+  if ([_viewControllers count] == 0 && !_allowHeadlessExecution) {
+    [self shutDownEngine];
+  }
+}
+
+- (FlutterViewController*)viewControllerForId:(uint64_t)viewId {
+  FlutterViewController* controller = [_viewControllers objectForKey:@(viewId)];
+  NSAssert(controller == nil || controller.id == viewId,
+           @"The stored controller has unexpected view ID.");
```
loic-sharma, line 499: As per [this comment](https://github.com/flutter/engine/pull/37976/files#r1096338601), consider:

```suggestion
  NSAssert(controller != nil && controller.id == viewId,
           @"The stored controller has unexpected view ID.");
```

shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
```diff
@@ -408,39 +464,77 @@ - (void)loadAOTData:(NSString*)assetsDir {
...
-  } else if (_viewController != nil && controller == nil) {
+    [self registerViewController:controller forId:kFlutterDefaultViewId];
+  } else if (currentController != nil && controller == nil) {
     // From non-nil to nil.
-    [_viewController detachFromEngine];
-    _viewController = nil;
-    if (!_allowHeadlessExecution) {
-      [self shutDownEngine];
-    }
+    [self deregisterViewControllerForId:kFlutterDefaultViewId];
```
loic-sharma, line 521: Should this use the `currentController`'s ID?

shell/platform/darwin/macos/framework/Source/FlutterEngine_Internal.h
```diff
@@ -48,9 +48,39 @@
...
+ * assertion.
+ *
+ * TODO(dkwingsmt): Move this method to the public API once stable. This method
+ * should be a public method but its behavior is "broken" before we implement
+ * the proper embedder API. See its implementation for detail.
+ */
+- (void)removeViewController:(nonnull FlutterViewController*)viewController;
+
+/**
+ * Informs the engine that the specified view controller's view size has changed.
```
loic-sharma, line 81: Consider generalizing this comment as there are some metrics unrelated to size:

```suggestion
 * Informs the engine that the specified view controller's window metrics have changed.
```

loic-sharma (approved):
LGTM but please also get an approval from someone that's more familiar with the macOS embedder.

shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
dkwingsmt: Well I thought in that comment we agreed that it's ok for this method to return null?

shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
dkwingsmt: The `setViewController` method should only operate on the default view ID. The `currentController` is bound to have the default view ID, although yeah I can assert here.

shell/platform/darwin/macos/framework/Headers/FlutterEngine.h
```diff
@@ -76,10 +79,10 @@ FLUTTER_DARWIN_EXPORT
 - (BOOL)runWithEntrypoint:(nullable NSString*)entrypoint;
 
 /**
- * The default `FlutterViewController` associated with this engine, if any.
+ * The default `FlutterViewController` of this engine, if any.
  *
- * The default view always has ID kFlutterDefaultViewId, and is the view
- * operated by the APIs that do not have a view ID specified.
+ * The default view is the first view added to the engine, always has ID
+ * kFlutterDefaultViewId, and is used by legacy APIs that assume a single view.
```
goderbauer, line 85: nit: seems odd to talk about the "first" view here since this can only have one (for now)?

shell/platform/darwin/macos/framework/Source/FlutterEngine_Internal.h
```diff
@@ -48,9 +48,39 @@
...
 
 /**
- * Informs the engine that the associated view controller's view size has changed.
+ * Attach a view controller to the engine as its default controller.
+ *
+ * Practically, since FlutterEngine can only be attached with one controller,
+ * the given controller, if successfully attached, will always have the default
+ * view ID kFlutterDefaultViewId.
+ *
+ *  The engine holds a weak reference to the attached view controller.
```
goderbauer, line 57: nit: remove extra space?

goderbauer (approved):
LGTM

shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
loic-sharma: Right, but from my understanding this assertion fails instead of returning `null`.

shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
loic-sharma: Good point. An assertion and maybe a quick comment seems like a good idea :)

shell/platform/darwin/macos/framework/Headers/FlutterPluginRegistrarMacOS.h
```diff
@@ -36,12 +36,18 @@ FLUTTER_DARWIN_EXPORT
...
- * The view displaying Flutter content. May return |nil|, for instance in a headless environment.
+ * The default view displaying Flutter content.
  *
- * WARNING: If/when multiple Flutter views within the same application are supported (#30701), this
- * API will change.
+ * This method may return |nil|, for instance in a headless environment.
+ *
+ * The default view is a special view operated by single-view APIs.
+ */
+- (nullable NSView*)view;
```
stuartmorgan, line 45: I don't know if it's the same as the internal breakage, but this broke `file_selector_macos`, and would potentially break any other Swift plugin as well. In Obj-C the same syntax can be used to call methods and access properties, but that's not the case in Swift, so changing a property to a method is a breaking change, which would make supporting both `master` and `stable` with a single plugin impossible. We should not change this without a *very* compelling reason.

dkwingsmt: Yes, this is exactly why this PR was reverted. I did this only because I thought it was equivalent (for objC). I'll change it back to proprty syntax.

PR #37980, opened 2022-11-30
[deps] Add `package:intl`

dcharkes:
We've added a dependency on `package:intl` in `package:dartdev`:

* https://github.com/dart-lang/sdk/commit/1dd71b412f585bbe0353bebaa44f65d6d938cd1e

`package:intl` was not being checked out in the Flutter DEPS file, causing the builds to fail.

This PR adds `package:intl` as DEPS in Flutter.

cc @devoncarew


mkustermann (approved):

devoncarew:
Thanks for the fix here!

PR #37982, opened 2022-11-30
Revert "Roll Dart SDK from 88473a3b63ff to a55712b2de3f (0 revision)"

zanderso:
Reverts flutter/engine#37973

This looks like a mistaken roll-back of Dart to a known bad commit, which is causing the engine to fail to roll into the framework.


zanderso:
I'm going to TBR this, since getting caught up again could take many hours.

aam (approved):

PR #37986, opened 2022-11-30
Full implementation of text-input-test

lin-erik:
This PR connects the text-input-view with text-input-test.cc

- Remove `text-input-realm` and substitute it with an RealmBuilder environment
- `text-input-view` doesn't implement a `TextField` that displays text, it currently only validates that we are receiving through `onPlatformMessage` and sending back the correct keystrokes


skia-gold:
Gold has detected about 72 new digest(s) on patchset 6.
View them at https://flutter-engine-gold.skia.org/cl/github/37986

shell/platform/fuchsia/flutter/tests/integration/text-input/text-input-view/lib/text_input_view.dart
```diff
@@ -8,14 +8,98 @@
...
+    // Build the scene
+    final picture = recorder.endRecording();
+    final sceneBuilder = SceneBuilder()
+      ..pushClipRect(physicalBounds)
+      ..addPicture(Offset.zero, picture)
+      ..pop();
+    window.render(sceneBuilder.build());
+  }
+
+  void platformMessage(String name, ByteData data) async {
```
akbiggs, line 83: Is there another name we can pick for this method? I'm not sure what it does from the name.

shell/platform/fuchsia/flutter/tests/integration/text-input/text-input-view/lib/text_input_view.dart
```diff
@@ -8,14 +8,98 @@
...
+        ));
+      }
+    }
+
+    window.scheduleFrame();
+  }
+
+  void _respond(test_text.KeyboardInputListenerReportTextInputRequest request) async {
+    print('text-input-view reporting keyboard input to KeyboardInputListener');
+    await _responseListener.reportTextInput(request);
```
akbiggs, line 103: So just confirming, the test will fail if it doesn't receive the correct platform message, I guess because this will time out?

Just want to make sure it won't pass if the platform message doesn't get sent.

shell/platform/fuchsia/flutter/tests/integration/text-input/text-input-view/lib/text_input_view.dart
```diff
@@ -8,14 +8,98 @@
...
+  }
+
+  void platformMessage(String name, ByteData data) async {
+    final buffer = data.buffer;
+    var list = buffer.asUint8List(data.offsetInBytes, data.lengthInBytes);
+    var decoded = utf8.decode(list);
+    var decodedJson = json.decode(decoded);
+    print('received ${name} platform message: ${decodedJson}');
+
+    if (name == "flutter/keyevent" && decodedJson["type"] == "keydown") {
```
akbiggs, line 90: Should we also have a test for the `flutter/textinput` channel? My impression was that we send messages along both `flutter/keyevent` and `flutter/textinput`.

akbiggs (approved):
Nice. @filmil FYI this sets up an integration test for text input in Flutter Engine. Because we can't depend on the Flutter Framework directly in the Engine repo, there are some changes in the approach: instead of creating a text input widget and checking that it receives the input, we just check in the Dart component that the platform message for keyevents looks like what we expect it to.

filmil:
Yay!

shell/platform/fuchsia/flutter/tests/integration/text-input/text-input-view/lib/text_input_view.dart
lin-erik: Yep, it'll hang + the test won't receive an expected output so it'll fail (see L101 in `text-input-test.cc` for the matching logic for received vs expected)

shell/platform/fuchsia/flutter/tests/integration/text-input/text-input-view/lib/text_input_view.dart
lin-erik: Renamed to `decodeAndReportPlatformMessage`

PR #37990, opened 2022-11-30
[Impeller] add --enable-impeller-3d flag to support scene experimentation

jonahwilliams:
Off by default. Analysis doesn't quite work.


jonahwilliams:
I could not find another way to do this without copying compositing. I could re-arrange compositing.dart to separate scenebuilder so that there is less duplicated code though

lib/ui/dart_ui.gni
```diff
@@ -2,26 +2,55 @@
...
-  "//flutter/lib/ui/pointer.dart",
-  "//flutter/lib/ui/semantics.dart",
-  "//flutter/lib/ui/text.dart",
-  "//flutter/lib/ui/ui.dart",
-  "//flutter/lib/ui/window.dart",
-]
+import("//flutter/impeller/tools/impeller.gni")
 
-dart_ui_path = "//flutter/lib/ui/ui.dart"
+if (!impeller_enable_3d) {
```
zanderso, line 7: You can factor out the common lists and use `+` to add the ones that are different.
```gn
_common_dart_ui_files = [
  ...
]

if () {
  dart_ui_files = _common_dart_ui_files + [
  ]
} else {
}
```

Also, think positive:
```gn
if (impeller_enable_3d) {
} else {
}
```

lib/ui/dart_ui.gni
```diff
@@ -2,26 +2,55 @@
...
+    "//flutter/lib/ui/ui.dart",
+    "//flutter/lib/ui/window.dart",
+  ]
+
+  dart_ui_path = "//flutter/lib/ui/ui.dart"
+} else {
+  dart_ui_files = [
+    "//flutter/lib/ui/annotations.dart",
+    "//flutter/lib/ui/channel_buffers.dart",
+    "//flutter/lib/ui/compositing.dart",
```
zanderso, line 35: Does this need both `compositing.dart` and `experiments/compositing_32.dart`?

lib/ui/experiments/compositing_3d.dart
```diff
@@ -0,0 +1,884 @@
...
+    double height = 0.0,
+  }) {
+    assert(offset != null, 'Offset argument was null');
+    _addPlatformView(offset.dx, offset.dy, width, height, viewId);
+  }
+
+  @FfiNative<Void Function(Pointer<Void>, Double, Double, Double, Double, Int64)>('SceneBuilder::addPlatformView', isLeaf: true)
+  external void _addPlatformView(double dx, double dy, double width, double height, int viewId);
+
+  void addModelLayer(
```
zanderso, line 821: Could `experimental/compositing_3d.dart` just contain an extension method on `SceneBuilder`, and leave the rest of this file in `compositing.dart`? (An exception to the ban on extension methods seems less bad than duplicating all the code, in this case...)

jonahwilliams: extension method no, but I could try to make this a static method. Different API on the Flutter side but that doesn't really matter for this.  Good call

jonahwilliams: oh wait, no that won't work. Needs to be an instance method

jonahwilliams: what if we leave the method here, but make it un-reachable without a separate static method?

jonahwilliams:
Updated to use static method and everything seems to work. Made the current implementation just append a transform layer to verify it works correctly

jonahwilliams:
I removed the changes to dart_ui.gni since that is only for constructing the sdk structure for analysis

zanderso (approved):
lgtm to land after branching

skia-gold:
Gold has detected about 1 new digest(s) on patchset 8.
View them at https://flutter-engine-gold.skia.org/cl/github/37990

PR #37998, opened 2022-12-01
[Impeller] Use glyph bounds to compute the max glyph size instead of font metrics

bdero:
Font metrics like top/bottom/ascend/descend are not suitable for determining the min/max coverage area of glyphs in a font. Font glyphs can and often do exceed these boundaries as part of normal behavior. (Remember Zalgo text?)

This finally fixes the long standing issue with some typefaces containing glyphs that overlap. However, we should follow up to this by packing the atlas using the per-glyph boundaries instead of the min/maxed boundary. Doing so will _substantially_ reduce the size of atlas textures.

I expect this will make some CPU metrics a little worse.

Before:
Screen Shot 2022-11-30 at 10 17 20 PM

After:
Screen Shot 2022-11-30 at 10 14 14 PM


jonahwilliams (approved):
LGTM

Do we have an API to get per-glyph metrics, or do we essentially need to reverse engineer them from knowing which glyphs are in a given run?

chinmaygarde (approved):

bdero:
> Do we have an API to get per-glyph metrics, or do we essentially need to reverse engineer them from knowing which glyphs are in a given run?

There's an method on the Skia font that allows looking up a glyph's bounds (I'm gonna start referring to it as baseline-relative coverage) given the 16 bit glyph id, but we should probably just pull them out of Skia and track baseline-relative glyph coverage as part of the glyph state that gets tracked around the rest of Typographer (e.g. the glyph pair struct).

We can also get rid of all other metrics-related state except for the point size + scale.

PR #38017, opened 2022-12-01
[Impeller] Incorporate difference clips in stencil coverage

bdero:
Culls more aggressively for cases where a difference clip intersects with the current clip coverage rect such that the coverage boundaries become smaller.

This may help reduce redundant rendering work in cases where there's a full screen platform view in the scene.


bdero:
This isn't quite right, actually. This solution would only work if the difference clip geometry is a rectangle and the the CTM is scale-only.

PR #38041, opened 2022-12-02
[ui] make vertices validation slightly cheaper

jonahwilliams:
Old godbolt compile: https://godbolt.org/z/cvdGP4vPb
New godbolt compile: https://godbolt.org/z/cGnxqWcYs

This appears to generate better code since we don't use a closure. Additional removes the `<0` check which is unnecessary because the int is read from an unsigned int list.


jonahwilliams:
@Hixie is going to fold this into a pending change he has. Closing!

PR #38055, opened 2022-12-03
Make ColorFilterLayer accept an offset

jonahwilliams:

## Pre-launch Checklist

- [ ] I signed the [CLA].

### DIFF
--- a/flow/layers/color_filter_layer.cc
+++ b/flow/layers/color_filter_layer.cc
@@ -11,22 +11,24 @@
 
 namespace flutter {
 
-ColorFilterLayer::ColorFilterLayer(std::shared_ptr<const DlColorFilter> filter)
+ColorFilterLayer::ColorFilterLayer(std::shared_ptr<const DlColorFilter> filter, const SkPoint& offset)
     : CacheableContainerLayer(
           RasterCacheUtil::kMinimumRendersBeforeCachingFilterLayer,
           true),
-      filter_(std::move(filter)) {}
+      filter_(std::move(filter)),
+      offset_(offset) {}
 
 void ColorFilterLayer::Diff(DiffContext* context, const Layer* old_layer) {
   DiffContext::AutoSubtreeRestore subtree(context);
   auto* prev = static_cast<const ColorFilterLayer*>(old_layer);
   if (!context->IsSubtreeDirty()) {
     FML_DCHECK(prev);
-    if (NotEquals(filter_, prev->filter_)) {
+    if (NotEquals(filter_, prev->filter_) || offset_ != prev->offset_) {
       context->MarkSubtreeDirty(context->GetOldLayerPaintRegion(old_layer));
     }
   }
 
+  context->PushTransform(SkMatrix::Translate(offset_.fX, offset_.fY));
   if (context->has_raster_cache()) {
     context->SetTransform(
         RasterCacheUtil::GetIntegralTransCTM(context->GetTransform()));
@@ -38,6 +40,9 @@ void ColorFilterLayer::Diff(DiffContext* context, const Layer* old_layer) {
 }
 
 void ColorFilterLayer::Preroll(PrerollContext* context) {
+  auto mutator = context->state_stack.save();
+  mutator.translate(offset_);
+
   Layer::AutoPrerollSaveLayerState save =
       Layer::AutoPrerollSaveLayerState::Create(context);
   AutoCache cache = AutoCache(layer_raster_cache_item_.get(), context,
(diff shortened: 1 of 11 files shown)


skia-gold:
Gold has detected about 37 new digest(s) on patchset 3.
View them at https://flutter-engine-gold.skia.org/cl/github/38055

lib/web_ui/lib/src/engine/html/color_filter.dart
```diff
@@ -10,15 +10,17 @@ import '../dom.dart';
...
 import 'path_to_svg_clip.dart';
 import 'shaders/shader.dart';
 import 'surface.dart';
+import 'surface_stats.dart';
 
 /// A surface that applies an [ColorFilter] to its children.
 class PersistedColorFilter extends PersistedContainerSurface
     implements ui.ColorFilterEngineLayer {
-  PersistedColorFilter(PersistedColorFilter? super.oldLayer, this.filter);
+  PersistedColorFilter(PersistedColorFilter? super.oldLayer, this.filter, this.offset);
```
jonahwilliams, line 23: @yjbanov do you have any pointers on this? I tried to base it off of the ImageFilterLayer change but this class works a bit differently

flow/layers/color_filter_layer.cc
```diff
@@ -38,6 +41,9 @@ void ColorFilterLayer::Diff(DiffContext* context, const Layer* old_layer) {
 }
 
 void ColorFilterLayer::Preroll(PrerollContext* context) {
+  auto mutator = context->state_stack.save();
+  mutator.translate(offset_);
+
```
flar, line 46: You need to adjust the paint bounds accordingly. The paint_bounds are measured in the parent coordinate system (which used to be a transform layer so it would be the agent applying the offset to the child's bounds to create its own bounds). But now that this layer is doing its own translation, it needs to adjust the bounds it reports.

flar (changes requested):

flow/layers/color_filter_layer.cc
jonahwilliams: Done, but I noticed that there is a difference in how the image filter layer and this layer behave w.r.t child bounds transformation. I assume that is because matrix filters or blur can increase the size of things beyond the offseT?

flar: The bounds should always be offset because the translation is always applied in Paint. There is a bug in image_filter_layer in that it doesn't offset the bounds if there is no filter. Do not replicate that bug here the way you have - and fix IFL while you're at it if you please.

The bounds need to match what is rendered (modulo we haven't been replicating the integral transform code into the layer Preroll methods, which has been a topic of discussion with @knopp as we migrate Diff to use the LayerStateStack).

But, this offset is always applied in Paint and should always be applied to the paint bounds in both CFL and IFL.

flar (changes requested):

jonahwilliams:
I've updated both image and color filter layers to always apply the offset in preroll and added unit tests for both cases, PTAL

flow/layers/color_filter_layer.cc
```diff
@@ -54,18 +62,23 @@ void ColorFilterLayer::Preroll(PrerollContext* context) {
...
   if (filter_) {
+    set_paint_bounds(child_bounds);
+
     context->renderable_state_flags =
         filter_->can_commute_with_opacity()
             ? LayerStateStack::kCallerCanApplyOpacity
             : 0;
+  } else {
+    // else - we can apply whatever our children can apply.
+    set_paint_bounds(child_bounds);
```
flar, line 73: This line is not conditional, why is it applied in both halves of the if?

flow/layers/color_filter_layer.cc
```diff
@@ -54,18 +62,23 @@ void ColorFilterLayer::Preroll(PrerollContext* context) {
...
   }
-  // else - we can apply whatever our children can apply.
 }
 
 void ColorFilterLayer::Paint(PaintContext& context) const {
   FML_DCHECK(needs_painting(context));
 
   auto mutator = context.state_stack.save();
+  mutator.translate(offset_);
```
flar, line 81: There is a bug in IFL that we are currently fixing where this line here causes the offset to be applied doubly when caching children. See https://github.com/flutter/engine/pull/38567 on our discussion about how the interplay of offset/integralTransform/caching should play out.

flar (changes requested):
A minor fix and then we should work with the IFL PR below to make sure we manage translation during caching correctly. I only skimmed a couple of the files on this pass pending those eventual changes.

flow/layers/color_filter_layer.cc
jonahwilliams: fixed

PR #38079, opened 2022-12-05
[ImpellerC] Print the working directory that was tried when failing to open

bdero:
Also, resolve the source language before attempting to open the directory, to prevent the source language error message from showing up incorrectly.

### DIFF
--- a/impeller/compiler/switches.cc
+++ b/impeller/compiler/switches.cc
@@ -130,10 +130,6 @@ Switches::Switches(const fml::CommandLine& command_line)
                                                       "0"))),
       entry_point(
           command_line.GetOptionValueWithDefault("entry-point", "main")) {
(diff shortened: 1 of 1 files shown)


impeller/compiler/switches.cc
```diff
@@ -196,7 +196,9 @@ bool Switches::AreValid(std::ostream& explain) const {
   }
 
   if (!working_directory || !working_directory->is_valid()) {
-    explain << "Could not figure out working directory." << std::endl;
+    explain << "Could not open the working directory: "
```
zanderso, line 199: Consider wrapping in quotes to make it clear if/when the string is empty.

zanderso (approved):
lgtm w/ small suggestion

impeller/compiler/switches.cc
bdero: Done.

PR #38085, opened 2022-12-06
Update glfw to head and imgui to the docking branch

bdero:
The GLFW backend in the ImGui docking branch relies on `glfwGetMonitorWorkarea`, which was introduced later than the hash we're running. I tried to perform this GLFW upgrade earlier when I first added ImGui as a test dependency, but ran into some trouble. Things seem to be going more smoothly this time, but we'll see if I can get it passing on all targets. 🤷‍♂️

### DIFF
--- a/DEPS
+++ b/DEPS
@@ -236,7 +236,7 @@ allowed_hosts = [
 ]
 
 deps = {
-  'src': 'https://github.com/flutter/buildroot.git' + '@' + 'decc4f1635a48b7d7aa5c25ecc5c167d35ca2fd3',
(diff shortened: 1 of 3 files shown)


skia-gold:
Gold has detected about 35 new digest(s) on patchset 3.
View them at https://flutter-engine-gold.skia.org/cl/github/38085

bdero:
It looks to me like there are a bunch of new variations of the same license caused by the years being updated in some files and not in others over time. Additionally, sometimes the licenses have a Google LLC line and sometimes they don't.

skia-gold:
Gold has detected about 72 new digest(s) on patchset 4.
View them at https://flutter-engine-gold.skia.org/cl/github/38085

PR #38092, opened 2022-12-06
Disable an extension in Xvfb to work around errors seen when running Impeller/Vulkan unit tests

jason-simmons:
See https://github.com/flutter/flutter/issues/116130


jonahwilliams (approved):
RSLGTM

iskakaushik (approved):

PR #38105, opened 2022-12-06
Roll Dart SDK from 35a9facce191 to e517487c5679 (Dart 3.0)

a-siva:
https://dart.googlesource.com/sdk.git/+log/35a9facce191..e517487c5679

2022-12-06 dart-internal-merge@dart-ci-internal.iam.gserviceaccount.com Version 3.0.0-0.0.dev


zanderso:
Looks like the dependency on `package:quiver` needs to be updated. https://github.com/google/quiver-dart/commit/c855b10d18d1aa6cb568657b065d49226882c9e1 is 3.1.0 and has been migrated to null-safety.

zanderso:
Unfortunately, it also looks like the generated FIDL bindings are not ready for sound null-safety? It might be simpler to pass `--no-sound-null-safety` to the VM in the `flutter_runner` and/or `dart_runner`. I'm not sure where to do that, but maybe it's here? https://github.com/flutter/engine/blob/main/shell/platform/fuchsia/flutter/component_v1.cc#L426

cc @akbiggs for advice.

zanderso (approved):

a-siva:
Let me update the version of Dart now and see if the tests succeed.

akbiggs:
Sorry, I forgot to hit post on my reply. We do need to update component_v2 for Flutter https://github.com/flutter/engine/blob/main/shell/platform/fuchsia/flutter/component_v2.cc#L489 as well as the Dart runner https://github.com/flutter/engine/blob/main/shell/platform/fuchsia/dart_runner/dart_runner.cc#L210

PR #38126, opened 2022-12-07
[web] Trivial fix for non-static interop JS interop class.

joshualitt:


ditman (approved):
Subtle! LGTM!

Hixie:
test-exempt: if i understand correctly, we will in due course be testing this kind of thing but for now this is not enforced.

PR #38127, opened 2022-12-07
Remove dlCanvasRecorder from flutter::PictureRecorder

flar:
Fixes: https://github.com/flutter/flutter/issues/116670

The Paragraph code was the last known use of the DisplayListCanvasRecorder that was held by the PictureRecorder/Canvas objects. Now that the text code talks directly to the DisplayListBuilder, this vestigial object can be removed in favor of allocating a DisplayListBuilder directly.

This is a deletion of code paths that are no longer used so no new unit tests are needed as long as it passes existing tests.


jason-simmons (approved):

jason-simmons:
That code path would be used if the user selected Libtxt instead of SkParagraph.

That isn't an officially supported option, but you can force it by patching the engine or by setting a hidden engine flag.

I wanted to keep this option around until the current release branch was created.  But now that it's done I've been planning to remove everything related to Libtxt soon (probably at the start of next year).
We could wait until then to remove the SkCanvas recorder.  But I'd also be fine with just doing it now.

PR #38128, opened 2022-12-07
[web] Use js_util.promiseToFuture with CanvasKitInit()

mdebbar:
Instead of dealing with the JS Promise, let `js_util` convert it to a Future for us.


hterkelsen (approved):
LGTM

PR #38130, opened 2022-12-07
[web] Make Canvaskit's malloc more useful

mdebbar:
Open up Canvaskit's malloc to be used for other types of lists, not just float32.


lib/web_ui/lib/src/engine/canvaskit/canvaskit_api.dart
```diff
@@ -1267,34 +1267,46 @@ Float32List toSkColorStops(List<double>? colorStops) {
...
-  int size,
-);
+external Object _mallocList(_NativeArrayType nativeArrayType, int size);
 
 /// Allocates a [Float32List] backed by WASM memory, managed by
 /// a [SkFloat32List].
 ///
-/// To free the allocated array use [freeFloat32List].
+/// To free the allocated array use [freeList].
 SkFloat32List mallocFloat32List(int size) {
```
yjbanov, line 1287: Let's document if `size` is in bytes or element count. Also, if it's the latter, then perhaps `length` is a more idiomatic name for the parameter (and I think `sizeInBytes` is more common for the other one).

lib/web_ui/lib/src/engine/canvaskit/canvaskit_api.dart
```diff
@@ -1267,34 +1267,46 @@ Float32List toSkColorStops(List<double>? colorStops) {
...
 
-/// Frees the WASM memory occupied by a [SkFloat32List].
+/// Frees the WASM memory occupied by a [SkFloat32List] or [SkUint32List].
 ///
 /// The [list] is no longer usable after calling this function.
 ///
 /// Use this function to free lists owned by the engine.
 @JS('window.flutterCanvasKit.Free')
-external void freeFloat32List(SkFloat32List list);
+external void freeList(_MallocObj list);
```
yjbanov, line 1305: It's strange for a public function to take a private type. Can `_MallocObj` be public?

Also, nit: avoid abbreviations - `MallocObject`.

yjbanov, line 1305: nit: do we need `List` in the name? In the future, we could extend the API to include other malloc-ed types, including non-List kinds, no?

yjbanov (approved):
lgtm

lib/web_ui/lib/src/engine/canvaskit/canvaskit_api.dart
mdebbar: I'll check what CanvasKit calls the parameter.

lib/web_ui/lib/src/engine/canvaskit/canvaskit_api.dart
mdebbar: Sure, I'll make it public.

As for the abbreviation, I used whatever name CanvasKit uses: https://github.com/google/skia/blob/0e95fd2482f8a456b2bcabf350efd51a042c3b97/modules/canvaskit/npm_build/types/index.d.ts#L786

mdebbar: We don't need `List` in the name. I'll remove it.

PR #38131, opened 2022-12-08
[const_finder] Ignore constructor invocations from generated tear-off declarations

christopherfujino:
Fixes https://github.com/flutter/flutter/issues/116807

Skip non-const locations that are:
1. within the target class (in production, this will be `IconData`
2. within a procedure that is:
  a. static
  b. a method
  c. named `_#new#tearOff`

### DIFF
--- a/tools/const_finder/lib/const_finder.dart
+++ b/tools/const_finder/lib/const_finder.dart
@@ -57,6 +57,8 @@ class _ConstVisitor extends RecursiveVisitor<void> {
   /// }
   bool inTargetClass = false;
 
+  bool inTargetTearOff = false;
+
   /// The name of the name of the class of the annotation marking classes
   /// whose constant references should be ignored.
   final String? annotationClassName;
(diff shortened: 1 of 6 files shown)


tools/const_finder/test/fixtures/.dart_tool/package_config.json
```diff
@@ -4,12 +4,12 @@
         {
             "name": "const_finder_fixtures",
             "rootUri": "../lib/",
-            "languageVersion": "2.12"
+            "languageVersion": "2.17"
```
christopherfujino, line 7: Needed in order to use constructor tear-offs

tools/const_finder/lib/const_finder.dart
```diff
@@ -58,6 +85,18 @@ class _ConstVisitor extends RecursiveVisitor<void> {
   // Avoid visiting the same constant more than once.
   final Set<Constant> _cache = LinkedHashSet<Constant>.identity();
 
+  @override
+  void visitProcedure(Procedure node) {
+    final bool isTearOff = node.isStatic &&
+        node.kind == ProcedureKind.Method &&
+        node.name.text == '_#new#tearOff';
```
zanderso, line 92: Will one of the tests below break if the compiler starts using a different name for this?

christopherfujino: Yes, the web tests will pick up new non-const locations not in the expectations

zanderso (approved):

PR #38137, opened 2022-12-08
Revert "Remove dlCanvasRecorder from flutter::PictureRecorder"

flar:
Reverts flutter/engine#38127

It looks like some internal down-stream testing was affected, but we can't figure out why it was dependent on the libtxt code. I'll hold off on this code removal until we've successfully weaned down-stream code off of dependencies.


jason-simmons:
The affected test was building the engine without defining the `FLUTTER_ENABLE_SKSHAPER` #ifdef and was therefore always selecting Libtxt

jason-simmons (approved):

flar:
So, if we remove the code that decides which shaper to instantiate then that test should work modulo golden file off-by-1s?

PR #38138, opened 2022-12-08
[Windows] Synthesize modifier keys events on pointer events

bleroux:
## Description

This PR implements modifier keys states synchronization based on pointer events.
It is the Windows implementation for https://github.com/flutter/flutter/issues/115066.
Win32 mouse events only expose flags for Shift and Control, so only those two modifiers are sync.

## Related Issue

https://github.com/flutter/flutter/issues/115066

## Tests

Adds 4 tests.


dkwingsmt (approved):
LGTM, thank you!

PR #38146, opened 2022-12-08
[Impeller] Use DrawPath instead of Rect geometry when the paint style is stroke

bdero:
Before:
Screen Shot 2022-12-08 at 2 47 28 PM

After:
Screen Shot 2022-12-08 at 2 44 19 PM


jonahwilliams (approved):
Lets add a test for this?

bdero:
Added a better playground for this to make regressions here easier to spot.

zanderso (approved):

PR #38147, opened 2022-12-09
Massage the JS interop around `didCreateEngineInitializer`

eyebrowsoffire:
Massage JS interop around `didCreateEngineInitializer` to work better with dart2wasm.


lib/web_ui/lib/src/engine/js_interop/js_loader.dart
```diff
@@ -20,11 +21,17 @@ typedef DidCreateEngineInitializerFn = void Function(FlutterEngineInitializer);
...
+class FlutterLoader {}
 
-@JS('_flutter.loader.didCreateEngineInitializer')
-external DidCreateEngineInitializerFn? get didCreateEngineInitializer;
+extension FlutterLoaderExtension on FlutterLoader {
+  external void didCreateEngineInitializer(FlutterEngineInitializer initializer);
+  bool get isAutostart => !js_util.hasProperty(this, 'didCreateEngineInitializer');
+}
+
+@JS('_flutter.loader')
```
ditman, line 33: It is not guaranteed that `_flutter.loader` exists in window (apps before flutter.js).

IIRC, if `window._flutter` is undefined, this crashes, rather than returning null or doing something sensible. You can verify by doing this instead of using `flutter.js`:

```html
<script src="main.dart.js"></script>
```

I think the solution is just to define $JS('_flutter') as a getter of something that might have a FlutterLoader, rather than defining the raw `loader` in the getter @ line 34.

lib/web_ui/lib/initialization.dart
```diff
@@ -82,7 +82,7 @@ Future<void> webOnlyWarmupEngine({
   // Should the app "autoStart"?
   bool autoStart = true;
   if (engine.flutter != null && engine.loader != null) {
-    autoStart = engine.didCreateEngineInitializer == null;
+    autoStart = engine.loader!.isAutostart;
```
ditman, line 85:
```suggestion
    autoStart = engine.loader?.isAutostart ?? false;
```

Since anything in javascript can be null at any time, the above is a little bit safer, rather than using `!`

ditman (approved):
Both `_flutter` and `loader` may be null and I think that this would crash in that case (because of the `@JS('_flutter.loader')` bit). The fix is slightly more verbose, but it's the same idea you have here!

lib/web_ui/lib/initialization.dart
eyebrowsoffire: But this isn't a JavaScript value. This is the result of `hasProperty` which is explicitly a dart boolean.

lib/web_ui/lib/src/engine/js_interop/js_loader.dart
eyebrowsoffire: I think I see what you're saying. Let me try a change.

ditman (approved):
LGTM!

lib/web_ui/lib/initialization.dart
ditman: Ah, gotcha, I totally missed the null check literally above this line (x_X)

Hixie:
test-exempt: change to make things work at all for wasm

lib/web_ui/lib/src/engine/js_interop/js_loader.dart
```diff
@@ -15,16 +16,27 @@ import 'js_promise.dart';
...
-external Object? get loader;
+@JS()
+@staticInterop
+class FlutterLoader {}
 
-@JS('_flutter.loader.didCreateEngineInitializer')
-external DidCreateEngineInitializerFn? get didCreateEngineInitializer;
+extension FlutterLoaderExtension on FlutterLoader {
+  external void didCreateEngineInitializer(FlutterEngineInitializer initializer);
+  bool get isAutostart => !js_util.hasProperty(this, 'didCreateEngineInitializer');
```
joshualitt, line 38: `autoStart`? or rename the bool above to be `autostart`?

lib/web_ui/lib/initialization.dart
```diff
@@ -91,7 +91,7 @@ Future<void> webOnlyWarmupEngine({
   } else {
     // Yield control of the bootstrap procedure to the user.
     engine.domWindow.console.debug('Flutter Web Bootstrap: Programmatic.');
-    engine.didCreateEngineInitializer!(bootstrap.prepareEngineInitializer());
+    engine.flutter!.loader!.didCreateEngineInitializer(bootstrap.prepareEngineInitializer());
```
joshualitt, line 94: Can we refactor this to make the invariants a bit clearer and reduce the number of `!` operators?
```
FlutterLoader? loader = engine.flutter?.loader;
if (loader == null || loader.isAutostart()) {
   ... autostart ...
} else {
  loader.didCreateEngineInitializer(...)
}
```

joshualitt (approved):

PR #38149, opened 2022-12-09
Don't use sync*, as it is unimplemented in dart2wasm.

eyebrowsoffire:


tools/gen_web_locale_keymap/lib/common.dart
```diff
@@ -159,22 +159,20 @@ class _StringStream {
...
 Map<String, int> _unmarshallCodeMap(_StringStream stream) {
   final int entryNum = stream.readIntAsVerbatim();
-  return Map<String, int>.fromEntries((() sync* {
-    for (int entryIndex = 0; entryIndex < entryNum; entryIndex += 1) {
-      yield MapEntry<String, int>(stream.readEventKey(), stream.readIntAsChar());
-    }
-  })());
+  return Map<String, int>.fromEntries(
+    Iterable<int>.generate(entryNum).map((final int index) =>
+      MapEntry<String, int>(stream.readEventKey(), stream.readIntAsChar())
```
joshualitt, line 164: Here and below, can we pass the generator function as the second argument to `generate`? `generate` may be lazy, but why risk it? Unless there is some reason we can't use the generator function here.

dkwingsmt (approved):
LTGM, thank you!

third_party/web_locale_keymap/lib/web_locale_keymap/key_mappings.g.dart
```diff
@@ -167,22 +168,20 @@ class _StringStream {
...
   final int entryNum = stream.readIntAsVerbatim();
-  return Map<String, int>.fromEntries((() sync* {
-    for (int entryIndex = 0; entryIndex < entryNum; entryIndex += 1) {
-      yield MapEntry<String, int>(stream.readEventKey(), stream.readIntAsChar());
-    }
-  })());
+  return Map<String, int>.fromEntries(
+    Iterable<MapEntry<String, int>>.generate(entryNum, (final int index) =>
+      MapEntry<String, int>(stream.readEventKey(), stream.readIntAsChar())
+  ));
```
Hixie, line 174: if we're looking for performance, this would probably be even faster (you'd have to benchmark it to be sure) if we used a regular c-style `for` loop.

Hixie:
test-exempt: refactor with no semantic change

third_party/web_locale_keymap/lib/web_locale_keymap/key_mappings.g.dart
kevmoo: You could/should use collection literal syntax, here!

```dart
return <String, int>{
  for (var i = 0; i < entryNum; i++)
    stream.readEventKey() : stream.readIntAsChar(),
};
```

third_party/web_locale_keymap/lib/web_locale_keymap/key_mappings.g.dart
```diff
@@ -6,7 +6,7 @@
 // DO NOT EDIT -- DO NOT EDIT -- DO NOT EDIT
 //
 // This file is auto generated by flutter/engine:flutter/tools/gen_web_keyboard_layouts based on
```
dkwingsmt, line 8: Can you help me fix this line by changing gen_web_keyboard_layouts to gen_web_locale_keymap? It was using the old livrary name. Make sure to change the tmpl too.

third_party/web_locale_keymap/lib/web_locale_keymap/key_mappings.g.dart
dkwingsmt: I never thought it could be done this clean!

yjbanov (approved):
lgtm

PR #38159, opened 2022-12-09
Fix premature LayerStateStack layer culling

flar:
The LayerStateStack introduced some aggressive clip and opacity culling to the Paint methods of some layers. Unfortunately, PlatformViews require that their Paint method is called if their Preroll method is called. Since we Preroll all layers regardless of culling, then we must Paint all PlatformViewLayers regardless of culling. The culling introduced by the LayerStateStack did not check if there was a PlatformView in the subtree before returning early, so this PR eliminates those optimizations.


jonahwilliams (approved):
LGTM

PR #38165, opened 2022-12-09
[cpp20] Fix incompatible aggregate initialization

CaseyHillers:
Upstream of a recommended Google internal patch to unblock the internal cpp20 migration.

https://github.com/flutter/flutter/issues/114625

## Pre-launch Checklist

- [X] I signed the [CLA].

### DIFF
--- a/third_party/accessibility/ax/ax_tree_update.h
+++ b/third_party/accessibility/ax/ax_tree_update.h
@@ -50,7 +50,6 @@ namespace ui {
 //        before or after an AXTreeUpdate.
 template <typename AXNodeData, typename AXTreeData>
 struct AXTreeUpdateBase {
-  AXTreeUpdateBase() = default;
   ~AXTreeUpdateBase() = default;
 
   // If |has_tree_data| is true, the value of |tree_data| should be used


cbracken:
Can you provide details on the issue this is causing? I'm assuming somewhere we're using `axtree = {foo, bar, baz}` initialisation which is raising an issue?

cbracken (commented):
Can you provide details on the issue this is causing? I'm assuming somewhere we're using `axtree = {foo, bar, baz}` initialisation which is raising an issue? In that case, we should eliminate the dtor too to make it clearly a POD type.

third_party/accessibility/ax/ax_tree_update.h
```diff
@@ -50,7 +50,6 @@ namespace ui {
 //        before or after an AXTreeUpdate.
 template <typename AXNodeData, typename AXTreeData>
 struct AXTreeUpdateBase {
-  AXTreeUpdateBase() = default;
   ~AXTreeUpdateBase() = default;
```
cbracken, line 53:
```suggestion
```

CaseyHillers:
> Can you provide details on the issue this is causing? I'm assuming somewhere we're using axtree = {foo, bar, baz} initialisation which is raising an issue? In that case, we should eliminate the dtor too to make it clearly a POD type.

cl/491751896 is the internal patch with the reasoning.

cbracken:
Sounds good. I've amended the PR description to include the reasoning.

third_party/accessibility/ax/ax_tree_update.h
```diff
@@ -50,8 +50,6 @@ namespace ui {
 //        before or after an AXTreeUpdate.
 template <typename AXNodeData, typename AXTreeData>
 struct AXTreeUpdateBase {
-  AXTreeUpdateBase() = default;
-  ~AXTreeUpdateBase() = default;
```
cbracken, line 53:
```suggestion
```

cbracken (approved):
lgtm modulo one more fix for the clang-format lint.

CaseyHillers:
> lgtm modulo one more fix for the clang-format lint.

Thanks for your help!

PR #38167, opened 2022-12-09
Some fixes for dart2wasm

eyebrowsoffire:


lib/web_ui/lib/src/engine/js_interop/js_promise.dart
```diff
@@ -6,18 +6,39 @@
...
+
+extension PromiseRejecterExtension on PromiseRejecter {
+  void reject(Object? error) => js_util.callMethod(this, 'call', <Object>[this, if (error != null) error]);
+}
 
 /// Type-safe JS Promises
 @JS('Promise')
 @staticInterop
-abstract class Promise<T> {
+abstract class Promise<T extends Object> {
```
kevmoo, line 30: js-promise values can never be null?

eyebrowsoffire: Hmmm... I think I can enable nullability here, and we'll interpret dart null as undefined (not passing the argument at all to `resolve`)

kevmoo: IIUC `extends Object` only ever really means "this will never be `null`"

eyebrowsoffire: Well we don't have any Promise objects currently that have a nullable object. And how nullability maps to JS from wasm is kind of fuzzy right now, sometimes it means JS null, sometimes it means undefined. So I was kind of trying to avoid it, but I think allowing it to be nullable and using undefined here is probably okay.

lib/web_ui/lib/src/engine/canvaskit/canvaskit_api.dart
```diff
@@ -176,9 +176,9 @@ class ColorSpace {}
 @staticInterop
 class SkWebGLContextOptions {
   external factory SkWebGLContextOptions({
-    required int antialias,
+    required double antialias,
```
joshualitt, line 179: This really shouldn't be necessary, but doesn't seem harmful.

lib/web_ui/lib/src/engine/js_interop/js_promise.dart
```diff
@@ -6,18 +6,39 @@
...
+
+extension PromiseRejecterExtension on PromiseRejecter {
+  void reject(Object? error) => js_util.callMethod(this, 'call', <Object>[this, if (error != null) error]);
+}
 
 /// Type-safe JS Promises
 @JS('Promise')
 @staticInterop
-abstract class Promise<T> {
+abstract class Promise<T extends Object?> {
```
joshualitt, line 30: Are any of these type parameters really adding value? It might make sense just to take `Object` values.

joshualitt (approved):

Hixie:
test-exempt: wasm compatibility

PR #38169, opened 2022-12-09
[web] Remove dependency on 'profiling' subfolder of canvaskit.

hterkelsen:
Removes the dependency on the `profiling/` subdirectory in the CanvasKit directory. There are several reasons to do this:

1. Crashes from CanvasKit should be very rare. In the rare case that someone crashes the CanvasKit module and gets a stacktrace from WASM, it is even rarer that the unobfuscated stacktrace would allow them to debug and fix the problem.
2. If the profile build of CanvasKit is actually required, you can still build `wasm_profile` or `wasm_debug` and use that as your local engine.

## Pre-launch Checklist

- [x] I signed the [CLA].


eyebrowsoffire (approved):

mdebbar (approved):

lib/web_ui/dev/canvaskit_roller.dart
```diff
@@ -85,8 +85,6 @@ data:
 const List<String> _canvaskitAssets = <String>[
   'canvaskit.js',
   'canvaskit.wasm',
-  'profiling/canvaskit.js',
-  'profiling/canvaskit.wasm',
```
yjbanov, line 89: My recommendation is to keep the profile build but remove all the runtime switcheroos in the engine. This way people can still profile their apps by pointing to the profile build using the `canvasKitBaseUrl` configuration parameter.

hterkelsen: Done

PR #38170, opened 2022-12-09
Fix sampler offsets

jonahwilliams:
Sampler offsets also need to be configured, as a sampled image is both an image and a sampler and they use different fields.

Additionally remove remmaping of floats. Even though this corrects the order in the shader, the logic we have for writing the floats into the storage buffer apparently accounts for this.


jonahwilliams:
@zanderso do you know what the right thing to do with the license is for the third_party dir?

zanderso:
It looks like the license check script is happy with how you have things here modulo wanting the new license added to the golden. However, since the test shaders won't appear in the engine binary, as the license check script does in other places, I think it can skip the test shaders directory.

zanderso (approved):
LGTM

PR #38230, opened 2022-12-12
Always set orientation preferences on iOS 16+

jmagman:
Previous to iOS 16, there was a hack in place to explicitly set the device orientation.  In iOS 16 there's now an API to set the orientation "preferences" (portrait only, all but upside down, etc), and the OS decides if the screen needs rotation after the preference is set, which was adopted in https://github.com/flutter/engine/pull/36874.  Instead of checking the current orientation and only setting the preference if that orientation isn't included in the preference, instead always set it and let the OS do its thing.

iOS 16 orientation code introduced in https://github.com/flutter/engine/pull/36874
Fixes https://github.com/flutter/flutter/issues/116711

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/shell/platform/darwin/ios/framework/Source/FlutterViewController.mm
+++ b/shell/platform/darwin/ios/framework/Source/FlutterViewController.mm
@@ -1544,21 +1544,17 @@ static flutter::PointerData::DeviceKind DeviceKindFromTouchType(UITouch* touch) 
           continue;
         }
         UIWindowScene* windowScene = (UIWindowScene*)scene;
-        UIInterfaceOrientationMask currentInterfaceOrientation =
-            1 << windowScene.interfaceOrientation;
-        if (!(_orientationPreferences & currentInterfaceOrientation)) {
-          [self setNeedsUpdateOfSupportedInterfaceOrientations];
-          UIWindowSceneGeometryPreferencesIOS* preference =
-              [[UIWindowSceneGeometryPreferencesIOS alloc]
-                  initWithInterfaceOrientations:_orientationPreferences];
-          [windowScene
-              requestGeometryUpdateWithPreferences:preference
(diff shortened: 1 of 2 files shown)


shell/platform/darwin/ios/framework/Source/FlutterViewController.mm
```diff
@@ -1544,21 +1544,17 @@ - (void)performOrientationUpdate:(UIInterfaceOrientationMask)new_preferences {
           continue;
         }
         UIWindowScene* windowScene = (UIWindowScene*)scene;
-        UIInterfaceOrientationMask currentInterfaceOrientation =
-            1 << windowScene.interfaceOrientation;
-        if (!(_orientationPreferences & currentInterfaceOrientation)) {
```
jmagman, line 1549: These are the lines that were removed.  Always set `-requestGeometryUpdateWithPreferences:` if the preference mask changes since the OS should handle orientation based on those preferences.

hellohuanlin: Hmmmm, will this fix the issue?   

```
if (_orientationPreferences & currentInterfaceOrientation) // removed `!`
```

jmagman: If the current preference is "portrait" and the current orientation is "portrait", and the preference is updated to "all orientations", then we always need to set the preference so as soon as the user rotates to landscape, it works.  Likewise if the preference is "all orientations" and the current orientation is "portrait", and the preference is updated to "just portrait" then we also need to set it so that when the user tries to rotate to landscape it stays in portrait.

That's the bug this is fixing--it was only updating the preference based on the current orientation.

hellohuanlin (approved):

PR #38253, opened 2022-12-13
[Impeller] Append to existing atlas if room exists, reuse texture

jonahwilliams:
How it works: keep the Skbitmap and Rectanizer on the glyph atlas context. Use these to determine if any additional glyphs fit and draw them (and only them) onto the atlas if they do. Should reduce CPU time by reducing how much software rasteriation happens. The texture is also reused which seems to be working.

https://github.com/flutter/flutter/issues/116815


jonahwilliams:
FYI @dnfield

impeller/typographer/backends/skia/text_render_context_skia.cc
```diff
@@ -98,9 +96,48 @@ static size_t PairsFitInAtlasOfSize(const FontGlyphPair::Vector& pairs,
...
+  TRACE_EVENT0("impeller", __FUNCTION__);
+  if (!rect_packer || atlas_size.IsEmpty()) {
+    return false;
+  }
+
+  // We assume that all existing glyphs will fit. After all, they fit before.
+  // The glyph_positions only contains the values for the additional glyphs
+  // from extra_pairs.
+  glyph_positions.clear();
+  glyph_positions.reserve(extra_pairs.size());
```
dnfield, line 114: Why pass this in at all? It seems strange to have this method clearing a vector it was handed.

dnfield: I see now.

Perhaps instead of clearing it, just DCHECK that it's empty?

impeller/typographer/typographer_unittests.cc
```diff
@@ -168,5 +168,41 @@ TEST_P(TypographerTest, GlyphAtlasWithLotsOfdUniqueGlyphSize) {
...
+                                TextFrameFromTextBlob(blob2));
+  ASSERT_EQ(atlas, next_atlas);
+  auto* second_texture = next_atlas->GetTexture().get();
+
+  auto new_bitmap = atlas_context->GetBitmap();
+  auto new_packer = atlas_context->GetRectPacker();
+
+  ASSERT_EQ(second_texture, first_texture);
+  ASSERT_EQ(old_bitmap, new_bitmap);
+  ASSERT_EQ(old_packer, new_packer);
```
dnfield, line 204: nit: this seems like implementation details that don't necessarily need to be tested.

All we really care about is that the texture is the same, right?

dnfield (approved):
This will increase memory usage to keep around the SkBitmap.

I'm slightly concerned about a case like showing a full screen of text that uses up just over half of the available atlas size, and then scrolling down a bit to show lots of new text (and new characters) but preserving all of the old glyphs that aren't necessarily still on the screen. But this is probably over all an improvement.

jonahwilliams:
I suspect we could get away with not keeping around the old bitmap if we could instead partially update the texture. Though we'd still end up creating a new SkBitmap each frame

impeller/typographer/backends/skia/text_render_context_skia.cc
jonahwilliams: Done

impeller/typographer/typographer_unittests.cc
jonahwilliams: Done

PR #38255, opened 2022-12-13
Reland arc cf2 (Draft for discussion)

cyanglaz:
This is a draft for discussion only. The result will be cooperated in https://github.com/flutter/engine/pull/37883.

### DIFF
--- a/shell/platform/darwin/common/framework/Source/FlutterStandardCodec.mm
+++ b/shell/platform/darwin/common/framework/Source/FlutterStandardCodec.mm
@@ -4,6 +4,8 @@
 
 #import "flutter/shell/platform/darwin/common/framework/Source/FlutterStandardCodec_Internal.h"
 
+FLUTTER_ASSERT_ARC
+
 #pragma mark - Codec for basic message channel
 
 @implementation FlutterStandardMessageCodec {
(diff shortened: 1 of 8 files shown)


shell/platform/darwin/common/framework/Source/FlutterStandardCodec.mm
```diff
@@ -387,70 +395,79 @@ - (FlutterStandardTypedData*)readTypedDataOfType:(FlutterStandardDataType)type {
...
     case FlutterStandardFieldInt64Data:
     case FlutterStandardFieldFloat32Data:
     case FlutterStandardFieldFloat64Data:
-      return [self readTypedDataOfType:FlutterStandardDataTypeForField(field)];
+      return (__bridge CFTypeRef)[codec readTypedDataOfType:FlutterStandardDataTypeForField(field)];
     case FlutterStandardFieldList: {
-      UInt32 length = [self readSize];
-      NSMutableArray* array = [NSMutableArray arrayWithCapacity:length];
+      UInt32 length = [codec readSize];
+      CFMutableArrayRef array = CFArrayCreateMutable(kCFAllocatorDefault, length, &kCFTypeArrayCallBacks);
```
gaaclarke: If you want to minimize the diff you could say:
```objc
CFMutableArrayRef array = (__bridge CFMutableArrayRef)[NSMutableArray arrayWithCapacity:length];
```

Then remove the `CFAutorelease` below.

Same thing for the dictionary below.

shell/platform/darwin/common/framework/Source/FlutterStandardCodec.mm
```diff
@@ -364,16 +366,22 @@ - (UInt32)readSize {
...
 }
 
-- (NSData*)readData:(NSUInteger)length {
+- (CFDataRef)readDataRef:(NSUInteger)length {
   _range.length = length;
   NSData* data = [_data subdataWithRange:_range];
   _range.location += _range.length;
-  return data;
+  // return (CFDataRef)CFAutorelease((__bridge CFDataRef)data);
+  return (CFDataRef)CFRetain((__bridge CFDataRef)data);
```
gaaclarke, line 374: So `data` is already on the autorelease pool here.  You should just return it directly with the bridge cast.

`return (__bridge CFDataRef)data;`

shell/platform/darwin/common/framework/Source/FlutterStandardCodec.mm
```diff
@@ -364,16 +366,22 @@ - (UInt32)readSize {
...
+}
+
+- (NSData*)readData:(NSUInteger)length {
+  return (NSData *)CFBridgingRelease([self readDataRef:length]);
 }
 
 - (NSString*)readUTF8 {
-  NSData* bytes = [self readData:[self readSize]];
-  return [[NSString alloc] initWithData:bytes encoding:NSUTF8StringEncoding];
+  CFDataRef bytes = (CFDataRef)[self readDataRef:[self readSize]];
```
gaaclarke, line 382: No need for a cast here.

shell/platform/darwin/common/framework/Source/FlutterStandardCodec.mm
```diff
@@ -364,16 +366,22 @@ - (UInt32)readSize {
...
+  NSData *data = (NSData*)CFBridgingRelease(bytes);
```
gaaclarke: Why do you do `CFBridgingRelease` here instead of just using `(__bridge)`?  `bytes` is already on the autorelease pool and `initWithData:encoding:` should retain it.  There's no need to involve ARC.  That's what we are trying to avoid =)

jmagman:
@cyanglaz can this be closed?

PR #38259, opened 2022-12-13
Generate font fallback data to be const.

eyebrowsoffire:
This is friendlier toward dart2wasm, and prevents it from generating some pathological functions that V8 chokes on.


yjbanov (approved):
lgtm

Hixie:

PR #38271, opened 2022-12-13
[web] Don't overwrite editing state with semantic updates

htoor3:
We currently handle syncing text input editing state updates in 2 places: `text_editing.dart` (non-semantic path) and `text_field.dart` (semantic mode path).  This redundancy is causing several text input issues in semantic mode, especially as it relates to text field value and selection state.  Delegating this responsibility solely to `text_editing.dart` will fix this.  

Fixes https://github.com/flutter/flutter/issues/107634
Fixes https://github.com/flutter/flutter/issues/90794

## Pre-launch Checklist

- [x] I signed the [CLA].


lib/web_ui/test/engine/semantics/text_field_test.dart
```diff
@@ -159,7 +159,6 @@ void testMain() {
       expect(domDocument.activeElement, flutterViewEmbedder.glassPaneElement);
       expect(appHostNode.activeElement, strategy.domElement);
       expect(textField.editableElement, strategy.domElement);
-      expect((textField.editableElement as dynamic).value, 'hello');
```
yjbanov, line 162: This expectation used to verify that the semantics text value gets updated when the framework begins editing a text field. I think to cover that with the new approach we need a new test (or maybe just add it to this test) that sends the text editing state through the text input channel and checks that the editing state is applied to `textField.editableElement`.

htoor3: This is a good point.  I wrote it this way because we're delegating the responsibility of editing state updates to `text_editing.dart`, so it seems like the best place to test framework -> engine editing state updates would be in `text_editing_test.dart` which has existing tests that verify that behavior. 

But is the idea here that we want to explicitly test whether that behavior applies specifically to semantic nodes?

mdebbar: I think we want to test the behavior of `text_editing.dart` when using a semantics editing strategy to make sure that the element's value is still being updated correctly.

lib/web_ui/test/engine/semantics/text_field_test.dart
```diff
@@ -188,6 +193,92 @@ void testMain() {
...
+      expect(editableElement.value, '');
+      expect(editableElement.selectionStart, 0);
+      expect(editableElement.selectionEnd, 0);
+
+      strategy.disable();
+      semantics().semanticsEnabled = false;
+    });
+
+    test(
+        'Updates editing state when receiving framework messages from the text input channel',
```
htoor3, line 231: cc @yjbanov @mdebbar

mdebbar (approved):
LGTM

nbayati (approved):
LGTM!

lib/web_ui/test/engine/semantics/text_field_test.dart
yjbanov: Perfect! Thanks.

yjbanov (approved):
lgtm

PR #38284, opened 2022-12-14
Implement ITextProvider and ITextRangeProvider for UIA

yaakovschectman:
In order for text providing nodes to properly interact with screen reader commands via UIA, we need to implement text and textedit pattern providers. As an example, this allows screen readers to narrate the character after the carat when the user moves the carat in an edit field with the arrow keys.

Unit tests for `AXPlatformNodeText(Range)ProviderWin` are modeled after those found in the Chromium source. Details for some required modification to account for differences between the two codebases, and a handful of the newly introduced test cases are currently disabled, which is tracked in [flutter/flutter#117012](https://github.com/flutter/flutter/issues/117012).

Functionality for searching for text in a text range does not yet respect the `ignore_case` option: [flutter/flutter#117013](https://github.com/flutter/flutter/issues/117013).

Part of [flutter/flutter#116219](https://github.com/flutter/flutter/issues/116219)

## Pre-launch Checklist

- [ ] I signed the [CLA].


shell/platform/common/accessibility_bridge.h
```diff
@@ -176,7 +204,7 @@ class AccessibilityBridge
   std::unordered_map<AccessibilityNodeId,
                      std::shared_ptr<FlutterPlatformNodeDelegate>>
       id_wrapper_map_;
-  ui::AXTree tree_;
+  std::unique_ptr<ui::AXTree> tree_;
```
yaakovschectman, line 207: Changed this to a pointer so we can return a pointer in the implementation for AX Tree Manager.

shell/platform/common/accessibility_bridge.cc
```diff
@@ -524,7 +533,7 @@ void AccessibilityBridge::SetTreeData(const SemanticsNode& node,
   // 1. this text field has a valid selection
   // 2. this text field doesn't have a valid selection but had selection stored
   //    in the tree.
-  if (flags & FlutterSemanticsFlag::kFlutterSemanticsFlagIsTextField) {
+  if (flags & FlutterSemanticsFlag::kFlutterSemanticsFlagIsTextField && flags & FlutterSemanticsFlag::kFlutterSemanticsFlagIsFocused) {
```
yaakovschectman, line 536: We want to only update these values for the focused node so that the text selection data points to the focused node, not to whichever text node happens to be latest in the update.

shell/platform/windows/accessibility_bridge_windows.cc
```diff
@@ -41,10 +41,13 @@ void AccessibilityBridgeWindows::OnAccessibilityEvent(
...
-      DispatchWinAccessibilityEvent(
-          win_delegate, ax::mojom::Event::kDocumentSelectionChanged);
-      break;
+    case ui::AXEventGenerator::Event::DOCUMENT_SELECTION_CHANGED: {
+        ui::AXNode::AXID focus_id = GetAXTreeData().sel_focus_object_id;
+        auto focus_delegate = GetFlutterPlatformNodeDelegateFromID(focus_id).lock();
+        DispatchWinAccessibilityEvent(
+            std::static_pointer_cast<FlutterPlatformNodeDelegateWindows>(focus_delegate), ax::mojom::Event::kDocumentSelectionChanged);
+        break;
+      }
```
yaakovschectman, line 50: Document selection change event should be fired on the node containing the changed selection, not the root.

third_party/accessibility/ax/ax_node_position_unittest.cc
```diff
@@ -1078,7 +1079,7 @@ TEST_F(AXPositionTest, GetMaxTextOffsetAndGetTextWithGeneratedContent) {
   root_1.role = ax::mojom::Role::kRootWebArea;
   root_1.child_ids = {text_field_2.id};
 
-  text_field_2.role = ax::mojom::Role::kTextField;
+  text_field_2.role = ax::mojom::Role::kGroup;
```
yaakovschectman, line 1082: Chromium uses kTextField for parent nodes containing further text-containing children. We use it to represent an editable text field; we use kGroup for general grouping nodes.

third_party/accessibility/ax/platform/ax_platform_node_win.cc
```diff
@@ -5599,10 +5601,12 @@ AXPlatformNodeWin::GetPatternProviderFactoryMethod(PATTERNID pattern_id) {
       }
       break;
 
-      // TODO(schectman): add implementations for ITextProvider and
-      // ITextRangeProvider interfaces.
-      // https://github.com/flutter/flutter/issues/114547 and
-      // https://github.com/flutter/flutter/issues/109804
+    case UIA_TextEditPatternId:
+    case UIA_TextPatternId:
+      if (IsText() || IsTextField() || data.role == ax::mojom::Role::kRootWebArea) {
```
yaakovschectman, line 5606: We do not use `kRootWebArea`, but it is expected by unittests.

shell/platform/common/accessibility_bridge.h
```diff
@@ -106,6 +109,33 @@ class AccessibilityBridge
   const std::vector<ui::AXEventGenerator::TargetedEvent> GetPendingEvents()
       const;
 
+  // AXTreeManager methods.
+
+  ui::AXNode* GetNodeFromTree(const ui::AXTreeID tree_id,
```
loic-sharma, line 114: It seems the style guide is to label each method with `// |AXTreeManager|` to declare which type these methods implement.

third_party/accessibility/ax/platform/ax_platform_node_textprovider_win.cc
```diff
@@ -0,0 +1,354 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
```
loic-sharma, line 3: This is Chromium code, right? Should these have Chromium license headers?

shell/platform/common/accessibility_bridge.cc
```diff
@@ -428,6 +436,11 @@ void AccessibilityBridge::SetBooleanAttributesFromFlutterUpdate(
       ax::mojom::BoolAttribute::kEditableRoot,
       flags & FlutterSemanticsFlag::kFlutterSemanticsFlagIsTextField &&
           (flags & FlutterSemanticsFlag::kFlutterSemanticsFlagIsReadOnly) == 0);
+
+  // TODO(schectman): figure out when we actually want this attribute set or
+  // not.
+  node_data.AddBoolAttribute(ax::mojom::BoolAttribute::kIsLineBreakingObject,
+                             true);
```
loic-sharma, line 443: Just to double-check, should this be kept?

yaakovschectman: Yeah, I probably will need to set up a new issue for this if there isn't one yet. I don't think we currently have a good way to keep track of this attribute that is used by the AX code, so this is a placeholder as I think commonly used elements usually behave as their own line wrt a11y.

shell/platform/common/accessibility_bridge.h
```diff
@@ -39,6 +41,7 @@ namespace flutter {
 class AccessibilityBridge
     : public std::enable_shared_from_this<AccessibilityBridge>,
       public FlutterPlatformNodeDelegate::OwnerBridge,
+      public ui::AXPlatformTreeManager,
```
loic-sharma, line 44: For multi-view work, we will need a `AXTree` for each view. @dkwingsmt's plan was to make each view own their own `AccessibilityBridge`. From my understanding, there should only be a single `AXPlatfromTreeManager` for all `AXTree`s. Is that correct? Does that mean that having the view own its own `AccessibilityBridge` won't be a viable solution going forward?

shell/platform/common/accessibility_bridge.h
```diff
@@ -106,6 +109,33 @@ class AccessibilityBridge
...
+
+  ui::AXTreeID GetParentTreeID() const override;
+
+  ui::AXNode* GetRootAsAXNode() const override;
+
+  ui::AXNode* GetParentNodeFromParentTreeAsAXNode() const override;
+
+  ui::AXTree* GetTree() const override;
+
+  // AXPlatformTreeManger methods.
```
loic-sharma, line 129:
```suggestion
  // AXPlatformTreeManager methods.
```

shell/platform/common/flutter_platform_node_delegate.cc
```diff
@@ -114,4 +117,50 @@ FlutterPlatformNodeDelegate::GetOwnerBridge() const {
   return bridge_;
 }
 
+gfx::NativeViewAccessible
+FlutterPlatformNodeDelegate::GetLowestPlatformAncestor() const {
+  auto bridge_ptr = bridge_.lock();
+  BASE_DCHECK(bridge_ptr);
+  auto lowest_platform_acnestor = ax_node_->GetLowestPlatformAncestor();
```
loic-sharma, line 124:
```suggestion
  auto lowest_platform_ancestor = ax_node_->GetLowestPlatformAncestor();
```

shell/platform/common/accessibility_bridge.h
yaakovschectman: The usage of the class suggests that there should be an instance of `AXPlatformTreeManager` for each `AXTreeID` that is used. If each view owns an `AccessibilityBridge` and has its own `AXTree`, this sounds consistent with its use.

third_party/accessibility/ax/ax_position.h
```diff
@@ -1316,7 +1331,8 @@ class AXPosition {
           child_position->affinity_ = ax::mojom::TextAffinity::kUpstream;
           break;
         }
-        child_position = text_position->CreateChildPositionAt(i);
+        AXPositionInstance child = text_position->CreateChildPositionAt(i);
+        child_position = std::move(child);
```
loic-sharma, line 1335: Can this be undone?

third_party/accessibility/ax/ax_position.h
```diff
@@ -1935,6 +1951,12 @@ class AXPosition {
           parent_affinity = ax::mojom::TextAffinity::kDownstream;
         }
 
+        AXPositionInstance dummy_position =
+            CreateTextPosition(tree_id, parent_id, 0, parent_affinity);
+        max_text_offset_in_parent = dummy_position->MaxTextOffset();
+        if (parent_offset > max_text_offset_in_parent) {
+          parent_offset = max_text_offset_in_parent;
+        }
```
loic-sharma, line 1959: From my understanding, this is new code that's not from Chromium? Are we okay with `third_party/accessiblity` drifting away from Chromium's implementation? Should we try to keep our fork somewhat in sync with Chromium? /cc @chunhtai

chunhtai: I am ok drifting away from chromium. I don't think it is possible to keep them in sync in the long run.

third_party/accessibility/ax/platform/ax_platform_node_textrangeprovider_win.cc
```diff
@@ -0,0 +1,1698 @@
...
+  std::vector<gfx::NativeViewAccessible> descendants;
+
+  AXPlatformNodeWin* start_anchor =
+      GetPlatformNodeFromAXNode(start()->GetAnchor());
+  AXPlatformNodeWin* end_anchor = GetPlatformNodeFromAXNode(end()->GetAnchor());
+  AXPlatformNodeWin* common_anchor = GetLowestAccessibleCommonPlatformNode();
+  if (!common_anchor || !start_anchor || !end_anchor)
+    return UIA_E_ELEMENTNOTAVAILABLE;
+
+  SAFEARRAY* safe_array = SafeArrayCreateVector(VT_UNKNOWN, 0, 0);
```
loic-sharma, line 1036: Why is this different from the Chromium implementation?

shell/platform/common/accessibility_bridge.h
loic-sharma: Gotcha, thanks!

third_party/accessibility/ax/platform/ax_platform_node_textrangeprovider_win.cc
yaakovschectman: I'll add a TODO. This would require us to implement a method to get all children nodes that are between two particular nodes on the tree in traversal order, which I am not aware of currently having.

third_party/accessibility/ax/ax_position.h
```diff
@@ -1935,6 +1950,12 @@ class AXPosition {
...
```
loic-sharma, line 1958: Could you add a comment as to why this is necessary?

third_party/accessibility/ax/platform/ax_platform_node_textprovider_win.cc
```diff
@@ -0,0 +1,354 @@
...
+  BASE_DCHECK(start->GetAnchor() == end->GetAnchor());
+
+  // SAFEARRAYs are not dynamic, so fill the visible ranges in a vector
+  // and then transfer to an appropriately-sized SAFEARRAY
+  std::vector<Microsoft::WRL::ComPtr<ITextRangeProvider>> ranges;
+
+  auto current_line_start = start->Clone();
+  while (!current_line_start->IsNullPosition() && *current_line_start < *end) {
+    auto current_line_end = current_line_start->CreateNextLineEndPosition(
+        AXBoundaryBehavior::CrossBoundary);
```
loic-sharma, line 143: Could you explain why `AXBoundaryDetection::kDontCheckInitialPosition` was removed here?

third_party/accessibility/ax/platform/ax_platform_node_textprovider_win_unittest.cc
```diff
@@ -0,0 +1,946 @@
...
+  };
+  NearestTextIndexTestData nodes[] = {
+      {text_node,
+       {{0, 0}, {2, 0}, {3, 1}, {4, 1}, {5, 2}, {8, 2}, {9, 3}, {10, 3}}},
+      {root_node,
+       {{0, 0}, {2, 0}, {3, 0}, {4, 0}, {5, 0}, {8, 0}, {9, 0}, {10, 0}}}};
+  for (auto data : nodes) {
+    if (!data.node->IsText() && !data.node->data().IsTextField()) {
+      continue;
+    }
```
loic-sharma, line 400: Is this necessary? If looks like the original test passed without this

third_party/accessibility/ax/platform/ax_platform_node_textprovider_win.cc
yaakovschectman: The method signature for our version of `CreateNextLineEndPosition` takes a parameter of a different type than Chrome's current version of the same. We take a `AXBoundaryBehavior`, while Chromium takes a struct consisting of `AXBoundaryBehavior, AXBoundaryDetection`

third_party/accessibility/ax/platform/ax_platform_node_textprovider_win_unittest.cc
yaakovschectman: I think it is because we treat the `kRootWebArea` role differently than in Chrome.

yaakovschectman:
I will attempt to split this PR into several smaller PRs with more organized commits.

PR #38304, opened 2022-12-15
Display list culling prototype based on op indices

flar:
This is a variant of https://github.com/flutter/engine/pull/38189 that uses indices of the ops rather than offsets into the buffer, to control which ops re rendered or culled. The description that follows is essentially the same as the other PR.

This PR creates the underpinnings for a mechanism that can be used to cull DisplayLists based on the indices of the rendering operations that are desired. The RTree that the DisplayList produces will return indices within its list of the rectangles that are involved in the operation and a simple vector of "op indices" maintained by the RTree accumulator can be used to translate into an op index with a simple vector lookup.

Note that only the indices of the rendering ops are required.

The following ops will cull themselves if they are not needed:

- all rendering ops (Draw...Op)
- all saveLayers with their matching restores
- all save operations with their matching restores
- all transform ops
- all clip ops

The following ops will not cull themselves and will always be executed. The tracking to prevent them from being executed would require a lot of overhead that is unlikely to save much time:

- all set/clear attribute calls, most of which assign a simple value into a field in the dispatcher, some of which may cause a shared object to be created, though.

The culling is tested in its simplest form by the included unit test.

The performance of the overhead of this mechanism has not (yet) been measured.

There is a lot of bookkeeping cruft in this PR, mostly in the form of having to add a `count_op()` method to the DisplayListCalculator to reconstruct information that is known implicitly by the DisplayListBuilder. When the bounds and rtree calculation are merged into the Builder class via https://github.com/flutter/engine/pull/34365, that cruft can be eliminated in favor of the Builder already knowing the indices of all of its ops.

The commits for this PR start with the same commits used for the previous "buffer offset" variant and then convert to using indices partway through because I implemented this variant first as a modification to those previous code changes.

### DIFF
--- a/display_list/display_list.cc
+++ b/display_list/display_list.cc
@@ -273,7 +273,8 @@ static bool CompareOps(uint8_t* ptrA,
 }
 
 void DisplayList::RenderTo(DisplayListBuilder* builder,
-                           SkScalar opacity, bool cull) {
+                           SkScalar opacity,
+                           bool cull) {
   // TODO(100983): Opacity is not respected and attributes are not reset.
   if (!builder) {
     return;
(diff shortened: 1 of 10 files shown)


flar:
Closing in favor of https://github.com/flutter/engine/pull/38429

PR #38325, opened 2022-12-15
make sure DisplayListCanvasRecorder updates clip bounds methods

flar:
Fixes: https://github.com/flutter/flutter/issues/117121

While implementing display list render culling I noticed that the clip bounds methods on DLCanvasRecorder were not set up correctly. This PR will make sure that the clip bounds are updated both for the initial cull rect and the subsequent calls to clipRect/RRect/Path.


jonahwilliams (approved):
LGTM!

PR #38335, opened 2022-12-15
[Impeller] drawAtlas blend mode.

jonahwilliams:
Complete support for all blend modes in drawAtlas. Currently investigating if 1) we can share these blend mode shaders with drawVertices 2) if we can skip the normal blends by drawing colored trianges and then blending the image.

* https://github.com/flutter/flutter/issues/108047
* https://github.com/flutter/flutter/issues/109956


impeller/entity/contents/atlas_contents.cc
```diff
@@ -123,26 +123,137 @@ bool AtlasContents::Render(const ContentContext& renderer,
...
 
   Command cmd;
   cmd.label = "DrawAtlas";
-  cmd.pipeline =
-      renderer.GetAtlasPipeline(OptionsFromPassAndEntity(pass, entity));
+  switch (blend_mode_) {
+    case BlendMode::kClear:
+      return true;
+    case BlendMode::kSource:
+      // Color only, just use vertices.
```
jonahwilliams, line 140: TODO:

impeller/entity/contents/vertices_contents.cc
```diff
@@ -30,62 +30,202 @@ void VerticesContents::SetGeometry(std::unique_ptr<VerticesGeometry> geometry) {
...
+      cmd.pipeline = renderer.GetAtlasBlendColorPipeline(
+          OptionsFromPassAndEntity(pass, entity));
+      break;
+    case BlendMode::kLuminosity:
+      cmd.pipeline = renderer.GetAtlasBlendLuminosityPipeline(
+          OptionsFromPassAndEntity(pass, entity));
+      break;
   }
-  pass.AddCommand(std::move(cmd));
+  auto src_texture = src_contents_->RenderToSnapshot(renderer, entity);
```
jonahwilliams, line 177: This almost works, or rather it works in the playground example I've set up but in a sample app the gradient starts/stops don't seem to be lined up correctly. I think I've probably either forgotten to apply a transform somewhere, or else I'm double transforming - that said, where. 😓

jonahwilliams:
Here is an example of this not quite working:

flutter_01


```dart
// Copyright 2014 The Flutter Authors. All rights reserved.
// found in the LICENSE file.
import 'dart:ui' as ui;

void main() =>
  runApp(MaterialApp(
    home: Scaffold(
      body:CustomPaint(painter: FooPainter())
    )
  ));


class FooPainter extends CustomPainter {
  final ui.Vertices vertices = ui.Vertices(
    VertexMode.triangles,
    <Offset>[
      Offset(0, 0),
      Offset(0, 100),
      Offset(100, 100),
      Offset(50, 50),
      Offset(50, 150),
      Offset(150, 150),
    ],
    colors: [
      Colors.red,
      Colors.green,
      Colors.blue,
      Colors.red,
      Colors.green,
      Colors.blue,
    ].map((x) => x.withAlpha(255)).toList(),
  );

  @override
  void paint(ui.Canvas canvas, ui.Size size) {
    canvas.drawVertices(vertices, BlendMode.srcATop, Paint()
    // ..color = Colors.black.withAlpha(100)
    ..shader = LinearGradient(colors: [
      Colors.yellow,
      Colors.green,
    ]).createShader(Rect.fromLTWH(0, 0, size.width, size.height)));
  }

  @override
    return true;
  }
}

```

cmkweber:
> Here is an example of this not quite working:
> 
> flutter_01
> 
> ```dart
> // Copyright 2014 The Flutter Authors. All rights reserved.
> // Use of this source code is governed by a BSD-style license that can be
> // found in the LICENSE file.
> import 'dart:ui' as ui;
> import 'package:flutter/material.dart';
> 
> void main() =>
>   runApp(MaterialApp(
>     home: Scaffold(
>       body:CustomPaint(painter: FooPainter())
>     )
>   ));
> 
> 
> class FooPainter extends CustomPainter {
>   final ui.Vertices vertices = ui.Vertices(
>     VertexMode.triangles,
>     <Offset>[
>       Offset(0, 0),
>       Offset(0, 100),
>       Offset(100, 100),
>       Offset(50, 50),
>       Offset(50, 150),
>       Offset(150, 150),
>     ],
>     colors: [
>       Colors.red,
>       Colors.green,
>       Colors.blue,
>       Colors.red,
>       Colors.green,
>       Colors.blue,
>     ].map((x) => x.withAlpha(255)).toList(),
>   );
> 
>   @override
>   void paint(ui.Canvas canvas, ui.Size size) {
>     canvas.drawVertices(vertices, BlendMode.srcATop, Paint()
>     // ..color = Colors.black.withAlpha(100)
>     ..shader = LinearGradient(colors: [
>       Colors.yellow,
>       Colors.green,
>     ]).createShader(Rect.fromLTWH(0, 0, size.width, size.height)));
>   }
> 
>   @override
>   bool shouldRepaint(covariant CustomPainter oldDelegate) {
>     return true;
>   }
> }
> ```

Could it be that you are creating the gradient shader at the painters size (size.width, size.height) instead of the triangles size (150, 150) ?

jonahwilliams:
Yeah, the shader was a bit out of bounds, however the biggest issues was that I was accidentally dividing twice here: https://github.com/flutter/engine/pull/38335/commits/0c23ff48a6b0bef72d6323182bffd47c181468ae which made all the UVs tiny.

jonahwilliams:
TODO: 
 * texcoordinates
 * Always treat src color as fully opaque and pull out opacity according to https://api.skia.org/classSkCanvas.html#ae971eebc0c442a0310b830a7711b8798

jonahwilliams:
I cheated a bit for src blend modes but that might have bitten me for supporting texture coordinates.

jonahwilliams:
TODO figure out if texture coordinates apply if there are no per-vertex colors (I would think they do)

jonahwilliams:
updates: texcoordinates should be treated as being in the coordinate space of whatever is being sampled and not already normalized 0-1. It should apply on src and it should apply if there are no colors, but not if there is only a solid color paint obviously

jonahwilliams:
also as an optimization we should avoid calling makeSnapshot if the paint shader just contains an image

jonahwilliams:
We need more information from the color source than is currently present in order to correctly implement texture coordinates.

jonahwilliams:
Slightly blocked by https://github.com/flutter/flutter/issues/118476, but we can also ignore texture coordinates for now

jonahwilliams:
Getting there, but the alpha is still off for atlas. First is Skia next is Impeller

flutter_01

flutter_02

impeller/entity/contents/atlas_contents.cc
```diff
@@ -81,18 +85,133 @@ const SamplerDescriptor& AtlasContents::GetSamplerDescriptor() const {
...
+  auto src_texture = renderer.MakeSubpass(
+      ISize::Ceil(size),
+      [&contents = *this, &coverage](const ContentContext& renderer,
+                                     RenderPass& pass) -> bool {
+        Entity sub_entity;
+        sub_entity.SetBlendMode(BlendMode::kSourceOver);
+        sub_entity.SetTransformation(
+            Matrix::MakeTranslation(Vector3(-coverage.origin)));
+        return contents.RenderTexture(renderer, sub_entity, pass, 1.0);
+      });
```
bdero, line 144: Rather than rendering the textures here, any Contents can be passed to a filter as a ContentsFilterInput. Doing so has some nice performance benefits -- for example, these inputs evaluate lazily, and filters won't fetch the snapshot unless it's actually needed.
Also, the `ContentsFilterInput` evaluation uses `Contents::RenderToSnapshot`, which allows contents to make optimizations that avoid  creating a new texture in some cases by deferring parts of the rendering down the filter chain via other snapshot fields, like the transform.

jonahwilliams: Makes sense, I should probably create a subcontents for both the src and dst and then I should be able to combine all of these cases into one. Perhaps that will fix the issue with destination alpha looking off, but any ideas on that?

jonahwilliams: Though for vertices, I'll still probably need to manually create the texture in order to implement texture coordinate support.

jonahwilliams: ahh yeah, re-arranging it this way looks much better. Alpha on dest looks like skia now

impeller/entity/contents/filters/blend_filter_contents.cc
```diff
@@ -109,7 +110,8 @@ static std::optional<Snapshot> AdvancedBlend(
     auto sampler = renderer.GetContext()->GetSamplerLibrary()->GetSampler({});
     FS::BindTextureSamplerDst(cmd, dst_snapshot->texture, sampler);
     blend_info.dst_y_coord_scale = dst_snapshot->texture->GetYCoordScale();
-    blend_info.dst_input_alpha = absorb_opacity ? dst_snapshot->opacity : 1.0f;
+    blend_info.dst_input_alpha =
+        absorb_opacity ? dst_snapshot->opacity : alpha.value_or(1.0f);
```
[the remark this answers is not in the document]
jonahwilliams: Done

impeller/entity/shaders/vertices.frag
```diff
@@ -4,10 +4,15 @@
 
 #include <impeller/types.glsl>
 
+uniform FragInfo {
+  float alpha;
```
bdero, line 8: Might make sense to tack this onto `position_color.vert` instead and multiply the alpha in there.

impeller/entity/contents/atlas_contents.cc
```diff
@@ -121,31 +189,103 @@ bool AtlasContents::Render(const ContentContext& renderer,
...
+  cmd.label = "AtlasColors";
+
+  auto& host_buffer = pass.GetTransientsBuffer();
+
+  VS::VertInfo vert_info;
+  vert_info.mvp = Matrix::MakeOrthographic(pass.GetRenderTargetSize()) *
+                  entity.GetTransformation();
+
+  FS::FragInfo frag_info;
+  frag_info.alpha = alpha_;
```
bdero, line 279: Ok, so looking over this, I _think_ the destination color is getting double premultiplied. First it's getting multiplied here with the `GeometryColorPipeline` , and then the same value is getting multiplied into the destination via the new alpha in the blend filter.

I think the better place to keep the `SetAlpha` business is here in this contents, and the new `SetAlpha` stuff can be removed from the blend filter.

If it still doesn't look like Skia after fixing the double premul, then I'd wonder if this alpha is supposed to be deferred to the final blend as opposed to impacting the destination color of the vertex color blend (I suspect you may have worked this out already, though).

jonahwilliams: That is what I thought, but I don't set the alpha here at all if I'm blending - its only used for the src or dst case. Double checking this

bdero (approved):
Overall LGTM!

This gets things mostly working and I don't think it should introduce any regressions, and so following up with a fix for the advanced blend case seems reasonable.

jonahwilliams:
I guess a fair question would be why even use the colors and srcOver if you know your image has no alpha.

bdero:
> Do we have any tags/data on our images to be able to tell this?

Huh, I feel like @chinmaygarde and I have briefly mentioned doing this several times, but I guess we never made a bug about it. Added one here: https://github.com/flutter/flutter/issues/118854

jonahwilliams:
It would probably useful for tracking in general. For example on the vertices blend modes, we could apply similar optimziations based on knowing if the color source has any transparency. Probably low priority compared to other optimizations though

impeller/entity/contents/filters/blend_filter_contents.cc
```diff
@@ -109,7 +110,8 @@ static std::optional<Snapshot> AdvancedBlend(
     auto sampler = renderer.GetContext()->GetSamplerLibrary()->GetSampler({});
     FS::BindTextureSamplerDst(cmd, dst_snapshot->texture, sampler);
     blend_info.dst_y_coord_scale = dst_snapshot->texture->GetYCoordScale();
-    blend_info.dst_input_alpha = absorb_opacity ? dst_snapshot->opacity : 1.0f;
+    blend_info.dst_input_alpha =
+        (absorb_opacity ? dst_snapshot->opacity : 1.0) * alpha.value_or(1.0);
```
bdero, line 114: Since this `alpha` value is no longer being used by `AtlasContents`, we could repurpose it to just forward down the filter chain by multiplying it into the returned `Snapshot::opacity`, which would allow us to remove the need for the extra subpass as mentioned in my other comment (this snapshot alpha forwarding can be supported for both the pipeline and advanced blend case with no inconsistencies/ill effects).

jonahwilliams: ahh, Yeah that is what I missed. I had a similar technique for the pipeline blends which didn't work since it would just blend the alpha with one of the inputs. Will fix so both are applied to the snapshot.

impeller/entity/shaders/vertices.frag
jonahwilliams: Let me revist this when I do DrawVertices next, since I need to adjust the contents there anyway.

impeller/entity/contents/atlas_contents.cc
```diff
@@ -78,30 +83,123 @@ const SamplerDescriptor& AtlasContents::GetSamplerDescriptor() const {
...
+                                     FilterInput::Make(src_contents)});
+
+          return filter_contents->Render(renderer, sub_entity, pass);
+        });
+    if (!subpass_texture) {
+      return false;
+    }
+    auto texture_contents = TextureContents::MakeRect(coverage);
+    texture_contents->SetTexture(subpass_texture);
+    texture_contents->SetOpacity(alpha_);
```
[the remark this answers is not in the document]
jonahwilliams: Done

impeller/entity/contents/filters/color_filter_contents.h
```diff
@@ -33,8 +33,17 @@ class ColorFilterContents : public FilterContents {
...
+  ///        inherited opacity.
+  ///
+  ///        This is only used for advanced blends.
+  void SetAlpha(Scalar alpha);
+
+  std::optional<Scalar> GetAlpha() const;
+
  private:
   bool absorb_opacity_ = false;
+  std::optional<Scalar> alpha_;
```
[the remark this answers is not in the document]
jonahwilliams: This is used now

impeller/entity/contents/content_context.h
```diff
@@ -27,8 +27,7 @@
 #include "impeller/entity/advanced_blend_saturation.frag.h"
 #include "impeller/entity/advanced_blend_screen.frag.h"
 #include "impeller/entity/advanced_blend_softlight.frag.h"
-#include "impeller/entity/atlas_fill.frag.h"
-#include "impeller/entity/atlas_fill.vert.h"
+
```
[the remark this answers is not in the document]
jonahwilliams: Done

jonahwilliams:
Filled https://github.com/flutter/flutter/issues/118903 for the blend issue, I verified by modifying the shader the the dst needs to be in the second slot for advanced blends, but the first for pipeline blends. Don't see the issue elsewhere though

jonahwilliams:
This is updated now, PTAL @bdero

jonahwilliams:
Filled https://github.com/flutter/flutter/issues/118914 for the remaing alpha issue

bdero (approved):
LGTM!

jonahwilliams:
yaaaay!

chinmaygarde:
🎉 We can update the section about this in go/impeller-wiki#status

PR #38349, opened 2022-12-16
Use DisplayListMatrixClipTracker in DisplayListBuilder

ColdPaleLight:
fix https://github.com/flutter/flutter/issues/117096

## Pre-launch Checklist

- [ ] I signed the [CLA].


ColdPaleLight:
@flar After the modification, the clip-related test in `canvas_test` failed, because the default value of `is_aa` of `Canvas.clipXXX` methods is true. In DLBuilder, `is_aa` will be ignored, but it will be processed in `DisplayListMatrixClipTracker`. 

Should we modify those tests to make them pass?

display_list/display_list_builder.cc
```diff
@@ -67,16 +67,16 @@ sk_sp<DisplayList> DisplayListBuilder::Build() {
 }
 
 DisplayListBuilder::DisplayListBuilder(const SkRect& cull_rect,
-                                       bool prepare_rtree) {
+                                       bool prepare_rtree)
+    : tracker_(cull_rect.isEmpty() ? SkRect::MakeEmpty() : cull_rect,
+               SkMatrix::I()) {
```
flar, line 72: Maybe move the NaN protection to the tracker?

display_list/display_list_builder.cc
```diff
@@ -636,11 +635,10 @@ void DisplayListBuilder::transform2DAffine(
     Push<Transform2DAffineOp>(0, 1,
                               mxx, mxy, mxt,
                               myx, myy, myt);
-    current_layer_->matrix().preConcat(SkM44(mxx, mxy,  0,  mxt,
-                                             myx, myy,  0,  myt,
-                                             0,   0,   1,   0,
-                                             0,   0,   0,   1));
-    current_layer_->update_matrix33();
+    tracker_.transform(SkM44(mxx, mxy,  0,  mxt,
```
flar, line 638: Use `transform(SkMatrix)` otherwise you promote it to an SkM44 before it needs it.

If you want to avoid scalars->SkMatrix->SkM44, then you could ask it if it is already an M44 and call either `transform(SkMatrix)` or `transform(SkM44)` accordingly.

And maybe we should have transform(6 scalars), transform(16 scalars) in the Tracker so that it can figure out which to use without the caller having to ask.

display_list/display_list_builder.cc
```diff
@@ -794,7 +769,7 @@ bool DisplayListBuilder::quickReject(const SkRect& bounds) const {
   }
   SkRect dev_bounds;
   matrix.mapRect(bounds).roundOut(&dev_bounds);
-  return !current_layer_->clip_bounds().intersects(dev_bounds);
+  return !tracker_.device_cull_rect().intersects(dev_bounds);
```
flar, line 772: Perhaps add `Tracker::device_content_culled`? Or `Tracker::device_bounds_culled`? It doesn't save much, but it reads better in the calling code.

display_list/display_list_builder.cc
```diff
@@ -753,33 +749,12 @@ void DisplayListBuilder::clipPath(const SkPath& path,
...
-}
-void DisplayListBuilder::intersect(const SkRect& rect) {
-  SkRect dev_clip_bounds = getTransform().mapRect(rect);
-  if (!current_layer_->clip_bounds().intersect(dev_clip_bounds)) {
-    current_layer_->clip_bounds().setEmpty();
-  }
-}
-SkRect DisplayListBuilder::getLocalClipBounds() {
-  SkM44 inverse;
-  if (current_layer_->matrix().invert(&inverse)) {
```
flar, line 777: I noticed that Data4x4 inverts the 3x3 part rather than the entire 4x4 matrix as this code does. I wonder if that matters and if we should file a bug against Data4x4?

display_list/display_list_builder.h
```diff
@@ -386,32 +387,21 @@ class DisplayListBuilder final : public virtual Dispatcher,
...
                        size_t save_layer_offset = 0,
                        bool has_layer = false,
                        std::shared_ptr<const DlImageFilter> filter = nullptr)
-        : LayerInfo(current_layer->matrix_,
-                    current_layer->matrix33_,
-                    current_layer->clip_bounds_,
-                    save_layer_offset,
-                    has_layer,
-                    filter) {}
+        : LayerInfo(save_layer_offset, has_layer, filter) {}
```
flar, line 404: Is this constructor even needed any more?

flar: (It doesn't use the current_layer argument)

display_list/display_list_builder.h
```diff
@@ -640,6 +622,7 @@ class DisplayListBuilder final : public virtual Dispatcher,
   DlPaint current_;
   // If |current_blender_| is set then ignore |current_.getBlendMode()|
   sk_sp<SkBlender> current_blender_;
+  DisplayListMatrixClipTracker tracker_;
```
flar, line 625: Maybe locate this near the current_layer_/layer_stack_ declarations up above (around line 500)?

display_list/display_list_matrix_clip_tracker.cc
```diff
@@ -323,7 +336,9 @@ SkRect Data3x3::local_cull_rect() const {
     // cull rect.
     return DisplayListBuilder::kMaxCullRect;
   }
-  return inverse.mapRect(cull_rect_);
+  SkRect expended_rect;
```
flar, line 339: What about the version in Data4x4?

flar (changes requested):
This looks great! Minor changes noted and fix or file an issue against Data4x4 if it is doing the inverse wrong (I didn't go through the math to see if it matters).

display_list/display_list_builder.cc
flar: Also, even if the math works out the same for the coordinates, would doing the invert on the full 4x4 fail in some circumstances that an invert of the 3x3 succeeds? Perhaps if the unused entries caused a zero determinant? Or if they were non-finite?

display_list/display_list_matrix_clip_tracker.cc
flar: This is dubious, btw, because we could be computing local bounds for something that won't yet have its final transform, so the roundOut may be too much or not enough depending on a scale that happens outside the picture recording...

flar:
> @flar After the modification, the clip-related test in `canvas_test` failed, because the default value of `is_aa` of `Canvas.clipXXX` methods is true. In DLBuilder, `is_aa` will be ignored, but it will be processed in `DisplayListMatrixClipTracker`. Please see https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8794613483671935009/+/u/test:_Host_Tests_for_host_debug/stdout?format=raw and search "Canvas.clipRect" for detail.
> 
> Should we modify those tests to make them pass? Or should we make DisplayListMatrixClipTracker ignore is_aa? Please let me know what you do you think. thanks!

This is all about whether to do a roundOut when setting an AA clip. Ugh, I can see some arguments both ways. Let me think about it some more.

flar:
One issue with roundOut is that it is assuming that it is working in device pixels. But, these measurements are relative to the environment in which the recorded DL will be displayed. In most cases that will involve a DPI adjustment that we don't see and so the roundOut is larger than it should be. But, we could just as well be compiling a huge picture that will be shown at a fractional scale - in which case roundOut won't be as large as it should be.

flar:
I think we should proceed with clip(AA) doing a roundout (ignoring the fact that we aren't really talking about pixels in the builder/recorder) and DL bounds doing a roundout for AA primitives. And not doing those things on non-AA clips/primitives.

The test should be changed to specify non-AA for the clip that it executes.

display_list/display_list_builder.cc
ColdPaleLight: Done.

display_list/display_list_builder.cc
ColdPaleLight: Done.

display_list/display_list_builder.h
ColdPaleLight: Done.l

display_list/display_list_builder.cc
ColdPaleLight: I found that SkCanvas also converts M44 to M33 and then inverse, so maybe the code in tracker should be adopted here.
c.f.
https://github.com/google/skia/blob/3724b5964eed55f5297fd5101a077f4180bfdcf5/src/core/SkCanvas.cpp#L1730

display_list/display_list_builder.cc
ColdPaleLight: It looks like we can directly call the existing `Tracker::content_culled` method.

display_list/display_list_matrix_clip_tracker.cc
ColdPaleLight: There is roundOut logic in DLBuilder. If it is not modified here, some tests in display_list_unittests.cc and canvas_test.dart will fail.

So we should change the tests instead of here, right?

flar (changes requested):
Should we duplicate the clip bounds tests and have a version that tests AA and another that tests non-AA?

display_list/display_list_builder.cc
```diff
@@ -1370,7 +1341,7 @@ void DisplayListBuilder::AccumulateOpBounds(SkRect& bounds,
 }
 void DisplayListBuilder::AccumulateBounds(SkRect& bounds) {
   getTransform().mapRect(&bounds);
-  if (bounds.intersect(current_layer_->clip_bounds())) {
+  if (bounds.intersect(tracker_.device_cull_rect())) {
```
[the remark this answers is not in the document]
ColdPaleLight: Done.

display_list/display_list_builder.cc
```diff
@@ -792,9 +764,8 @@ bool DisplayListBuilder::quickReject(const SkRect& bounds) const {
   if (!matrix.invert(nullptr)) {
     return true;
   }
-  SkRect dev_bounds;
-  matrix.mapRect(bounds).roundOut(&dev_bounds);
-  return !current_layer_->clip_bounds().intersects(dev_bounds);
+
+  return tracker_.content_culled(bounds);
```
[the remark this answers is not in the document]
ColdPaleLight: Done.

display_list/display_list_builder.cc
```diff
@@ -67,16 +67,15 @@ sk_sp<DisplayList> DisplayListBuilder::Build() {
...
   if (prepare_rtree) {
     accumulator_ = std::make_unique<RTreeBoundsAccumulator>();
   } else {
     accumulator_ = std::make_unique<RectBoundsAccumulator>();
   }
 
-  // isEmpty protects us against NaN as we normalize any empty cull rects
-  SkRect cull = cull_rect.isEmpty() ? SkRect::MakeEmpty() : cull_rect;
-  layer_stack_.emplace_back(SkM44(), SkMatrix::I(), cull);
+  layer_stack_.emplace_back(LayerInfo());
```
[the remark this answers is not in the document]
ColdPaleLight: Changed it to `layer_stack_.emplace_back();`

The DISALLOW_COPY_etc macro has not been added yet since it needs to change the restore to avoid copying, maybe this can be done in the future.

display_list/display_list_matrix_clip_tracker.cc
```diff
@@ -80,12 +81,15 @@ class Data3x3 : public DisplayListMatrixClipTracker::Data {
     FML_CHECK(false) << "SkM44 was set without upgrading Data";
   }
   void setIdentity() override { matrix_.setIdentity(); }
+  bool mapRect(const SkRect& rect, SkRect* mapped) const override {
+    return matrix_.mapRect(mapped, rect);
+  }
+  virtual bool canBeInverted() const override {
```
flar, line 87: nit - virtual not needed here (I was wondering why this line was longer than the Data4x4 version)

flar (approved):
LGTM - just a nit

display_list/display_list_matrix_clip_tracker.cc
ColdPaleLight: Ahh, I see, good catch, thanks!

flar:
Something about this change causes the DisplayList.FlutterSvgIssue661BoundsWereEmpty test to fail now on my development MBP with an M1 chip. The bounds computed and compared [here](https://github.com/flutter/engine/blob/4724a91afdc0ca8f1bdabaee9a6615cdb199fd7e/display_list/display_list_unittests.cc#L1234) are very slightly off. It looks like the new code performs computations that are sensitive to the FP differences on Apple Silicon?

PR #38359, opened 2022-12-16
Remove DisplayList's dependency on SkAutoTMalloc

jason-simmons:
SkAutoTMalloc is not a public Skia API


flar (approved):
LGTM

flar:
This is a code switch from one mechanism to another equivalent mechanism so there is no way to test the change itself. Its correctness will be tested by existing changes...

PR #38360, opened 2022-12-16
[Impeller] move Atlas geometry calculation to geometry class

jonahwilliams:
This will make it easier to have drawVertices and drawAtlas shader code (in the contents), as part of an investigation for finishing the implementation.

Part of https://github.com/flutter/engine/pull/38335


impeller/entity/geometry.h
```diff
@@ -84,6 +84,15 @@ class VerticesGeometry : public Geometry {
                                              RenderPass& pass) = 0;
 };
 
+/// @brief A geometry that is created from a draw atlas call.
+class AtlasGeometry : public Geometry {
+ public:
+  virtual GeometryResult GetPositionColorBuffer(const ContentContext& renderer,
```
jonahwilliams, line 90: Unfortunately this and vertices geometry are now _slightly_ different :(

bdero:
Since `drawAtlas` is a special case in that it doesn't apply to all `ColorSourceContents` (correct me if I'm mistaken), would it make sense to keep the whole thing as a special case separate from the `Geometry`/`ColorSource` base classes? Ie., will `AtlasGeometry` ever used in conjunction with a `Contents` other than `AtlasContents`?

DrawVertices is not as exceptional, as the given DlVertices does not imply a specific fragment rendering behavior and can apply to any `ColorSourceContents`.

jonahwilliams:
The gist of go/impeller-draw-atlas-vertices is that some of the drawAtlas and drawVertices behavior involves blending a per-vertex color with a texture. so rather than have an atlas contents and a vertices contents special case this, we can have a per-vertex color source + texture contents that can accept either atlas or vertices geometry.

Maybe this won't actually work out, but I think its probably worth a shot.

jonahwilliams:
Updated to include my idea of how to make more blend modes work. Sort of seems to work, but I need to think about the opacity and I think advanced blends need to be separate.

impeller/entity/contents/atlas_contents.cc
```diff
@@ -85,59 +64,54 @@ bool AtlasContents::Render(const ContentContext& renderer,
...
 
-  using VS = AtlasFillVertexShader;
-  using FS = AtlasFillFragmentShader;
+  auto& host_buffer = pass.GetTransientsBuffer();
 
-  const auto texture_size = texture_->GetSize();
-  if (texture_size.IsEmpty()) {
-    return true;
-  }
+  if (geometry_->GetVertexType() == GeometryVertexType::kColor) {
```
jonahwilliams, line 69: So the idea is, since the color is the dst, draw that first, then draw the image with a configured blend mode. Something like that.

bdero:
I had a bit of a thonk about this...

For the trouble "`VerticesGeometry` with colors + `TiledTextureContents`" case, one option would be to just add a `GetPositionColorUVBuffer` as part of the `Geometry` base class and add a branch in `TiledTextureContents` to handle the `GeometryVertexType::kPositionColorUV` case (and other contents, depending on how the gradients are supposed to work).

Ideally, `VerticesContents` can also be removed and entirely absorbed by the `ColorSourceContents` (that was the intent of http://flutter.dev/go/impeller-geometry, anyhow).

As an aside, I also noticed that `GetPositionColorBuffer` and `GetPositionUVBuffer` are not part of the base `Geometry` class right now, but they should be. If `TiledTextureContents`+the gradients supplies paths for `GetPositionUVBuffer/GetPositionColorUVBuffer` (`GetPositionBuffer` should not be used by these Contents -- the UVs can always be supplied or generated by the `Geometry`) and `ColorSourceContents` supplies paths for `GetPositionBuffer/GetPositionColorBuffer`, I believe this covers all of the ColorSource x Geometry cases. So just http://flutter.dev/go/impeller-geometry, except with an extra `GetPositionColorUVBuffer` that `TiledTextureContents`+gradients will use and handle separately if the vertex type happens to be `kPositionColorUV`.

These cases can just get handled as switches against the `GeometryVertexType`. This will all be much less of a pain when we have vertex layout customization on render pipelines.

impeller/entity/geometry.h
```diff
@@ -84,6 +84,14 @@ class VerticesGeometry : public Geometry {
                                              RenderPass& pass) = 0;
 };
 
+/// @brief A geometry that is created from a draw atlas call.
+class AtlasGeometry : public Geometry {
```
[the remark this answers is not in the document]
jonahwilliams: Yeah, looking at this some more there isn't really that much overlap. I still like the idea of encapsulating this outside of the contents but it should be in different type

jonahwilliams:
> Ideally, VerticesContents can also be removed and entirely absorbed by the ColorSourceContents (that was the intent of http://flutter.dev/go/impeller-geometry, anyhow).

I think we could do this, but it depends on if my hack if drawing vertices color + draw solid color with blend mode works.


> As an aside, I also noticed that GetPositionColorBuffer and GetPositionUVBuffer are not part of the base Geometry class right now, but they should be

I didn't actually find a way to use these at the time. Its starting to make more sense to me how we would do this but its taken me a little while.

jonahwilliams:
Coming back to this after break a bit refreshed: anyway some thoughts:

- Atlas and Vertices are the only contents that could use either GetPositionColorBuffer or GetPositionUVBuffer. Every other geometry/contents could only use GetPositionBuffer. I don't know how we could possibly use this API elsewhere.

- Vertices should be a Geometry, since under some configurations it can accept regular color sources, but in all other cases it should probably just have specific methods.

- Atlas shouldn't be a Geometry. Its not really compatible with tiled texture and every other behavior is separate or more similar to vertices.

Given that, I'm thinking of the following solution to round out support for drawAtlas/Vertices:

 1. Introduce either a subtype or a new interface for Vertices/Atlas that provides generic access to a PositionColor buffer or a PositionUV buffer.
   1a. Remove CPU blending of drawVertices with solid color.
 2. Introduce Vertices/Atlas contents that handles blending by drawing dst then src
 3. Introduce optimization special cases (BlendMode.src/dst/clear) and existing case where Vertices is just used as the position buffer for regular color sources.

jonahwilliams:
Its easier to start to see more similaries starting from vertices first: https://github.com/flutter/engine/pull/38644

jonahwilliams:
Closing for now

PR #38363, opened 2022-12-16
Fix signature logic in license tool

Hixie:


Hixie:
test-exempt: optimization

...the real problem here is that there's only one really valid test and that's checking that it actually works on CI. :-)

zanderso (approved):

zanderso:
Autosubmit failed on https://github.com/flutter/flutter/issues/117318. Landing manually.

PR #38365, opened 2022-12-16
Use FML_DCHECK in place of assert in codec utils

cbracken:
Even though the file is pure C code, it's useful to use a C++ or Objective-C++ filename in order to use FML (assertions) in the implementation. 

Three implementation changes:
* Since the file is now C++, some implicit void* conversions now require an explicit cast.
* A malloc/free of a buffer has been replaced with a std::vector.
* An `assert()` was changed to `FML_DCHECK` for consistency with the rest of the embedder.

No test changes since there are no semantic changes. 

Existing tests are in:
https://github.com/flutter/engine/blob/main/shell/platform/darwin/common/framework/Source/flutter_standard_codec_unittest.mm

## Pre-launch Checklist

- [X] I signed the [CLA].

### DIFF
--- a/shell/platform/darwin/common/framework/Source/FlutterStandardCodecHelper.cc
+++ b/shell/platform/darwin/common/framework/Source/FlutterStandardCodecHelper.cc
@@ -5,6 +5,16 @@
 #include "flutter/shell/platform/darwin/common/framework/Source/FlutterStandardCodecHelper.h"
 #include <stdint.h>
 
+#include <vector>
+
+#include "flutter/fml/logging.h"
+
+// The google-runtime-int lint suggests uint64_t in place of unsigned long,
+// however these functions are frequently used with NSUInteger, which is
+// defined as an unsigned long.
+//
+// NOLINTBEGIN(google-runtime-int)
+
 void FlutterStandardCodecHelperReadAlignment(unsigned long* location,
                                              uint8_t alignment) {
   uint8_t mod = *location % alignment;
@@ -25,7 +35,7 @@ void FlutterStandardCodecHelperReadBytes(unsigned long* location,
                                          void* destination,
                                          CFDataRef data) {
   CFRange range = CFRangeMake(*location, length);
-  CFDataGetBytes(data, range, destination);
+  CFDataGetBytes(data, range, static_cast<UInt8*>(destination));
   *location += length;
 }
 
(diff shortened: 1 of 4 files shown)


cbracken:
This could just as well (and perhaps more accurately) be a .cc file.

gaaclarke (commented):
The reason I named it .c is because I want to assert we don't accidentally introduce extraneous ARC code generation.  ARC code generation was found to slow down the codec so by putting as much logic into a .c file I could make sure I wasn't accidentally introducing ARC overhead.  I would prefer to keep it as .c if possible, my second preference would be a .cc file.

shell/platform/darwin/common/framework/Source/FlutterStandardCodecHelper.cc
```diff
@@ -210,12 +212,12 @@ void FlutterStandardCodecHelperWriteUTF8(CFMutableDataRef data,
     CFIndex used_length = 0;
     // UTF16 length times 3 will fit all UTF8.
     CFIndex buffer_length = length * 3;
-    UInt8* buffer = (UInt8*)malloc(buffer_length * sizeof(UInt8));
+    std::vector<UInt8> buffer;
```
gaaclarke, line 215: Can you use the constructor that takes a size instead of a separate call to reserve.  I suspect it is as fast or faster and this is highly trafficked code.

gaaclarke (changes requested):

gaaclarke (commented):
Weren't you going to change an `assert` call too?

shell/platform/darwin/common/framework/Source/FlutterStandardCodecHelper.cc
cbracken: Haha yes just sent you the same over chat.

gaaclarke: Actually, it looks like using that constructor might zero out the data.  That's not what we want.  Maybe reserve is better.

cbracken:
Was going to do the assert in a global patch that does the entire FML migration for the macOS embedder, but can stick this one in this patch.

shell/platform/darwin/common/framework/Source/FlutterStandardCodecHelper.cc
cbracken: Tha ctor will zero the buffer at creation time which the current code doesn't do. reserve just makes sure the vector's buffer is sized as requested.

gaaclarke (approved):
Thanks chris!

Hixie:

jmagman (approved):

PR #38375, opened 2022-12-17
[Impeller] Move GetSecondsElapsed to Playground

bdero:
PlaygroundTest isn't the right abstraction for time utils like this.

### DIFF
--- a/impeller/playground/playground.h
+++ b/impeller/playground/playground.h
@@ -8,6 +8,7 @@
 
 #include "flutter/fml/closure.h"
 #include "flutter/fml/macros.h"
+#include "flutter/fml/time/time_delta.h"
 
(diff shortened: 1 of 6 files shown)


impeller/playground/playground.h
```diff
@@ -9,6 +9,7 @@
 #include "flutter/fml/closure.h"
 #include "flutter/fml/macros.h"
 
+#include "fml/time/time_delta.h"
```
zanderso, line 12: nit: Group with the fml includes above and prefix with "flutter/".

zanderso (approved):
Lgtm w/ nit

impeller/playground/playground.h
bdero: Done.

PR #38393, opened 2022-12-19
Reduce the size of Overlay FlutterImageView in HC mode

Nayuta403:
fix: https://github.com/flutter/flutter/issues/117303

In HC mode, for the intersecting layers on top of the Platformview, we create an OverlayImageView display as shown below
image
But the current size of the Overlay is the size of the Layer, the rectangle where the Layer intersects the Platformview can be calculated to reduce the size of the Overlay. As shown below

image
## Pre-launch Checklist

- [x] I signed the [CLA].


shell/platform/darwin/ios/framework/Source/FlutterPlatformViews.mm
```diff
@@ -674,7 +674,8 @@ static bool ClipBoundsContainsPlatformViewBoundingRect(const SkRect& clip_bounds
       // https://github.com/flutter/flutter/issues/52510
       if (allocation_size > kMaxLayerAllocations) {
         SkRect joined_rect;
-        for (const SkRect& rect : intersection_rects) {
+        for (SkRect rect : intersection_rects) {
```
cyanglaz, line 677: I have not validated this but I believe this is handled in line 689?: `joined_rect.intersect(platform_view_rect);`

shell/platform/android/external_view_embedder/external_view_embedder.cc
```diff
@@ -145,7 +145,8 @@ void AndroidExternalViewEmbedder::SubmitFrame(
       //
       // In this case, the rects are merged into a single one that is the union
       // of all the rects.
-      for (const SkRect& rect : intersection_rects) {
+      for (SkRect rect : intersection_rects) {
+        rect.intersect(current_view_rect);
```
cyanglaz, line 149: The fix should be similar to what iOS does: `joined_rect.intersect(platform_view_rect);`

cyanglaz (changes requested):

shell/platform/darwin/ios/framework/Source/FlutterPlatformViews.mm
flar: This gets complicated.

Technically if this was any set of rects then we could end up with the situation where a bunch of rects that surrounds the view without intersecting it could join into a rect that spans the view and then intersecting it would produce a copy of the view rect. But none of them intersected the view. So, early intersection prevents them from creating that spanning space from the empty area between them.

However, in this case, the list of intersection_rects came from a query of rectangles that intersected the view rect, so every rect being joined does actually intersect it. I can't think of any set of rectangles that intersect the view rect and combine into a rectangle that contains space that would have been eliminated by an early intersection.

shell/platform/android/external_view_embedder/external_view_embedder.cc
flar: I agree with @cyanglaz here. A single intersect on the final joined rect should be enough per the comment I made in the iOS code.

flar (changes requested):

shell/platform/android/external_view_embedder/external_view_embedder.cc
Nayuta403: Yes, thanks for the response, you're right. 
Sorry, I didn't notice this on ios, but I just noticed it on Android and changed it synchronically. `joined_rect.intersect(platform_view_rect)`  it's a very good deal, I'll do it later.

Nayuta403:
@cyanglaz @flar  Thanks for the response, you're right.
Sorry, I didn't notice this on ios, but I just noticed it on Android and changed it synchronically. `joined_rect.intersect(platform_view_rect)` it's a very good deal, I'll do it later.

shell/platform/darwin/ios/framework/Source/FlutterPlatformViews.mm
Nayuta403: Done  thanks

shell/platform/android/external_view_embedder/external_view_embedder.cc
Nayuta403: Done thanks~

Hixie:
test-exempt: is tested (for some reason the script didn't detect that?)

shell/platform/android/external_view_embedder/external_view_embedder.cc
```diff
@@ -151,7 +151,7 @@ void AndroidExternalViewEmbedder::SubmitFrame(
 
       // Get the intersection rect between the current rect
       // and the platform view rect.
-      joined_rect.intersect(current_view_rect)
+      joined_rect.intersect(current_view_rect);
```
flar, line 154: This is not quite the same thing as iOS does. Here you are joining all of the sets of rects for each view all into one rect. In pseudo-code:

iOS:
```
  for (each inter-view slice/picture) {
    for (each view behind that slice) {
      query rects that overlay that underlying view
      (potentially) join them all
      intersect the (potentially joined) rects with the underlying view bounds
      inform iOS of that intersection
    }
  }
```

android:
```
  for (each inter-view slice/picture) {
    create a global join rect
    for (each view behind that slice) {
      query rects that overlay that underlying view
      join them all into the global join rect
    }
    inform Android of that intersection
  }
```

The joins are batched differently on Android. An intersection is needed on the join of the rects from a given trip through the inner loop, so you'd essentially need to create a global and a local join rect and operate on them as follows:

android:
```
  for (each inter-view slice/picture) {
    create a global join rect
    for (each view behind that slice) {
      create a local join rect
      query rects that overlay that underlying view
      join them all into the local join rect
      intersect local join rect with the view bounds
      join the local join rect into the global join rect
    }
    inform Android of the intersection in the global join rect
  }
```

flar (changes requested):

shell/platform/android/external_view_embedder/external_view_embedder.cc
Nayuta403: Oh, great, I got it. As you said, I shouldn't just calculate the intersection of one rectangle here, otherwise I will have a problem when I have multiple Platformviews. I wrote a demo to find this problem（The green rectangle should be above the two red rectangles）

image

Nayuta403: After making the change, display it correctly
image

flar (approved):
LGTM

shell/platform/android/external_view_embedder/external_view_embedder.cc
flar: Should you maybe add a test case that covers this scenario since it points out that a future maintainer might make the same mistake and we wouldn't have a current unit test that catches it.

Nayuta403:
@flar  Okay, I added tests
image
There is something wrong with ci, let me confirm

cyanglaz:
`android_external_view_embedder_unittests --repeat=2" this is the failing test command.
You need to run out/<your_host_build_dir>/android_external_view_embedder_unittests

Nayuta403:
@cyanglaz  I see the error, but I can't reproduce the test error locally right now. I didn't find `android_external_view_embedder_unittests` in `out/<your_host_build_dir>/`. 

I did it by reference to this document：https://github.com/flutter/flutter/wiki/Testing-the-engine

The command to compile the engine is
```
gclient sync -D -f && ./flutter/tools/gn --unoptimized --no-prebuilt-dart-sdk && ./flutter/tools/gn --android --unoptimized

ninja -C out/host_debug_unopt -j 10 && ninja -C out/android_debug_unopt -j 10

```
run test:
```
testing/run_tests.py --type=engine
```

anything wrong?

cyanglaz:
You should find `android_external_view_embedder_unittests` under `out/host_debug_unopt`

Nayuta403:
@cyanglaz Thanks, I finally found it, the tests have been added again, welcome to review

cyanglaz (approved):
LGTM

PR #38407, opened 2022-12-19
[WIP] Windows DirectComposition PlatformViews

ds84182:
WIP. Lots of commented out code. Does not handle input, just visuals.

- [ ] Land needed buildroot changes for d3dallocator and spirv_cross
- [ ] Update Windows embedder template to disable the redirection bitmap on the root window where nothing is drawn
- [ ] Allow WindowSurface to control the presence of the redirection bitmap, ANGLE and DX12 do not need it but SW renderer does
- [ ] Separate compositor from WindowSurface, reuse with ANGLE
- [ ] Improve somewhat risky DirectComposition Visual reordering code

flutter/flutter#108486

## Pre-launch Checklist

- [ ] I signed the [CLA].


Hixie:
Closing for now, @ds84182 is not actively working on this at this time.

yaakovschectman:
@ds84182 We are looking into platform views for Windows at the moment. Would you be able to provide a high-level overview of how these changes could be used, or how we would interface with them from the framework and/or application side?

ds84182:
> @ds84182 We are looking into platform views for Windows at the moment. Would you be able to provide a high-level overview of how these changes could be used, or how we would interface with them from the framework and/or application side?

Unfortunately these changes aren't a good base for implementing Platform Views. While it definitely does function, there are some unsolved issues regarding input. It is also wildly inefficient (high memory usage, memory bandwidth, etc.). Flutter needs better integration with DirectComposition to mitigate this.

See flutter/flutter#108486 and flutter/flutter#124457

yaakovschectman:
For now, I think we need to investigate approaches we can use going forward for platform views on Windows. If we consider using DirectComposition for this, I think this PR would be useful as a resource to demonstrate how to incorporate it. Do you disagree as it may be too misleading toward an inadequate solution? If not, would you be able to explain how we would have been able to use this?

ds84182:
> For now, I think we need to investigate approaches we can use going forward for platform views on Windows. If we consider using DirectComposition for this, I think this PR would be useful as a resource to demonstrate how to incorporate it. Do you disagree as it may be too misleading toward an inadequate solution? If not, would you be able to explain how we would have been able to use this?

Sure, this PR is a decent starting point. The resulting performance is way less than ideal and I would not recommend anyone use this for shipping an application intended to be used on any Windows laptop or tablet.

On the embedder side, the embedder would be responsible for providing a DirectComposition visual for what it wants to draw (this does not have to be a window, it can be a video surface). It may also supply a HWND to position behind the visual to accept input. The HWND can exist in a separate message loop within the same process, or inside another process. The same is true for visuals.

There is a major caveat with this: Win32 windows cannot be transformed. They are restricted to axis-aligned rectangles. Platform Views with an associated HWND will not be interactive when their visual undergoes a non-trivial transform. Translation and scale are fine, and they are pixel snapped.

Another caveat is lack of gesture input disambiguation for touch and pen. This one is described in more detail within the associated Github issue. It is impossible for Windows Platform Views to attain parity with the Android & iOS implementations due to this. Any attempts to workaround this will not go well, at all. I do not recommend even attempting it.

While Windows has an API for pointer routing, but I have not figured out how to make it work yet. See [IDirectManipulationCompositor2::AddContentWithCrossProcessChaining](https://learn.microsoft.com/en-us/windows/win32/api/directmanipulation/nf-directmanipulation-idirectmanipulationcompositor2-addcontentwithcrossprocesschaining).

yaakovschectman:
Am I correct that the plugin would register the visual using `PlatformViewRegistrar::Register`? If so, once the visual is registered, what would be needed in order to display it within the application? I think we would still need to create a new widget type to represent this, in which case, how would the widget need to interact with the `PlatformViewWindows` instance to render it?

PR #38429, opened 2022-12-20
Display list R-Tree culling

flar:
Fixes: https://github.com/flutter/flutter/issues/92740

This PR uses an R-Tree (when computed by the Builder) to cull the dispatching of a DisplayList similar to what is done in SkPicture objects when constructed with a bbox factory.

Some additional changes:
- DlRTree no longer depends on the Skia bbox hierarchy base class
- DlRTree now uses a more streamlined approach to managing its data now that it is not relying on a generic superclass
- basic tests of R-Tree culling added, and some unit tests. Is something more sophisticated warranted?
- the code that tags rectangles with the rendering op index is currently executed post-`Push` which means the rectangles need to be tagged with `op_index_ - 1`. I eventually want to move the bounds calculations to before the `Push` calls and then potentially use them to proactively cull operations that will never be visible due to clipping.

The basic principle here is to add the "index" of every rendering op to the DisplayList R-Tree. At rendering time, if the DL is a super-set of the current clip bounds then an R-Tree search is done on those bounds and only the indicated rendering ops are executed. Also executed are all attribute ops, any saveLayer/restore enclosing a chosen rendering op, and any transform/clip operations that will be followed by a matched rendering op before their next restore.

Previous PR prototypes (now obsolete):

- https://github.com/flutter/engine/pull/38189 (based on buffer offsets)
- https://github.com/flutter/engine/pull/38304 (based on rendering op indices upon which this PR is based)

### DIFF
--- a/display_list/display_list.cc
+++ b/display_list/display_list.cc
@@ -38,18 +38,23 @@ DisplayList::DisplayList(DisplayListStorage&& storage,
       op_count_(op_count),
       nested_byte_count_(nested_byte_count),
       nested_op_count_(nested_op_count),
+      unique_id_(next_unique_id()),
       bounds_(bounds),
       can_apply_group_opacity_(can_apply_group_opacity),
-      rtree_(std::move(rtree)) {
-  static std::atomic<uint32_t> next_id{1};
-  do {
-    unique_id_ = next_id.fetch_add(+1, std::memory_order_relaxed);
-  } while (unique_id_ == 0);
-}
+      rtree_(std::move(rtree)) {}
 
 DisplayList::~DisplayList() {
   uint8_t* ptr = storage_.get();
   DisposeOps(ptr, ptr + byte_count_);
+}
+
+uint32_t DisplayList::next_unique_id() {
+  static std::atomic<uint32_t> next_id{1};
+  uint32_t id;
+  do {
+    id = next_id.fetch_add(+1, std::memory_order_relaxed);
+  } while (id == 0);
+  return id;
 }
 
 class Culler {
@@ -107,7 +112,7 @@ void DisplayList::Dispatch(Dispatcher& ctx) const {
   uint8_t* ptr = storage_.get();
   Dispatch(ctx, ptr, ptr + byte_count_, NopCuller::instance);
 }
-void DisplayList::Dispatch(Dispatcher& ctx, const SkRect& cull_rect) {
+void DisplayList::Dispatch(Dispatcher& ctx, const SkRect& cull_rect) const {
   if (cull_rect.isEmpty()) {
     return;
   }
@@ -255,7 +260,8 @@ static bool CompareOps(uint8_t* ptrA,
   return true;
 }
 
-void DisplayList::RenderTo(DisplayListBuilder* builder, SkScalar opacity) {
+void DisplayList::RenderTo(DisplayListBuilder* builder,
+                           SkScalar opacity) const {
   // TODO(100983): Opacity is not respected and attributes are not reset.
   if (!builder) {
     return;
@@ -267,7 +273,7 @@ void DisplayList::RenderTo(DisplayListBuilder* builder, SkScalar opacity) {
   }
 }
 
-void DisplayList::RenderTo(SkCanvas* canvas, SkScalar opacity) {
+void DisplayList::RenderTo(SkCanvas* canvas, SkScalar opacity) const {
   DisplayListCanvasDispatcher dispatcher(canvas, opacity);
   if (has_rtree()) {
     Dispatch(dispatcher, canvas->getLocalClipBounds());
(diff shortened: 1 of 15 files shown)


display_list/display_list_unittests.cc
```diff
@@ -1232,11 +1232,11 @@ TEST(DisplayList, FlutterSvgIssue661BoundsWereEmpty) {
   // improved and maintained, then we can eliminate this test and
   // just rely on the "rounded out" bounds test that follows.
   EXPECT_EQ(display_list->bounds(),
-            SkRect::MakeLTRB(0, 0.00189208984375, 99.9839630127, 100));
+            SkRect::MakeLTRB(0, 0.00190162658691406, 99.9839553833008, 100));
```
flar, line 1235: A note to self: Figure out why these bounds changed subtly. Nothing about the new code should have caused these bounds to change AFAICT, but the test doesn't pass without this edit.

flar: It looks like this change is causing failures in the presubmit checks. I'm double checking, but it looks like my machine will fail this test without this change to the expected rect in this test, but the test machines are perfectly happy with the old values. This may be an M1 vs Intel discrepancy that we'll have to get to the bottom of or work around it.

flar: Fixed in https://github.com/flutter/engine/pull/38435

flar:
@endless7 does this solution answer your concerns about the correlation of ops to bounds in the R-Tree?

display_list/display_list_rtree.cc
```diff
@@ -8,43 +8,166 @@
...
 namespace flutter {
 
-DlRTree::DlRTree() : bbh_{SkRTreeFactory{}()}, all_ops_count_(0) {}
+DlRTree::DlRTree(const SkRect rects[],
+                 int N,
+                 const int ids[],
+                 bool p(int),
+                 int invalid_id)
+    : leaf_count_(0), invalid_id_(invalid_id) {
+  if (N <= 0) {
```
flar, line 17: I've determined empirically that 1952257857 is the largest number of leaves we can deal with without the number of nodes overflowing an int. I could switch this all to uint32 or uint64 internally or simply reject the construction of an R-Tree with more leaves than that.

Thoughts?

flar: It looks like 2^31 leaves generates less than 2^32 total nodes due to the reduction by 11 on every generation of nodes, so using uint32_t internally might fix this overflow potential.

flar: fixed using `uint32_t` in latest commit for node indexing past the leaf counting stage...

display_list/display_list.cc
```diff
@@ -166,18 +255,25 @@ static bool CompareOps(uint8_t* ptrA,
...
   }
-  Dispatch(*builder);
+  if (has_rtree()) {
+    Dispatch(*builder, builder->getLocalClipBounds());
+  } else {
+    Dispatch(*builder);
+  }
 }
 
-void DisplayList::RenderTo(SkCanvas* canvas, SkScalar opacity) const {
```
ColdPaleLight, line 178: nit: We can continue to mark these "Dispatch/RenderTo" methods const, if we mark `bounds()`, `rtree()`, `has_rtree()` as const

display_list/display_list.cc
```diff
@@ -52,17 +52,105 @@ DisplayList::~DisplayList() {
...
+ public:
+  virtual bool init(DispatchContext& context) = 0;
+  virtual void update(DispatchContext& context) = 0;
+};
+class NopCuller : public Culler {
+ public:
+  static NopCuller instance;
+
+  bool init(DispatchContext& context) override {
+    context.next_render_index = 0;
```
ColdPaleLight, line 65: nit: It would be helpful to potential maintainers in the future if a comment can be added here to explain why `next_render_index` is set to 0 here.

ColdPaleLight (approved):
Fantastic! LGTM!

display_list/display_list.cc
flar: Done. I made pretty much all fields and methods const.

display_list/display_list.cc
flar: Done. I added comments in a few places where DispatchContext is defined and modified.

flar:
This PR shows a net gain in the builder benchmarks for the kRtree and kBoundsAndRtree sub-measurements:

https://flutter-engine-perf.skia.org/e/?begin=1671661938&end=1671754421&keys=X789f7ff76f30f8ccc672464f335fe09b&num_commits=50&request_type=1&xbaroffset=31974

endless7:
> @endless7 does this solution answer your concerns about the correlation of ops to bounds in the R-Tree?

I got it. Thanks!!

PR #38435, opened 2022-12-21
Fix floating point rounding issue in unit test bounds check

flar:
A DisplayList unit test started failing on M1 Macs after https://github.com/flutter/engine/pull/38349 most likely due to a slight difference in the way that Apple Silicon performs some floating point operations.

Since the exact floating point precision bounds aren't a required part of the test (the bug being tested was that the bounds used to be completely empty), I switched the test to use a min/max bounds that sandwich the desired output bounds so that the test would be resistant to changes in floating point computations.


ColdPaleLight (approved):
LGTM

PR #38446, opened 2022-12-21
[web] Delete web_ui/BUILD.gn

mdebbar:
Is this being used anywhere?


eyebrowsoffire (approved):
Yes, I think this is unused. LGTM, thanks for catching this!

PR #38453, opened 2022-12-21
Remove single view assumptions from `window.dart`

a-wallen:
Removes `FlutterView` to support multi-window.

Fixes https://github.com/flutter/flutter/issues/112204

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/lib/ui/platform_dispatcher.dart
+++ b/lib/ui/platform_dispatcher.dart
@@ -226,6 +226,7 @@ class PlatformDispatcher {
     }
     _viewConfigurations[id] = previousConfiguration.copyWith(
       view: _views[id],
+      window: _views[id],
       devicePixelRatio: devicePixelRatio,
       geometry: Rect.fromLTWH(0.0, 0.0, width, height),
       viewPadding: WindowPadding._(
@@ -1349,7 +1350,7 @@ class ViewConfiguration {
   final double devicePixelRatio;
 
   /// The geometry requested for the view on the screen or within its parent
-  /// view, in logical pixels.
+  /// window, in logical pixels.
   final Rect geometry;
 
   /// Whether or not the view is currently visible on the screen.
--- a/lib/ui/platform_dispatcher.dart
+++ b/lib/ui/platform_dispatcher.dart
@@ -222,7 +222,7 @@ class PlatformDispatcher {
     final ViewConfiguration previousConfiguration =
         _viewConfigurations[id] ?? const ViewConfiguration();
     if (!_views.containsKey(id)) {
-      _views[id] = FlutterWindow._(id, this);
+      _views[id] = FlutterView._(id, this);
     }
     _viewConfigurations[id] = previousConfiguration.copyWith(
       window: _views[id],
--- a/lib/ui/platform_dispatcher.dart
+++ b/lib/ui/platform_dispatcher.dart
@@ -1291,11 +1291,11 @@ class PlatformConfiguration {
 class ViewConfiguration {
   /// A const constructor for an immutable [ViewConfiguration].
   const ViewConfiguration({
-    this.view,
+    FlutterView? view,
     @Deprecated('''
       Use the `view` property instead.
     ''')
-    this.window,
+    FlutterView? window,
     this.devicePixelRatio = 1.0,
     this.geometry = Rect.zero,
     this.visible = false,
@@ -1305,7 +1305,8 @@ class ViewConfiguration {
     this.padding = WindowPadding.zero,
     this.gestureSettings = const GestureSettings(),
     this.displayFeatures = const <DisplayFeature>[],
-  }) : assert(!(window != null && view != null));
+  }) : assert(!(window != null && view != null)),
+    _view = view ?? window;
 
   /// Copy this configuration with some fields replaced.
   ViewConfiguration copyWith({
@@ -1323,7 +1324,6 @@ class ViewConfiguration {
   }) {
     return ViewConfiguration(
       view: view ?? this.view,
-      window: window ?? this.window,
       devicePixelRatio: devicePixelRatio ?? this.devicePixelRatio,
       geometry: geometry ?? this.geometry,
       visible: visible ?? this.visible,
@@ -1343,8 +1343,10 @@ class ViewConfiguration {
   @Deprecated('''
       Use the `view` property instead.
   ''')
-  final FlutterView? window;
-  final FlutterView? view;
+  FlutterView? get window => _view;
+  FlutterView? get view => _view;
+
+  final FlutterView?  _view;
 
   /// The pixel density of the output surface.
   final double devicePixelRatio;
--- a/lib/ui/platform_dispatcher.dart
+++ b/lib/ui/platform_dispatcher.dart
@@ -1290,8 +1290,8 @@ class PlatformConfiguration {
 class ViewConfiguration {
   /// A const constructor for an immutable [ViewConfiguration].
   ///
-  /// When constructing a view configuration, supply either the `view` or the
-  /// `window` property, but not both since the `view` and `window` property
+  /// When constructing a view configuration, supply either the [view] or the
+  /// [window] property, but not both since the [view] and [window] property
   /// are backed by the same instance variable.
   const ViewConfiguration({
     FlutterView? view,
--- a/lib/ui/platform_dispatcher.dart
+++ b/lib/ui/platform_dispatcher.dart
@@ -1344,6 +1344,11 @@ class ViewConfiguration {
       Use the `view` property instead.
   ''')
   FlutterView? get window => _view;
+
+  /// The top level view into which the view is placed and its geometry is
+  /// relative to.
+  ///
+  /// If null, then this configuration represents a top level view itself.
   FlutterView? get view => _view;
 
   final FlutterView?  _view;
--- a/lib/ui/platform_dispatcher.dart
+++ b/lib/ui/platform_dispatcher.dart
@@ -1309,7 +1309,7 @@ class ViewConfiguration {
     this.padding = WindowPadding.zero,
     this.gestureSettings = const GestureSettings(),
     this.displayFeatures = const <DisplayFeature>[],
-  }) : assert(!(window != null && view != null)),
+  }) : assert(window == null || view == null),
     _view = view ?? window;
 
   /// Copy this configuration with some fields replaced.
--- a/lib/ui/platform_dispatcher.dart
+++ b/lib/ui/platform_dispatcher.dart
@@ -1329,7 +1329,7 @@ class ViewConfiguration {
   }) {
     assert(view == null || window == null);
     return ViewConfiguration(
-      view: view ?? this.view,
+      view: view ?? window ?? _view,
       devicePixelRatio: devicePixelRatio ?? this.devicePixelRatio,
       geometry: geometry ?? this.geometry,
       visible: visible ?? this.visible,
--- a/lib/ui/platform_dispatcher.dart
+++ b/lib/ui/platform_dispatcher.dart
@@ -1290,6 +1290,10 @@ class PlatformConfiguration {
 /// An immutable view configuration.
 class ViewConfiguration {
   /// A const constructor for an immutable [ViewConfiguration].
+  ///
+  /// When constructing a view configuration, supply either the `view` or the
+  /// `window` property, but not both since the `view` and `window` property
+  /// are backed by the same instance variable.
   const ViewConfiguration({
     FlutterView? view,
     @Deprecated('''
--- a/lib/ui/platform_dispatcher.dart
+++ b/lib/ui/platform_dispatcher.dart
@@ -225,7 +225,7 @@ class PlatformDispatcher {
       _views[id] = FlutterView._(id, this);
     }
     _viewConfigurations[id] = previousConfiguration.copyWith(
-      window: _views[id],
+      view: _views[id],
       devicePixelRatio: devicePixelRatio,
       geometry: Rect.fromLTWH(0.0, 0.0, width, height),
       viewPadding: WindowPadding._(
@@ -1290,7 +1290,7 @@ class PlatformConfiguration {
 class ViewConfiguration {
   /// A const constructor for an immutable [ViewConfiguration].
   const ViewConfiguration({
-    this.window,
+    this.view,
     this.devicePixelRatio = 1.0,
     this.geometry = Rect.zero,
     this.visible = false,
@@ -1304,7 +1304,7 @@ class ViewConfiguration {
 
   /// Copy this configuration with some fields replaced.
   ViewConfiguration copyWith({
-    FlutterView? window,
+    FlutterView? view,
     double? devicePixelRatio,
     Rect? geometry,
     bool? visible,
@@ -1316,7 +1316,7 @@ class ViewConfiguration {
     List<DisplayFeature>? displayFeatures,
   }) {
     return ViewConfiguration(
-      window: window ?? this.window,
+      view: view ?? this.view,
       devicePixelRatio: devicePixelRatio ?? this.devicePixelRatio,
       geometry: geometry ?? this.geometry,
       visible: visible ?? this.visible,
@@ -1333,7 +1333,7 @@ class ViewConfiguration {
   /// relative to.
   ///
   /// If null, then this configuration represents a top level view itself.
-  final FlutterView? window;
+  final FlutterView? view;
 
   /// The pixel density of the output surface.
   final double devicePixelRatio;
--- a/lib/ui/platform_dispatcher.dart
+++ b/lib/ui/platform_dispatcher.dart
@@ -1316,6 +1316,11 @@ class ViewConfiguration {
   /// Copy this configuration with some fields replaced.
   ViewConfiguration copyWith({
     FlutterView? view,
+    @Deprecated('''
+      Use the `view` property instead.
+      This change is related to adding multi-view support in Flutter.
+      This feature was deprecated after 3.7.0-1.2.pre.
+    ''')
     FlutterView? window,
     double? devicePixelRatio,
     Rect? geometry,
--- a/lib/ui/platform_dispatcher.dart
+++ b/lib/ui/platform_dispatcher.dart
@@ -1298,6 +1298,8 @@ class ViewConfiguration {
     FlutterView? view,
     @Deprecated('''
       Use the `view` property instead.
+      This change is related to adding multi-view support in Flutter.
+      This feature was deprecated after 3.7.0-1.2.pre.
     ''')
     FlutterView? window,
     this.devicePixelRatio = 1.0,
@@ -1345,7 +1347,9 @@ class ViewConfiguration {
   ///
   /// If null, then this configuration represents a top level view itself.
   @Deprecated('''
-      Use the `view` property instead.
+    Use the `view` property instead.
+    This change is related to adding multi-view support in Flutter.
+    This feature was deprecated after 3.7.0-1.2.pre.
   ''')
   FlutterView? get window => _view;
 
--- a/lib/ui/platform_dispatcher.dart
+++ b/lib/ui/platform_dispatcher.dart
@@ -1349,7 +1349,7 @@ class ViewConfiguration {
   final double devicePixelRatio;
 
   /// The geometry requested for the view on the screen or within its parent
-  /// window, in logical pixels.
+  /// view, in logical pixels.
   final Rect geometry;
 
   /// Whether or not the view is currently visible on the screen.
@@ -1437,7 +1437,7 @@ class ViewConfiguration {
 
   @override
   String toString() {
-    return '$runtimeType[window: $window, geometry: $geometry]';
+    return '$runtimeType[view: $view, geometry: $geometry]';
   }
 }
 
@@ -1676,7 +1676,7 @@ enum AppLifecycleState {
   /// On Android, this corresponds to an app or the Flutter host view running
   /// in the foreground inactive state.  Apps transition to this state when
   /// another activity is focused, such as a split-screen app, a phone call,
-  /// a picture-in-picture app, a system dialog, or another window.
+  /// a picture-in-picture app, a system dialog, or another view.
   ///
   /// Apps in this state should assume that they may be [paused] at any time.
   inactive,
--- a/lib/ui/platform_dispatcher.dart
+++ b/lib/ui/platform_dispatcher.dart
@@ -226,7 +226,6 @@ class PlatformDispatcher {
     }
     _viewConfigurations[id] = previousConfiguration.copyWith(
       view: _views[id],
-      window: _views[id],
       devicePixelRatio: devicePixelRatio,
       geometry: Rect.fromLTWH(0.0, 0.0, width, height),
       viewPadding: WindowPadding._(
@@ -1328,6 +1327,7 @@ class ViewConfiguration {
     GestureSettings? gestureSettings,
     List<DisplayFeature>? displayFeatures,
   }) {
+    assert(view == null || window == null);
     return ViewConfiguration(
       view: view ?? this.view,
       devicePixelRatio: devicePixelRatio ?? this.devicePixelRatio,
--- a/lib/ui/platform_dispatcher.dart
+++ b/lib/ui/platform_dispatcher.dart
@@ -1305,7 +1305,7 @@ class ViewConfiguration {
     this.padding = WindowPadding.zero,
     this.gestureSettings = const GestureSettings(),
     this.displayFeatures = const <DisplayFeature>[],
-  });
+  }) : assert(!(window != null && view != null));
 
   /// Copy this configuration with some fields replaced.
   ViewConfiguration copyWith({
--- a/lib/ui/platform_dispatcher.dart
+++ b/lib/ui/platform_dispatcher.dart
@@ -1293,7 +1293,7 @@ class ViewConfiguration {
   const ViewConfiguration({
     this.view,
     @Deprecated('''
-      Renaming window to view as the class `FlutterWindow` has been deprecated.
+      Renaming window to view since `FlutterWindow` has been removed from dart::ui.
     ''')
     this.window,
     this.devicePixelRatio = 1.0,
@@ -1341,7 +1341,7 @@ class ViewConfiguration {
   ///
   /// If null, then this configuration represents a top level view itself.
   @Deprecated('''
-    Renaming window to view as the class `FlutterWindow` has been deprecated.
+    Renaming window to view since `FlutterWindow` has been removed from dart::ui.
   ''')
   final FlutterView? window;
   final FlutterView? view;
--- a/lib/ui/platform_dispatcher.dart
+++ b/lib/ui/platform_dispatcher.dart
@@ -1293,7 +1293,7 @@ class ViewConfiguration {
   const ViewConfiguration({
     this.view,
     @Deprecated('''
-      Renaming window to view since `FlutterWindow` has been removed from dart::ui.
+      Use the `view` property instead.
     ''')
     this.window,
     this.devicePixelRatio = 1.0,
@@ -1341,7 +1341,7 @@ class ViewConfiguration {
   ///
   /// If null, then this configuration represents a top level view itself.
   @Deprecated('''
-    Renaming window to view since `FlutterWindow` has been removed from dart::ui.
+      Use the `view` property instead.
   ''')
   final FlutterView? window;
   final FlutterView? view;
--- a/lib/ui/platform_dispatcher.dart
+++ b/lib/ui/platform_dispatcher.dart
@@ -1342,10 +1342,9 @@ class ViewConfiguration {
     );
   }
 
-  /// The top level view into which the view is placed and its geometry is
-  /// relative to.
+  /// The top level view for which this [ViewConfiguration]'s properties apply to.
   ///
-  /// If null, then this configuration represents a top level view itself.
+  /// If this property is null, this [ViewConfiguration] is a top level view.
   @Deprecated('''
     Use the `view` property instead.
     This change is related to adding multi-view support in Flutter.
@@ -1353,10 +1352,9 @@ class ViewConfiguration {
   ''')
   FlutterView? get window => _view;
 
-  /// The top level view into which the view is placed and its geometry is
-  /// relative to.
+  /// The top level view for which this [ViewConfiguration]'s properties apply to.
   ///
-  /// If null, then this configuration represents a top level view itself.
+  /// If this property is null, this [ViewConfiguration] is a top level view.
   FlutterView? get view => _view;
 
   final FlutterView?  _view;
--- a/lib/ui/platform_dispatcher.dart
+++ b/lib/ui/platform_dispatcher.dart
@@ -1291,6 +1291,10 @@ class ViewConfiguration {
   /// A const constructor for an immutable [ViewConfiguration].
   const ViewConfiguration({
     this.view,
+    @Deprecated('''
+      Renaming window to view as the class `FlutterWindow` has been deprecated.
+    ''')
+    this.window,
     this.devicePixelRatio = 1.0,
     this.geometry = Rect.zero,
     this.visible = false,
@@ -1305,6 +1309,7 @@ class ViewConfiguration {
   /// Copy this configuration with some fields replaced.
   ViewConfiguration copyWith({
     FlutterView? view,
+    FlutterView? window,
     double? devicePixelRatio,
     Rect? geometry,
     bool? visible,
@@ -1317,6 +1322,7 @@ class ViewConfiguration {
   }) {
     return ViewConfiguration(
       view: view ?? this.view,
+      window: window ?? this.window,
       devicePixelRatio: devicePixelRatio ?? this.devicePixelRatio,
       geometry: geometry ?? this.geometry,
       visible: visible ?? this.visible,
@@ -1333,6 +1339,10 @@ class ViewConfiguration {
   /// relative to.
   ///
   /// If null, then this configuration represents a top level view itself.
+  @Deprecated('''
+    Renaming window to view as the class `FlutterWindow` has been deprecated.
+  ''')
+  final FlutterView? window;
   final FlutterView? view;
 
   /// The pixel density of the output surface.
--- a/lib/ui/window.dart
+++ b/lib/ui/window.dart
@@ -5,8 +5,8 @@ part of dart.ui;
 
 /// A view into which a Flutter [Scene] is drawn.
 ///
-/// Each [FlutterView] has its own layer tree that is shown whenever [render]
-/// is called on it with a [Scene].
+/// Each [FlutterView] has its own layer tree that is rendered 
+/// whenever [render] is called on it with a [Scene].
 ///
 /// ## Insets and Padding
 ///
--- a/lib/ui/window.dart
+++ b/lib/ui/window.dart
@@ -55,8 +55,7 @@ class FlutterView {
   FlutterView._(this._viewId, this.platformDispatcher);
 
   /// The opaque ID for this view.
-  final Object _viewId;
-  Object get viewId => _viewId;
+  final Object viewId;
 
   /// The platform dispatcher that this view is registered with, and gets its
   /// information from.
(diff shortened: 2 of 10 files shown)


lib/ui/window.dart
```diff
@@ -6,7 +6,7 @@ part of dart.ui;
 /// A view into which a Flutter [Scene] is drawn.
 ///
 /// Each [FlutterView] has its own layer tree that is rendered into an area
-/// inside of a [FlutterWindow] whenever [render] is called with a [Scene].
+/// inside of a [SingletonFlutterWindow] whenever [render] is called with a [Scene].
```
goderbauer, line 9: nit: let's reword this paragraph without mentioning SingletonFlutterWindow or FlutterWindow. It doesn't make super sense, FlutterViews can stand alone without any window.

lib/ui/window.dart
```diff
@@ -54,15 +54,24 @@ part of dart.ui;
 ///
 /// See also:
 ///
-///  * [FlutterWindow], a special case of a [FlutterView] that is represented on
+///  * [FlutterView], a special case of a [FlutterView] that is represented on
```
goderbauer, line 57: A FlutterView is a special case of a FlutterView? :P

lib/web_ui/lib/platform_dispatcher.dart
```diff
@@ -31,7 +31,7 @@ abstract class PlatformDispatcher {
   VoidCallback? get onPlatformConfigurationChanged;
   set onPlatformConfigurationChanged(VoidCallback? callback);
 
-  Iterable<FlutterView> get views;
+  Map<Object, FlutterView> get views;
```
goderbauer, line 34: We can't expose this as a Map. Maps are mutable and we don't want the framework/apps to be able to mutate this object. It's also an (unnecessary) API break. 

We have however talked about exposing a new method, something like `getViewById` that allows you to, well, get a view by passing in an ID without exposing the entire Maps interface.

lib/ui/platform_dispatcher.dart
```diff
@@ -1333,7 +1333,7 @@ class ViewConfiguration {
   /// relative to.
   ///
   /// If null, then this configuration represents a top level view itself.
-  final FlutterView? window;
+  final FlutterView? view;
```
goderbauer, line 1336: I am all in favor of this rename. As-is, this may be hard to land because presumably it breaks the framework and hence the auto-roller. You may have to do this in two stages where view/window exist in parallel, then you switch over the framework, and then you remove the window property.

goderbauer (commented):
I know, this is still a draft, just some fly-by comments.

lib/ui/platform_dispatcher.dart
a-wallen: Is it possible to mark the property as deprecated?

goderbauer: Yeah, you should be able to use `@Deprecated(<message>)`

lib/web_ui/lib/platform_dispatcher.dart
a-wallen: https://github.com/flutter/engine/blob/d76ccb8f44c911b2bf35813bd6fb5cb791218e89/lib/web_ui/lib/src/engine/window.dart#L49

a-wallen:
```dart  
@override
  Iterable<ui.FlutterView> get views => _windows.values;
  Map<Object, ui.FlutterWindow> get windows => _windows;
  final Map<Object, ui.FlutterWindow> _windows = <Object, ui.FlutterWindow>{};
```dart

Comment (a-wallen) [reply lib/web_ui/lib/platform_dispatcher.dart:34]: Although the views aren't currently mutable, there's places in the code where we mutate entries of `windows`. 

Ideally, I'd like to keep all the same variables as above, but having a `windows` property doesn't make sense without `FlutterWindow` and renaming it to `views` creates a conflict.

File: lib/ui/window.dart (hunk 2 above)
Comment (a-wallen) [reply lib/ui/window.dart:57]: 😅

File: lib/ui/platform_dispatcher.dart [hunk 5]
```diff
@@ -1290,6 +1291,10 @@ class PlatformConfiguration {
 class ViewConfiguration {
   /// A const constructor for an immutable [ViewConfiguration].
   const ViewConfiguration({
+    this.view,
+    @Deprecated('''
+      Renaming window to view since `FlutterWindow` has been removed from dart::ui.
```
dkwingsmt, line 1296: 1. I don't understand this deprecation message. It should be something actionable in the form of "Use xxxx instead." For example, "Use the `view` argument instead. The `window` argument has been renamed to `view`."
2. If it's a rename, shouldn't we point them to the same value? The two parameter should be exclusive (assert that only one of them is non-null) and be assigned to the same field (`_view`) and both getter should return this field.

a-wallen: Yeah, you're right 👍

a-wallen: How about 
```
      `FlutterWindow` has been removed from dart::ui, and is illogical semantically
       use the `view` property instead.
```

lib/ui/window.dart
```diff
@@ -328,7 +300,7 @@ class FlutterWindow extends FlutterView {
 /// `WidgetsBinding.instance.platformDispatcher` over a static reference to
 /// [window], or [PlatformDispatcher.instance]. See the documentation for
 /// [PlatformDispatcher.instance] for more details about this recommendation.
-class SingletonFlutterWindow extends FlutterWindow {
+class SingletonFlutterWindow extends FlutterView {
```
dkwingsmt, line 303: I'm wondering if inheritance is a good idea here. I don't think a window _is_ a view. I propose that we keep `FlutterWindow` and let it own a `FlutterView view` property. This is probably a less breaking change as well, so that we can discuss how to refine it in phase 2.

lib/ui/platform_dispatcher.dart
dkwingsmt: In my opinion the explanation is unnecessary here. Just tell the user how to change it. If you really want to explain it, add a link to the design doc, or to this PR with an expanded paragraph of description.

loic-sharma: > For example, "Use the `view` argument instead. The `window` argument has been renamed to `view`."

+1 for this deprecation message

loic-sharma: Reopening as I don't think the second part was addressed:

> If it's a rename, shouldn't we point them to the same value? The two parameter should be exclusive (assert that only one of them is non-null) and be assigned to the same field (_view) and both getter should return this field.

lib/ui/window.dart
```diff
@@ -5,8 +5,8 @@ part of dart.ui;
 
 /// A view into which a Flutter [Scene] is drawn.
 ///
-/// Each [FlutterView] has its own layer tree that is rendered into an area
-/// inside of a [FlutterWindow] whenever [render] is called with a [Scene].
+/// Each [FlutterView] has its own layer tree that is shown whenever [render]
+/// is called on it with a [Scene].
```
loic-sharma, line 9: Sorry for being pedantic, but there's no guarantee a rendered scene is shown. What do you think of something like this?

```suggestion
/// Each [FlutterView] has its own layer tree that is rendered 
/// whenever [render] is called on it with a [Scene].
```

lib/ui/platform_dispatcher.dart
```diff
@@ -1290,7 +1291,11 @@ class PlatformConfiguration {
 class ViewConfiguration {
   /// A const constructor for an immutable [ViewConfiguration].
   const ViewConfiguration({
-    this.window,
+    FlutterView? view,
+    @Deprecated('''
```
goderbauer, line 1295: This is missing the version in which this was deprecated, see https://github.com/flutter/flutter/wiki/Tree-hygiene#deprecations.

lib/ui/platform_dispatcher.dart
```diff
@@ -1300,10 +1305,12 @@ class ViewConfiguration {
     this.padding = WindowPadding.zero,
     this.gestureSettings = const GestureSettings(),
     this.displayFeatures = const <DisplayFeature>[],
-  });
+  }) : assert(!(window != null && view != null)),
```
goderbauer, line 1308: could be simplified to `assert(window == null || view == null)`, no?

lib/ui/platform_dispatcher.dart
```diff
@@ -1333,7 +1340,13 @@ class ViewConfiguration {
   /// relative to.
   ///
   /// If null, then this configuration represents a top level view itself.
-  final FlutterView? window;
+  @Deprecated('''
+      Use the `view` property instead.
+  ''')
+  FlutterView? get window => _view;
+  FlutterView? get view => _view;
```
goderbauer, line 1347: `view` needs a doc comment.

lib/ui/platform_dispatcher.dart
```diff
@@ -1333,7 +1340,13 @@ class ViewConfiguration {
   /// relative to.
   ///
   /// If null, then this configuration represents a top level view itself.
-  final FlutterView? window;
+  @Deprecated('''
+      Use the `view` property instead.
```
goderbauer, line 1344: missing version (see previous comment)

lib/ui/platform_dispatcher.dart
```diff
@@ -1333,7 +1340,13 @@ class ViewConfiguration {
   /// relative to.
```
goderbauer, line 1340: Not directly related to your PR, but this doc comment doesn't make much sense: "[The] view into which the view is placed"?

goderbauer: Similarly, the comment below: What is the "top level view itself"?

lib/ui/window.dart
```diff
@@ -51,18 +51,22 @@ part of dart.ui;
...
-/// See also:
-///
-///  * [FlutterWindow], a special case of a [FlutterView] that is represented on
-///    the platform as a separate window which can host other [FlutterView]s.
-abstract class FlutterView {
+class FlutterView {
+  FlutterView._(this._viewId, this.platformDispatcher);
+
+  /// The opaque ID for this view.
+  final Object _viewId;
```
goderbauer, line 58: move the private field below the getter. Our style is to have the following order: getter, private field, setter. The setter may be optional, though.

lib/ui/platform_dispatcher.dart
dkwingsmt, line 1347: Add doc to `view`.

lib/ui/platform_dispatcher.dart
```diff
@@ -1290,7 +1291,11 @@ class PlatformConfiguration {
 class ViewConfiguration {
   /// A const constructor for an immutable [ViewConfiguration].
```
dkwingsmt, line 1292: Explain in the doc that only one of `view` and `window` should be non-null, and that they're equivalent.

lib/ui/window.dart
goderbauer: Unless this is actively breaking anything I think this is fine. SingletonFlutterWindow was a hack to avoid a massive breaking change. It will get deprecated in the next few weeks as we continue to remove the single window assumption from the framework.

lib/web_ui/lib/platform_dispatcher.dart
```diff
@@ -190,6 +190,10 @@ class PlatformConfiguration {
 
 class ViewConfiguration {
   const ViewConfiguration({
+    this.view,
+    @Deprecated('''
+      Renaming window to view as the class `FlutterWindow` has been deprecated.
```
goderbauer, line 195: nit: keep these consistent with the other file?

lib/web_ui/lib/platform_dispatcher.dart
```diff
@@ -215,6 +220,7 @@ class ViewConfiguration {
     List<DisplayFeature>? displayFeatures,
   }) {
     return ViewConfiguration(
```
goderbauer, line 222: assert here that at most of view, window is not-null.

lib/ui/window.dart
dkwingsmt, line 58: Is this intentionally made an `Object` instead of a number?

lib/web_ui/lib/platform_dispatcher.dart
```diff
@@ -190,6 +190,10 @@ class PlatformConfiguration {
 
 class ViewConfiguration {
   const ViewConfiguration({
+    this.view,
```
dkwingsmt, line 193: The web side should be treated the same as `ui`.

lib/ui/platform_dispatcher.dart
```diff
@@ -1300,10 +1305,12 @@ class ViewConfiguration {
...
     this.gestureSettings = const GestureSettings(),
     this.displayFeatures = const <DisplayFeature>[],
-  });
+  }) : assert(!(window != null && view != null)),
+    _view = view ?? window;
 
   /// Copy this configuration with some fields replaced.
   ViewConfiguration copyWith({
+    FlutterView? view,
     FlutterView? window,
```
goderbauer, line 1314: copy with should do something with the window property instead of silently ignoring it.

lib/ui/platform_dispatcher.dart
```diff
@@ -1316,7 +1323,7 @@ class ViewConfiguration {
     List<DisplayFeature>? displayFeatures,
   }) {
     return ViewConfiguration(
```
goderbauer, line 1325: should probably assert that at least one of window, view is null

lib/web_ui/lib/platform_dispatcher.dart
```diff
@@ -228,7 +234,11 @@ class ViewConfiguration {
     );
   }
 
-  final FlutterWindow? window;
+  @Deprecated('''
+    Renaming window to view as the class `FlutterWindow` has been deprecated.
```
goderbauer, line 238: keep these consistent.

lib/ui/platform_dispatcher.dart
goderbauer, line 1295: Ultimately, it's up to you, but I would probably do the rename (with deprecation) in a separate PR that's dedicated to just that. That may make it easier to land these two changes and potentially roll back if there are issues.

lib/web_ui/lib/src/engine/platform_dispatcher.dart
```diff
@@ -138,9 +138,10 @@ class EnginePlatformDispatcher extends ui.PlatformDispatcher {
 
   /// The current list of windows,
   @override
-  Iterable<ui.FlutterView> get views => _windows.values;
-  Map<Object, ui.FlutterWindow> get windows => _windows;
-  final Map<Object, ui.FlutterWindow> _windows = <Object, ui.FlutterWindow>{};
+  Iterable<ui.FlutterView> get views => _views.values;
+  final Map<Object, ui.FlutterView> _views = <Object, ui.FlutterView>{};
+  ui.FlutterView? getViewById(Object id) => _views[id];
+  void addView(Object id, ui.FlutterView view) => _views[id] = view;
```
goderbauer, line 144: Since this is all private, you could just replace all of this with something like:

```suggestion
  final Map<Object, ui.FlutterView> idToView = <Object, ui.FlutterView>{};
```

... and then elsewhere operate directly with that map.

lib/web_ui/lib/src/engine/window.dart
```diff
@@ -340,11 +343,14 @@ class EngineSingletonFlutterWindow extends EngineFlutterWindow {
 }
 
 /// A type of [FlutterView] that can be hosted inside of a [FlutterWindow].
```
goderbauer, line 345: update the comment since FlutterWindow doesn't exist anymore.

The name of the class "EngineFlutterWindowView" also seems odd with the "Window" in there. Not sure what this is used for, but is this still needed?

lib/ui/window.dart
a-wallen: The `viewId` on `FlutterWindow` was an `Object`. I think that using types like string or a number makes sense as a `viewId` so I just left it.

lib/ui/platform_dispatcher.dart
a-wallen: Added in 56e558cca6d26765dac6ac5bf815074f04d30bca, it's just a duplicate of the doc for `window` since they are both the same.

lib/ui/platform_dispatcher.dart
a-wallen: You can resolve this if you're satisfied with the doc added in 278f7249916621f6611e8da715e76511eae39178.

lib/ui/platform_dispatcher.dart
a-wallen: vacation...

a-wallen: 23be98b9c5f2eee6ceb810a7f6f9a67ea3383555 should fix it. Thanks!

lib/ui/platform_dispatcher.dart
a-wallen: I think that this is a duplicate of https://github.com/flutter/engine/pull/38453#discussion_r1060960086

lib/ui/platform_dispatcher.dart
a-wallen: Got it, thanks for sharing. I need to bookmark that page.

This should resolve this comment 40922b5c8b5b029fe619801b8435d1aa37b8d926.

lib/ui/platform_dispatcher.dart
a-wallen: 40922b5c8b5b029fe619801b8435d1aa37b8d926

lib/ui/window.dart
a-wallen: Sure thing. 

Is that documented somewhere?

lib/ui/platform_dispatcher.dart
a-wallen: How about this 75d1cbe7771e687410113c9d64626f2c30b7523a?

lib/ui/platform_dispatcher.dart
a-wallen: Added 7d2ad3ebb21e9b735bc20810991044fe34b72159.

lib/ui/platform_dispatcher.dart
a-wallen: Sounds like a good idea. I'd like to address all the feedback and get an LGTM before I cherry-pick the commits from this PR into another one.

loic-sharma (approved):
LGTM but please get sign off from other folks that are more familiar with this :)

lib/ui/window.dart
goderbauer: I actually couldn't find this in the style guide, but it is something we try to follow. We should add it.

goderbauer:
Looks like the checks are unhappy now.

lib/ui/platform_dispatcher.dart
```diff
@@ -1315,8 +1328,9 @@ class ViewConfiguration {
     GestureSettings? gestureSettings,
     List<DisplayFeature>? displayFeatures,
   }) {
+    assert(view == null || window == null);
     return ViewConfiguration(
-      window: window ?? this.window,
+      view: view ?? this.view,
```
goderbauer, line 1332: you need to deal with the window here in addition to the view for backwards compatibility reasons:

```dart
window: window ?? this.window,
```

lib/ui/platform_dispatcher.dart
```diff
@@ -1289,8 +1289,18 @@ class PlatformConfiguration {
 /// An immutable view configuration.
 class ViewConfiguration {
   /// A const constructor for an immutable [ViewConfiguration].
+  ///
+  /// When constructing a view configuration, supply either the `view` or the
+  /// `window` property, but not both since the `view` and `window` property
+  /// are backed by the same instance variable.
```
goderbauer, line 1295: nit: since window and view are actual properties of this class put them in `[]` instead of `. That way, dartdoc directly links to them.

lib/web_ui/lib/platform_dispatcher.dart
```diff
@@ -215,7 +223,7 @@ class ViewConfiguration {
     List<DisplayFeature>? displayFeatures,
   }) {
     return ViewConfiguration(
-      window: window ?? this.window,
+      view: view ?? this.view,
```
goderbauer, line 226: assert that only one is provided and add

```dart
window: window ?? this.window,
```

lib/ui/platform_dispatcher.dart
a-wallen: I'll get an error that the `assert(window != null || view != null)` cannot be optimized since both the `window` and `view` properties have the same backing property.

a-wallen:
```dart
assert(window != null || view != null);
return ViewConfiguration(
	view: view ?? window ?? this.view,
);
```

dkwingsmt: Yeah this is also what I would go for.

a-wallen: Thanks @dkwingsmt 👍

a-wallen: 8199e721121231d2e4871a0576628d729209e843 has the changes requested.

lib/web_ui/lib/platform_dispatcher.dart
a-wallen: Resolved. See https://github.com/flutter/engine/pull/38453#discussion_r1061041530.

lib/ui/platform_dispatcher.dart
```diff
@@ -1300,10 +1310,12 @@ class ViewConfiguration {
     this.padding = WindowPadding.zero,
     this.gestureSettings = const GestureSettings(),
     this.displayFeatures = const <DisplayFeature>[],
-  });
+  }) : assert(window == null || view == null),
+    _view = view ?? window;
 
   /// Copy this configuration with some fields replaced.
   ViewConfiguration copyWith({
+    FlutterView? view,
```
goderbauer, line 1318: Deprecate the `window` parameter in the copyWith functions as well? That one will also need to be removed at some point, right?

lib/ui/window.dart
```diff
@@ -51,18 +51,22 @@ part of dart.ui;
...
+  Object get viewId => _viewId;
```
goderbauer, line 59: The getter that just returns a final private variable can be simplified to:

```suggestion
  /// The opaque ID for this view.
  final Object viewId;
```

goderbauer: Alternatively, move the field below the getter and the doc comment should be on the getter.

lib/web_ui/lib/platform_dispatcher.dart
```diff
@@ -200,10 +206,12 @@ class ViewConfiguration {
...
     this.gestureSettings = const GestureSettings(),
     this.displayFeatures = const <DisplayFeature>[],
-  });
+  }) : assert(window == null || view == null),
+  _view = view ?? window;
 
   ViewConfiguration copyWith({
-    FlutterWindow? window,
+    FlutterView? view,
+    FlutterView? window,
```
goderbauer, line 214: Deprecate window param?

goderbauer (approved):
LGTM after remaining comments are addressed.

dkwingsmt:
Yeah we probably also want some tests to make sure that the `view` parameter works as well as `window`.

lib/web_ui/lib/platform_dispatcher.dart
a-wallen: https://github.com/flutter/engine/pull/38453/commits/5c2e0a9d2a7250ab23e45dccd573c684fbf0791c

lib/ui/platform_dispatcher.dart
a-wallen: https://github.com/flutter/engine/pull/38453/commits/5c2e0a9d2a7250ab23e45dccd573c684fbf0791c

lib/web_ui/test/engine/platform_dispatcher_test.dart
```diff
@@ -160,6 +162,34 @@ void testMain() {
...
+    }, throwsAssertionError);
+  });
+
+  test("A ViewConfiguration's view and window are backed with the same property", () {
+    final ui.FlutterView window = EngineFlutterWindow(0, ui.PlatformDispatcher.instance);
+    final ui.ViewConfiguration viewConfiguration = ui.ViewConfiguration(view: window);
+    // ignore: deprecated_member_use
+    expect(viewConfiguration.window, window);
+    // ignore: deprecated_member_use
+    expect(viewConfiguration.window, viewConfiguration.view);
```
dkwingsmt, line 183: Can you also make sure that the `window` argument works? Just copy the test but initialize with `window`.

lib/web_ui/test/engine/platform_dispatcher_test.dart
```diff
@@ -160,6 +162,34 @@ void testMain() {
...
+  });
+
+  test('Calling copyWith on a ViewConfiguration with both a window and view throws an error', () {
+    final ui.FlutterView window = EngineFlutterWindow(0, ui.PlatformDispatcher.instance);
```
dkwingsmt, line 187: Can you add positive test cases for `copyWith`? Test that `copyWith` works with `window` and with `view`.

dkwingsmt (approved):
LGTM except for two suggestions for tests.

lib/web_ui/test/engine/platform_dispatcher_test.dart
a-wallen: https://github.com/flutter/engine/pull/38453/commits/6d531ba35e27e11034e5ed3ee44d283dae81b3e0 should resolve this.

lib/web_ui/test/engine/platform_dispatcher_test.dart
a-wallen: I don't think that there's a way to create another `FlutterView` to copy over the `ViewConfiguration`'s owned view since Flutter is only single-view for the moment. 

Let me know if I'm missing something, but I'm not sure if I can resolve this suggestion.

lib/web_ui/lib/src/engine/window.dart
```diff
@@ -58,7 +58,10 @@ class EngineFlutterWindow extends ui.SingletonFlutterWindow {
     });
   }
 
-  final Object _windowId;
+  final Object _viewId;
+
+  @override
+  Object get viewId => _viewId;
```
ditman, line 64: Isn't this the same, but a little bit shorter?

```suggestion
  @override
  final Object viewId;
```

lib/web_ui/lib/src/engine/window.dart
```diff
@@ -339,25 +342,6 @@ class EngineSingletonFlutterWindow extends EngineFlutterWindow {
   double? _debugDevicePixelRatio;
 }
 
-/// A type of [FlutterView] that can be hosted inside of a [FlutterWindow].
-class EngineFlutterWindowView extends ui.FlutterWindow {
```
ditman, line 343: I like this simplification, however we left this here because it seemed that this was going to be used by the new multi-view hierarchy of classes. What happened? Is there a design doc for this?

lib/web_ui/lib/src/engine/platform_dispatcher.dart
```diff
@@ -138,9 +138,8 @@ class EnginePlatformDispatcher extends ui.PlatformDispatcher {
 
   /// The current list of windows,
   @override
-  Iterable<ui.FlutterView> get views => _windows.values;
-  Map<Object, ui.FlutterWindow> get windows => _windows;
-  final Map<Object, ui.FlutterWindow> _windows = <Object, ui.FlutterWindow>{};
+  Iterable<ui.FlutterView> get views => viewData.values;
+  final Map<Object, ui.FlutterView> viewData = <Object, ui.FlutterView>{};
```
ditman, line 142: I think the iteration order of a plain HashMap in Dart is not guaranteed, so `viewData` *must* be a `LinkedHashMap` if we want indices to be consistent through the `views` Iterable, right?

https://api.dart.dev/be/175791/dart-core/Map-class.html

ditman: (Or is `views` a stopgap measure until we retrieve directly from the viewData by their ViewId?)

Note that this also exists in the mobile lib/ui/platform_dispatcher.dart (~line 165)

lib/web_ui/lib/src/engine/platform_dispatcher.dart
```diff
@@ -478,10 +477,10 @@ class EnginePlatformDispatcher extends ui.PlatformDispatcher {
         final MethodCall decoded = codec.decodeMethodCall(data);
         switch (decoded.method) {
           case 'SystemNavigator.pop':
-            // TODO(gspencergoog): As multi-window support expands, the pop call
+            // TODO(a-wallen): As multi-window support expands, the pop call
             // will need to include the window ID. Right now only one window is
```
ditman, line 481: Should "window ID" be "view ID"?

lib/web_ui/test/engine/platform_dispatcher_test.dart
```diff
@@ -160,6 +162,40 @@ void testMain() {
...
+    // ignore: deprecated_member_use
+    expect(viewConfiguration.window, window);
+    // ignore: deprecated_member_use
+    expect(viewConfiguration.window, viewConfiguration.view);
+  });
+
+  test('Initialize a ViewConfiguration with a window', () {
+    final ui.FlutterView window = EngineFlutterWindow(0, ui.PlatformDispatcher.instance);
+    // ignore: deprecated_member_use
+    ui.ViewConfiguration(window: window);
```
ditman, line 189: This needs an expectation too:

```suggestion
    ui.ViewConfiguration(window: window);
    expect(viewConfiguration.view, window);
```

lib/web_ui/test/engine/platform_dispatcher_test.dart
```diff
@@ -160,6 +162,40 @@ void testMain() {
...
+        view: window,
+      );
+    }, throwsAssertionError);
+  });
+
+  test("A ViewConfiguration's view and window are backed with the same property", () {
+    final ui.FlutterView window = EngineFlutterWindow(0, ui.PlatformDispatcher.instance);
+    final ui.ViewConfiguration viewConfiguration = ui.ViewConfiguration(view: window);
+    // ignore: deprecated_member_use
+    expect(viewConfiguration.window, window);
```
ditman, line 181:
```suggestion
    expect(viewConfiguration.view, window);
```

The next line already tests that .window (the deprecated value) matches .view, no need to access the deprecated value twice IMO.

ditman (commented):
LGTM, left some small suggestions in `web_ui`.

Two questions:

1. Small concern about the iteration order / indexed access of the `views` property of the platform_dispatcher (both on Web and VM), once we have more than one view :)
2. This PR seems to be removing classes that were added in previous multi-view work. Do we have a design doc/class diagram of what's the desired end state of this series of refactors?

lib/web_ui/lib/src/engine/platform_dispatcher.dart
jonahwilliams: map literals are LinkedHashmaps and the iteration order is gauranteed.

lib/web_ui/lib/src/engine/platform_dispatcher.dart
```diff
@@ -478,10 +477,10 @@ class EnginePlatformDispatcher extends ui.PlatformDispatcher {
...
             // supported.
-            (_windows[0]! as EngineFlutterWindow)
+            (viewData[0]! as EngineFlutterWindow)
```
mdebbar, line 483: `0` seems to have a special meaning here that's not documented.

The code looks like it's trying to access the "first" element in a list, but that's not what's happening. `viewData` (and previously `_windows`) is a map, and the key `0` is basically a special hard-coded key that's shared between here and:
https://github.com/flutter/engine/blob/bb4015269d3d252160133b377b72e315e89f2d4d/lib/web_ui/lib/src/engine/window.dart#L366-L367

I suggest we make that explicit through a constant (e.g. `const int kSingletonViewId = 0`), and stop using literal `0` elsewhere.

(I know this existed before your PR, but I wanted to recommend it since you are working on this part of the code anyway).

lib/web_ui/lib/src/engine/platform_dispatcher.dart
timsneath: (This doesn't seem to be [documented accurately in the Dart language tour](https://dart.dev/guides/language/language-tour#maps:~:text=Dart%20infers%20that%20gifts%20has%20the%20type%20Map%3CString%2C%20String%3E%20and%20nobleGases%20has%20the%20type%20Map%3Cint%2C%20String%3E). Adding @mit-mit to take a look and file an issue if necessary.)

lib/web_ui/lib/src/engine/platform_dispatcher.dart
```diff
@@ -138,9 +138,8 @@ class EnginePlatformDispatcher extends ui.PlatformDispatcher {
 
   /// The current list of windows,
```
loic-sharma, line 139: The comma here looks unintentional, consider fixing it while you're here :)

```suggestion
  /// The current list of windows.
```

lib/web_ui/lib/src/engine/window.dart
a-wallen: It isn't being used right now. If it's needed it'll be easy to re-implement.

https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#lazy-programming

lib/web_ui/lib/src/engine/platform_dispatcher.dart
a-wallen: https://github.com/flutter/engine/pull/38453/commits/8fa153bb72ad174e862c2f92ef886a74788213e3

lib/web_ui/lib/src/engine/platform_dispatcher.dart
ditman: It's documented in the Map *constructor* [here](https://api.flutter.dev/flutter/dart-core/Map/Map.html), but I couldn't find that piece of lore looking at the main pages of Map/LinkedHashMap, apologies for the noise in the PR!!

flar:
This PR caused the engine roll to break: https://github.com/flutter/flutter/pull/118054

ditman:
Wow, the errors are unfortunate :/

```
dartdoc:stdout: Generating docs for library foundation from package:flutter/foundation.dart...
dartdoc:stderr:   error: unresolved doc reference [ui.FlutterWindow]
dartdoc:stderr:     from foundation.BindingBase.window: (file:///b/s/w/ir/x/w/%C3%81%20flutter%20sdk/packages/flutter/lib/src/foundation/binding.dart:202:33)

dartdoc:stdout: Generating docs for library flutter_test from package:flutter_test/flutter_test.dart...
dartdoc:stderr:   error: unresolved doc reference [ui.FlutterWindow]
dartdoc:stderr:     from flutter_test.TestWidgetsFlutterBinding.window: (file:///b/s/w/ir/x/w/%C3%81%20flutter%20sdk/packages/flutter_test/lib/src/binding.dart:191:18)
dartdoc:stderr:     in documentation inherited from binding.BindingBase.window: (file:///b/s/w/ir/x/w/%C3%81%20flutter%20sdk/packages/flutter/lib/src/foundation/binding.dart:202:33)
```

lib/web_ui/lib/src/engine/platform_dispatcher.dart
mit-mit: @lrhn promised to take a look at the API docs

ditman:
🎉 just saw the notification that this landed. Kudos!

lib/web_ui/lib/platform_dispatcher.dart
```diff
@@ -214,8 +227,9 @@ class ViewConfiguration {
     GestureSettings? gestureSettings,
     List<DisplayFeature>? displayFeatures,
   }) {
+    assert(view == null || window == null);
```
lrhn, line 230: Personally, I'd use runtime check to make absolutely sure that no production code ends up with both being non-`null`. Asserts can be ignored. I assume the check will be removed against when `window` is removed.

PR #38455, opened 2022-12-21
Remove single window/view assumption from `SceneBuilder`

a-wallen:
fixes https://github.com/flutter/flutter/issues/112202

## Pre-launch Checklist

- [ ] I signed the [CLA].

### DIFF
--- a/lib/ui/compositing.dart
+++ b/lib/ui/compositing.dart
@@ -878,12 +878,16 @@ class SceneBuilder extends NativeFieldWrapperClass1 {
   ///
   /// After calling this function, the scene builder object is invalid and
   /// cannot be used further.
-  Scene build() {
+  Scene build({
+    required Rect bounds,
+    double displayPixelRatio = 1.0,
+    double touchSlop = -1.0,
+  }) {
     final Scene scene = Scene._();
-    _build(scene);
+    _build(scene, bounds.width, bounds.height, displayPixelRatio, touchSlop);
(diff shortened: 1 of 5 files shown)


jonahwilliams (commented):
Even though the Scene needs a physical size and dpr, it can actually be given these values directly instead of taking them from the window. See https://github.com/flutter/flutter/blob/master/packages/flutter/lib/src/rendering/layer.dart#L1428-L1505

It would be odd the make these cases require supplying a dummy window. 

Instead I would consider deferring the lookup of physical size/dpr until the scene is rendered instead of doing it eagerly and then throwing this information away in some cases

a-wallen:
> Instead I would consider deferring the lookup of physical size/dpr until the scene is rendered instead of doing it eagerly and then throwing this information away in some cases

@jonahwilliams if the scene is given different values than those in the view it's being rendered into should those values be overridden?

jonahwilliams:
conceptually Scenes don't have these metrics, they take them from either the arguments to toImage or from the window they are rendered into. Its sort of an implementation detail that constructing a scene eagerly looks up the viewport metrics but this could be done later

jonahwilliams:
None of the methods that reference the viewport metrics are called until the scene is moved to the raster thread.

Hixie:
I'm going to close this because @dkwingsmt has an alternative approach he's chasing down.

PR #38497, opened 2022-12-25
[Impeller] Set adaptive tolerance when rendering FillPathGeometry

luckysmg:
https://github.com/flutter/flutter/issues/99769

## Pre-launch Checklist

- [x] I signed the [CLA].


jonahwilliams (approved):
LGTM

impeller/entity/geometry.cc
```diff
@@ -53,10 +53,13 @@ GeometryResult FillPathGeometry::GetPositionBuffer(
     const ContentContext& renderer,
     const Entity& entity,
     RenderPass& pass) {
+  auto tolerance =
+      kDefaultCurveTolerance / entity.GetTransformation().GetMaxBasisLength();
```
bdero, line 57: Does the max basis length need to be bottom-bound to e.g. `0.1f` for this? What happens when the transform is scaled to 0 and the tolerance becomes `NaN`?

bdero:
FYI @zanderso, this is expected to increase tessellation time for unscaled fill curves on high DPI displays. Flutter scales the transform for everything up by 3x on my iPhone 12 Mini.

zanderso:
How much does it impact tessellation time? Tessellation is already pretty high up in CPU profiles. The circle does look a lot better with this change, but it would be good to know how much bigger the tessellation problem is getting.

impeller/entity/geometry.cc
luckysmg: Hello @bdero 
It becomes below when I set `auto tolerance = kDefaultCurveTolerance / 0;`
image

And this change is almost the same like code in 
https://github.com/flutter/engine/blob/main/impeller/entity/geometry.cc#L420-#L422

luckysmg:
Thx @zanderso to pointing out the high CPU Profile usage about the tessellation.
And I had some try to modify the `tolerance`, since **we need to keep `tolerance` as big as possible while also keeping it visually perfect, since bigger `tolerance` can reduce the CPU usage.** So I think we could find a balance between that.O(∩_∩)O

Here is the formula (`kDefaultCurveTolerance = .1f`).
`auto tolerance =
      kDefaultCurveTolerance * x / entity.GetTransformation().GetMaxBasisLength();`


<details>

<summary>See the result when "x" is change </summary>

**x = 1**


**x = 2**


**x = 4**


**x = 6**



**x = 10**



</details>

**Here is the image zip, you can open them all and switch between them to see the difference on your computer.**
[images.zip](https://github.com/flutter/engine/files/10341603/imageZip.zip)

With result above I think maybe we can use `x = 2` or `x = 4` to decrease the points. Since I can't see very trasparent difference between `x=1` and `x=2` and `x=4` 

LMK If I missed something O(∩_∩)O

jonahwilliams:
I think we should probably eat the cost of the triangulation in order to get reasonable fidelity. If anything, it will make it easier to prioritize compute shaders/buffer management/caching improvements

jonahwilliams:
Whereas we don't have any other route for improving circle fidelity in the general case (except for drawCircle, which we could use prototypes)

impeller/entity/geometry.cc
bdero: Thanks for checking! This seems like good behavior.

bdero (approved):
I think it makes sense to land this, especially since we can easily tune this down later if we need to win back some headroom in short order. :)

For future onlookers: Note that the "circle" in this test isn't produced by the path builder. `PathBuilder` produces more accurate ovals with colinear control lines for overlapping end points, and so there shouldn't be any discontinuities when smoothed, unlike in the `BezierCircleScaled` test.

PR #38538, opened 2022-12-28
Implement ITextProvider and ITextRangeProvider for UIA

yaakovschectman:
This is the initial subset of #38284, handling importing files from Chromium, making modifications necessary to compile, and temporarily disabling the newly added unittests.

In order for text providing nodes to properly interact with screen reader commands via UIA, we need to implement text and textedit pattern providers. As an example, this allows screen readers to narrate the character after the carat when the user moves the carat in an edit field with the arrow keys.

Unit tests for `AXPlatformNodeText(Range)ProviderWin` are modeled after those found in the Chromium source. Details for some required modification to account for differences between the two codebases, and a handful of the newly introduced test cases are currently disabled, which is tracked in [flutter/flutter#117012](https://github.com/flutter/flutter/issues/117012).

Functionality for searching for text in a text range does not yet respect the `ignore_case` option: [flutter/flutter#117013](https://github.com/flutter/flutter/issues/117013).

Part of [flutter/flutter#116219](https://github.com/flutter/flutter/issues/116219)

## Pre-launch Checklist

- [ ] I signed the [CLA].


third_party/accessibility/ax/ax_node.h
```diff
@@ -436,6 +440,8 @@ class AX_EXPORT AXNode final {
   // Finds and returns a pointer to ordered set containing node.
   AXNode* GetOrderedSet() const;
 
+  AXNode* GetLowestPlatformAncestor() const;
```
loic-sharma, line 443: Can we keep Chromium's comment?

```suggestion
  // If this node is exposed to the platform's accessibility layer, returns this
  // node. Otherwise, returns the lowest ancestor that is exposed to the
  // platform. (See `IsLeaf` and `IsIgnored` for information on what is
  // exposed to platform APIs.)
  AXNode* GetLowestPlatformAncestor() const;
```

third_party/accessibility/ax/ax_node.h
```diff
@@ -120,6 +120,7 @@ class AX_EXPORT AXNode final {
   size_t GetUnignoredChildCount() const;
   AXNode* GetUnignoredChildAtIndex(size_t index) const;
   AXNode* GetUnignoredParent() const;
+  OwnerTree::Selection GetUnignoredSelection() const;
```
loic-sharma, line 123: Can we keep Chromium's comment?

```suggestion
// Gets the unignored selection from the accessibility tree, meaning the
  // selection whose endpoints are on unignored nodes. (An "ignored" node is a
  // node that is not exposed to platform APIs: See `IsIgnored`.)
  OwnerTree::Selection GetUnignoredSelection() const;
```

third_party/accessibility/ax/platform/ax_platform_node_textrangeprovider_win_unittest.cc
```diff
@@ -0,0 +1,7565 @@
...
+      &count));
+  ASSERT_EQ(25, count);
+  EXPECT_UIA_TEXTRANGE_EQ(text_range_provider, L"");
+
+  ASSERT_HRESULT_SUCCEEDED(
+      text_range_provider->ExpandToEnclosingUnit(TextUnit_Line));
+  EXPECT_UIA_TEXTRANGE_EQ(text_range_provider, L"not line #1");
+}
+
+// TOOD(schectman) https://github.com/flutter/flutter/issues/117012
```
loic-sharma, line 1258:
```suggestion
// TODO(schectman) https://github.com/flutter/flutter/issues/117012
```

loic-sharma: Same comment for other TODOs :)

third_party/accessibility/ax/platform/ax_platform_node_textprovider_win.cc
```diff
@@ -0,0 +1,354 @@
...
+  BASE_DCHECK(start->GetAnchor() == end->GetAnchor());
+
+  // SAFEARRAYs are not dynamic, so fill the visible ranges in a vector
+  // and then transfer to an appropriately-sized SAFEARRAY
+  std::vector<Microsoft::WRL::ComPtr<ITextRangeProvider>> ranges;
+
+  auto current_line_start = start->Clone();
+  while (!current_line_start->IsNullPosition() && *current_line_start < *end) {
+    auto current_line_end = current_line_start->CreateNextLineEndPosition(
+        AXBoundaryBehavior::CrossBoundary);
```
loic-sharma, line 143: Follow-up to this conversation: https://github.com/flutter/engine/pull/38284#discussion_r1057955169

> @loic-sharma: Could you explain why AXBoundaryDetection::kDontCheckInitialPosition was removed here?

> @yaakovschectman:
The method signature for our version of CreateNextLineEndPosition takes a parameter of a different type than Chrome's current version of the same. We take a AXBoundaryBehavior, while Chromium takes a struct consisting of AXBoundaryBehavior, AXBoundaryDetection

Why are our method signatures different? Should we update our signatures to match Chromium's?

yaakovschectman: This is not only a superficial signature difference, our implementation is different also. We do not have the `AXMovementOptions` struct that exists in Chromium. Updating it in our existing code may warrant a separate PR.

third_party/accessibility/ax/platform/ax_platform_node_textrangeprovider_win.h
```diff
@@ -0,0 +1,283 @@
...
+  END_COM_MAP()
+
+  AXPlatformNodeTextRangeProviderWin();
+  ~AXPlatformNodeTextRangeProviderWin();
+
+  static ITextRangeProvider* CreateTextRangeProvider(
+      AXNodePosition::AXPositionInstance start,
+      AXNodePosition::AXPositionInstance end);
+
+  static ITextRangeProvider* CreateTextRangeProviderForTesting(
```
loic-sharma, line 37: Can we keep Chromium's comment here?

```suggestion
  // Creates an instance of the class for unit tests, where AXPlatformNodes	
  // cannot be queried automatically from endpoints.
  static ITextRangeProvider* CreateTextRangeProviderForTesting(
```

third_party/accessibility/ax/platform/ax_platform_node_textrangeprovider_win.cc
```diff
@@ -0,0 +1,1701 @@
...
+  explicit AXRangePhysicalPixelRectDelegate(
+      AXPlatformNodeTextRangeProviderWin* host)
+      : host_(host) {}
+
+  gfx::Rect GetInnerTextRangeBoundsRect(
+      AXTreeID tree_id,
+      AXNode::AXID node_id,
+      int start_offset,
+      int end_offset,
+      AXOffscreenResult* offscreen_result) override {
```
loic-sharma, line 68: Should we add the `clipping_behavior` parameter to `AXRangeRectDelegate`?

third_party/accessibility/ax/platform/ax_platform_node_textrangeprovider_win.cc
```diff
@@ -0,0 +1,1701 @@
...
+  return S_OK;
+}
+
+static bool StringSearch(const std::u16string& search_string,
+                         const std::u16string& find_in,
+                         size_t* find_start,
+                         size_t* find_length,
+                         bool ignore_case,
+                         bool backwards) {
+  // TODO(schectman) Respect ignore_case
```
loic-sharma, line 440: Do we also need to add internationalization support?

third_party/accessibility/ax/platform/ax_platform_node_textrangeprovider_win.cc
```diff
@@ -0,0 +1,1701 @@
...
+  bool is_start_endpoint = endpoint == TextPatternRangeEndpoint_Start;
+  AXPositionInstance position_to_move =
+      is_start_endpoint ? start()->Clone() : end()->Clone();
+
+  AXPositionInstance new_position;
+  switch (unit) {
+    case TextUnit_Character:
+      new_position =
+          MoveEndpointByCharacter(position_to_move, count, units_moved);
+      break;
```
loic-sharma, line 816: Why was `TextUnit_Format` removed? Should there be a TODO?

yaakovschectman: I don't know, I will see if it can be added

third_party/accessibility/ax/platform/ax_platform_node_textrangeprovider_win.cc
yaakovschectman: Maybe. I do not know what actual use case calls this method.

third_party/accessibility/ax/platform/ax_platform_node_textrangeprovider_win.cc
yaakovschectman: Probably. It does not look like we actually use the class outside of this file

third_party/accessibility/ax/platform/ax_platform_node_textrangeprovider_win.cc
yaakovschectman: I see. We do not actually have `kFormatStart` and `kFormatEnd` in our `TextBoundary` enum.

third_party/accessibility/ax/platform/ax_platform_node_textrangeprovider_win.cc
yaakovschectman: I will put it to a TODO for now.

third_party/accessibility/ax/platform/ax_platform_node_textrangeprovider_win.cc
```diff
@@ -0,0 +1,1701 @@
...
+      start_->tree_id() != end_->tree_id()) {
+    AddObserver(end_);
+  }
+}
+
+void AXPlatformNodeTextRangeProviderWin::TextRangeEndpoints::AddObserver(
+    const AXPositionInstance& position) {
+  auto tree = AXTreeManagerMap::GetInstance()
+                  .GetManager(position->GetAnchor()->tree()->GetAXTreeID())
+                  ->GetTree();
```
loic-sharma, line 1554: It looks like Chromium finds the tree ID using the equivalent of `position->tree_id()`. Can you not use that here? Why not? Otherwise, would it make sense to make `AddObserver`'s and `RemoveObserver`'s input a tree ID instead of a position?

yaakovschectman: It appears the discrepancy is because Chromium recently changed and reverted these methods and their signatures. At the time I imported this code, Chromium used the version of the method seen here which takes the position instance as a parameter, but currently has changed it to take a tree ID instead.

third_party/accessibility/ax/platform/ax_platform_node_textprovider_win.cc
loic-sharma: Right. What are the resulting behavior differences while we don't have `AXMovementOptions`?

loic-sharma:
The code changes seem good, nice work! Could you detail your test plan once text providers are completed? Will it be manual validation using a screen reader? If so, what scenarios will you test?

third_party/accessibility/ax/platform/ax_platform_node_textprovider_win.cc
yaakovschectman: Presumably, there will not be an inherent difference between checking or not checking the initial position when moving by a boundary. I think for our purposes a similar effect can be achieved by the conditions used where the enclosing methods are called.

yaakovschectman:
@loic-sharma In particular, hovering the cursor over an element, moving the focus using the keyboard, and triggering a read of the current element (e.g. NVDA+Up). Once TextProviders are fully implemented, this will also include moving the carat, erasing characters, and changing the selection using the keyboard.

third_party/accessibility/ax/platform/ax_platform_node_textrangeprovider_win.cc
loic-sharma: Should we create an issue and leave a TODO then? This seems like a useful thing: https://github.com/MicrosoftDocs/win32/blob/docs/desktop-src/WinAuto/uiauto-uiautomationtextunits.md#format

third_party/accessibility/ax/platform/ax_platform_node_textrangeprovider_win.cc
loic-sharma: What do you think of updating the issue/comment to mention both internalization and case insensitivity?

loic-sharma (approved):
LGTM modulo pending comments. Nice work!

PR #21526, opened 2020-09-30
iOS: only add explicit transactions when there are platform views (only on main threads)

cyanglaz:
## Description

We only need to add explicit CATransactions on the main thread since platform views are only going to be on the main thread.
Also, adding explicit transactions seem to crash in certain image picker scenarios on iOS 14, so remove these transactions on the main thread will fix those crashes (on metal only) 

## Related Issues

https://github.com/flutter/flutter/issues/66647

## Tests

Not sure what we can do to test the rendering synchronization. I added a bunch of FML_DCHECKs so at least, the [scenario app tests](https://github.com/flutter/engine/tree/master/testing/scenario_app) in the engine should catch un-balanced CATransactions or CATransactions on the background thread. 

## Checklist

Before you create this PR confirm that it meets all requirements listed below by checking the relevant checkboxes (`[x]`). This will ensure a smooth and quick review process.

- [x] I read the [contributor guide] and followed the process outlined there for submitting PRs.
- [x] I signed the [CLA].
- [x] I read and followed the [C++, Objective-C, Java style guides] for the engine.
- [x] I read the [tree hygiene] wiki page, which explains my responsibilities.
- [x] I updated/added relevant documentation.
- [x] All existing and new tests are passing.
- [x] I am willing to follow-up on review comments in a timely manner.

## Breaking Change

Did any tests fail when you ran them? Please read [handling breaking changes].

- [x] No, no existing tests failed, so this is *not* a breaking change.
- [ ] Yes, this is a breaking change. *If not, delete the remainder of this section.*
   - [ ] I wrote a design doc: https://flutter.dev/go/template *Replace this with a link to your design doc's short link*
   - [ ] I got input from the developer relations team, specifically from: *Replace with the names of who gave advice*
   - [ ] I wrote a migration guide: https://flutter.dev/go/breaking-changes-template *Replace this with a link to a pull request that adds the migration guide to https://flutter.dev/docs/release/breaking-changes*

[issue database]: https://github.com/flutter/flutter/issues
[contributor guide]: https://github.com/flutter/flutter/wiki/Tree-hygiene#overview
[testing the engine]: https://github.com/flutter/flutter/wiki/Testing-the-engine
[C++, Objective-C, Java style guides]: https://github.com/flutter/engine/blob/master/CONTRIBUTING.md#style
[CLA]: https://cla.developers.google.com/
[tree hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[...]

PR #38551, opened 2022-12-29
[WIP][iOS] Avoid nextDrawable take much time.

luckysmg:
# WIP & Needs Triage🤔🤔🤔
### **This needs triage and needs more investigation. Since this change may have side effect, need to check that.**
**If this change is reasonable, I will modify the code to make it looks can be landed.**

### Before

https://user-images.githubusercontent.com/49340347/209905535-2c16eef6-2067-4587-b38b-7d670ab9d409.mp4

### After

https://user-images.githubusercontent.com/49340347/209905415-07d26180-ee58-436b-976f-72b43a332ea0.mp4

## This patch can solve:
- [x] The constant junk when scrolling randomly.
- [x] The constant junk when from bg to fg or swipe from control center.

## Physical devices I using for verify:
These all devices **can repo the constant junk bug and junk disappears with this patch:**

- [x] iPhone13Pro 
- [x] iPhone12 
- [x] iPhoneXS Max 
- [x] iPad Pro 

## The Pure Native iOS code to show this mechanism

<details>

<summary>Demo code</summary>

```swift

class TestVC: UIViewController {
    
    let metalLayer:CAMetalLayer = CAMetalLayer()
    
    var displayLink:CADisplayLink!
    
    var thread:Thread!
    
    override func touchesBegan(_ touches: Set<UITouch>, with event: UIEvent?) {
        self.displayLink.isPaused = true
    }
    
    override func touchesEnded(_ touches: Set<UITouch>, with event: UIEvent?) {
        self.displayLink.isPaused = false
    }
    
    override func viewDidLoad() {
        super.viewDidLoad()
        self.view.backgroundColor = .red
        let scale = UIScreen.main.scale
        metalLayer.frame = self.view.bounds
        metalLayer.drawableSize = CGSize(width: self.view.frame.width * scale, height: self.view.frame.height * scale)
        self.view.layer.addSublayer(metalLayer)
        
        self.addDisplayLinkInUITaskRunner()
        
        // Wait screen loading complete,and next,You can:
        
        // 1. Swipe down the control center, you will see a lot log print in console, and it will
        //                      constantly exists.
        
        // 2. Move the app to bg, and then move to fg, you will see a lot of log,
        //      it also will constantly exist for a long time.
        
        // 3. Just don't do anything, you might see it the log print constantly....
        //      And when you tap the screen, it may disappear, and After a while, you may see log again....
        
        
        // 4. Uncomment next line and rerun, and do same things in 1,2,3,
        // you will find that only a few of log will print, most of time there is not log ^_^.
        // If these are logs print, it is not constant, it will recover as normal after a short time,
        // which is what we want.
        // self.metalLayer.presentsWithTransaction = true
    }
    
    func addDisplayLinkInUITaskRunner() {
        self.thread = Thread(block: {
            RunLoop.current.add(NSMachPort(), forMode: .common)
            RunLoop.current.run()
        })
        self.thread.name = ""
        self.thread.start()
        
        self.perform(#selector(addDisplayLink), on: thread, with: nil, waitUntilDone: false)
    }
    
    @objc func addDisplayLink() {
        self.displayLink = CADisplayLink(target: self, selector: #selector(onDisplayLink))
        if #available(iOS 15.0, *) {
            self.displayLink.preferredFrameRateRange = .init(minimum: 60, maximum: 120, preferred: 120)
        } else {
            self.displayLink.preferredFramesPerSecond = 120
        }
        self.displayLink.add(to: .current, forMode: .common)
    }
    
    @objc private func onDisplayLink() {
        // Here should not be in main thread, because in flutter, layer renders in raster thread.
        assert(!Thread.current.isMainThread)
        
        let startTime = CACurrentMediaTime()
        let frameDrawable = metalLayer.nextDrawable()!
        let timeUsed = CACurrentMediaTime() - startTime
        
        // If time used to get next drawble over 3ms,
        // we print it here to indicate this method take much time!
        if (timeUsed > 0.003) {
            print("CAMetalLayer.nextDrawable take much time!! -> \(String(format: "%.2f", timeUsed * 1000)) ms")
        }
        frameDrawable.present()
    }
}

```

</details>

## Pre-launch Checklist

- [x] I signed the [CLA].


knopp:
This seems like it might need more investigation? Opting out of CATransaction should not affect frame rendering time.

luckysmg:
> This seems like it might need more investigation? Opting out of CATransaction should not affect frame rendering time.

Yes @knopp , this change needs triage and now this PR is just a draft and just show why it behave like this. Still Needs more investigation O(∩_∩)O

cyanglaz:
This seems also fixes https://github.com/flutter/flutter/issues/115036.

Is there a benefit of not rendering with transactions?

luckysmg:
> Is there a benefit of not rendering with transactions?

I don't know if there is any benefit render without transactions in raster thread (without platform view).... The comment just say `in raster thread there is not transactions like this`. But from result, this did impact the rendering time.

knopp:
I think the transactions enabled was breaking image picker at some point?

https://github.com/flutter/engine/pull/21526

luckysmg:
@knopp Thx for point that out. I will have a check later. O(∩_∩)O

knopp:
That is unfortunately. Conceivably we might need to do same thing we do on macOS, which is (bit simplified)
- from rater thread schedule block on main thread in which the surface is flipped
- block raster thread until it is done

luckysmg:
> That is unfortunately. Conceivably we might need to do same thing we do on macOS, which is (bit simplified)
> 
> * from rater thread schedule block on main thread in which the surface is flipped
> * block raster thread until it is done

Thx. I will have a try on that. ^_^

luckysmg:
Did some try and invesigation. Because the `PlatformView::NotifyCreated` and `PlatformView::NotifyDestroyed` use a latch to sync platform thread and raster thread. 

And in `IOSSurfaceMetalSkia::PresentDrawable`

```cpp

bool IOSSurfaceMetalSkia::PresentDrawable(GrMTLHandle drawable) const {
  if (drawable == nullptr) {
    FML_DLOG(ERROR) << "Could not acquire next Metal drawable from the SkSurface.";
    return false;
  }

  auto command_buffer =
      fml::scoped_nsprotocol<id<MTLCommandBuffer>>([[command_queue_ commandBuffer] retain]);
  [command_buffer.get() commit];
  [command_buffer.get() waitUntilScheduled];

  // Use sync and we can wait this operation end.
  dispatch_sync(dispatch_get_main_queue(), ^{
    FML_DCHECK([NSThread isMainThread]);
    [CATransaction begin];
    [CATransaction setDisableActions:YES];
    [reinterpret_cast<id<CAMetalDrawable>>(drawable) present];
    [CATransaction commit];
  });
  return true;
}

```

And it is easy to get dead lock in add-to-app scene.....

knopp:
This shouldn't dead lock on `PresentDrawable`, right? Because at that point rasterizer is not yet presenting anything, we're just creating surface, am I wrong?

For teardown it is more complicated. The code is structured differently on desktop so I'm not sure how this would translate, but on shutdown, we first call engine->NotifyDestroyed, which down the line results in FlutterView getting notified, which unblocks the raster thread:

https://github.com/flutter/engine/blob/a19cf360a4706cdd351682cd7b9a9a6804a26d33/shell/platform/darwin/macos/framework/Source/FlutterThreadSynchronizer.mm#L105-L110

Since this is called from platform thread, it executes all blocks that raster thread is waiting for and then makes sure that raster thread won't block on anything after.

luckysmg:
Have some try. Present on platform thread is ok, emm but when platform thread has some work to do, the drawable will be delayed to be presented... Motion more sensitive for users on a mobile phone than on a desktop.....

And recently, I did some investigation, surprisingly found this change will also work, and seems has no side effect. Maybe can also take a look that. 
- https://github.com/flutter/engine/pull/39359

PR #38562, opened 2022-12-29
[flutter roll] Revert "[web] Don't overwrite editing state with semantic updates"

XilaiZhang:
Reverts flutter/engine#38271.

fails education tests. context: https://buganizer.corp.google.com/issues/263966258

would be great if web team could help take a look. Thank you!


htoor3 (approved):

XilaiZhang:
this was confirmed via reverting both copies of source file. Merging now

PR #38573, opened 2022-12-30
[web] use a render target instead of a new surface for Picture.toImage

jonahwilliams:
Attempts to fix https://github.com/flutter/flutter/issues/117786

I don't have full context into the original problem this change was solving, but in the general case it is incredibly slow. Using a different render target instead of a new glContext is significantly faster.

This does hit an issue where we are perhaps setting up some of the rendering state too late, but that only seems to happen once.


jonahwilliams:
This currently doesn't handle the destruction of the surface when resizing the window

jonahwilliams:
We shouldn't be losing the context on resize anyway: https://github.com/flutter/flutter/issues/117803

lib/web_ui/lib/src/engine/canvaskit/picture.dart
```diff
@@ -102,26 +100,13 @@ class CkPicture extends ManagedSkiaObject<SkPicture> implements ui.Picture {
   @override
   ui.Image toImageSync(int width, int height) {
     assert(debugCheckNotDisposed('Cannot convert picture to image.'));
-    final Surface surface = SurfaceFactory.instance.pictureToImageSurface;
-    final CkSurface ckSurface =
-      surface.createOrUpdateSurface(ui.Size(width.toDouble(), height.toDouble()));
+    final CkSurface ckSurface = SurfaceFactory.instance.baseSurface
```
hterkelsen, line 103: dispose of this surface once we have created the image?

jonahwilliams:
update: holding off on this change util runtime effect support lands in the engine so I can easily test it

jonahwilliams:
Verified that samplers still work with this change, marking as ready for review

jonahwilliams:
Update: working on the case where there is no GL context, we may need differend code to fall back to a surface.

lib/web_ui/lib/src/engine/canvaskit/surface.dart
```diff
@@ -141,6 +141,37 @@ class Surface {
   ui.Size? _currentSurfaceSize;
   double _currentDevicePixelRatio = -1;
 
+  /// This is only valid after the first frame or if [ensureSurface] has been
+  /// called
+  bool get usingSoftwareBackend => _glContext == null ||
+      _grContext == null || webGLVersion == -1 || !configuration.canvasKitForceCpuOnly;
```
hterkelsen, line 147: don't invert `configuration.canvasKitForceCpuOnly`

lib/web_ui/lib/src/engine/canvaskit/surface.dart
```diff
@@ -141,6 +141,37 @@ class Surface {
...
+  void ensureSurface() {
+    // If the GrContext hasn't been setup yet then we need to force initialization
+    // of the canvas and initial surface.
+    if (_surface != null) {
+      return;
+    }
+    // TODO(jonahwilliams): this is somewhat wasteful. We should probably
+    // eagerly setup this surface instead of delaying until the first frame?
+    // Or at least cache the estimated window size.
+    createOrUpdateSurface(const ui.Size(1, 1));
```
hterkelsen, line 160: maybe `ensureSurface` can take an optional `Size`?

hterkelsen (approved):
LGTM with small change

lib/web_ui/lib/src/engine/canvaskit/surface.dart
jonahwilliams: Oops, good catch. Fixed

lib/web_ui/lib/src/engine/canvaskit/surface.dart
jonahwilliams: Done

PR #38576, opened 2022-12-30
[web] retain GL/Gr context on window resize

jonahwilliams:
The current implementation of the canvaskit renderer abandons the current gl context and creates a new one when resizing the window. This is an expensive operation, and causes the loss of all GPU backed textures associated with the previous context. At least part of the issue in https://github.com/flutter/flutter/issues/117786 is due to using a separate context, which I think is due partially to needing to deal with this on resize.

I don't think this should be necessary, its fairly straightforward to resize a canvas using webgl/webgl2 (see https://webgl2fundamentals.org/webgl/lessons/webgl-resizing-the-canvas.html ). If we're missing APIs in Skia/Canvaskit then we should get those added instead of working around the problem.

Fixes https://github.com/flutter/flutter/issues/117803


hterkelsen:
mostly LGTM but we should probably touch this class up now that we should be a one-to-one relationship between Surface and HTML Canvas (whereas before one Surface could have many HTML canvases over its lifetime) so things like `createNewHtmlCanvas` should probably be refactored out

jonahwilliams:
Its a bit hard to tell in flutter/gallery, because at a certain size it switches layouts and that frame is fairly expensive anyway

Before:

image

After:

image

My read of the profiles is that the long frame is roughly 2x as long in the former one, its just not all in JS time

jonahwilliams:
It also seems like we can keep using the same CkSurface as long as the canvas size didn't change

jonahwilliams:
@hterkelsen this is (mostly) updated now. It seems like we do need to create new CkSurfaces whenever the size of the underlying canvas changes, so we can still have multiple CkSurfaces. If we can fix the lookup of the GlParameters on surface creation this will be significantly faster too

jonahwilliams:
Looks like we do need to recreate the surface, filled https://github.com/flutter/flutter/issues/117934 as a follow up

hterkelsen (approved):

jonahwilliams:
Thanks!

PR #38577, opened 2022-12-30
[Impeller Scene] Import materials, load embedded textures

bdero:
Adds automated import/deserialization of materials and embedded textures! Removed the material hacks from the playgrounds.

The "Flutter Logo" playground loads the unlit material/embedded texture automatically during the deserialization routine, and the "Two Triangles" playground is now displaying the material base color factors:

https://user-images.githubusercontent.com/919017/210055963-38dbeaca-b645-4766-8d0a-beaff58f7e71.mov

The vertex color blending is also now turned off by default when an albedo/base color texture is present.


chinmaygarde (approved):
Awesome!

PR #38602, opened 2023-01-02
PoC: SVG canvas

yjbanov:
A proof-of-concept for a canvas that uses SVG exclusively to render all pictures. Reuses HTML renderer's layer tree and text layout. Only replaces the `PersistedPicture` implementation. Does not implement things that are not implementable in SVG with expected correctness and performance characteristics: `drawVertices` and `drawPoints`.

This is currently in a state where it can render the gallery 99% correctly, but probably contains a long tail of small issues that would need to be fixed before this can turn into anything serious. In particular, this needs a higher coverage of `Paint` features, such as gradients.

## Notes on performance

Tested on M1 MacBook Pro, iPhone 7, and Pixel 6 Pro.

Good news:

* Eliminates virtually all 2d canvases (required https://github.com/flutter/engine/pull/38640).
* Uses less GPU memory compared to HTML (on the Gallery drops from ~40 MB down to ~30 MB).
* Just eyeballing the frame rate, feels competitive with the HTML renderer on Firefox and Chromium-based browsers.

Bad news:

* In Safari there are cases where performance has severely regressed. One example is the [Reply demo in the gallery](https://butterfly-flutters.web.app/svg/#/reply). The transition when expanding and collapsing emails is super-janky.
* The material text fields demo severely regressed on Android.

## Next steps

I don't think this is ready for anything close to production. We should at least eliminate the obvious regressions. We could land this experimentally without enabling it, so we can iterate on it. All changes outside `lib/src/engine/html/svg` directory are meant to be non-breaking. They just generalize existing code so it can be used both in the HTML mode and SVG mode. If we decide to abandon this experiment, we can simply delete `lib/src/engine/html/svg` and a couple of lines that refer to it, and forget this ever happened. The actual "SVG canvas" part is ~450 LoC (`svg_canvas.dart` + `svg_picture.dart`).

The other thing is that this should _not_ become yet another canvas, but a replacement for both `dom_canvas.dart` and `bitmap_canvas.dart`. If we can't get it to a point that it fully replaces the old canvases, we should drop it.


lib/web_ui/lib/src/engine/text/canvas_paragraph.dart
```diff
@@ -185,6 +187,66 @@ class CanvasParagraph implements ui.Paragraph {
...
+        final SVGTextElement spanElement = createSVGTextElement();
+        applyTextStyleToSvgElement(element: spanElement, style: fragment.style);
+
+        final ui.Rect boxRect = fragment.toPaintingTextBox().toRect();
+        spanElement.addX(boxRect.left);
+        // SVG places <text> elements such that the text baseline coincides with
+        // the Y coordinate. However, Flutter positions the top edge of the text
+        // line at the Y coordinate. For the SVG text to match what Flutter
+        // expects, the baseline is added to the Y coordinate.
+        spanElement.addY(boxRect.top + line.baseline);
```
mdebbar, line 235: (I think you already did this but haven't pushed it yet)
```suggestion
        spanElement.addY(line.baseline);
```

mdebbar (commented):
This is amazing! Thanks for hacking on this, @yjbanov!

lib/web_ui/lib/src/engine/embedder.dart
```diff
@@ -171,8 +171,9 @@ class FlutterViewEmbedder {
     _glassPaneShadow = glassPaneElementHostNode;
 
     // Don't allow the scene to receive pointer events.
-    _sceneHostElement = domDocument.createElement('flt-scene-host')
-      ..style.pointerEvents = 'none';
+    _sceneHostElement = domDocument.createElement('flt-scene-host');
+      // DO_NOT_SUBMIT: revert this before submitting.
```
ditman, line 175: Making the PR red :P

ditman (changes requested):

lib/web_ui/lib/src/engine/dom.dart
```diff
@@ -1740,3 +1762,215 @@ DomV8BreakIterator createV8BreakIterator() {
...
+@JS()
+@anonymous
+@staticInterop
+class DOMPointInit {
+  external factory DOMPointInit({
+    double x = 0,
+    double y = 0,
+    double z = 0,
+    double w = 1,
+  });
```
ditman, line 1785: I normally don't separate the `external factory` from the staticInterop class that it creates. I think you should be able to do this:

```dart
@JS()
@staticInterop
class DOMPoint {
  external factory DOMPoint({
    double x = 0,
    double y = 0,
    double z = 0,
    double w = 1,
  });
}

extension DOMPointExtension on DOMPoint {
  external double get x;
  external double get y;
  external double get z;
  external double get w;
}
```

(And then the types of stuff elsewhere can receive a `DOMPoint` and a `DOMMatrix` rather than the `...Init` versions?)

lib/web_ui/lib/src/engine/embedder.dart
```diff
@@ -171,8 +171,9 @@ class FlutterViewEmbedder {
...
+      // ..style.pointerEvents = 'none';
```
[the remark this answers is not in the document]
Hixie: fwiw, one trick for this kind of thing is to use `// TODO: revert before submitting` because then the todo style lint will make the analyzer shard red, preventing you from landing.

yjbanov:
Couldn't schedule time to investigate this further. In particular, it's still unclear how to deal with the poor Safari performance until Safari itself optimizes SVG filters (there are bugs filed in WebKit dating back to 2012 on this, but no movement so far). Finally, the recent improvements in the size and performance of CanvasKit are getting closer to removing the need for an HTML renderer altogether. So spending time on this has dubious value. Closing for now.

PR #38617, opened 2023-01-03
Roll quiver to 3.2.1

jason-simmons:
Need to roll quiver to a version that doesn't depend on BidirectionalIterator, which is being removed in https://dart-review.googlesource.com/c/sdk/+/276741?tab=checks.


jason-simmons:
@mit-mit

zanderso (approved):

mit-mit:
Thanks!

PR #38634, opened 2023-01-04
Limit selection change to focused node on Windows

yaakovschectman:
When the text selection of a node is changed, we should only set the AX tree's selection to that of the focused node. Currently, we set it for every node in the update, which means that whichever node happens to be last will have the selection that actually persists, even if it is not the focused node. The order of nodes when iterating through an update is meaningless. The focused flag is added to one unit test as well.

Part of https://github.com/flutter/flutter/issues/116219

## Pre-launch Checklist

- [ ] I signed the [CLA].

### DIFF
--- a/shell/platform/windows/accessibility_bridge_windows.cc
+++ b/shell/platform/windows/accessibility_bridge_windows.cc
@@ -54,8 +54,8 @@ void AccessibilityBridgeWindows::OnAccessibilityEvent(
             std::static_pointer_cast<FlutterPlatformNodeDelegateWindows>(
                 focus_delegate);
       }
-      DispatchWinAccessibilityEvent(win_delegate,
-                                    ax::mojom::Event::kDocumentSelectionChanged);
+      DispatchWinAccessibilityEvent(
+          win_delegate, ax::mojom::Event::kDocumentSelectionChanged);
       break;
     }
     case ui::AXEventGenerator::Event::FOCUS_CHANGED:
--- a/shell/platform/windows/accessibility_bridge_windows.cc
+++ b/shell/platform/windows/accessibility_bridge_windows.cc
@@ -41,10 +41,16 @@ void AccessibilityBridgeWindows::OnAccessibilityEvent(
       DispatchWinAccessibilityEvent(win_delegate,
                                     ax::mojom::Event::kChildrenChanged);
       break;
-    case ui::AXEventGenerator::Event::DOCUMENT_SELECTION_CHANGED:
+    case ui::AXEventGenerator::Event::DOCUMENT_SELECTION_CHANGED: {
+      ui::AXNode::AXID focus_id = GetAXTreeData().sel_focus_object_id;
+      auto focus_delegate =
+          GetFlutterPlatformNodeDelegateFromID(focus_id).lock();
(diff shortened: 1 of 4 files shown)


shell/platform/windows/accessibility_bridge_windows.cc
```diff
@@ -41,10 +41,16 @@ void AccessibilityBridgeWindows::OnAccessibilityEvent(
       DispatchWinAccessibilityEvent(win_delegate,
                                     ax::mojom::Event::kChildrenChanged);
       break;
-    case ui::AXEventGenerator::Event::DOCUMENT_SELECTION_CHANGED:
+    case ui::AXEventGenerator::Event::DOCUMENT_SELECTION_CHANGED: {
+      ui::AXNode::AXID focus_id = GetAXTreeData().sel_focus_object_id;
+      auto focus_delegate =
+          GetFlutterPlatformNodeDelegateFromID(focus_id).lock();
```
loic-sharma, line 47: Are we guaranteed to get a focus delegate? I see the corresponding Chrome logic checks if it was able to find a focus object: https://source.chromium.org/chromium/chromium/src/+/main:content/browser/accessibility/browser_accessibility_manager_win.cc;l=235-237?q=case%20ui::AXEventGenerator::Event::DOCUMENT_SELECTION_CHANGED&ss=chromium

shell/platform/windows/accessibility_bridge_windows.cc
```diff
@@ -41,10 +41,16 @@ void AccessibilityBridgeWindows::OnAccessibilityEvent(
       DispatchWinAccessibilityEvent(win_delegate,
                                     ax::mojom::Event::kChildrenChanged);
       break;
-    case ui::AXEventGenerator::Event::DOCUMENT_SELECTION_CHANGED:
+    case ui::AXEventGenerator::Event::DOCUMENT_SELECTION_CHANGED: {
```
loic-sharma, line 44: Could we add a comment similar to Chromium's comment? https://source.chromium.org/chromium/chromium/src/+/main:content/browser/accessibility/browser_accessibility_manager_win.cc;l=230-234?q=case%20ui::AXEventGenerator::Event::DOCUMENT_SELECTION_CHANGED&ss=chromium

loic-sharma: Does this part of the comment make sense for us too?

> This is because the focus is the only endpoint that can move, and because the caret (if present) is at the focus.

yaakovschectman: I do not think so, as we do not have the carat as its own node.

shell/platform/windows/accessibility_bridge_windows.cc
```diff
@@ -41,10 +41,23 @@ void AccessibilityBridgeWindows::OnAccessibilityEvent(
...
+      // An event indicating a change in document selection should be fired
+      // only for the focused node whose selection has changed. If a valid
+      // carat and selection exist in the app tree, they must both be within
```
loic-sharma, line 47:
```suggestion
      // caret and selection exist in the app tree, they must both be within
```

loic-sharma (approved):

PR #38635, opened 2023-01-04
[flutter roll] quick fix to import map

XilaiZhang:
context: https://b.corp.google.com/issues/264388135

quick PR to unblock flutter roll.

### DIFF
--- a/impeller/scene/animation/animation_player.h
+++ b/impeller/scene/animation/animation_player.h
@@ -6,7 +6,7 @@
 
 #include <memory>
 #include <optional>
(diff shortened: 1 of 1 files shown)


zanderso (approved):
This looks right to me if the presubmits are green. Is this the only instance that needs to be updated?

XilaiZhang:
> This looks right to me if the presubmits are green. Is this the only instance that needs to be updated?

Thanks for the quick review! Yeah animation_player.h is the only place

bdero (approved):
Thanks!

PR #38640, opened 2023-01-04
[web] Fix canvas2d leaks in text measurement

mdebbar:
Fixes https://github.com/flutter/flutter/issues/117164
Could also help with https://github.com/flutter/flutter/issues/117714


lib/web_ui/lib/src/engine/text/layout_service.dart
```diff
@@ -16,6 +16,19 @@ import 'paragraph.dart';
...
+/// A single canvas2d context to use for all text measurements.
+final DomCanvasRenderingContext2D _context = _createTextMeasurementCanvas();
+
+DomCanvasRenderingContext2D _createTextMeasurementCanvas() {
+  final DomCanvasElement canvas = createDomCanvasElement();
+  // We don't use this canvas to draw anything, so let's make it as small as
+  // possible to save memory.
+  canvas
+    ..width = 0
+    ..height = 0;
```
yjbanov, line 28: I think we should be able to test that we use a 0x0 canvas.

mdebbar: Added a test that checks for number of canvases created as well as the size.

yjbanov (approved):
lgtm

mdebbar:
Safari does weird things (I was able to confirm this on a jsfiddle). When you do:
```dart
context.font = "bold 20px FontFamily1";
```
the canvas measures and renders the text correctly using the given style (bold, 20px, FonFamily1). But when you read the value again:
```dart
print(context.font); // "20px FontFamily1" -- no "bold".
```
the "bold" part of the string disappears!

I adjusted the test to avoid this.

PR #38644, opened 2023-01-04
[Impeller] support vertices per-vertex color blending with all color sources

jonahwilliams:
I believe this correctly solves _most_ of https://github.com/flutter/flutter/issues/108047 for Vertices, assuming that the approach I'm taking with multiple contents is reasonable. I'm sure there are ways to optimize this a bit more for certain blends.

This also doesn't cover the texture coordinates issue.

If this works, ideally we'd use a similar approach to drawAtlas since it also allows for a per-vertex color that acts as a destination.


jonahwilliams:
Working on a larger playground for this right now. I suspect I'd need to do some sort of subpass to make the blend mode work correctly, though this is at least approximately the right thing for some of the simple modes

jonahwilliams:
For example:

image

The background color changes, probably because I'm applying the blend in a weird way

jonahwilliams:
Closing in favor of https://github.com/flutter/engine/pull/38335

PR #38647, opened 2023-01-04
[fuchsia][scenic] Use infinite hit region

jaeheon:
fxbug.dev/118729

This patchset uses the new "infinite" hit region offered by Flatland. It makes the input interceptor less error prone. 

Issue 118006

## Pre-launch Checklist

- [x I read and followed the [Flutter Style Guide] and the [C++, Objective-C, Java style guides].
- [x] I signed the [CLA].

### DIFF
--- a/shell/platform/fuchsia/flutter/tests/flatland_external_view_embedder_unittests.cc
+++ b/shell/platform/fuchsia/flutter/tests/flatland_external_view_embedder_unittests.cc
@@ -38,8 +38,10 @@ using fuchsia::scenic::scheduling::FuturePresentationTimes;
 using fuchsia::scenic::scheduling::PresentReceivedInfo;
 using ::testing::_;
 using ::testing::AllOf;
+using ::testing::Contains;
 using ::testing::ElementsAre;
 using ::testing::Eq;
+using ::testing::Field;
 using ::testing::FieldsAre;
 using ::testing::IsEmpty;
 using ::testing::Matcher;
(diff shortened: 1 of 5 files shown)


abrush21 (commented):
Changes LGTM. I think we could add a unit test where intercept_all_input = true, which would exercise the new behavior.

jaeheon:
Yes, agreed. Will work on it, then push a next version.

Jaeheon

jaeheon:
Alex, PTAL. Fakes updated, test cobbled together. I tried trimming the test down, but LMK if there's more I can cut out. Also I filed fxbug.dev/118945 to track getting Flutter repo using Fuchsia SDK's Fake Flatland. Likely Emircan will get to it soon.

shell/platform/fuchsia/flutter/tests/flatland_external_view_embedder_unittests.cc
```diff
@@ -1476,4 +1491,143 @@ TEST_F(FlatlandExternalViewEmbedderTest, SimpleScene_OverlappingHitRegions) {
...
+  const fuchsia::math::SizeU frame_size{
+      static_cast<uint32_t>(frame_size_signed.width()),
+      static_cast<uint32_t>(frame_size_signed.height())};
+  DrawFrameWithView(
+      external_view_embedder, frame_size_signed, kDPR, child_view_id,
+      child_view_params,
+      [](SkCanvas* canvas) {
+        const SkSize canvas_size = SkSize::Make(canvas->imageInfo().width(),
+                                                canvas->imageInfo().height());
+        SkPaint rect_paint;
```
abrush21, line 1577: I'm not 100% sure if the test would still work, but you _might_ be able to cut out the draw operation (here and below), in which case you would just pass an empty vector for the expected hit regions to `IsImageLayer`.

Besides that, I'm not sure what else you could remove from this test to make it more concise.

abrush21 (approved):

shell/platform/fuchsia/flutter/tests/flatland_external_view_embedder_unittests.cc
jaeheon: Thank you for the review. Good suggestion, I played around with it a bit, but didn't work. I kinda recall that optimizations look for draw calls, so removing them might even be undesirable if it warps the transform structure we want to check against.

PR #38665, opened 2023-01-05
Remove strict equality check for SkMatrix comparison

lin-erik:
Fixes b/262417298

### DIFF
--- a/shell/platform/fuchsia/flutter/gfx_external_view_embedder.cc
+++ b/shell/platform/fuchsia/flutter/gfx_external_view_embedder.cc
@@ -328,15 +328,13 @@ void GfxExternalViewEmbedder::SubmitFrame(
             kScenicZElevationBetweenLayers * scenic_layer_index +
             embedded_views_height;
 
-        if (view_mutators.total_transform != view_params.transformMatrix()) {
-          FML_LOG(ERROR) << "Failed assertion: view_mutators.total_transform "
-                            "!= view_params.transformMatrix()";
-          FML_LOG(ERROR) << "view_mutators.total_transform:";
-          view_mutators.total_transform.dump();
-          FML_LOG(ERROR) << "view_params.transformMatrix():";
-          view_params.transformMatrix().dump();
-          FML_LOG(FATAL) << "view_mutators.total_transform "
-                            "!= view_params.transformMatrix()";
+        // Use built-in get method for SkMatrix to get values
+        // See:
+        // https://source.corp.google.com/piper///depot/google3/third_party/skia/HEAD/include/core/SkMatrix.h;l=391
+        for (int index = 0; index < 9; index++) {
+          FML_CHECK(SkScalarNearlyEqual(
+              view_mutators.total_transform.get(index),
+              view_params.transformMatrix().get(index), 0.0005f));
         }
 
         // Set clips for the platform view.


shell/platform/fuchsia/flutter/gfx_external_view_embedder.cc
```diff
@@ -328,15 +328,13 @@ void GfxExternalViewEmbedder::SubmitFrame(
...
-          FML_LOG(ERROR) << "view_mutators.total_transform:";
-          view_mutators.total_transform.dump();
-          FML_LOG(ERROR) << "view_params.transformMatrix():";
-          view_params.transformMatrix().dump();
-          FML_LOG(FATAL) << "view_mutators.total_transform "
-                            "!= view_params.transformMatrix()";
+        // Use built-in get method for SkMatrix to get values
+        // See:
+        // https://source.corp.google.com/piper///depot/google3/third_party/skia/HEAD/include/core/SkMatrix.h;l=391
+        for (int index = 0; index < 9; index++) {
```
akbiggs, line 334: Let's surround this in `#ifdef NDEBUG` to make it a debug only assertion. I don't see any reason to do this on every frame in a release build.

shell/platform/fuchsia/flutter/gfx_external_view_embedder.cc
```diff
@@ -328,15 +328,13 @@ void GfxExternalViewEmbedder::SubmitFrame(
...
-        if (view_mutators.total_transform != view_params.transformMatrix()) {
-          FML_LOG(ERROR) << "Failed assertion: view_mutators.total_transform "
-                            "!= view_params.transformMatrix()";
-          FML_LOG(ERROR) << "view_mutators.total_transform:";
-          view_mutators.total_transform.dump();
-          FML_LOG(ERROR) << "view_params.transformMatrix():";
-          view_params.transformMatrix().dump();
-          FML_LOG(FATAL) << "view_mutators.total_transform "
-                            "!= view_params.transformMatrix()";
+        // Use built-in get method for SkMatrix to get values
```
akbiggs, line 331: A comment first like `// Verify that we're unpacking the mutators' transform matrix correctly on debug builds.` might be useful.

shell/platform/fuchsia/flutter/gfx_external_view_embedder.cc
```diff
@@ -328,15 +328,13 @@ void GfxExternalViewEmbedder::SubmitFrame(
...
+        // See:
+        // https://source.corp.google.com/piper///depot/google3/third_party/skia/HEAD/include/core/SkMatrix.h;l=391
+        for (int index = 0; index < 9; index++) {
+          FML_CHECK(SkScalarNearlyEqual(
```
akbiggs, line 335: FML_CHECK doesn't give good feedback on why the values were unequal, so should we do something like:

```c++
for (...) {
  const float mutators_transform_value = view_mutators.total_transform.get(index);
  const float params_transform_value = view_params.transformMatrix().get(index);
  if (!SkScalarNearlyEqual(mutators_transform_value, params_transform_value, 0.0005f)) {
    FML_LOG(FATAL)
        << "Assertion failed: view_mutators.total_transform[" << index
        << "] (" << mutators_transform_value << ") != view_params.transformMatrix()[" << index
        << "] (" << params_transform_value << "). This likely means there is a bug with the "
        << "logic for parsing embedded views' transform matrices.";
  }
}
```

akbiggs (approved):

PR #38682, opened 2023-01-06
Ability to disable browser context menu

justinmc:
The goal of this PR is to create a way to disable the browser-rendered context menu on web.

The main reason that someone might want to do this is to use a custom Flutter-rendered context menu instead, which this PR will unblock.

After this PR:

https://user-images.githubusercontent.com/389558/211122149-1b211cd0-faaa-4013-a60a-8d3d9ce1194b.mov

### The current, basic approach
I'm just calling `event.preventDefault()` on `contextmenu` events at the root `domWindow` level.  I've tied this to platform channel methods that could be exposed in the framework.

```dart
  static final DomEventListener _handleContextMenu = allowInterop((DomEvent event) {
    event.preventDefault();
  });

  static void disableContextMenu() {
    domWindow.addEventListener('contextmenu', _handleContextMenu);
  }

  static void enableContextMenu() {
    domWindow.removeEventListener('contextmenu', _handleContextMenu);
  }
```

### Alternative, more fully featured approach
It might be nice if, instead of toggling the context menus for the whole app, users could toggle it for a widget subtree.  Would it be at all possible to do this?  I guess there would have to be a specific DOM element for every given widget that we could add/remove the listener on.

With either approach, please excuse my lack of web engine knowledge and let me know if I'm on the right track!

### References

Part of: https://github.com/flutter/flutter/issues/78671
Framework PR: https://github.com/flutter/flutter/pull/118194

<details>

<summary>Example code</summary>

```dart
import 'package:flutter/services.dart';

void main() {
  runApp(const MyApp());
}

  const MyApp({super.key});

  @override
    return MaterialApp(
      title: 'Flutter Demo',
      theme: ThemeData(
        primarySwatch: Colors.blue,
      ),
      home: const MyHomePage(title: 'Flutter Demo Home Page'),
    );
  }
}

class MyHomePage extends StatefulWidget {
  const MyHomePage({super.key, required this.title});

  final String title;

  @override
  State<MyHomePage> createState() => _MyHomePageState();
}

class _MyHomePageState extends State<MyHomePage> {
  bool _contextMenuEnabled = true;

  void _toggleContextMenu() {
    if (_contextMenuEnabled) {
      ContextMenu.disableContextMenu();
    } else {
      ContextMenu.enableContextMenu();
    }
    setState(() {
      _contextMenuEnabled = !_contextMenuEnabled;
    });
  }

  @override
    return Scaffold(
      appBar: AppBar(
        title: Text(widget.title),
      ),
      floatingActionButton: FloatingActionButton(
        onPressed: _toggleContextMenu,
        tooltip: _contextMenuEnabled ? 'Disable' : 'Enable',
        child: Icon(_contextMenuEnabled ? Icons.thumb_down: Icons.thumb_up),
      ),
      body: Center(
        child: Column(
          mainAxisAlignment: MainAxisAlignment.center,
          children: <Widget>[
            const TextField(),

            // This container does not show a context menu.
            GestureDetector(
              onSecondaryTapDown: (TapDownDetails details) async {
                await ContextMenu.disableContextMenu();
                await Future.delayed(const Duration(seconds: 0));
                ContextMenu.enableContextMenu();
              },
              child: Container(
                width: 200,
                height: 200,
                color: Colors.tealAccent,
              ),
            ),
          ],
        ),
      ),
    );
  }
}
```

</details>


lib/web_ui/lib/src/engine/context_menu.dart
```diff
@@ -0,0 +1,37 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
```
mdebbar, line 1: The path of this file needs to be added here:
https://github.com/flutter/engine/blob/eb5c6f0b4ed94325e9c9811cc8681d687e0f5c8a/ci/licenses_golden/licenses_flutter#L1910-L1914

lib/web_ui/lib/src/engine/context_menu.dart
```diff
@@ -0,0 +1,37 @@
...
+    event.preventDefault();
+  });
+
+  /// Disables the browser's context menu.
+  ///
+  /// By default, when a Flutter web app starts, the context menu is enabled.
+  ///
+  /// Can be re-enabled by calling [enableContextMenu].
+  static void disableContextMenu() {
+    domWindow.addEventListener('contextmenu', _handleContextMenu);
```
mdebbar, line 26: The flutter app could be embedded inside a host element (instead of taking over the entire window). We should add the event listener to the host element. cc @ditman for the right way to do this.

mdebbar (commented):
Both PRs (engine & framework) look good to me overall. They are still in draft mode, so I'll assume that tests will be added later.

mdebbar:
> It might be nice if, instead of toggling the context menus for the whole app, users could toggle it for a widget subtree. Would it be at all possible to do this?

This isn't possible because the entire app could be rendered on a canvas element.

To achieve this, you want the pointer event listener on the framework side to decide whether it wants to show a context menu or not (e.g. based on which widget was hit tested). And it should immediately tell the engine that a context menu shouldn't be shown for this event.

justinmc:
Thanks for the review!  Understood about specific widget subtrees.  I just played around with using the API in this PR to quickly disable and then reenable the context menu for a specific GestureDetector in the framework, and it worked, which is a good endorsement for this API.

I'm going to go ahead with cleaning these two PRs up and writing tests for them when I'm back next week, and I'll let you know when they're ready for a real review.

lib/web_ui/lib/src/engine/context_menu.dart
ditman: The `contextmenu` event bubbles ([spec](https://w3c.github.io/uievents/#event-type-contextmenu)), so this should be enough?

If this needs to be aware of the embedding, the enable/disable methods could be added to the `EmbeddingStrategy` abstract class (and its implementations): 

https://github.com/flutter/engine/blob/main/lib/web_ui/lib/src/engine/view_embedder/embedding_strategy/embedding_strategy.dart#L23

mdebbar: > The `contextmenu` event bubbles ([spec](https://w3c.github.io/uievents/#event-type-contextmenu)), so this should be enough?

Yes, but it's more than enough :) It disables the context menu for the entire page, which we don't want in the host element mode.

ditman: You're very right :P

lib/web_ui/test/engine/view_embedder/embedding_strategy/custom_element_embedding_strategy_test.dart
```diff
@@ -121,4 +121,33 @@ void doTests() {
...
+
+    tearDown(() {
+      target.remove();
+    });
+
+    test('disableContextMenu and enableContextMenu can toggle the context menu', () {
+      expect(strategy.contextMenuEnabled, isTrue);
+
+      strategy.disableContextMenu();
+      expect(strategy.contextMenuEnabled, isFalse);
```
justinmc, line 141: Is there anything else I could expect here, like could I somehow verify that the event listener was added to the dom element?

lib/web_ui/lib/src/engine/context_menu.dart
justinmc: Thanks for the advice here!  I've moved the enable/disable methods to the embedding strategy and made them accessible via FlutterViewEmbedder.

lib/web_ui/test/engine/view_embedder/embedding_strategy/custom_element_embedding_strategy_test.dart
ditman: You could add your own `contextmenu` Event handler to the root DOM element of the strategy in the test, and check  if [`isDefaultPrevented`](https://developer.mozilla.org/en-US/docs/Web/API/Event/defaultPrevented) has been set to `true`. (IIRC, once you prevent the default in a handler, the event will be modified to set `defaultPrevented` to true.)

In order to programmatically trigger the event, you should be able to [`dispatchEvent`](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/dispatchEvent) a [`contextmenu` event](https://developer.mozilla.org/en-US/docs/Web/API/Element/contextmenu_event) using the root DOM element of the strategy.

mdebbar: There are some examples in our code base. E.g. https://github.com/flutter/engine/blob/7bff5dd115d794200266c523f332c408c1da25c8/lib/web_ui/test/raw_keyboard_test.dart#L57

You don't have to create an event listener on the root element. Just check `defaultPrevented` on the event you dispatched.

lib/web_ui/lib/src/engine/platform_dispatcher.dart
```diff
@@ -531,6 +530,21 @@ class EnginePlatformDispatcher extends ui.PlatformDispatcher {
         textEditing.channel.handleTextInput(data, callback);
         return;
 
+      case 'flutter/contextmenu':
```
justinmc, line 533: I've omitted any web or browser namespacing (not "flutter/webcontextmenu") because other methods don't seem to include platform names like that.  And maybe in the future this could be used on other platforms.

In the framework PR I'll make the API specific (BrowserContextMenu.disable or something like that).

lib/web_ui/test/engine/view_embedder/embedding_strategy/custom_element_embedding_strategy_test.dart
justinmc: That worked, thank you all for the guidance here!  I think the tests are much more robust this way, and I don't have to use `@visibleForTest`.

ditman: > Just check defaultPrevented on the event you dispatched.

Damn this approach is good! Thanks for the tests @justinmc!

justinmc:
@mdebbar This and the framework PR (https://github.com/flutter/flutter/pull/118194) are now green and ready for re-review when you have a chance.  Thank you!

lib/web_ui/lib/src/engine/view_embedder/embedding_strategy/full_page_embedding_strategy.dart
```diff
@@ -48,6 +48,12 @@ class FullPageEmbeddingStrategy extends EmbeddingStrategy {
     registerElementForCleanup(resourceHost);
   }
 
+  @override
+  void disableContextMenu() => disableContextMenuOn(domDocument.body!);
```
mdebbar, line 52: We usually set these event listeners on `window` not `document.body`. In order to do that, you want to change the signature of `disableContextMenuOn`/`enableContextMenuOn` to take a `DomEventTarget` instead of a `DomElement`. Then do:
```suggestion
  void disableContextMenu() => disableContextMenuOn(domWindow);
```

lib/web_ui/test/engine/view_embedder/embedding_strategy/custom_element_embedding_strategy_test.dart
```diff
@@ -121,4 +121,54 @@ void doTests() {
...
+      expect(event.defaultPrevented, isFalse);
+      target.dispatchEvent(event);
+      expect(event.defaultPrevented, isFalse);
+
+      // Disabling the context menu causes contextmenu events to be prevented.
+      strategy.disableContextMenu();
+      event = createDomEvent('Event', 'contextmenu');
+      expect(event.defaultPrevented, isFalse);
+      target.dispatchEvent(event);
+      expect(event.defaultPrevented, isTrue);
```
mdebbar, line 149: Another important test is to dispatch an event outside of `target` (either create a sibling element or just dispatch on `document.body`) and make sure that the event wasn't prevented.

mdebbar (approved):

lib/web_ui/test/engine/view_embedder/embedding_strategy/custom_element_embedding_strategy_test.dart
justinmc: Great idea, thanks!

PR #38686, opened 2023-01-06
Running macOS resize listener in background queue to avoid frozen animations when the main thread is blocked

jmatth:
This changes part of the macOS embedder to run its resize listener in a background queue rather than the main thread. This change is necessary because some AppKit APIs must be called in the main thread and block until they return, which currently causes the app to freeze during that time. Out of the box I've noticed this while interacting with the macOS menu bar, which causes brief but noticeable jank. A more serious problem is that this makes implementing plugins that use certain native functionality impractical, as they can cause anything from jank to freezing the entire app for extended periods of time. This is the problem I'm currently encountering: showing a native context menu via [`NSMenu.popUpContextMenu`][2] freezes the entire app until the menu is dismissed.

Although the tests pass and I haven't observed any issues running a real app with these changes, I assume the `isMainThread` assertions I took out were there for a reason and this can't be merged as-is. I'm opening this PR as a draft in the hopes that someone more familiar with the codebase knows what those assertions were for and can point me in the right direction so I can work on a proper implementation w/ test coverage. I'd especially appreciate any input from:

- @knopp, who recently merged a major refactor of the macOS rendering infrastructure
- @cbracken, who weighed in on my original bug report on this issue and was a reviewer on knopp's recent macOS refactor PR
- @dkwingsmt, who was also a reviewer on knopp's PR

Closes flutter/flutter#104638

## Pre-launch Checklist

- [x] I signed the [CLA].

[1]: https://developer.apple.com/documentation/dispatch/dispatchqos/1780708-userinteractive
[2]: https://developer.apple.com/documentation/appkit/nsmenu/1518170-popupcontextmenu

### DIFF
--- a/shell/platform/darwin/macos/framework/Source/FlutterSurfaceManager.mm
+++ b/shell/platform/darwin/macos/framework/Source/FlutterSurfaceManager.mm
@@ -78,8 +78,6 @@
 }
 
 - (void)commit:(NSArray<FlutterSurfacePresentInfo*>*)surfaces {
-  FML_DCHECK([NSThread isMainThread]);
-
   // Release all unused back buffer surfaces and replace them with front surfaces.
   [_backBufferCache replaceSurfaces:_frontSurfaces];
 
--- a/shell/platform/darwin/macos/framework/Source/FlutterThreadSynchronizer.mm
+++ b/shell/platform/darwin/macos/framework/Source/FlutterThreadSynchronizer.mm
@@ -25,8 +25,6 @@
 @implementation FlutterThreadSynchronizer
 
 - (void)drain {
-  FML_DCHECK([NSThread isMainThread]);
-
   [CATransaction begin];
   [CATransaction setDisableActions:YES];
   for (dispatch_block_t block : _scheduledBlocks) {
@@ -88,7 +86,7 @@
     if (_beginResizeWaiting) {
       _condBlockBeginResize.notify_all();
     } else {
-      dispatch_async(dispatch_get_main_queue(), ^{
+      dispatch_async(dispatch_get_global_queue(QOS_CLASS_USER_INTERACTIVE, 0), ^{
         std::unique_lock<std::mutex> lock(_mutex);
         [self drain];
       });


knopp:
This seems like a wrong solution to the problem. It is definitely not going to work. The surfaces must be switched on main thread. On top of it this will do platform view mutations which definitely need to be in main thread.

`[NSMenu popUpContextMenu]` doesn't freeze the main thread. It keeps pumping the event loop. The problem is that unlike the event loop, which is re-entrant, dispatch queue is not reentrant. So you can't call `[NSMenu.popUpContextMenu]` from dispatch queue callback directly, because it will block any subsequent dispatch queue callbacks until the menu closes. 

Have you tried calling the `[NSMenu popupContextMenu` from `[NSRunLoop performBlock:]`?

jmatth:
@knopp: Thank you for taking the time to reply. Moving my call to `popUp` into the Swift version of`[NSRunLoop performBlock:]` did indeed solve my problem.

If you don't mind, a follow up question: the [Flutter engine architecture wiki][1] claims that

> ...blocking the platform thread for inordinate amounts of time will not block the Flutter rendering pipeline...

Do you think the current state of the engine on macOS fails to fulfil that claim, and therefore there's a bug to be fixed here? Or is the current state fine and failing to use the main runloop is just PEBKAC on the part of plugin authors?

[1]: https://github.com/flutter/flutter/wiki/The-Engine-architecture#platform-task-runner

knopp:
> @knopp: Thank you for taking the time to reply. Moving my call to `popUp` into the Swift version of`[NSRunLoop performBlock:]` did indeed solve my problem.
> 
> If you don't mind, a follow up question: the [Flutter engine architecture wiki](https://github.com/flutter/flutter/wiki/The-Engine-architecture#platform-task-runner) claims that
> 
> > ...blocking the platform thread for inordinate amounts of time will not block the Flutter rendering pipeline...
> 
> Do you think the current state of the engine on macOS fails to fulfil that claim, and therefore there's a bug to be fixed here? Or is the current state fine and failing to use the main runloop is just PEBKAC on the part of plugin authors?

That statement doesn't seem to be correct. It depends on platform. On iOS for example it is only true unless there are platform views (in which case the platform thread is merged with raster thread). 

in any case, it is never a good idea to block platform thread. But that's not what [NSMenu popup] does anyway. But in your case it blocks the dispatch queue.

gspencergoog:
@jmatth There's been a lot of refactoring done on our rendering code in macOS in the last few months, so the rebase for this will be complicated. Are you planning on continuing with this patch?  If not, could you please close it to get it off of our triage radar?

(Desktop Triage)

jmatth:
Sorry, I let this get away from me. One followup question I had for @knopp: what would you think of refactoring the Flutter engine to process platform messages in the main run loop instead of the main queue? It seems like that would remove a footgun for plugin developers but I have no idea if it might break anything else. If you think it's a worthwhile change I can start working on it, if not I'll just drop the issue.

@gspencergoog I'll go ahead and close this either way. The implementation here doesn't make sense, and if there is work to be done I'll open a new PR. That said, it seems like the bit about `blocking the platform thread for inordinate amounts of time will not block the Flutter rendering pipeline` should be removed from the [Flutter architecture wiki][1]? Based on @knopp's responses above, that statement is false on macOS and only conditionally true on iOS.

[1]: https://github.com/flutter/flutter/wiki/The-Engine-architecture#platform-task-runner

PR #38737, opened 2023-01-10
Set the ivar to nil after release it in dealloc method. 

endless7:
As https://github.com/flutter/engine/pull/37666#discussion_r1024396618 mentioned, the correct pattern is the set the ivar to nil after release it in dealloc method.

Related issue: https://github.com/flutter/flutter/issues/118217

## Pre-launch Checklist

- [✓] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [✓] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [✓] I read and followed the [Flutter Style Guide], including [Features we expect every widget to implement].
- [✓] I signed the [CLA].
- [✓] I listed at least one issue that this PR fixes in the description above.
- [✓] I updated/added relevant documentation (doc comments with `///`).
- [✓] I added new tests to check the change I am making, or this PR is [test-exempt].
- [✓] All existing and new tests are passing.

### DIFF
--- a/shell/platform/darwin/ios/framework/Source/FlutterEngine.mm
+++ b/shell/platform/darwin/ios/framework/Source/FlutterEngine.mm
@@ -261,20 +261,27 @@ static constexpr int kNumProfilerSamplesPerSec = 5;
       }];
 
   [_labelPrefix release];
+  _labelPrefix = nil;
   [_initialRoute release];
+  _initialRoute = nil;
   [_pluginPublications release];
+  _pluginPublications = nil;
   [_registrars release];
+  _registrars = nil;
   _binaryMessenger.parent = nil;
   _textureRegistry.parent = nil;
   [_binaryMessenger release];
+  _binaryMessenger = nil;
   [_textureRegistry release];
   _textureRegistry = nil;
   [_isolateId release];
+  _isolateId = nil;
 
   NSNotificationCenter* center = [NSNotificationCenter defaultCenter];
   if (_flutterViewControllerWillDeallocObserver) {
     [center removeObserver:_flutterViewControllerWillDeallocObserver];
     [_flutterViewControllerWillDeallocObserver release];
+    _flutterViewControllerWillDeallocObserver = nil;
   }
   [center removeObserver:self];
 


endless7:
Hi, @cyanglaz. I have some difficulties writing the test. Is there any way to retain the ivar in the outside of `FlutterEngine`?

cyanglaz:
Yes there are hacks to do that in objective-C. https://medium.com/@AdamRoberts/accessing-private-properties-and-private-ivars-in-objective-c-from-another-class-bc59edcc36cb

jmagman:
@cyanglaz I'm surprised this is necessary per your comment in https://github.com/flutter/engine/pull/37666#discussion_r1024396618, it's being dealloc'd so none of these ivars should _never_ be used after this point.  Even in MRC code I've worked in before we didn't nil out ivars in dealloc...

hellohuanlin:
it's been a while since I wrote MRC code, but iirc you only need to nil out things like delegates to prevent dangling pointers: 

```
- (void) dealloc {
  [super dealloc]
  _tableView.delegate = nil 
  _tableView.dataSource = nil
}
```

You don't need to nil out `_tableView` here, because when object is deallocated, its `_tableView` pointer will be gone.

cyanglaz:
> @cyanglaz I'm surprised this is necessary per your comment in [#37666 (comment)](https://github.com/flutter/engine/pull/37666#discussion_r1024396618), it's being dealloc'd so none of these ivars should _ever_ be used after this point. Even in MRC code I've worked in before we didn't nil out ivars in dealloc...

It is to prevent tricky dangling pointers. Examples like:
 ```objc
- (void) dealloc {
  [_propertyA release];
  [self aCleanUpMethodDoesSomethingWithPropertyA];

  [super dealloc];
}
```

or a pointer pointing to this variable outside of the class.

Even for ivars that only defined in .mm files, there is no guarantee that they won't be exposed in a public getter later. When that happens, it is easy to forget to nil out the ivar in dealloc.

jmagman:
> ```objc
>  [self aCleanUpMethodDoesSomethingWithPropertyA];
> ```

That's the pattern that justifies "never call `self` in `init` or `dealloc`" which seems like the real bug, not that anything needs to get `nil`d.  Definitely when this is migrated to ARC we won't have to do this.

knopp:
> Even for ivars that only defined in .mm files, there is no guarantee that they won't be exposed in a public getter later. When that happens, it is easy for the author to forget to nil out the ivar in dealloc.

Not sure I follow. How would you access these ivars through public getter after calling dealloc?

Also as @jmagman pointed out, calling methods on self within dealloc goes against google objc guidelines. I had a brief look through `dealloc` methods within engine, and there are couple of instances of this, but it's usually just methods that clear/deregister single ivar.

knopp:
I would think that accessing ivar after releasing it is generally a bug, and setting it to `nil` just hides it so it won't be discoverable even with NSZombieEnabled.

chinmaygarde (changes requested):
This almost always indicates there is some other issue. If the instance owning the variable has been dealloc-ed, then someone managed to get a reference to that variable with incorrectly retaining it or reasoning about its lifecycle.

jmagman:
Let's close this, and stop using this `nil`ing out pattern going forward.  @endless7 Thanks for following up on the comment from the other review and sitting through this discussion 🙂

PR #38745, opened 2023-01-10
Merge MSAA alert functionality with UIA

yaakovschectman:
Merges the existing MSAA alert implementation with UIA making use of the existing `AXFragmentRoot` class rather than using separate implementations for each. Also use the above class for holding alert nodes, making `AccessibilityRootNode` and `AccessibilityAlert` redundant.

Part of https://github.com/flutter/flutter/issues/116220

## Pre-launch Checklist

- [ ] I signed the [CLA].


shell/platform/windows/flutter_windows_view.h
```diff
@@ -221,10 +224,7 @@ class FlutterWindowsView : public WindowBindingHandlerDelegate,
   virtual std::unique_ptr<TextInputPlugin> CreateTextInputPlugin(
       BinaryMessenger* messenger);
 
-  virtual void NotifyWinEventWrapper(DWORD event,
-                                     HWND hwnd,
-                                     LONG idObject,
-                                     LONG idChild);
+  virtual void NotifyWinEventWrapper(ui::AXPlatformNodeWin* node, ax::mojom::Event event);
```
yaakovschectman, line 227: `AXPlatformNodeWin` already has a helper method for dispatching accessibility events, so rather than dispatching them directly from the view, we use this helper method. This also means UIA events are handled alongside MSAA events.

third_party/accessibility/ax/platform/ax_platform_node_win.cc
```diff
@@ -5408,7 +5418,7 @@ AXPlatformNodeWin* AXPlatformNodeWin::GetTargetFromChildID(
 
   AXPlatformNodeBase* base =
       FromNativeViewAccessible(node->GetNativeViewAccessible());
-  if (base && !IsDescendant(base))
+  if (base && !base->IsDescendantOf(this))
```
yaakovschectman, line 5421: Allows using the descendant's override of `IsDescendantOf` to handle whichever delegate type it has.

shell/platform/windows/flutter_window.h
```diff
@@ -151,7 +151,10 @@ class FlutterWindow : public Window, public WindowBindingHandler {
   void SendInitialAccessibilityFeatures() override;
 
   // |WindowBindingHandler|
-  AccessibilityRootNode* GetAccessibilityRootNode() override;
+  void Alert(const std::wstring& text) override;
+
+  // |WindowBindingHandler|
+  ui::AXPlatformNodeWin* GetAlert() override;
```
yaakovschectman, line 157: The two separate methods make unit testing simpler.

shell/platform/common/alert_platform_node_delegate.h
```diff
@@ -0,0 +1,34 @@
...
+#define FLUTTER_SHELL_PLATFORM_COMMON_ALERT_PLATFORM_NODE_DELEGATE_H_
+
+#include "flutter/third_party/accessibility/ax/ax_node_data.h"
+#include "flutter/third_party/accessibility/ax/platform/ax_platform_node_delegate_base.h"
+
+namespace flutter {
+
+class AlertPlatformNodeDelegate : public ui::AXPlatformNodeDelegateBase {
+ public:
+  AlertPlatformNodeDelegate(ui::AXPlatformNodeDelegate* parent_delegate);
```
loic-sharma, line 15:
```suggestion
  explicit AlertPlatformNodeDelegate(ui::AXPlatformNodeDelegate* parent_delegate);
```

See: https://google.github.io/styleguide/cppguide.html#Implicit_Conversions

loic-sharma, line 15: Should this disallow copy and assign?

shell/platform/windows/flutter_window_unittests.cc
```diff
@@ -415,15 +416,13 @@ TEST(FlutterWindowTest, AlertNode) {
...
-              NotifyWinEventWrapper(EVENT_SYSTEM_ALERT, nullptr, OBJID_CLIENT,
-                                    AccessibilityRootNode::kAlertChildId))
-      .Times(1);
   std::wstring message = L"Test alert";
+  EXPECT_CALL(view, NotifyWinEventWrapper(_, ax::mojom::Event::kAlert))
+      .Times(1);
   view.AnnounceAlert(message);
-  IAccessible* alert = root_node->GetOrCreateAlert();
+
+  IAccessible* alert = view.AlertNode();  // root_node->GetOrCreateAlert();
```
loic-sharma, line 425:
```suggestion
  IAccessible* alert = view.AlertNode();
```

shell/platform/windows/flutter_platform_node_delegate_windows.cc
```diff
@@ -11,6 +11,7 @@
 #include "flutter/third_party/accessibility/ax/ax_clipping_behavior.h"
 #include "flutter/third_party/accessibility/ax/ax_coordinate_system.h"
 #include "flutter/third_party/accessibility/ax/platform/ax_fragment_root_win.h"
+#include "flutter/third_party/accessibility/ax/platform/ax_platform_node_win.h"
```
loic-sharma, line 14: Is this necessary? Feel free to close if yes

shell/platform/windows/window.cc
```diff
@@ -678,11 +677,15 @@ bool Window::GetHighContrastEnabled() {
...
-void Window::CreateAccessibilityRootNode() {
-  if (accessibility_root_) {
-    accessibility_root_->Release();
+void Window::CreateAlertNode() {
+  if (alert_delegate_) {
+    return;
   }
-  accessibility_root_ = AccessibilityRootNode::Create();
+  alert_delegate_ =
+      std::make_unique<AlertPlatformNodeDelegate>(ax_fragment_root_.get());
```
loic-sharma, line 685: What happens if the framework requests an alert before the app receives its first `WM_GETOBJECT` message? From my understanding, `ax_fragment_root_` will be `nullptr`, and the alernt node's parent will never be updated. Is that correct?

shell/platform/windows/flutter_windows_view.cc
```diff
@@ -661,28 +662,24 @@ FlutterWindowsEngine* FlutterWindowsView::GetEngine() {
...
-void FlutterWindowsView::NotifyWinEventWrapper(DWORD event,
-                                               HWND hwnd,
-                                               LONG idObject,
-                                               LONG idChild) {
-  if (hwnd) {
-    NotifyWinEvent(EVENT_SYSTEM_ALERT, hwnd, OBJID_CLIENT,
-                   AccessibilityRootNode::kAlertChildId);
+  binding_handler_->Alert(text);
+  ui::AXPlatformNodeWin* alert_node = binding_handler_->GetAlert();
+  NotifyWinEventWrapper(alert_node, ax::mojom::Event::kAlert);
```
loic-sharma, line 667: Why doesn't `WindowBindingHandler::Alert` call `AXPlatformNodeWin::NotifyAccessibilityEvent`?

shell/platform/windows/window.cc
yaakovschectman: In that case nothing would happen, but if there has been no getobject message, I believe that would mean there is no screen reader open, so an alert would do nothing, is that correct?

shell/platform/windows/flutter_windows_view.cc
yaakovschectman: For mocking in the unit tests.

loic-sharma: Taking a step back, the value of the `FlutterWindowsView` / `FlutterWindow` split is to enable mock testing by decoupling Flutter from native win32 APIs. We can't mock win32 APIs, so instead we wrap them in `FlutterWindow` and mock that.

Ideally `FlutterWindow` is as small as possible. Any code that's unit testable should be moved out of `FlutterWindow` and into `FlutterWindowsView`.

Currently, the alerting logic is split across `FlutterWindow` and `FlutterWindowsView`. Do you think we could refactor `FlutterWindow` to push up more alerting logic to `FlutterWindowsView`?

shell/platform/windows/window.cc
```diff
@@ -204,35 +204,39 @@ LRESULT Window::OnGetObject(UINT const message,
...
+      }
 #endif  // FLUTTER_ENGINE_USE_UIA
-  } else if (is_msaa_request && root_view) {
-    // Create the accessibility root if it does not already exist.
-    if (!accessibility_root_) {
-      CreateAccessibilityRootNode();
+    } else if (is_msaa_request) {
+      // Create the accessibility root if it does not already exist.
+      // Return the IAccessible for the root view.
+      // Microsoft::WRL::ComPtr<IAccessible> root(root_view);
```
loic-sharma, line 234:
```suggestion
```

third_party/accessibility/ax/platform/ax_fragment_root_win.cc
```diff
@@ -289,7 +289,7 @@ class AXFragmentRootMapWin {
 
 AXFragmentRootWin::AXFragmentRootWin(gfx::AcceleratedWidget widget,
                                      AXFragmentRootDelegateWin* delegate)
-    : widget_(widget), delegate_(delegate) {
+    : widget_(widget), delegate_(delegate), alert_node_(nullptr) {
```
loic-sharma, line 292: What's the benefit of lazily creating the alert node? Could we create an alert node every time we create the root?

yaakovschectman:
With the logic of dispatching the alert message moved to the windows view, I will need to change the platform message unit test a bit more.

shell/platform/windows/window.cc
loic-sharma: To clarify, here's the scenario I'm concerned about:

1. The Windows embedder receives an alert message from the framework
2. The user opens a screen reader & the Windows embedder receives the first `WM_GETOBJECT` message
3. The Windows embedder receives a second alert message from the framework

In this scenario, I expect that the screen reader is notified of an alert for step 3. However, I suspect it won't. From my understanding, step 1 creates an alert delegate that will always have a `nullptr` parent delegate. Step 2 creates the `ax_fragment_root_`, but, the alert delegate is never notified of this parent. Finally for step 3, both `AlertPlatformNodeDelegate::GetTargetForNativeAccessibilityEvent` and `AlertPlatformNodeDelegate::GetParent` will return `nullptr` when called by `AXPlatformNodeWin::NotifyAccessibilityEvent`. Please let me know if I'm missing something!

third_party/accessibility/ax/platform/ax_fragment_root_win.cc
yaakovschectman: The alert node, delegate, and fragment root refer to each other, so one pointer or more will be set after creation.

third_party/accessibility/ax/platform/ax_fragment_root_win.h
```diff
@@ -55,6 +56,11 @@ class AX_EXPORT AXFragmentRootWin : public ui::AXPlatformNodeDelegateBase {
   // If a child node is available, return its delegate.
   AXPlatformNodeDelegate* GetChildNodeDelegate() const;
 
+  // AXPlatformNodeDelegate override.
+  gfx::AcceleratedWidget GetTargetForNativeAccessibilityEvent() override;
+
+  void SetAlertNode(AXPlatformNodeWin* alert_node);
```
cbracken, line 62: Add a doc comment specifically describing `alert_node_`'s:
* Ownership model - since this is a raw/weak pointer, who is the owner of the node (transitively, the AXTree I think)
* Any restrictions on which node(s) should be passed.

third_party/accessibility/ax/platform/ax_platform_node_win.cc
```diff
@@ -5701,4 +5711,28 @@ AXPlatformNodeWin* AXPlatformNodeWin::GetFirstTextOnlyDescendant() {
...
 }
 
+bool AXPlatformNodeWin::IsDescendantOf(AXPlatformNode* ancestor) const {
+  if (AXPlatformNodeBase::IsDescendantOf(ancestor)) {
+    return true;
+  }
+
+  if (!ancestor) {
+    return false;
+  }
```
cbracken, line 5721: Does the base `IsDescendantOf` method ever return true for nullptr? If not, then I'd move this above that check since it's cheaper than a function call.

third_party/accessibility/ax/platform/ax_platform_node_win.cc
```diff
@@ -5701,4 +5711,28 @@ AXPlatformNodeWin* AXPlatformNodeWin::GetFirstTextOnlyDescendant() {
...
+
+  // Test if the ancestor is an IRawElementProviderFragmentRoot and if it
+  // matches this node's root fragment.
+  IRawElementProviderFragmentRoot* root;
+  if (SUCCEEDED(
+          const_cast<AXPlatformNodeWin*>(this)->get_FragmentRoot(&root))) {
+    AXPlatformNodeWin* root_win;
+    if (SUCCEEDED(root->QueryInterface(__uuidof(AXPlatformNodeWin),
+                                       (void**)&root_win))) {
```
cbracken, line 5730: Use a reinterpret_cast here.

third_party/accessibility/ax/platform/ax_fragment_root_win.cc
loic-sharma: Right... but the creation of the fragment root node and alert node is decoupled: a fragment root is created only when I attach a screen reader, and, an alert node is created only when the framework sends an alert. Why is that? Why doesn't creating a fragment root also create the alert node? Do we need to create an alert node if there's no screen reader?

If we do need to create an alert node even without a screen reader, perhaps that should also create a fragment root? That would address concerns like: https://github.com/flutter/engine/pull/38745#discussion_r1066448031

shell/platform/windows/flutter_windows_view.cc
```diff
@@ -661,28 +662,28 @@ FlutterWindowsEngine* FlutterWindowsView::GetEngine() {
...
+}
+
+void FlutterWindowsView::NotifyWinEventWrapper(ui::AXPlatformNodeWin* node,
+                                               ax::mojom::Event event) {
+  if (node) {
+    node->NotifyAccessibilityEvent(event);
   }
 }
```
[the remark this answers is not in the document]
yaakovschectman: The problem is that `AXPlatformNodeWin::NotifyAccessibilityEvent` is not const, nor are methods it calls.

third_party/accessibility/ax/platform/ax_fragment_root_win.cc
```diff
@@ -425,4 +427,18 @@ int AXFragmentRootWin::GetIndexInParentOfChild() const {
   return 0;
 }
 
+void AXFragmentRootWin::SetAlertNode(AXPlatformNodeWin* alert_node) {
+  alert_node_ = alert_node;
```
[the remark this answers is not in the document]
yaakovschectman: I do not think we have a use case for this, but setting to nullptr would remove the alert node from the root. The only time it is accessed by `AXFragmentRootWin` is in `ChildAtIndex`, where it is checked for nullity.

shell/platform/windows/flutter_windows_view.cc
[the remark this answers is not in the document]
cbracken: Ah interesting. I'm guessing that deeper down then, we are actually mutating something (or we have a cascade of methods where we accidentally missed out on marking em const). sgtm.

third_party/accessibility/ax/platform/ax_fragment_root_win.cc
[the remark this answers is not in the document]
cbracken: If it would currently be a developer error to set it null, consider adding an `assert(alert_node != nullptr);` so we catch it in dev mode. If doing so is harmless, but unexpected, maybe just covering that in the doc comment is sufficient.

shell/platform/common/alert_platform_node_delegate.h
```diff
@@ -0,0 +1,41 @@
...
+ public:
+  explicit AlertPlatformNodeDelegate(
+      ui::AXPlatformNodeDelegate* parent_delegate);
+  ~AlertPlatformNodeDelegate();
+
+  AlertPlatformNodeDelegate(const AlertPlatformNodeDelegate& other) = delete;
+  AlertPlatformNodeDelegate operator=(const AlertPlatformNodeDelegate& other) =
+      delete;
+
+  void SetText(const std::u16string& text);
```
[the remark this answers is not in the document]
loic-sharma: I'm also leaning towards keeping this a string reference, but, the string reference is also copied when `AXNodeData` stores it as an attribute.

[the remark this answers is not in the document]
cbracken: > The AXNodeData methods that this method calls take string& parameters. Am I mistaken that constructing a string from a string_view effectively copies it?

Yes - to benefit from this, you'd need any methods this gets passed to to also take a view rather than a string ref. If the current implementation is doing so, then leave as-is; at some point we can do a sweep of the codebase to clean these up across the board.

loic-sharma:
@yaakovschectman Please tag me when this is ready for another review 😄

shell/platform/windows/flutter_windows_view.cc
```diff
@@ -661,28 +662,29 @@ FlutterWindowsEngine* FlutterWindowsView::GetEngine() {
...
 
 ui::AXFragmentRootDelegateWin* FlutterWindowsView::GetAxFragmentRootDelegate() {
+  auto bridge = engine_->accessibility_bridge().lock().get();
```
loic-sharma, line 682: Can this be removed?

shell/platform/common/alert_platform_node_delegate.h
```diff
@@ -0,0 +1,41 @@
...
+
+#include "flutter/third_party/accessibility/ax/ax_node_data.h"
+#include "flutter/third_party/accessibility/ax/platform/ax_platform_node_delegate_base.h"
+
+namespace flutter {
+
+class AlertPlatformNodeDelegate : public ui::AXPlatformNodeDelegateBase {
+ public:
+  explicit AlertPlatformNodeDelegate(
+      ui::AXPlatformNodeDelegate* parent_delegate);
```
loic-sharma, line 16: The parent delegate should never be `nullptr`, right? Consider making this a `const ui::AXPlatformNodeDelegate&` instead and updating the `AlertPlatformMessage` test to pass a mocked object instead.

shell/platform/common/alert_platform_node_delegate.h
```diff
@@ -0,0 +1,41 @@
...
+
+#ifndef FLUTTER_SHELL_PLATFORM_COMMON_ALERT_PLATFORM_NODE_DELEGATE_H_
+#define FLUTTER_SHELL_PLATFORM_COMMON_ALERT_PLATFORM_NODE_DELEGATE_H_
+
+#include "flutter/third_party/accessibility/ax/ax_node_data.h"
+#include "flutter/third_party/accessibility/ax/platform/ax_platform_node_delegate_base.h"
+
+namespace flutter {
+
+class AlertPlatformNodeDelegate : public ui::AXPlatformNodeDelegateBase {
```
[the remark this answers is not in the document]
loic-sharma: Please do a pass, there's several other fields that are also missing comments

shell/platform/common/alert_platform_node_delegate.h
yaakovschectman: I do not think we can make this `const` without a more thorough sweeping change to `AXPlatformNodeDelegate` and classes that touch it. `AlertPlatformNodeDelegate` needs to call several methods on its `parent_delegate_` that are not `const`.

loic-sharma: Ah ok. My meta point is to disallow `nullptr`, could we make it a non-constant reference?

yaakovschectman: We should be able to.

third_party/accessibility/ax/platform/ax_platform_node_win.h
```diff
@@ -473,6 +473,9 @@ class AX_EXPORT __declspec(uuid("26f5641a-246d-457b-a96d-07f3fae6acf2"))
   static std::optional<PROPERTYID> MojoEventToUIAProperty(
       ax::mojom::Event event);
 
+  // |AXPlatformNodeBase|
+  bool IsDescendantOf(AXPlatformNode* acnestor) const override;
```
loic-sharma, line 477:
```suggestion
  bool IsDescendantOf(AXPlatformNode* ancestor) const override;
```

loic-sharma (approved):
LGTM, nice work!

cbracken (approved):
Nice!

PR #38783, opened 2023-01-11
delete include of private GrMtlTypes header

flar:
Fixes https://github.com/flutter/flutter/issues/118330

Removes unnecessary includes of what is now a private Skia header file.

No tests because this is just a source compatibility change and is covered by existing unit tests.


jonahwilliams (approved):
RSLGTM

chinmaygarde (approved):
LGTM. For context on why this was included, this header included a macro that gated the availability of Metal backend related code behind a runtime check for iOS 10. We dropped support for older versions of iOS and hence the macro as well (which may have already moved).

PR #38786, opened 2023-01-11
[web] Update felt to use generated JS runtime for Dart2Wasm.

joshualitt:


joshualitt:
@eyebrowsoffire ptal. I tested this locally and it seems to work.

joshualitt:
> Nice, didn't even have to change the compile command.

Yea, Dart2Wasm defaults to generating the `.mjs` file alongside the `.wasm` file with just a suffix change. It's nice that it seems to get hosted automatically.  It looks like we don't currently support a way to rename the runtime. It would be trivial to add a flag if we need that functionality, but in this case it doesn't seem worthwhile.

Hixie:
test-exempt: is the test system

PR #38797, opened 2023-01-12
Adjust to changes to CanvasKit getRectsForRange API.

eyebrowsoffire:
Adjust to the changes made in this PR https://skia-review.googlesource.com/c/skia/+/628979

### DIFF
--- a/lib/web_ui/lib/src/engine/canvaskit/canvaskit_api.dart
+++ b/lib/web_ui/lib/src/engine/canvaskit/canvaskit_api.dart
@@ -2170,6 +2170,16 @@ extension SkLineMetricsExtension on SkLineMetrics {
 @JS()
 @anonymous
 @staticInterop
+class SkRectWithDirection {}
+
+extension SkRectWithDirectionExtension on SkRectWithDirection {
+  external Float32List rect;
+  external double direction;
+}
+
+@JS()
+@anonymous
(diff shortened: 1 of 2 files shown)


lib/web_ui/lib/src/engine/canvaskit/canvaskit_api.dart
```diff
@@ -2167,6 +2167,16 @@ extension SkLineMetricsExtension on SkLineMetrics {
   external double get lineNumber;
 }
 
+@JS()
+@anonymous
+@staticInterop
+class SkRectWithDirection {}
+
+extension SkRectWithDirectionExtension on SkRectWithDirection {
+  external Float32List rect;
```
yjbanov, line 2176: Does the meaning of elements change based on the value of `direction`? E.g. is is always LTRB, or can it also be RTLB? Might be worth leaving a comment explaining it.

yjbanov (approved):

lib/web_ui/lib/src/engine/canvaskit/canvaskit_api.dart
kjlubick: Rectangles are always LTRB

eyebrowsoffire:
This was included in a manual skia roll and doesn't need to actually merge.

PR #38822, opened 2023-01-12
Fix iOS safari keyboard issue when semantics is enabled

nbayati:
Fixes Flutter/Flutter#110284

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/lib/web_ui/lib/src/engine/semantics/text_field.dart
+++ b/lib/web_ui/lib/src/engine/semantics/text_field.dart
@@ -384,7 +384,7 @@ class TextField extends RoleManager {
     }
 
     _initializeEditableElement();
-    activeEditableElement.style.transform = 'translate(-999px,-999px)';
+    activeEditableElement.style.transform = 'translate(${offScreenOffset}px, ${offScreenOffset}px)';
     _positionInputElementTimer?.cancel();
     _positionInputElementTimer = Timer(_delayBeforePlacement, () {
       editableElement?.style.transform = '';
--- a/lib/web_ui/lib/src/engine/semantics/text_field.dart
+++ b/lib/web_ui/lib/src/engine/semantics/text_field.dart
@@ -244,7 +244,10 @@ class TextField extends RoleManager {
   Timer? _positionInputElementTimer;
   static const Duration _delayBeforePlacement = Duration(milliseconds: 100);
 
-  void _createEditableElement() {
+  void _initializeEditableElement() {
+    assert(editableElement == null,
+        'Editable element has already been initialized');
+
     editableElement = semanticsObject.hasFlag(ui.SemanticsFlag.isMultiline)
         ? createDomHTMLTextAreaElement()
         : createDomHTMLInputElement();
@@ -295,7 +298,7 @@ class TextField extends RoleManager {
   /// When in browser gesture mode, the focus is forwarded to the framework as
   /// a tap to initialize editing.
   void _initializeForBlink() {
-    _createEditableElement();
+    _initializeEditableElement();
     activeEditableElement.addEventListener('focus',
         allowInterop((DomEvent event) {
           if (semanticsObject.owner.gestureMode != GestureMode.browserGestures) {
@@ -380,13 +383,17 @@ class TextField extends RoleManager {
       return;
     }
 
-    _createEditableElement();
+    _initializeEditableElement();
     activeEditableElement.style.transform = 'translate(-999px,-999px)';
     _positionInputElementTimer?.cancel();
     _positionInputElementTimer = Timer(_delayBeforePlacement, () {
       editableElement?.style.transform = '';
+      _positionInputElementTimer = null;
     });
 
+    // Can not have both activeEditableElement and semanticsObject.element
+    // represent the same text field. It will confuse VoiceOver, so `role` needs to
+    // be assigned and removed, based on whether or not editableElement exists.
     activeEditableElement.focus();
     semanticsObject.element.removeAttribute('role');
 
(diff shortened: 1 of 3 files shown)


lib/web_ui/lib/src/engine/semantics/text_field.dart
```diff
@@ -341,75 +375,103 @@ class TextField extends RoleManager {
...
+  void _invokeIosWorkaround() {
+    if (editableElement != null) {
+      return;
+    }
+
+    _createEditableElement();
+    activeEditableElement.style.transform = 'translate(-999px,-999px)';
+    _positionInputElementTimer?.cancel();
+    _positionInputElementTimer = Timer(_delayBeforePlacement, () {
+      editableElement?.style.transform = '';
```
yjbanov, line 405: We might want to null out `_positionInputElementTimer` when it fires to avoid having a stale timer object lurking in the system.

lib/web_ui/lib/src/engine/semantics/text_field.dart
```diff
@@ -341,75 +375,103 @@ class TextField extends RoleManager {
...
+    });
+
+    activeEditableElement.focus();
+    semanticsObject.element.removeAttribute('role');
```
yjbanov, line 409: Is this to avoid having two elements with the "textfield" role representing the same text field? If so, might be worth leaving a comment explaining it for a future reader.

lib/web_ui/lib/src/engine/semantics/text_field.dart
```diff
@@ -341,75 +375,103 @@ class TextField extends RoleManager {
...
-        // Unlike `editableElement.focus()` we don't need to schedule `blur`
-        // post-update because `document.activeElement` implies that the
-        // element is already attached to the DOM. If it's not, it can't
-        // possibly be focused and therefore there's no need to blur.
-        editableElement.blur();
-      }
-      _hasFocused = false;
+  @override
+  void update() {
+    if (editableElement != null) {
```
yjbanov, line 427: I think all the `editableElement != null` checks can be combined into one top-level check, so that we don't need to check it again multiple times.

lib/web_ui/lib/src/engine/semantics/text_field.dart
```diff
@@ -341,75 +375,103 @@ class TextField extends RoleManager {
...
+      element.removeAttribute('aria-label');
     }
   }
 
   @override
   void dispose() {
-    editableElement.remove();
+    _positionInputElementTimer?.cancel();
+    _positionInputElementTimer = null;
+    activeEditableElement.remove();
```
yjbanov, line 474: Are we guaranteed to have an `activeEditableElement`? If not, this can throw. I can imagine a text field to never be focused and disposed. Also, we might want to have a test for disposing both focused and unfocused text fields.

lib/web_ui/test/engine/semantics/text_field_test.dart
```diff
@@ -434,9 +365,332 @@ void testMain() {
...
+
+      final DomElement textField = appHostNode
+          .querySelector('flt-semantics[role="textbox"]')!;
+
+      simulateTap(textField);
+      expect(await logger.idLog.first, 0);
+      expect(await logger.actionLog.first, ui.SemanticsAction.tap);
+    });
+
+      test('Syncs editing state from framework', () {
```
yjbanov, line 412: There's an extra indent compared to sigling tests.

lib/web_ui/test/engine/semantics/text_field_test.dart
```diff
@@ -434,9 +365,332 @@ void testMain() {
         expect(appHostNode.activeElement, tester.getTextField(2).editableElement);
         expect(strategy.domElement, tester.getTextField(2).editableElement);
       }
+    });
+  }, skip: isIosSafari);
+
+  group('$SemanticsTextEditingStrategy in iOS', () {
```
yjbanov, line 371: Since we're only emulating iOS here, we probably want to limit the number of browsers we run this on to avoid browser quirks. I think we should pick one browser that's closest to iOS Safari, which I think is macOS Safari.

lib/web_ui/lib/src/engine/semantics/text_field.dart
nbayati: Good point! Done!

lib/web_ui/test/engine/semantics/text_field_test.dart
nbayati: Done!

lib/web_ui/lib/src/engine/semantics/text_field.dart
nbayati: Good point! Done!

lib/web_ui/lib/src/engine/semantics/text_field.dart
```diff
@@ -205,35 +208,59 @@ class SemanticsTextEditingStrategy extends DefaultTextEditingStrategy {
...
+  ///
+  /// In iOS, the virtual keyboard shifts the screen up if the focused input
+  /// element is under the keyboard or very close to the keyboard. Before the
+  /// focus is called we are positioning it offscreen. The location of the input
+  /// in iOS is set to correct place, 100ms after focus. We use this timer for
+  /// timing this delay.
+  Timer? _positionInputElementTimer;
+  static const Duration _delayBeforePlacement = Duration(milliseconds: 100);
+
+  void _createEditableElement() {
```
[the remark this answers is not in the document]
nbayati: Done!

lib/web_ui/lib/src/engine/semantics/text_field.dart
nbayati: Yes, VoiceOver starts acting weird if we don't remove the `role`.
Sure! I added some comments.

lib/web_ui/lib/src/engine/semantics/text_field.dart
nbayati: Yep, you're right.

lib/web_ui/test/engine/semantics/text_field_test.dart
nbayati: Done!

lib/web_ui/lib/src/engine/semantics/text_field.dart
```diff
@@ -341,75 +378,109 @@ class TextField extends RoleManager {
...
+    SemanticsTextEditingStrategy.instance.setEditingState(editingState);
+  }
+
+  void _invokeIosWorkaround() {
+    if (editableElement != null) {
+      return;
+    }
+
+    _initializeEditableElement();
+    activeEditableElement.style.transform = 'translate(-999px,-999px)';
```
yjbanov, line 405: With 8k screen resolutions getting to 7680x4320 these days, I wonder if 999 is a sufficient  #offset. Would using a bigger number, e.g. 20000, work?

yjbanov (approved):
lgtm

lib/web_ui/lib/src/engine/semantics/text_field.dart
nbayati: You're right. I've increased it to 9999 to match what we have in the non semantic iOS strategy.

PR #38832, opened 2023-01-13
[web] remove runtime effect check

jonahwilliams:
Fixes https://github.com/flutter/flutter/issues/114260


yjbanov (approved):
lgtm

feinstein:
The docs on `InkSparkle` mention it's not supported on CanvasKit and we should look [this issue](https://github.com/flutter/flutter/issues/85238). But the issue is closed, so shouldn't the docs be updated?

jonahwilliams:
yes, this is all done and should work. Thanks for the heads up on the docs, we missed that one.

feinstein:
My pleasure.

PR #38855, opened 2023-01-13
[macOS] Move A11yBridge to FVC

dkwingsmt:
This PR moves `AccessibilityBridgeMac` from `FlutterEngine` to `FlutterViewController`.

Currently, whether `FlutterEngine` has a non-null `AccessibilityBridge` only depends on the value of `semanticsEnabled`. The `_bridge` is initialized the moment `semanticsEnabled` turns true, even if `viewController` is null. Similarly, setting the view controller to null does not affect `_bridge` either. This is problematic because conceptually an `AccessibilityBridge` manages the a11y information for one view. Accessibility information does not make much sense without a view.

By moving the `AccessibilityBridge` into `FlutterViewController`, the lifecycles of these 2 classes are tied. The bridge exists only when the view controller exists _and_ semantics has been enabled. This is important because accessibility information should be view-specific.

| Before | After |
|----|----|
|`FlutterEngine#accessibilityBridge`|`FlutterViewController#accessibilityBridge`|
|`FlutterEngine#updateSemantics:`|`FlutterViewController#updateSemantics:`|
|`FlutterEngine#createAccessibilityBridge:viewController:`|`FlutterViewController#createAccessibilityBridgeWithEngine:`|
|-|`FlutterViewController#notifySemanticsEnabledChanged:`|

## Pre-launch Checklist

- [ ] I signed the [CLA].

### DIFF
--- a/shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
+++ b/shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
@@ -194,9 +194,6 @@ static void OnPlatformMessage(const FlutterPlatformMessage* message, FlutterEngi
   // The embedding-API-level engine object.
   FLUTTER_API_SYMBOL(FlutterEngine) _engine;
 
-  // The private member for accessibility.
-  std::shared_ptr<flutter::AccessibilityBridgeMac> _bridge;
-
   // The project being run by this engine.
   FlutterDartProject* _project;
 
@@ -315,7 +312,7 @@ static void OnPlatformMessage(const FlutterPlatformMessage* message, FlutterEngi
   flutterArguments.update_semantics_callback = [](const FlutterSemanticsUpdate* update,
                                                   void* user_data) {
     FlutterEngine* engine = (__bridge FlutterEngine*)user_data;
-    [engine updateSemantics:update];
+    [engine.viewController updateSemantics:update];
   };
   flutterArguments.custom_dart_entrypoint = entrypoint.UTF8String;
   flutterArguments.shutdown_dart_vm_when_done = true;
@@ -543,10 +540,6 @@ static void OnPlatformMessage(const FlutterPlatformMessage* message, FlutterEngi
   return _embedderAPI;
 }
 
-- (std::weak_ptr<flutter::AccessibilityBridgeMac>)accessibilityBridge {
-  return _bridge;
-}
-
 - (nonnull NSString*)executableName {
   return [[[NSProcessInfo processInfo] arguments] firstObject] ?: @"Flutter";
 }
@@ -586,22 +579,8 @@ static void OnPlatformMessage(const FlutterPlatformMessage* message, FlutterEngi
     return;
   }
   _semanticsEnabled = enabled;
-  // Remove the accessibility children from flutter view before reseting the bridge.
-  if (!_semanticsEnabled && self.viewController.viewLoaded) {
-    self.viewController.flutterView.accessibilityChildren = nil;
-  }
-  if (!_semanticsEnabled && _bridge) {
-    _bridge.reset();
-  } else if (_semanticsEnabled && !_bridge) {
-    _bridge = [self createAccessibilityBridge:self viewController:self.viewController];
-  }
+  [_viewController notifySemanticsEnabledChanged];
   _embedderAPI.UpdateSemanticsEnabled(_engine, _semanticsEnabled);
-}
-
-- (std::shared_ptr<flutter::AccessibilityBridgeMac>)
-    createAccessibilityBridge:(nonnull FlutterEngine*)engine
-               viewController:(nonnull FlutterViewController*)viewController {
-  return std::make_shared<flutter::AccessibilityBridgeMac>(engine, _viewController);
 }
 
 - (void)dispatchSemanticsAction:(FlutterSemanticsAction)action
@@ -937,36 +916,6 @@ static void OnPlatformMessage(const FlutterPlatformMessage* message, FlutterEngi
 
 - (BOOL)unregisterTextureWithID:(int64_t)textureID {
   return _embedderAPI.UnregisterExternalTexture(_engine, textureID) == kSuccess;
-}
-
-- (void)updateSemantics:(const FlutterSemanticsUpdate*)update {
-  NSAssert(_bridge, @"The accessibility bridge must be initialized.");
-  for (size_t i = 0; i < update->nodes_count; i++) {
-    const FlutterSemanticsNode* node = &update->nodes[i];
-    _bridge->AddFlutterSemanticsNodeUpdate(node);
-  }
-
-  for (size_t i = 0; i < update->custom_actions_count; i++) {
-    const FlutterSemanticsCustomAction* action = &update->custom_actions[i];
-    _bridge->AddFlutterSemanticsCustomActionUpdate(action);
-  }
-
-  _bridge->CommitUpdates();
-
-  // Accessibility tree can only be used when the view is loaded.
-  if (!self.viewController.viewLoaded) {
-    return;
-  }
-  // Attaches the accessibility root to the flutter view.
-  auto root = _bridge->GetFlutterPlatformNodeDelegateFromID(0).lock();
-  if (root) {
-    if ([self.viewController.flutterView.accessibilityChildren count] == 0) {
-      NSAccessibilityElement* native_root = root->GetNativeViewAccessible();
-      self.viewController.flutterView.accessibilityChildren = @[ native_root ];
-    }
-  } else {
-    self.viewController.flutterView.accessibilityChildren = nil;
-  }
 }
 
 #pragma mark - Task runner integration
--- a/shell/platform/darwin/macos/framework/Source/FlutterEngineTest.mm
+++ b/shell/platform/darwin/macos/framework/Source/FlutterEngineTest.mm
@@ -348,6 +348,31 @@ TEST_F(FlutterEngineTest, CanToggleAccessibilityWhenHeadless) {
   EXPECT_EQ(engine.viewController, nil);
 }
 
+TEST_F(FlutterEngineTest, ProducesAccessibilityTreeWhenAddingViews) {
+  FlutterEngine* engine = GetFlutterEngine();
+  EXPECT_TRUE([engine runWithEntrypoint:@"main"]);
+
+  // Enable the semantics without attaching a view controller.
+  bool enabled_called = false;
+  engine.embedderAPI.UpdateSemanticsEnabled =
+      MOCK_ENGINE_PROC(UpdateSemanticsEnabled, ([&enabled_called](auto engine, bool enabled) {
+                         enabled_called = enabled;
+                         return kSuccess;
+                       }));
+  engine.semanticsEnabled = YES;
+  EXPECT_TRUE(enabled_called);
+
+  EXPECT_EQ(engine.viewController, nil);
+
+  // Assign the view controller after enabling semantics
+  FlutterViewController* viewController = [[FlutterViewController alloc] initWithEngine:engine
+                                                                                nibName:nil
+                                                                                 bundle:nil];
+  engine.viewController = viewController;
+
+  EXPECT_NE(viewController.accessibilityBridge.lock(), nullptr);
+}
+
 TEST_F(FlutterEngineTest, NativeCallbacks) {
   fml::AutoResetWaitableEvent latch;
   bool latch_called = false;
--- a/shell/platform/darwin/macos/framework/Source/FlutterTextInputPluginTest.mm
+++ b/shell/platform/darwin/macos/framework/Source/FlutterTextInputPluginTest.mm
@@ -1414,14 +1414,14 @@ TEST(FlutterTextInputPluginTest, CanWorkWithFlutterTextField) {
 
   engine.semanticsEnabled = YES;
 
-  auto bridge = engine.accessibilityBridge.lock();
+  auto bridge = viewController.accessibilityBridge.lock();
   FlutterPlatformNodeDelegateMac delegate(bridge, viewController);
   ui::AXTree tree;
   ui::AXNode ax_node(&tree, nullptr, 0, 0);
   ui::AXNodeData node_data;
   node_data.SetValue("initial text");
   ax_node.SetData(node_data);
-  delegate.Init(engine.accessibilityBridge, &ax_node);
+  delegate.Init(viewController.accessibilityBridge, &ax_node);
   {
     FlutterTextPlatformNode text_platform_node(&delegate, viewController);
 
@@ -1480,14 +1480,14 @@ TEST(FlutterTextInputPluginTest, CanNotBecomeResponderIfNoViewController) {
 
   engine.semanticsEnabled = YES;
 
-  auto bridge = engine.accessibilityBridge.lock();
+  auto bridge = viewController.accessibilityBridge.lock();
   FlutterPlatformNodeDelegateMac delegate(bridge, viewController);
   ui::AXTree tree;
   ui::AXNode ax_node(&tree, nullptr, 0, 0);
   ui::AXNodeData node_data;
   node_data.SetValue("initial text");
   ax_node.SetData(node_data);
-  delegate.Init(engine.accessibilityBridge, &ax_node);
+  delegate.Init(viewController.accessibilityBridge, &ax_node);
   FlutterTextPlatformNode text_platform_node(&delegate, viewController);
 
   FlutterTextField* textField = text_platform_node.GetNativeViewAccessible();
--- a/shell/platform/darwin/macos/framework/Source/FlutterViewController.mm
+++ b/shell/platform/darwin/macos/framework/Source/FlutterViewController.mm
@@ -287,11 +287,14 @@ void OnKeyboardLayoutChanged(CFNotificationCenterRef center,
   // The project to run in this controller's engine.
   FlutterDartProject* _project;
 
+  std::shared_ptr<flutter::AccessibilityBridgeMac> _bridge;
+
   uint64_t _id;
 }
 
 @dynamic id;
 @dynamic view;
+@dynamic accessibilityBridge;
 
 /**
  * Performs initialization that's common between the different init paths.
@@ -314,6 +317,7 @@ static void CommonInit(FlutterViewController* controller, FlutterEngine* engine)
   controller->_mouseTrackingMode = FlutterMouseTrackingModeInKeyWindow;
   controller->_textInputPlugin = [[FlutterTextInputPlugin alloc] initWithViewController:controller];
   [controller initializeKeyboard];
+  [controller notifySemanticsEnabledChanged];
   // macOS fires this message when changing IMEs.
   CFNotificationCenterRef cfCenter = CFNotificationCenterGetDistributedCenter();
   __weak FlutterViewController* weakSelf = controller;
@@ -437,6 +441,26 @@ static void CommonInit(FlutterViewController* controller, FlutterEngine* engine)
   [self initializeKeyboard];
 }
 
+- (void)notifySemanticsEnabledChanged {
+  BOOL mySemanticsEnabled = !!_bridge;
+  BOOL newSemanticsEnabled = _engine.semanticsEnabled;
+  if (newSemanticsEnabled == mySemanticsEnabled) {
+    return;
+  }
+  if (newSemanticsEnabled) {
+    _bridge = [self createAccessibilityBridgeWithEngine:_engine];
+  } else {
+    // Remove the accessibility children from flutter view before resetting the bridge.
+    _flutterView.accessibilityChildren = nil;
+    _bridge.reset();
+  }
+  NSAssert(newSemanticsEnabled == !!_bridge, @"Failed to update semantics for the view.");
+}
+
+- (std::weak_ptr<flutter::AccessibilityBridgeMac>)accessibilityBridge {
+  return _bridge;
+}
+
 - (void)attachToEngine:(nonnull FlutterEngine*)engine withId:(uint64_t)viewId {
   NSAssert(_engine == nil, @"Already attached to an engine %@.", _engine);
   _engine = engine;
@@ -450,6 +474,39 @@ static void CommonInit(FlutterViewController* controller, FlutterEngine* engine)
 
 - (BOOL)attached {
   return _engine != nil;
+}
+
+- (void)updateSemantics:(const FlutterSemanticsUpdate*)update {
+  NSAssert(_engine.semanticsEnabled, @"Semantics must be enabled.");
+  if (!_engine.semanticsEnabled) {
+    return;
+  }
+  for (size_t i = 0; i < update->nodes_count; i++) {
+    const FlutterSemanticsNode* node = &update->nodes[i];
+    _bridge->AddFlutterSemanticsNodeUpdate(node);
+  }
+
+  for (size_t i = 0; i < update->custom_actions_count; i++) {
+    const FlutterSemanticsCustomAction* action = &update->custom_actions[i];
+    _bridge->AddFlutterSemanticsCustomActionUpdate(action);
+  }
+
+  _bridge->CommitUpdates();
+
+  // Accessibility tree can only be used when the view is loaded.
+  if (!self.viewLoaded) {
+    return;
+  }
+  // Attaches the accessibility root to the flutter view.
+  auto root = _bridge->GetFlutterPlatformNodeDelegateFromID(0).lock();
+  if (root) {
+    if ([self.flutterView.accessibilityChildren count] == 0) {
+      NSAccessibilityElement* native_root = root->GetNativeViewAccessible();
+      self.flutterView.accessibilityChildren = @[ native_root ];
+    }
+  } else {
+    self.flutterView.accessibilityChildren = nil;
+  }
 }
 
 #pragma mark - Private methods
@@ -711,6 +768,11 @@ static void CommonInit(FlutterViewController* controller, FlutterEngine* engine)
     // back.
     [self.view addSubview:_textInputPlugin];
   }
+}
+
+- (std::shared_ptr<flutter::AccessibilityBridgeMac>)createAccessibilityBridgeWithEngine:
+    (nonnull FlutterEngine*)engine {
+  return std::make_shared<flutter::AccessibilityBridgeMac>(engine, self);
 }
 
 - (nonnull FlutterView*)createFlutterViewWithMTLDevice:(id<MTLDevice>)device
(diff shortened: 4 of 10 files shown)


shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
```diff
@@ -68,6 +68,58 @@ - (instancetype)initWithConnection:(NSNumber*)connection
...
+- (BOOL)semanticsEnabled {
+  return !!_bridge;
+}
+
+- (void)setSemanticsEnabled:(BOOL)semanticsEnabled {
+  if (semanticsEnabled == [self semanticsEnabled]) {
+    return;
+  }
+  if (semanticsEnabled) {
+    _bridge = [_engine createAccessibilityBridge:_engine viewController:_controller];
```
loic-sharma, line 111: Why does this accept an engine parameter? Can it not use `self`?

shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
```diff
@@ -68,6 +68,58 @@ - (instancetype)initWithConnection:(NSNumber*)connection
 
 #pragma mark -
 
+// The information of a view that has been attached to the engine.
+@interface FlutterEngineViewRecord : NSObject
```
loic-sharma, line 72: Could you explain the benefits of adding this new `FlutterEngineViewRecord` interface? Alternatively, the `FlutterViewController` could own the bridge and track whether semantics is enabled.

shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
```diff
@@ -531,18 +590,18 @@ - (void)sendInitialSettings {
 }
 
 - (std::weak_ptr<flutter::AccessibilityBridgeMac>)accessibilityBridge {
-  return _bridge;
+  return _viewRecord.bridge;
```
loic-sharma, line 593: Can we remove this method entirely?

shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
loic-sharma, line 111: If the accessibility bridge is tied to the view controller, does it make sense for the engine to create accessibility bridges? Should we move `createAccessibilityBridge:` to `FlutterEngineViewRecord` or `FlutterViewController`?

shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
```diff
@@ -603,6 +654,48 @@ - (FlutterPlatformViewController*)platformViewController {
 
 #pragma mark - Private methods
 
+- (FlutterEngineViewRecord*)viewRecordForId:(uint64_t)viewId {
+  if (viewId == kFlutterDefaultViewId) {
+    return _viewRecord;
+  }
+  return nil;
```
loic-sharma, line 661: Consider adding an assertion here. Currently it's unexpected to get the non-default view ID.

shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
```diff
@@ -68,6 +68,58 @@ - (instancetype)initWithConnection:(NSNumber*)connection
...
+                            engine:(nonnull FlutterEngine*)engine
+                  semanticsEnabled:(BOOL)semanticsEnabled;
+
+- (BOOL)semanticsEnabled;
+
+- (void)setSemanticsEnabled:(BOOL)semanticsEnabled;
+
+@property(nonatomic, readonly, weak) FlutterViewController* controller;
+
+@property(nonatomic, readonly) std::shared_ptr<flutter::AccessibilityBridgeMac> bridge;
```
loic-sharma, line 83: Can this be `std::unique_ptr`?

shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
```diff
@@ -603,6 +654,48 @@ - (FlutterPlatformViewController*)platformViewController {
...
+  }
+  return nil;
+}
+
+- (void)updateSemantics:(const FlutterSemanticsUpdate*)update forViewId:(uint64_t)viewId {
+  FlutterEngineViewRecord* record = [self viewRecordForId:viewId];
+  if (record == nil) {
+    return;
+  }
+  std::shared_ptr<flutter::AccessibilityBridgeMac> bridge = record.bridge;
```
loic-sharma, line 669: The previous code asserted that the bridge had been created: https://github.com/flutter/engine/pull/38855/files#diff-9f979d53c3e186c5c7ab9e9f71924b50535d029f0f994ceafc21fb1f88011c01L921

Should we re-add that?

shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
```diff
@@ -603,6 +654,48 @@ - (FlutterPlatformViewController*)platformViewController {
...
+    return _viewRecord;
+  }
+  return nil;
+}
+
+- (void)updateSemantics:(const FlutterSemanticsUpdate*)update forViewId:(uint64_t)viewId {
+  FlutterEngineViewRecord* record = [self viewRecordForId:viewId];
+  if (record == nil) {
+    return;
+  }
```
loic-sharma, line 668: Should the rest of this method be moved to the `FlutterEngineViewRecord` or `FlutterViewController`? It seems reasonable for the engine to receive semantics updates and dispatch them to the view, but actually applying the semantics update to a view seems like it should be outside the engine's scope.

shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
```diff
@@ -68,6 +68,58 @@ - (instancetype)initWithConnection:(NSNumber*)connection
...
 
+// The information of a view that has been attached to the engine.
+@interface FlutterEngineViewRecord : NSObject
+- (instancetype)initWithController:(nonnull FlutterViewController*)controller
+                            engine:(nonnull FlutterEngine*)engine
+                  semanticsEnabled:(BOOL)semanticsEnabled;
+
+- (BOOL)semanticsEnabled;
+
+- (void)setSemanticsEnabled:(BOOL)semanticsEnabled;
```
cbracken, line 79: This feels a little misleading. Whether semantics is enabled or not is a global setting at the OS level. Within the Flutter app, it feels like we should be modelling this as a single property maybe on FlutterEngine even if we have an accessibility bridge per view.

I kind of wonder if we could actually maintain a single accessibility bridge but have it handle the view/subtree mapping for all the views across the app?

dkwingsmt: It's more like a problem of naming. In the end we need a class to handle or store view-specific a11y information, whether it's called AccessibilityBridge or not. I'm open to better names.

dkwingsmt: But I'll change this setter with a `notifyUpdated` method to follow SSOT.

a-wallen:
Are we already committed to integers for viewIds?

dkwingsmt:
> Are we already committed to integers for viewIds?

Not "committed", but what is your suggestion? Strings are not a good option for their cost in comparison and transmission.

sergiy-sc:
>Last, with this PR, FlutterViewController always initializes its keyboard manager and attaches itself to the engine, even if the engine has not been launched. Although @sergiy-sc's previous PR https://github.com/flutter/engine/pull/36410 seems to have carefully avoided it, I don't see a reason not to do this (and it's much simpler). @sergiy-sc Do you remember why you made it the way?

That's because `initializeKeyboard` is called from `launchEngine`, but if engine is already running, it won't be launched again. So I initialized the keyboard in `initWithEngine` if it's running.

dkwingsmt:
> > Last, with this PR, FlutterViewController always initializes its keyboard manager and attaches itself to the engine, even if the engine has not been launched. Although @sergiy-sc's previous PR #36410 seems to have carefully avoided it, I don't see a reason not to do this (and it's much simpler). @sergiy-sc Do you remember why you made it the way?
> 
> That's because `initializeKeyboard` is called from `launchEngine`, but if engine is already running, it won't be launched again. So I initialized the keyboard in `initWithEngine` if it's running.

Thank you. In that case, I think the change in this PR is safe.

shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
dkwingsmt: You're absolutely right. Thanks for pointing it out.

shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
dkwingsmt: That's a good idea, and it saves me from writing alternatives to the `accessibilityBridge` getter.

shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
dkwingsmt: Besides potentially moving it to `FVC`, I don't think we can remove this method entirely. The `accessibilityBridge` is used by `FlutterPlatformNodeDelegate` and `FlutterPlatformNodeDelegateMac`, for example [here](https://github.com/flutter/engine/blob/88a8d9a36705b0331ab2bdf7fdf60a0ba69f59a2/shell/platform/common/flutter_platform_node_delegate.cc#L41). So many methods are used that exposing `AccessibilityBridgeMac` is probably better than hiding it and adding every method to `FVC`.

This also answers [your other question](https://github.com/flutter/engine/pull/38855#discussion_r1070169895): `_bridge` needs to be a `shared_ptr`, not a `unique_ptr`, so that other class can get a `weak_ptr` out of it. (If we use `unique_ptr` other class can only get a raw pointer... not a modern C++ solution.)

shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
dkwingsmt: This is a `shared_ptr` because other classes (`FlutterPlatformNodeDelegate`) need to get a `weak_ptr` to use it. But I think we might be able to change to a reference, since we're asserting that the bridge is non-null anyway. I'll experiment and report the result.

a-wallen:
> Not "committed", but what is your suggestion? Strings are not a good option for their cost in comparison and transmission.

I don't have anything in mind at the moment. Did you mention anything about it in your design doc?

dkwingsmt:
> > Not "committed", but what is your suggestion? Strings are not a good option for their cost in comparison and transmission.
> 
> I don't have anything in mind at the moment. Did you mention anything about it in your design doc?

It does have [a section "Surface ID"](https://docs.google.com/document/d/1U_ciMTHJgDrnw8kIo9sGKhiKFOycWRmz1yVVI_osCXg/edit#heading=h.ycqyrgvgvn86) that says the ID would be a `uint64_t`. Honestly I didn't think about other options since a `uint64_t` is short and easy to use. The ID is meant to be opaque too (it is generated by the engine, not assigned by the user.)

shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
dkwingsmt: No, we can't. It turns out that `FlutterPlatformNodeDelegateMac` takes a `weak_ptr` as an argument and uses the engine in the future.

shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
dkwingsmt: The bridge has been moved to `FVC`, as well as `createAccessibilityBridge`.

shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
dkwingsmt: This method has been moved to `FVC`.

shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
dkwingsmt: The entire `updateSemantics` has been moved to `FVC`.

shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
dkwingsmt: Added it back.

shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
```diff
@@ -120,6 +122,8 @@ - (void)setUpPlatformViewChannel;
  */
 - (void)handleMethodCall:(FlutterMethodCall*)call result:(FlutterResult)result;
 
+- (void)updateSemantics:(const FlutterSemanticsUpdate*)update forViewId:(uint64_t)viewId;
```
a-wallen: I'd rename the parameter `forViewId` to just `view` since we're updating the semantics of the `view` itself, and not the `viewId`

a-wallen: `forView` works as well.

shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
```diff
@@ -603,6 +583,17 @@ - (FlutterPlatformViewController*)platformViewController {
 
 #pragma mark - Private methods
 
+- (FlutterViewController*)viewControllerForId:(uint64_t)viewId {
+  // TODO(dkwingsmt): The engine only supports single-view, therefore it
+  // only processes the default ID. After the engine supports multiple views,
+  // this method should be able to return the view for any IDs.
+  NSAssert(viewId == kFlutterDefaultViewId, @"Unexpected view ID %llu", viewId);
```
a-wallen, line 590: IIUC, this should be removed when we officially support multi-window - and we're checking here whether the viewId is that of the root window (window 0).

If all of that's true, we might say that "Multiple view controllers aren't currently supported. viewControllerForId got an unexpected viewId: %llu"

shell/platform/darwin/macos/framework/Source/FlutterViewController.mm
```diff
@@ -286,21 +286,31 @@ - (NSArray*)accessibilityChildren {
...
  * Performs initialization that's common between the different init paths.
  */
-static void CommonInit(FlutterViewController* controller) {
-  if (!controller->_engine) {
-    controller->_engine = [[FlutterEngine alloc] initWithName:@"io.flutter"
-                                                      project:controller->_project
-                                       allowHeadlessExecution:NO];
+static void CommonInit(FlutterViewController* controller, FlutterEngine* engine) {
+  if (!engine) {
+    engine = [[FlutterEngine alloc] initWithName:@"io.flutter"
```
a-wallen, line 301: Could you elaborate as to why we are allocating a `FlutterEngine` if the client of this call does not pass a `FlutterEngine?`

a-wallen:
Nice work 👍

shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
dkwingsmt: This method is no longer present in the latest revision. Now `FlutterViewController` has a `updateSemantics:` with no view ID or view parameters.

shell/platform/darwin/macos/framework/Source/FlutterViewController.mm
dkwingsmt: This is a convention with a long history. In this way, the user can simply create a `FlutterViewController` and assigns it to the window, and Flutter just works. The user doesn't have to know that `FVC` creates an engine, and the engine runs a Dart VM, etc. In fact, this is the sample runner that Flutter creates for a default user.

(Although IIRC, the history worked the other way around. At first Flutter only had a `FlutterViewController`, and `FlutterEngine` was split from it afterwards.)

shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
dkwingsmt: Well, there is no way for the user to send a request with a different ID anyway. The framework doesn't support so. So this is really just an internal error.

shell/platform/darwin/macos/framework/Source/FlutterViewController.mm
dkwingsmt: And also, `CommonInit` is only called by the various initializers of `FVC`. Your question might be better phrased as "why we are allocating a `FlutterEngine` for `FVC`'s initializers that do not receive an engine".

shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
```diff
@@ -573,24 +566,11 @@ - (void)setSemanticsEnabled:(BOOL)enabled {
...
-  // Remove the accessibility children from flutter view before reseting the bridge.
-  if (!_semanticsEnabled && self.viewController.viewLoaded) {
-    self.viewController.flutterView.accessibilityChildren = nil;
-  }
-  if (!_semanticsEnabled && _bridge) {
-    _bridge.reset();
-  } else if (_semanticsEnabled && !_bridge) {
-    _bridge = [self createAccessibilityBridge:self viewController:self.viewController];
-  }
+  NSLog(@"Eng %d %d", _semanticsEnabled, self.semanticsEnabled);
```
loic-sharma, line 569: Should this be removed?

shell/platform/darwin/macos/framework/Source/FlutterTextInputPluginTest.mm
```diff
@@ -407,72 +410,65 @@ - (bool)testFirstRectForCharacterRange {
...
+
+  // NSRect rect = [plugin firstRectForCharacterRange:NSMakeRange(0, 0) actualRange:nullptr];
+  // @try {
+  //   OCMVerify(  // NOLINT(google-objc-avoid-throwing-exception)
+  //       [windowMock convertRectToScreen:NSMakeRect(28, 10, 2, 19)]);
+  // } @catch (...) {
+  //   return false;
+  // }
+
+  // return NSEqualRects(rect, NSMakeRect(38, 20, 2, 19));
```
loic-sharma, line 455: Looks like this needs to be updated

shell/platform/darwin/macos/framework/Source/FlutterViewController.mm
a-wallen: Sounds good, I had concerns about multiple engines being created, but I think the API is clear enough that it shouldn't be the case. I would hope that the allocation, if the `FlutterEngine` passed is null, is documented somewhere. That's all.

shell/platform/darwin/macos/framework/Source/FlutterView.h
```diff
@@ -18,7 +18,7 @@
  * backward compatibility, single-view APIs will always operate the view with
  * this ID. Also, the first view assigned to the engine will also have this ID.
  */
-constexpr uint64_t kFlutterDefaultViewId = 0;
+constexpr uint64_t kFlutterDefaultViewId = 1;
```
loic-sharma, line 21: Just curious, why was this changed? Does the framework have any expectations of the default view having an ID of `0`?

shell/platform/darwin/macos/framework/Source/FlutterViewController.mm
```diff
@@ -427,12 +429,80 @@ - (void)onPreEngineRestart {
...
+- (void)notifySemanticsEnabledChanged {
+  BOOL mySemanticsEnabled = !!_bridge;
+  BOOL newSemanticsEnabled = _engine.semanticsEnabled;
+  if (newSemanticsEnabled == mySemanticsEnabled) {
+    return;
+  }
+  if (newSemanticsEnabled) {
+    _bridge = [self createAccessibilityBridgeWithEngine:_engine];
+  } else {
+    // Remove the accessibility children from flutter view before reseting the bridge.
```
loic-sharma, line 441:
```suggestion
    // Remove the accessibility children from flutter view before resetting the bridge.
```

shell/platform/darwin/macos/framework/Source/FlutterViewController.mm
```diff
@@ -427,12 +429,80 @@ - (void)onPreEngineRestart {
   [self initializeKeyboard];
 }
 
+- (void)notifySemanticsEnabledChanged {
+  BOOL mySemanticsEnabled = !!_bridge;
+  BOOL newSemanticsEnabled = _engine.semanticsEnabled;
```
loic-sharma, line 434: Consider making the engine pass its `semanticsEnabled` value as parameter to this method. This would align better with the corresponding embedder API. I don't feel strongly about this, feel free to skip if you prefer as is

dkwingsmt:
@loic-sharma @a-wallen I've split a great part of this PR into https://github.com/flutter/engine/pull/38981, so that this PR will contain only accessibility changes. For more discussion on non-accessibility changes, such as the IDs, let's move to that PR.

shell/platform/darwin/macos/framework/Headers/FlutterViewController.h
```diff
@@ -31,9 +31,19 @@ FLUTTER_DARWIN_EXPORT
 
 /**
  * The Flutter engine associated with this view controller.
+ *
+ * The engine is strongly referenced by the FlutterViewController, and weakly
+ * vice versa.
```
a-wallen, line 36: The engine is strongly referenced by the FlutterViewController, and weakly referenced otherwise.

shell/platform/darwin/macos/framework/Source/FlutterViewController_Internal.h
```diff
@@ -18,14 +21,40 @@
...
+
+- (void)detachFromEngine;
+
+/**
+ * Called by the associated FlutterEngine when FlutterEngine#semanticsEnabled
+ * has changed.
+ */
+- (void)notifySemanticsEnabledChanged;
+
+- (void)updateSemantics:(nonnull const FlutterSemanticsUpdate*)update;
```
a-wallen, line 41: Does `attachToEngine`, `detachFromEngine`, and `updateSemantics` need to be documented?

shell/platform/darwin/macos/framework/Source/FlutterViewController.mm
dkwingsmt: Yeah I prefer the current way. Since `FVC` already references the engine anyway, the current way follows SSOT, so that `FVC` doesn't have to worry about this argument being different from its `engine.semanticsEnabled`.

shell/platform/darwin/macos/framework/Source/FlutterView.h
dkwingsmt: Moving this discussion to https://github.com/flutter/engine/pull/38981#discussion_r1081691327 .

shell/platform/darwin/macos/framework/Source/FlutterViewController.mm
dkwingsmt: I think it's intentionally designed that multiple `FlutterEngine` instances can exist in a single context (for example, in the case of [`FlutterEngineGroup`](https://api.flutter.dev/javadoc/io/flutter/embedding/engine/FlutterEngineGroup.html), a workaround for multi-view in the single-view era), otherwise we would've used a static variable for the singleton engine. 

The documentation has been added to the initializers.

dkwingsmt:
@cbracken The depended PR has been merged, and this PR should be ready for review.

shell/platform/darwin/macos/framework/Source/FlutterViewController.mm
```diff
@@ -437,6 +441,26 @@ - (void)onPreEngineRestart {
...
+    return;
+  }
+  if (newSemanticsEnabled) {
+    _bridge = [self createAccessibilityBridgeWithEngine:_engine];
+  } else {
+    // Remove the accessibility children from flutter view before resetting the bridge.
+    _flutterView.accessibilityChildren = nil;
+    _bridge.reset();
+  }
+  NSAssert(newSemanticsEnabled == !!_bridge, @"Failed to update semantics for the view.");
```
chunhtai, line 457:
```suggestion
  NSAssert(newSemanticsEnabled == mySemanticsEnabled, @"Failed to update semantics for the view.");
```

chunhtai (approved):
LGTM, thanks for refactoring this!

shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
```diff
@@ -586,24 +579,10 @@ - (void)setSemanticsEnabled:(BOOL)enabled {
...
-  // Remove the accessibility children from flutter view before reseting the bridge.
-  if (!_semanticsEnabled && self.viewController.viewLoaded) {
-    self.viewController.flutterView.accessibilityChildren = nil;
-  }
-  if (!_semanticsEnabled && _bridge) {
-    _bridge.reset();
-  } else if (_semanticsEnabled && !_bridge) {
-    _bridge = [self createAccessibilityBridge:self viewController:self.viewController];
-  }
+  [_viewController notifySemanticsEnabledChanged];
```
loic-sharma, line 582: Thoughts on calling the view controller's method `setSemanticsEnabled` and passing `enabled` as a parameter? (Apologies if this is a repeat comment, I thought I had left this comment before but I couldn't find it)

shell/platform/darwin/macos/framework/Source/FlutterEngineTest.mm
```diff
@@ -348,6 +348,31 @@ - (nonnull NSView*)createWithViewIdentifier:(int64_t)viewId arguments:(nullable
...
+  EXPECT_EQ(engine.viewController, nil);
+
+  // Assign the view controller after enabling semantics
+  FlutterViewController* viewController = [[FlutterViewController alloc] initWithEngine:engine
+                                                                                nibName:nil
+                                                                                 bundle:nil];
+  engine.viewController = viewController;
+
+  EXPECT_NE(viewController.accessibilityBridge.lock(), nullptr);
+}
```
loic-sharma, line 374: Nice test 👍

shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
loic-sharma: Ah found it here: https://github.com/flutter/engine/pull/38855#discussion_r1081863524

Current approach is to have a single source of truth. Marking as resolved

loic-sharma (approved):
Nice work! 🥳

shell/platform/darwin/macos/framework/Source/FlutterViewController.mm
dkwingsmt: But `mySemanticsEnabled` is not automatically updated, and this change is wrong. I can extract `mySemanticsEnabled` to be a getter, but I'm thinking that it's not worth it for all its uses in this single method. What's your opinion?

chunhtai: ah nvm, I missed it.

PR #38905, opened 2023-01-16
Add more flexible image loading API

tvolkert:
This adds a new `instantiateImageCodecWithSize` method, which can be used to decode an image into a size, where the target size isn't known until the caller is allowed to inspect the image descriptor.

This enables the use case of loading an image whose aspect ratio isn't known at load time, and which needs to be resized to fit within a bounding box while also maintaining its original aspect ratio.

https://github.com/flutter/flutter/issues/118543

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/lib/ui/painting.dart
+++ b/lib/ui/painting.dart
@@ -2145,6 +2145,13 @@ Future<Codec> instantiateImageCodec(
 ///
 /// The returned future can complete with an error if the image decoding has
 /// failed.
+///
+/// ## Compatibility note on the web
+///
+/// When running Flutter on the web, only the CanvasKit renderer supports image
+/// resizing capabilities (not the HTML renderer). So if image resizing is
+/// critical to your use case, and you're deploying to the web, you should
+/// build using the CanvasKit renderer.
 Future<Codec> instantiateImageCodecFromBuffer(
   ImmutableBuffer buffer, {
   int? targetWidth,
@@ -2153,13 +2160,13 @@ Future<Codec> instantiateImageCodecFromBuffer(
 }) {
   return instantiateImageCodecWithSize(
     buffer,
-    getTargetSize: (ImageDescriptor descriptor) {
+    getTargetSize: (int intrinsicWidth, int intrinsicHeight) {
       if (!allowUpscaling) {
-        if (targetWidth != null && targetWidth! > descriptor.width) {
-          targetWidth = descriptor.width;
+        if (targetWidth != null && targetWidth! > intrinsicWidth) {
+          targetWidth = intrinsicWidth;
         }
-        if (targetHeight != null && targetHeight! > descriptor.height) {
-          targetHeight = descriptor.height;
+        if (targetHeight != null && targetHeight! > intrinsicHeight) {
+          targetHeight = intrinsicHeight;
         }
       }
       return TargetImageSize(width: targetWidth, height: targetHeight);
@@ -2171,23 +2178,23 @@ Future<Codec> instantiateImageCodecFromBuffer(
 ///
 /// This method is a convenience wrapper around the [ImageDescriptor] API.
 ///
-/// The [buffer] parameter is the binary image data (e.g a PNG or GIF binary data).
-/// The data can be for either static or animated images. The following image
-/// formats are supported: {@macro dart.ui.imageFormats}
+/// The [buffer] parameter is the binary image data (e.g a PNG or GIF binary
+/// data). The data can be for either static or animated images. The following
+/// image formats are supported: {@macro dart.ui.imageFormats}
 ///
-/// The [buffer] will be disposed by this method once the codec has been created,
-/// so the caller must relinquish ownership of the [buffer] when they call this
-/// method.
+/// The [buffer] will be disposed by this method once the codec has been
+/// created, so the caller must relinquish ownership of the [buffer] when they
+/// call this method.
 ///
 /// The [getTargetSize] parameter, when specified, will be invoked and passed
-/// the image's [ImageDescriptor] to determine the size to decode the image to.
+/// the image's intrinsic size to determine the size to decode the image to.
 /// The width and the height of the size it returns must be positive values
-/// greater than or equal to 1, or null. It is valid to return a [TargetImageSize]
-/// that specifies only one of `width` and `height` with the other remaining
-/// null, in which case the omitted dimension will be scaled to maintain the
-/// aspect ratio of the original dimensions. When both are null or omitted,
-/// the image will be decoded at its native resolution (as will be the case if
-/// the [getTargetSize] parameter is omitted).
+/// greater than or equal to 1, or null. It is valid to return a
+/// [TargetImageSize] that specifies only one of `width` and `height` with the
+/// other remaining null, in which case the omitted dimension will be scaled to
+/// maintain the aspect ratio of the original dimensions. When both are null or
+/// omitted, the image will be decoded at its native resolution (as will be the
+/// case if the [getTargetSize] parameter is omitted).
 ///
 /// Scaling the image to larger than its intrinsic size should usually be
 /// avoided, since it causes the image to use more memory than necessary.
@@ -2195,6 +2202,13 @@ Future<Codec> instantiateImageCodecFromBuffer(
 ///
 /// The returned future can complete with an error if the image decoding has
 /// failed.
+///
+/// ## Compatibility note on the web
+///
+/// When running Flutter on the web, only the CanvasKit renderer supports image
+/// resizing capabilities (not the HTML renderer). So if image resizing is
+/// critical to your use case, and you're deploying to the web, you should
+/// build using the CanvasKit renderer.
 Future<Codec> instantiateImageCodecWithSize(
   ImmutableBuffer buffer, {
   TargetImageSizeCallback? getTargetSize,
@@ -2202,7 +2216,7 @@ Future<Codec> instantiateImageCodecWithSize(
   getTargetSize ??= _getDefaultImageSize;
   final ImageDescriptor descriptor = await ImageDescriptor.encoded(buffer);
   try {
-    final TargetImageSize targetSize = getTargetSize(descriptor);
+    final TargetImageSize targetSize = getTargetSize(descriptor.width, descriptor.height);
     assert(targetSize.width == null || targetSize.width! > 0);
     assert(targetSize.height == null || targetSize.height! > 0);
     return descriptor.instantiateCodec(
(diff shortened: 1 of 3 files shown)


tvolkert:
@yjbanov I'm wondering how we can support this API in the web engine.  Thoughts?

lib/ui/painting.dart
```diff
@@ -2146,21 +2150,119 @@ Future<Codec> instantiateImageCodecFromBuffer(
...
+}
+
+/// Signature for a callback that determines the size to which an image should
+/// be decoded given its [ImageDescriptor].
+///
+/// See also:
+///
+///  * [instantiateImageCodecWithSize], which used this signature for its
+///    `getTargetSize` argument.
+typedef TargetImageSizeProducer = TargetImageSize Function(ImageDescriptor descriptor);
```
jonahwilliams, line 2224: I think we generally use "Callback" as the suffix for function types

lib/ui/painting.dart
```diff
@@ -2146,21 +2150,119 @@ Future<Codec> instantiateImageCodecFromBuffer(
...
+/// avoided, since it causes the image to use more memory than necessary.
+/// Instead, prefer scaling the [Canvas] transform.
+///
+/// The returned future can complete with an error if the image decoding has
+/// failed.
+Future<Codec> instantiateImageCodecWithSize(
+  ImmutableBuffer buffer, {
+  TargetImageSizeProducer? getTargetSize,
 }) async {
+  getTargetSize ??= (ImageDescriptor descriptor) => const TargetImageSize();
```
jonahwilliams, line 2202: Instead of using `??=` here, create a static function and make it the argument default value:

```
TargetImageSize _defaultGetTargetSize(ImageDescriptor descriptor) {
  return const TargetImageSize();
}

```
TargetImageSizeProducer getTargetSize = _defaultGetTargetSize,
```

```

tvolkert: Heh - I did that at first, but then the `package:flutter/painting.dart` API wrapper has to do something like this:

```dart
  Future<ui.Codec> instantiateImageCodecWithSize(
    ui.ImmutableBuffer buffer, {
    ui.TargetImageSizeProducer? getTargetSize,
  }) {
    if (getTargetSize == null) {
      return ui.instantiateImageCodecWithSize(buffer);
    } else {
      return ui.instantiateImageCodecWithSize(buffer, getTargetSize: getTargetSize);
    }
  }
```

Either that, or the static function has to be public (or worse: copied).  After discussing with Bob, he suggested that he frequently resorts to the method I've done here.  What do you think given those tradeoffs?

jonahwilliams: ahh yeah, that makes sense. I think the approach is fine. I would still use a static function instead of creating a new closure on each run though

jonahwilliams (commented):
I think overall this approach LGTM. How does this look on the web side of things?

lib/ui/painting.dart
tvolkert: Done

lib/ui/painting.dart
tvolkert: Done

tvolkert:
> How does this look on the web side of things?

I added a web impl.  It's not pretty.... definitely interested in knowing if there's a better way.

lib/web_ui/lib/painting.dart
```diff
@@ -491,6 +491,50 @@ Future<Codec> instantiateImageCodecFromBuffer(
...
+  const _SizeOnlyImageDescriptor(this.width, this.height);
+
+  @override
+  final int width;
+
+  @override
+  final int height;
+
+  @override
+  dynamic noSuchMethod(Invocation invocation) => throw UnsupportedError('ImageDescriptor.${invocation.memberName} is not supported on web within a TargetImageSizeCallback.');
```
jonahwilliams, line 504: Nooooooooo, can we avoid using noSuchMethod? :)

tvolkert: TIL!  I'll update it.  Reason I went with this is because it was forcing me to implement all the private methods too, which was a bummer.

jonahwilliams:
Talked about this a bit offline with @chinmaygarde . I think we can already accomplish this with our existing API, but its not very obvious how to do so. For example, you can set one of targetWidth and targetHeight to a maximum value, but don't set the other one - this will preserve the aspect ratio. Then set allowUpscalling to false, this will prevent the image from being resized to be larger.

Given that, I suspect you could probably implement this API in the framework without any dart:ui changes

lib/ui/painting.dart
```diff
@@ -2146,21 +2150,121 @@ Future<Codec> instantiateImageCodecFromBuffer(
...
+///    class when consulted by image decoding methods such as
+///    [instantiateImageCodecWithSize].
+class TargetImageSize {
+  /// Creates a new instance of this class.
+  ///
+  /// The `width` and `height` may both be null, but if they're non-null, they
+  /// must be positive.
+  const TargetImageSize({this.width, this.height})
+      : assert(width == null || width > 0),
+        assert(width == null || width > 0);
```
alanwutang11, line 2242: nit: I think you meant for this to be `height`

lib/web_ui/lib/painting.dart
```diff
@@ -491,6 +491,50 @@ Future<Codec> instantiateImageCodecFromBuffer(
...
+        targetWidth: targetSize.width, targetHeight: targetSize.height, allowUpscaling: false);
+  }
+}
+
+typedef TargetImageSizeCallback = TargetImageSize Function(ImageDescriptor descriptor);
+
+class TargetImageSize {
+  const TargetImageSize({this.width, this.height})
+      : assert(width == null || width > 0),
+        assert(width == null || width > 0);
```
alanwutang11, line 532: nit: `height` here too

lib/ui/painting.dart
```diff
@@ -2146,21 +2150,121 @@ Future<Codec> instantiateImageCodecFromBuffer(
...
+/// aspect ratio of the original dimensions. When both are null or omitted,
+/// the image will be decoded at its native resolution (as will be the case if
+/// the [getTargetSize] parameter is omitted).
+///
+/// Scaling the image to larger than its intrinsic size should usually be
+/// avoided, since it causes the image to use more memory than necessary.
+/// Instead, prefer scaling the [Canvas] transform.
+///
+/// The returned future can complete with an error if the image decoding has
+/// failed.
```
alanwutang11, line 2197: maybe note here that image resizing capabilities are only available with the canvaskit renderer and not html renderer (we should probably update the api docs for `instantiateImageCodec` to note this too)

tvolkert:
> Talked about this a bit offline with @chinmaygarde . I think we can already accomplish this with our existing API, but its not very obvious how to do so. For example, you can set one of targetWidth and targetHeight to a maximum value, but don't set the other one - this will preserve the aspect ratio. Then set allowUpscalling to false, this will prevent the image from being resized to be larger.
> 
> Given that, I suspect you could probably implement this API in the framework without any dart:ui changes

That only works if you at least know the aspect ratio of the image.  If I want to size an image to fit within e.g. a square bounding box of 100x100, but I don't know if the image is portrait or landscape orientation, then I don't know whether to say "load the image with width=100 and let height be flexible" or "load the image at height=100 and let the width be flexible".

This was the use case that drove this change.  It comes about when you load a stream of random images over the network via a public API (e.g. Google Photos, Flickr, etc.).

jonahwilliams:
so in that case you'd have to say maxWidth = 100, maxHeight = 100. but then that doesn't preserve the aspect ratio?

tvolkert:
Correct - that stretches the image to a square.

tvolkert:
Note that the current API doesn't specify maxWidth & maxHeight -- it specifies targetWidth & targetHeight

jonahwilliams:
with allowUpscaling= false then targetSise becomes maxSize essentially

jonahwilliams:
Could we accomplish something similar with a preserveAspectRatio flag that behaved like you expected?

tvolkert:
No, allowUpscaling=false only ensures that you don't increase the target width beyond the intrinsic width (or the target height beyond the intrinsic height)

https://github.com/flutter/engine/blob/7d40e77d003540fb3e8908d730110197e5b72c51/lib/ui/painting.dart#L2151-L2158

tvolkert:
> Could we accomplish something similar with a preserveAspectRatio flag that behaved like you expected?

We already do preserve the aspect ratio if you only specify _one of_ the target width or height.  But the problem is, you don't know which one to pass if you don't already know its aspect ratio.

We could (and I considered) add a new parameter (or new signature) that had the caller pass a bounding box, and we'd fix the target image within the bounding box while maintaining the aspect ratio.  But such an API change would be just as disruptive as the change in this PR, but less flexible.

lib/ui/painting.dart
tvolkert: Done

lib/web_ui/lib/painting.dart
tvolkert: Done - I got rid of the need for the ImageDescriptor at all.

lib/web_ui/lib/painting.dart
```diff
@@ -491,6 +491,42 @@ Future<Codec> instantiateImageCodecFromBuffer(
...
+      }
+    } finally {
+      codec.dispose();
+    }
+  }
+}
+
+typedef TargetImageSizeCallback = TargetImageSize Function(int intrinsicWidth, int intrinsicHeight);
+
+class TargetImageSize {
```
cbracken, line 521: Are there other places in dart:ui we could use an int-based size? Might make sense to generalise the name so others could use it later, if so.

I can’t think of a great name though. Maybe something like `PhysicalSize`.

cbracken (approved):
Overall LGTM, and looks like all comments have been addressed. One nit about whether we could generalise `TargetImageSize` to avoid adding more such classes later elsewhere.

lib/web_ui/lib/painting.dart
tvolkert: I thought about that, but a real `PhysicalSize` class would want both width & height to be non-null.  This is unique in how it allows one or both of those to be null.

jonahwilliams:
I think that this is a good feature, but I think it should actually be implemented in foundation instead of dart:ui. It does not require any new native bindings as far as I can tell.

tvolkert:
There's an optimization I've discussed with @yjbanov and @alanwutang11 that will need to be done in the web engine, so it's probably best left here.

jonahwilliams:
Which one?

tvolkert:
If the user passes a callback, but it returns a `TargetImageSize` with either both width & height null or the same size as the intrinsic size, we can pass the existing codec without discarding it and having to re-create one.  It relies on a web engine implementation detail whereby we can avoid getting the first frame in order to know the intrinsic size of the image.

lib/web_ui/lib/painting.dart
jonahwilliams: More similar to an IntConstraints type, not quite a size. We do really need an ISize type, but probably unrelated.

jonahwilliams:
Interesting! Please file a bug/todo for tracking this

jonahwilliams (approved):
LGTM

tvolkert:
https://github.com/flutter/flutter/issues/118985

PR #38919, opened 2023-01-16
[WIP] Remove references to Observatory

bkonyi:
Observatory is being deprecated for Dart 3.0 so it should no longer be referenced in tooling messaging / flags.

See https://github.com/dart-lang/sdk/issues/50233


shell/platform/darwin/ios/framework/Source/FlutterDartVMServicePublisher.mm
```diff
@@ -83,18 +83,18 @@ - (void)publishServiceProtocolPort:(NSURL*)url {
   // Physical devices need to request all interfaces.
   uint32_t interfaceIndex = 0;
 #endif  // TARGET_IPHONE_SIMULATOR
-  const char* registrationType = "_dartobservatory._tcp";
+  const char* registrationType = "_dartVmService._tcp";
```
bkonyi, line 86: @jmagman I'd like to change the mDNS entry name to stop referencing Observatory. Are there any gotchas I should be aware of? Should we be supporting both `_dartVmService._tcp` and `_dartobservatory._tcp` until flutter_tools is updated?

jmagman: Yes, you'll need to support both until the tool is updated or the engine->flutter roll will have a wall of test failures.  I don't see any usages in g3 and I can't think of gotchas outside of the tool (please add me as a reviewer when you get to that PR).

cc @vashworth who is working on that area of the tool at the moment https://github.com/flutter/flutter/pull/118104

shell/platform/embedder/tests/embedder_config_builder.cc
```diff
@@ -113,7 +113,7 @@ EmbedderConfigBuilder::EmbedderConfigBuilder(
     SetSemanticsCallbackHooks();
     SetLogMessageCallbackHook();
     SetLocalizationCallbackHooks();
-    AddCommandLineArgument("--disable-observatory");
+    AddCommandLineArgument("--disable-vm-service");
```
jmagman, line 116: You should also support `--disable-observatory` until the tool is updated.

jmagman: Oh this is in `tests`, not sure if that's necessary then.

shell/platform/darwin/ios/framework/Source/FlutterDartVMServicePublisher.mm
bkonyi: Makes sense. I have a WIP PR ready for updating the tool here: https://github.com/flutter/flutter/pull/118577, so I'll add you to that PR once this change lands and I've confirmed things are working as expected.

shell/platform/embedder/tests/embedder_config_builder.cc
bkonyi: Yeah, this is internal to the engine so we should be good to just change this as part of this PR.

jmagman (approved):

shell/platform/android/io/flutter/embedding/engine/FlutterJNI.java
```diff
@@ -230,14 +230,14 @@ public boolean getIsSoftwareRenderingEnabled() {
...
-   * Observatory URI for the VM instance.
+   * VM Service URI for the VM instance.
    *
    * <p>Its value is set by the native engine once {@link #init(Context, String[], String, String,
    * String, long)} is run.
    */
   @Nullable
-  public static String getObservatoryUri() {
-    return observatoryUri;
+  public static String getVMServiceUri() {
```
bparrishMines, line 239: @bkonyi The removal of this is breaking the `espresso` plugin: https://github.com/flutter/plugins/blob/main/packages/espresso/android/src/main/java/androidx/test/espresso/flutter/action/FlutterViewAction.java#L108

And is breaking the Flutter to plugins roller. 

Could we add the `getObservatoryUri` back and mark it as deprecated until `getVMServiceUri` makes it to stable?

zanderso: @bparrishMines if this issue is blocking rolls, please file a P0 on GitHub, and assign @bkonyi.

bparrishMines: Created https://github.com/flutter/flutter/issues/118896

PR #38941, opened 2023-01-17
[windows] Fix crash when calling `FlutterDesktopMessengerSetCallback` after engine shutdown

jnschulze:
Fixes https://github.com/flutter/flutter/issues/118611

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/shell/platform/windows/flutter_windows.cc
+++ b/shell/platform/windows/flutter_windows.cc
@@ -292,10 +292,12 @@ void FlutterDesktopMessengerSetCallback(FlutterDesktopMessengerRef messenger,
                                         const char* channel,
                                         FlutterDesktopMessageCallback callback,
                                         void* user_data) {
-  flutter::FlutterDesktopMessenger::FromRef(messenger)
-      ->GetEngine()
-      ->message_dispatcher()
-      ->SetMessageCallback(channel, callback, user_data);
+  auto engine =
+      flutter::FlutterDesktopMessenger::FromRef(messenger)->GetEngine();
+  if (engine) {
+    engine->message_dispatcher()->SetMessageCallback(channel, callback,
+                                                     user_data);
+  }
 }
 
 FlutterDesktopMessengerRef FlutterDesktopMessengerAddRef(
--- a/shell/platform/windows/flutter_windows_unittests.cc
+++ b/shell/platform/windows/flutter_windows_unittests.cc
@@ -279,5 +279,25 @@ TEST_F(WindowsTest, GetGraphicsAdapter) {
   ASSERT_TRUE(SUCCEEDED(dxgi_adapter->GetDesc(&desc)));
 }
 
+// Verify that calling FlutterDesktopMessengerSetCallback
+// after shutting the engine down is safe.
+// Prevent regression: https://github.com/flutter/flutter/issues/118611
+TEST_F(WindowsTest, SetMessengerCallbackAfterEngineShutdown) {
+  auto& context = GetContext();
+  WindowsConfigBuilder builder(context);
+  EnginePtr engine{builder.InitializeEngine()};
+  ASSERT_NE(engine, nullptr);
+
+  auto messenger = FlutterDesktopEngineGetMessenger(engine.get());
+  FlutterDesktopMessengerAddRef(messenger);
+  ASSERT_TRUE(FlutterDesktopMessengerIsAvailable(messenger));
+
+  engine.reset();
+
+  ASSERT_FALSE(FlutterDesktopMessengerIsAvailable(messenger));
+  FlutterDesktopMessengerSetCallback(messenger, "my_channel", nullptr, nullptr);
+  FlutterDesktopMessengerRelease(messenger);
+}
+
 }  // namespace testing
 }  // namespace flutter


shell/platform/windows/flutter_windows.cc
```diff
@@ -292,10 +292,12 @@ void FlutterDesktopMessengerSetCallback(FlutterDesktopMessengerRef messenger,
...
-  flutter::FlutterDesktopMessenger::FromRef(messenger)
-      ->GetEngine()
-      ->message_dispatcher()
-      ->SetMessageCallback(channel, callback, user_data);
+  auto engine =
+      flutter::FlutterDesktopMessenger::FromRef(messenger)->GetEngine();
+  if (engine) {
+    engine->message_dispatcher()->SetMessageCallback(channel, callback,
+                                                     user_data);
+  }
```
loic-sharma, line 300: If we move forward with this approach, we should also update `FlutterDesktopMessengerSendWithReply` and `FlutterDesktopMessengerSendResponse`.

loic-sharma:
Thanks for reporting this and sending a pull request! Here are some potential alternatives we should consider before moving forward with this:

1. Improve but keep the crashes. Assert that `FlutterDesktopMessengerIsAvailable` before doing messenger operations. Document which messenger operations are illegal if the engine is shutdown.
2. Destroy plugins before shutting down the engine.
3. Move this logic up to C++ wrapper around the desktop APIs.

Let me do some more digging to see which approach would best. /cc @gaaclarke @cbracken in case you have thoughts already here.

loic-sharma:
@gaaclarke The engine's shutdown invalidates the messenger before plugins are shutdown:

https://github.com/flutter/engine/blob/5932badce44a20deb43e78443d0bd4f33910c2ac/shell/platform/windows/flutter_windows_engine.cc#L217-L220

Could that be swapped so that plugins have access to a valid messenger during shutdown? That would prevent these kinds of issues.

Here's the call stack to shut down plugins:

```
app.dll!camera_windows::CameraPlugin::~CameraPlugin()
...
app.dll!flutter::PluginRegistrar::ClearPlugins()
app.dll!flutter::PluginRegistrarWindows::~PluginRegistrarWindows()
...
app.dll!flutter::PluginRegistrarManager::OnRegistrarDestroyed(FlutterDesktopPluginRegistrar * registrar)
flutter_windows.dll!flutter::FlutterWindowsEngine::Stop() 
flutter_windows.dll!flutter::FlutterWindowsEngine::~FlutterWindowsEngine()
flutter_windows.dll!flutter::FlutterWindowsEngine::~FlutterWindowsEngine()
```

shell/platform/windows/flutter_windows.cc
gaaclarke: I think there is a difference between setting the handler and sending a message.  I think the OP was trying to set the handler to `nullptr` so that they know the proper cleanup is happening instead of relying on the engine to do so.  They wanted to make that not an error.  I think it would be acceptable to make that not a crash, but to keep sending messages in the destructors a crash (we can improve the crash messaging with an assert if we don't already).

gaaclarke:
> Could that be swapped so that plugins have access to a valid messenger during shutdown? That would prevent these kinds of issues.

We could look into it, but I don't think the end goal is worth the risk or investment to research.  I think sending messages in the destructor should be an error.  I guess it's debatable if we want it to be a crash or just a log message.

jnschulze:
> 1. Improve but keep the crashes. Assert that `FlutterDesktopMessengerIsAvailable` before doing messenger operations. Document which messenger operations are illegal if the engine is shutdown.

I think the C++ client wrapper should implicitly handle this scenario. According to the documentation, "the caller is responsible for unregistering the handler if it should no longer be called". Why should there be an restriction that forbids calling it from a plugin destructor?

My use case is a video player plugin which has one event channel per player instance. Each event channel is owned by the corresponding player instance. Hence, it's not just about removing the `SetStreamHandler(...)` in the plugin destructor, but introducing a second destruction path for players which gets used upon plugin destruction.

> 3. Move this logic up to C++ wrapper around the desktop APIs.

Yes, moving the check to `BinaryMessengerImpl::SetMessageHandler` (and using `FlutterDesktopMessengerIsAvailable`) is certainly an option as well.

loic-sharma:
> > 1. Improve but keep the crashes. Assert that `FlutterDesktopMessengerIsAvailable` before doing messenger operations. Document which messenger operations are illegal if the engine is shutdown.
> 
> I think the C++ client wrapper should implicitly handle this scenario. According to the documentation, "the caller is responsible for unregistering the handler if it should no longer be called". Why should there be a restriction that forbids calling it from a plugin destructor?

In hindsight, [that comment](https://github.com/flutter/engine/blob/5932badce44a20deb43e78443d0bd4f33910c2ac/shell/platform/common/client_wrapper/include/flutter/method_channel.h#L94-L96) is easily misunderstood. It's supposed to mean: 

> Destroying the `MethodChannel` does not deregister the handler as the lifetimes of the `MethodChannel` and its handler are fully decoupled. The handler is destroyed when it is deregistered or when the engine is shut down.

We should update that :)

> Hence, it's not just about removing the `SetStreamHandler(...)` in the plugin destructor, but introducing a second destruction path for players which gets used upon plugin destruction.

Interesting, could you tell us more about how players get used upon plugin destruction?

jnschulze:
> Interesting, could you tell us more about how players get used upon plugin destruction?

The left players just get destroyed (and so do their event channels).
However, I don't want to distinguish between destroying them explicitly (i.e. by invoking some "destroy" method channel method) and destroying them when the app gets shut down.

loic-sharma:
@gaaclarke and I chatted and our plan is:

Short-term: add asserts to the Flutter desktop API to ensure the messenger is valid before doing operations on it. Also, improve relevant comments.

Long-term: consider making the messenger valid during engine shut down, however, disallow sending messages. In other words, engine shutdown would have two separate phases:

1. Notify components of engine shutdown. Allow things like plugins to run shutdown logic
2. Once shutdown logic is completed, cleanup resources like the messenger

This requires investigation to see if it's a good idea.

PR #38948, opened 2023-01-17
[web] dont look up webgl params if no GPU is available

jonahwilliams:
Potentially fixes https://github.com/flutter/flutter/issues/118633

The glcontext being 0 indicates we are going to use software rendering, so don't try to look up gl parameters.


yjbanov (approved):
If this is testable, would be good to have a test, but if we don't have the right test environment (e.g. it needs non-headless browser) then maybe this is the best we can do.

LGTM.

PR #38964, opened 2023-01-18
[Impeller] Fix contiguous clip restoration

bdero:
Because we treat the stencil buffer like a heightmap, clip restores work by flattening out everything above a certain height on the stencil buffer. This means when we have a group of contiguous clip restore entities, we can just render the last one and get identical results -- this is what we do.

We track a stack of stencil coverage rectangles to cull draw calls that obviously won't impact any pixels due to the stencil heightmap. The code that handles popping this stack assumed that the restore would always be `current_stencil_depth - 1`, which is wrong -- a single clip restore entity may reduce the stencil height by any amount, and so we need to do the same for the stencil coverage stack.


chinmaygarde (approved):

zanderso (approved):
LGTM

PR #38981, opened 2023-01-18
[macOS] Formalize FlutterViewController's initialization flow, and prohibit replacing

dkwingsmt:
1. This PR adds `FlutterViewController.id`, and formalizes how `FlutterViewController` is initialized by `FlutterEngine`:
* When `FlutterViewController` is set to the engine, the engine calls `FlutterViewController#attachToEngine:withId:`.
* When `FlutterViewController` is unset from the engine, the engine calls `FlutterViewController#detachFromEngine`.

This allows the ID to be generated by the engine.

2. The default View ID is now 1 instead of 0. In this way, we can use 0 as an empty value, the value for `FlutterViewController.id` when the FVC is not attached to an engine. (IDs are meant to be opaque, and views should be a map, not an array anyway.)

3. This PR moves almost all initialization code into `FVC`'s `CommonInit` to ensure that all init path produces the expected result.

4. With this PR, FlutterEngine no long supports replacing viewController (from non-null to a different non-null). Setting the view controller (from null to non-null) is intended to allow pre-warming an engine before displaying it in add-to-app, but replacing the view controller does not make much sense.

- If the developer wants to persist the state of the view between different windows, they could just attach the view controller to different windows.
- In fact, Flutter doesn't even have a way to create a FlutterViewController that is unattached to an engine. All initializers either require an engine or create an engine.

As a result, unit tests that used to create `FVC` with the wrong initializer are fixed.

## Pre-launch Checklist

- [ ] I signed the [CLA].

### DIFF
--- a/shell/platform/darwin/macos/framework/Headers/FlutterEngine.h
+++ b/shell/platform/darwin/macos/framework/Headers/FlutterEngine.h
@@ -67,6 +67,15 @@ FLUTTER_DARWIN_EXPORT
  *
  * The default view always has ID kFlutterDefaultViewId, and is the view
  * operated by the APIs that do not have a view ID specified.
+ *
+ * Setting this field from nil to a non-nil view controller also updates
+ * the view controller's engine and ID.
+ *
+ * Setting this field to non-nil to nil will terminate the engine if
+ * allowHeadlessExecution is NO.
+ *
+ * Setting this field from non-nil to a different non-nil is prohibited and
+ * will throw an assertion error.
  */
 @property(nonatomic, nullable, weak) FlutterViewController* viewController;
 
--- a/shell/platform/darwin/macos/framework/Headers/FlutterEngine.h
+++ b/shell/platform/darwin/macos/framework/Headers/FlutterEngine.h
@@ -71,11 +71,11 @@ FLUTTER_DARWIN_EXPORT
  * Setting this field from nil to a non-nil view controller also updates
  * the view controller's engine and ID.
  *
- * Setting this field to non-nil to nil will terminate the engine if
+ * Setting this field from non-nil to nil will terminate the engine if
  * allowHeadlessExecution is NO.
  *
- * Setting this field from non-nil to a different non-nil is prohibited and
- * will throw an assertion error.
+ * Setting this field from non-nil to a different non-nil FlutterViewController
+ * is prohibited and will throw an assertion error.
  */
 @property(nonatomic, nullable, weak) FlutterViewController* viewController;
 
--- a/shell/platform/darwin/macos/framework/Headers/FlutterEngine.h
+++ b/shell/platform/darwin/macos/framework/Headers/FlutterEngine.h
@@ -7,6 +7,8 @@
 
 #import <Foundation/Foundation.h>
 
+#include <stdint.h>
+
 #import "FlutterBinaryMessenger.h"
 #import "FlutterDartProject.h"
 #import "FlutterMacros.h"
@@ -14,6 +16,17 @@
 #import "FlutterTexture.h"
 
 // TODO: Merge this file with the iOS FlutterEngine.h.
+
+/**
+ * The view ID for APIs that don't support multi-view.
+ *
+ * Some single-view APIs will eventually be replaced by their multi-view
+ * variant. During the deprecation period, the single-view APIs will coexist with
+ * and work with the multi-view APIs as if the other views don't exist.  For
+ * backward compatibility, single-view APIs will always operate the view with
+ * this ID. Also, the first view assigned to the engine will also have this ID.
+ */
+extern const uint64_t kFlutterDefaultViewId;
 
 @class FlutterViewController;
 
--- a/shell/platform/darwin/macos/framework/Headers/FlutterViewController.h
+++ b/shell/platform/darwin/macos/framework/Headers/FlutterViewController.h
@@ -25,22 +25,41 @@ typedef NS_ENUM(NSInteger, FlutterMouseTrackingMode) {
 
 /**
  * Controls a view that displays Flutter content and manages input.
+ *
+ * A FlutterViewController works with a FlutterEngine. Upon creation, the view
+ * controller is always added to an engine, either a given engine, or it implicitly
+ * creates an engine and add itself to that engine.
+ *
+ * The FlutterEngine assigns each view controller attached to it a unique ID.
+ * Each view controller corresponds to a view, and the ID is used by the framework
+ * to specify which view to operate.
+ *
+ * A FlutterViewController can also be unattached to an engine after it is manually
+ * unset from the engine, or transiently during the initialization process.
+ * An unattached view controller is invalid. Whether the view controller is attached
+ * can be queried using FlutterViewController#attached.
+ *
+ * The FlutterViewController strongly references the FlutterEngine, while
+ * the engine weakly the view controller. When a FlutterViewController is deallocated,
+ * it automatically removes itself from its attached engine. When a FlutterEngine
+ * has no FlutterViewControllers attached, it might shut down itself or not depending
+ * on its configuration.
  */
 FLUTTER_DARWIN_EXPORT
 @interface FlutterViewController : NSViewController <FlutterPluginRegistry>
 
 /**
  * The Flutter engine associated with this view controller.
- *
- * The engine is strongly referenced by the FlutterViewController, and the
- * ViewController weakly from the engine.
  */
 @property(nonatomic, nonnull, readonly) FlutterEngine* engine;
 
 /**
  * The identifier for this view controller.
  *
- * The ID is assigned when the view controller is added to FlutterEngine.
+ * The ID is assigned by FlutterEngine when the view controller is attached.
+ *
+ * If the view controller is unattached (see FlutterViewController#attached),
+ * reading this property throws an assertion.
  */
 @property(nonatomic, readonly) uint64_t id;
 
@@ -82,6 +101,12 @@ FLUTTER_DARWIN_EXPORT
 - (nonnull instancetype)initWithEngine:(nonnull FlutterEngine*)engine
                                nibName:(nullable NSString*)nibName
                                 bundle:(nullable NSBundle*)nibBundle NS_DESIGNATED_INITIALIZER;
+
+/**
+ * Return YES if the view controller is attached to an engine.
+ */
+- (BOOL)attached;
+
 /**
  * Invoked by the engine right before the engine is restarted.
  *
--- a/shell/platform/darwin/macos/framework/Headers/FlutterViewController.h
+++ b/shell/platform/darwin/macos/framework/Headers/FlutterViewController.h
@@ -4,6 +4,8 @@
 
 #import <Cocoa/Cocoa.h>
 
+#include <memory>
+
 #import "FlutterEngine.h"
 #import "FlutterMacros.h"
 #import "FlutterPlatformViews.h"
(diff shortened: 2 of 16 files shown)


shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
```diff
@@ -411,29 +410,28 @@ - (void)loadAOTData:(NSString*)assetsDir {
...
+    [_viewController attachToEngine:self withId:kFlutterDefaultViewId];
+  } else if (_viewController != nil && controller == nil) {
+    [_viewController detachFromEngine];
+    _viewController = nil;
+    if (!_allowHeadlessExecution) {
       [self shutDownEngine];
     }
+  } else {
+    NSLog(@"Failed to set view controller to the engine: "
+          @"Replacing the view controller of the engine is not supported.");
```
loic-sharma, line 428: Should this crash in debug mode?

shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
```diff
@@ -679,9 +688,7 @@ - (void)shutDownEngine {
     return;
   }
 
-  if (_viewController && _viewController.flutterView) {
-    [_viewController.flutterView shutdown];
-  }
+  [self.viewController.flutterView shutdown];
```
loic-sharma, line 691: I may be missing something, but won't `viewController` be nil in headless mode here?

shell/platform/darwin/macos/framework/Source/FlutterView.h
```diff
@@ -18,7 +18,7 @@
  * backward compatibility, single-view APIs will always operate the view with
  * this ID. Also, the first view assigned to the engine will also have this ID.
  */
-constexpr uint64_t kFlutterDefaultViewId = 0;
+constexpr uint64_t kFlutterDefaultViewId = 1;
```
loic-sharma, line 21: Just so this comment isn't lost: https://github.com/flutter/engine/pull/38855#discussion_r1073934050

Consider double-checking with @goderbauer that the framework does not have any expectations that the default view has an ID of `0`.

shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
dkwingsmt: I think assertions are for internal errors, and "misuses" should trigger something else, either an exception (which is discouraged in C++ and ObjC), an error code, or a log. But I can change it if you still support so.

loic-sharma (approved):
LGTM, but please get sign-off from folks that are more familiar with the macOS embedder

shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
loic-sharma: image

dkwingsmt: Yeah, this feature has screwed me up more than I like it. But it exists, so...

shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
loic-sharma: I'm worried a developer might try to replace a flutter view controller without realizing their code doesn't actually work. I don't think the NSLog is a strong enough signal.

FYI it looks like there are `NSAssert`s to validate misuses, for example: https://github.com/flutter/engine/blob/78bbea005d27e77193793cda562d5d8ef53070b4/shell/platform/darwin/ios/framework/Source/FlutterView.mm#L23-L26

a-wallen (approved):
LGTM

shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
dkwingsmt: Done. I've also added fixing instructions to the assertions.

shell/platform/darwin/macos/framework/Headers/FlutterEngine.h
```diff
@@ -67,6 +67,15 @@ FLUTTER_DARWIN_EXPORT
  *
  * The default view always has ID kFlutterDefaultViewId, and is the view
  * operated by the APIs that do not have a view ID specified.
+ *
+ * Setting this field from nil to a non-nil view controller also updates
+ * the view controller's engine and ID.
+ *
+ * Setting this field to non-nil to nil will terminate the engine if
```
goderbauer, line 74: typo: to -> from ?

shell/platform/darwin/macos/framework/Headers/FlutterViewController.h
```diff
@@ -43,6 +53,12 @@ FLUTTER_DARWIN_EXPORT
 /**
  * Initializes a controller that will run the given project.
  *
+ * In this initializer, this controller creates an engine, and is attached to
+ * this engine as the default controller. In this way, this controller can not
+ * be set to other engines. This initializer is suitable for the first Flutter
+ * view controller of the app. To use the controller for an existing engine,
```
goderbauer, line 59: nit: for an existing... -> with an existing...?

shell/platform/darwin/macos/framework/Source/FlutterView.h
goderbauer: The framework makes no assumption about the value of the id. Why does this have to change from 0 to 1, though?

shell/platform/darwin/macos/framework/Source/FlutterViewController.mm
```diff
@@ -427,12 +429,23 @@ - (void)onPreEngineRestart {
...
+           _engine, _id);
+  _engine = engine;
+  _id = viewId;
+}
+
+- (void)detachFromEngine {
+  NSAssert(_engine != nil && _id != 0, @"Not attached to an engine, engine %@ ID %llu.", _engine,
+           _id);
+  _engine = nil;
+  _id = 0;
```
goderbauer, line 443: Ah, are you shifting the default so 0 becomes available to denote "not attached"? Should we use a constant for this? Might also be worthwhile to mention that in the doc on the "id" property, i.e. that accessing the ID is only valid when an engine is attached? Maybe we can even assert if somebody tries to access the ID while no engine is attached.

goderbauer: Also, if we make accessing "ID" invalid when no engine is attached, do we really need to reserve a value for that case? It feels a little odd to me that view IDs don't include zero...

shell/platform/darwin/macos/framework/Headers/FlutterViewController.h
```diff
@@ -31,9 +31,19 @@ FLUTTER_DARWIN_EXPORT
 
 /**
  * The Flutter engine associated with this view controller.
+ *
+ * The engine is strongly referenced by the FlutterViewController, and weakly
+ * vice versa.
```
cbracken, line 36: Maybe:
```suggestion
 * The engine is strongly referenced by the FlutterViewController, and the
 * ViewController weakly from the engine.
```

shell/platform/darwin/macos/framework/Headers/FlutterViewController.h
```diff
@@ -31,9 +31,19 @@ FLUTTER_DARWIN_EXPORT
...
  */
 @property(nonatomic, nonnull, readonly) FlutterEngine* engine;
 
+/**
+ * The identifier for this view controller.
+ *
+ * The ID is assigned when the view controller is added to FlutterEngine.
```
cbracken, line 43: Maybe expand on this, like:
```suggestion
 * The ID is assigned when the view controller is assigned to the                             
 * FlutterEngine.viewController property.
```

shell/platform/darwin/macos/framework/Headers/FlutterViewController.h
```diff
@@ -43,6 +53,12 @@ FLUTTER_DARWIN_EXPORT
 /**
  * Initializes a controller that will run the given project.
  *
+ * In this initializer, this controller creates an engine, and is attached to
+ * this engine as the default controller. In this way, this controller can not
```
cbracken, line 57:
```suggestion
 * that engine as the default controller. In this way, this controller can not
```

shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
```diff
@@ -411,29 +410,38 @@ - (void)loadAOTData:(NSString*)assetsDir {
 }
 
 - (void)setViewController:(FlutterViewController*)controller {
-  if (_viewController != controller) {
+  if (_viewController == controller) {
+    // From nil to nil, or from non-nil to the same controller.
+    return;
+  }
+  if (_viewController == nil && controller != nil) {
```
cbracken, line 417: ~You can simplify this since if `_viewController` is nil, at this point we know that `controller` is non-nil or they'd have been equal in the check above.~

Actually, regardless of the fact that we _can_ simplify these, I think from a readability point of view this is easier to reason about. I'd be tempted to add comments though just for easy scanning of the code.
```suggestion
  // From nil to non-nil.
  if (_viewController == nil && && controller != nil) {
```

shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
```diff
@@ -411,29 +410,38 @@ - (void)loadAOTData:(NSString*)assetsDir {
...
+             controller.engine);
     _viewController = controller;
-
-    if (_semanticsEnabled && _bridge) {
-      _bridge->UpdateDefaultViewController(_viewController);
-    }
-
-    if (!controller && !_allowHeadlessExecution) {
+    [_viewController attachToEngine:self withId:kFlutterDefaultViewId];
+  } else if (_viewController != nil && controller == nil) {
```
cbracken, line 426: ~Similarly, we know that _viewController is non-nil due to the check right above this.~

Similar comment as above; I think I'd leave this check as-is but add the comment.
```suggestion
  } else if (_viewController != nil && controller == nil) {
    // From non-nil to nil.
```

shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
```diff
@@ -411,29 +410,38 @@ - (void)loadAOTData:(NSString*)assetsDir {
...
+    [_viewController detachFromEngine];
+    _viewController = nil;
+    if (!_allowHeadlessExecution) {
       [self shutDownEngine];
     }
+  } else {
```
cbracken, line 432:
```suggestion
  } else {
    // From non-nil to a different non-nil view controller.
```

shell/platform/darwin/macos/framework/Source/FlutterViewController.mm
dkwingsmt: Currently the `id` property is a normal property, neither getter or setter overridden, which I think it's good simplicity. Technically we don't need to a special value for an invalid ID value, we can instead tell the user to not trust the ID value if the FVC is unattached. Do you prefer this way?

goderbauer: I think I would prefer that over another magic value. We should document that, though. And I would still argue for properly asserting if someone accesses the ID while the controller is detached (AFAIK, currently, ObjC just generates the getter/setter, making those explicit, shouldn't be that complex).

dkwingsmt: Done. Reading the ID now asserts that the engine is attached.

shell/platform/darwin/macos/framework/Headers/FlutterViewController.h
dkwingsmt: The mechanism is a little more complicated than this, so I decided to explain everything in FlutterViewController's doc. Can you take a look?

dkwingsmt:
@cbracken @goderbauer I should have addressed all your comments. Can you take another look?

shell/platform/darwin/macos/framework/Headers/FlutterEngine.h
```diff
@@ -15,6 +17,17 @@
 
 // TODO: Merge this file with the iOS FlutterEngine.h.
 
+/**
+ * The view ID for APIs that don't support multi-view.
+ *
+ * Some single-view APIs will eventually be replaced by their multi-view
+ * variant. During the deprecation period, the single-view APIs will coexist with
+ * and work with the multi-view APIs as if the other views don't exist.  For
+ * backward compatibility, single-view APIs will always operate the view with
```
goderbauer, line 26: nit: operate _on_ (or something like that)?

goderbauer (approved):
LGTM

PR #38989, opened 2023-01-18
[ios_platform_view] MaskView pool to reuse maskViews.

cyanglaz:
The maskViews are constructed and destructed every frame for every PlatformView on the screen. It slows down the performance quite a lot.
This PR adds a pool that caches maximum of 5 maskViews. These maskViews are reused every frame.
If there are more than 5 maskViews required in a frame, extra maskViews are constructed and destructed each frame. (This scenario is very unlikely in real life application)

Part of https://github.com/flutter/flutter/issues/116640

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/shell/platform/darwin/ios/framework/Source/FlutterPlatformViews.mm
+++ b/shell/platform/darwin/ios/framework/Source/FlutterPlatformViews.mm
@@ -18,6 +18,8 @@
 #import "flutter/shell/platform/darwin/ios/framework/Source/FlutterViewController_Internal.h"
 #import "flutter/shell/platform/darwin/ios/ios_surface.h"
 
+static const NSUInteger kFlutterClippingMaskViewPoolCapacity = 5;
+
 @implementation UIView (FirstResponder)
 - (BOOL)flt_hasFirstResponderInViewHierarchySubtree {
   if (self.isFirstResponder) {
@@ -451,6 +453,17 @@ int FlutterPlatformViewsController::CountClips(const MutatorsStack& mutators_sta
   return clipCount;
 }
 
+void FlutterPlatformViewsController::ClipViewAddMaskView(UIView* clipView) {
+  if (clipView.maskView) {
+    return;
+  }
+  UIView* flutterView = flutter_view_.get();
+  CGRect frame =
+      CGRectMake(-clipView.frame.origin.x, -clipView.frame.origin.y,
+                 CGRectGetWidth(flutterView.bounds), CGRectGetHeight(flutterView.bounds));
+  clipView.maskView = [mask_view_pool_.get() getMaskViewWithFrame:frame];
+}
+
 void FlutterPlatformViewsController::ApplyMutators(const MutatorsStack& mutators_stack,
                                                    UIView* embedded_view,
                                                    const SkRect& bounding_rect) {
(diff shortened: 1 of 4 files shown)


shell/platform/darwin/ios/framework/Source/FlutterPlatformViews_Internal.mm
```diff
@@ -447,3 +451,63 @@ - (void)clipPath:(const SkPath&)path matrix:(const SkMatrix&)matrix {
 }
 
 @end
+
+@interface FlutterClippingMaskViewPool ()
+
+@property(assign, nonatomic) NSUInteger capacity;
+@property(retain, nonatomic) NSMutableArray<FlutterClippingMaskView*>* pool;
```
jmagman, line 458: Would this retain the `FlutterClippingMaskView` even after the `ChildClippingView` is dealloc'd?
Would `NSPointerArray` with `NSPointerFunctionsWeakMemory` (I think zeroes it out with MRC but worth testing) work better here?  If you could reliably know the mask views in the pool are used (because they are retained), then you could get rid of `recycleMaskViews` and during insert  `compact`, then insert at `count` if there was capacity.

shell/platform/darwin/ios/framework/Source/FlutterPlatformViews_Internal.h
```diff
@@ -328,6 +353,7 @@ class FlutterPlatformViewsController {
   fml::scoped_nsobject<FlutterMethodChannel> channel_;
   fml::scoped_nsobject<UIView> flutter_view_;
   fml::scoped_nsobject<UIViewController> flutter_view_controller_;
+  fml::scoped_nsobject<FlutterClippingMaskViewPool> mask_view_pool_;
```
jmagman, line 356: Is there a reason to make this a `scoped_nsobject`?

shell/platform/darwin/ios/framework/Source/FlutterPlatformViews_Internal.mm
jmagman: Well never mind, retaining it after the `ChildClippingView` is dealloc'd is the whole point of this PR...
Can you confirm that deallocating `FlutterPlatformViewsController` releases the pool and the mask views?

shell/platform/darwin/ios/framework/Source/FlutterPlatformViews_Internal.h
cyanglaz: No, I should change it to properties. I was blindly following the current pattern.

shell/platform/darwin/ios/framework/Source/FlutterPlatformViews_Internal.mm
```diff
@@ -447,3 +451,64 @@ - (void)clipPath:(const SkPath&)path matrix:(const SkMatrix&)matrix {
...
+    _capacity = capacity;
+    _availableIndex = 0;
+  }
+  return self;
+}
+
+- (FlutterClippingMaskView*)getMaskViewWithFrame:(CGRect)frame {
+  FML_DCHECK(self.availableIndex <= self.capacity);
+  FlutterClippingMaskView* maskView;
+  if (self.availableIndex == self.capacity) {
```
hellohuanlin, line 478: it looks like the `capacity` is not updated? should it be increased when you add a new one?

hellohuanlin: we can probably just compare with `pool.count`, and do not keep this capacity ivar?

shell/platform/darwin/ios/framework/Source/FlutterPlatformViews.mm
```diff
@@ -451,6 +453,17 @@ static bool ClipRRectContainsPlatformViewBoundingRect(const SkRRect& clip_rrect,
   return clipCount;
 }
 
+void FlutterPlatformViewsController::ClipViewAddMaskView(UIView* clipView) {
```
hellohuanlin, line 456: supernit: `ClipViewSetMaskView`?

hellohuanlin (commented):
Just a question on the pool capacity.

shell/platform/darwin/ios/framework/Source/FlutterPlatformViews_Internal.mm
cyanglaz: The capacity is a static limit set for the pool. My intention is to limit the memory used by the pool.

shell/platform/darwin/ios/framework/Source/FlutterPlatformViews_Internal.h
cyanglaz: Oh, actually, it is in a c++ class. So I think scoped_nsobject is better that the lifecycle is automatically managed.

jmagman: Which class is c++?  Your new `FlutterClippingMaskViewPool` is a Obj-C class.

cyanglaz: FlutterPlatformViewsController is the c++ class.

shell/platform/darwin/ios/framework/Source/FlutterPlatformViews_Internal.mm
hellohuanlin: gotcha. I mistakenly thought that this capacity increases like an NSMutableArray.

hellohuanlin (approved):

shell/platform/darwin/ios/framework/Source/FlutterPlatformViews_Internal.mm
cyanglaz: Ah, yeah it could be confusing, I added a comment.

shell/platform/darwin/ios/framework/Source/FlutterPlatformViews.mm
cyanglaz: done

jmagman (approved):

PR #39005, opened 2023-01-19
Shader analysis with malioc

zanderso:
This PR adds logic to the GN build to analyze shaders and output the analysis results in the build output directory unconditionally when the build is configured with the path to `malioc` using the `--malioc-path` flag passed to the `gn` script. `malioc` is quite fast, so this does not affect build times.

This PR also adds a script `impeller/tools/malioc_diff.py` that detects before/after changes in the results of `malioc`'s analysis, or generate a analysis results golden file. In a subsequent PR, the plan is to check in such a golden file, and then compare shader analysis results on PRs/commits against the golden file in CI.

### DIFF
--- a/impeller/tools/malioc.gni
+++ b/impeller/tools/malioc.gni
@@ -1,0 +1,124 @@
+# Copyright 2013 The Flutter Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+import("//build/compiled_action.gni")
+import("//flutter/common/config.gni")
+import("//flutter/testing/testing.gni")
+
+declare_args() {
+  # Path to the Mali offline compiler tool 'malioc'.
+  impeller_malioc_path = ""
+
+  impeller_malioc_cores = []
+}
+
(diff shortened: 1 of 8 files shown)


jonahwilliams (approved):
how does this handle the SSBO shaders?

LGTM

impeller/tools/malioc.gni
```diff
@@ -0,0 +1,124 @@
...
+                rebase_path(core_list_file),
+              ])
+  _mali_cores = read_file(core_list_file, "json")
+  foreach(mali_core, _mali_cores.cores) {
+    impeller_malioc_cores += [ mali_core.core ]
+  }
+}
+
+template("malioc_analyze_shaders") {
+  # TODO(zra): Check that gles_language_version is in the supported set. For now
```
zanderso, line 34: @jonahwilliams ssbos requires a higher language version, so when any language version is set, analysis is skipped. This TODO is for tightening this logic to match what the analyzer supports.

jonahwilliams: makes sense!

PR #39019, opened 2023-01-20
Revert "Extract WideToUTF16String/UTF16StringToWide to FML"

godofredoc:
Reverts flutter/engine#39006


fluttergithubbot (approved):

cbracken:
LGTM thanks for reverting. This was caused by a race between two PRs: the one being reverted which removed the old function and another one that landed which added a new use of it. I've relanded this patch with the fix.

PR #39023, opened 2023-01-20
[Impeller] Check the correct stencil coverage when deciding whether to elide a restore

bdero:
Just noticed this while carefully reviewing the stencil coverage stack again today. This is much more rare/a subset of the original issue that was fixed by https://github.com/flutter/engine/pull/38964, but could still result in lots of draw calls getting culled. Slightly modified the playground to detect this case too.

Before
Screenshot 2023-01-19 at 5 49 35 PM

After
Screenshot 2023-01-19 at 5 51 42 PM

### DIFF
--- a/impeller/aiks/aiks_unittests.cc
+++ b/impeller/aiks/aiks_unittests.cc
@@ -278,11 +278,11 @@ TEST_P(AiksTest, CanRenderWithContiguousClipRestores) {
 
   canvas.Save();
 
-  // Append two clips. First with empty coverage.
+  // Append two clips, the second resulting in empty coverage.
   canvas.ClipPath(
       PathBuilder{}.AddRect(Rect::MakeXYWH(100, 100, 100, 100)).TakePath());
   canvas.ClipPath(
-      PathBuilder{}.AddRect(Rect::MakeXYWH(100, 100, 100, 100)).TakePath());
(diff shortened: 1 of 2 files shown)


chinmaygarde (approved):

bdero:
Correction: This wouldn't result in excess draw calls getting _culled_ -- it actually results in some stuff getting inverse clipped by anything above the restored value because the stencil check uses strict equality.

Mercury really needs to be in retrograde for an app to be affected by this, so it's not as much of an emergency, but could cause big weird problems for the poor soul that runs into it:
1. We can only hit this if the stencil stack prior to restore happened to have empty coverage, and
2. just like the original bug, multiple clips need to be getting restored by one restore entity.

bdero:
@zanderso Should we do a cherry pick for this one too? This isn't present on gallery/wondrous, but I won't say it doesn't worry me.

zanderso:
@bdero Yeah, we can. I only hadn't because I interpreted https://github.com/flutter/engine/pull/39023#issuecomment-1397857753 as saying that it wasn't worth a CP. I'll draw up the issue.

bdero:
@zanderso Makes sense, thanks. When this came across my mind today I realized this is doesn't seem terribly rare. _Any_ restore from empty stencil coverage will break subsequent draws (including clips). The stencil buffer will basically be in a bad state that will only recover the next time a restore happens from non-empty stencil coverage.

PR #39038, opened 2023-01-20
[impeller] change input order in ColorFilterContents::MakeBlend

jonahwilliams:
Fixes https://github.com/flutter/flutter/issues/118903

Both pipeline and advanced blends internally used the order dst, src, src, however the ColorFilterContents::MakeFilter would jumble the order and the only usage in the engine compensated for this.

(Of course, still need to double check)

### DIFF
--- a/impeller/entity/contents/filters/blend_filter_contents.cc
+++ b/impeller/entity/contents/filters/blend_filter_contents.cc
@@ -159,7 +159,7 @@ static std::optional<Snapshot> PipelineBlend(
   using VS = BlendPipeline::VertexShader;
   using FS = BlendPipeline::FragmentShader;
 
-  auto input_snapshot = inputs[0]->GetSnapshot(renderer, entity);
+  auto dst_snapshot = inputs[0]->GetSnapshot(renderer, entity);
 
   ContentContext::SubpassCallback callback = [&](const ContentContext& renderer,
                                                  RenderPass& pass) {
@@ -213,7 +213,7 @@ static std::optional<Snapshot> PipelineBlend(
     // Draw the first texture using kSource.
     options.blend_mode = BlendMode::kSource;
     cmd.pipeline = renderer.GetBlendPipeline(options);
-    if (!add_blend_command(input_snapshot)) {
+    if (!add_blend_command(dst_snapshot)) {
       return true;
     }
 
@@ -225,8 +225,8 @@ static std::optional<Snapshot> PipelineBlend(
 
       for (auto texture_i = inputs.begin() + 1; texture_i < inputs.end();
            texture_i++) {
-        auto input = texture_i->get()->GetSnapshot(renderer, entity);
-        if (!add_blend_command(input)) {
+        auto src_input = texture_i->get()->GetSnapshot(renderer, entity);
+        if (!add_blend_command(src_input)) {
           return true;
         }
       }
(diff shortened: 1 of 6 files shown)


impeller/entity/contents/filters/blend_filter_contents.cc
```diff
@@ -159,7 +159,7 @@ static std::optional<Snapshot> PipelineBlend(
   using VS = BlendPipeline::VertexShader;
   using FS = BlendPipeline::FragmentShader;
 
-  auto input_snapshot = inputs[0]->GetSnapshot(renderer, entity);
+  auto dst_snapshot = inputs[0]->GetSnapshot(renderer, entity);
```
jonahwilliams, line 162: No change here, just renaming a bit for clarity

impeller/entity/contents/filters/color_filter_contents.cc
```diff
@@ -37,7 +37,7 @@ std::shared_ptr<ColorFilterContents> ColorFilterContents::MakeBlend(
   std::shared_ptr<BlendFilterContents> new_blend;
   for (auto in_i = inputs.begin() + 1; in_i < inputs.end(); in_i++) {
     new_blend = std::make_shared<BlendFilterContents>();
-    new_blend->SetInputs({*in_i, blend_input});
+    new_blend->SetInputs({blend_input, *in_i});
```
jonahwilliams, line 40: This was the bug. We start with input 0, and input 1, and but we shifted input 1 and input 0. This was worked around in entity_pass.cc.

impeller/entity/entity_unittests.cc
```diff
@@ -856,7 +856,7 @@ TEST_P(EntityTest, Filters) {
 
     auto blend1 = ColorFilterContents::MakeBlend(
         BlendMode::kScreen,
-        {fi_bridge, FilterInput::Make(blend0), fi_bridge, fi_bridge});
+        {FilterInput::Make(blend0), fi_bridge, fi_bridge, fi_bridge});
```
jonahwilliams, line 859: This changes the output of this test. I think that is what we want, the way the previous algorithm worked for 2+ inputs was given the following filter inputs:

N0 N1 N2 N3

First N1 becomes dst and N0 becomes src to produce F1
Then N2 becomes dst and F1 becomes src to produce F2
Then N3 becomes dst and F2 becomes src to produce F3

Now this is changed (and this output looks different), first

N0 becomes dst and N1 becomes src to produce F1
F1 becomes dst and N2 becomes src to produce F2
F2 becomes dst and N3 becomes src to produce F3.

I think the latter is more intuitive initially, but the former also sort of makes sense as a kind of polish notation (or reverse, I don't remember which). But of course this ends up with inconsistencies for the most common 2 input case since that works differently if you're just using the  pipeline blends.

jonahwilliams: Maybe we could retain the old behavior if desired with a new constructor, since you can't really express one in terms of the other directly

impeller/entity/contents/filters/color_filter_contents.cc
jonahwilliams: (This re-arranging doesn't happen for pipeline blends which exit above. We could also change this code to not do the re-arranging for advanced blends with 2 inputs, but that has consistency implications

impeller/entity/entity_unittests.cc
bdero: Since we don't rely on more then 2 inputs anywhere right now (I think the filter graph has bigger dreams, so maybe someday we will), changing the behavior now SGTM. I agree the latter seems more intuitive.

bdero (approved):
Assuming `AiksTest.ColorWheel` (which applies the blend via a savelayer) has identical behavior prior to this fix, LGTM!

jonahwilliams:
Looks good so far

PR #39077, opened 2023-01-23
[Impeller] uniform offsets account for size

jonahwilliams:
Fixes https://github.com/flutter/flutter/issues/118847

When reading from the uniform buffer account for vec2/3/4 uniforms being larger than a single float.

### DIFF
--- a/impeller/entity/entity_unittests.cc
+++ b/impeller/entity/entity_unittests.cc
@@ -2097,11 +2097,11 @@ TEST_P(EntityTest, RuntimeEffect) {
     contents->SetRuntimeStage(runtime_stage);
 
     struct FragUniforms {
-      Scalar iTime;
       Vector2 iResolution;
+      Scalar iTime;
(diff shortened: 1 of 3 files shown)


impeller/entity/entity_unittests.cc
```diff
@@ -2097,11 +2097,11 @@ TEST_P(EntityTest, RuntimeEffect) {
     contents->SetRuntimeStage(runtime_stage);
 
     struct FragUniforms {
-      Scalar iTime;
       Vector2 iResolution;
```
jonahwilliams, line 2100: By re-arranging these uniforms we can observe that the shader does not work without this fix. Thus, tested.

jonahwilliams: this doesn't animate without the fix, because the first uniform gets data in 0, 1, and the second uniform gets the fixed data in 1 instead of at 2.

PR #39103, opened 2023-01-24
Get the vulkan backend working again

iskakaushik:
* Disables modern features as SSBOs are wired on Vulkan yet.
* Sub-optimal presentation mode must not be a fatal issue.
* Allow for shader outputs not to be consumed.

### DIFF
--- a/impeller/tools/impeller.gni
+++ b/impeller/tools/impeller.gni
@@ -18,7 +18,7 @@ declare_args() {
   impeller_enable_opengles = is_mac || is_linux || is_win || is_android
 
   # Whether the Vulkan backend is enabled.
-  impeller_enable_vulkan = is_linux || is_android
+  impeller_enable_vulkan = is_mac || is_linux || is_android
 
   # Whether to use a prebuilt impellerc.
   # If this is the empty string, impellerc will be built.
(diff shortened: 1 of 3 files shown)


impeller/tools/impeller.gni
```diff
@@ -18,7 +18,7 @@ declare_args() {
   impeller_enable_opengles = is_mac || is_linux || is_win || is_android
 
   # Whether the Vulkan backend is enabled.
-  impeller_enable_vulkan = is_linux || is_android
+  impeller_enable_vulkan = is_mac || is_linux || is_android
```
bdero, line 21: Is the Vulkan playground harness configured to use the SwiftShader ICD by default on MacOS? If not, might be worth adding a doc about setting up Vulkan for MacOS [here](https://github.com/flutter/engine/tree/main/impeller/docs) (getting LunarG SDK, environment vars for MoltenVK, etc). FYI @chinmaygarde

bdero (approved):
LGTM

impeller/tools/impeller.gni
iskakaushik: added a issue https://github.com/flutter/flutter/issues/119083

PR #39104, opened 2023-01-24
[Impeller] Switch to nearest sampling for the text atlas

bdero:
Resolves https://github.com/flutter/flutter/issues/118613.

* Remove residual position offsets. These were used to help mitigate the infamous glyph overlap problem, which has since been solved.
* Always pixel snap the source position.
* Switch from linear sampling to nearest sampling.

Before (top), after (bottom):
Screenshot 2023-01-24 at 12 21 11 PM

Before:
Screen Shot 2023-01-24 at 12 14 44 PM

After:
Screen Shot 2023-01-24 at 12 17 31 PM

### DIFF
--- a/impeller/entity/shaders/glyph_atlas.vert
+++ b/impeller/entity/shaders/glyph_atlas.vert
@@ -26,7 +26,8 @@ void main() {
   gl_Position = IPPositionForGlyphPosition(
       frame_info.mvp, unit_position, destination_position, destination_size);
   v_unit_position = unit_position;
-  v_source_position = source_position;
+  // Pixel snap the source (sampling) start position.
+  v_source_position = round(source_position);
   v_source_glyph_size = source_glyph_size;
   v_has_color = has_color;
 }
--- a/impeller/entity/contents/text_contents.cc
+++ b/impeller/entity/contents/text_contents.cc
@@ -78,8 +78,8 @@ static bool CommonRender(
   VS::BindFrameInfo(cmd, pass.GetTransientsBuffer().EmplaceUniform(frame_info));
 
   SamplerDescriptor sampler_desc;
-  sampler_desc.min_filter = MinMagFilter::kLinear;
-  sampler_desc.mag_filter = MinMagFilter::kLinear;
+  sampler_desc.min_filter = MinMagFilter::kNearest;
+  sampler_desc.mag_filter = MinMagFilter::kNearest;
   sampler_desc.mip_filter = MipFilter::kNone;
 
   typename FS::FragInfo frag_info;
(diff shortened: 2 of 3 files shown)


bdero:
Unfortunately, given the nature of using an atlas, rotated text still looks crunchy.

image

Possible ways to solve:
* Render larger glyphs to the atlas and scale them down.
* Turn on linear sampling whenever the transform contains stuff other than a translation.
* Render the fully transformed glyphs to the atlas.

jonahwilliams:
I have some thoughts!

> Easiest solution: Turn on linear sampling whenever the transform contains stuff other than a translation. Downside: Text is slightly blurry when rotated.

This seems like a good approach at least for now, though hopefully this doesn't need to be done for all text? I'm assuming this won't make us any worse off than we are today.

> Render larger glyphs to the atlas and scale them down. Downside: We 4x the atlas area.

I'd be really skeptical about any sort of supersampling techniques. Populating the glyph atlas is already very expensive, I'd actually be less worried about the memory usage.

> Render the fully transformed glyphs to the atlas. Downside: Reduced cacheability of atlases.

Since rotated text is actually quite rare, I don't think this would be the worst thing. Though we'd go through a lot of churn in the cases where this is animated. You might want something like incremental appends first : https://github.com/flutter/engine/pull/38253

There are a few places in the framework where this happens, like the cupertino picker thingy.

Separately:

* Does rendering rotated text work better if we're using a signed distance field approach?
* What does Skia do?

impeller/entity/shaders/glyph_atlas.vert
```diff
@@ -26,7 +26,8 @@ void main() {
   gl_Position = IPPositionForGlyphPosition(
       frame_info.mvp, unit_position, destination_position, destination_size);
   v_unit_position = unit_position;
-  v_source_position = source_position;
+  // Pixel snap the source (sampling) start position.
+  v_source_position = round(source_position);
```
chinmaygarde, line 30: ~Shouldn't this be floor instead of round?~

Eh, nevermind.

chinmaygarde (approved):
> Turn on linear sampling whenever the transform contains stuff other than a translation.

This seems like the best solution for now. But I'd be curious to see what the Skia behavior is (after we land this).

> Does rendering rotated text work better if we're using a signed distance field approach?

We'll have more options but not all text can be rendered using SDF.

jonahwilliams:
Right, bitmap emojis and colr fonts.

bdero:
> This seems like a good approach at least for now, though hopefully this doesn't need to be done for all text? I'm assuming this won't make us any worse off than we are today.

SGTM, throwing together a solution for this in a follow-up. We can avoid doing this whenever text is scale/translation only, which is the most common case.

> Does rendering rotated text work better if we're using a signed distance field approach?

Yes. For SDF, we always we linear sample the distance to determine if we're inside/outside the shape and then perform AA ourselves based on the distance. This problem only exists if the AA is pre-rendered by something else and we need to somewhat preserve it. (On this note, we could also turn off AA to lessen the blurriness being added by the non-pixel aligned linear sampling, but this would still have oscillating crunchiness that weaves in an out with the pixel alignment)

> What does Skia do?

I'm not sure if Skia caches things to intermediary atlases and/or SDFs, but what I do know for sure is that it _must_ do path rendering as a first step for TTFs, since TTF glyphs are just paths!

PR #39109, opened 2023-01-24
`HtmlImage.toByteData`: handle blob rawUnmodified format / add png support

maRci002:
Resolves https://github.com/flutter/flutter/issues/119065.

This PR adds the following features:

- If `DomHTMLImageElement.src` is `blob` and `format = ui.ImageByteFormat.rawUnmodified` then we can restore original bytes
- `ui.ImageByteFormat.png` format is handled properly

## Pre-launch Checklist

- [x] I signed the [CLA].


lib/web_ui/lib/src/engine/html_image_codec.dart
```diff
@@ -194,16 +194,54 @@ class HtmlImage implements ui.Image {
...
+        }
+
+        final DomCanvasElement canvas = createDomCanvasElement()
+          ..width = width.toDouble()
+          ..height = height.toDouble();
+        final DomCanvasRenderingContext2D ctx = canvas.context2D;
+        ctx.drawImage(imgElement, 0, 0);
+        // canvas.toDataURL defaults to: 'image/png'
+        final String dataUrl = canvas.toDataURL();
+        return _byteDataFromDataString(dataUrl);
```
mdebbar, line 209: This might be useful for CanvasKit as well. cc @yjbanov

yjbanov: Yep, but I'm OK taking one step at a time.

lib/web_ui/lib/src/engine/html_image_codec.dart
```diff
@@ -194,16 +198,44 @@ class HtmlImage implements ui.Image {
...
+          return _byteDataFromDataString(imgElement.src!);
+        }
+
+        final DomCanvasElement canvas = createDomCanvasElement()
+          ..width = width.toDouble()
+          ..height = height.toDouble();
+        final DomCanvasRenderingContext2D ctx = canvas.context2D;
+        ctx.drawImage(imgElement, 0, 0);
+        // canvas.toDataURL defaults to: 'image/png'
+        final String dataUrl = canvas.toDataURL();
```
yjbanov, line 212: After reading the image data back, set `canvas.width` and `canvas.height` to zero. Safari on iOS will crash if the GC doesn't collect 2D canvases fast enough. Setting the image size to zero causes Safari to free internal image memory immediately.

yjbanov (changes requested):

Hixie:
@maRci002 Will you be able to continue working on this? It's totally fine if not; we can take it over (though it might take longer). I'm just asking because I'm trying to clean out our review queue.

maRci002:
> @maRci002 Will you be able to continue working on this? It's totally fine if not; we can take it over (though it might take longer). I'm just asking because I'm trying to clean out our review queue.

Unfortunately I don't have any free time in the near future.

I think this part of the PR is invalid, since it should return raw decoded bytes and not the encoded ones.

> * If `DomHTMLImageElement.src` is `blob` and `format = ui.ImageByteFormat.rawUnmodified` then we can restore original bytes

PR #39111, opened 2023-01-25
[Impeller] Adds wide gamut support for iOS.

gaaclarke:
issue https://github.com/flutter/flutter/issues/117182
Integration test: https://github.com/flutter/flutter/pull/119657

What is yet to be implemented:
1) 32bit per pixel wide gamut images with no opacity
1) SaveLayer support
1) Gradient support
1) On by default.

## Pre-launch Checklist

- [x] I signed the [CLA].


shell/platform/darwin/ios/ios_surface_metal_impeller.mm
```diff
@@ -4,8 +4,75 @@
...
+namespace {
+using std::shared_ptr;
+using namespace impeller;
+
+// This appears to be the only safe way to override the
+// GetColorAttachmentPixelFormat method.  It is assumed in the Context that
+// there will be one pixel format for the whole app which is not true.  So, it
+// is unsafe to mutate the Context and you cannot clone the Context at this
+// level since the Context does not safely manage the MTLDevice.
+class CustomColorAttachmentPixelFormatContext : public Context {
```
gaaclarke, line 20: I think we should refactor this in a different PR.  It isn't clear to me the best way to change it though.

gaaclarke:
I'm not quite sure what happened with the scenario_app tests.  `-[RenderingSelectionTest testMetalRendering]` for example runs fine for me locally.  I'll try rerunning.

shell/platform/darwin/ios/framework/Source/FlutterView.mm
```diff
@@ -16,8 +16,16 @@
 #import "flutter/shell/platform/darwin/ios/ios_surface_software.h"
 #include "third_party/skia/include/utils/mac/SkCGUtils.h"
 
+static BOOL IsWideGamutSupported() {
+  // This predicates the decision on the capabilities of the iOS device's
+  // display.  This means external displays will not support wide gamut if the
+  // device's display doesn't support it.  It practice that should be never.
+  return UIScreen.mainScreen.traitCollection.displayGamut != UIDisplayGamutSRGB;
```
bdero, line 23: What will happen when moving the window to a monitor that doesn't support wide gamut?

bdero (commented):
The layer settings in `layoutSubviews` seem good to me. Mainly code nits and questions.

shell/platform/darwin/ios/framework/Source/FlutterView.mm
gaaclarke: I believe the extended range values will be clamped in the driver.  So it is technically wasteful if your display doesn't support wide gamut but it doesn't require handling operating system callbacks about transitioning surfaces from one display to another.  We don't have a final cost value for the wide gamut support since I'm still working on optimizations, but I think this is the right tradeoff with the information we have.

shell/platform/darwin/ios/ios_surface_metal_impeller.mm
```diff
@@ -4,8 +4,75 @@
...
+ public:
+  CustomColorAttachmentPixelFormatContext(const shared_ptr<Context>& context,
+                                          PixelFormat color_attachment_pixel_format)
+      : context_(context), color_attachment_pixel_format_(color_attachment_pixel_format) {}
+
+  virtual bool IsValid() const override { return context_->IsValid(); }
```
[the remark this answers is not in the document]
gaaclarke: Done, fwiw omitting the `virtual` has no change in the generated code.  This is just syntactic sugar, right?

shell/platform/darwin/ios/ios_surface_metal_impeller.mm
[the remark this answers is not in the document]
gaaclarke: done

shell/platform/darwin/ios/ios_surface_metal_impeller.mm
```diff
@@ -4,8 +4,75 @@
 
 #import "flutter/shell/platform/darwin/ios/ios_surface_metal_impeller.h"
 
+#include "flutter/impeller/renderer/backend/metal/formats_mtl.h"
+#include "flutter/impeller/renderer/context.h"
 #include "flutter/shell/gpu/gpu_surface_metal_impeller.h"
 
+namespace {
+using std::shared_ptr;
+using namespace impeller;
```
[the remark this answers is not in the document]
gaaclarke: done

shell/common/rasterizer.cc
```diff
@@ -711,22 +711,32 @@ Rasterizer::Screenshot Rasterizer::ScreenshotLastLayerTree(
...
 
   sk_sp<SkData> data = nullptr;
+  std::string format;
 
   GrDirectContext* surface_context =
       surface_ ? surface_->GetContext() : nullptr;
 
   switch (type) {
     case ScreenshotType::SkiaPicture:
+      format = "ScreenshotType::SkiaPicture";
```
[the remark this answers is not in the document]
gaaclarke: I'm not sure how they are used.  I suspect they are strings since the values are read in dart code.  That way we don't have to worry about synchronizing enums values across languages.

shell/gpu/gpu_surface_metal_impeller.mm
```diff
@@ -129,4 +133,49 @@
   return aiks_context_.get();
 }
 
+Surface::SurfaceData GPUSurfaceMetalImpeller::GetSurfaceData() const {
+  if (last_drawable_) {
+    id<CAMetalDrawable> metal_drawable;
+    if ([last_drawable_ conformsToProtocol:@protocol(CAMetalDrawable)]) {
+      metal_drawable = static_cast<id<CAMetalDrawable>>(last_drawable_);
+    } else {
```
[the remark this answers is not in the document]
gaaclarke: done

shell/gpu/gpu_surface_metal_impeller.mm
```diff
@@ -129,4 +133,49 @@
   return aiks_context_.get();
 }
 
+Surface::SurfaceData GPUSurfaceMetalImpeller::GetSurfaceData() const {
+  if (last_drawable_) {
```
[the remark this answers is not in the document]
gaaclarke: done

shell/platform/darwin/ios/framework/Source/FlutterEngine.mm
```diff
@@ -633,6 +634,37 @@ - (void)setupChannels {
...
+            result(@[ width, height, format, data ]);
+          } else {
+            result([FlutterError errorWithCode:@"failure"
+                                       message:@"Unable to get screenshot."
+                                       details:nil]);
+          }
+        } else {
+          result([FlutterError errorWithCode:@"invalid_state"
+                                     message:@"Requesting screenshot while engine is not running."
+                                     details:nil]);
```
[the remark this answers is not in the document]
gaaclarke: Done.  My concern here is being able to easily look at the code and make sure all branches call the asynchronous handler before exiting.  When you have early exits for async functions someone can easily forget calling the callback or interrupting execution.  I've tried to minimize that risk by putting `return` before all callback invocations.

gaaclarke:
(rebasing to hopefully flush out infra errors)

shell/platform/darwin/ios/ios_surface_metal_impeller.mm
```diff
@@ -35,9 +100,10 @@
 
 // |IOSSurface|
 std::unique_ptr<Surface> IOSSurfaceMetalImpeller::CreateGPUSurface(GrDirectContext*) {
-  return std::make_unique<GPUSurfaceMetalImpeller>(this,              //
-                                                   impeller_context_  //
-
+  auto context = std::make_shared<CustomColorAttachmentPixelFormatContext>(
+      impeller_context_, FromMTLPixelFormat(layer_.get().pixelFormat));
```
bdero, line 104: Since this default attachment format is backend-specific, and different backends have different techniques for resolving them (see `ContextVK`, for example, which queries device capabilities when making the decision), prefer to override this behavior in `ContextMTL::GetColorAttachmentPixelFormat` directly instead of creating a wrapper.

This could be as simple as adding a `color_attachment_pixel_format_` field/setter in `ContextMTL`, overriding `GetColorAttachmentPixelFormat` to return it, and then calling the new setter here with a backend cast:

```cpp
ContextMTL::Cast(*impeller_context_).SetColorAttachmentPixelFormat(
    FromMTLPixelFormat(layer_.get().pixelFormat));
```

shell/platform/darwin/ios/framework/Source/FlutterView.mm
gaaclarke: Yea, I have to almost believe the display driver has to clamp the values but this is just speculation.  We should test it out.  We'll never be able to have an integration test for this unfortunately.  If it doesn't work this isn't the appropriate PR to address it.

Filed an issued: https://github.com/flutter/flutter/issues/120485

shell/platform/darwin/ios/ios_surface_metal_impeller.mm
```diff
@@ -4,10 +4,75 @@
...
+  shared_ptr<Context> context_;
+  PixelFormat color_attachment_pixel_format_;
+};
+}  // namespace
+}  // namespace impeller
+
 namespace flutter {
 
+using impeller::CustomColorAttachmentPixelFormatContext;
+using impeller::FromMTLPixelFormat;
```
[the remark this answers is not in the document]
gaaclarke: Are you referring to the Google C++ Style guide?

shell/platform/darwin/ios/ios_surface_metal_impeller.mm
gaaclarke: I can make that change 2.  It does however introduce a non-safe cast and a mutable ivar.  I kind of think the proxy value is better, then trying to refactor things so we can know the proper value at creation later.  Let's chat about it quickly.

gaaclarke:
Just chatted with @bdero:
1) we checked the google style guide regarding `using` directive rules
1) agreed that we need to test and make sure that we don't crash with an external monitor (going to follow up in a different issue, linked above)
1) said I'd look at potentially making the color attachment pixel format mutable if I can get the correct pointer without unsafe casts (although he was fine with them) otherwise we can go ahead with the proxy pattern which isn't perfect but there is hope that stuff will get reworked out as we chase down vulkan support and mixmatch surface/offscreen buffer pixel formats.

shell/platform/darwin/ios/ios_surface_metal_impeller.mm
```diff
@@ -4,10 +4,75 @@
 
 #import "flutter/shell/platform/darwin/ios/ios_surface_metal_impeller.h"
 
+#include "flutter/impeller/renderer/backend/metal/formats_mtl.h"
+#include "flutter/impeller/renderer/context.h"
 #include "flutter/shell/gpu/gpu_surface_metal_impeller.h"
 
+namespace impeller {
+namespace {
+using std::shared_ptr;
```
bdero, line 13: Nit: Remove this?

shell/platform/darwin/ios/ios_surface_metal_impeller.mm
```diff
@@ -4,10 +4,75 @@
 
 #import "flutter/shell/platform/darwin/ios/ios_surface_metal_impeller.h"
 
+#include "flutter/impeller/renderer/backend/metal/formats_mtl.h"
+#include "flutter/impeller/renderer/context.h"
 #include "flutter/shell/gpu/gpu_surface_metal_impeller.h"
 
+namespace impeller {
```
bdero, line 11: Nit: Could we remove the use of the Impeller namespace here, since symbols being defined here (`CustomColorAttachmentPixelFormatContext`) are not part of Impeller.

shell/platform/darwin/ios/ios_surface_metal_impeller.mm
gaaclarke: Yep, done.  I'm still mystified how this got here.  I wonder if my ide automatically added it.

shell/platform/darwin/ios/ios_surface_metal_impeller.mm
gaaclarke: Oh, looking into this I remember why we can't mutate the context.  Because the context is shared across multiple surfaces and surfaces may have different pixel formats.  So I can't mutate it, I have to make a new one.

shell/platform/darwin/ios/ios_surface_metal_impeller.mm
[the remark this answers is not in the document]
bdero: Discussed offline: The style guide seems to have an example of this kind of namespace stripping. These are all nits in general and shouldn't block stuff.

shell/platform/darwin/ios/ios_surface_metal_impeller.mm
bdero: Discussed offline: This case should be safe because we're working in the iOS surface, and we won't be supporting a non-Metal backend on iOS. I don't have a very strong opinion about which solution we land here, but I think the current solution introduces new code navigation friction (especially if we replicate this solution for other backends). A castless solution would be to add an override field on the abstract `Context` instead.

bdero (approved):

shell/platform/darwin/ios/ios_surface_metal_impeller.mm
gaaclarke: If I do that the code is going to explode.  There are tons of symbols from `::impeller` in it.  FWIW this class isn't really in `::impeller`.  It's in `::impeller::(anonymous namespace)`.  Once you are in an an anonymous namespace the other parts of your namespace are inconsequential except for having convenient access to symbols.  I could pull this out of the anonymous namespace and into its own file, but considering how we want to fix the issue that lead to us needing this, that's probably overklll. 

I've filed an issue to try to capture this: https://github.com/flutter/flutter/issues/120499

shell/platform/darwin/ios/ios_surface_metal_impeller.mm
bdero: > Oh, looking into this I remember why we can't mutate the context. Because the context is shared across multiple surfaces and surfaces may have different pixel formats. So I can't mutate it, I have to make a new one.

Ah, that makes sense. This might eventually happen if we have multiple windows on different monitors, right? The more I think about this, the more I suspect our best bet is to get rid of this this value from the context altogether and supply it when making the render call per-frame. The renderer already configures the attachments for the root pass using the pixel format on the drawable itself every frame, and we aught to do something similar for the offscreen textures instead of arbitrarily storing this value in the context where it doesn't belong. After all, the embedder is what's responsible for providing the surface every frame.

PR #39112, opened 2023-01-25
[Impeller] Linear sample atlas glyphs when the CTM isn't translation/scale only

bdero:
Follow-up to https://github.com/flutter/engine/pull/39104.

Before:
image

After:
image


jonahwilliams (approved):
LGTM!

PR #39113, opened 2023-01-25
Remove unnecessary null checks

goderbauer:

Dart 3 drops support for non-null safe code, so we can finally turn on the unnecessary_null_comparison lint and remove the unnecessary checks it flags.


testing/dart/color_test.dart
```diff
@@ -27,23 +27,27 @@ void main() {
...
+    Object? exception;
     try {
       const Color c = Color(1 << 1000000);
       final Paint p = Paint();
       p.color = c;
     } catch (e) {
-      expect(e != null, equals(true));
+      exception = e;
     }
+    expect(exception != null, equals(true));
```
goderbauer, line 50: I thought the original intend of these two tests was that there would be an exception, but rewriting the tests to actually assert that (as shown here) makes the test fail. Needs some more investigation...

goderbauer: Interestingly, on the web SDK side these tests are explicitly testing that you can create these with out-of-bounds values: 

https://github.com/flutter/engine/blob/bffb98352446ae620ea40dbef0e38562b24bde1e/lib/web_ui/test/ui/color_test.dart#L37-L47

Maybe that is the intend here after all?

goderbauer: Since the color constructor is just removing anything above `0xFFFFFFFF` I am going to assume that creating a color with out-of-bounds values should be legal and not throw. I'll adjust these tests accordingly. 

https://github.com/flutter/engine/blob/bffb98352446ae620ea40dbef0e38562b24bde1e/lib/ui/painting.dart#L108

lib/ui/platform_dispatcher.dart
```diff
@@ -1943,8 +1943,7 @@ class Locale {
   const Locale(
     this._languageCode, [
     this._countryCode,
-  ]) : assert(_languageCode != null),
-       assert(_languageCode != ''),
+  ]) : assert(_languageCode != ''),
```
cbracken, line 1946: I wonder whether `assert(_languageCode != '')` or `assert(_languageCode.isNotEmpty)` is more efficient? My guess is that string equals special cases the zero length check anyway, so probably no real difference and it's a debug-only check anyway.

lib/web_ui/lib/platform_dispatcher.dart
```diff
@@ -443,16 +443,14 @@ class Locale {
   const Locale(
     this._languageCode, [
     this._countryCode,
-  ])  : assert(_languageCode != null),
-        assert(_languageCode != ''),
+  ])  : assert(_languageCode != ''),
```
cbracken, line 446: Similar question to above.

lib/ui/platform_dispatcher.dart
goderbauer: @rmacnak-google Do you happen to know?

tools/licenses/lib/main.dart
```diff
@@ -98,7 +98,7 @@ class _RepositorySourceFile extends _RepositoryLicensedFile {
   @override
   Iterable<Assignment> assignLicenses() {
     _internalLicenses ??= determineLicensesFor(_contents, name, parent, origin: io.fullName);
-    final List<License> licenses = _internalLicenses!;
+    final List<License>? licenses = _internalLicenses;
```
cbracken, line 101: Oops. That's embarrassing. 🤦‍♂️

lib/ui/platform_dispatcher.dart
cbracken: I'd go with `isNotEmpty` from the readability point of view too.

goderbauer: Sounds good, since this PR is so giant, I'll send a follow-up PR with that fix. Thanks, @rmacnak-google.

goderbauer: Sounds good, since this PR is so giant, I'll send a follow-up PR with that fix. Thanks, @rmacnak-google.

cbracken (approved):

lib/ui/platform_dispatcher.dart
goderbauer: Oh, I guess we cannot make that change:

```
platform_dispatcher.dart:446:16: Error: Not a constant expression.
  ])  : assert(_languageCode.isNotEmpty),
               ^^^^^^^^^^^^^
```

PR #39124, opened 2023-01-25
[Impeller] Use the start/end tangent of contours to compute stroke contour cap normals

bdero:
Resolves https://github.com/flutter/flutter/issues/118945.

* Instead of using the direction from/to the start/end points of the contour for a cap normal, use the curve's analytical start/end tangent for perfect results.
* Append points to draw the `Cap::kButt` cap, since it may no longer align perfectly with the last line segment direction.
* This also fixes similar slightly wrong results for closed path shapes such as ovals when using `Cap::kButt`.

Before:

https://user-images.githubusercontent.com/919017/214541644-6f8b0414-4c50-4d50-aafe-1b100dace66e.mov

After:

https://user-images.githubusercontent.com/919017/214541675-207324f2-af59-4c21-bc14-68c720bfd0c8.mov

### DIFF
--- a/impeller/display_list/display_list_unittests.cc
+++ b/impeller/display_list/display_list_unittests.cc
@@ -109,11 +109,13 @@ TEST_P(DisplayListTest, CanDrawArc) {
   auto callback = [&]() {
     static float start_angle = 45;
     static float sweep_angle = 270;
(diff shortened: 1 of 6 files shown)


jonahwilliams (approved):
LGTM

PR #39035, opened 2023-01-20
Revert "Remove references to Observatory (#38919)"

bkonyi:
Reason for revert: breaking Flutter -> plugins roll

This reverts commit 88e61d8bd166a0f4c6d5387ddc2df95f2ff1359b.

PR #39139, opened 2023-01-25
Reland "Remove references to Observatory (#38919)"

bkonyi:
This reverts commit 5dd945442b230a2c0549ee6d798fda71e031a4a9.


bkonyi:
I reverted the changes to the public APIs and marked them as deprecated. I _think_ I caught them all, but I'm not sure where the public API boundaries are, so please let me know if I missed any.

jmagman:
What was breaking in the Flutter -> plugins roll? https://github.com/flutter/engine/pull/39035

shell/platform/android/io/flutter/embedding/engine/FlutterJNI.java
```diff
@@ -230,14 +230,28 @@ public boolean getIsSoftwareRenderingEnabled() {
...
+   * VM Service URI for the VM instance.
+   *
+   * <p>Its value is set by the native engine once {@link #init(Context, String[], String, String,
+   * String, long)} is run.
+   *
+   * @deprecated replaced by {@link #getVMServiceUri()}.
+   */
+  @Deprecated
+  @Nullable
   public static String getObservatoryUri() {
```
bkonyi, line 253: @jmagman renaming this method had broken the espresso plugin.

shell/platform/darwin/ios/framework/Source/FlutterDartVMServicePublisher.mm
```diff
@@ -83,18 +88,20 @@ - (void)publishServiceProtocolPort:(NSURL*)url {
...
-                               FlutterObservatoryPublisher.serviceName.UTF8String, registrationType,
-                               domain, NULL, htons(port), txtData.length, txtData.bytes,
-                               RegistrationCallback, NULL);
+                               FlutterDartVMServicePublisher.serviceName.UTF8String,
+                               registrationType, domain, NULL, htons(port), txtData.length,
+                               txtData.bytes, RegistrationCallback, NULL);
 
   if (err != 0) {
-    FML_LOG(ERROR) << "Failed to register observatory port with mDNS with error " << err << ".";
+    FML_LOG(ERROR) << "Failed to register Dart VM Service port with mDNS with error " << err << ".";
```
vashworth, line 104: Networking in iOS is a protected resource and therefore requires permissions, specifically NSBonjourServices. Without it, this will always error and show this error message to users.

In the Flutter tools, we add in NSBonjourServices to the Info.plist on build here: 
https://github.com/flutter/flutter/blob/c31856bc48596d251ade554c33a09184e69185d9/packages/flutter_tools/bin/xcode_backend.dart#L247-L285

To avoid errors, the tool should be updated first to include the new `_dartVmService._tcp` (and also keep the original `_dartobservatory._tcp`).

jmagman: This code is falling back to `_dartobservatory._tcp` with the `legacyRegistrationType` check below so I think it should still work. However, it shouldn't log any errors like this of or the iOS 14 one until both fail.

@vashworth's suggestion to add both to the tool first would also work.

jmagman (changes requested):
Requesting changes to remove the error logs until both the `_dartVmService._tcp` AND `_dartobservatory._tcp` checks fail.

shell/platform/darwin/ios/framework/Source/FlutterDartVMServicePublisher.mm
bkonyi: Updated to only output the error message after both registration attempts have failed.

shell/platform/android/io/flutter/embedding/engine/FlutterJNI.java
jmagman: Interesting... https://github.com/flutter/plugins/blob/7686be7ecbe1cc049423715b5e2d897cb63a43c1/packages/espresso/android/src/main/java/androidx/test/espresso/flutter/action/FlutterViewAction.java#L108

jmagman (approved):
LGTM
@vashworth were your concerns addressed?

vashworth:
> LGTM @vashworth were your concerns addressed?

Yup, LGTM too!

PR #39162, opened 2023-01-26
[Web][HTML] Add mirrored characters support for RTL languages

bleroux:
## Description

Handling RTL text in the HTML renderer was introduced by https://github.com/flutter/engine/pull/26811.

It seems that there is a missing piece related to mirrored characters (see https://www.w3.org/International/articles/inline-bidi-markup/uba-basics#mirroring).

This PR adds this feature by relying on the HTML “dir” attribute.

## Before/after

Here is a simplified sample (compared to the one in https://github.com/flutter/flutter/issues/108431) to illustrate the issue:

<details>
<summary>Code sample</summary>

```dart

void main() {
  runApp(const MyApp());
}

  const MyApp({Key? key}) : super(key: key);

  @override
    return const MaterialApp(
      title: 'Flutter Demo',
      home: MyHomePage(),
      debugShowCheckedModeBanner: false,
    );
  }
}

class MyHomePage extends StatelessWidget {
  const MyHomePage({Key? key}) : super(key: key);

  @override
    return const Scaffold(
      backgroundColor: Colors.deepPurple,
      body: Center(
        child: Text(
          "(1)",
          textDirection: TextDirection.rtl,
          style: TextStyle(fontSize: 40, fontWeight: FontWeight.bold, color: Colors.white),
        ),
      ),
    );
  }
}

```
</details

Before (using HTML Renderer):
Capture d’écran du 2023-01-26 16-44-58

After (using HTML Renderer):
Capture d’écran du 2023-01-26 16-45-48

## Implementation choice

@mdebbar
 I was not able to figure out if there is already some web_ui code to handle mirrored characters.
So the solution I choosed is to rely on the HTML dir attribute and apply it to “flt-span” which is created for each text fragment (adding it only when a text fragment direction is RTL).

## Related Issue

Fixes: https://github.com/flutter/flutter/issues/108431

## Tests

Adds 1 test.

### DIFF
--- a/lib/web_ui/lib/src/engine/text/canvas_paragraph.dart
+++ b/lib/web_ui/lib/src/engine/text/canvas_paragraph.dart
@@ -174,6 +174,9 @@ class CanvasParagraph implements ui.Paragraph {
         }
 
         final DomElement spanElement = domDocument.createElement('flt-span');
+        if (fragment.textDirection == ui.TextDirection.rtl) {
+          spanElement.setAttribute('dir', 'rtl');
+        }
         applyTextStyleToElement(element: spanElement, style: fragment.style);
         _positionSpanElement(spanElement, line, fragment);
 
--- a/lib/web_ui/test/text/canvas_paragraph_builder_test.dart
+++ b/lib/web_ui/test/text/canvas_paragraph_builder_test.dart
@@ -460,9 +460,8 @@ Future<void> testMain() async {
   // Set dir attribute for RTL fragments in order to let the browser
   // handle mirrored characters.
   test('Sets "dir" attribute for RTL fragment', () {
-    const double fontSize = 20.0;
     final EngineParagraphStyle style = EngineParagraphStyle(
-      fontSize: fontSize,
+      fontSize: 20.0,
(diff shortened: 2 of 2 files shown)


mdebbar (approved):
The PR looks good to me. Thanks for contributing this fix!

There are some golden triage issues (bug: https://bugs.chromium.org/p/skia/issues/detail?id=13907). Let's wait a bit to see what the Skia team says. I'll triage them later.

bleroux:
> The PR looks good to me. Thanks for contributing this fix!
> 
> There are some golden triage issues (bug: https://bugs.chromium.org/p/skia/issues/detail?id=13907). Let's wait a bit to see what the Skia team says. I'll triage them later.

Great! I was trying to figure out if my PR broke those golden and noticed all golden add this message 'Computing closest positive and negative image. Check back later.'. It might be related to the Tree being currently broken and some golden are stuck?

mdebbar:
> It might be related to the Tree being currently broken and some golden are stuck?

It's not related to tree status. This is a bug in the Gold system where some images lose their triage status after a while.

hterkelsen:
LGTM

PR #39174, opened 2023-01-26
[Impeller] support overlapping stops with SSBO gradients.

jonahwilliams:
Makes the gradient shaders more expensive, but fully supports overlapping stops and reduces CPU workload. This can be clawed back later via shader variants.

Fixes https://github.com/flutter/flutter/issues/119175

overlapping


chinmaygarde (approved):
I don't have a framework for reasoning about how much more expensive this is. But seems like a fine tradeoff.

jonahwilliams:
@zanderso when's malioc?

zanderso:
> @zanderso when's malioc?

If you have a local install you can follow the instructions on the PR here https://github.com/flutter/engine/pull/39005. As for when we can get it in CI, that is going to be possible, but there's some paperwork to do that I need to prioritize.

PR #39198, opened 2023-01-27
[impellerc] remove remaining remap sampler support

jonahwilliams:
Usage removed in https://github.com/flutter/flutter/pull/119346


jonahwilliams:
I don't believe I need a test exception because this is a delete-only PR

zanderso (approved):

zanderso:
> I don't believe I need a test exception because this is a delete-only PR

Is this a bug in the bot?

jonahwilliams:
bug and/or limitation 😄

zanderso:
Removing the flag from building the shaders for the text fixture is also updating the test.

PR #36279, opened 2022-09-20
[Impeller] Defer applying opacity when saving layer

ColdPaleLight:
fix issue: https://github.com/flutter/flutter/issues/112175

1. Snapshot adds an `opacity` parameter
2. ImageFilter (Include ImageFilterColorFilter) will pass opacity, and ColorFilter will absorb opacity.
3. The opacity needs to be applied lazily only in save layer scenarios.

#### Test: SaveLayerWithColorMatrixFiltersAndAlphaDrawCorrectly

https://user-images.githubusercontent.com/31977171/192499376-816a7982-ea9d-4f0a-b857-d25cc45ca108.mp4

#### Test: SaveLayerWithBlendFiltersAndAlphaDrawCorrectly

https://user-images.githubusercontent.com/31977171/192499395-0dc5e5b4-a519-4522-9474-b19664963763.mp4

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [Flutter Style Guide] and the [C++, Objective-C, Java style guides].
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I added new tests to check the change I am making or feature I am adding, or Hixie said the PR is test-exempt. See [testing the engine] for instructions on
writing and running engine tests.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I signed the [CLA].
- [x] All existing and new tests are passing.

PR #39237, opened 2023-01-29
[Impeller] let images with opacity and filters passthrough.

jonahwilliams:
Fixes https://github.com/flutter/flutter/issues/119474

I'm not actually sure if this is correct! But this improves wonderous frame times on the wonders screen by ~20 FPS on the iPad pro.


jonahwilliams:
Need to double check but I think specifically for images we document the opacity as applying at the end so it may be OK

chinmaygarde (approved):

impeller/aiks/paint.cc
```diff
@@ -107,4 +107,8 @@ std::shared_ptr<FilterContents> Paint::MaskBlurDescriptor::CreateMaskBlur(
                                             effect_transform);
 }
 
+bool Paint::HasFilters() const {
+  return color_filter.has_value();
```
bdero, line 111: Why color filters specifically? Not all color filters absorb opacity IIRC.

impeller/aiks/paint.h
```diff
@@ -106,6 +106,11 @@ struct Paint {
   std::shared_ptr<Contents> CreateContentsForGeometry(
       std::unique_ptr<Geometry> geometry) const;
 
+  /// @brief   Whether this paint has a filter that can apply opacity
+  ///
+  ///          Only color filters can absorb opacity currently.
+  bool HasFilters() const;
```
bdero, line 112: This just checks for a color filter, not any filter. So perhaps name `HasColorFilter` instead?

impeller/aiks/canvas.cc
```diff
@@ -306,6 +306,7 @@ void Canvas::DrawImageRect(const std::shared_ptr<Image>& image,
   contents->SetSourceRect(source);
   contents->SetSamplerDescriptor(std::move(sampler));
   contents->SetOpacity(paint.color.alpha);
+  contents->SetDeferApplyingOpacity(paint.HasFilters());
```
bdero, line 309: This feature was added in order to apply the opacity in the correct order in some cases. Unfortunately, opacity is starting to get kind of hard to reason about in Impeller.

Would this still work correctly when:
1. An image filter is present along with a color filter?
2. An advanced blend + a color filter is being used?

cc @ColdPaleLight , who had to contend with opacity deferral here: https://github.com/flutter/engine/pull/36279

impeller/aiks/paint.cc
jonahwilliams: Good point, I should check for this with https://github.com/flutter/engine/pull/39266

The general answer is that the high level framework API makes it extrenely easy to apply color blend filters to images, and no other type of color or image filter. see color and colorBlendMode in https://api.flutter.dev/flutter/widgets/Image-class.html

impeller/aiks/canvas.cc
jonahwilliams: 1/2 essentially don't matter for performance (though they certainly do for correctness). The issue we need to contend with is that Flutter makes it very easy to add a color+blend to an image, and its expected that this is only marginally more expensive than drawing the image itself (probably due to skia genering shaders for this).

So one partial solution is what I have here, but perhaps this is pushing the implementation in the wrong direction. In all of the cases I'm worried about, we'll get some drawing primitive (drawRect + tiled texture), drawImageRect, drawImageNine, and then a paint with a color filter on it.

Perhaps the way to go about this is to recognize specifically that case (requiring that color sources and color filters are typed and not just opaque procs), and then create a specialized contents for it. That would also make it easier to fix a simialr problem with tiled textures where we tile then color filter instead of color filter then tile.

impeller/aiks/paint.cc
ColdPaleLight: As of now, all color filters absorb opacity.

impeller/aiks/canvas.cc
ColdPaleLight: 1. An image filter is present along with a color filter?
If the ImageFilter and ColorFilter work together, the ColorFilter will be applied first and then the ImageFilter will be applied. And as of now, all ColorFilter will absorb opacity, so I think in this scenario, the performance of this PR will be correct.

2. An advanced blend + a color filter is being used?
Since advanced blend is still applied after ColorFilter, there will be no problem

So I think this PR is correct.

ColdPaleLight (approved):
LGTM

impeller/aiks/paint.h
jonahwilliams: Done

PR #36346, opened 2022-09-22
Web trackpad pan

moffatman:
Use a kind of heuristic to guess at whether an event is from mouse or trackpad.

Part of https://github.com/flutter/flutter/issues/64210

## Sequence

1. https://github.com/flutter/flutter/pull/112171
2. This PR

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [Flutter Style Guide] and the [C++, Objective-C, Java style guides].
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I added new tests to check the change I am making or feature I am adding, or Hixie said the PR is test-exempt. See [testing the engine] for instructions on
writing and running engine tests.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I signed the [CLA].
- [x] All existing and new tests are passing.

PR #39260, opened 2023-01-30
Add unique device id for trackpad on web

Piinks:
Fixes https://github.com/flutter/flutter/issues/118051

After https://github.com/flutter/engine/pull/36346, it was possible to trip an assertion in the framework since we were differentiating mouse & trackpad on web - but weren't giving them unique IDs. This adds that.

## Pre-launch Checklist

- [x] I signed the [CLA].


lib/web_ui/test/engine/pointer_binding_test.dart
```diff
@@ -1330,6 +1334,7 @@ void testMain() {
           packets[4].data[0].signalKind, equals(ui.PointerSignalKind.scroll));
       expect(
           packets[4].data[0].kind, equals(ui.PointerDeviceKind.trackpad));
+      expect(packets[4].data[0].pointerIdentifier, equals(-2));
```
moffatman, line 1337: Should be device?

Piinks: Oh 🤦‍♀️  - thanks. Still recovering from jet lag.

Piinks, line 1337:
```suggestion
      expect(packets[4].data[0].device, equals(-2));
```

moffatman:
I think the test problem is that now there are two different device IDs, so the first mouse event will have a synthesized add event.

moffatman:
Wrong button 😁

yjbanov (approved):
LGTM, but cc @mdebbar @htoor3 who own the input area.

htoor3 (approved):
LGTM

mdebbar (approved):

Piinks:
> I think the test problem is that now there are two different device IDs, so the first mouse event will have a synthesized add event.

Hmm. I will take another look tomorrow. Thanks!

Piinks:
> I think the test problem is that now there are two different device IDs, so the first mouse event will have a synthesized add event.

@moffatman did you mean that the test just need to be updated? I figured the prior synthesized event may need to be removed or cancelled, but that doesn't seem right. There could be a mouse and a trackpad both at play at the same time.

moffatman:
Yeah, the test just needs to be updated for the new behaviour. packets[5] now has [add, scroll] instead of just [scroll]. Since it's now using a different device ID from the earlier events.

PR #39266, opened 2023-01-30
[Impeller] make color filter objects inspectable.

jonahwilliams:
Would be required to determine whether or not to use specialized contents for drawing an image with a color filter.


jonahwilliams:
actually that may be pointless, I forgot about the blend mode - so we'd actually need a whole shitload of shaders. Probably not a great approach unless we wanted to canonicalize whatever blendmode the framework used by default

jonahwilliams:
Haven't found a use for this yet so closing

PR #39267, opened 2023-01-30
Add support for double tap action from Apple Pencil 2

LouiseHsu:
This PR adds engine support for the double tap action on the Apple Pencil 2.

Addresses https://github.com/flutter/flutter/issues/73172

The Apple Pencil 2 has a feature which allows the user to double tap the side of the pencil to perform one of 4 preset actions, which can be set in Settings. The double tap is treated as a pointer event, with a new field that represents the preferred action of the user. 

Framework PR will come after this is merged and will be linked here :)

## Pre-launch Checklist

- [x] I signed the [CLA].


lib/ui/platform_dispatcher.dart
```diff
@@ -417,6 +417,7 @@ class PlatformDispatcher {
         panDeltaY: packet.getFloat64(kStride * offset++, _kFakeHostEndian),
         scale: packet.getFloat64(kStride * offset++, _kFakeHostEndian),
         rotation: packet.getFloat64(kStride * offset++, _kFakeHostEndian),
+        preferredAction: PointerPreferredAction.values[packet.getInt64(kStride * offset++, _kFakeHostEndian)],
```
hellohuanlin, line 420: it's not clear what "preferredAction" or `PointerPreferredAction` means in this context. How about `preferredStylusAuxilliaryAction` and `PointerPreferredStylusAuxilliaryAction`

lib/ui/pointer.dart
```diff
@@ -133,10 +133,21 @@ enum PointerSignalKind {
   /// A pointer-generated scale event (e.g. trackpad pinch).
   scale,
 
+  /// A stylus generated action (e.g. double tap on Apple Pencil 2)
+  stylusAction,
```
hellohuanlin, line 137: same comment on the naming here.

lib/ui/pointer.dart
```diff
@@ -133,10 +133,21 @@ enum PointerSignalKind {
...
   /// An unknown pointer signal kind.
   unknown
 }
 
+  /// The preferred action for stylus action
+enum PointerPreferredAction {
+  ignore,
+  showColorPalette,
+  switchEraser,
+  switchPrevious
```
hellohuanlin, line 148: `switchToEraser` and `switchToPrevious`

shell/platform/darwin/ios/framework/Source/FlutterViewController.mm
```diff
@@ -956,7 +962,7 @@ - (void)goToApplicationLifecycle:(nonnull NSString*)state {
     case UITouchTypeDirect:
     case UITouchTypeIndirect:
       return flutter::PointerData::DeviceKind::kTouch;
-    case UITouchTypeStylus:
+    case UITouchTypePencil:
```
hellohuanlin, line 965: is this change intended?

shell/platform/darwin/ios/framework/Source/FlutterViewController.mm
```diff
@@ -1211,6 +1217,40 @@ - (void)invalidateTouchRateCorrectionVSyncClient {
   _touchRateCorrectionVSyncClient = nil;
 }
 
+#pragma mark - Stylus Events
+
+#pragma clang diagnostic push
+
+- (void)pencilInteractionDidTap:(UIPencilInteraction*)interaction API_AVAILABLE(ios(13.4)) {
+  flutter::PointerData pointer_data = [self generatePointerDataAtLastMouseLocation];
```
hellohuanlin, line 1225: do we need the mouse location for this pointer_data?

shell/platform/darwin/ios/framework/Source/FlutterViewController.mm
```diff
@@ -1211,6 +1217,40 @@ - (void)invalidateTouchRateCorrectionVSyncClient {
...
+    case UIPencilPreferredActionShowColorPalette:
+      pointer_data.preferred_action = flutter::PointerData::PreferredAction::kShowColorPalette;
+      break;
+    case UIPencilPreferredActionSwitchEraser:
+      pointer_data.preferred_action = flutter::PointerData::PreferredAction::kSwitchEraser;
+      break;
+    case UIPencilPreferredActionSwitchPrevious:
+      pointer_data.preferred_action = flutter::PointerData::PreferredAction::kSwitchPrevious;
+    default:
+      break;
```
hellohuanlin, line 1240: if apple come up with a new case, will this cause trouble on the framework side? Or should there be a "kUnknown" type?

shell/platform/darwin/ios/framework/Source/FlutterViewController.mm
```diff
@@ -1680,7 +1720,7 @@ - (void)pressesCancelled:(NSSet<UIPress*>*)presses
 
 #pragma mark - Orientation updates
 
-- (void)onOrientationPreferencesUpdated:(NSNotification*)notification {
+- (void)onOrientationPreferencesUpdated:(NSNotification*)notification API_AVAILABLE(ios(12.1)) {
```
hellohuanlin, line 1723: is this intended?

shell/platform/darwin/ios/framework/Source/FlutterViewControllerTest.mm
```diff
@@ -1479,6 +1480,27 @@ - (void)testMouseSupport API_AVAILABLE(ios(13.4)) {
...
+                                                                     bundle:nil];
+  XCTAssertNotNil(vc);
+
+  id mockPencilInteraction = OCMClassMock([UIPencilInteraction class]);
+  XCTAssertNotNil(mockPencilInteraction);
+
+  [vc pencilInteractionDidTap:mockPencilInteraction];
+
+  [[[self.mockEngine verify] ignoringNonObjectArgs]
+      dispatchPointerDataPacket:std::make_unique<flutter::PointerDataPacket>(0)];
```
hellohuanlin, line 1501: missing test on the preferred action part.

hellohuanlin (commented):
Overall looks good! just some questions around the naming

shell/platform/darwin/ios/framework/Source/FlutterViewController.mm
```diff
@@ -93,7 +95,7 @@ @interface FlutterViewController () <FlutterBinaryMessenger, UIScrollViewDelegat
 // Trackpad rotating
 @property(nonatomic, retain)
     UIRotationGestureRecognizer* rotationGestureRecognizer API_AVAILABLE(ios(13.4));
-
+@property(nonatomic, retain) UIPencilInteraction* pencilInteraction API_AVAILABLE(ios(13.4));
```
jmagman, line 98: This is manually reference counted code, so you'll need to manually release `_pencilInteraction` (the backing instance variable that gets synthesized from a property declaration).
```
/Volumes/Work/s/w/ir/cache/builder/src/flutter/shell/platform/darwin/ios/framework/Source/FlutterViewController.mm:891:3: error: The '_pencilInteraction' ivar in 'FlutterViewController' was retained by a synthesized property but not released before '[super dealloc]' [clang-analyzer-osx.cocoa.Dealloc,-warnings-as-errors]
  [super dealloc];
```
https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8790504096978028977/+/u/test:_lint_ios_debug_sim/stdout

So in the `dealloc` method below, add:
```objc
  [_pencilInteraction release];
```

lib/ui/pointer.dart
```diff
@@ -176,6 +187,7 @@ class PointerData {
     this.panDeltaY = 0.0,
     this.scale = 0.0,
     this.rotation = 0.0,
+    this.preferredAction = PointerPreferredAction.ignore,
```
jmagman, line 190:
```
Warning: lib/ui/ui.dart PointerData.Unnamed Constructor has a different parameter length than in lib/web_ui/ui.dart.
```
https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8790504096880865457/+/u/analyze_dart_ui/stdout

lib/ui/pointer.dart
jmagman: Trailing commas.
```suggestion
  switchPrevious,
```

lib/ui/pointer.dart
```diff
@@ -133,10 +133,21 @@ enum PointerSignalKind {
...
 
+  /// A stylus generated action (e.g. double tap on Apple Pencil 2)
+  stylusAction,
+
   /// An unknown pointer signal kind.
   unknown
 }
 
+  /// The preferred action for stylus action
+enum PointerPreferredAction {
```
jmagman, line 144: Should this specify stylus then?
```suggestion
enum PointerPreferredStylusAction {
```

jmagman (commented):
I know you have work-in-progress design doc out for this feature: https://docs.google.com/document/d/1r4P5r-jGt2Sjqro3ldCU2axUiHTpu3yhIycnI94OKQw/edit

I think you should finish that up, add the missing sections, and pass it around to the framework team before this gets merged.

lib/ui/platform_dispatcher.dart
jmagman: https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8790504096978029009/+/u/test:_Host_Tests_for_host_debug_unopt/stdout
```
[0;32m[ RUN      ] [mShellTest.CanCorrectlySynthesizePointerPacket
[ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: RangeError (index): Invalid value: Not in inclusive range 0..3: 1731047719413088311
#0      _Array.[] (dart:core-patch/array.dart:10:36)
#1      PlatformDispatcher._unpackPointerDataPacket (dart:ui/platform_dispatcher.dart:420:55)
#2      PlatformDispatcher._dispatchPointerDataPacket (dart:ui/platform_dispatcher.dart:364:9)
#3      _dispatchPointerDataPacket (dart:ui/hooks.dart:91:31)
```

lib/ui/pointer.dart
LouiseHsu: @hellohuanlin suggested preferredStylusAuxilliaryAction or PointerPreferredStylusAuxilliaryAction, what do you think?

jmagman: I think you have an extra "l" in "Auxiliary".  But what does "Auxiliary" mean in this context?

hellohuanlin: I view stylus touching on screen as the "primary action" and Apple Pencil double tap as "auxiliary action". If we just call it "stylus action"  or "pointer action", it sounds like stylus touching on screen (rather finger tapping on pencil in our case)

jmagman: This would be a good question for the framework team, I'm sure they have prior art for this problem.

lib/ui/platform_dispatcher.dart
LouiseHsu: I think its actually really confusing if both the actual event and the preferredAction are called PointerPreferredStylusAuxilliaryAction and PointerStylusAuxilliaryAction respectively,  I got confused just now between the two as Im trying to change the names because then the only difference them is the "preferred", and its pretty verbose

LouiseHsu: How about StylusSecondaryGesture (double tap) + ConfiguredStylusAction (the preferred action)?

lib/ui/pointer.dart
LouiseHsu: This path "lib/web_ui/ui.dart" doesnt exist, but I think this is because I need to update pointer packet on the Android side too

shell/platform/darwin/ios/framework/Source/FlutterViewController.mm
LouiseHsu: Yeah, UITouchTypeStylus is deprecated and is suggested https://developer.apple.com/documentation/uikit/uitouchtype/uitouchtypestylus

lib/ui/pointer.dart
hellohuanlin: One is udner `flutter/lib/ui/pointer.dart`, another one is under `flutter/lib/web_ui/pointer.dart`

shell/platform/darwin/ios/framework/Source/FlutterViewController.mm
```diff
@@ -1211,6 +1219,46 @@ - (void)invalidateTouchRateCorrectionVSyncClient {
   _touchRateCorrectionVSyncClient = nil;
 }
 
+#pragma mark - Stylus Events
+
+#pragma clang diagnostic push
```
jmagman, line 1224: Is this pragma push pop here for something? Can it be removed?

shell/platform/darwin/ios/framework/Source/FlutterViewControllerTest.mm
```diff
@@ -1479,6 +1480,36 @@ - (void)testMouseSupport API_AVAILABLE(ios(13.4)) {
       dispatchPointerDataPacket:std::make_unique<flutter::PointerDataPacket>(0)];
 }
 
+- (void)testPencilSupport API_AVAILABLE(ios(13.4)) {
+  if (@available(iOS 13.4, *)) {
+    // noop
+  } else {
+    return;
+  }
```
jmagman, line 1488: I don't think you need this with the `API_AVAILABLE` in the method signature.

shell/platform/darwin/ios/framework/Source/FlutterViewControllerTest.mm
```diff
@@ -1479,6 +1480,36 @@ - (void)testMouseSupport API_AVAILABLE(ios(13.4)) {
...
+
+  FlutterViewController* vc = [[FlutterViewController alloc] initWithEngine:self.mockEngine
+                                                                    nibName:nil
+                                                                     bundle:nil];
+  XCTAssertNotNil(vc);
+
+  id mockPencilInteraction = OCMClassMock([UIPencilInteraction class]);
+  XCTAssertNotNil(mockPencilInteraction);
```
jmagman, line 1496: You don't need to assert that the mock you just made isn't nil, then you're just testing that OCMock works.

shell/platform/darwin/ios/framework/Source/FlutterViewControllerTest.mm
```diff
@@ -1479,6 +1480,36 @@ - (void)testMouseSupport API_AVAILABLE(ios(13.4)) {
...
+
+  [vc pencilInteractionDidTap:mockPencilInteraction];
+
+  flutter::PointerData pointer_data;
+
+  pointer_data.kind = flutter::PointerData::DeviceKind::kStylus;
+  pointer_data.signal_kind = flutter::PointerData::SignalKind::kStylusAuxiliaryAction;
+  pointer_data.preferred_auxiliary_stylus_action =
+      flutter::PointerData::PreferredStylusAuxiliaryAction::kIgnore;
```
jmagman, line 1505: Where was `ignore` mocked?

shell/platform/darwin/ios/framework/Source/FlutterViewController.mm
hellohuanlin: looks like this is not resolved - we don't need the mouse location here. Can probably just do

```
  flutter::PointerData pointer_data;
  pointer_data.Clear();
```

shell/platform/darwin/ios/framework/Source/FlutterViewControllerTest.mm
LouiseHsu: This section is in every other test in this file too, they also all have the API_AVAILABLE in the signature. should I remove it for the others too?

jmagman: Yes, that would be great, assuming they all pass.  Thank you!

shell/platform/darwin/ios/framework/Source/FlutterViewControllerTest.mm
LouiseHsu: Ignore is the default value of UIPencilInteraction.preferredAction - it's a readonly property so I'm not sure how I would mock it to be a different enum.

hellohuanlin: You can look into how to mock static methods using ocmock.

jmagman: > Ignore is the default value of UIPencilInteraction.preferredAction

We wouldn't want this test to fail based on real settings on the target device.

jmagman: See https://ocmock.org/reference/#mocking-class-methods
It would be something like:
```objc
OCMStub([mockPencilInteraction preferredAction]).andReturn(UIPencilPreferredActionIgnore);
```
And at the end of the test, reset the class method mock:
```objc
[mockPencilInteraction stopMocking];
```

shell/platform/darwin/ios/framework/Source/FlutterViewControllerTest.mm
```diff
@@ -1479,6 +1452,33 @@ - (void)testMouseSupport API_AVAILABLE(ios(13.4)) {
...
+  pointer_data.Clear();
+  pointer_data.kind = flutter::PointerData::DeviceKind::kStylus;
+  pointer_data.signal_kind = flutter::PointerData::SignalKind::kStylusAuxiliaryAction;
+  pointer_data.preferred_auxiliary_stylus_action =
+      flutter::PointerData::PreferredStylusAuxiliaryAction::kShowColorPalette;
+
+  auto packet = std::make_unique<flutter::PointerDataPacket>(1);
+  packet->SetPointerData(/*index=*/0, pointer_data);
+
+  [[[self.mockEngine verify] ignoringNonObjectArgs] dispatchPointerDataPacket:std::move(packet)];
```
jmagman, line 1478: `ignoringNonObjectArgs` is ignoring the value of the `packet` and accepting anything, so you aren't actually testing anything here except that `dispatchPointerDataPacket` was called at least once.  When I swap the action to `kSwitchEraser` it still passes.

This doesn't pass but something like:


```objc
  id mockPencilInteraction = OCMClassMock([UIPencilInteraction class]);

  OCMStub([mockPencilInteraction preferredTapAction])
      .andReturn(UIPencilPreferredActionShowColorPalette);

  XCTestExpectation* pointerDataPacketExpectation =
      [self expectationWithDescription:@"pointerDataPacket"];
  OCMStub(
      [self.mockEngine dispatchPointerDataPacket:std::make_unique<flutter::PointerDataPacket>(0)])
      .ignoringNonObjectArgs()
      .andDo(^(NSInvocation* invocation) {
        flutter::PointerData packet;
        [invocation getArgument:&packet atIndex:2];
        XCTAssertEqual(packet.kind, flutter::PointerData::DeviceKind::kStylus);
        XCTAssertEqual(packet.signal_kind,
                       flutter::PointerData::SignalKind::kStylusAuxiliaryAction);
        XCTAssertEqual(packet.preferred_auxiliary_stylus_action,
                       flutter::PointerData::PreferredStylusAuxiliaryAction::kShowColorPalette);
        [pointerDataPacketExpectation fulfill];
      });

  [vc pencilInteractionDidTap:mockPencilInteraction];

  [self waitForExpectationsWithTimeout:1 handler:nil];

  [mockPencilInteraction stopMocking];
```

jmagman:
@camsim99 would you mind taking a look at the Android java part of this PR?

shell/platform/darwin/ios/framework/Source/FlutterViewControllerTest.mm
jmagman: Hm, this still doesn't pass but it's closer, it needs to get the pointer data off the packet.
```objc
    std::unique_ptr<flutter::PointerDataPacket> packet;
    [invocation getArgument:&packet atIndex:2];
    flutter::PointerData pointer_data = packet->GetPointerData(0);

    XCTAssertEqual(pointer_data.kind, flutter::PointerData::DeviceKind::kStylus);
    XCTAssertEqual(pointer_data.signal_kind,
                   flutter::PointerData::SignalKind::kStylusAuxiliaryAction);
    XCTAssertEqual(pointer_data.preferred_auxiliary_stylus_action,
                   flutter::PointerData::PreferredStylusAuxiliaryAction::kShowColorPalette);
    [pointerDataPacketExpectation fulfill];
```

hellohuanlin: I did some research just now - it looks like there's no such OCMock API for dereferencing smart pointers at all. I think OCMock only supports dereferencing raw pointers. If that's the case, `verify` maybe the only choice here for this test. 

Another idea is to break out the main logic into a helper function like `getAxiliaryActionPacket()` so it's easier to test.

jmagman: > I could be wrong, but I wasn't able to find any examples online. If that's the case, `verify` maybe the only choice here for this test.

How do you suggest getting `verify` working?  It would need to do some equality check.

camsim99 (commented):
Android part looks good to me!

shell/platform/darwin/ios/framework/Source/FlutterViewControllerTest.mm
hellohuanlin: I was suggesting to just verify that it is called (like [this](https://github.com/flutter/engine/commit/a8a7534eabbad4eedbc3fa2963cdefd604541709#diff-9df3e6878a0f23383e1f2de991386935e435c5b77c863de75bd67a4ced38591dR697-R698)). 

For the equality check, i think we can factor out a helper function `getAxiliaryActionPacket()` and test that instead.

lib/ui/pointer.dart
```diff
@@ -414,6 +441,7 @@ class PointerData {
              'panDeltaY: $panDeltaY, '
              'scale: $scale, '
              'rotation: $rotation'
```
jmagman, line 443:
```suggestion
             'rotation: $rotation, '
```

lib/web_ui/lib/pointer.dart
```diff
@@ -146,6 +166,7 @@ class PointerData {
            'panDeltaY: $panDeltaY, '
            'scale: $scale, '
            'rotation: $rotation'
```
jmagman, line 168:
```suggestion
           'rotation: $rotation, '
```

shell/platform/darwin/ios/framework/Source/FlutterViewController.mm
```diff
@@ -1211,6 +1219,50 @@ - (void)invalidateTouchRateCorrectionVSyncClient {
   _touchRateCorrectionVSyncClient = nil;
 }
 
+#pragma mark - Stylus Events
+
+- (void)pencilInteractionDidTap:(UIPencilInteraction*)interaction API_AVAILABLE(ios(13.4)) {
+  flutter::PointerData pointer_data = [self createAuxillaryStylusActionData];
+
+  auto packet = std::make_unique<flutter::PointerDataPacket>(1);
+  packet->SetPointerData(/*index=*/0, pointer_data);
```
hellohuanlin, line 1228: you can probably put this packet creation into the helper so that it's also unit tested.

jmagman:
> sigh whitespace

You can run the engine formatters locally instead of waiting to see it fails in CI:
```
$ ci/format.sh --fix
```

shell/platform/darwin/ios/framework/Source/FlutterViewControllerTest.mm
```diff
@@ -123,6 +123,8 @@ - (void)performOrientationUpdate:(UIInterfaceOrientationMask)new_preferences;
 - (void)handlePressEvent:(FlutterUIPressProxy*)press
               nextAction:(void (^)())next API_AVAILABLE(ios(13.4));
 - (void)discreteScrollEvent:(UIPanGestureRecognizer*)recognizer;
+- (void)pencilInteractionDidTap:(UIPencilInteraction*)interaction;
```
jmagman, line 126: The interface can implement the delegate protocol instead:
```objc
@interface FlutterViewController (Tests) <UIPencilInteractionDelegate>
```
```suggestion
```

shell/platform/darwin/ios/framework/Source/FlutterViewController.mm
```diff
@@ -1220,6 +1228,50 @@ - (void)invalidateTouchRateCorrectionVSyncClient {
   _touchRateCorrectionVSyncClient = nil;
 }
 
+#pragma mark - Stylus Events
+
+- (void)pencilInteractionDidTap:(UIPencilInteraction*)interaction API_AVAILABLE(ios(13.4)) {
+  auto packet = [self createAuxillaryStylusActionData];
```
jmagman, line 1234: I think this failure
```
2023-02-10 13:00:50.880393-0800 IosUnitTests[97230:523827] *** NSForwarding: warning: object 0x7ff7b43dd258 of class 'nil' does not implement methodSignatureForSelector: -- did you forget to declare the superclass of 'nil'?
2023-02-10 13:00:50.881393-0800 IosUnitTests[97230:523827] *** NSForwarding: warning: selector (0x7fefbeb1ca00) for message '0^*\^H' does not match selector known to Objective C runtime (0x600000c044d0)-- abort
2023-02-10 13:00:50.882077-0800 IosUnitTests[97230:523827] *** NSForwarding: warning: object 0x7ff7b43dd258 of class 'nil' does not implement doesNotRecognizeSelector: -- abort
```
is because `-pencilInteractionDidTap:` is called on a partial mock, and OCMock is choking trying to figure out the smart pointer signature of `createAuxillaryStylusActionData`.

It passes if you go back to returning `flutter::PointerData` from this method (and update the tests)
https://github.com/flutter/engine/pull/39267#discussion_r1102090058

```objc
- (flutter::PointerData)createAuxillaryStylusActionData
```

hellohuanlin: Interesting find. @LouiseHsu can you try not returning the smart pointer format? Just return the raw packet type should work.

LouiseHsu: I tried returning the raw packet type in an this earlier commit https://github.com/flutter/engine/pull/39267/commits/1cdf0ec7ec0dfc74049c8195dffc9f830de3058b
but I ran into compile errors

hellohuanlin: > I tried returning the raw packet type in an this earlier commit https://github.com/flutter/engine/commit/1cdf0ec7ec0dfc74049c8195dffc9f830de3058b

Did you paste the wrong commit?

hellohuanlin (commented):
This is getting very close. just a few comments.

jmagman (approved):
LGTM, thanks for spending so much time on this @LouiseHsu!

lib/ui/platform_dispatcher.dart
```diff
@@ -417,6 +417,7 @@ class PlatformDispatcher {
         panDeltaY: packet.getFloat64(kStride * offset++, _kFakeHostEndian),
         scale: packet.getFloat64(kStride * offset++, _kFakeHostEndian),
         rotation: packet.getFloat64(kStride * offset++, _kFakeHostEndian),
+        preferredStylusAuxiliaryAction: PointerPreferredStylusAuxiliaryAction.values[packet.getInt64(kStride * offset++, _kFakeHostEndian)],
```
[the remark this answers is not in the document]
LouiseHsu: its already in the design doc, and they didnt object - if they change their mind, I think it wouldnt be a big deal to change the naming.

hellohuanlin (approved):
Nice! Great work!

PR #39268, opened 2023-01-30
Download emsdk for web framework tests.

eyebrowsoffire:
This is in preparation for changes that will use `--local-web-sdk` instead of `--local-engine` for the web framework tests.


hterkelsen (approved):
LGTM

PR #39279, opened 2023-01-31
Use the x64 toolchain when building host artifacts on arm64 mac

zanderso:
This is needed to use arm64 macs in CI, and for the reasons noted in the comment.

### DIFF
--- a/tools/gn
+++ b/tools/gn
@@ -342,22 +342,21 @@ def to_gn_args(args):
     gn_args['host_cpu'] = 'x86'
     gn_args['current_cpu'] = 'x86'
 
-  if is_host_build(args) and gn_args['host_os'] == 'mac':
-    # macOS host builds (whether x64 or arm64) must currently be built under
-    # Rosetta on Apple Silicon Macs.
-    # TODO(cbracken): https://github.com/flutter/flutter/issues/103386
+  # When building binaries to run on a macOS host (like gen_snapshot), always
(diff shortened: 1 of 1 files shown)


zanderso:
cc @godofredoc

jmagman (commented):
LGTM in theory, @cbracken should have the deciding review though.

tools/gn
```diff
@@ -342,6 +342,17 @@ def to_gn_args(args):
...
+  # When building binaries to run on a macOS host (like gen_snapshot), always
+  # use the x64 toolchain. This is done for two reasons:
+  #  1. goma currently only supports the clang_x64 toolchain.
+  #  2. gen_snapshot cannot crossbuild from arm64 to x64. Its host architecture
+  #     must be x64 to target x64.
+  # However, if native arm64 binaries are needed when not building the engine
+  # for macOS (see the logic below), pass the flag "--mac-cpu arm64".
+  # https://github.com/flutter/flutter/issues/103386 is also relevant here.
+  if get_host_os() == 'mac' and not args.mac_cpu == 'arm64':
+    gn_args['host_cpu'] = 'x64'
```
cbracken, line 354: Haven't I already covered this for host builds in lines 356-360?

zanderso: I simplified the logic, and with the simpler logic, hopefully the comment makes more sense now. PTAL.

cbracken: Looks good. Thanks!

cbracken (approved):

PR #39288, opened 2023-01-31
[web] Update the within() matcher to be more useful

mdebbar:
Instead of hardcoding the error tolerance value everywhere, let's use the `moreOrLessEquals` matcher like flutter/flutter does: https://github.com/flutter/flutter/blob/2c34a88eb89628e6a80caa50dd03797f9b6aa5e5/packages/flutter_test/lib/src/matchers.dart#L251-L253

Note: the matcher were copied straight from the flutter/flutter repo.


lib/web_ui/test/geometry_test.dart
```diff
@@ -32,33 +34,24 @@ void testMain() {
...
-    expect(Offset.fromDirection(pi / 4.0).dx,
-        closeTo(1.0 / math.sqrt(2.0), 1e-12));
-    expect(Offset.fromDirection(pi / 4.0).dy,
-        closeTo(1.0 / math.sqrt(2.0), 1e-12));
-    expect(Offset.fromDirection(-pi / 4.0).dx,
-        closeTo(1.0 / math.sqrt(2.0), 1e-12));
-    expect(Offset.fromDirection(-pi / 4.0).dy,
-        closeTo(-1.0 / math.sqrt(2.0), 1e-12));
+    expect(Offset.fromDirection(pi / 4.0).dx, moreOrLessEquals(1.0 / math.sqrt(2.0)));
+    expect(Offset.fromDirection(pi / 4.0).dy, moreOrLessEquals(1.0 / math.sqrt(2.0)));
```
yjbanov, line 43: We also have `within` that supports types other than numbers. For example, it could collapse the above two lines into one. It doesn't have a built-in epsilon value, but we could teach it. We own the implementation.

mdebbar: `within` already takes a `distance` parameter which would basically be the epsilon value in this case:
```dart
expect(
  Offset.fromDirection(pi / 4.0),
  within(
    distance: epsilon,
    from: Offset(1.0 / math.sqrt(2.0), 1.0 / math.sqrt(2.0)),
  ),
);
```
Maybe we should make `moreOrLessEquals` generic? Like:
```dart
Matcher moreOrLessEquals<T>(
  T value, {
  double epsilon = precisionErrorTolerance,
  DistanceFunction<T>? distanceFunction,
}) {
  ...
}
```
Then the above two lines become:

```dart
expect(
  Offset.fromDirection(pi / 4.0),
  moreOrLessEquals(Offset(1.0 / math.sqrt(2.0), 1.0 / math.sqrt(2.0))),
);
```

ditman: If we make `distance` in `within` default to `precisionErrorTolerance`, then it becomes moreOrLessEquals to `moreOrLessEquals`, right?

ditman: (I would force `AnyDistanceFunction` to return a `double` rather than `num`, but that's just me)

mdebbar: Great idea @ditman !

lib/web_ui/test/matchers.dart
```diff
@@ -496,3 +503,54 @@ final Matcher throwsAssertionError = throwsA(isAssertionError);
...
+///
+/// See also:
+///
+///  * [closeTo], which is identical except that the epsilon argument is
+///    required and not named.
+///  * [inInclusiveRange], which matches if the argument is in a specified
+///    range.
+///  * [rectMoreOrLessEquals] and [offsetMoreOrLessEquals], which do something
+///    similar but for [Rect]s and [Offset]s respectively.
+Matcher moreOrLessEquals(double value, { double epsilon = precisionErrorTolerance }) {
```
ditman, line 526: This is not needed anymore?

mdebbar: We still use this in many tests. Are you suggesting we use `within()` for everything?

ditman: In flutter/engine/web_ui we don't need to introduce the matcher now, or do we? I only see it used in one test in this PR (and that could use `within` as well?)

ditman (approved):

PR #39294, opened 2023-01-31
[web] Hide autofill overlay

htoor3:
There's an issue currently where the autofill overlay renders on top of our Flutter-rendered text inputs.  We previously were preventing these overlays from rendering via a `transition-delay` - however, scrolling immediately after autofilling short circuits these transition delays and renders the overlay anyways. 

This PR is changing the styling to instead hide the overlays rather than delay their rendering.  We do this by making the overlays completely transparent, which solves this issue.  

Fixes https://github.com/flutter/flutter/issues/118337

## Pre-launch Checklist

- [x] I signed the [CLA].


mdebbar (approved):

lib/web_ui/lib/src/engine/host_node.dart
```diff
@@ -317,16 +317,16 @@ void applyGlobalCssRulesToSheet(
...
-  // See: https://github.com/flutter/flutter/issues/61132.
+  // This CSS makes the autofill overlay transparent in order to prevent it
+  // from overlaying on top of Flutter-rendered text inputs.
+  // See: https://github.com/flutter/flutter/issues/118337.
   if (browserHasAutofillOverlay()) {
     sheet.insertRule('''
       $cssSelectorPrefix .transparentTextEditing:-webkit-autofill,
       $cssSelectorPrefix .transparentTextEditing:-webkit-autofill:hover,
       $cssSelectorPrefix .transparentTextEditing:-webkit-autofill:focus,
       $cssSelectorPrefix .transparentTextEditing:-webkit-autofill:active {
```
ditman, line 328: Should we start targeting the [standard](https://developer.mozilla.org/en-US/docs/Web/CSS/:autofill) `:autofill` pseudo-class as well?

ditman (approved):
So firefox and safari understand `-webkit-autofill`, but internally they call it just `autofill`, all right then!

lib/web_ui/lib/src/engine/host_node.dart
htoor3: It does mention here https://developer.mozilla.org/en-US/docs/Web/CSS/:autofill to use both for best browser compatibility, but in manual testing:
- I was able to use the CSSOM to read the above `:-webkit-autofill` style on every browser.  Like you mentioned above, safari and firefox just read it internally as `:autofill`
- Interestingly enough, Chrome wouldn't read it at all if we just used `:autofill`...chrome and edge both need the webkit prefix.

So we can certainly conditionally target the `:autofill` selector for Firefox and Safari for "completeness", but it works fine as is too.  The one downside I see in targeting `:autofill` as well is that it won't work for older versions of those browsers. 

What do you think?

ditman: I think that until firefox and safari break because they don't recognize `-webkit-autofill`, we're fine :P

PR #39297, opened 2023-01-31
Remove uses of `--full-dart-sdk` from the engine_v2 builders.

eyebrowsoffire:
This updates some of the engine_v2 builders to get them away from the old "host build to make the web sdk" pattern.


ci/builders/linux_host_engine.json
```diff
@@ -78,7 +78,6 @@
                         "out/host_debug/zip_archives/linux-x64/font-subset.zip",
                         "out/host_debug/zip_archives/flutter_patched_sdk.zip",
                         "out/host_debug/zip_archives/dart-sdk-linux-x64.zip",
-                        "out/host_debug/zip_archives/flutter-web-sdk-linux-x64.zip",
```
godofredoc, line 81: Does this mean we don't need this artifact anymore?

eyebrowsoffire: That's correct. The flutter tool no longer uses any of the host specific flutter-web-sdk artifacts (i.e. `flutter-web-sdk-{platform-name}.zip`). It now just uses `flutter-web-sdk.zip` which is host-agnostic, and that is produced by the linux_host_engine builder.

godofredoc: Thanks for the explanation. LGTM.

godofredoc (approved):

ci/builders/linux_host_engine.json
eyebrowsoffire: Thank you!

PR #39345, opened 2023-02-02
[impellerc] sort uniforms on metal backend

jonahwilliams:
Fixes https://github.com/flutter/flutter/issues/119173, https://github.com/flutter/flutter/issues/118847

Use add_msl_resource_binding to place metal samplers in the same decoration order that we use on the SkSL backend. This should ensure that float/sampler uniforms from user authored fragment shaders behave identically across backends.


impeller/compiler/reflector.cc
```diff
@@ -359,23 +360,30 @@ std::shared_ptr<RuntimeStageData> Reflector::GenerateRuntimeStageData() const {
...
-        data->AddUniformDescription(std::move(uniform_description));
-      });
+
+  // Sort the IR so that the uniforms are in declaration order.
+  std::vector<spirv_cross::ID> uniforms =
+      SortUniforms(ir_.get(), compiler_.GetCompiler());
+
+  for (auto& sorted_id : uniforms) {
+    ir_->for_each_typed_id<spirv_cross::SPIRVariable>(
+        [&](uint32_t id, const spirv_cross::SPIRVariable& var) {
```
jonahwilliams, line 370: OK I KNOW HOW THIS LOOKS 😆 , will fix once I find the right public API

impeller/entity/contents/runtime_effect_contents.cc
```diff
@@ -182,7 +182,7 @@ bool RuntimeEffectContents::Render(const ContentContext& renderer,
 
         ShaderUniformSlot uniform_slot;
         uniform_slot.name = uniform.name.c_str();
-        uniform_slot.ext_res_0 = buffer_index;
+        uniform_slot.ext_res_0 = uniform.location;
```
jonahwilliams, line 185: In practice buffer_index and uniform.location are the same now but I believe this is more correct.

impeller/compiler/compiler.cc
```diff
@@ -36,6 +37,43 @@ static CompilerBackend CreateMSLCompiler(const spirv_cross::ParsedIR& ir,
       spirv_cross::CompilerMSL::Options::make_msl_version(1, 2);
   sl_compiler->set_msl_options(sl_options);
 
+  auto floats =
```
zanderso, line 40: Some comments about why this is being done here could be helpful.

zanderso (approved):
lgtm w/ a request for a comment. We should also get this queued up for a CP.

impeller/compiler/compiler.cc
jonahwilliams: Done

zanderso:
I tried a bunch of shaders with this patched in, and everything seems A-OK.

PR #39356, opened 2023-02-03
Revert "[impellerc] sort uniforms on metal backend"

bdero:
Reverts flutter/engine#39345

I believe this is causing the engine roll to fail: https://github.com/flutter/flutter/pull/119894


jonahwilliams (approved):
LGTM

PR #39358, opened 2023-02-03
[Impeller] Use minimal coverage for stencil restores after overdraw prevention

bdero:
I'm hoping this'll eliminate much of the delta in https://github.com/flutter/flutter/issues/119881#issuecomment-1414812797.

Up until now, clip restores have always just covered the whole render target. But in some cases (like with overdraw prevention when drawing stroke paths), the restore area may be comparatively tiny. In the reduced test case linked above, this is a worst case scenario, because each dash takes up only a tiny sliver of pixels, and the interleaving restores repeatedly hammer the raster pipeline by stencil checking the whole screen.

Frame capture times haven't been historically reliable for me, but I seem to be getting a consistent order of magnitude difference on my iPhone 12 mini.

As an aside, I may come back around and try to refactor the overdraw prevention bits for dryness later.

Before:
Screenshot 2023-02-02 at 9 16 56 PM

After:
Screenshot 2023-02-02 at 10 26 28 PM

### DIFF
--- a/impeller/entity/contents/clip_contents.cc
+++ b/impeller/entity/contents/clip_contents.cc
@@ -136,6 +136,11 @@ ClipRestoreContents::ClipRestoreContents() = default;
 
 ClipRestoreContents::~ClipRestoreContents() = default;
 
+void ClipRestoreContents::SetRestoreCoverage(
+    std::optional<Rect> restore_coverage) {
+  restore_coverage_ = restore_coverage;
+}
+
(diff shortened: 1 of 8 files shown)


jonahwilliams (approved):
LGTM!

jonahwilliams:
Tested on the iPad pro, with this change we're down to 0.5-1ms which is inline where things are without the restore at all. I think we could call this fixed.

🎊 awesome find 🎆

PR #39359, opened 2023-02-03
[iOS] Ensure FlutterView's background color is not nil to avoid CAMetalLayer nextDrawable being time-consuming

luckysmg:
To Fix
- https://github.com/flutter/flutter/issues/103847

This looks crazy. 

I think that the view's background color is something related with CoreAnimation and `CAMetalDrawable` also related with it. And the thing draw's on it. Change it to a color not nil maybe have something to do with present behavior. So just set it to UIColor.clearcolor which is the same visually but can help reduce junk a lot.

Below video is tested on iPhone 13 Pro with local compiled engine in debug mode.
After this change the constant junk will reduce a lot on iOS.

### Before

https://user-images.githubusercontent.com/49340347/216589281-9a0b4b02-af44-419d-b186-a571e84b9022.mp4

### After
https://user-images.githubusercontent.com/49340347/216589285-18b7d207-cae9-4d70-b219-3fc589251742.mp4

- [x] I signed the [CLA].

### DIFF
--- a/shell/platform/darwin/ios/framework/Source/FlutterView.mm
+++ b/shell/platform/darwin/ios/framework/Source/FlutterView.mm
@@ -47,6 +47,11 @@
   if (self) {
     _delegate = delegate;
     self.layer.opaque = opaque;
+
+    // This line is necessary. CoreAnimation(or UIKit) may take this to do
+    // something to compute the final frame presented on screen, if we don't set this,
+    // it will make it take long time for us to take next CAMetalDrawable and will
+    // cause constant junk during rendering.
     self.backgroundColor = UIColor.clearColor;
   }
 
--- a/shell/platform/darwin/ios/framework/Source/FlutterView.mm
+++ b/shell/platform/darwin/ios/framework/Source/FlutterView.mm
@@ -47,6 +47,7 @@
   if (self) {
     _delegate = delegate;
     self.layer.opaque = opaque;
+    self.backgroundColor = UIColor.clearColor;
   }
 
   return self;
--- a/shell/platform/darwin/ios/framework/Source/FlutterViewTest.mm
+++ b/shell/platform/darwin/ios/framework/Source/FlutterViewTest.mm
@@ -49,4 +49,10 @@
   XCTAssertTrue(delegate.callbackCalled);
 }
 
+- (void)testFlutterViewBackgroundColorIsNotNil {
+  FakeDelegate* delegate = [[FakeDelegate alloc] init];
+  FlutterView* view = [[FlutterView alloc] initWithDelegate:delegate opaque:NO];
+  XCTAssertNotNil(view.backgroundColor);
+}
+
 @end


delfme:
This should also be pretty safe 👏

shell/platform/darwin/ios/framework/Source/FlutterView.mm
```diff
@@ -47,6 +47,7 @@ - (instancetype)initWithDelegate:(id<FlutterViewEngineDelegate>)delegate opaque:
   if (self) {
     _delegate = delegate;
     self.layer.opaque = opaque;
+    self.backgroundColor = UIColor.clearColor;
```
jonahwilliams, line 50: I would add a short comment here that explains why this color is important

jonahwilliams (approved):
LGTM

jonahwilliams:
Curious to see what this looks like when it rolls into the devicelab benchmarks, IIRC we have a few platform view ones there. If it doesn't work for whatever reason, well its a small change and doesn't seem risky so I think its worth a shot?

cyanglaz (approved):

jmagman:
> So just set it to UIColor.clearcolor which is the same visually but can help reduce junk a lot.

How did you figure this out?

jmagman (approved):
LGTM once the comment @jonahwilliams suggested it added.

luckysmg:
> > So just set it to UIColor.clearcolor which is the same visually but can help reduce junk a lot.
> 
> How did you figure this out?

When I was debugging a add-to-app scene. Where I set the bg color coincidentally to make UIVIewController opaque or not, I found one of them is smooth(Of course, I didn't know it is bg color to do this.....Have a very long debugging to find this....)

shell/platform/darwin/ios/framework/Source/FlutterView.mm
luckysmg: Thx for reviewing this, I will add a comment immediately^_^

luckysmg: Comment added.

luckysmg:
> Curious to see what this looks like when it rolls into the devicelab benchmarks

@jonahwilliams Would you mind sharing how to get that benchmark to see ? Maybe one link? O(∩_∩)O

jonahwilliams:
This will run on our benchmarks after its submitted, and if there are any changes we'll make sure they're shared here. That said, I'm not sure if our devicelab has phones that will exhibit this problem. I had trouble reproducing the original issue, though @cyanglaz managed to reproduce both the original issue and confirm this improves things

luckysmg:
The junk is reproduced much easier for a iphone with high refresh rate. Because they take frames to draw and present more frequently. I switch my iPhone 13 Pro to low power mode (will refresh 60HZ), the constant junk is less than low power mode off (120HZ).But it will also have constant junk

b3nni97:
Is there a technical reason why setting the background colour makes such a difference in performance?

luckysmg:
See pr desc maybe a reason.Maybe Apple do something when compositing the frame and setting bg color has something to do with it.

ayyoubelamrani4:
Do we know when this will be available in stable?

jmagman:
> Do we know when this will be available in stable?

You can follow along in the hotfix request at https://github.com/flutter/flutter/issues/120130.

jmagman:
This is available in Flutter stable 3.7.2.  It has not yet been hotfixed to the beta channel, pending at https://github.com/flutter/flutter/issues/120139

PR #39366, opened 2023-02-03
[impellerc] sort uniforms on metal backend 

jonahwilliams:
Fixes https://github.com/flutter/flutter/issues/119173, https://github.com/flutter/flutter/issues/118847

Use add_msl_resource_binding to place metal samplers in the same decoration order that we use on the SkSL backend. This should ensure that float/sampler uniforms from user authored fragment shaders behave identically across backends.


impeller/compiler/spirv_sksl.cc
```diff
@@ -219,47 +220,14 @@ bool CompilerSkSL::emit_uniform_resources() {
...
-        // we need to make sure they are emitted only after the other uniforms.
-        if (type.basetype == SPIRType::SampledImage) {
-          shader_uniforms.push_back(var.self);
-        } else {
-          regular_uniforms.push_back(var.self);
-        }
-        emitted = true;
-      }
-    }
+  std::vector<ID> regular_uniforms =
```
jonahwilliams, line 223: The revert was required because this change was filtering out non float uniforms, which we would never emit, and therefore never error check. We could move the error checking earlier in the pipeline, now that I'm more familiar with the API that is probably a better place to do it.

That said for now, the minimal change is to sort all unfiorms except sampled images instead of all float uniforms.

zanderso (approved):
lgtm

PR #39378, opened 2023-02-03
[Impeller] Retain embolden/skew font properties when rendering text glyphs

bdero:
Resolves: https://github.com/flutter/flutter/issues/119489

These options are more like implementation hacks we use to tickle Skia into rendering text the way we want. They're not generally applicable to fonts/typefaces, and so I'm working on a patch that just makes us hold onto a backend-specific font resource instead of decomposing/reconstructing from metrics. In the meantime, this patch fixes the most egregious rendering error.

I had some trouble getting a repro of this issue in a playground, but I'll add one in a follow-up if I end up getting one working.

Before:

Screen Shot 2023-02-03 at 3 20 53 AM

After:

Screen Shot 2023-02-03 at 3 01 19 PM

### DIFF
--- a/impeller/typographer/backends/skia/text_frame_skia.cc
+++ b/impeller/typographer/backends/skia/text_frame_skia.cc
@@ -27,6 +27,8 @@ static Font ToFont(const SkTextBlobRunIterator& run, Scalar scale) {
   Font::Metrics metrics;
   metrics.scale = scale;
   metrics.point_size = font.getSize();
+  metrics.embolden = font.isEmbolden();
+  metrics.skewX = font.getSkewX();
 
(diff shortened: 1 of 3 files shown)


jonahwilliams (approved):
LGTM

How does this look on the arabic text example?

jonahwilliams:
Ahh, I see in the chat you mentioned it doesn't specifically fix the bold examples. Thats fine, just curious

bdero:
Yeah, I just checked and I'm pretty sure the arabic text bug (https://github.com/flutter/flutter/issues/119805) is just a duplicate of https://github.com/flutter/flutter/issues/119234.

Impeller before/Impeller after/Skia:
image

PR #39383, opened 2023-02-04
[Impeller] Make dark text appear less emboldened

bdero:
Resolves https://github.com/flutter/flutter/issues/119234.
Resolves https://github.com/flutter/flutter/issues/119805.

I think what's going on here is that Skia applies a curve to the antialiased edges of text based on its color (see [this old chromium bug](https://bugs.chromium.org/p/chromium/issues/detail?id=898928), for example). Since we were asking Skia to draw white text, it biased the edges towards being more saturated.

By flipping the paint color from white to black, Skia's antialiasing behaves like we're drawing black text, which does the opposite.

This makes black text on a white background look almost identical to Skia, but note that white text now looks a bit too thin (seems like a better outcome than making the text too thick?).

For a more perfect solution, we can incorporate the expected text color into the glyphs. I'm experimenting with this at the moment, but this PR at least makes the most common case behave better.

Top = Impeller before
Middle = Impeller after
Bottom = Skia

Screenshot 2023-02-03 at 7 01 23 PM

Screenshot 2023-02-03 at 7 32 31 PM

### DIFF
--- a/impeller/typographer/backends/skia/text_frame_skia.cc
+++ b/impeller/typographer/backends/skia/text_frame_skia.cc
@@ -29,6 +29,7 @@ static Font ToFont(const SkTextBlobRunIterator& run, Scalar scale) {
   metrics.point_size = font.getSize();
   metrics.embolden = font.isEmbolden();
   metrics.skewX = font.getSkewX();
+  metrics.scaleX = font.getScaleX();
 
   return Font{std::move(typeface), metrics};
 }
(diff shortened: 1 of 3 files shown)


jonahwilliams:
Is the scaleX change related?

bdero:
> Is the scaleX change related?

It had no impact on the text I tested for this, but just noticed I had missed adding it while forwarding the missing properties along.

jonahwilliams (approved):
LGTM

michaelsoliman1:
Any chance this will be cherry picked to the stable channel? my production app depends heavily on arabic text and we cannot return back to skia performance 😅

PR #39445, opened 2023-02-07
[Impeller] Special case color filter in tiled texture.

jonahwilliams:
Fixes https://github.com/flutter/flutter/issues/119874

Currently there are two inefficiencies in the main wonderous carosel page:

1. We require two pass to tile the texture with opacity, then apply a color filter, these operations should be combined.
2. The final coverage is used for intermediate render passes, but tilling could be applied at the end.

Fixing these increases the FPS I see from ~90 up to 110. Getting closer!

I first attempt to generalize the passthrough code to handle tiled textures. In summary, this was a huge mess:  https://github.com/flutter/engine/compare/main...jonahwilliams:engine:uv_stuff?expand=1 

I tried to create a simpler implementation that takes advantage of the fact that the color filters we care about are almost always set on the same paint that is drawing the tiled texture. The tiled texture contents can be responsible for applying color filter in a subpass that is sized to the texture.


jonahwilliams:
TBD how we should test this, texture allocation or whatnot.

jonahwilliams:
## before

before_patch


## after 

after_patch

impeller/renderer/allocator.cc
```diff
@@ -45,6 +45,14 @@ std::shared_ptr<DeviceBuffer> Allocator::CreateBuffer(
...
 }
 
+uint64_t Allocator::GetAllocatedSize() const {
+  return allocated_size_;
+}
+
+void Allocator::ResetAllocatedSize() {
+  allocated_size_ = 0;
+}
+
```
jonahwilliams, line 55: Perhaps we should add/remove these with some ifdef.

jonahwilliams: Another idea is that if we do this only in builds for engine unit tests, we could probably track a vector of allocated sizes which make assertions easier

impeller/renderer/allocator.h
```diff
@@ -37,6 +37,15 @@ class Allocator {
   ///
   virtual uint16_t MinimumBytesPerRow(PixelFormat format) const;
 
+  // Debug/Profile only functionality for tracking the size of allocations
+#if (FLUTTER_RUNTIME_MODE != FLUTTER_RUNTIME_MODE_RELEASE)
+  const std::vector<ISize>& GetAllocatedSizes() const;
```
chinmaygarde, line 42: I am not sure if you intend to [incorporate the allocator instrumentation into this patch](https://github.com/flutter/flutter/issues/120502#issuecomment-1426545766). If it helps speeds thing along for this patch, maybe we can back this out and land this patch while working on the instrumentation separately. Your call.

jonahwilliams: I think now that I have an idea of how we will test this, I feel more comfortable backing this change out and working on it separately

impeller/entity/contents/tiled_texture_contents.h
```diff
@@ -33,11 +37,17 @@ class TiledTextureContents final : public ColorSourceContents {
...
+
  private:
+  std::optional<std::shared_ptr<Texture>> CreateFilterTexture(
+      const ContentContext& renderer) const;
+
   std::shared_ptr<Texture> texture_;
   SamplerDescriptor sampler_descriptor_ = {};
   Entity::TileMode x_tile_mode_ = Entity::TileMode::kClamp;
   Entity::TileMode y_tile_mode_ = Entity::TileMode::kClamp;
+  std::optional<ColorFilterProc> color_filter_;
```
chinmaygarde, line 50: It'd be good to note why this this is here as a naive implementation (like the one earlier) shouldn't have to care about the filter.

chinmaygarde (approved):
LGTM after backing out the allocator instrumentation is backed out. We discussed that this is would be fixed in a separate patch.

impeller/aiks/paint.cc
```diff
@@ -99,6 +99,11 @@ std::shared_ptr<Contents> Paint::WithImageFilter(
 std::shared_ptr<Contents> Paint::WithColorFilter(
     std::shared_ptr<Contents> input,
     bool absorb_opacity) const {
+  // Image input types will directly set their color filter,
+  // if any. See `TiledTextureContents.SetColorFilter`.
+  if (color_source_type == ColorSourceType::kImage) {
+    return input;
```
bdero, line 105: Won't this will break if the filter is set after the color source?

bdero: Oh, it's capturing a ref to the paint.

jonahwilliams: We can lose this optimization if its making canvas weird. now that we're handling filters better in general its not as necessary.

jonahwilliams: We can lose this optimization if its making canvas weird. now that we're handling filters better in general its not as necessary.

bdero: Nah I think it's alright. Just had to rethink a refactor I'm working on.

PR #39460, opened 2023-02-07
[macOS] Support text input autocomplete

cbracken:
By default, autocomplete is enabled during text input on macOS. On Macs with the touchbar enabled, the current text input and any suggested autocompletions are listed in the touchbar.

This adds support for disabling autocomplete when autofill is disabled, when obscureText is set in the text input configuration, and when the autofill hint type is "password" or "username". When an AutofillGroup is in use, we disable autocomplete for all fields within the group when any of the fields disables autocomplete.

While OS-level autocomplete support is far more robust on iOS, this behaviour matches out enable/disable state management behaviour on that platform.

Issue: https://github.com/flutter/flutter/issues/119824

## Pre-launch Checklist

- [X] I signed the [CLA].

### DIFF
--- a/shell/platform/darwin/macos/framework/Source/FlutterTextInputPlugin.mm
+++ b/shell/platform/darwin/macos/framework/Source/FlutterTextInputPlugin.mm
@@ -20,6 +20,7 @@
 
 static NSString* const kTextInputChannel = @"flutter/textinput";
 
+#pragma mark - Textinput channel method names
 // See https://api.flutter.dev/flutter/services/SystemChannels/textInput-constant.html
 static NSString* const kSetClientMethod = @"TextInput.setClient";
 static NSString* const kShowMethod = @"TextInput.show";
@@ -35,14 +36,12 @@ static NSString* const kPerformAction = @"TextInputClient.performAction";
 static NSString* const kPerformSelectors = @"TextInputClient.performSelectors";
 static NSString* const kMultilineInputType = @"TextInputType.multiline";
 
-static NSString* const kTextAffinityDownstream = @"TextAffinity.downstream";
-static NSString* const kTextAffinityUpstream = @"TextAffinity.upstream";
-
+#pragma mark - TextInputConfiguration field names
+static NSString* const kSecureTextEntry = @"obscureText";
 static NSString* const kTextInputAction = @"inputAction";
 static NSString* const kEnableDeltaModel = @"enableDeltaModel";
 static NSString* const kTextInputType = @"inputType";
 static NSString* const kTextInputTypeName = @"name";
-
 static NSString* const kSelectionBaseKey = @"selectionBase";
 static NSString* const kSelectionExtentKey = @"selectionExtent";
 static NSString* const kSelectionAffinityKey = @"selectionAffinity";
@@ -51,6 +50,17 @@ static NSString* const kComposingBaseKey = @"composingBase";
 static NSString* const kComposingExtentKey = @"composingExtent";
 static NSString* const kTextKey = @"text";
 static NSString* const kTransformKey = @"transform";
+static NSString* const kAssociatedAutofillFields = @"fields";
+
+// TextInputConfiguration.autofill and sub-field names
+static NSString* const kAutofillProperties = @"autofill";
+static NSString* const kAutofillId = @"uniqueIdentifier";
+static NSString* const kAutofillEditingValue = @"editingValue";
+static NSString* const kAutofillHints = @"hints";
+
+// TextAffinity types
+static NSString* const kTextAffinityDownstream = @"TextAffinity.downstream";
+static NSString* const kTextAffinityUpstream = @"TextAffinity.upstream";
 
 /**
  * The affinity of the current cursor position. If the cursor is at a position representing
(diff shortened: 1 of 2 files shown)


shell/platform/darwin/macos/framework/Source/FlutterTextInputPlugin.mm
```diff
@@ -77,6 +87,53 @@ typedef NS_ENUM(NSUInteger, FlutterTextAffinity) {
...
+  NSString* contentType = GetAutofillContentType(configuration);
+  if ([contentType isEqualToString:@"password"] || [contentType isEqualToString:@"username"]) {
+    return NO;
+  }
+
+  // Disable if autofill properties specify a blank autofill type.
+  if ([contentType isEqualToString:@""]) {
+    return NO;
+  }
+  return YES;
```
cbracken, line 116: This block could just be a `return ![contentType ...];` but this reads very marginally better IMHO.

shell/platform/darwin/macos/framework/Source/FlutterTextInputPlugin.mm
```diff
@@ -77,6 +87,53 @@ typedef NS_ENUM(NSUInteger, FlutterTextAffinity) {
...
+    return NO;
+  }
+
+  // Disable if autofill properties indicate a username/password.
+  NSString* contentType = GetAutofillContentType(configuration);
+  if ([contentType isEqualToString:@"password"] || [contentType isEqualToString:@"username"]) {
+    return NO;
+  }
+
+  // Disable if autofill properties specify a blank autofill type.
```
loic-sharma, line 112: When is the autofill type blank? Consider adding a comment, but I don't feel strongly

shell/platform/darwin/macos/framework/Source/FlutterTextInputPlugin.mm
```diff
@@ -77,6 +87,53 @@ typedef NS_ENUM(NSUInteger, FlutterTextAffinity) {
...
+// the specified TextInputConfiguration. Autocomplete is enabled by default, but will be disabled
+// if the field is password-related, or if the autofill hint is the empty string.
+static BOOL EnableAutocompleteForTextInputConfiguration(NSDictionary* configuration) {
+  // Disable if obscureText is set.
+  if ([configuration[kSecureTextEntry] boolValue]) {
+    return NO;
+  }
+
+  // Disable if autofill properties indicate a username/password.
+  NSString* contentType = GetAutofillContentType(configuration);
```
loic-sharma, line 107: Consider checking for `nil` here if there are no autofill content types.

loic-sharma (approved):
LGTM but please also get approval from someone that's more familiar with the text input plugin

shell/platform/darwin/macos/framework/Source/FlutterTextInputPlugin.mm
cbracken: In Obj-C, sending a message to nil always results in nil, so it's more idiomatic just to go ahead and fire the message. In this case, the ensuing checks will all be false in that case.

shell/platform/darwin/macos/framework/Source/FlutterTextInputPlugin.mm
cbracken: Nice catch. I actually had the same question, spent a bunch of time poking at five layers deep of blame history this morning, and finally got the right answer from @LongCatIsLooong, then promptly forgot to add the doc.

The answer is that while setting obscureText or a content type of password/username disables autofill, sometimes you want to disable it explicitly. This is how we do it.

shell/platform/darwin/macos/framework/Source/FlutterTextInputPlugin.mm
```diff
@@ -77,6 +87,54 @@ typedef NS_ENUM(NSUInteger, FlutterTextAffinity) {
...
+  if ([contentType isEqualToString:@"password"] || [contentType isEqualToString:@"username"]) {
+    return NO;
+  }
+
+  // Disable if autofill properties specify a blank autofill type. The framework sends blank content
+  // type when autofill should be disabled.
+  if ([contentType isEqualToString:@""]) {
```
LongCatIsLooong, line 114: Sorry I wasn't clear the framework doesn't use `@""` to indicate for disabling autofill, that is specific to iOS: https://github.com/flutter/engine/blob/bf8547dd9bcdf14c96ac778a24d36072f038e789/shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm#L820-L826

The framework will send you an input configuration without an `autofill` field

cbracken: Ah, sorry I'd understood that as a framework implementation detail when we talked. In that case, I'll kill that edge-case. 

Any idea why we do that in the iOS embedder? It seems a little overly complicated as yet another means of disabling autocorrect.

LongCatIsLooong: it's for disabling autofill. In the framework there's an option to explicitly disable autofill but there's no real API to disable autofill on iOS per text field so IIRC that's a workaround for that.

LongCatIsLooong: https://developer.apple.com/documentation/uikit/uitextinputtraits/1624453-autocorrectiontype is for disabling autocorrect

cbracken: I think we'll be in the same boat for macOS will we not? The available content types for macOS are a small subset of those available for iOS, but 'none' is not one of them:
* macOS: https://developer.apple.com/documentation/appkit/nstextcontenttype?language=objc#
* iOS: https://developer.apple.com/documentation/uikit/uitextcontenttype?language=objc#

cbracken: Updated to avoid this, and to disable autofill when autocorrect is disabled, similar to iOS.

shell/platform/darwin/macos/framework/Source/FlutterTextInputPlugin.mm
```diff
@@ -77,6 +87,53 @@ typedef NS_ENUM(NSUInteger, FlutterTextAffinity) {
...
+  NSString* contentType = GetAutofillContentType(autofill);
+  if ([contentType isEqualToString:@"password"] || [contentType isEqualToString:@"username"]) {
+    return NO;
+  }
+  return YES;
+}
+
+// Returns YES if configuration describes a field for which autocomplete should be enabled.
+// Autocomplete is enabled by default, but will be disabled if the field is password-related, or if
+// the configuration contains no autofill settings.
```
LongCatIsLooong, line 121: nit: maybe link to the issue or provide a simple explanation as to why this is disabled for username/passwords?

LongCatIsLooong (approved):
LGTM

shell/platform/darwin/macos/framework/Source/FlutterTextInputPlugin.mm
cbracken: Good call; done. Also filed https://github.com/flutter/flutter/issues/120252, which should be relatively straightforward and nice to have.

PR #39468, opened 2023-02-08
Fix FlView self variables using non-standard names.

robert-ancell:
By convention, use `self` for the pointer to the current object. Fixes this inconsistency in FlView.


cbracken (approved):

Hixie:

PR #39467, opened 2023-02-08
Simplify FlView

robert-ancell:
This PR simplifies the FlView GL interactions. It removes some code paths that were never being used (possibly for future use with platform views?) and complex internal APIs that could be done in a single method call. This is motivated in investigating https://github.com/ubuntu-flutter-community/software/issues/868.

I plan to take this further, since there doesn't seem to be any reason why multiple FlGLArea widgets are used - we can draw all the layers in one widget.

For reviewers - the individual commits aren't particularly interesting, I just did it that way to be sure I wasn't breaking anything along the way and help to diagnose any regressions in the future.

The structural changes that this PR is simplifying came in f94cf140c52c0cc3769d9281d5579e935886fbac.

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [Flutter Style Guide] and the [C++, Objective-C, Java style guides].
- [ ] I listed at least one issue that this PR fixes in the description above.
- [ ] I added new tests to check the change I am making or feature I am adding, or Hixie said the PR is test-exempt. See [testing the engine] for instructions on writing and running engine tests.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I signed the [CLA].
- [x] All existing and new tests are passing.

PR #39473, opened 2023-02-08
Use a single FlGLArea

robert-ancell:
Further simplification from https://github.com/flutter/engine/pull/39467

### DIFF
--- a/shell/platform/linux/public/flutter_linux/fl_view.h
+++ b/shell/platform/linux/public/flutter_linux/fl_view.h
@@ -16,7 +16,7 @@
 
 G_BEGIN_DECLS
 
-G_DECLARE_FINAL_TYPE(FlView, fl_view, FL, VIEW, GtkContainer)
+G_DECLARE_FINAL_TYPE(FlView, fl_view, FL, VIEW, GtkBox)
 
 /**
  * FlView:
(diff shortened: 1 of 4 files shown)


robert-ancell:
@jpnurmi please try this and see if it resolved your first frame issue.

robert-ancell:
Fast clicking and dragging on the (+) button is causing crashes with this branch, which don't seem reproducible with the previous branches.

robert-ancell:
The crashes seem to be related to getting leave events after releasing the button. I don't know why GTK is taking focus away :/

jpnurmi:
Thanks! I'm no longer able to reproduce the blank window issue with this.

robert-ancell:
Fixed crash, should be good to review once https://github.com/flutter/engine/pull/39467 is landed.

shell/platform/linux/public/flutter_linux/fl_view.h
```diff
@@ -16,7 +16,7 @@
 
 G_BEGIN_DECLS
 
-G_DECLARE_FINAL_TYPE(FlView, fl_view, FL, VIEW, GtkContainer)
+G_DECLARE_FINAL_TYPE(FlView, fl_view, FL, VIEW, GtkBox)
```
robert-ancell, line 19: While this is changing the public API, I don't think this causes any ongoing issues - the class couldn't be derived and it only accessed via pointers and functions. There is no reason for an app to be looking inside this widget, and since it is a GtkBox it is still a GtkContainer.

cbracken:
Looks like this will need to be rebased to tip-of-tree to deal with a few conflicts.

cbracken (approved):

PR #39483, opened 2023-02-08
[Impeller] Fix sampling management problems

bdero:
Resolves https://github.com/flutter/flutter/issues/118614.
Resolves https://github.com/gskinnerTeam/flutter-wonderous-app/issues/69.

* Add a `Contents::RenderToSnapshot` sampler descriptor param, so that `TextureContents::RenderToSnapshot` can apply the right sampler settings even when it's unable to passthrough.
* Make the blend filter apply the snapshot's sampling mode when rendering. Then, use nearest sampling for the final render (since the blend filter absorbs the transform).
* Treat the Cubic DL sampling mode as Linear instead of falling back to Nearest. Impeller doesn't support `kCubic` sampling (and perhaps we shouldn't).
* Make a default snapshot SamplerDescriptor have a clear label so that it's easy to spot in GPU captures.

### DIFF
--- a/impeller/display_list/display_list_dispatcher.cc
+++ b/impeller/display_list/display_list_dispatcher.cc
@@ -136,6 +136,9 @@ static impeller::SamplerDescriptor ToSamplerDescriptor(
       desc.label = "Nearest Sampler";
       break;
     case flutter::DlImageSampling::kLinear:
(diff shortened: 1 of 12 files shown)


jonahwilliams (approved):
LGTM

chinmaygarde (approved):

PR #39490, opened 2023-02-08
Revert "[ios_platform_view] MaskView pool to reuse maskViews."

cyanglaz:
Reverts flutter/engine#38989

This revert is to try confirming the original commit causing a benchmark regression, see: https://github.com/flutter/flutter/issues/120207


jonahwilliams (approved):
LGTM

PR #39491, opened 2023-02-08
[Impeller] specialize decal lookup in advanced blend shader.

jonahwilliams:
From my previous investigation of shader performance, I found that providing a constant to IPSampleWithTileMode was not sufficient to cause the other code paths to be compiled out. instead use a  specialized sample function


jonahwilliams:
Will recreate, easier than figuring out the rebase

PR #39506, opened 2023-02-09
[web] Fixes invisible platform view compositing edge cases.

ditman:
This PR fixes a couple of edge case with the canvaskit platform view embedder where if an app would have an "invisible" platform view as its first platform view in the scene, it would prevent "visible" views from being composited at the right spot of the DOM.

The issue is fixed by:

* Simplifying the logic to determine whether a canvaskit overlay is needed (it's now only triggered by visible platform views) in the preroll/composite stages.

The above introduces a new corner case, for apps that only contain invisible platform views, that is fixed by:

* Excluding platform view groups without any visible platform view from the "viewsNeedingOverlays" computation when updating the overlays in the submitFrame stage.

### Related issues

* Fixes https://github.com/flutter/flutter/issues/118452

### Testing

* Unit tests are updated with the new behavior where apps with only invisible views don't render a second overlay at all.
* Our example apps for Link Widget and Pointer Interceptor seem to work fine with the change.

This PR also modifies the test matcher used to test this feature to improve its assertions (expected and actual were reversed, actual values look now very similar to the expected ones, and error messages are easier to understand), and to make test cases easier to find by adding an optional `reason` field.

<details>
<summary>Click to expand sample test failure output before and after this PR</summary>

**Before**

```
00:04 +17 -1: HtmlViewEmbedder does not create overlays for invisible platform views [E]                          
  Expected: an object with length of <4>
    Actual: [
              _EmbeddedViewMarker:_EmbeddedViewMarker.overlay,
              _EmbeddedViewMarker:_EmbeddedViewMarker.platformView,
              _EmbeddedViewMarker:_EmbeddedViewMarker.platformView
            ]
     Which: has length of <3>
```

Note that the "Actual" above is the *expected* in the test. See below.

**After**

```
00:04 +17 -1: HtmlViewEmbedder does not create overlays for invisible platform views [E]                          
  Expected: [
              _EmbeddedViewMarker:_EmbeddedViewMarker.overlay,
              _EmbeddedViewMarker:_EmbeddedViewMarker.platformView,
              _EmbeddedViewMarker:_EmbeddedViewMarker.platformView
            ]
    Actual: [
              _EmbeddedViewMarker:_EmbeddedViewMarker.overlay,
              _EmbeddedViewMarker:_EmbeddedViewMarker.platformView,
              _EmbeddedViewMarker:_EmbeddedViewMarker.platformView,
              _EmbeddedViewMarker:_EmbeddedViewMarker.overlay
            ]
     Which: at location [3] is [
              _EmbeddedViewMarker:_EmbeddedViewMarker.overlay,
              _EmbeddedViewMarker:_EmbeddedViewMarker.platformView,
              _EmbeddedViewMarker:_EmbeddedViewMarker.platformView,
              _EmbeddedViewMarker:_EmbeddedViewMarker.overlay
            ] which longer than expected
  Overlay created after a group containing a visible view.
```

</summary>

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/lib/web_ui/lib/src/engine/canvaskit/embedded_views.dart
+++ b/lib/web_ui/lib/src/engine/canvaskit/embedded_views.dart
@@ -133,9 +133,8 @@ class HtmlViewEmbedder {
           'displayed at once. '
           'You may experience incorrect rendering.');
     }
-    // We need an overlay for the first platform view no matter what. The first
-    // visible platform view doesn't need to create a new one if we already
-    // created one.
+    // We need an overlay for each visible platform view. Invisible platform
+    // views will be grouped with (at most) one visible platform view later.
     final bool needNewOverlay = platformViewManager.isVisible(viewId);
     if (needNewOverlay && hasAvailableOverlay) {
       final CkPictureRecorder pictureRecorder = CkPictureRecorder();
@@ -164,10 +163,11 @@ class HtmlViewEmbedder {
   CkCanvas? compositeEmbeddedView(int viewId) {
     final int overlayIndex = _context.visibleViewCount;
     _compositionOrder.add(viewId);
+    // Keep track of the number of visible platform views.
     if (platformViewManager.isVisible(viewId)) {
       _context.visibleViewCount++;
     }
-    // We need a new overlay if this is a visible view or if we don't have one yet.
+    // We need a new overlay if this is a visible view.
     final bool needNewOverlay = platformViewManager.isVisible(viewId);
     CkPictureRecorder? recorderToUseForRendering;
     if (needNewOverlay) {
@@ -415,8 +415,8 @@ class HtmlViewEmbedder {
     _updateOverlays(diffResult);
     assert(
       _context.pictureRecorders.length == _overlays.length,
-      'There should be the same picture recorders (${_context.pictureRecorders.length}) '
-      'than overlays (${_overlays.length}).',
+      'There should be the same number of picture recorders '
+      '(${_context.pictureRecorders.length}) as overlays (${_overlays.length}).',
     );
     int pictureRecorderIndex = 0;
 
@@ -593,15 +593,13 @@ class HtmlViewEmbedder {
       // overlays.
       return;
     }
-    final List<List<int>> overlayGroups = getOverlayGroups(_compositionOrder);
-    final List<int> viewsNeedingOverlays = overlayGroups
-        .where(
-          (List<int> group) => group.any(
-            (int viewId) => platformViewManager.isVisible(viewId)
-          )
-        )
-        .map((List<int> group) => group.last)
-        .toList();
+    // Group platform views from their composition order.
+    // Each group contains one visible view, and any number of invisible views
+    // before or after that visible view.
+    final List<OverlayGroup> overlayGroups =
+        getOverlayGroups(_compositionOrder);
+    final List<int> viewsNeedingOverlays =
+        overlayGroups.map((OverlayGroup group) => group.last).toList();
     // If there were more visible views than overlays, then the last group
     // doesn't have an overlay.
     if (viewsNeedingOverlays.length > SurfaceFactory.instance.maximumOverlays) {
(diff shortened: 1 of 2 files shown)


lib/web_ui/lib/src/engine/canvaskit/embedded_views.dart
```diff
@@ -424,7 +413,11 @@ class HtmlViewEmbedder {
             ? null
             : diffViewList(_activeCompositionOrder, _compositionOrder);
     _updateOverlays(diffResult);
-    assert(_context.pictureRecorders.length == _overlays.length);
+    assert(
+      _context.pictureRecorders.length == _overlays.length,
+      'There should be the same picture recorders (${_context.pictureRecorders.length}) '
+      'than overlays (${_overlays.length}).',
```
Juliotati, line 419: I'm finding having `and` as opposed to `than` in the assert makes it easier to read and understand.

Reading `than`, makes me suspect that the assert is supposed to assert that:

- `_context.pictureRecorders.length` is supposed to be greater/lower **`than`** `_overlays.length`
- and not that `_context.pictureRecorders.length` and `_overlays.length` should be the same. _(the intended)_

Makes sense?

lib/web_ui/lib/src/engine/canvaskit/embedded_views.dart
```diff
@@ -601,8 +594,14 @@ class HtmlViewEmbedder {
...
-    final List<int> viewsNeedingOverlays =
-        overlayGroups.map((List<int> group) => group.last).toList();
+    final List<int> viewsNeedingOverlays = overlayGroups
+        .where(
+          (List<int> group) => group.any(
+            (int viewId) => platformViewManager.isVisible(viewId)
+          )
+        )
+        .map((List<int> group) => group.last)
+        .toList();
```
yjbanov, line 604: I don't know if this is on a performance hot path, but keep in mind that each `where`, `any`, and `map` cost an extra closure and a lazy-evaluated `Iterable` wrapper.

I'm also a little confused by the "overlay group" concept. If it's an overlay group, then presumably it already needs an overlay (or _is_ an overlay). Maybe these should be called `viewGoups` (and similarly the function `getOverlayGroups` be renamed to `getViewGroups`)?

Even better might be to introduce a `ViewGroup` class that has a `containsVisiblePlatformViews` boolean flag that's populated by `getOverlayGroups` as it's iterating over all views.

lib/web_ui/lib/src/engine/canvaskit/embedded_views.dart
ditman: I quickly ran grammarly through this, and it definitely prefers "same number of" instead of just "same" (I lost energy typing there), and it doesn't care about "as" or "and", but it does flag "than" as wrong, so

"There should be the same number of picture recorders (5) as overlays (7)." it is!

lib/web_ui/lib/src/engine/canvaskit/embedded_views.dart
```diff
@@ -136,12 +136,7 @@ class HtmlViewEmbedder {
     // We need an overlay for the first platform view no matter what. The first
     // visible platform view doesn't need to create a new one if we already
     // created one.
-    final bool needNewOverlay = (platformViewManager.isVisible(viewId) &&
```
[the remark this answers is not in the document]
ditman: Updated comments about only needing an overlay for visible platform views.

lib/web_ui/lib/src/engine/canvaskit/embedded_views.dart
ditman: I like the ViewGroup idea, let me see how much breakage I introduce :)

hterkelsen (approved):
LGTM

lib/web_ui/lib/src/engine/canvaskit/embedded_views.dart
ditman: I added a small OverlayGroup class that should help this bit of the code.

Juliotati (commented):
LGTMT

yjbanov (approved):
lgtm

PR #39516, opened 2023-02-09
Revert "Remove deprecated TextInputClient scribble method code"

justinmc:
The framework side of this was reverted long ago and still hasn't been relanded (https://github.com/flutter/flutter/pull/115296).  The result is that scribble does not currently work.  This PR will add back the original method channel calls so that it works in the meantime until the framework PR can be relanded.

### DIFF
--- a/shell/platform/darwin/ios/framework/Source/FlutterEngine.mm
+++ b/shell/platform/darwin/ios/framework/Source/FlutterEngine.mm
@@ -986,6 +986,10 @@ static void SetEntryPoint(flutter::Settings* settings, NSString* entrypoint, NSS
 #pragma mark - FlutterViewEngineDelegate
 
 - (void)flutterTextInputView:(FlutterTextInputView*)textInputView showToolbar:(int)client {
+  // TODO(justinmc): Remove the TextInputClient usage when the framework has
+  // finished transitioning to using the Scribble channel.
+  // https://github.com/flutter/flutter/pull/104128
+  [_textInputChannel.get() invokeMethod:@"TextInputClient.showToolbar" arguments:@[ @(client) ]];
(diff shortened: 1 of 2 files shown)


jmagman:
To add to the paper trail, this is reverting #36970 and I think will fix https://github.com/flutter/flutter/issues/119180.

Once this lands it's worth a hotfix.

jmagman (approved):
LGTM, thanks @justinmc!

jmagman:
@justinmc this should also fix https://github.com/flutter/flutter/issues/119590, right?

chinmaygarde (approved):

justinmc:
@jmagman Yes I believe this will fix https://github.com/flutter/flutter/issues/119590 too.

Looking for a test exemption.

Hixie:
test-exempt: reverting to known good state

...but really this should have tests, because the only reason the thing it's reverting was test-exempt itself is that it was removing untested code...

justinmc:
+1, the new Scribble-specific methods are tested and I plan to migrate back over to them ASAP.

justinmc:
I intend to try out Scribble on a real iPad + Apple Pencil before I merge this.

jmagman:
Validated https://github.com/flutter/flutter/issues/119590 no longer reproduces on 0af8f79

justinmc:
Awesome, thank you!  Something wasn't happy with the duplicate calls.  I'm going to take a quick look at if there's anything else I can do to make the transition back to the Scribble channel go smoothly, otherwise I'll just merge this in a few minutes.

justinmc:
I didn't end up with any ideas that worked for a perfect transition.  I'll figure that out when I get to it.  Merging for now.

PR #39517, opened 2023-02-09
Fix classes that shouldn't be extended/instantiated/mixedin

goderbauer:
This was discovered when investigating which classes could potentially be used as a mixin and may need a mixin class modifier in Dart 3. These classes came up as potential mixin candidates even though they really shouldn't be candidates.


chinmaygarde (approved):

Hixie:

PR #39523, opened 2023-02-09
[Impeller] use IPSampleDecal in advanced blends.

jonahwilliams:
Slightly improves performance and tests malioc on CI


chinmaygarde (approved):

impeller/tools/malioc_diff.py
```diff
@@ -116,7 +116,10 @@ def read_malioc_file(malioc_tree, json_file):
     if shader['hardware']['core'] not in CORES:
       continue
     result = {}
-    result['filename'] = os.path.relpath(shader['filename'], build_gen_dir)
+    filename = os.path.relpath(shader['filename'], build_gen_dir)
```
jonahwilliams, line 119: @zanderso I tried to get the paths right with various combinations of relative and absolute paths, plus a few different working directories, but no dice.

This seems like a harmless change though.

zanderso: What were `shader['filename']` and `build_gen_dir` here when you ran from a cwd of `src/` (that is the directory containing `out`) ?

jonahwilliams: build_gen_dir: out/android_debug_arm64/gen/malioc
filename: ../flutter/impeller/scene/shaders/gles/unskinned.vert.gles

PR #29930, opened 2021-11-24
[ios platform view] fix overlay zPosition

cyanglaz:
Explicitly set every platform view and overlay view's zPosition so we are sure the overlays are on top of the platform views.

Fixes: https://github.com/flutter/flutter/issues/86787

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [Flutter Style Guide] and the [C++, Objective-C, Java style guides].
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I added new tests to check the change I am making or feature I am adding, or Hixie said the PR is test-exempt. See [testing the engine] for instructions on
writing and running engine tests.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I signed the [CLA].
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/flutter/wiki/Tree-hygiene#overview
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[C++, Objective-C, Java style guides]: https://github.com/flutter/engine/blob/main/CONTRIBUTING.md#style
[testing the engine]: https://github.com/flutter/flutter/wiki/Testing-the-engine
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/wiki/Chat

PR #39527, opened 2023-02-09
[platform_view] fix overlapping platform view not touchable

hellohuanlin:
Note: Added integration test [here](https://github.com/flutter/flutter/pull/120844), but needs engine roll of this PR in order to pass that test. 

It turns out that the rendering Z order was correct, but the touch behavior is incorrect. So instead of setting the `zPosition`, we actually change subview order in the view hierarchy, which works for both z order and touch behavior. 

Added a unit test for this particular issue. The existing scenario/screenshot tests should already be sufficient. 

https://github.com/flutter/flutter/issues/118366

## Pre-launch Checklist

- [x ] I read and followed the [Flutter Style Guide] and the [C++, Objective-C, Java style guides].
- [x] I signed the [CLA].

### DIFF
--- a/shell/platform/darwin/ios/framework/Source/FlutterPlatformViews.mm
+++ b/shell/platform/darwin/ios/framework/Source/FlutterPlatformViews.mm
@@ -777,7 +777,6 @@ void FlutterPlatformViewsController::BringLayersIntoView(LayersMap layer_map) {
     int64_t platform_view_id = composition_order_[i];
     std::vector<std::shared_ptr<FlutterPlatformViewLayer>> layers = layer_map[platform_view_id];
     UIView* platform_view_root = root_views_[platform_view_id].get();
-    // Simply having a higher zPosition is not enough since it does not affect touch behavior.
     // `addSubview` will automatically reorder subview if it is already added.
     [flutter_view addSubview:platform_view_root];
     for (const std::shared_ptr<FlutterPlatformViewLayer>& layer : layers) {
--- a/shell/platform/darwin/ios/framework/Source/FlutterPlatformViews.mm
+++ b/shell/platform/darwin/ios/framework/Source/FlutterPlatformViews.mm
@@ -771,29 +771,17 @@ bool FlutterPlatformViewsController::SubmitFrame(GrDirectContext* gr_context,
 void FlutterPlatformViewsController::BringLayersIntoView(LayersMap layer_map) {
   FML_DCHECK(flutter_view_);
   UIView* flutter_view = flutter_view_.get();
-  auto zIndex = 0;
   // Clear the `active_composition_order_`, which will be populated down below.
   active_composition_order_.clear();
   for (size_t i = 0; i < composition_order_.size(); i++) {
     int64_t platform_view_id = composition_order_[i];
     std::vector<std::shared_ptr<FlutterPlatformViewLayer>> layers = layer_map[platform_view_id];
     UIView* platform_view_root = root_views_[platform_view_id].get();
-
-    if (platform_view_root.superview != flutter_view) {
-      [flutter_view addSubview:platform_view_root];
-    }
-    // Make sure the platform_view_root is higher than the last platform_view_root in
-    // composition_order_.
-    platform_view_root.layer.zPosition = zIndex++;
-
+    // Simply having a higher zPosition is not enough since it does not affect touch behavior.
+    // `addSubview` will automatically reorder subview if it is already added.
+    [flutter_view addSubview:platform_view_root];
     for (const std::shared_ptr<FlutterPlatformViewLayer>& layer : layers) {
-      if ([layer->overlay_view_wrapper.get() superview] != flutter_view) {
-        [flutter_view addSubview:layer->overlay_view_wrapper];
-      }
-      // Make sure all the overlays are higher than the platform view.
-      layer->overlay_view_wrapper.get().layer.zPosition = zIndex++;
-      FML_DCHECK(layer->overlay_view_wrapper.get().layer.zPosition >
-                 platform_view_root.layer.zPosition);
+      [flutter_view addSubview:layer->overlay_view_wrapper];
     }
     active_composition_order_.push_back(platform_view_id);
   }
--- a/testing/ios/IosUnitTests/Tests/FlutterEngineConfig.xcconfig
+++ b/testing/ios/IosUnitTests/Tests/FlutterEngineConfig.xcconfig
@@ -1,2 +1,3 @@
 FLUTTER_ENGINE[arch=x86_64]=ios_debug_sim_unopt
 FLUTTER_ENGINE[arch=arm64]=ios_debug_sim_unopt_arm64
+FLUTTER_ENGINE=ios_debug_sim_unopt
--- a/testing/scenario_app/lib/src/platform_view.dart
+++ b/testing/scenario_app/lib/src/platform_view.dart
@@ -1145,7 +1145,7 @@ class PlatformViewForOverlappingPlatformViewsScenario extends Scenario
   @override
   void onDrawFrame() {
     _frameCount += 1;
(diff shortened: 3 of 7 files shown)


shell/platform/darwin/ios/framework/Source/FlutterPlatformViews.mm
```diff
@@ -744,18 +744,19 @@ static bool ClipBoundsContainsPlatformViewBoundingRect(const SkRect& clip_bounds
     int64_t platform_view_id = composition_order_[i];
     std::vector<std::shared_ptr<FlutterPlatformViewLayer>> layers = layer_map[platform_view_id];
     UIView* platform_view_root = root_views_[platform_view_id].get();
-
-    if (platform_view_root.superview != flutter_view) {
-      [flutter_view addSubview:platform_view_root];
-    }
+    //    if (platform_view_root.superview == flutter_view) {
+    //      [platform_view_root removeFromSuperview];
```
hellohuanlin, line 748: @cyanglaz it turns out it's the `removeFromSuperview` call making the platform view not touchable. So now we simply `addSubview`, which just reorder the subview if already added.

loic-sharma:
@hellohuanlin Would you mind putting slightly more information in the pull request title even if it is a work in progress? This makes it easier to manage our inboxes 😄

hellohuanlin:
@loic-sharma oops. sorry i was not expecting this to trigger any email notifications yet. it just came from an offline discussion. Let me make it a draft.

testing/ios/IosUnitTests/Tests/FlutterEngineConfig.xcconfig
```diff
@@ -1,2 +1,3 @@
 FLUTTER_ENGINE[arch=x86_64]=ios_debug_sim_unopt
 FLUTTER_ENGINE[arch=arm64]=ios_debug_sim_unopt_arm64
+FLUTTER_ENGINE=ios_debug_sim_unopt
```
hellohuanlin, line 3: @jmagman Not sure if you want to keep this or fix it in a separate PR?

shell/platform/darwin/ios/framework/Source/FlutterPlatformViews.mm
```diff
@@ -786,29 +786,17 @@ static bool ClipRRectContainsPlatformViewBoundingRect(const SkRRect& clip_rrect,
...
     UIView* platform_view_root = root_views_[platform_view_id].get();
-
-    if (platform_view_root.superview != flutter_view) {
-      [flutter_view addSubview:platform_view_root];
-    }
-    // Make sure the platform_view_root is higher than the last platform_view_root in
-    // composition_order_.
-    platform_view_root.layer.zPosition = zIndex++;
-
+    // Simply having a higher zPosition is not enough since it does not affect touch behavior.
```
jmagman, line 795: I don't think this comment about zindex makes sense without also looking at the previous code, since now the current code has no reference to a zindex.
I think you can just remove this line, the point is the next sentence: it will be reordered if it's already added.

shell/platform/darwin/ios/framework/Source/FlutterPlatformViews.mm
```diff
@@ -786,29 +786,17 @@ static bool ClipRRectContainsPlatformViewBoundingRect(const SkRRect& clip_rrect,
...
+    [flutter_view addSubview:platform_view_root];
     for (const std::shared_ptr<FlutterPlatformViewLayer>& layer : layers) {
-      if ([layer->overlay_view_wrapper.get() superview] != flutter_view) {
-        [flutter_view addSubview:layer->overlay_view_wrapper];
-      }
-      // Make sure all the overlays are higher than the platform view.
-      layer->overlay_view_wrapper.get().layer.zPosition = zIndex++;
-      FML_DCHECK(layer->overlay_view_wrapper.get().layer.zPosition >
-                 platform_view_root.layer.zPosition);
+      [flutter_view addSubview:layer->overlay_view_wrapper];
```
jmagman, line 799: Can you manually test that this change didn't regress https://github.com/flutter/flutter/issues/86787 (see #29930)

testing/ios/IosUnitTests/Tests/FlutterEngineConfig.xcconfig
jmagman: Bleh may as well keep it, my attempt to do the right thing based on architecture was totally thwarted by Xcode passing in `ARCH=unknown` at certain build steps.

cyanglaz (approved):
LGTM

hellohuanlin:
@cyanglaz Bring the discussion back here since that PR is stamped. Let me know what you think about https://github.com/flutter/flutter/pull/120844#issuecomment-1433636433

cyanglaz:
The framework integration test does not add additional verification of the engine behavior. Since we know the issue was due to re-ordering the platform views and the bug is in the engine. The scenario test is sufficient to test against this bug.
The framework integration test is still valuable as I commented 

>(1) In an ideal world, these 2 tests verify the same engine behavior; however, in practice I'd argue that scenario tests will not give us as much confidence about engine behavior, since we could made mistakes mimicking the FakeUI behavior. (e.g. engine may pass the test, but under a wrong FakeUI command)

The FakeUI command only needs to create the necessary composition to reproduce the bug, we don't need to match the framework exactly.

> (2) Another drawback with scenario test is that framework is not part of the integration. it doesn't give any confidence that the UiKitView is indeed tappable, which is the very purpose of this bugfix.

The PlatformView can listen to the tap event, right?

hellohuanlin:
> The PlatformView can listen to the tap event, right?

@cyanglaz Not sure if i follow this. Are you talking about dart side? I think I am able to create 2 frames and reorder platform views using `addPlatformView` ([like this]https://github.com/flutter/engine/blob/7db8c2bb876e52995cd9fb27eba5240567e54545/testing/scenario_app/lib/src/platform_view.dart#L268-L273)), but may need some help with the tapping part 😆

cyanglaz:
Yes, here is one example: https://github.com/flutter/engine/blob/main/testing/scenario_app/lib/src/platform_view.dart#L992-L1017.

When you send `acceptGesture`, the platform view will receive the gesture.

hellohuanlin:
@cyanglaz just looked into `onPointerDataPacket` and it makes sense. Will add the scenario test here.

testing/scenario_app/lib/src/platform_view.dart
```diff
@@ -1001,6 +1001,133 @@ class PlatformViewForTouchIOSScenario extends Scenario
...
+    window.render(scene);
+    scene.dispose();
+  }
+
+  int _frameCount = 0;
+
+  @override
+  void onDrawFrame() {
+    _frameCount += 1;
+    // TODO: (hellohuanlin) Need further investigation - the first 2 frames are dropped for some reason.
```
jmagman, line 1093:
```
   info - scenario_app/lib/src/platform_view.dart:1093:5 - To-do comment doesn't follow the Flutter style. Try following the Flutter style for to-do comments. - flutter_style_todos
```

cyanglaz (approved):
LGTM! Thanks for looking into the scenario test.

shell/platform/darwin/ios/framework/Source/FlutterPlatformViews.mm
hellohuanlin: I can confirm that it didn't regress https://github.com/flutter/flutter/issues/86787. 

Though I found a different regression (see pic below). The navigation bar becomes transparent, so the platform view is displayed underneath. This seems to be similar to https://github.com/flutter/flutter/issues/119485. 

It is NOT caused by this PR because I get the same behaviors with or without this change; but it is a regression, because the released build is working fine. I will comment in that issue about this finding. 

Simulator Screen Shot - iPhone 14 Pro - 2023-02-21 at 13 32 53

PR #39540, opened 2023-02-10
[Impeller] Use minimal coverage for EntityPass clip restores

bdero:
This makes clip restores a bit cheaper by only covering the area that's actually impacted by the restore as opposed to the whole pass texture. I don't think this is making much of a difference in our usual testing apps, but it's also not hard to construct somewhat pathological cases where the current behavior is surprisingly slow (not dissimilar to https://github.com/flutter/engine/pull/39358).


jonahwilliams (approved):
LGTM

PR #39543, opened 2023-02-10
[Impeller] Start stroke tessellation in compute

dnfield:
See comment below for update on what this does.

Fixes https://github.com/flutter/flutter/issues/120241


impeller/renderer/backend/metal/compute_pass_mtl.mm
```diff
@@ -241,23 +241,23 @@ static bool Bind(ComputePassBindingsCache& pass,
...
+    auto width = grid_size.width;
+    auto height = grid_size.height;
+    while (width * height >
+           static_cast<int64_t>(
+               pass_bindings.GetPipeline().maxTotalThreadsPerThreadgroup)) {
+      width /= 2;
+      height /= 2;
+    }
+    auto size = MTLSizeMake(width, height, 1);
+    [encoder dispatchThreadgroups:size threadsPerThreadgroup:size];
```
dnfield, line 259: This was a bug. Previously, only the last command in the pass was being dispatched.

dnfield:
Ok. This patch still only does things in tests, but it verifies the following:

- A compute shader that can generate quadratic bezier curves from cubics.
- A compute shader that can generate a polyline from quadratic bezier curve definitions
- A compute shader that can generate stroke vertices for a single contour stroke without any special caps or joins.

And a test that strings them all together.

This feels like enough that I'd like to get it reviewed and tested on CI. Some next steps:

- Improve the estimation of the sizes of the buffers
- Actually feed that vertex buffer into a vertex shader to draw the stroke
- Extend the stroke shader to support multiple contours/caps/joins
- Prove out using the interface for cubics/quads from our actual path definitions, especailly batching up all the paths for a scene to be used in various passes.

impeller/compiler/compiler.cc
```diff
@@ -34,7 +34,7 @@ static CompilerBackend CreateMSLCompiler(const spirv_cross::ParsedIR& ir,
   // If this version specification changes, the GN rules that process the
   // Metal to AIR must be updated as well.
   sl_options.msl_version =
-      spirv_cross::CompilerMSL::Options::make_msl_version(1, 2);
+      spirv_cross::CompilerMSL::Options::make_msl_version(2, 1);
```
jonahwilliams, line 37: Need to make these configurable and add a new target for it.

dnfield: Ahh right I have to fix this.

impeller/compiler/compiler.cc
```diff
@@ -357,9 +379,9 @@ Compiler::Compiler(const fml::Mapping& source_mapping,
           shaderc_optimization_level::shaderc_optimization_level_performance);
       spirv_options.SetTargetEnvironment(
           shaderc_target_env::shaderc_target_env_vulkan,
-          shaderc_env_version::shaderc_env_version_vulkan_1_0);
+          shaderc_env_version::shaderc_env_version_vulkan_1_1);
       spirv_options.SetTargetSpirv(
-          shaderc_spirv_version::shaderc_spirv_version_1_0);
+          shaderc_spirv_version::shaderc_spirv_version_1_3);
```
dnfield, line 384: @chinmaygarde or @iskakaushik - is this ok?

I'm not sure why we'd be targetting Vulkan 1.0 here but Vulkan 1.1 above.

I'm also not sure I understand the implications of using SPIR-V version 1.0 here. I need 1.3 for subgroups. I think it's probably ok but don't know where to look it up.

dnfield: If it's not ok all the time, I can parameterize this the way I did with the MSL version above.

impeller/compiler/compiler.cc
```diff
@@ -437,7 +459,11 @@ Compiler::Compiler(const fml::Mapping& source_mapping,
                    << ShaderCErrorToString(spv_result_->GetCompilationStatus())
                    << ". " << spv_result_->GetNumErrors() << " error(s) and "
                    << spv_result_->GetNumWarnings() << " warning(s).";
-    if (spv_result_->GetNumErrors() > 0 || spv_result_->GetNumWarnings() > 0) {
+    // It should normally be enough to check that there are errors or warnings,
+    // but some cases result in no errors or warnings and still have an error
+    // message. If there's a message we should print it.
+    if (spv_result_->GetNumErrors() > 0 || spv_result_->GetNumWarnings() > 0 ||
+        !spv_result_->GetErrorMessage().empty()) {
```
dnfield, line 466: This is the part that fixes https://github.com/flutter/flutter/issues/120241

impeller/compiler/compiler.cc
dnfield: Ok. This is now parameterized here and elsewhere.

Because it's now parameterized, the comment above no longer applies - the MSL version is set once and flows through the GN templates to here and the build_metal_library.py script.

jonahwilliams (approved):
LGTM

dnfield:
I think I finally got Windows to build.

PR #39553, opened 2023-02-11
[dart:ui] Introduce `PlatformDispatcher.implicitView`

loic-sharma:
This introduces `PlatformDispatcher.implicitView` (see https://github.com/flutter/flutter/issues/120306).

This is a low-level primitive for the framework's bootstrapping. Most code, including the framework after bootstrapping, will use `View.of(context)` instead of this new API. This new primitive will let us deprecate the `window` global.

Goals:

1. **Enable multi-window**. The `PlatformDispatcher.implicitView` is nullable. If `null`, the app must create a window to get a view it can draw into.
2. **Backwards compatibility**. For "single window" apps, `PlatformDispatcher.instance.implicitView` should behave as similar to `window` as possible.
    1. The `PlatformDispatcher.instance.implicitView.viewId` should be `0`.
    1. The `PlatformDispatcher.instance.implicitView` must be available synchronously at root isolate startup. This allows the framework to determine if it can make single window assumptions at startup.
    2. The `PlatformDispatcher.instance.implicitView` reference must not change after startup: if it is null at startup, it must always be null; if it is non-null at startup, it must always be non-null. If "single window" app enters headless mode, the implicit view must remain non-null.

In the future, the embedder will control whether an implicit view is created: mobile & legacy desktop apps will have an implicit view, multi-window desktop apps won't have an implicit view. This requires updating the engine's embedder API and is out-of-scope for this change. For now, all apps will have an implicit view.

Part of https://github.com/flutter/flutter/issues/120306

## Manual test

<details>
<summary>Manual test plan...</summary>

Modify: https://github.com/flutter/flutter/blob/master/examples/layers/raw/hello_world.dart

```diff
diff --git a/examples/layers/raw/hello_world.dart b/examples/layers/raw/hello_world.dart
index 20a53b0c7c..0940ebc49f 100644
--- a/examples/layers/raw/hello_world.dart
+++ b/examples/layers/raw/hello_world.dart
@@ -8,8 +8,9 @@
 import 'dart:ui' as ui;

 void beginFrame(Duration timeStamp) {
-  final double devicePixelRatio = ui.window.devicePixelRatio;
-  final ui.Size logicalSize = ui.window.physicalSize / devicePixelRatio;
+  ui.FlutterView? window = ui.PlatformDispatcher.instance.implicitView;
+  final double devicePixelRatio = window!.devicePixelRatio;
+  final ui.Size logicalSize = window!.physicalSize / devicePixelRatio;

   final ui.ParagraphBuilder paragraphBuilder = ui.ParagraphBuilder(
     ui.ParagraphStyle(textDirection: ui.TextDirection.ltr),
@@ -35,7 +36,7 @@ void beginFrame(Duration timeStamp) {
     ..addPicture(ui.Offset.zero, picture)
     ..pop();

-  ui.window.render(sceneBuilder.build());
+  window!.render(sceneBuilder.build());
 }

 // This function is the primary entry point to your application. The engine
```

✅ Verify that Windows works as expected:

```
flutter run .\raw\hello_world.dart -d windows --local-engine host_debug_unopt
```

✅ Verify that Web works as expected:

```
flutter run .\raw\hello_world.dart --local-engine host_debug_unopt -d chrome
```

</details>

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/lib/ui/platform_dispatcher.dart
+++ b/lib/ui/platform_dispatcher.dart
@@ -168,6 +168,36 @@ class PlatformDispatcher {
   // A map of opaque platform view identifiers to view configurations.
   final Map<Object, ViewConfiguration> _viewConfigurations = <Object, ViewConfiguration>{};
 
+  /// The [FlutterView] provided by the engine if the platform is unable to
+  /// create windows, or, for backwards compatibility.
+  ///
+  /// If the platform provides an implicit view, it can be used to bootstrap
+  /// the framework. This is common for platforms designed for single-view
+  /// applications like mobile devices with a single display.
+  ///
+  /// Applications and libraries must not rely on this property being set
+  /// as it may be null depending on the engine's configuration. Instead,
+  /// consider using [View.of] to lookup the [FlutterView] the current
+  /// [BuildContext] is drawing into.
+  ///
+  /// While the properties on the referenced [FlutterView] may change,
+  /// the reference itself is guaranteed to never change over the lifetime
+  /// of the application: if this property is null at startup, it will remain
+  /// so throughout the entire lifetime of the application. If it points to a
+  /// specific [FlutterView], it will continue to point to the same view until
+  /// the application is shut down (although the engine may replace or remove
+  /// the underlying backing surface of the view at its discretion).
+  ///
+  /// See also:
+  ///
+  /// * [View.of], for accessing the current view.
+  /// * [PlatformDisptacher.views] for a list of all [FlutterView]s provided
+  ///   by the platform.
+  FlutterView? get implicitView => _implicitViewEnabled() ? _views[0] : null;
+
+  @Native<Handle Function()>(symbol: 'PlatformConfigurationNativeApi::ImplicitViewEnabled')
+  external static bool _implicitViewEnabled();
+
   /// A callback that is invoked whenever the [ViewConfiguration] of any of the
   /// [views] changes.
   ///
--- a/lib/ui/window/platform_configuration.cc
+++ b/lib/ui/window/platform_configuration.cc
@@ -22,6 +22,8 @@
 namespace flutter {
 namespace {
 
+constexpr int kImplicitViewId = 0;
+
 Dart_Handle ToByteData(const fml::Mapping& buffer) {
   return tonic::DartByteData::Create(buffer.GetMapping(), buffer.GetSize());
 }
@@ -67,8 +69,13 @@ void PlatformConfiguration::DidCreateIsolate() {
                   Dart_GetField(library, tonic::ToDart("_drawFrame")));
   report_timings_.Set(tonic::DartState::Current(),
                       Dart_GetField(library, tonic::ToDart("_reportTimings")));
-  windows_.insert(std::make_pair(
-      0, std::make_unique<Window>(0, ViewportMetrics{1.0, 0.0, 0.0, -1})));
+
+  // TODO(loicsharma): This should only be created if the embedder enables the
+  // implicit view.
+  // See: https://github.com/flutter/flutter/issues/120306
+  windows_.emplace(kImplicitViewId,
+                   std::make_unique<Window>(
+                       kImplicitViewId, ViewportMetrics{1.0, 0.0, 0.0, -1}));
 }
 
 void PlatformConfiguration::UpdateLocales(
@@ -425,6 +432,16 @@ Dart_Handle PlatformConfigurationNativeApi::ComputePlatformResolvedLocale(
            ->ComputePlatformResolvedLocale(supportedLocales);
 
   return tonic::DartConverter<std::vector<std::string>>::ToDart(results);
+}
+
+Dart_Handle PlatformConfigurationNativeApi::ImplicitViewEnabled() {
+  UIDartState::ThrowIfUIOperationsProhibited();
+  bool enabled = UIDartState::Current()
+                     ->platform_configuration()
+                     ->client()
+                     ->ImplicitViewEnabled();
+
+  return Dart_NewBoolean(enabled);
 }
 
 std::string PlatformConfigurationNativeApi::DefaultRouteName() {
--- a/lib/web_ui/lib/src/engine/platform_dispatcher.dart
+++ b/lib/web_ui/lib/src/engine/platform_dispatcher.dart
@@ -148,6 +148,34 @@ class EnginePlatformDispatcher extends ui.PlatformDispatcher {
   final Map<Object, ui.ViewConfiguration> _windowConfigurations =
       <Object, ui.ViewConfiguration>{};
 
+  /// The [FlutterView] provided by the engine if the platform is unable to
+  /// create windows, or, for backwards compatibility.
+  ///
+  /// If the platform provides an implicit view, it can be used to bootstrap
+  /// the framework. This is common for platforms designed for single-view
+  /// applications like mobile devices with a single display.
+  ///
+  /// Applications and libraries must not rely on this property being set
+  /// as it may be null depending on the engine's configuration. Instead,
+  /// consider using [View.of] to lookup the [FlutterView] the current
+  /// [BuildContext] is drawing into.
+  ///
+  /// While the properties on the referenced [FlutterView] may change,
+  /// the reference itself is guaranteed to never change over the lifetime
+  /// of the application: if this property is null at startup, it will remain
+  /// so throughout the entire lifetime of the application. If it points to a
+  /// specific [FlutterView], it will continue to point to the same view until
+  /// the application is shut down (although the engine may replace or remove
+  /// the underlying backing surface of the view at its discretion).
+  ///
+  /// See also:
+  ///
+  /// * [View.of], for accessing the current view.
+  /// * [PlatformDisptacher.views] for a list of all [FlutterView]s provided
+  ///   by the platform.
+  @override
+  ui.FlutterView? get implicitView => viewData[kImplicitViewId];
+
   /// A callback that is invoked whenever the platform's [devicePixelRatio],
   /// [physicalSize], [padding], [viewInsets], or [systemGestureInsets]
   /// values change, for example when the device is rotated or when the
@@ -479,7 +507,7 @@ class EnginePlatformDispatcher extends ui.PlatformDispatcher {
             // TODO(a-wallen): As multi-window support expands, the pop call
             // will need to include the view ID. Right now only one view is
             // supported.
-            (viewData[kSingletonViewId]! as EngineFlutterWindow)
+            (viewData[kImplicitViewId]! as EngineFlutterWindow)
                 .browserHistory
                 .exit()
                 .then((_) {
@@ -584,7 +612,7 @@ class EnginePlatformDispatcher extends ui.PlatformDispatcher {
         // TODO(a-wallen): As multi-window support expands, the navigation call
         // will need to include the view ID. Right now only one view is
         // supported.
-        (viewData[kSingletonViewId]! as EngineFlutterWindow)
+        (viewData[kImplicitViewId]! as EngineFlutterWindow)
             .handleNavigationMessage(data)
             .then((bool handled) {
           if (handled) {
@@ -1173,7 +1201,7 @@ class EnginePlatformDispatcher extends ui.PlatformDispatcher {
   @override
   String get defaultRouteName {
     return _defaultRouteName ??=
-        (viewData[kSingletonViewId]! as EngineFlutterWindow).browserHistory.currentPath;
+        (viewData[kImplicitViewId]! as EngineFlutterWindow).browserHistory.currentPath;
   }
 
   /// Lazily initialized when the `defaultRouteName` getter is invoked.
--- a/lib/web_ui/test/window_test.dart
+++ b/lib/web_ui/test/window_test.dart
@@ -41,6 +41,7 @@ void testMain() {
     await window.resetHistory();
   });
 
+  // For now, web always has an implicit view provided by the web engine.
   test('EnginePlatformDispatcher.instance.implicitView should be non-null', () async {
     expect(EnginePlatformDispatcher.instance.implicitView, isNotNull);
     expect(EnginePlatformDispatcher.instance.implicitView?.viewId, 0);
--- a/lib/web_ui/test/window_test.dart
+++ b/lib/web_ui/test/window_test.dart
@@ -41,6 +41,12 @@ void testMain() {
     await window.resetHistory();
   });
 
+  test('EnginePlatformDispatcher.instance.implicitView should be non-null', () async {
+    expect(EnginePlatformDispatcher.instance.implicitView, isNotNull);
+    expect(EnginePlatformDispatcher.instance.implicitView?.viewId, 0);
+    expect(window.viewId, 0);
+  });
+
   test('window.defaultRouteName should work with JsUrlStrategy', () async {
     dynamic state = <dynamic, dynamic>{};
     final JsUrlStrategy jsUrlStrategy = JsUrlStrategy(
--- a/shell/common/engine.cc
+++ b/shell/common/engine.cc
@@ -431,6 +431,14 @@ void Engine::SetAccessibilityFeatures(int32_t flags) {
   runtime_controller_->SetAccessibilityFeatures(flags);
 }
 
+bool Engine::ImplicitViewEnabled() {
+  // TODO(loicsharma): This value should be provided by the embedder
+  // when it launches the engine. For now, assume the embedder always creates a
+  // view.
+  // See: https://github.com/flutter/flutter/issues/120306
+  return true;
+}
+
 std::string Engine::DefaultRouteName() {
   if (!initial_route_.empty()) {
     return initial_route_;
--- a/shell/platform/embedder/fixtures/main.dart
+++ b/shell/platform/embedder/fixtures/main.dart
@@ -45,8 +45,15 @@ void executableNameNotNull() {
   notifyStringValue(Platform.executable);
 }
 
+@pragma('vm:entry-point')
+void implicitViewNotNull() {
+  notifyBoolValue(PlatformDispatcher.instance.implicitView != null);
+}
+
 @pragma('vm:external-name', 'NotifyStringValue')
 external void notifyStringValue(String value);
+@pragma('vm:external-name', 'NotifyBoolValue')
+external void notifyBoolValue(bool value);
 
 @pragma('vm:entry-point')
 void invokePlatformTaskRunner() {
--- a/shell/platform/embedder/tests/embedder_unittests.cc
+++ b/shell/platform/embedder/tests/embedder_unittests.cc
@@ -174,6 +174,30 @@ TEST_F(EmbedderTest, ExecutableNameNotNull) {
   latch.Wait();
 }
 
+TEST_F(EmbedderTest, ImplicitViewNotNull) {
+  // TODO(loicsharma): Update this test when embedders can opt-out
+  // of the implicit view.
+  // See: https://github.com/flutter/flutter/issues/120306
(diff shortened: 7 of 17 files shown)


lib/ui/window/platform_configuration.cc
```diff
@@ -427,6 +427,17 @@ Dart_Handle PlatformConfigurationNativeApi::ComputePlatformResolvedLocale(
...
 
+Dart_Handle PlatformConfigurationNativeApi::ImplicitViewId() {
+  UIDartState::ThrowIfUIOperationsProhibited();
+
+  auto enabled = UIDartState::Current()
+                     ->platform_configuration()
+                     ->client()
+                     ->ImplicitViewEnabled();
+
+  return enabled ? Dart_NewInteger(0) : Dart_Null();
```
loic-sharma, line 438: TODO: Make `0` an "implicit view ID" constant, share with above.

lib/web_ui/lib/src/engine/platform_dispatcher.dart
```diff
@@ -148,6 +148,10 @@ class EnginePlatformDispatcher extends ui.PlatformDispatcher {
   final Map<Object, ui.ViewConfiguration> _windowConfigurations =
       <Object, ui.ViewConfiguration>{};
 
+  // TODO: Add comment + TODO
+  @override
+  ui.FlutterView? get implicitView => viewData[kSingletonViewId];
```
loic-sharma, line 153: Change to `kImplicitViewId`

ditman:
web_ui bits look good! Thanks!

lib/ui/platform_dispatcher.dart
```diff
@@ -168,6 +168,23 @@ class PlatformDispatcher {
...
 
+  /// The implicit [FlutterView] provided by the platform, if any.
+  ///
+  /// This is used to bootstrap the framework for applications where the
+  /// platform provides a view, such as applications designed for
+  /// single-display mobile devices.
+  ///
+  /// While the [FlutterView]'s properties may change, this reference is
+  /// guaranteed to never change.
+  ///
```
goderbauer, line 179: Should we add some kind of a warning here that applications and packages should not rely on this property being set and instead acquire the view via other means (e.g. `View.of(context)` is appropriate for most use cases)? This property is basically just interesting for low-level app bootstrapping.

goderbauer: (I guess you're sort of hinting at this with the "see also" section)

lib/ui/platform_dispatcher.dart
```diff
@@ -168,6 +168,23 @@ class PlatformDispatcher {
...
+  /// See also:
+  ///
+  /// * [View.of], for accessing the current view.
+  FlutterView? get implicitView => _implicitViewEnabled() ? _views[0] : null;
+
+  @Native<Handle Function()>(symbol: 'PlatformConfigurationNativeApi::ImplicitViewEnabled')
+  external static bool _implicitViewEnabled();
```
goderbauer, line 186: Wondering if this maybe should just return the implicit view ID (or null) to avoid that we have to hard-code the magic number here and in platform_configuration.cc.

lib/ui/platform_dispatcher.dart
```diff
@@ -168,6 +168,23 @@ class PlatformDispatcher {
...
   final Map<Object, ViewConfiguration> _viewConfigurations = <Object, ViewConfiguration>{};
 
+  /// The implicit [FlutterView] provided by the platform, if any.
+  ///
+  /// This is used to bootstrap the framework for applications where the
+  /// platform provides a view, such as applications designed for
+  /// single-display mobile devices.
+  ///
+  /// While the [FlutterView]'s properties may change, this reference is
+  /// guaranteed to never change.
```
goderbauer, line 178: nit: maybe add: "... for the lifetime of the application."

shell/common/engine.cc
```diff
@@ -431,6 +431,13 @@ void Engine::SetAccessibilityFeatures(int32_t flags) {
   runtime_controller_->SetAccessibilityFeatures(flags);
 }
 
+bool Engine::ImplicitViewEnabled() {
+  // TODO(loicsharma): This value should be provided by the embedder
+  // when it launches the engine. For now, assume the embedder always creates a
+  // view.
```
goderbauer, line 437: Maybe link this to the issue?

goderbauer (approved):
LGTM

lib/ui/platform_dispatcher.dart
loic-sharma: I had this approach originally but I switched to a boolean instead as we don't have any scenarios where the implicit view will have a non-zero ID. I don't feel strongly either ways though, so let me know if you have a preference.

shell/common/engine.cc
loic-sharma: Whoops, I meant to do that. Will fix!

lib/ui/platform_dispatcher.dart
goderbauer: No preference, really. It's all private and internal anyways, so easy to change if it has to.

lib/ui/platform_dispatcher.dart
loic-sharma: Could you take another look? I updated this comment to push people towards `View.of` instead. The new comment is less clear as to when an implicit view is provided by the platform though...

lib/ui/platform_dispatcher.dart
loic-sharma: Added

shell/common/engine.cc
loic-sharma: Added

lib/ui/platform_dispatcher.dart
loic-sharma: Sounds good. I'll leave this open so others can chime in :)

yaakovschectman (approved):
LGTM with one question

lib/ui/platform_dispatcher.dart
goderbauer: I took a stab below at how I would write the documentation for this property. Feel free to copy it as a whole or bits and pieces that you like or ignore it completely.

---

The implicit [FlutterView] provided by the platform, if any.

If the platform provides an implicit view, it can be used to bootstrap the framework. This is common for platforms designed for single-view applications (e.g. on mobile devices with a single display). 

Applications and libraries must not rely on this property being set as it may be null depending on the engine's configuration. Instead, consider using [View.of] to lookup the [FlutterView] the current [BuildContext] is drawing into.

While the properties on the referenced [FlutterView] may change, the reference itself is guaranteed to never change over the lifetime of the application: If this property is null at startup, it will remain so throughout the entire lifetime of the application. If it points to a specific [FlutterView], it will continue to point to the same view until the application is shut down (although the engine may replace or remove the underlying backing surface of the view at its discretion).

See also:

 * [PlatformDisptacher.views] for a list of all [FlutterView]s provided by the platform.

lib/ui/platform_dispatcher.dart
```diff
@@ -168,6 +168,23 @@ class PlatformDispatcher {
...
+  ///
+  /// See also:
+  ///
+  /// * [View.of], for accessing the current view.
+  FlutterView? get implicitView => _implicitViewEnabled() ? _views[0] : null;
```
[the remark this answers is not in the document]
loic-sharma: Yup, if there is an implicit view it is guaranteed to have view ID `0`. Note that this is a map and not a list. Furthermore, the implicit view & underlying view exist even if the app transitions to headless mode - this matches today's `window` behavior.

"Modern" desktop apps won't have an implicit view. These are apps that use the new runner templates that doesn't exist yet. For these future apps, view ID `0` won't have any special meaning.

lib/ui/platform_dispatcher.dart
loic-sharma: That's perfect, will update!

loic-sharma: Updated! FYI, I added `[View.of]` to the "See also" section. It's a little repetitive, but I want to nudge folks towards `View.of` as much as possible :)

lib/ui/platform_dispatcher.dart
```diff
@@ -168,6 +168,35 @@ class PlatformDispatcher {
   // A map of opaque platform view identifiers to view configurations.
   final Map<Object, ViewConfiguration> _viewConfigurations = <Object, ViewConfiguration>{};
 
+  /// The implicit [FlutterView] provided by the platform, if any.
```
cbracken, line 171: This feels a bit tautological.

I suspect that the "implicit view" is the view that, should the user not manually configure any other views, is the initial view automatically provided by Flutter itself, but @goderbauer and I were discussing this problem a few weeks back so I've got a bit of insider knowledge that an API doc reader may not have.

The name `implicitView` is probably fine, but we should define it clearly in the first line of the doc comment, which is all that's shown in some cases. 

Maybe something like:
```
The [FlutterView] in which content is rendered in the absence of any user-configured views.
```

lib/ui/platform_dispatcher.dart
```diff
@@ -168,6 +168,35 @@ class PlatformDispatcher {
...
+  ///
+  /// If the platform provides an implicit view, it can be used to bootstrap
+  /// the framework. This is common for platforms designed for single-view
+  /// applications (e.g. on mobile devices with a single display).
+  ///
+  /// Applications and libraries must not rely on this property being set
+  /// as it may be null depending on the engine's configuration. Instead,
+  /// consider using [View.of] to lookup the [FlutterView] the current
+  /// [BuildContext] is drawing into.
```
cbracken, line 180: Overall this is good. Rather than pointing out that it may be null (clear from the method signature already), I'd explain _when_ this will be null or non-null, and follow up with your existing suggestion to avoid this in favour of `View.of`.

lib/ui/platform_dispatcher.dart
cbracken: Added just a couple more suggestions above but this is looking pretty great!

lib/ui/window/platform_configuration.cc
```diff
@@ -22,6 +22,8 @@
 namespace flutter {
 namespace {
 
+const int kImplicitViewId = 0;
```
cbracken, line 25:
```suggestion
constexpr int kImplicitViewId = 0;
```

lib/ui/window/platform_configuration.cc
```diff
@@ -67,8 +69,14 @@ void PlatformConfiguration::DidCreateIsolate() {
...
-  windows_.insert(std::make_pair(
-      0, std::make_unique<Window>(0, ViewportMetrics{1.0, 0.0, 0.0, -1})));
+
+  // TODO(loicsharma): This should only be created if the embedder enables the
+  // implicit view.
+  // See: https://github.com/flutter/flutter/issues/120306
+  windows_.insert(
+      std::make_pair(kImplicitViewId,
+                     std::make_unique<Window>(
+                         kImplicitViewId, ViewportMetrics{1.0, 0.0, 0.0, -1})));
```
cbracken, line 79: You should be able to use map::emplace() here, I'd think.

lib/web_ui/lib/src/engine/platform_dispatcher.dart
```diff
@@ -148,6 +148,33 @@ class EnginePlatformDispatcher extends ui.PlatformDispatcher {
...
+  /// so throughout the entire lifetime of the application. If it points to a
+  /// specific [FlutterView], it will continue to point to the same view until
+  /// the application is shut down (although the engine may replace or remove
+  /// the underlying backing surface of the view at its discretion).
+  ///
+  /// See also:
+  ///
+  /// * [View.of], for accessing the current view.
+  /// * [PlatformDisptacher.views] for a list of all [FlutterView]s provided
+  ///   by the platform.
```
cbracken, line 174: Just a note to remember to update here to match the lib/ui version.

shell/platform/embedder/tests/embedder_unittests.cc
```diff
@@ -174,6 +174,25 @@ TEST_F(EmbedderTest, ExecutableNameNotNull) {
...
 
+TEST_F(EmbedderTest, ImplicitViewNotNull) {
+  auto& context = GetEmbedderContext(EmbedderTestContextType::kSoftwareContext);
+
+  fml::AutoResetWaitableEvent latch;
+  context.AddNativeCallback(
+      "NotifyBoolValue", CREATE_NATIVE_ENTRY([&](Dart_NativeArguments args) {
+        const auto dart_bool = tonic::DartConverter<bool>::FromDart(
+            Dart_GetNativeArgument(args, 0));
+        EXPECT_TRUE(dart_bool);
```
cbracken, line 185: This is fine though I'd be tempted to move the bool out, capture by reference then keep the checks at the top-level of the test. Captures are possible since you've used the `CREATE_NATIVE_ENTRY` macro.

cbracken (approved):
Overall looks good - just a few C++ and doc nits.

lib/ui/platform_dispatcher.dart
loic-sharma: > I'd explain _when_ this will be null or non-null

It's hard to explain when this is null as it is rather arbitrary from the perspective of the engine/framework.

The current plan is:

1. Mobile apps will be non-null
2. Desktop apps using today's runner will be non-null, even if they have multiple windows. This ensures backwards compatibility
3. Desktop apps using a future runner that doesn't exist yet will be null

However, a brand new desktop embedder could choose to never provide an implicit view. In other words, whether there is an implicit view depends on your embedder & your runner app. 

Let me think on this some more...

lib/ui/platform_dispatcher.dart
loic-sharma: > I suspect that the "implicit view" is the view that, should the user not manually configure any other views, is the initial view automatically provided by Flutter itself

That's not quite right. It's a view provided for backwards compatibility reasons, or, by embedders that cannot create windows. See: https://github.com/flutter/engine/pull/39553#discussion_r1107565652

lib/ui/platform_dispatcher.dart
```diff
@@ -168,6 +168,35 @@ class PlatformDispatcher {
   // A map of opaque platform view identifiers to view configurations.
   final Map<Object, ViewConfiguration> _viewConfigurations = <Object, ViewConfiguration>{};
 
+  /// The implicit [FlutterView] provided by the platform, if any.
+  ///
+  /// If the platform provides an implicit view, it can be used to bootstrap
+  /// the framework. This is common for platforms designed for single-view
+  /// applications (e.g. on mobile devices with a single display).
```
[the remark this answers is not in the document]
loic-sharma: Today I learned! After some reflection, I switched to `like` instead as new English speakers may not be familiar with `e.g.`

lib/ui/platform_dispatcher.dart
loic-sharma: I tried to clarify this, see: https://github.com/flutter/engine/pull/39553#discussion_r1107582717

lib/web_ui/lib/src/engine/platform_dispatcher.dart
goderbauer: Actually, I think you can just delete this copy of the doc? It should just get inherited from the super method.

goderbauer (approved):
LGTM 🚀

lib/web_ui/lib/src/engine/platform_dispatcher.dart
loic-sharma: @ditman Do you know why web's engine `PlatformDispatcher` has the doc comments? Is it safe to remove this comment?

goderbauer: > Just in case it's not clear, this overrides the web_ui's PlatformDispatcher, not ui's `PlatformDispatcher.

I did indeed miss that.

Still, I don't think the doc comments from web_ui are used anywhere (well, except maybe for people reading the source directly). I defer to the web team to decide whether these comments should actually be duplicated here.

lib/ui/window/platform_configuration.cc
loic-sharma: Good idea, that's much better!

shell/platform/embedder/tests/embedder_unittests.cc
loic-sharma: Fixed!

shell/platform/embedder/fixtures/main.dart
```diff
@@ -45,8 +45,15 @@ void executableNameNotNull() {
   notifyStringValue(Platform.executable);
 }
 
+@pragma('vm:entry-point')
+void implicitViewNotNull() {
+  notifyBoolValue(PlatformDispatcher.instance.implicitView != null);
```
dkwingsmt, line 50: What does this test do? To ensure that `PlatformDispatcher.instance.implicitView` isn't null? But this is only because we don't have an option to control it right? Should we add a TODO or at least comment to explain that this test might (should) be broken once we allow non-implicitView?

lib/web_ui/test/window_test.dart
```diff
@@ -41,6 +41,12 @@ void testMain() {
     await window.resetHistory();
   });
 
+  test('EnginePlatformDispatcher.instance.implicitView should be non-null', () async {
```
dkwingsmt, line 44: Same.

dkwingsmt (approved):
LGTM, nice work!

shell/platform/embedder/fixtures/main.dart
loic-sharma: This is just a fixture to check whether the implicit view is non-null. The real test is `EmbedderTest.ImplicitViewNotNull`.

> But this is only because we don't have an option to control it right?

Yup that's correct.

> Should we add a TODO or at least comment to explain that this test might (should) be broken once we allow non-implicitView?

Good idea, added a TODO to `EmbedderTest.ImplicitViewNotNull`.

lib/web_ui/lib/src/engine/platform_dispatcher.dart
ditman: > Do you know why web's engine PlatformDispatcher has the doc comments? Is it safe to remove this comment?

Not really, I remember some pushes in the past to remove documentation from web_ui that had drifted with regards to the proper dart:ui version (which is the one that is published in the docsite, anyways), and letting the IDE pick it up as suggested by @goderbauer.

(Unfortunately I cannot find a specific commit doing what I describe, so maybe it was done in small increments (sorry!))

If the IDE picks up the documentation from dart:ui, I'm fine with it being removed from the web_ui source.

lib/web_ui/test/window_test.dart
dkwingsmt: > All other platforms - mobile and web - will have an implicit view instead. In other words, this test will be correct for the foreseeable future.

I think we'd rather not assume that (and we don't have to, right?). The user might very possibly attach multiple views from the native code, and the framework will handle them. And with that, the user might even choose not to use the implicit view.

ditman: > All other platforms - mobile and web - will have an implicit view instead.

Web will (possibly) want to render to multiple different target DIVs at some point, so I guess we're going to have multiple "views" too?

loic-sharma: > 



> > All other platforms - mobile and web - will have an implicit view instead.
> 
> Web will (possibly) want to render to multiple different target DIVs at some point, so I guess we're going to have multiple "views" too?

To clarify there's two semi-related things here:

1. Multi-view: all platforms, including web, will have multi-view support
2. Implicit view: this replaces the `window` global

Having an implicit view does not block multi-view. It's just the replacement for the `window` global.

ditman: Thanks for the clarification @loic-sharma (expect a quick call from me soon-ish though, I want to start redoing stuff in the web engine so the multi-view lands more easily!)

PR #39560, opened 2023-02-11
[Impeller] Return entities from filters instead of snapshots

bdero:
This is a pass collapse optimization I've been meaning to get around to.

By outputting a deferred rendering command, filters are given full control over how its results should be applied to the parent pass; whether a subpass is necessary is entirely up to the filter.

This PR is a pure refactor that shouldn't change anything about how we render. Note that thanks to our other texture passthrough optimizations, this does _not_ introduce an extra pass when filters are chained. I plan to land the actual pass removals in follow-ups.


jonahwilliams (approved):
LGTM

PR #31098, opened 2022-01-27
Add Animated PNG demuxer

bdero:
Adds a multiframe image decoder that demuxes APNG streams into regular PNG-compliant streams. The actual color decoding is done by Skia's existing PNG decoder.

A single lazy pass is performed over the full stream as the frames are decoded for the first time.

Resolves https://github.com/flutter/flutter/issues/37247.

PR #39579, opened 2023-02-13
Revert "Add Animated PNG demuxer"

itsjustkevin:
Reverts flutter/engine#31098 as it is causing breakages.


bdero:
Should be fixed with cl/509118678.

PR #39592, opened 2023-02-13
[web] Glue code for CanvasKit Chromium

mdebbar:
This PR provides all client-ICU data to CanvasKit from browser APIs:

* **Line breaks**: from `v8BreakIterator` (only available in Chromium).
* **Word boundaries**: from `Intl.Segmenter` (available in all browsers except Firefox).
* **Grapheme boundaries**: from `Intl.Segmenter` (available in all browsers except Firefox).

Fixes https://github.com/flutter/flutter/issues/118801


lib/web_ui/lib/src/engine/canvaskit/text_fragmenter.dart
```diff
@@ -0,0 +1,91 @@
...
+  word,
+}
+
+class CkIntlFragmenter extends CkTextFragmenter {
+  CkIntlFragmenter(super.text, this.granularity)
+      : assert(domIntl.Segmenter != null);
+
+  final IntlSegmenterGranularity granularity;
+
+  String get _granularityString {
```
yjbanov, line 35: This can become a `final String _granularityString` to avoid putting the switch block on every `fragment()`  invocation.

lib/web_ui/lib/src/engine/canvaskit/text_fragmenter.dart
```diff
@@ -0,0 +1,91 @@
...
+        return 'grapheme';
+      case IntlSegmenterGranularity.word:
+        return 'word';
+    }
+  }
+
+  @override
+  Uint32List fragment() {
+    final DomIteratorWrapper<DomSegment> iterator =
+        createIntlSegmenter(granularity: _granularityString)
```
yjbanov, line 47: Same question here. Is `IntlSegmenter` stateful? Can we memoize one for each granularity and avoid creating them on every string? This will become more relevant with wasm, where we probably want to avoid too many wasm-js hops.

lib/web_ui/lib/src/engine/canvaskit/text_fragmenter.dart
```diff
@@ -0,0 +1,91 @@
...
+  CkLineBreakFragmenter(super.text) : _fragmenter = LineBreakFragmenter(text);
+
+  final LineBreakFragmenter _fragmenter;
+
+  @override
+  Uint32List fragment() {
+    final List<LineBreakFragment> fragments = _fragmenter.fragment();
+
+    final int size = (fragments.length + 1) * 2;
+    final Uint32List typedArray = Uint32List(size);
```
yjbanov, line 75: Consider using `CanvasKit.Malloc` to avoid creating a temporary array and extra copying into wasm.

lib/web_ui/lib/src/engine/dom.dart
```diff
@@ -1790,17 +1793,123 @@ class _DomTouchListWrapper<T> extends Iterable<T> {
...
   ///
   /// See: https://code.google.com/archive/p/v8-i18n/wikis/BreakIterator.wiki
   external Object? get v8BreakIterator;
 }
 
+/// Similar to [js_util.callMethod] but allows for providing a non-string method
+/// name.
+T jsCallMethod<T>(Object o, Object method, [List<Object?> args = const <Object?>[]]) {
+  final Object actualMethod = js_util.getProperty(o, method);
+  return js_util.callMethod<T>(actualMethod, 'apply', <Object?>[o, args]);
```
yjbanov, line 1822: Let's check with @joshualitt if this is future-proof.

lib/web_ui/lib/src/engine/dom.dart
```diff
@@ -1790,17 +1793,123 @@ class _DomTouchListWrapper<T> extends Iterable<T> {
...
+
   /// This is a V8-only API for segmenting text.
   ///
   /// See: https://code.google.com/archive/p/v8-i18n/wikis/BreakIterator.wiki
   external Object? get v8BreakIterator;
 }
 
+/// Similar to [js_util.callMethod] but allows for providing a non-string method
+/// name.
+T jsCallMethod<T>(Object o, Object method, [List<Object?> args = const <Object?>[]]) {
```
yjbanov, line 1820: Can this be moved to `safe_browser_api.dart` and wrapped as to not be publicly available to all of the engine code?

yjbanov: If it's too difficult to move, can it stay private to this file?

yjbanov (commented):
I also recommend running an A/B test using some of our text benchmarks.

lib/web_ui/lib/src/engine/dom.dart
joshualitt: I'm committed to migrating the engine shim to whatever future interop looks like. That said, I'm hoping to have a replacement for js_util based on JS types, and when I do I can migrate this use case.

lib/web_ui/lib/src/engine/canvaskit/text.dart
```diff
@@ -976,6 +977,18 @@ class CkParagraphBuilder implements ui.ParagraphBuilder {
...
+    if (useClientICU) {
+      final String text = _paragraphBuilder.getText();
+      _paragraphBuilder.setWordsUtf16(
+        CkIntlFragmenter(text, IntlSegmenterGranularity.word).fragment(),
+      );
+      _paragraphBuilder.setGraphemeBreaksUtf16(
+        CkIntlFragmenter(text, IntlSegmenterGranularity.grapheme).fragment(),
+      );
+      _paragraphBuilder.setLineBreaksUtf16(
+        CkLineBreakFragmenter(text).fragment(),
```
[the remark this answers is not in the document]
mdebbar: Makes sense. I made all these methods stateless and reused the same `Intl.Segmenter` instance for each `granularity`. Good idea!

lib/web_ui/lib/src/engine/canvaskit/text_fragmenter.dart
mdebbar: Not applicable anymore.

lib/web_ui/lib/src/engine/canvaskit/text_fragmenter.dart
mdebbar: Done.

Regarding the wasm thing: I don't think this memoization makes a difference. The majority of chatting with js happens *after* creating the `Intl.Segmenter` (calling `segment()` and iterating over the result). If it becomes an issue, we can implement a JS function that wraps this entire block of code and does all the chatting on the JS side and returns the final result (i.e. `Uint32List`) to wasm. Not sure how this can be done in practice with dart2wasm.

cc @eyebrowsoffire to keep in mind that this might be needed for the skwasm project. It would be nice to have some kind of annotation on methods to tell the compiler that it's better to keep the entire method in JS and not compile it to wasm. I may write some thoughts around this in a new issue.

lib/web_ui/lib/src/engine/canvaskit/text_fragmenter.dart
mdebbar: Up until today, I didn't really understand the benefit of using CK's Malloc. It seemed to me like the CK bindings always end up copying the arrays anyway.

But I took a closer look now, and I think I understand where this optimization is happening on the CK side:
https://github.com/google/skia/blob/deb1cf8d127986fee464946a12b248d6f17fb629/modules/canvaskit/memory.js#L126-L128

What threw me off was the name `copy1dArray`. But it turns out this function doesn't always *copy*. It first checks if the array was malloc'd and avoids the copying in that case.

lib/web_ui/lib/src/engine/dom.dart
mdebbar: I understand the benefit of making it private, so I'll make it private.

Would you mind clarifying what's the benefit of moving it to `safe_browser_api.dart`?

mdebbar:
> I also recommend running an A/B test using some of our text benchmarks.

Agreed. @eyebrowsoffire is working on a PR to run tests against skwasm as well as the chromium variant of CanvasKit.

That said, all of this code is behind a flag that's not enabled yet, but we definitely need to land those tests before opening this up to users.

lib/web_ui/lib/src/engine/dom.dart
yjbanov: This function can execute arbitrary code. It's good to avoid ad hoc calls out to JS from all over the codebase. So libraries like `dom.dart` and `safe_browser_api.dart` should provide a safe subset of the API to the rest of the engine.

lib/web_ui/lib/src/engine/canvaskit/text.dart
```diff
@@ -976,6 +977,16 @@ class CkParagraphBuilder implements ui.ParagraphBuilder {
 
   /// Builds the CkParagraph with the builder and deletes the builder.
   SkParagraph _buildSkParagraph() {
+    if (browserSupportsCanvaskitChromium) {
+      final String text = _paragraphBuilder.getText();
```
yjbanov, line 981: Why do we need to read the text string back? Don't we already have it?

Now that I'm looking at the Skia API, it feels a little too chatty. I feel like we should be able to provide all this information - text, word breaks, graphemes, line breaks - in a single method call.

But maybe we can revisit this later. If this works, let's keep it. It's not public API anyway. Or maybe I just need some likbez on why this should be this way.

yjbanov (approved):
lgtm

hterkelsen (approved):
LGTM

lib/web_ui/lib/src/engine/canvaskit/text.dart
hterkelsen: For anyone curious: https://en.wikipedia.org/wiki/Likbez

joshualitt (approved):
JS Interop LGTM

lib/web_ui/lib/src/engine/dom.dart
mdebbar: I agree with making it private so it doesn't get called everywhere.

Does it make a difference if we put it in `dom.dart` vs `safe_browser_api.dart`? Like is there a special handling for that file or something?

lib/web_ui/lib/src/engine/canvaskit/text.dart
mdebbar: Thanks @hterkelsen for the link 😄 I was wondering about that word myself, and thought it could've been a typo.

The reason we need to ask Skia to give us the string is because SkParagraphBuilder inserts extra text into the string. E.g. when we call `addPlaceholder(...)`, SkParagraphBuilder inserts a magical character into the string (and this is an implementation detail that could change or even go away in the future).

A little bit of history to explain how the API became chatty:

Initially, we had a single method called `buildWithClientInfo()` (to be called instead of `build()`) which takes parameters for all the information it needs (i.e. line breaks, word breaks, graphemes, (and bidi at the time)).

Then we started playing with ICU4X for the bidi information. ICU4X operates on UTF8, but browser APIs operate on UTF16. So we needed a way to indicate to Skia whether the array we are passing is UTF8 or UTF16 indices. That's how the `set***Utf8()` and `set***Utf16()` APIs were born.

PR #39603, opened 2023-02-13
[Impeller] Fix multi-function compute

dnfield:
Adding multiple kernels to a single compute pass was only running the last one - the call to actually dispatch the method was nested improperly and only being invoked for the final command in the buffer.

Adds a test covering this situation, including sharing data between the kernel functions and verifying the output of both.


jonahwilliams (approved):
LGTM

PR #39607, opened 2023-02-14
Revert "Add support for double tap action from Apple Pencil 2"

jonahwilliams:
Reverts flutter/engine#39267

This change is causing the framework to fail its analysis checks because a switch over the PointerSignalKind is no longer exhaustive due to the addition of stylusAuxiliaryAction.

### DIFF
--- a/lib/ui/platform_dispatcher.dart
+++ b/lib/ui/platform_dispatcher.dart
@@ -371,7 +371,7 @@ class PlatformDispatcher {
   //  * pointer_data.cc
   //  * pointer.dart
   //  * AndroidTouchProcessor.java
-  static const int _kPointerDataFieldCount = 36;
+  static const int _kPointerDataFieldCount = 35;
 
   static PointerDataPacket _unpackPointerDataPacket(ByteData packet) {
     const int kStride = Int64List.bytesPerElement;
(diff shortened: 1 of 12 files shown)


jonahwilliams:
See https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8789227906726691521/+/u/Analyze/analyze/test_stdout

jonahwilliams:
This looks like the failure is only in one place, so if you have a patch ready we could quickly land that.

hellohuanlin (approved):

hellohuanlin:
Interesting. May have to look into git history and see how other fields are added.

jonahwilliams:
Generally the way you would do it is to:

1. add a `default` case to the switch statement with the behavior you expect the new enum to have.
2. Land the engine change that adds the new case
3. Once that engine change has rolled into the framework, remove the default case and add the case for the new variant.

hellohuanlin:
@jonahwilliams that makes sense!

hellohuanlin:
@jonahwilliams im wondering how do we avoid this break in the future? maybe having some basic framework test in the engine CI?

jmagman:
> @jonahwilliams im wondering how do we avoid this break in the future? maybe having some basic framework test in the engine CI?

It was caught during a engine -> framework roll, which is the earliest these kinds of failures get caught at the moment. 
 https://github.com/flutter/flutter/pull/120646

There are some web tests that run framework tests, they clone the newest version of the framework that's older than the engine version, it's very complicated and somewhat error prone go/flutter-web-engine-break https://github.com/flutter/engine/blob/master/tools/clone_flutter.sh

hellohuanlin:
Gotcha. it sounds like mono-repo maybe something we want in the future where engine and framework share the same repo.

PR #39629, opened 2023-02-14
Fix build output path of the shader analysis core list

zanderso:
`$root_gen_dir` isn't guaranteed to exist when GN runs this, and the script won't create it before trying to write to the file. So, this PR changes the output path to one that is guaranteed to exist.


jonahwilliams (approved):
RSLGTM

PR #39636, opened 2023-02-14
[Impeller] Make RenderTarget::CreateOffscreen utilities add a stencil by default

bdero:
This fixes the problem noted in this comment: https://github.com/flutter/flutter/issues/120586#issuecomment-1428023566

A breakage was introduced in https://github.com/flutter/engine/pull/39537 that affects `Picture::ToImage` and `Picture::RenderToTexture`. Because the default behavior was changed to remove the stencil, these utilities end up trying to render an `EntityPass` to a `RenderTarget` with no configured stencil attachment.

This fix inverts the default behavior and explicitly removes the stencil from `ContentContext::MakeSubpass` instead.


jonahwilliams (approved):
LGTM

PR #39651, opened 2023-02-15
Remove "bringup: true" from "Linux Fuchsia FEMU"

zanderso:
The emulator flakes appear to have been addressed.

There is a still a test flaking like here https://ci.chromium.org/ui/p/flutter/builders/staging/Linux%20Fuchsia%20FEMU/532/overview, but it is at a much lower rate, and from chat, I think it will soon be fixed by an upcoming Fuchsia SDK roll.


eyebrowsoffire (approved):
LGTM

PR #39654, opened 2023-02-15
[Impeller] align SSBOs to 16 bytes to fix iOS shader validation issue.

jonahwilliams:
Fix https://github.com/flutter/flutter/issues/120817 by doing what the nice shader validation error message says

### DIFF
--- a/impeller/entity/contents/gradient_generator.h
+++ b/impeller/entity/contents/gradient_generator.h
@@ -20,10 +20,6 @@ namespace impeller {
 
 class Context;
 
-/// Lower iOS version require alignment to be a multiple of 16, so the alignof
-/// operator will not give correct results for most structs.
-constexpr size_t kSSBOAlignOf = 16u;
-
 /**
  * @brief Create a host visible texture that contains the gradient defined
  * by the provided gradient data.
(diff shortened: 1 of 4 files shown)


chinmaygarde (changes requested):
I believe `DefaultUniformAlignment` will give you the right alignment. The name is an unfortunate artifact of a time when we didn't have SSBOs. Just uniforms which have the same alignment rules. I see `EmplaceUniform` and `EmplaceStorageBuffer` which apply this alignment correctly but I guess we didn't use it here because these don't have a `vector` specialization.

`EmplaceUniform` and `EmplaceStorageBuffer` are identical and should be aliased. They should also have a vector or contiguous collection specialization. Can we file a bug to fix both and bring the naming conventions of `impeller::HostBuffer` into the modern age?

W.r.t this patch, lets replace `kSSBOAlignOf` with `impeller::DefaultUniformAlignment` (also unfortunately named and we can fix in a refactoring pass later) so its right on more platforms. Then we can land this while we wait for a refactor.

jonahwilliams:
Done

chinmaygarde (approved):

PR #39657, opened 2023-02-15
[web] consolidate network code into httpFetch

yjbanov:
Consolidate all of web engine HTTP code into a single `httpFetch` function that provides a standard way for handling errors:

- Provide `httpFetch` and a few specialize convenience functions around it for making HTTP calls.
- Provide a standard set of mock classes for mocking HTTP requests in unit tests.
- Migrate all of our code to `httpFetch` (fonts, assets, network images,

### DIFF
--- a/lib/web_ui/dev/test_platform.dart
+++ b/lib/web_ui/dev/test_platform.dart
@@ -329,7 +329,7 @@ class BrowserPlatform extends PlatformPlugin {
   Future<shelf.Response> _testPayloadGenerator(shelf.Request request) async {
     if (!request.requestedUri.path.endsWith('/long_test_payload')) {
       return shelf.Response.notFound(
-          'This request is not handled by the screenshot handler');
+          'This request is not handled by the test payload generator');
     }
 
     final int payloadLength = int.parse(request.requestedUri.queryParameters['length']!);
--- a/lib/web_ui/dev/test_platform.dart
+++ b/lib/web_ui/dev/test_platform.dart
@@ -90,6 +90,12 @@ class BrowserPlatform extends PlatformPlugin {
         // This handler goes last, after all more specific handlers failed to handle the request.
         .add(_createAbsolutePackageUrlHandler())
         .add(_screenshotHandler)
+
+        // Generates and serves a test payload of given length, split into chunks
+        // of given size. Reponds to requests to /long_test_payload.
+        .add(_testPayloadGenerator)
+
+        // If none of the handlers above handled the request, return 404.
         .add(_fileNotFoundCatcher);
 
     server.mount(cascade.handler);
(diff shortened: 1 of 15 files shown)


hterkelsen (approved):
LGTM

lib/web_ui/dev/test_platform.dart
```diff
@@ -320,6 +326,46 @@ class BrowserPlatform extends PlatformPlugin {
     };
   }
 
+  Future<shelf.Response> _testPayloadGenerator(shelf.Request request) async {
+    if (!request.requestedUri.path.endsWith('/long_test_payload')) {
+      return shelf.Response.notFound(
+          'This request is not handled by the screenshot handler');
```
mdebbar, line 332: copy-paste?

mdebbar (changes requested):

mdebbar (commented):
Looks great!

lib/web_ui/dev/test_platform.dart
yjbanov: Yep! Fixed.

mdebbar (approved):

lib/web_ui/lib/src/engine/dom.dart
```diff
@@ -756,56 +766,400 @@ extension DomCanvasGradientExtension on DomCanvasGradient {
...
+/// payload handling, including error checking, is sufficient, consider using
+/// convenience functions [httpFetchByteBuffer], [httpFetchJson], or
+/// [httpFetchText] instead.
+Future<HttpFetchResponse> httpFetch(String url) async {
+  if (mockHttpFetchResponseFactory != null) {
+    return mockHttpFetchResponseFactory!(url);
+  }
+  try {
+    final _DomResponse domResponse = await _rawHttpGet(url);
+    return HttpFetchResponseImpl._(url, domResponse);
```
ditman, line 796: Since _DomResponse is a `@staticInterop` object, you may be able to simplify this a bunch by just making the `mockHttpFetchResponseFactory` return a

```
js_util.jsify({
  // Something that resembles a _DomResponse
}) as _DomResponse;
```

(Or even create a real _DomResponse object with its [constructor](https://developer.mozilla.org/en-US/docs/Web/API/Response/Response)) 

That would allow you to simplify the objects returned by fetch quite a bit, and also:

```
try {
  _DomResponse response;
  if (mockHttpFetchResponseFactory != null) {
    response = await mockHttpFetchResponseFactory!(url);
  } else {
    response = await _rawHttpGet(url);
  }
  return HttpFetchResponseImpl._(url, domResponse);
} catch (...) {
  //...
}
```

test what happens when the mockHttpRequestFactory throws.

ditman (commented):
🚀

PR #39632, opened 2023-02-14
[macOS] Set textfield autofill type

cbracken:
macOS supports three autofill types: username, password, and oneTimeCode. Apply these content types when the equivalent Flutter autofill hints are set.

Note that enabling autocomplete of username and password in applications will still require setting the "Autofill Credential Provider" and "Associated Domains" entitlements, and for the FlutterTextInputPlugin to be updated to use an NSSecureTextInput input type.

Issue: https://github.com/flutter/flutter/issues/120252

## Pre-launch Checklist

- [X] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [X] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [X] I read and followed the [Flutter Style Guide] and the [C++, Objective-C, Java style guides].
- [X] I listed at least one issue that this PR fixes in the description above.
- [X] I added new tests to check the change I am making or feature I am adding, or Hixie said the PR is test-exempt. See [testing the engine] for instructions on writing and running engine tests.
- [X] I updated/added relevant documentation (doc comments with `///`).
- [X] I signed the [CLA].
- [X] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/flutter/wiki/Tree-hygiene#overview
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[C++, Objective-C, Java style guides]: https://github.com/flutter/engine/blob/main/CONTRIBUTING.md#style
[testing the engine]: https://github.com/flutter/flutter/wiki/Testing-the-engine
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/wiki/Chat

PR #39664, opened 2023-02-16
[macOS] Improve TextInputPlugin test readability

cbracken:
This just extracts arguments to the TextInputClient.setClient async method call into an NSDictionary local to reduce the deep indentation of the formatted code.

While the bulk of these calls are identical and could be extracted to a utility function, or a class extension method like `handleDefaultSetClientCall`, the point of these tests is to exercise FlutterTextInputPlugin, and the message protocol over the channel are precisely what's  being tested, so it makes sense to keep these explicit.

Followup patch to https://github.com/flutter/engine/pull/39632.

Issue: https://github.com/flutter/flutter/issues/120252

## Pre-launch Checklist

- [X] I signed the [CLA].

### DIFF
--- a/shell/platform/darwin/macos/framework/Source/FlutterTextInputPluginTest.mm
+++ b/shell/platform/darwin/macos/framework/Source/FlutterTextInputPluginTest.mm
@@ -69,14 +69,12 @@
   FlutterTextInputPlugin* plugin =
       [[FlutterTextInputPlugin alloc] initWithViewController:viewController];
 
-  [plugin handleMethodCall:[FlutterMethodCall
-                               methodCallWithMethodName:@"TextInput.setClient"
-                                              arguments:@[
-                                                @(1), @{
-                                                  @"inputAction" : @"action",
-                                                  @"inputType" : @{@"name" : @"inputName"},
-                                                }
-                                              ]]
+  NSDictionary* setClientConfig = @{
+    @"inputAction" : @"action",
+    @"inputType" : @{@"name" : @"inputName"},
+  };
+  [plugin handleMethodCall:[FlutterMethodCall methodCallWithMethodName:@"TextInput.setClient"
+                                                             arguments:@[ @(1), setClientConfig ]]
                     result:^(id){
                     }];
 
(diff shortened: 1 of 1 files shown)


shell/platform/darwin/macos/framework/Source/FlutterTextInputPluginTest.mm
```diff
@@ -431,30 +419,22 @@ - (bool)testAutocompleteEnabledWhenAutofillSet {
...
                     result:^(id){
                     }];
 
   // Verify autocomplete is enabled.
   EXPECT_TRUE([plugin isAutomaticTextCompletionEnabled]);
-
-  // Verify content type is nil for unsupported content types.
-  if (@available(macOS 11.0, *)) {
-    EXPECT_EQ([plugin contentType], nil);
-  }
```
loic-sharma, line 456: Just to confirm, is this deletion intentional?

cbracken: Good catch - no! That code showed up about 50 lines down below and caused another test to fail; on looking at it there, I was thinking "why would I add this?!" Answer: apparently an inadvertent cut/paste!

shell/platform/darwin/macos/framework/Source/FlutterTextInputPluginTest.mm
```diff
@@ -658,46 +634,6 @@ - (bool)testContentTypeWhenAutofillTypeIsUsername {
   return true;
 }
 
-- (bool)testContentTypeWhenAutofillTypeIsOneTimeCode {
```
loic-sharma, line 661: Is this deletion intentional?

shell/platform/darwin/macos/framework/Source/FlutterTextInputPluginTest.mm
```diff
@@ -618,7 +595,7 @@ - (bool)testAutocompleteDisabledWhenAutofillGroupIncludesPassword {
   return true;
 }
 
-- (bool)testContentTypeWhenAutofillTypeIsUsername {
+- (bool)testContentTypeWhenAutofillGroupIncludesPassword {
```
loic-sharma, line 598: It loooks like this is two tests: username content type, and `obscureText` config. Should we split these up into two separate tests? I don't feel strongly either ways.

loic-sharma (approved):
LGTM

shell/platform/darwin/macos/framework/Source/FlutterTextInputPluginTest.mm
cbracken: I think the diff was playing tricks here due to the deleted test right next to it; after restoring and re-pushing it looks right to me. The test without obscureTest checks the content type is username. The one with obscureText checks password autofill type.

PR #39683, opened 2023-02-16
Remove extraneous if

dnfield:
The compiler probably does this anyway, but unless I'm misreading the code and still waking up I'm pretty sure this is what we actually want.


Hixie:

bdero (approved):

chinmaygarde (approved):

PR #39685, opened 2023-02-16
Match PlatformConfiguration properties to PlatformDispatcher ones

pdblasi-google:
* Updated PlatformDispatcher to use PlatformConfiguration to back `nativeSpellCheckServiceDefined` and `brieflyShowPassword`
* Updated web PlatformDispatcher/Configuration to match new API

Resolves [#120840](https://github.com/flutter/flutter/issues/120840)

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/testing/dart/platform_dispatcher_test.dart
+++ b/testing/dart/platform_dispatcher_test.dart
@@ -2,13 +2,8 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
-import 'dart:io';
 import 'dart:ui';
 
-import 'package:analyzer/dart/analysis/features.dart';
-import 'package:analyzer/dart/analysis/results.dart';
-import 'package:analyzer/dart/analysis/utilities.dart';
-import 'package:analyzer/dart/ast/ast.dart';
 import 'package:litetest/litetest.dart';
 
 void main() {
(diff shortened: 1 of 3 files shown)


loic-sharma (approved):

testing/dart/platform_dispatcher_test.dart
```diff
@@ -39,4 +44,60 @@ void main() {
...
+      final List<Pattern> exclusionCriteria = <Pattern>[
+        RegExp(r'^_.*'), // exclude any private members
+        RegExp(r'^on[A-Z].*'), // exclude any callbacks
+        RegExp(r'locale$'), // locale is a convenience getter for interacting with `locales` which _is_ backed by the config
+        'instance', // exclude the static instance of PlatformDispatcher
+        'configuration', // the configuration should not reference itself
+        'views', // views are backed by their own config and don't need to be referenced in the platform config
+        'initialLifecycleState', // default route is stored/retrieved from the platform, not the engine
+        'frameData', // framed data updates too often and would kick off too many config changes
+      ];
```
pdblasi-google, line 58: @loic-sharma could you give me a spot-check on the justifications for these exclusions? Specifically `initialLifecycleState` and `frameData`. I'm not sure on the reasons why those properties _weren't_ backed by the config, but their implementations were different enough that I expected they _were_ intended to be different.

testing/dart/platform_dispatcher_test.dart
```diff
@@ -2,8 +2,13 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+import 'dart:io';
```
ditman, line 5: I'm not entirely sure if this test is expected to run in a web browser, but importing dart:io probably will break that?

testing/dart/platform_dispatcher_test.dart
```diff
@@ -39,4 +44,60 @@ void main() {
...
+      final Set<String> configurationMembers = configurationClass.members
+        .where((ClassMember member) =>
+          member is FieldDeclaration ||
+          member is MethodDeclaration && member.isGetter)
+        .expand(getNameOfClassMember)
+        .toSet();
+
+      final Set<String> missingMembers = configurationMembers.difference(dispatcherMembers);
+      final Set<String> extraMembers = dispatcherMembers.difference(configurationMembers);
+      expect(missingMembers.isEmpty && extraMembers.isEmpty, isTrue,
```
ditman, line 97: This test looks a little bit funky to me, isn't the "exposes the same getters" an OOP concern (like a "PlatformConfigurationProvider" mixin or similar) rather than something that needs to be asserted in a test?

The compiler is good at flagging these things :P

ditman (commented):
The config code change, and exposing it through the platform dispatcher looks good, however the test for this looks a little bit weird to me? Isn't there a practical way of leveraging OOP more?

testing/dart/platform_dispatcher_test.dart
pdblasi-google: Hmmm. I don't _think_ this test is intended to run in the web. `web_ui` has it's own set of tests to run as far as I know, and this test should be extending `dart:ui` ones.

testing/dart/platform_dispatcher_test.dart
pdblasi-google: I'd normally agree with you, but the issue that let to this work in the first place was that someone added to a class without adding to another class. I don't think that adding an _additional_ place that things need to be added/implemented is going to resolve that, where this test does give a warning in the same situation of "someone added to PlatformDispatcher without following up elsewhere".

That said, the test requires `package:analyzer`, and I'm not sure I'll be able to swing pulling that dependency into the tests, so it may be a moot point.

ditman: I'm not trying to block this at all. I'm just saying that needing a test like this is a smell of something that needs to be solved elsewhere.

It might be an OOP/design issue (lacking an interface, or maybe exposing the `config` instance from the dispatcher so people use getters from it, rather than having this 1:1 mapping), or maybe having 1.3k lines in a file makes it harder for people to reason about *where* and how to add things (and the code reviewer to pick up those differences).

pdblasi-google: Fair points all. I've talked with @goderbauer a bit about this, and we were actually floating the idea of _removing_ the config part of the equation. We weren't able to track down the original reasoning for having the config in the first place though, so we weren't able to tell how impactful actually removing it would be.

I personally want to leave the config for now, possibly file an issue or something to re-evaluate whether `PlatformDispatcher` and `FlutterView` actually _need_ the configs, or whether they can be simplified.

testing/dart/platform_dispatcher_test.dart
pdblasi-google: Never mind, had to remove this test due to the `analyzer` dependency.

Hixie:

PR #39689, opened 2023-02-16
[impeller] support generating mip-maps on Vulkan

iskakaushik:
fix: https://github.com/flutter/flutter/issues/120134

### DIFF
--- a/impeller/renderer/backend/vulkan/allocator_vk.cc
+++ b/impeller/renderer/backend/vulkan/allocator_vk.cc
@@ -116,6 +116,7 @@ std::shared_ptr<Texture> AllocatorVK::OnCreateTexture(
   image_create_info.initialLayout = vk::ImageLayout::eUndefined;
   image_create_info.usage = vk::ImageUsageFlagBits::eSampled |
                             vk::ImageUsageFlagBits::eColorAttachment |
+                            vk::ImageUsageFlagBits::eTransferSrc |
                             vk::ImageUsageFlagBits::eTransferDst;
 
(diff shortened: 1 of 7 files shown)


iskakaushik:
image

impeller/renderer/backend/vulkan/allocator_vk.cc
```diff
@@ -116,6 +116,7 @@ std::shared_ptr<Texture> AllocatorVK::OnCreateTexture(
   image_create_info.initialLayout = vk::ImageLayout::eUndefined;
   image_create_info.usage = vk::ImageUsageFlagBits::eSampled |
```
chinmaygarde, line 117: This makes me think we should perhaps have a recommended usage as on optional on the texture descriptor. For later though.

shell/platform/android/platform_view_android.cc
```diff
@@ -47,7 +47,7 @@ std::unique_ptr<AndroidSurface> AndroidSurfaceFactoryImpl::CreateSurface() {
     case AndroidRenderingAPI::kOpenGLES:
       if (enable_impeller_) {
 // TODO(kaushikiska@): Enable this after wiring a preference for Vulkan backend.
-#if false
+#if true
```
chinmaygarde, line 50: Did you mean to leave this in?

chinmaygarde (approved):
LGTM but I think you forgot to back out the part where you opted into Vulkan on Android.

shell/platform/android/platform_view_android.cc
iskakaushik: nope, typo!

PR #39692, opened 2023-02-16
[web] Make glassPaneElement and glassPaneShadow non-nullable

mdebbar:


ditman (approved):
LGTM! If we're (almost) always using it with `!`, it's probably well initialized for all cases!

htoor3 (approved):
lgtm

PR #39696, opened 2023-02-16
[gn + codesign] add android artifact metadata to legacy build target

XilaiZhang:
[gn + codesign] add android artifact metadata to legacy build target

Currently, android target are still build through the legacy job target flutter/shell/platform/android:gen_snapshot.
And the metadata only exists on the new target flutter/shell/platform/android:new_gen_snapshot,
therefore codesign metadata is still missing in the old gn build target.


XilaiZhang:
Need to verify this with a local build + test. Currently android artifacts aren't uploaded back to gcloud.

zanderso (approved):

PR #39698, opened 2023-02-17
[gn + codesign] add metadata for path ops

XilaiZhang:
Adds lib path ops .dylib to meta data file.

Currently, when building "//flutter/shell/testing:testing", it adds a duplicate entry of  "impellerc" to the metadata file of artifacts.zip. I suspect there is a dependency of "//flutter/shell/testing:testing" on "//flutter/impeller/compiler:impellerc", but I traced a few hours and didn't track down why there is a duplicate entry of "impellerc".
Since the codesign handles duplicate entries already, I will leave the duplicate entry in the metadata file.


christopherfujino:
as I have no idea what's going on here or how gn works, I will defer approval to Godofredo :)

christopherfujino:
> Since the codesign handles duplicate entries already, I will leave the duplicate entry in the metadata file.

Does it handle it by skipping repeat entries?

XilaiZhang:
> > Since the codesign handles duplicate entries already, I will leave the duplicate entry in the metadata file.
> 
> Does it handle it by skipping repeat entries?

Yeah the entries are compiled into a set https://github.com/flutter/cocoon/blob/main/codesign/lib/src/file_codesign_visitor.dart#L65 to eliminate duplicates

godofredoc (commented):
Does this change need a rebase?

build/archives/BUILD.gn
```diff
@@ -16,6 +16,7 @@ generated_file("artifacts_entitlement_config") {
       "//flutter/impeller/compiler:impellerc",
       "//flutter/impeller/tessellator:tessellator_shared",
       "//flutter/shell/testing:testing",
+      "//flutter/tools/path_ops:path_ops",
```
godofredoc, line 19: Why do we need to add this dep?

XilaiZhang: yeah this dependency is needed to collect the metadata from tools/path_ops/BUILD.gn.

we defined a data_keys field in this target. So that if a matching data key is present in tools/path_ops/BUILD.gn, it is propagated and collected into the meta data file.

XilaiZhang:
> Does this change need a rebase?

I created it recently, but it doesn't hurt to rebase again. I will rebase now.

chinmaygarde (approved):

zanderso (approved):

PR #39718, opened 2023-02-17
Adds a Linux Fuchsia FEMU config that enables CSO

zanderso:
This configuration exposed Fuchsia issues that were not surfaced in any other environment when it was enabled prematurely last week. This change adds a `bringup: true` builder for the new configuration. When it stabilizes, `enable_cso` will become the default, and the `bringup: true` builder can be deleted.

`enable_cso` was added here https://flutter-review.git.corp.google.com/c/recipes/+/39400.

### DIFF
--- a/.ci.yaml
+++ b/.ci.yaml
@@ -156,6 +156,19 @@ targets:
       emulator_arch: "x64"
     timeout: 60
 
+  - name: Linux Fuchsia FEMU (cso)
+    recipe: engine/femu_test
+    bringup: true
+    properties:
+      add_recipes_cq: "true"
+      build_fuchsia: "true"
+      fuchsia_ctl_version: version:0.0.27
+      # ensure files from pre-production Fuchsia SDK tests are purged from cache
+      clobber: "true"
+      emulator_arch: "x64"
+      enable_cso: "true"
+    timeout: 60
+
   - name: Linux Fuchsia arm64 FEMU
     recipe: engine/femu_test
     bringup: true
--- a/.ci.yaml
+++ b/.ci.yaml
@@ -160,14 +160,13 @@ targets:
     recipe: engine/femu_test
     bringup: true
     properties:
-      add_recipes_cq: "true"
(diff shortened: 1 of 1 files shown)


.ci.yaml
```diff
@@ -156,6 +156,19 @@ targets:
       emulator_arch: "x64"
     timeout: 60
 
+  - name: Linux Fuchsia FEMU (cso)
+    recipe: engine/femu_test
+    bringup: true
+    properties:
+      add_recipes_cq: "true"
```
godofredoc, line 163: False? as this build will be removing caches it will slow down recipes testing. Femu_test recipe is already being covered by the other builders.

.ci.yaml
```diff
@@ -156,6 +156,19 @@ targets:
...
+      build_fuchsia: "true"
+      fuchsia_ctl_version: version:0.0.27
```
godofredoc, line 165: This is not needed, fuchsia_ctl was removed a few months ago.

.ci.yaml
```diff
@@ -156,6 +156,19 @@ targets:
...
+      # ensure files from pre-production Fuchsia SDK tests are purged from cache
+      clobber: "true"
+      emulator_arch: "x64"
+      enable_cso: "true"
+    timeout: 60
```
godofredoc, line 170: Would 60 mins be enough? usually checking out all the repositories from a clean cache takes ~14 mins.

.ci.yaml
zanderso: This is a copy paste from the other Fuchsia config. Let's remove in a separate cleanup.

.ci.yaml
zanderso: Increased to 90

.ci.yaml
zanderso: You tell me =) I have no idea what the flag is supposed to mean. I just copy-pasted from the other config.

godofredoc (approved):

PR #39738, opened 2023-02-18
[Windows] Mark move-only classes as such

cbracken:
Applies the FML_DISALLOW_COPY_AND_ASSIGN to non-POD types in the Windows embedder.
Also #includes fml/macros.h in a few places we were already using the macro, but missed int #include.

No new tests since no semantic change. Only has the compile-time effect of preventing copying of classes intended to be move-only.

## Pre-launch Checklist

- [X] I signed the [CLA].

### DIFF
--- a/shell/platform/windows/flutter_window.h
+++ b/shell/platform/windows/flutter_window.h
@@ -8,6 +8,7 @@
 #include <string>
 #include <vector>
 
+#include "flutter/fml/macros.h"
 #include "flutter/shell/platform/common/geometry.h"
 #include "flutter/shell/platform/embedder/embedder.h"
 #include "flutter/shell/platform/windows/flutter_windows_view.h"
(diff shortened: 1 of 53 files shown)


shell/platform/windows/flutter_window.h
```diff
@@ -168,6 +169,8 @@ class FlutterWindow : public Window, public WindowBindingHandler {
 
   // The cursor rect set by Flutter.
   RECT cursor_rect_;
+
+  FML_DISALLOW_COPY_AND_ASSIGN(FlutterWindow);
```
yaakovschectman, line 173: Should `Window` in `window.h` get the same treatment?

yaakovschectman (approved):
LGTM, mod one question

shell/platform/windows/flutter_window.h
cbracken: Good question! Window is an abstract class so already not copyable/movable.

cbracken: Actually, I'm going to pass on that for now as it would involve adding a bunch of default constructors since we're explicitly deleting other ctors.

PR #39765, opened 2023-02-20
[web] Add explicit JS type test for `JsStrategy`.

joshualitt:
Type checking is not well defined with JS interop. A safer approach is to type check JS values in JS using js_util. This CL implements such an explicit test.

### DIFF
--- a/lib/web_ui/lib/src/engine/initialization.dart
+++ b/lib/web_ui/lib/src/engine/initialization.dart
@@ -273,8 +273,18 @@ Future<void> _downloadAssetFonts() async {
 
 void _addUrlStrategyListener() {
   jsSetUrlStrategy = allowInterop((JsUrlStrategy? jsStrategy) {
-    customUrlStrategy =
-        jsStrategy == null ? null : CustomUrlStrategy.fromJs(jsStrategy);
+    if (jsStrategy == null) {
+      customUrlStrategy = null;
+    } else {
+      // Because `JSStrategy` could be anything, we check for the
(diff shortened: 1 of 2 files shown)


joshualitt:
@mdebbar ptal, I think you may have added this test a few years ago. I'm not sure how we want to handle this, but this tes relies on some unstable behavior of JS interop. In particular, JS interop APIs don't have the same type checking guarantees that Dart APIs do, and as we iterate on the new JS interop, these guarantees will be weakened further.

I'm not sure the best path forward. I think the best 'type checking' we can do is ensure that when we `get` this property, it has a `typeof === 'function'`.

mdebbar:
Would it work if we added type-checking here? : https://github.com/flutter/engine/blob/eb5e5624dedad5eebc7f4de4aea3ae51d8e9d9a5/lib/web_ui/lib/src/engine/initialization.dart#L275-L278



> I'm not sure the best path forward. I think the best 'type checking' we can do is ensure that when we `get` this property, it has a `typeof === 'function'`.

We are not type-checking the function here, we are type-checking the arguments passed to the function. We need to make sure the argument is a `JsUrlStrategy`, but I'm not sure how to do that in the static js interop world. Maybe check that some properties exist on the object? e.g. `js_util.hasProperty(obj, 'addPopStateListener')`

joshualitt:
Good call, I added a check for `addPopStateListener` in the non-null case. ptal.

mdebbar (approved):
Thanks for fixing this!

PR #39768, opened 2023-02-21
[iOS][Text Input] Avoid crash when its UIViewController.view is nil.

luckysmg:
Fix:
 - https://github.com/flutter/flutter/issues/106404

## Pre-launch Checklist

- [x] I signed the [CLA].


shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
```diff
@@ -2081,6 +2081,11 @@ - (void)removeEnableFlutterTextInputViewAccessibilityTimer {
 }
 
 - (void)handleMethodCall:(FlutterMethodCall*)call result:(FlutterResult)result {
+  if (_viewController.view == nil) {
```
hellohuanlin, line 2084: It isn't obvious why we need it here. Can you add a comment? you can also link to the github issue.

shell/platform/darwin/ios/framework/Source/FlutterTextInputPluginTest.mm
```diff
@@ -166,6 +166,20 @@ - (FlutterTextRange*)getLineRangeFromTokenizer:(id<UITextInputTokenizer>)tokeniz
...
 
+- (void)testTextInputPluginWillNotHandleMethodWhenViewControllerIsNil {
+  FlutterEngine* flutterEngine = [[FlutterEngine alloc] init];
+  FlutterTextInputPlugin* inputPlugin =
+      [[FlutterTextInputPlugin alloc] initWithDelegate:(id<FlutterTextInputDelegate>)flutterEngine];
+  XCTAssertNil(inputPlugin.viewController);
+  FlutterMethodCall* methodCall = [FlutterMethodCall methodCallWithMethodName:@"TextInput.show"
+                                                                    arguments:nil];
+  [inputPlugin handleMethodCall:methodCall
+                         result:^(id _Nullable result){
```
hellohuanlin, line 177: you may wanna validate that the result is actually called with nil argument.

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
luckysmg: Done O(∩_∩)O

shell/platform/darwin/ios/framework/Source/FlutterTextInputPluginTest.mm
luckysmg: Done O(∩_∩)O

hellohuanlin (approved):

LongCatIsLooong:
Could you help me understand when this could happen? From the stacktrace  here: https://github.com/flutter/flutter/issues/106404#issuecomment-1210983394, not handling 
```
-[FlutterTextInputPlugin setTextInputClient:withConfiguration:]
```
could be disastrous (since the framework always assumes the connection is successfully established), unless the engine is about to be torn down. But if the engine is being deallocated we're not supposed to get any platform messages right?

luckysmg:
In add-to-app scene. We will share one engine with multiple ViewController, and we will set engine's view controller to nil manually sometimes. Any if some message's coming will trigger this crash....

luckysmg:
Here is the demo code to reproduce 100% with this crash.
in `main.dart`:

<details>

<summary>Dart code</summary>

```dart

void main() {
  runApp(const MyApp());
}

  const MyApp({super.key});

  // This widget is the root of your application.
  @override
    return MaterialApp(
      title: 'Flutter Demo',
      theme: ThemeData(
        colorScheme: ColorScheme.fromSeed(seedColor: Colors.deepPurple),
        useMaterial3: true,
      ),
      home: const MyHomePage(title: 'Flutter Demo Home Page'),
    );
  }
}

class MyHomePage extends StatefulWidget {
  const MyHomePage({super.key, required this.title});

  final String title;

  @override
  State<MyHomePage> createState() => _MyHomePageState();
}

class _MyHomePageState extends State<MyHomePage> {

  FocusNode node = FocusNode();

  @override
  void initState() {
    super.initState();
    Future.delayed(const Duration(seconds: 5),() {
      node.requestFocus();
    });
  }

  @override
    return Scaffold(
      appBar: AppBar(
        backgroundColor: Theme.of(context).colorScheme.inversePrimary,
        title: Text(widget.title),
      ),
      body: Center(
        child: Column(
          mainAxisAlignment: MainAxisAlignment.center,
          children: <Widget>[
            const Text(
              'You have pushed the button this many times:',
            ),
            TextField(
              focusNode: node,
            ),
          ],
        ),
      ),
    );
  }
}

```

</details>





in `FlutterAppDelegate.swift`


<details>

<summary>swift code</summary>

```swift
import UIKit
import Flutter

@UIApplicationMain
@objc class AppDelegate: FlutterAppDelegate {
  override func application(
    _ application: UIApplication,
    didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?
  ) -> Bool {
    GeneratedPluginRegistrant.register(with: self)
      
      
      let vc = window.rootViewController as! FlutterViewController;
      DispatchQueue.main.asyncAfter(deadline: DispatchTime.now() + 3){
          let engine = vc.engine
          engine?.viewController = nil
      }
      
    return super.application(application, didFinishLaunchingWithOptions: launchOptions)
  }
}

```

</details>

It is very easy for developers to write code like this (delay some time request focus), if we detach the view controller from engine(This is very common in add-to-app scene), it will cause crash. I think we should do something to fallback, maybe not perfect, but at least, it will be better than crash directly ^_^

luckysmg:
I have modified the code. I think it's better than before. Plz re-review ^_^

hellohuanlin (approved):

PR #39779, opened 2023-02-21
[Impeller] Stencil formats aspect mask fix

iskakaushik:
Ideally the texture descriptor should have the information to indicate the intended texture usage but this will let us get to the finish line quicker.


chinmaygarde (changes requested):
I fixed stencils and depth and am working on MSAA and resource tracking in https://github.com/flutter/engine/compare/flutter:3aa112c...chinmaygarde:24d0e05#. I commented in chat before the weekend but not sure you saw. I think that patch is going to be a superset of this. Can we hold on landing this?

PR #39781, opened 2023-02-21
[Impeller] texture coordinates implementation

jonahwilliams:
~~A not-particularly-fast implementation of texture coordinates. This is missing a fairly reasonable fast path which is 1. texture coordinates + image shader + no per vertex colors should not require any subpasses.~~ Fast path added.

This implementation is a bit complicated by the fact that both the geometetry and the contents need to know the "size" of the color source that the texture coordinates are applied to. From experimentation with Skia, the extent of the gradient is used where possible, as is the size of the image - as the texture coordinates are always given in the local coordinate space.

For solid colors we should probably elide the subpass and bail, but that doesn't seem particularly important. For sweep gradients and runtime effect, its not clear what size we should use so I'm falling back to using the overall geometry coverage.

Fixes https://github.com/flutter/flutter/issues/109956


jonahwilliams:
still needs effect transform

jonahwilliams:
I should double check, but I suppose a reasonable way to interpret the texture coordinates for color sources without sizes is to take the maximum coverage from the texture coordinates. I'll do some experiments with Skia.

jonahwilliams:
Added fast path for ImageShaders and use texture coordinate extent for other shader types. Not going to spend much more time on that, I don't think its important.

Need to do effect transform still.

impeller/display_list/display_list_vertices_geometry.cc
```diff
@@ -225,19 +254,89 @@ GeometryResult DLVerticesGeometry::GetPositionColorBuffer(
...
+  std::vector<VS::PerVertexData> vertex_data(vertex_count);
+  {
+    for (auto i = 0; i < vertex_count; i++) {
+      auto sk_point = dl_vertices[i];
+      auto texture_coord = dl_texture_coordinates[i];
+      auto uv = Point((texture_coord.x() - origin.x) / size.width,
+                      (texture_coord.y() - origin.y) / size.height);
+      // From experimentation we need to clamp these values to < 1.0 or else
+      // there can be flickering.
+      vertex_data[i] = {
```
jonahwilliams, line 284: Observed flickering while scrolling when using vertices + texture coordinates and a gradient in a subpass. Geometry and stencil were correct

jonahwilliams:
Okay, i have hacked in support for the effect transform for textures!

jonahwilliams:
I removed the position_uv vertex shader because it doesn't do anything that the texture_fill.vert shader doesn't do

impeller/aiks/canvas.cc
```diff
@@ -386,7 +386,12 @@ void Canvas::DrawVertices(const std::shared_ptr<VerticesGeometry>& vertices,
   entity.SetStencilDepth(GetStencilDepth());
   entity.SetBlendMode(paint.blend_mode);
 
-  if (!vertices->HasVertexColors()) {
+  // If there are no vertex color or texture coordinates. Or if there
+  // are vertex coordinates then only if the contents are an image.
+  if (!vertices->HasVertexColors() &&
+      (!vertices->HasTextureCoordinates() ||
+       (vertices->HasTextureCoordinates() &&
+        paint.color_source_type == Paint::ColorSourceType::kImage))) {
```
dnfield, line 394: nit: I think this and the `if` below would read a little clearer if there was a method on `Vertices` to help sort these things out.

impeller/display_list/display_list_vertices_geometry.cc
```diff
@@ -225,19 +254,91 @@ GeometryResult DLVerticesGeometry::GetPositionColorBuffer(
...
+      auto texture_coord = dl_texture_coordinates[i];
+      auto uv = effect_transform *
+                Point((texture_coord.x() - origin.x) / size.width,
+                      (texture_coord.y() - origin.y) / size.height);
+      // From experimentation we need to clamp these values to < 1.0 or else
+      // there can be flickering.
+      vertex_data[i] = {
+          .position = Point(sk_point.x(), sk_point.y()),
+          .texture_coords = Point(std::clamp(uv.x, 0.0f, 0.995f),
+                                  std::clamp(uv.y, 0.0f, 0.995f)),
```
dnfield, line 289: Can the second constant be `1.0 - kEhCloseEnough`?

dnfield: (that would avoid introducing a new epsillon, which this feels like)

dnfield (approved):
LGTM with nits

impeller/display_list/display_list_vertices_geometry.cc
jonahwilliams: Yeah good idea. I was actually surprised a bit by this behavior, maybe I'm actually doing the math wrong?

impeller/aiks/canvas.cc
jonahwilliams: There's not really a good place to put this. it depends on the color source type which can't be used outside of aiks, and the vertices are a geometry object defined in entity. It can't go on the contents because it isn't created yet.

jonahwilliams: moved this to a static helper function

impeller/tools/malioc_diff.py
```diff
@@ -266,7 +268,7 @@ def main(argv):
   if args.update:
     # Write the new results to the file given by --before, then exit.
     with open(args.before, 'w') as file:
-      json.dump(after_json, file, sort_keys=True)
+      json.dump(after_json, file, sort_keys=True, indent=2)
```
jonahwilliams, line 271: Splitting each line out makes the diff view a lot easier to read

jonahwilliams: reverted this for now to avoid conflicts

PR #39785, opened 2023-02-22
Add vsync support to Linux shell

robert-ancell:
Implement vsync_callback to ensure that Flutter renders frames at the optimum time/speed.

### DIFF
--- a/shell/platform/linux/fl_engine.cc
+++ b/shell/platform/linux/fl_engine.cc
@@ -48,6 +48,12 @@ struct _FlEngine {
   FLUTTER_API_SYMBOL(FlutterEngine) engine;
   FlutterEngineProcTable embedder_api;
 
+  // Function to call when vsync is requested.
+  intptr_t vsync_baton;
+  FlEngineOnPreEngineRestartHandler vsync_handler;
+  gpointer vsync_handler_data;
+  GDestroyNotify vsync_handler_destroy_notify;
+
   // Function to call when a platform message is received.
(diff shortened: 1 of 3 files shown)


robert-ancell:
@jpnurmi can you try this and see if it helps with flickering issues you have seen?

jpnurmi:
It renders fine in Wayland but I get a blank window in X11. The values look like this:

Wayland:
- `frame_time_msec`: 12464413584
- `refresh_interval`: 16667
- `presentation_time`: 0

X11:
- `frame_time_msec`: 12280308676
- `refresh_interval`: 16674
- `presentation_time`: 12280359145

robert-ancell:
I've updated with a fix that hopefully works in X11. I don't really understand the information that GTK is giving me though.

jpnurmi:
Now rendering works in X11, thanks! Unfortunately, I'm still witnessing some flickering corners. :(

cbracken (commented):
After talking through it with @robert-ancell today we think this is probably unnecessary. GTK is already implicitly ensuring we're tied to vsync via draw callbacks, and the engine's `Pipeline` already ensures that calling PlatformDispatcher.scheduleFrame() already prevents duplicate frame builds between vsyncs.

Closing for now. There may be further work to do when we get to the point of multi-window.

HookedBehemoth:
I tried this patch but noticed that gdk timers would always emit 60fps timers.
Other GTK applications appear to also be locked to 60fps under X11 but on Wayland they are synched correctly.

PR #39788, opened 2023-02-22
[Windows] Stop engine before destructing FlutterWindowsView

scutlight:
Stop engine before destructing FlutterWindowsView. If not, a running engine on raster thread might call into FlutterWindowsView::PresentSoftwareBitmap while binding_handler_ had been released and led to crash.

fix: https://github.com/flutter/flutter/issues/121189

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/shell/platform/windows/flutter_windows_view.cc
+++ b/shell/platform/windows/flutter_windows_view.cc
@@ -49,6 +49,9 @@ FlutterWindowsView::FlutterWindowsView(
 }
 
 FlutterWindowsView::~FlutterWindowsView() {
+  if (engine_ && engine_->running()) {
+    engine_->Stop();
+  }
   DestroyRenderSurface();
 }
 


loic-sharma:
(Apologies, I misclicked the close button)

I don't think this is the right fix. The Flutter engine should run in "headless" mode when there are no visible views. Furthermore, we're adding support for multiple views from a single engine to enable multi-window apps. Closing a window/view should not stop the engine if there are other views.

scutlight:
> The Flutter engine should run in "headless" mode when there are no visible views. 
Would you please explain a more bit about it?

If so, we may just call `engine_->SetView(nullptr);` before `DestroyRenderSurface();` inside `FlutterWindowsView::~FlutterWindowsView()`, to unregister itself from the engine as mentioned [here](https://github.com/flutter/flutter/issues/121189#issuecomment-1440606165). But I have to say that this will leave some error logs output from Skia and Angle, as the engine is trying to draw something and swap buffer without a correct gl frame buffer.

Also, we can just exchange their declared order of binding_handler_ and engine_ in FlutterWindowsView, to make sure that engine_ is first released while destructing FlutterWindowsView. This is a bit tricky and someone might append another member following engine_ in the future and again trigger some bugs like this one.

cbracken:
A few examples of where we want to handle headless engines are add-to-app scenarios where the user fires up and engine and then attaches a view so we're ready with the first frame as soon as the view is present. That scenario is a bit different in that it's a startup scenario - but in a world where multi-window apps are present we expect apps to be able to add and remove (multiple) windows over their lifetimes.

Since this isn't the approach we're likely to take, I'm going to close this PR for now. There are some notes on the original issue on how we might want to approach a fix for this.

loic-sharma:
> If so, we may just call `engine_->SetView(nullptr);` before `DestroyRenderSurface();` inside `FlutterWindowsView::~FlutterWindowsView()`, to unregister itself from the engine as mentioned [here](https://github.com/flutter/flutter/issues/121189#issuecomment-1440606165).

Yup that sounds good. Feel free to send us a patch! :)

> But I have to say that this will leave some error logs output from Skia and Angle, as the engine is trying to draw something and swap buffer without a correct gl frame buffer.

I agree this is weird, but the Flutter engine expects the Windows embedder to support this scenario. In today's single-view world, the framework always has a `PlatformDispatcher.instance.views[0]` regardless of whether there's a window or not. The framework should be able to draw, the engine delegates the draw to the embedder, and the embedder should know to no-op.

Note this will change in a multi-window world. There, the `PlatformDispatcher.instance.views` will not contain any views in headless mode. See: https://github.com/flutter/flutter/issues/120306

PR #39799, opened 2023-02-22
[web] Copy everything under the local canvaskit folder

mdebbar:
Instead of hardcoding the list of canvaskit files to copy, let's copy all js and wasm files that are built locally. This is necessary to pick up the chromium variant of canvaskit (see: https://github.com/flutter/engine/pull/39796).


eyebrowsoffire (approved):
LGTM

PR #39805, opened 2023-02-22
[Impeller] Wire up a playground that can run Dart

bdero:
This took a bit of tinkering, but once I figured out the fixture/dart kernel situation and got everything to happen in the right order, the minimal solution turned out to be pretty simple.


chinmaygarde (approved):
> once I figured out the fixture/dart kernel situation

Yeah, this one grew out of the initial setup just for unit-tests.

jonahwilliams:
OMG.  I would love to not have to write tests in C++ if they're intended to be end2end ish tests.

PR #37129, opened 2022-10-28
[Embedder API] Introduce new semantics update callback and migrate Windows

loic-sharma:
This introduces a new embedder API to notify the embedder of semantic updates, and, migrates Windows to this new embedder API. The Linux and macOS migrations will be done in follow-up changes.

Part of https://github.com/flutter/flutter/issues/114201

## Background

In the old embedder API, the engine breaks up a batch of semantic updates and sends each update individually to the embedder. This is problematic as there may be dependencies across these updates, so the ordering of these updates is crucial to prevent crashes (see https://github.com/flutter/flutter/issues/103808 and https://github.com/flutter/engine/pull/35792). In other words, there was an implicit contract between the embedder and the engine on the ordering of these updates.

The new embedder API just sends the entire batch of semantic updates in one go, allowing us to remove ordering concerns from the embedder API.

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [Flutter Style Guide] and the [C++, Objective-C, Java style guides].
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I added new tests to check the change I am making or feature I am adding, or Hixie said the PR is test-exempt. See [testing the engine] for instructions on writing and running engine tests.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I signed the [CLA].
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/flutter/wiki/Tree-hygiene#overview
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[C++, Objective-C, Java style guides]: https://github.com/flutter/engine/blob/main/CONTRIBUTING.md#style
[testing the engine]: https://github.com/flutter/flutter/wiki/Testing-the-engine
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/wiki/Chat

PR #39807, opened 2023-02-23
[Embedder API] Introduce new update semantics callback

loic-sharma:
This introduces a new update semantics embedder API that is identical to [the previous update semantics embedder API](https://github.com/flutter/engine/pull/37129), except that its ABI is stable as:

1. The callback uses array of pointers to structs instead of array of structs. This ensures that array indexing does not break if new members are added to the structs.
2. The types `FlutterSemanticsNode` and `FlutterSemanticsCustomAction` were duplicated as `FlutterSemanticsNode2` and `FlutterSemanticsCustomAction2`. This is necessary as `FlutterSemanticsUpdate` has an array of `FlutterSemanticsNode` and an array of `FlutterSemanticsCustomAction`. Adding new members to these structs would break the ABI compatibility of array indexing.

This change does not migrate embedders to this new embedder API, this will be done in subsequent changes: https://github.com/flutter/flutter/issues/121176.

Part of: https://github.com/flutter/flutter/issues/121176
See [go/flutter-semantics-abi](http://go/flutter-semantics-abi).

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/shell/platform/common/accessibility_bridge.cc
+++ b/shell/platform/common/accessibility_bridge.cc
@@ -37,10 +37,25 @@ AccessibilityBridge::~AccessibilityBridge() {
 }
 
 void AccessibilityBridge::AddFlutterSemanticsNodeUpdate(
+    const FlutterSemanticsNode2& node) {
+  pending_semantics_node_updates_[node.id] = FromFlutterSemanticsNode(node);
+}
+
+void AccessibilityBridge::AddFlutterSemanticsCustomActionUpdate(
+    const FlutterSemanticsCustomAction2& action) {
+  pending_semantics_custom_action_updates_[action.id] =
+      FromFlutterSemanticsCustomAction(action);
+}
+
+// TODO(loicsharma): Remove this as FlutterSemanticsNode is deprecated.
+// See: https://github.com/flutter/flutter/issues/121176
+void AccessibilityBridge::AddFlutterSemanticsNodeUpdate(
     const FlutterSemanticsNode& node) {
   pending_semantics_node_updates_[node.id] = FromFlutterSemanticsNode(node);
 }
 
+// TODO(loicsharma): Remove this as FlutterSemanticsNode is deprecated.
+// See: https://github.com/flutter/flutter/issues/121176
 void AccessibilityBridge::AddFlutterSemanticsCustomActionUpdate(
     const FlutterSemanticsCustomAction& action) {
   pending_semantics_custom_action_updates_[action.id] =
@@ -578,6 +593,74 @@ void AccessibilityBridge::SetTreeData(const SemanticsNode& node,
 
 AccessibilityBridge::SemanticsNode
 AccessibilityBridge::FromFlutterSemanticsNode(
+    const FlutterSemanticsNode2& flutter_node) {
+  SemanticsNode result;
+  result.id = flutter_node.id;
+  result.flags = flutter_node.flags;
+  result.actions = flutter_node.actions;
+  result.text_selection_base = flutter_node.text_selection_base;
+  result.text_selection_extent = flutter_node.text_selection_extent;
+  result.scroll_child_count = flutter_node.scroll_child_count;
+  result.scroll_index = flutter_node.scroll_index;
+  result.scroll_position = flutter_node.scroll_position;
+  result.scroll_extent_max = flutter_node.scroll_extent_max;
+  result.scroll_extent_min = flutter_node.scroll_extent_min;
+  result.elevation = flutter_node.elevation;
+  result.thickness = flutter_node.thickness;
+  if (flutter_node.label) {
+    result.label = std::string(flutter_node.label);
+  }
+  if (flutter_node.hint) {
+    result.hint = std::string(flutter_node.hint);
+  }
+  if (flutter_node.value) {
+    result.value = std::string(flutter_node.value);
+  }
+  if (flutter_node.increased_value) {
+    result.increased_value = std::string(flutter_node.increased_value);
+  }
+  if (flutter_node.decreased_value) {
+    result.decreased_value = std::string(flutter_node.decreased_value);
+  }
+  if (flutter_node.tooltip) {
+    result.tooltip = std::string(flutter_node.tooltip);
+  }
+  result.text_direction = flutter_node.text_direction;
+  result.rect = flutter_node.rect;
+  result.transform = flutter_node.transform;
+  if (flutter_node.child_count > 0) {
+    result.children_in_traversal_order = std::vector<int32_t>(
+        flutter_node.children_in_traversal_order,
+        flutter_node.children_in_traversal_order + flutter_node.child_count);
+  }
+  if (flutter_node.custom_accessibility_actions_count > 0) {
+    result.custom_accessibility_actions = std::vector<int32_t>(
+        flutter_node.custom_accessibility_actions,
+        flutter_node.custom_accessibility_actions +
+            flutter_node.custom_accessibility_actions_count);
+  }
+  return result;
+}
+
+AccessibilityBridge::SemanticsCustomAction
+AccessibilityBridge::FromFlutterSemanticsCustomAction(
+    const FlutterSemanticsCustomAction2& flutter_custom_action) {
+  SemanticsCustomAction result;
+  result.id = flutter_custom_action.id;
+  result.override_action = flutter_custom_action.override_action;
+  if (flutter_custom_action.label) {
+    result.label = std::string(flutter_custom_action.label);
+  }
+  if (flutter_custom_action.hint) {
+    result.hint = std::string(flutter_custom_action.hint);
+  }
+  return result;
+}
+
+// TODO(loicsharma): Remove this as FlutterSemanticsNode is deprecated.
+// See: https://github.com/flutter/flutter/issues/121176
+AccessibilityBridge::SemanticsNode
+AccessibilityBridge::FromFlutterSemanticsNode(
     const FlutterSemanticsNode& flutter_node) {
   SemanticsNode result;
   result.id = flutter_node.id;
@@ -627,6 +710,9 @@ AccessibilityBridge::FromFlutterSemanticsNode(
   return result;
 }
 
+// TODO(loicsharma): Remove this as FlutterSemanticsCustomAction is
+// deprecated.
+// See: https://github.com/flutter/flutter/issues/121176
 AccessibilityBridge::SemanticsCustomAction
 AccessibilityBridge::FromFlutterSemanticsCustomAction(
     const FlutterSemanticsCustomAction& flutter_custom_action) {
--- a/shell/platform/common/accessibility_bridge.h
+++ b/shell/platform/common/accessibility_bridge.h
@@ -54,6 +54,27 @@ class AccessibilityBridge
   ///             To flush the pending updates, call the CommitUpdates().
   ///
   /// @param[in]  node           A reference to the semantics node update.
+  void AddFlutterSemanticsNodeUpdate(const FlutterSemanticsNode2& node);
+
+  //------------------------------------------------------------------------------
+  /// @brief      Adds a custom semantics action update to the pending semantics
+  ///             update. Calling this method alone will NOT update the
+  ///             semantics tree. To flush the pending updates, call the
+  ///             CommitUpdates().
+  ///
+  /// @param[in]  action           A reference to the custom semantics action
+  ///                              update.
+  void AddFlutterSemanticsCustomActionUpdate(
+      const FlutterSemanticsCustomAction2& action);
+
+  //------------------------------------------------------------------------------
+  /// @brief      Adds a semantics node update to the pending semantics update.
+  ///             Calling this method alone will NOT update the semantics tree.
+  ///             To flush the pending updates, call the CommitUpdates().
+  ///
+  /// @param[in]  node           A reference to the semantics node update.
+  //  TODO(loicsharma): Remove this as FlutterSemanticsNode is deprecated.
+  //  See: https://github.com/flutter/flutter/issues/121176
   void AddFlutterSemanticsNodeUpdate(const FlutterSemanticsNode& node);
 
   //------------------------------------------------------------------------------
@@ -64,6 +85,9 @@ class AccessibilityBridge
   ///
   /// @param[in]  action           A reference to the custom semantics action
   ///                              update.
+  //  TODO(loicsharma): Remove this as FlutterSemanticsCustomAction is
+  //  deprecated.
+  //  See: https://github.com/flutter/flutter/issues/121176
   void AddFlutterSemanticsCustomActionUpdate(
       const FlutterSemanticsCustomAction& action);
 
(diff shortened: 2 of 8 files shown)


shell/platform/embedder/embedder.cc
```diff
@@ -1301,6 +1301,49 @@ FlutterSemanticsNode CreateEmbedderSemanticsNode(
...
+      node.customAccessibilityActions.size(),
+      node.customAccessibilityActions.data(),
+      node.platformViewId,
+      node.tooltip.c_str(),
+  };
+}
+
 // Translates engine semantic custom actions to embedder semantic custom
 // actions.
 FlutterSemanticsCustomAction CreateEmbedderSemanticsCustomAction(
```
loic-sharma, line 1349: Add comment not to add new members.

shell/platform/embedder/embedder.cc
```diff
@@ -1301,6 +1301,49 @@ FlutterSemanticsNode CreateEmbedderSemanticsNode(
   };
 }
```
loic-sharma, line 1302: Add comment not to add new members

shell/platform/embedder/embedder.cc
```diff
@@ -1346,6 +1402,55 @@ CreateNewEmbedderSemanticsUpdateCallback(
...
 
+// Create a callback to notify the embedder of semantic updates
+// using the new embedder callback 'update_semantics_callback'.
```
loic-sharma, line 1406:
```suggestion
// using the new embedder callback 'update_semantics_callback2'.
```

shell/platform/embedder/embedder.h
```diff
@@ -1048,6 +1048,8 @@ extern const int32_t kFlutterSemanticsNodeIdBatchEnd;
 /// (i.e., during PipelineOwner.flushSemantics), which happens after
 /// compositing. Updates are then pushed to embedders via the registered
 /// `FlutterUpdateSemanticsCallback`.
+/// TODO: THIS IS DEPRECATED. DO NOT ADD NEW MEMBERS. USE 2 INSTEAD.
+/// TODO: Unit test new members are not added.
```
loic-sharma, line 1052: TODO

shell/platform/embedder/embedder.h
```diff
@@ -1130,6 +1205,8 @@ extern const int32_t kFlutterSemanticsCustomActionIdBatchEnd;
 /// Action overrides are custom actions that the application developer requests
 /// to be used in place of the standard actions in the `FlutterSemanticsAction`
 /// enum.
+/// TODO: THIS IS DEPRECATED. DO NOT ADD NEW MEMBERS. USE 2 INSTEAD.
+/// TODO: Unit test new members are not added.
```
loic-sharma, line 1209: TODO

shell/platform/embedder/embedder.h
```diff
@@ -1938,6 +2056,17 @@ typedef struct {
   /// If this callback is provided, update_semantics_node_callback and
   /// update_semantics_custom_action_callback must not be provided.
   FlutterUpdateSemanticsCallback update_semantics_callback;
```
loic-sharma, line 2058: TODO: Update comment above

shell/platform/embedder/tests/embedder_a11y_unittests.cc
```diff
@@ -50,6 +97,214 @@ TEST_F(EmbedderA11yTest, A11yTreeIsConsistent) {
...
+      8 + // Struct padding assuming 64 bit processor
+      3 * sizeof(size_t) + 5 * sizeof(double) + 5 * sizeof(int32_t) +
+      3 * sizeof(const int32_t*) + 6 * sizeof(const char*) +
+      sizeof(FlutterRect) + sizeof(FlutterTextDirection) +
+      sizeof(FlutterSemanticsFlag) + sizeof(FlutterTransformation) +
+      sizeof(FlutterSemanticsAction) + sizeof(FlutterPlatformViewIdentifier);
+
+  size_t expected_action_size = sizeof(size_t) + sizeof(int32_t) +
+                                2 * sizeof(const char*) +
+                                sizeof(FlutterSemanticsAction);
```
loic-sharma, line 292: Is there a better way to "lock" these structs to ensure no new members are added? In particular, `FlutterSemanticsNode` is problematic due to architecture-specific struct padding.

cbracken: https://en.cppreference.com/w/cpp/types/offsetof in a googletest maybe?

cbracken: A quick and dirty (though a bit brittle) hacky approach could be to just grep for the current struct as-is. You could keep a canonical set of text files each with a locked struct, and have a grep check. I'm not proud of suggesting it, but it's not particularly more awful than some of the horrors in the licence script.

shell/platform/embedder/embedder.h
```diff
@@ -1118,9 +1121,82 @@ typedef struct {
   const char* tooltip;
 } FlutterSemanticsNode;
 
+/// A node that represents some semantic data.
+///
+/// The semantics tree is maintained during the semantics phase of the pipeline
+/// (i.e., during PipelineOwner.flushSemantics), which happens after
+/// compositing. Updates are then pushed to embedders via the registered
+/// `FlutterUpdateSemanticsCallback2`.
+typedef struct {
```
yaakovschectman, line 1130: What is the reason for defining this new struct to use instead of the existing one? Is this only to keep the old and new ABIs separated?

loic-sharma: This is necessary as `FlutterSemanticsUpdate` has an array of `FlutterSemanticsNode` and an array of `FlutterSemanticsCustomAction`. Adding new members to these structs would break the ABI compatibility of array indexing. In other words, `FlutterSemanticsNode` and `FlutterSemanticsCustomAction` must never change.

However, `FlutterSemanticsUpdate2` has array of pointers to `FlutterSemanticsNode2` and an array of pointers to `FlutterSemanticsCustomAction2`. It is safe to add new members to `FlutterSemanticsNode2` and `FlutterSemanticsCustomAction2`. For more information, please see the Background section of https://github.com/flutter/flutter/issues/121176 or see [go/flutter-semantics-abi](http://go/flutter-semantics-abi).

yaakovschectman (approved):
Makes sense

shell/platform/common/accessibility_bridge.h
```diff
@@ -61,6 +61,27 @@ class AccessibilityBridge
   ///             To flush the pending updates, call the CommitUpdates().
   ///
   /// @param[in]  node           A pointer to the semantics node update.
+  void AddFlutterSemanticsNodeUpdate(const FlutterSemanticsNode2* node);
```
cbracken, line 64: While this isn't public API, these should really take a const ref rather than a const pointer, since this isn't an out param and null isn't an acceptable input.

Can you also update the old ones for consistency?

shell/platform/embedder/embedder.h
```diff
@@ -1051,6 +1051,9 @@ extern const int32_t kFlutterSemanticsNodeIdBatchEnd;
 /// (i.e., during PipelineOwner.flushSemantics), which happens after
 /// compositing. Updates are then pushed to embedders via the registered
 /// `FlutterUpdateSemanticsCallback`.
+///
+/// @deprecated     Use `FlutterSemanticsNode2` instead. Adding members to this
+///                 struct breaks the ABI of `FlutterUpdateSemantics`.
```
cbracken, line 1056: The message here is addressing two different audiences: users of this API should use `FlutterSemanticsNode2` but authors of this API need to know not to add to it because it will break ABI compatibility.

Maybe something like:
```suggestion
/// @deprecated     Use `FlutterSemanticsNode2` instead. In order to preserve
///                 ABI compatibility for existing users, no new fields will be
///                 added to this struct. New fields will continue to be added
///                 to `FlutterSemanticsNode2`.
```

shell/platform/embedder/embedder.h
```diff
@@ -1118,9 +1121,82 @@ typedef struct {
   const char* tooltip;
 } FlutterSemanticsNode;
 
+/// A node that represents some semantic data.
```
cbracken, line 1124:
```suggestion
/// A node in the Flutter semantics tree.
```

Yeah, it's still mostly tautological, but it does point out that this is the same thing as a `SemanticsNode` in the framework, and it's a couple characters shorter.

shell/platform/embedder/embedder.h
```diff
@@ -1118,9 +1121,82 @@ typedef struct {
...
+///
+/// The semantics tree is maintained during the semantics phase of the pipeline
+/// (i.e., during PipelineOwner.flushSemantics), which happens after
+/// compositing. Updates are then pushed to embedders via the registered
+/// `FlutterUpdateSemanticsCallback2`.
```
cbracken, line 1129: I don't think we've referred directly to framework code from here before, but I can't think of many reasons not to:

```suggestion
/// `FlutterUpdateSemanticsCallback2`.
///
/// @see https://api.flutter.dev/flutter/semantics/SemanticsNode-class.html
```

If we ever change our domain name we'll need to fix this, but that's happened only once. We could/should probably eventually add some form of URL liveness checking for this.

shell/platform/embedder/embedder.h
```diff
@@ -1133,6 +1209,9 @@ extern const int32_t kFlutterSemanticsCustomActionIdBatchEnd;
 /// Action overrides are custom actions that the application developer requests
 /// to be used in place of the standard actions in the `FlutterSemanticsAction`
 /// enum.
+///
+/// @deprecated     Use `FlutterSemanticsCustomAction2` instead. Adding members
+///                 to this struct breaks the ABI of `FlutterUpdateSemantics`.
```
cbracken, line 1214: Similar comment to above:
```suggestion
/// @deprecated     Use `FlutterSemanticsCustomAction2` instead. In order to
///                 preserve ABI compatility for existing users, no new fields  
///                 will be added to this struct. New fields will continue to         
///                 be added to `FlutterSemanticsCustomAction2`.
```

shell/platform/embedder/embedder.h
```diff
@@ -1791,9 +1916,9 @@ typedef struct {
   /// The callback will be invoked on the thread on which the `FlutterEngineRun`
   /// call is made.
   ///
-  /// @deprecated    Prefer using `update_semantics_callback` instead. If this
-  ///                calback is provided, `update_semantics_callback` must not
-  ///                be provided.
+  /// @deprecated    Use `update_semantics_callback2` instead. If this
+  ///                callback is provided, `update_semantics_callback2` and
+  ///                `update_semantics_callback` must not be provided.
```
cbracken, line 1921: I'll admit it took me a couple reads to be sure I was parsing this right. What do you think of the following note for all three of these (with the notable exception that the latest callback isn't marked deprecated)?
```suggestion
  /// @deprecated    Use `update_semantics_callback2` instead. Only one of      
  ///                `update_semantics_node_callback`,                          
  ///                `update_semantics_callback`, and                           
  ///                `update_semantics_callback2` may be provided; the others
  ///                should be set to null. 
```

shell/platform/embedder/embedder.h
```diff
@@ -1161,6 +1267,21 @@ typedef struct {
   FlutterSemanticsCustomAction* custom_actions;
 } FlutterSemanticsUpdate;
 
+/// A batch of updates to semantics nodes and custom actions.
+typedef struct {
+  /// The size of the struct. Must be sizeof(FlutterSemanticsUpdate2).
+  size_t struct_size;
+  /// The number of semantics node updates.
+  size_t nodes_count;
```
cbracken, line 1275: We should probably leave these as-is to make transitioning from the old one easier, but the nitpicker in me really wants to say `node_count` and `custom_action_count` :)

cbracken (commented):
Overall looks good - mostly nitpicks.

shell/platform/common/accessibility_bridge.h
loic-sharma: +1 but this turned out to be pretty impactful (it affects all desktop embedders and unit tests). I'd like to keep this pull request focused on the embedder API changes, so I'll do the reference refactoring in a separate pull request and then rebase this one on top that.

cbracken: Awesome - even better.

loic-sharma: Here's the pull request for this refactoring: https://github.com/flutter/engine/pull/40167. I'll land that and then rebase this one :)

shell/platform/embedder/embedder.h
loic-sharma: Nice, this is definitely clearer!

shell/platform/embedder/embedder.h
loic-sharma: Updated!

shell/platform/embedder/embedder.h
loic-sharma: Fixed

shell/platform/embedder/embedder.h
loic-sharma: I updated `FlutterSemanticsUpdate2` but let `FlutterSemanticsUpdate` as-is.

shell/platform/common/accessibility_bridge.h
loic-sharma: Rebased!

shell/platform/embedder/embedder.h
loic-sharma: Fixed

shell/platform/embedder/embedder.h
loic-sharma: Updated!

shell/platform/common/accessibility_bridge.cc
```diff
@@ -36,11 +36,26 @@ AccessibilityBridge::~AccessibilityBridge() {
...
+  pending_semantics_node_updates_[node.id] = FromFlutterSemanticsNode(node);
+}
+
+void AccessibilityBridge::AddFlutterSemanticsCustomActionUpdate(
+    const FlutterSemanticsCustomAction2& action) {
+  pending_semantics_custom_action_updates_[action.id] =
+      FromFlutterSemanticsCustomAction(action);
+}
+
+// TODO(loicsharma): Remove this as FlutterSemanticsNode is deprecated.
```
chunhtai, line 50: I don't think we have other customer relying on this, what's preventing us from removing it?

chunhtai: oh the migration of embedding will be a separate pr?

chunhtai (approved):
LGTM

shell/platform/common/accessibility_bridge.cc
loic-sharma: Yup, Windows and macOS will be in subsequent pull requests. I will remove this once all affected embedders have been updated.

loic-sharma:
@cbracken and @chinmaygarde This is ready for review. Please take your time as I'd like to make sure there aren't any further embedder API mistakes :)

shell/platform/embedder/embedder.cc
```diff
@@ -1354,6 +1409,54 @@ CreateNewEmbedderSemanticsUpdateCallback(
...
+// Create a callback to notify the embedder of semantic updates
+// using the new embedder callback 'update_semantics_callback2'.
+flutter::PlatformViewEmbedder::UpdateSemanticsCallback
+CreateNewEmbedderSemanticsUpdateCallback2(
+    FlutterUpdateSemanticsCallback2 update_semantics_callback,
+    void* user_data) {
+  return [update_semantics_callback, user_data](
+             const flutter::SemanticsNodeUpdates& nodes,
+             const flutter::CustomAccessibilityActionUpdates& actions) {
+    std::vector<FlutterSemanticsNode2> embedder_nodes;
```
cbracken, line 1421:
```suggestion
    std::vector<FlutterSemanticsNode2> embedder_nodes;
    embedder_nodes.reserve(nodes.size());
```

shell/platform/embedder/embedder.cc
```diff
@@ -1354,6 +1409,54 @@ CreateNewEmbedderSemanticsUpdateCallback(
...
+    for (const auto& value : nodes) {
+      embedder_nodes.push_back(CreateEmbedderSemanticsNode2(value.second));
+    }
+
+    std::vector<FlutterSemanticsCustomAction2> embedder_custom_actions;
```
cbracken, line 1426:
```suggestion
    std::vector<FlutterSemanticsCustomAction2> embedder_custom_actions;
    embedder_custom_actions.reserve(actions.size());
```

cbracken (approved):
lgtm with a teensy nit.

shell/platform/embedder/embedder.cc
loic-sharma: Updated!

shell/platform/embedder/embedder.cc
loic-sharma: Updated!

PR #39810, opened 2023-02-23
[Impeller] Add PolygonMode to render layer

bdero:
Allows for good wireframe drawing. I've reached for line strips when using playgrounds to interactively debug geometry issues many times in the past. But for stroke paths, this is often insufficient for visualizing problems with joins, caps, and disconnected contours.

https://user-images.githubusercontent.com/919017/220804041-62079a6e-e8ca-40b6-a266-01b7f2201647.mov


chinmaygarde (approved):
Handy. I've hacked the engine itself to view this in the past. But this seems cleaner.

PR #39817, opened 2023-02-23
Make FlutterException constructor public

xegrox:
The default constructor access modifier in java is package-private, which restricts access outside of this package
Change the constructor to `public` so that others may construct their own `FlutterException`

20230223_20h48m21s_grim


stuartmorgan:
I'll follow up in the Pigeon PR, but I don't think this is the right avenue here; the documentation for this class is "Thrown to indicate that a Flutter method invocation failed on the Flutter side", and you are trying to repurpose it for the opposite.

PR #39824, opened 2023-02-23
Unregister the FlutterWindowsView on its destruction

yaakovschectman:
`FlutterWindowsView` owns both the `FlutterWindowsEngine` and `WindowBindingHandler` as unique pointers, so each is released upon its destruction. However, in the brief time between the release of the handler and the engine, the engine may attempt to call a method that expects the handler to exist.

Solve this by unsetting the engine's view from `FlutterWindowsView::~FlutterWindowsView`, which should complete before destructing its members. Include a unit test to verify that the view is unset by the time the handler is released.

https://github.com/flutter/flutter/issues/121189

## Pre-launch Checklist

- [ ] I signed the [CLA].

### DIFF
--- a/shell/platform/windows/flutter_windows_view.cc
+++ b/shell/platform/windows/flutter_windows_view.cc
@@ -50,6 +50,9 @@ FlutterWindowsView::FlutterWindowsView(
 
 FlutterWindowsView::~FlutterWindowsView() {
   DestroyRenderSurface();
+  if (engine_) {
+    engine_->SetView(nullptr);
+  }
 }
 
 void FlutterWindowsView::SetEngine(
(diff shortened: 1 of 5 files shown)


shell/platform/windows/flutter_windows_view.cc
```diff
@@ -50,6 +50,9 @@ FlutterWindowsView::FlutterWindowsView(
 
 FlutterWindowsView::~FlutterWindowsView() {
   DestroyRenderSurface();
```
loic-sharma, line 52: This is fine but I think it'd be preferable if the view was deregistered from the engine before its state is torn down. This way the engine does not point to a "bad" view.

yaakovschectman: What are you considering when the state is torn down? At the start of the destructor being called?

loic-sharma: I meant that I would deregister the view before destroying the render surface. So yup, deregistering the view at the start of the destructor would work

loic-sharma (approved):
Nice work!

scutlight (approved):
LGTM!

PR #39827, opened 2023-02-23
[web] Download CanvasKit chromium when possible

mdebbar:
On Chromium-based browsers, try downloading the Chromium variant of CanvasKit. If not available, fallback to the normal variant.

This PR also introduce a new runtime config to control the filename of the `canvaskit.js` file. This allows users to force the Chromium of variant of CanvasKit by setting `canvasKitJsFileName: "chromium/canvaskit.js"` or force the non-Chromium one by doing `canvasKitJsFileName: "canvaskit.js"` (could be useful in tests as well).

This is still disabled so it should have no effect yet. Once all the pieces are ready, we need to set `useClientICU` to the correct value based on the browser we are running on.

Part of https://github.com/flutter/flutter/issues/118801


lib/web_ui/lib/src/engine/canvaskit/renderer.dart
```diff
@@ -28,6 +28,13 @@ import 'shader.dart';
 import 'text.dart';
 import 'vertices.dart';
 
+/// Whether we can use client-provided ICU information instead of CanvasKit's
+/// built-in ICU.
+const bool useClientICU = false;
```
eyebrowsoffire, line 33: I think I would name this something a little different, like `browserSupportsCanvaskitChromium` or something like that. This bool will only contain the information about whether *can* use the lite version of canvaskit, but the user might override that.

lib/web_ui/lib/src/engine/configuration.dart
```diff
@@ -249,6 +276,7 @@ class JsFlutterConfiguration {}
 
 extension JsFlutterConfigurationExtension on JsFlutterConfiguration {
   external String? get canvasKitBaseUrl;
+  external String? get canvasKitJsFileName;
```
eyebrowsoffire, line 279: This I feel is a bit too generic and unopinionated. I think instead, we should have a `canvasKitVariant` that the user specifies `'full'` or `'chromium'` or something like that. Tying this to the file name of the canvas kit JS stuff feels a little too fine grained, IMO.

eyebrowsoffire (changes requested):

lib/web_ui/lib/src/engine/configuration.dart
mdebbar: I have mixed feelings about this. But I think I agree with you. The file structure inside `canvaskit/` is an implementation detail, and we should be able to change it in the future without breaking apps.

So it's gonna be similar to web renderer i.e. 'auto' (default), 'full', 'chromium'.

lib/web_ui/lib/src/engine/canvaskit/renderer.dart
mdebbar: Done

lib/web_ui/lib/src/engine/canvaskit/canvaskit_api.dart
```diff
@@ -2666,40 +2667,83 @@ void patchCanvasKitModule(DomHTMLScriptElement canvasKitScript) {
...
+    case CanvasKitVariant.auto:
+      return <String>[
+        if (browserSupportsCanvaskitChromium) _kChromiumCanvasKitJsFileName,
+        _kFullCanvasKitJsFileName,
+      ];
+    case CanvasKitVariant.full:
+      return <String>[_kFullCanvasKitJsFileName];
+    case CanvasKitVariant.chromium:
+      return <String>[_kChromiumCanvasKitJsFileName];
+  }
```
ditman, line 2685: I think this mode is allowing people to configure their apps in a way that is quite breaking to part of their potential users.

I agree that we should offer "auto" (try to optimize) and "full" (opt out of the optimization), but "chromium" mode sounds like trouble down the line.

(Unless we want to bring back the "Best Viewed With Internet Explorer" type of badges :P)

lib/web_ui/lib/src/engine/canvaskit/canvaskit_api.dart
```diff
@@ -2666,40 +2667,83 @@ void patchCanvasKitModule(DomHTMLScriptElement canvasKitScript) {
...
-    locateFile: allowInterop((String file, String unusedBase) =>
-        canvasKitWasmModuleUrl(canvasKitBuildUrl, file)),
+    locateFile: allowInterop(canvasKitWasmModuleUrl),
   ));
 }
 
-/// Downloads the CanvasKit JavaScript file at [canvasKitBase].
-Future<void> _downloadCanvasKitJs() {
-  final String canvasKitJavaScriptUrl = canvasKitJavaScriptBindingsUrl;
+Future<void> _downloadOneOf(Iterable<String> urls) async {
```
ditman, line 2705: Add a small documentation comment to this function please, something like: 

```
// Attempts to download from [urls] until one succeeds.
```

lib/web_ui/lib/src/engine/canvaskit/canvaskit_api.dart
```diff
@@ -2666,40 +2667,83 @@ void patchCanvasKitModule(DomHTMLScriptElement canvasKitScript) {
...
+  late final DomEventListener errorCallback;
 
-  final Completer<void> canvasKitLoadCompleter = Completer<void>();
-  late DomEventListener callback;
   void loadEventHandler(DomEvent _) {
-    canvasKitLoadCompleter.complete();
-    canvasKitScript.removeEventListener('load', callback);
+    canvasKitLoadCompleter.complete(true);
+    canvasKitScript.removeEventListener('load', loadCallback);
+    canvasKitScript.removeEventListener('error', errorCallback);
```
ditman, line 2734: Since you're creating (and discarding) the `canvasKitScript` within this function, I don't think you need to worry about removing the listeners.

lib/web_ui/lib/src/engine/configuration.dart
```diff
@@ -249,6 +278,7 @@ class JsFlutterConfiguration {}
 
 extension JsFlutterConfigurationExtension on JsFlutterConfiguration {
   external String? get canvasKitBaseUrl;
+  external String? get canvasKitVariant;
```
ditman, line 281: This is going to need some documentation.

Will this hit stable at the same time as https://github.com/flutter/flutter/issues/114708 ? I am going to have some time to add to the doc site.

ditman (approved):
Question about the "chromium" mode. Do we want it?

ditman:
Also there's a failing test that is probably passing `'null'` (string) as a config value, rather than `null`?

```
00:00 [32m+0[0m: [1m[90mloading test/ui/path_test.dart[0m[0m                                                                                                                                                               
[CHROME STDERR]:[0223/152338.970262:WARNING:sandbox_linux.cc(393)] InitializeSandbox() called with multiple threads in process gpu-process.
[CHROME STDERR]:
[CHROME STDERR]:DevTools listening on ws://127.0.0.1:12345/devtools/browser/76779f24-94a2-49d6-960f-74fb15daedae

00:00 [32m+0[0m: test/ui/path_test.dart: (setUpAll)[0m                                                                                                                                                           
00:00 [32m+0[0m[31m -1[0m: test/ui/path_test.dart: (setUpAll) [1m[31m[E][0m[0m                                                                                                                                                    
  Invalid argument (canvasKitVariant): Unknown CanvasKit variant: null
  org-dartlang-sdk:///lib/_internal/js_runtime/lib/js_helper.dart 1135:37   Object.wrapException
  ../../../../lib/src/engine/configuration.dart 211:9                       FlutterConfiguration.canvasKitVariant
  ../../../../lib/src/engine/canvaskit/canvaskit_api.dart 2675:25           Object._canvasKitJsFileNames
  ../../../../lib/src/engine/canvaskit/canvaskit_api.dart 2688:10           Object._canvasKitJsUrls
  ../../../../lib/src/engine/canvaskit/canvaskit_api.dart 2698:24           <fn>
  org-dartlang-sdk:///lib/_internal/js_runtime/lib/async_patch.dart 308:19  _wrapJsFunctionForAsync.closure.$protected
  org-dartlang-sdk:///lib/_internal/js_runtime/lib/async_patch.dart 333:23  _wrapJsFunctionForAsync.<fn>
  org-dartlang-sdk:///lib/_internal/js_runtime/lib/async_patch.dart 238:3   Object._asyncStartSync
  ../../../../lib/src/engine/canvaskit/canvaskit_api.dart 2698:41           Object.downloadCanvasKit
  ../../../../lib/src/engine/canvaskit/renderer.dart 76:25                  <fn>
```

lib/web_ui/lib/src/engine/canvaskit/canvaskit_api.dart
eyebrowsoffire: My two cents, I think we should allow it, just for clarity and symmetry. If people really shoot themselves in the foot with this and make their app unusable outside of Chromium browsers, that's on them. This might be useful in some sort of testing edge cases or something.

ditman: > If people really shoot themselves in the foot with this and make their app unusable outside of Chromium browsers, that's on them.

I don't know, it's possible I might be jaded by the discussions I have with people on the internet about CORS and other basic things :/

In any case, "chromium" mode smells like YAGNI. If we add it for testing or other edge cases, I'd call it: "dangerously-set-chromium-breaks-other-browsers" :P

mdebbar: Good point. The only use case I could think of is testing.

For testing, we may want to force the chromium variant. If we leave it at auto, and we forget to copy canvaskit_chromium, then the tests will silently fallback to the "full" canvaskit and we won't notice.

I'm happy to add some warnings in the dartdocs. We could also change the string to "unsafe_chromium" but I'm hesitant to do it because I don't want people to think that using the new lite canvaskit is unsafe.

lib/web_ui/lib/src/engine/canvaskit/canvaskit_api.dart
mdebbar: Existing code used to remove the listeners, so I did the same. But now that I think about it, maybe we should remove the entire script tag instead?

lib/web_ui/lib/src/engine/configuration.dart
mdebbar: I'm getting close to landing this, yes. Happy to work with you on the documentation.

mdebbar:
> Also there's a failing test that is probably passing `'null'` (string) as a config value, rather than `null`?

Nah. It's a real `null`. I added a print statement and this is what I'm seeing:
```
variant: null (runtimeType: Null)
```

For some reason `case null:` isn't catching it. I'll keep digging.

mdebbar:
It turns out, the value is `undefined` not `null`. So Dart fails to match it with `case null:`.

https://dartpad.dev/66017295266de29cb7bf82d0136c7bb7

lib/web_ui/lib/src/engine/canvaskit/canvaskit_api.dart
mdebbar: Done

lib/web_ui/lib/src/engine/configuration.dart
```diff
@@ -183,6 +186,33 @@ class FlutterConfiguration {
...
+  ///
+  /// * `full` - the full variant of CanvasKit that can be used in any browser.
+  ///
+  /// * `chromium` - the lite variant of CanvasKit that can be used in
+  /// Chromium-based browsers.
+  CanvasKitVariant get canvasKitVariant {
+    final String? variant = _configuration?.canvasKitVariant;
+    switch (variant) {
+      case 'auto':
+      case null:
```
ditman, line 204: `case null` needs to be handled outside the switch. According to @rakudrama you found a bug in the compiler!

lib/web_ui/lib/src/engine/canvaskit/canvaskit_api.dart
ditman: OK, I think that if it's only for testing we should be able to override this differently from the engine build, but let's add it and see what happens :P

ditman (approved):
Let's go chromium mode! (small change required in the `switch` though!)

lib/web_ui/lib/src/engine/configuration.dart
ditman: Ah, no, this one is already using Dart only types, so it's fine?

mdebbar: My first workaround was using `js_util` so I didn't like it much. Now I changed it and removed the `case null:`. I also added a test explicitly to check for this case.

ditman:
```
   info - lib/src/engine/configuration.dart:48:8 - Unused import: 'package:js/js_util.dart'. Try removing the import directive. - unused_import
```

mdebbar: Oops!

ditman:
🚀

mdebbar:
@eyebrowsoffire it looks like I need your approval too (to override your previous "Request Changes" review).

lib/web_ui/lib/src/engine/configuration.dart
```diff
@@ -183,6 +185,32 @@ class FlutterConfiguration {
...
+  /// * `auto` - the default value. The engine will automatically detect the
+  /// best variant to use based on the browser.
+  ///
+  /// * `full` - the full variant of CanvasKit that can be used in any browser.
+  ///
+  /// * `chromium` - the lite variant of CanvasKit that can be used in
+  /// Chromium-based browsers.
+  CanvasKitVariant get canvasKitVariant {
+    final String variant = _configuration?.canvasKitVariant ?? 'auto';
+    switch (variant) {
```
eyebrowsoffire, line 201: Not a blocker, but instead of doing your own switch here, you can do `CanvasKitVariant.values.byName(variant)` here.

eyebrowsoffire (approved):
Looks great!

lib/web_ui/lib/src/engine/configuration.dart
mdebbar: Done

ditman:
Let's goooo!

PR #39836, opened 2023-02-24
Add platform channel `System.exitApplication` and  `System.requestAppExit` support

gspencergoog:
## Description

This adds support for `System.exitApplication` and sending and handling the response from `System.requestAppExit` to the platform channel.

It also introduces a `FlutterApplication` class that is a subclass of `NSApplication`, and overrides the `terminate` message so that it can ignore it if the framework decides to cancel the termination, and to allow for async communication with the framework before termination.

If the application doesn't use a `FlutterApplication` as its `NSPrincipalClass` in `Info.plist` for the application, then it all works as before: the application will be unable to cancel the request. In debug mode it does log once per run, if the application is accessed, that some capabilities are disabled (like exitApplication). I'm not really sure if we should be doing the log message. It seems like a good prompt to remind people to migrate, and how, but I could easily be convinced that it's log spam.

Paired with https://github.com/flutter/flutter/pull/121378 to enable the framework to respond properly. Before that lands, if anyone switches to use `FlutterApplication` as their principal class, termination requests will go unanswered.

## Related Issues
 - https://github.com/flutter/flutter/issues/30735

## Tests
- Adds tests for the round trip from framework request for exit, to application termination request from the engine to framework, and then a response to the engine to actually terminate.


shell/platform/darwin/macos/framework/Source/FlutterApplication.mm
```diff
@@ -0,0 +1,86 @@
...
+  if (!warned_about_flutter_application && ![NSApp isKindOfClass:self]) {
+    NSLog(@"NSApp should be of type %s, not %s. "
+           "Some application lifecycle requests (e.g. ServicesBinding.exitApplication) "
+           "and notifications will be unavailable.\n"
+           "Modify the application's NSPrincipleClass to be %s"
+           "in the Info.plist to fix this.",
+          [[self className] UTF8String], [[NSApp className] UTF8String],
+          [[self className] UTF8String]);
+    warned_about_flutter_application = true;
+  }
```
cbracken, line 42:
```suggestion
  static dispatch_once_t onceToken = 0;
  dispatch_once(&onceToken, ^{
    if (![NSApp isKindOfClass:self]) {
      NSLog(@"NSApp should be of type %s, not %s. "
             "Some application lifecycle requests (e.g. ServicesBinding.exitApplication) "
             "and notifications will be unavailable.\n"
             "Modify the application's NSPrincipleClass to be %s"
             "in the Info.plist to fix this.",
            [[self className] UTF8String], [[NSApp className] UTF8String],
            [[self className] UTF8String]);
    }
  });
```

shell/platform/darwin/macos/framework/Source/FlutterApplication.mm
```diff
@@ -0,0 +1,86 @@
...
+// Initialize NSApplication using the custom subclass.  Check whether NSApp was
+// already initialized using another class, because that would break some
+// things. Warn about the mismatch only once, and only in debug builds.
++ (NSApplication*)sharedApplication {
+  NSApplication* app = [super sharedApplication];
+
+  // +sharedApplication initializes the global NSApp, so if we're delivering
+  // something other than a FlutterApplication, warn the developer once.
+#ifndef FLUTTER_RELEASE
+  if (!warned_about_flutter_application && ![NSApp isKindOfClass:self]) {
```
cbracken, line 33: I wonder if rather than testing for the use of the FlutterApplication class, we can somehow test they've overridden that particular method. NSApplication itself will include a default implementation though. :/

shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
```diff
@@ -152,6 +158,73 @@ - (void)handleMethodCall:(FlutterMethodCall*)call result:(FlutterResult)result;
...
+  // any good way to set the process exit code other than calling exit(), and
+  // that bypasses all of the native applicationShouldExit shutdown events,
+  // etc., which we don't want to skip.
+
+  FlutterAppExitType exitType;
+  if ([type isEqualTo:@"cancelable"]) {
+    exitType = kFlutterAppExitTypeCancelable;
+  } else {
+    exitType = kFlutterAppExitTypeRequired;
+  }
```
cbracken, line 186: I wonder if a ternary would look better here?

shell/platform/darwin/macos/framework/Source/FlutterEngine_Internal.h
```diff
@@ -127,3 +131,42 @@
...
+/**
+ * An enum for defining the different responses the framework can give to an
+ * application exit request from the engine.
+ *
+ * Must match the entries in the `AppExitResponse` enum in the Dart code.
+ */
+typedef enum {
+  kFlutterAppExitResponseCancel = 0,
+  kFlutterAppExitResponseExit = 1,
+} FlutterAppExitResponse;
```
cbracken, line 146: Prefer `NS_ENUM(NSInteger, FlutterAppExitResponse) { ... };`

shell/platform/darwin/macos/framework/Source/FlutterEngine_Internal.h
```diff
@@ -127,3 +131,42 @@
...
+/**
+ * An enum for defining the different request types allowed when requesting an
+ * application exit.
+ *
+ * Must match the entries in the `AppExitType` enum in the Dart code.
+ */
+typedef enum {
+  kFlutterAppExitTypeCancelable = 0,
+  kFlutterAppExitTypeRequired = 1,
+} FlutterAppExitType;
```
cbracken, line 157: Similarly here: `NS_ENUM(NSInteger, FlutterAppExitType) { ... };`

shell/platform/darwin/macos/framework/Source/FlutterEngine_Internal.h
```diff
@@ -127,3 +131,42 @@
                        withData:(fml::MallocMapping)data;
 
 @end
+
+#pragma mark -
```
cbracken, line 135:
```suggestion
#pragma mark - Enumerations
```

shell/platform/darwin/macos/framework/Source/FlutterEngine_Internal.h
```diff
@@ -127,3 +131,42 @@
...
+ * An enum for defining the different request types allowed when requesting an
+ * application exit.
+ *
+ * Must match the entries in the `AppExitType` enum in the Dart code.
+ */
+typedef enum {
+  kFlutterAppExitTypeCancelable = 0,
+  kFlutterAppExitTypeRequired = 1,
+} FlutterAppExitType;
+
```
cbracken, line 158:
```suggestion

#pragma mark - FlutterEngineTerminationHandler

```

shell/platform/darwin/macos/framework/Source/FlutterEngine_Internal.h
```diff
@@ -78,7 +82,7 @@
 - (void)removeViewController:(nonnull FlutterViewController*)viewController;
 
 /**
- * The `FlutterViewController` associated with the given view ID, if any.
+ * The |FlutterViewController associated with the given view ID, if any.
```
cbracken, line 85:
```suggestion
 * The |FlutterViewController| associated with the given view ID, if any.
```

shell/platform/darwin/macos/framework/Source/FlutterEngine_Internal.h
```diff
@@ -2,17 +2,21 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+#include <Foundation/Foundation.h>
```
cbracken, line 5: This shouldn't be necessary with the Cocoa import below.
```suggestion
```

shell/platform/darwin/macos/framework/Source/FlutterAppDelegate_internal.h
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * Allows the engine to add a handler for termination requests
+ */
+- (void)setTerminationRequestHandler:(FlutterEngineTerminationHandler*)handler;
+
+/**
+ * Allows the delegate to respond to an attempt to terminate the application.
+ */
+- (void)tryToTerminateApplication:(FlutterApplication*)application
+                         exitType:(FlutterAppExitType)type;
```
cbracken, line 22: `tryToFoo:` isn't very idiomatic Obj-C naming. I think there's one or two cases of `tryFoo:` but just `foo:` is more typical (sometimes with accompanying `shouldFoo` etc.).

shell/platform/darwin/macos/framework/Source/FlutterAppDelegate.mm
```diff
@@ -13,10 +17,16 @@ - (NSString*)applicationName;
 
 @end
 
-@implementation FlutterAppDelegate
+@implementation FlutterAppDelegate {
+  FlutterEngineTerminationHandler* _terminationHandler;
+}
```
cbracken, line 22: Since this is private, I wonder if we could just make it a property.

shell/platform/darwin/macos/framework/Source/FlutterAppDelegate_internal.h
gspencergoog: Okay, that was copied from the Chrome naming in their ObjectiveC code. I'll think of another name.

shell/platform/darwin/macos/framework/Source/FlutterAppDelegate_internal.h
```diff
@@ -0,0 +1,25 @@
...
+@interface FlutterAppDelegate ()
+/**
+ * Allows the engine to add a handler for termination requests
+ */
+- (void)setTerminationRequestHandler:(FlutterEngineTerminationHandler*)handler;
+
+/**
+ * Allows the delegate to respond to an attempt to terminate the application.
+ */
+- (void)tryToTerminateApplication:(FlutterApplication*)application
```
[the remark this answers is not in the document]
gspencergoog: Done

shell/platform/darwin/macos/framework/Source/FlutterApplication.mm
```diff
@@ -0,0 +1,86 @@
...
+// found in the LICENSE file.
+
+#import "flutter/shell/platform/darwin/macos/framework/Headers/FlutterApplication.h"
+#include "embedder.h"
+#include "shell/platform/darwin/macos/framework/Headers/FlutterAppDelegate.h"
+#include "shell/platform/darwin/macos/framework/Source/FlutterAppDelegate_internal.h"
+
+#ifndef FLUTTER_RELEASE
+namespace {
+static bool warned_about_flutter_application = false;
```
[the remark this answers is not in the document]
gspencergoog: Eliminated

gspencergoog:
Ok, PTAL: I've added a test that goes the round trip: handles a call to exit the application from the framework, asks the framework if that's OK (so that if other listeners are registered, they get asked), and then either exits or doesn't depending on the response.

The only thing I couldn't figure out how to test was the actually call to `[[FlutterApplication sharedApplication] terminate:nil]`, since that requires that there be a real application running (making a mock subclass would just bypass the thing I'm trying to test). There's only a couple of lines of code there.

I was able to cut down the necessary additions to `FlutterAppDelegate` to be just an internal property that is the termination handler.

The termination handler is an internal class, and so is the property on `FlutterAppDelegate` for now. I considered making them public, but I'd like to see a use case first.

lib/ui/platform_dispatcher.dart
```diff
@@ -1733,6 +1733,42 @@ enum AppLifecycleState {
...
+  /// Cancel the exit: do not exit the application.
+  cancel,
+}
+
+/// The type of application exit to perform when calling
+/// `ServicesBinding.exitApplication`.
+// TODO(gspencergoog): Insert references here to ServicesBinding.exitApplication
+// that has landed in the framework.
+enum AppExitType {
+  /// Requests that the application start an orderly exit, sending a request
```
loic-sharma, line 1754:
```suggestion
  /// Requests that the application starts an orderly exit, sending a request
```

lib/ui/platform_dispatcher.dart
```diff
@@ -1733,6 +1733,42 @@ enum AppLifecycleState {
...
+  /// back to the framework through the [WidgetsBinding], and if that responds
+  /// with [AppExitResponse.exit], then proceed with the same steps as a
+  /// [required] exit. if that responds with [AppExitResponse.cancel], then the
```
loic-sharma, line 1757:
```suggestion
  /// [required] exit. If that responds with [AppExitResponse.cancel], then the
```

shell/platform/darwin/macos/framework/Headers/FlutterApplication.h
```diff
@@ -0,0 +1,58 @@
...
+ * necessary from FlutterApplication.mm.
+ */
+@interface FlutterApplication : NSApplication
+
+/**
+ * FlutterApplication's implementation of |terminate| doesn't terminate the
+ * application: that is left up to the engine, which will call this function if
+ * it decides that termination request is granted, which will start the regular
+ * Cocoa flow for terminating the application, calling
+ * |applicationShouldTermnate|, etc.
```
loic-sharma, line 51:
```suggestion
 * |applicationShouldTerminate|, etc.
```

shell/platform/darwin/macos/framework/Headers/FlutterApplication.h
```diff
@@ -0,0 +1,58 @@
...
+ *
+ * @param(sender) The id of the object requesting the termination, or nil.
+ */
+- (void)terminateApplication:(id)sender;
```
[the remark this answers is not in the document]
loic-sharma: I'm fine with the current name, but perhaps this could align to `AppExitType.required`? Perhaps something like `requiredTerminate`?

shell/platform/darwin/macos/framework/Source/FlutterAppDelegate.mm
```diff
@@ -18,6 +24,13 @@ @implementation FlutterAppDelegate
 // TODO(stuartmorgan): Implement application lifecycle forwarding to plugins here, as is done
 // on iOS. Currently macOS plugins don't have access to lifecycle messages.
```
loic-sharma, line 25: Consider updating the TODO owner and linking to https://github.com/flutter/flutter/issues/30735

shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
```diff
@@ -3,13 +3,19 @@
 // found in the LICENSE file.
 
 #import "flutter/shell/platform/darwin/macos/framework/Headers/FlutterEngine.h"
+#include "FlutterCodecs.h"
```
cbracken, line 6: This should move down to the other flutter headers that aren't directly associated with `FlutterEngine` and prefix with the full include path.

shell/platform/darwin/macos/framework/Source/FlutterEngineTest.mm
```diff
@@ -2,8 +2,12 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+#include "OCMock/OCMArg.h"
+#include "OCMock/OCMock.h"
```
cbracken, line 6: These should move down below the directly-associated `FlutterEngine` headers but above the C++ headers in their own group.

shell/platform/darwin/macos/framework/Source/FlutterEngineTest.mm
```diff
@@ -2,8 +2,12 @@
...
 #import "flutter/shell/platform/darwin/macos/framework/Headers/FlutterEngine.h"
 #import "flutter/shell/platform/darwin/macos/framework/Source/FlutterEngine_Internal.h"
+#include "gtest/gtest.h"
+#include "shell/platform/darwin/macos/framework/Headers/FlutterApplication.h"
```
cbracken, line 10: Flutter headers that aren't directly associated with this test file (i.e. anything but FlutterEngine.h and FlutterEngine_Internal.h should move down into the other Flutter headers. 

Also for this line, Objective-C headers should use `#import`.

cbracken (commented):
A couple quick comments before I grab dinner. Will finish the rest of the review in a bit!

lib/ui/platform_dispatcher.dart
cbracken: I think `start` is correct here since it's the subjunctive 🤓. 

It would be correct to write "Requests the the presenter put on his pants" rather than "Requests that the presenter puts on his pants".

shell/platform/darwin/macos/framework/Headers/FlutterApplication.h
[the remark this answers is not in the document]
gspencergoog: Hmm.  Maybe?  The framework can always send a "required" termination to call it, but Objective-C code might want access.

There is an internal property on the engine that reveals the `terminationHandler`, which could be made public if we want to allow Objective-C code to send a required exit.  It would allow us to do other things when they try and exit if we need to in the future (not sure what those would be, but maybe notifying the framework so it can do work before we call the real terminate? Seems unreliable in the same way that the exiting state was, though.).

Still, add this to a protocol might be a good idea.

shell/platform/darwin/macos/framework/Source/FlutterAppDelegate.mm
gspencergoog: Okay, done.  This TODO will hopefully not last long, since my next task for the lifecycle implementation is to do exactly that.

lib/ui/platform_dispatcher.dart
```diff
@@ -1733,6 +1733,42 @@ enum AppLifecycleState {
...
+
+  /// A non-cancelable orderly exit request. The engine will shut down the
+  /// engine and call the native UI toolkit's exit API.
+  ///
+  /// If you need an even faster and more dangerous exit, then call `dart:io`'s
+  /// `exit()` directly, and even the native toolkit's exit API won't be called.
+  /// This is quite dangerous, though, since it's possible that the engine will
+  /// crash because it hasn't been properly shut down, causing the app to crash
+  /// on exit.
+  required,
```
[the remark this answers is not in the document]
gspencergoog: From the internets: "Required" is typically used to describe something that is needed in order to complete a task, while "mandatory" is used to describe something that is required by law or regulation (i.e. from a "mandate").

So, I agree, `required` it remains.

shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
gspencergoog: Damn you VSCode! I hate when it adds headers for me. I've turned it off in every place I can find, but clearly I've missed a spot.
It's not even needed (anymore: it probably was needed at one point in my explorations).

shell/platform/darwin/macos/framework/Headers/FlutterApplication.h
gspencergoog: Done.

shell/platform/darwin/macos/framework/Source/FlutterEngineTest.mm
gspencergoog: They don't appear to be needed.

shell/platform/darwin/macos/framework/Source/FlutterEngineTest.mm
gspencergoog: And it should have the right path. Done.

lib/ui/platform_dispatcher.dart
```diff
@@ -1733,6 +1733,42 @@ enum AppLifecycleState {
   detached,
 }
 
+/// The possible responses to a request to exit the application.
+///
+/// The request is typically responded to by a [WidgetsBindingObserver].
+// TODO(gspencergoog): Insert references here to AppLifecycleListener and to the
+// actual function called on WidgetsBindingObserver once those have landed in
+// the framework.
```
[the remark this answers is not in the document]
gspencergoog: Done. Added issue https://github.com/flutter/flutter/issues/121721

lib/ui/platform_dispatcher.dart
gspencergoog: Done.

shell/platform/darwin/macos/framework/Headers/FlutterApplication.h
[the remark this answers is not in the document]
cbracken: Ah I think it's sufficient to put it in the header, I just wonder if it's necessary. i.e. can we expose this method just to ourselves and not add any public API for this?

shell/platform/darwin/macos/framework/Source/FlutterAppDelegate_Internal.h
```diff
@@ -0,0 +1,22 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef FLUTTER_FLUTTERAPPDELEGATE_INTERNAL_H_
+#define FLUTTER_FLUTTERAPPDELEGATE_INTERNAL_H_
+
+#include "embedder.h"
```
cbracken, line 8:
```suggestion
#include "flutter/shell/platform/embedder/embedder.h"
```

shell/platform/darwin/macos/framework/Headers/FlutterApplication.h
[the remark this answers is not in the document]
gspencergoog: Oh, I see, so you're asking if we can make it just an interface declaration with no methods (which we have to have so that they can use it in their Info.plist), and move the method to a private header?

shell/platform/darwin/macos/framework/Source/FlutterAppDelegate.mm
```diff
@@ -4,6 +4,12 @@
 
 #import "flutter/shell/platform/darwin/macos/framework/Headers/FlutterAppDelegate.h"
 
+#import <AppKit/AppKit.h>
+
+#include "flutter/fml/logging.h"
+#import "flutter/shell/platform/darwin/macos/framework/Source/FlutterAppDelegate_Internal.h"
```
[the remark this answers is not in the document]
gspencergoog: Done.

shell/platform/darwin/macos/framework/Source/FlutterAppDelegate_Internal.h
gspencergoog: It's actually not needed at all. Removed.

shell/platform/darwin/macos/framework/Headers/FlutterApplication.h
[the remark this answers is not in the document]
gspencergoog: Okay, I've done that.

lib/ui/platform_dispatcher.dart
```diff
@@ -1733,6 +1733,43 @@ enum AppLifecycleState {
...
+}
+
+/// The type of application exit to perform when calling
+/// `ServicesBinding.exitApplication`.
+// TODO(gspencergoog): Insert doc references here to
+// ServicesBinding.exitApplication that has landed in the framework.
+// https://github.com/flutter/flutter/issues/121721
+enum AppExitType {
+  /// Requests that the application start an orderly exit, sending a request
+  /// back to the framework through the [WidgetsBinding], and if that responds
```
loic-sharma, line 1756: I don't feel strongly about this, but consider splitting this into separate sentences:

```suggestion
  /// back to the framework through the [WidgetsBinding]. If that responds
```

Feel free to keep as is if you prefer it.

shell/platform/darwin/macos/framework/Source/FlutterApplication.mm
```diff
@@ -0,0 +1,90 @@
...
+#include "flutter/shell/platform/embedder/embedder.h"
+#import "shell/platform/darwin/macos/framework/Headers/FlutterAppDelegate.h"
+#import "shell/platform/darwin/macos/framework/Source/FlutterAppDelegate_Internal.h"
+#import "shell/platform/darwin/macos/framework/Source/FlutterEngine_Internal.h"
+
+// An NSApplication subclass that implements overrides necessary for some
+// Flutter features, like application lifecycle handling.
+@implementation FlutterApplication
+
+// Initialize NSApplication using the custom subclass.  Check whether NSApp was
```
loic-sharma, line 17:
```suggestion
// Initialize NSApplication using the custom subclass. Check whether NSApp was
```

shell/platform/darwin/macos/framework/Source/FlutterApplication.mm
```diff
@@ -0,0 +1,90 @@
...
+}
+
+// |terminate| is the entry point for orderly "quit" operations in Cocoa. This
+// includes the application menu's Quit menu item and keyboard equivalent, the
+// application's dock icon menu's Quit menu item, "quit" (not "force quit") in
+// the Activity Monitor, and quits triggered by user logout and system restart
+// and shutdown.
+//
+// We override the normal |terminate| implementation. Our implementation, which
+// is specific to the asyncronous nature of Flutter, works by asking the
```
loic-sharma, line 50:
```suggestion
// is specific to the asynchronous nature of Flutter, works by asking the
```

shell/platform/darwin/macos/framework/Source/FlutterApplication.mm
```diff
@@ -0,0 +1,90 @@
...
+    // If there's no termination handler, then just terminate.
+    [super terminate:sender];
+  }
+  FlutterEngineTerminationHandler* terminationHandler =
+      [static_cast<FlutterAppDelegate*>([self delegate]) terminationHandler];
+  [terminationHandler requestApplicationTermination:sender
+                                           exitType:kFlutterAppExitTypeCancelable
+                                             result:nil];
+  // Return, don't exit. The application delegate is responsible for exiting on
+  // its own by calling |-terminateApplication|.
```
loic-sharma, line 80: (Ignore this comment if this is intentional)

```suggestion
  // its own by calling |terminateApplication|.
```

cbracken (approved):
lgtm!

PR #39859, opened 2023-02-24
[Impeller] Avoid printing inactionable warning from Metal compiler

dnfield:


jonahwilliams (approved):
LGTM

dnfield:
I am also realizing that a simpler way to deal with this is to just use `check_output` which won't print to stdout by default... Argh.

chinmaygarde (approved):
The test failures are complaining about the formatting of the python file. Other than that, I don't mind dropping all stdout/stderr too. We might end up making it harder to debug in case of failure but that is just a theoretical concern.

dnfield:
It'll still print stderr/stdout on failure, just not print anything on success

dnfield:
I really thought I ran the format script on this but maybe not. Either way I've simplified the change a bit so that it will still print out if the command fails, but won't print out warnings/stdout unless the process has a non-zero exit code.

PR #39863, opened 2023-02-24
delete DL ops no longer needed after removing SkCanvasRecorder

flar:
This is pretty simple "obsolete code" fallout from removing DisplayListCanvasRecorder. Some of the DisplayList ops were only ever captured on a theoretical basis from an SkCanvas object. Since we no longer capture from SkCanvas, we can delete the methods and objects that aren't directly recordable from our own DisplayList APIs. These include:

- DLBuilder::setAttributesFromPaint(SkPaint)
- DL::setBlender (we only use the BlendMode enum, not custom blender objects)
- DL::drawPicture (no more SkPictures in the ui.Canvas/flow layers work flow)
- DL::drawSkVertices (no longer need to capture opaque SkVertices objects)
- DL::drawLattice (only ever appeared as a general form of drawImageNine)

The 1000 lines are almost all straight deletions, and maybe some comment updates to no longer discuss objects we never see. Lots of Impeller::UNIMPLEMENTED removals...

### DIFF
--- a/display_list/display_list.h
+++ b/display_list/display_list.h
@@ -38,7 +38,7 @@
 //                       Any class implementing Dispatcher can inherit from
 //                       these utility classes to simplify its creation
 //
-// The Flutter DisplayList mechanism can be used in place of the Skia
+// The Flutter DisplayList mechanism is used in a similar manner to the Skia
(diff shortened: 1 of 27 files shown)


jonahwilliams (approved):
LGTM

PR #39865, opened 2023-02-24
Impeller drawAtlas performance improvements

jason-simmons:
* Remove construction of a temporary vector in TRect::MakePointBounds
* Avoid repeated int-to-float conversion of texture_size in AtlasTextureContents::Render
* Cache the AtlasContents bounding box


impeller/entity/contents/atlas_contents.cc
```diff
@@ -65,14 +67,17 @@ std::optional<Rect> AtlasContents::GetCoverage(const Entity& entity) const {
 }
 
 Rect AtlasContents::ComputeBoundingBox() const {
```
jonahwilliams, line 69: I suspect that we already know the exact bounding box in some part of the display list, but I took a shortcut here for simplicity.

This change LGTM, but fyi @flar in case we could plumb something through the dispatcher API if it exists

jonahwilliams (approved):
LGTM

impeller/entity/contents/atlas_contents.cc
flar: It would be odd to pass along a bounds to a draw call, but perhaps we could make an Atlas object that computed its bounds on construction and was passed through the DlCanvas/DL ops vector?

flar: The object could also vet that the array sizes all matched their common expectations...

flar: See https://github.com/flutter/flutter/issues/121512

jonahwilliams: Yeah that change sounds reasonable to me. If we are computing the bounds of these or other objects like paths for display list bounds though I think we should be passing them through somehow, or allowing impeller to query the bounds in the dispatcher somehow. Its wasteful if we end up doing nearly the same work twice.

flar: First, despite the bad naming, the dispatcher is the Impeller object which is receiving the data and so wouldn't know the answer. The object providing the info is the DisplayList and it is stateless (the Dispatch internal implementation has the state, but there is no "object" to represent its current state of dispatch as that is all inside the method call), so it would have no easy way to know which operation you are querying about. There are some other reasons we might want to provide "additional state" with the Dispatch() process, but there is no architectural mechanism to do so at this time.

The DisplayList doing the dispatching doesn't normally store any info that isn't provided in the method call. The API of the data going in and going out is identical for symmetry. Though we might change that in the future, that would require some re-plumbing. DisplayLists do store the bounds of all ops *IF* you create one with an RTree, but again, that is only used for culling within the implementation of the Dispatch() method call and there is no place to provide the info to the receiver without a lot of rework.

The easiest place to put the bounds would be in an object like we do with Path (via SkPath, but that feature will likely continue on with a DlPath replacement). I think, for many reasons, a DlAtlas object should be created and providing the bounds across the DL recording and playback process would be a nice side benefit. It would not only implicitly carry the local bounds of the operation along to the receiver, but it would also mean that the data consistency only has to be vetted once when the data is received from Dart rather than every step in the process that receives a bunch of raw arrays.

PR #39922, opened 2023-02-27
[Windows] Ignore case optionally in `AXPlatformNodeTextRangeProviderWin::FindText`

yaakovschectman:
When `ignore_case` is `true`, convert needle and haystack strings to lowercase before performing search,

https://github.com/flutter/flutter/issues/117013

## Pre-launch Checklist

- [ ] I signed the [CLA].


third_party/accessibility/ax/platform/ax_platform_node_textrangeprovider_win.cc
```diff
@@ -433,14 +434,20 @@ HRESULT AXPlatformNodeTextRangeProviderWin::FindAttributeRange(
   return S_OK;
 }
 
-static bool StringSearch(const std::u16string& search_string,
-                         const std::u16string& find_in,
+static bool StringSearch(std::u16string& search_string,
+                         std::u16string& find_in,
```
yaakovschectman, line 438: Please let me know if you can think of a better way to allow us to optionally copy the needle and haystack strings and then continue with the rest of the method. I attempted using `string_view`, but the problem with it is that if we create new copies of the strings within the if statement to convert to lowercase and assign to the views, the underlying memory is freed when the if statement completes.

cbracken: Rather than us trying to implement a case-insensitive compare, I wonder if we shouldn't just delegate to underlying Win32/macOS APIs. The idea of "case" is really finnicky -- e.g. Chinese, Japanese, Korean don't even have a concept of case, and even in French there have historically been differences between France and Canada's handling of accented characters (though IIRC France now matches Canada's handling), Turkish dotted vs non-dotted 'i' etc.

Does Chromium implement case comparison itself?

If we are going to do it ourselves, we should pass the string by value rather than mutate an incoming string.

cbracken: On Windows, I see [lstrcmpiW](https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-lstrcmpiw) which calls out to [CompareStringEx](https://learn.microsoft.com/en-us/windows/win32/api/stringapiset/nf-stringapiset-comparestringex), which looks like it does what we'd want.

cbracken: The notes on `lstrcmpiW` suggest some of the horrors intrinsic in implementing this sort of thing on our own, e.g. should ignore case be interpreted as ignore kana type in Japan? According to that function's implementation, apparently the answer is yes. e.g. か and カ are both the phonetic "ka" sound expressed using two different phonetic alphabets.

yaakovschectman: These functions compare strings, and `StringSearch` needs to search for a substring. Chromium appears to have an entire `i18n` library which provides a `StringSearch` method (see https://github.com/chromium/chromium/blob/main/base/i18n/string_search.h#L44 )

cbracken: Looking at their implementation, it looks like they're building on top of ICU and doing things like accent handling etc. As I say, I'd be tempted to avoid trying to recreate ICU/Win32 APIs here - between just languages I'm familiar with (French and Japanese) there are already difficulties, but I imagine a huge ton more in languages that I have no idea about.

cbracken: fwiw their search implementation does look very standalone and we could probably simply adopt it as an option.

third_party/accessibility/ax/platform/ax_platform_node_textrangeprovider_win_unittest.cc
```diff
@@ -5109,19 +5108,17 @@ TEST_F(AXPlatformNodeTextRangeProviderTest, TestITextRangeProviderFindText) {
...
 
   // Test searching for leaf content from ancestor.
   GetTextRangeProviderFromTextNode(range, root_node);
   EXPECT_UIA_FIND_TEXT(range, L"some text", false, owner);
-  // EXPECT_UIA_FIND_TEXT(range, L"SoMe TeXt", true, owner);
+  EXPECT_UIA_FIND_TEXT(range, L"SoMe TeXt", true, owner);
   EXPECT_UIA_FIND_TEXT(range, L"more text", false, owner);
-  // EXPECT_UIA_FIND_TEXT(range, L"MoRe TeXt", true, owner);
+  EXPECT_UIA_FIND_TEXT(range, L"MoRe TeXt", true, owner);
   EXPECT_UIA_FIND_TEXT(range, L"more", false, owner);
```
cbracken, line 5122: If we adopt Chromium's ICU-based StringSearch implementation, we should add further tests like checking that `eleve` is found in the string `élève`.

third_party/accessibility/ax/platform/ax_platform_node_textrangeprovider_win.cc
yaakovschectman: What are you looking at as being standalone? It appears that `StringSearch` creates and calls to `FixedPatternStringSearch`, which makes use of at least `UStringSearch` and `UCollator` from ICU

cbracken: Correct, I was looking at string_search.h,cc which depend only on ICU (which we already use) and a few FML-like utilities (which should be easy to replace).

yaakovschectman:
I expect the current test to fail, I just want to check before messing with the BUILD files more.

yaakovschectman:
@cbracken Now that we are using fuller ICU data on desktop, can you give this PR a look?

cbracken (approved):
Looks great - this was a painful one! Thanks for fixing!

PR #39926, opened 2023-02-27
Actually print out if xcrun metal fails

dnfield:
My previous commit was exiting with an unhelpful error that did not contain the failure output. This prints the failure output more nicely.

### DIFF
--- a/impeller/tools/build_metal_library.py
+++ b/impeller/tools/build_metal_library.py
@@ -120,7 +120,6 @@ def main():
   return 0
 
 
-
 if __name__ == '__main__':
   if sys.platform != 'darwin':
     raise Exception('This script only runs on Mac')
--- a/impeller/tools/build_metal_library.py
+++ b/impeller/tools/build_metal_library.py
@@ -113,9 +113,9 @@ def main():
 
   try:
     subprocess.check_output(command, stderr=subprocess.STDOUT, text=True)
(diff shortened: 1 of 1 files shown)


jonahwilliams (approved):
RSLGTM

impeller/tools/build_metal_library.py
```diff
@@ -111,10 +111,17 @@ def main():
 
   command += args.source
 
-  subprocess.check_output(command, stderr=subprocess.STDOUT)
+  try:
+    subprocess.check_output(command, stderr=subprocess.STDOUT, text=True)
```
zanderso, line 115: The presubmits will tell, but hopefully CI has a recent enough python for `text=True`.

zanderso (approved):

impeller/tools/build_metal_library.py
dnfield: I'm _pretty_ sure we do. I was going to use subprocess.run here but I'm not sure we have a recent enough for that. If it fails I'll change this.

dnfield:
```
impeller/tools/build_metal_library.py:116:2: C0103: Variable name "e" doesn't conform to '[a-z_][a-z0-9_]{2,30}$' pattern (invalid-name)
```

For real?

PR #39930, opened 2023-02-27
Revert "[lint + clang tidy] add lint check for python file changes"

jmagman:
Reverts flutter/engine#39736

The actual fix was related to glob behavior in https://github.com/flutter/cocoon/pull/2506.

We don't need `*.py` to run in both `Linux Host clang-tidy` and `Linux Android clang-tidy`, so remove the former.

### DIFF
--- a/.ci.yaml
+++ b/.ci.yaml
@@ -229,7 +229,6 @@ targets:
       - "**.cc"
       - "**.fbs"
       - "**.frag"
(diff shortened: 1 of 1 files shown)


godofredoc (approved):

XilaiZhang (approved):
images (1)
thanks for identifying and landing the fix so fast!

PR #39988, opened 2023-03-01
iOS embedder api migration umbrella PR

cyanglaz:
This is a temporary PR to see the diffs for the whole migration. This PR is used for reference during the migration and not intended to land.

The real migration will happen in smaller pathces.


Hixie:
I'm going to close this since it was not intended to land and it's been a few months now. Feel free to reopen if you still need this!

PR #40004, opened 2023-03-02
[canvaskit] read pixels back in Picture.toImage

yjbanov:
This undoes the optimization made in https://github.com/flutter/engine/pull/38573, and adds a test-case that would have failed when that optimization was made. The issue is that CanvasKit loses track of which resources belong to which context that leads to WebGL errors followed by UI corruption.

Fixes https://github.com/flutter/flutter/issues/121758

### DIFF
--- a/lib/web_ui/lib/src/engine/canvaskit/picture.dart
+++ b/lib/web_ui/lib/src/engine/canvaskit/picture.dart
@@ -100,28 +100,7 @@ class CkPicture extends ManagedSkiaObject<SkPicture> implements ui.Picture {
   }
 
   @override
-  ui.Image toImageSync(int width, int height) {
-    SurfaceFactory.instance.baseSurface.ensureSurface();
-    if (SurfaceFactory.instance.baseSurface.usingSoftwareBackend) {
-      return toImageSyncSoftware(width, height);
-    }
-    return toImageSyncGPU(width, height);
-  }
-
-  ui.Image toImageSyncGPU(int width, int height) {
-    assert(debugCheckNotDisposed('Cannot convert picture to image.'));
-
-    final CkSurface ckSurface = SurfaceFactory.instance.baseSurface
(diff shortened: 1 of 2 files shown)


jonahwilliams (commented):
Nit: this says its about toByteData, but the changes are in toImage.

Is there a way to defer the usage of MakeRasterDirectSurface until we're calling toByteData? ITs more or less expected that toByteData can be slow, but toImageSync is useful for composite effects using fragment shaders - and running those in software is not going to be feasible.

yjbanov:
> Nit: this says its about toByteData, but the changes are in toImage.
> 
> Is there a way to defer the usage of MakeRasterDirectSurface until we're calling toByteData? ITs more or less expected that toByteData can be slow, but toImageSync is useful for composite effects using fragment shaders - and running those in software is not going to be feasible.

Couldn't find a way to defer it. I'm going to start a thread with the Skia team to figure this out. In the meantime, I'm reverting https://github.com/flutter/engine/pull/38573 and adding a new unit-test that catches the issue. Hopefully, next time we attempt it, it won't regress.

yjbanov:
For posterity, I tried the following solutions that didn't work:

- Use `CanvasKit.MakeRasterDirectSurface` - this fixes https://github.com/flutter/flutter/issues/121758 but it regresses [texture-backed images](https://github.com/flutter/engine/blob/8885149cf97a9a97bcc54aa4c58a654a190411de/lib/web_ui/test/canvaskit/image_golden_test.dart#L745).
- I tried getting the `VideoFrame` pixels so `SkImage`s can be instantiated from in-memory pixels - that fixed both problems, but it required that I manually convert YUV encoded images to RGBA. Due to https://github.com/w3c/webcodecs/issues/351 there's no way to tell if images need to be reoriented during the RGBA conversion, so some of them came out wrong. This method is also likely to slow down image decoding (a more common use-case than `Picture.toImage`), so it was only good for experimentation.

hterkelsen (approved):
LGTM

PR #40007, opened 2023-03-02
Do not end the frame in the raster cache if ScopedFrame::Raster returns kResubmit

jason-simmons:
kResubmit (along with kSkipAndRetry) indicate that ScopedFrame::Raster did not actually paint the frame.

Fixes https://github.com/flutter/flutter/issues/121678


iskakaushik (approved):
LGTM! Thanks for the fix @jason-simmons

PR #40009, opened 2023-03-02
Don't include other files outside the flutter_web_sdk.

eyebrowsoffire:
This should fix the current engine -> framework roll.


goderbauer (approved):
LGTM

PR #40012, opened 2023-03-02
Remove PlatformConfiguration and ViewConfiguration

goderbauer:
Fixes https://github.com/flutter/flutter/issues/121742.

To be submitted after https://github.com/flutter/flutter/pull/121751 (framework tests will fail before that is submitted).


gspencergoog (approved):

goderbauer:
@ditman Could you take a look at the web_ui changes, when you have a chance? Thanks.

lib/web_ui/lib/platform_dispatcher.dart
```diff
@@ -96,7 +95,7 @@ abstract class PlatformDispatcher {
 
   Locale get locale;
 
-  List<Locale> get locales => configuration.locales;
+  List<Locale> get locales;
```
ditman, line 98: Thanks for doing this, the indirection here with the `configuration` getter that was overridden in the "engine" platform dispatcher was very confusing!

lib/web_ui/lib/src/engine/platform_dispatcher.dart
```diff
@@ -1320,3 +1318,104 @@ double findBrowserTextScaleFactor() {
   final num fontSize = parseFontSize(domDocument.documentElement!) ?? _defaultRootFontSize;
   return fontSize / _defaultRootFontSize;
 }
+
+class ViewConfiguration {
```
ditman, line 1322: I've been thinking of doing `class WebViewConfiguration extends ViewConfiguration` to be able to store some additional web-only view configuration (for example: the embedding strategy that we want to use for a given view).

Do you think it's OK to extend the ViewConfiguration like that, or should I put my additional view configuration in a separate data structure within the web engine?

lib/web_ui/lib/src/engine/semantics/semantics.dart
```diff
@@ -1579,7 +1579,7 @@ class EngineSemanticsOwner {
     final EngineAccessibilityFeatures original =
         EnginePlatformDispatcher.instance.configuration.accessibilityFeatures
         as EngineAccessibilityFeatures;
-    final ui.PlatformConfiguration newConfiguration =
+    final PlatformConfiguration newConfiguration =
```
ditman, line 1582: 💯

lib/web_ui/lib/src/engine/window.dart
```diff
@@ -202,15 +202,32 @@ class EngineFlutterWindow extends ui.SingletonFlutterWindow {
...
     final EnginePlatformDispatcher engineDispatcher =
         platformDispatcher as EnginePlatformDispatcher;
     assert(engineDispatcher.windowConfigurations.containsKey(viewId));
     return engineDispatcher.windowConfigurations[viewId] ??
-        const ui.ViewConfiguration();
+        const ViewConfiguration();
   }
 
+  @override
+  ui.Rect get physicalGeometry => _viewConfiguration.geometry;
```
ditman, line 214: I suspect some of these should come from the DimensionsProvider. Where can I read more about this new batch of view attributes?

lib/web_ui/lib/window.dart
```diff
@@ -6,17 +6,16 @@ part of ui;
 
 abstract class FlutterView {
```
ditman, line 7: Again, thanks for making this fully abstract.

ditman (approved):
The changes look good to me. Especially removing the indirection between the base abstract class, and the implementation via the "configuration" object. That was something that always made me confused when I looked at the View class hierarchy on the web!

lib/web_ui/lib/src/engine/platform_dispatcher.dart
goderbauer: Extending ViewConfiguration should be perfectly fine. I believe in the current setup you could even put the extra data right into the ViewConfiguration class. At this point it is essentially a private class of the web engine that only shares a name with the (also private) class in the non-web engine.

lib/web_ui/lib/src/engine/window.dart
goderbauer: They used to be implemented by the base FlutterView class in `lib/web_ui/lib/window.dart`. Since that one is now fully abstract, I moved the exact same implementation into the concrete class here. There should be no difference in where the value is coming from compared to what it was before. I am actually not familiar with the `DimensionsProvider`.

goderbauer: Oh, and these are documented on FlutterView, but they have been around for some time: https://master-api.flutter.dev/flutter/dart-ui/FlutterView-class.html

lib/ui/platform_dispatcher.dart
```diff
@@ -1247,9 +1242,8 @@ class PlatformDispatcher {
 /// Configuration of the platform.
 ///
 /// Immutable class (but can't use @immutable in dart:ui)
-class PlatformConfiguration {
-  /// Const constructor for [PlatformConfiguration].
-  const PlatformConfiguration({
+class _PlatformConfiguration {
```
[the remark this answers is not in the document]
goderbauer: I missed this comment just before I slapped the autosubmit label on. Sorry! You're right, though, we should be able to remove the private classes altogether now and store the values directly on the object. The only complication may be the window singleton, which currently makes use of the fact that two FlutterView instances can share the same ViewConfiguration. That could be resolved, though.

I may look at that clean-up in a follow-up.

lib/web_ui/lib/src/engine/window.dart
ditman: Totally missed that these came from the default implementation, I guess web is not relying in any of these (for now :P)

lib/web_ui/lib/src/engine/platform_dispatcher.dart
ditman: This is the perfect place for the EmbeddingStrategy object that I have for each View so it knows how to stamp itself into the DOM. Thanks!

PR #40025, opened 2023-03-02
[ios] add setting flag to enable embedder api.

cyanglaz:
First patch for iOS Embedder api migration.
The iOS Embedder API implementation is going to be hidden behind this flag during the migration. 
This PR introduced:
1. A command line flag `enable-embedder-api`, which can also be used for Android when we want to migrate it to embedder API.
2. An info.plist flag `FLTEnableIOSEmbedderAPI`, which is used on iOS.
Both flag acts the same on iOS and either one of them being true should run iOS embedder with the embedder API.
This flag defaults to false.

First step of https://github.com/flutter/flutter/issues/112232

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/common/settings.h
+++ b/common/settings.h
@@ -316,6 +316,11 @@ struct Settings {
   /// If it is not 0 or 1, it must be one of 2, 4, 8, or 16. However, if the
   /// GPU does not support the requested sampling value, MSAA will be disabled.
   uint8_t msaa_samples = 0;
+
+  /// Enable embedder api on the embedder.
+  ///
+  /// This is ignored on the platforms that use the embedder_api by default.
+  bool enable_embedder_api = false;
 };
 
 }  // namespace flutter
--- a/common/settings.h
+++ b/common/settings.h
@@ -319,7 +319,7 @@ struct Settings {
 
   /// Enable embedder api on the embedder.
   ///
-  /// This is ignored on the platforms that use the embedder_api by default.
+  /// This is currently only used by iOS.
   bool enable_embedder_api = false;
 };
 
--- a/shell/common/switches.h
+++ b/shell/common/switches.h
@@ -273,6 +273,9 @@ DEF_SWITCH(
     "Setting this value to 0 or 1 disables MSAA. If it is not 0 or 1, it must "
     "be one of 2, 4, 8, or 16. However, if the GPU does not support the "
     "requested sampling value, MSAA will be disabled.")
+DEF_SWITCH(EnableEmbedderAPI,
+           "enable-embedder-api",
+           "Enable the embedder api. Defaults to false. iOS only.")
 DEF_SWITCHES_END
 
 void PrintUsage(const std::string& executable_name);
--- a/shell/platform/darwin/ios/framework/Source/FlutterEngine.mm
+++ b/shell/platform/darwin/ios/framework/Source/FlutterEngine.mm
@@ -98,6 +98,12 @@ static constexpr int kNumProfilerSamplesPerSec = 5;
 @property(nonatomic, readwrite, copy) NSString* isolateId;
 @property(nonatomic, copy) NSString* initialRoute;
 @property(nonatomic, retain) id<NSObject> flutterViewControllerWillDeallocObserver;
+
+#pragma mark - Embedder API properties
+
+// Function pointers for interacting with the embedder.h API.
+@property(nonatomic, assign) BOOL enableEmbedderAPI;
+@property(nonatomic) FlutterEngineProcTable& embedderAPI;
 @end
 
 @implementation FlutterEngine {
@@ -184,6 +190,13 @@ static constexpr int kNumProfilerSamplesPerSec = 5;
     _dartProject.reset([[FlutterDartProject alloc] init]);
   } else {
     _dartProject.reset([project retain]);
+  }
+
+  _enableEmbedderAPI = _dartProject.get().settings.enable_embedder_api;
+  if (_enableEmbedderAPI) {
+    NSLog(@"============== iOS: enable_embedder_api is on ==============");
+    _embedderAPI.struct_size = sizeof(FlutterEngineProcTable);
+    FlutterEngineGetProcAddresses(&_embedderAPI);
   }
 
   if (!EnableTracingIfNecessary([_dartProject.get() settings])) {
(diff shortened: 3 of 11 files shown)


shell/common/switches.h
```diff
@@ -273,6 +273,12 @@ DEF_SWITCH(
     "Setting this value to 0 or 1 disables MSAA. If it is not 0 or 1, it must "
     "be one of 2, 4, 8, or 16. However, if the GPU does not support the "
     "requested sampling value, MSAA will be disabled.")
+DEF_SWITCH(EnableIOSEmbedderAPI,
+           "enable-embedder-api",
+           "Enable the embedder api. Defaults to false. Ignored if the "
+           "platform enables the"
```
jmagman, line 279:
```suggestion
           "platform enables the "
```

shell/platform/darwin/ios/framework/Source/FlutterEngineTest.mm
```diff
@@ -291,4 +291,26 @@ - (void)testThreadPrioritySetCorrectly {
   method_setImplementation(method, originalSetThreadPriority);
 }
 
+- (void)testCanEnableDisableEmbedderAPIThroughInfoPlist {
+  {
+    // Not enable embedder API by default
```
jmagman, line 296:
```suggestion
    // embedder API not enabled by default
```

shell/common/switches.h
```diff
@@ -273,6 +273,12 @@ DEF_SWITCH(
     "Setting this value to 0 or 1 disables MSAA. If it is not 0 or 1, it must "
     "be one of 2, 4, 8, or 16. However, if the GPU does not support the "
     "requested sampling value, MSAA will be disabled.")
+DEF_SWITCH(EnableIOSEmbedderAPI,
+           "enable-embedder-api",
+           "Enable the embedder api. Defaults to false. Ignored if the "
```
jmagman, line 278: Alternatively you could say "iOS only" for now, up to you though.

jmagman (commented):
Minor nits.  Looks like the `FlutterEngineTest` are crashing though.

shell/common/switches.h
cyanglaz: My first version was actually "iOS only" but wanted to keep it future proof for android. I will revert back to "iOS only", it's less confusing.

shell/platform/darwin/ios/framework/Source/FlutterEngineTest.mm
```diff
@@ -28,7 +28,7 @@ - (void)tearDown {
 }
 
 - (void)testCreate {
-  id project = OCMClassMock([FlutterDartProject class]);
+  FlutterDartProject* project = [[FlutterDartProject alloc] init];
```
cyanglaz, line 31: @jmagman The test failure was due to the mock project doesn't have the setting ivar mocked. For these tests, I either removed the project completely or used a real project object. There was no reason to mock the project at the first place.

jmagman (approved):

hellohuanlin:
> Both flag acts the same on iOS and either one of them being true should run iOS embedder with the embedder API.

Can you clarify why do we need both flags and when to use which one?

shell/platform/darwin/ios/framework/Source/FlutterEngine.mm
```diff
@@ -98,6 +98,12 @@ @interface FlutterEngine () <FlutterIndirectScribbleDelegate,
 @property(nonatomic, readwrite, copy) NSString* isolateId;
 @property(nonatomic, copy) NSString* initialRoute;
 @property(nonatomic, retain) id<NSObject> flutterViewControllerWillDeallocObserver;
+
+#pragma mark - Embedder API properties
+
+// Function pointers for interacting with the embedder.h API.
+@property(nonatomic, assign) BOOL enableEmbedderAPI;
```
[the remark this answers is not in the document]
hellohuanlin: probably also wanna remove `assign` if do `readonly`

jmagman:
> > Both flag acts the same on iOS and either one of them being true should run iOS embedder with the embedder API.
> 
> Can you clarify why do we need both flags and when to use which one?

The tool flag would be for a single `flutter run` invocation, and it gets passed on as a launch argument to the app, read by the engine.  The Info.plist bool makes it persistent for the app across runs, so you can open it in Xcode and have it be passed into the engine without actually setting a launch argument in the scheme.
We do the same pattern for other experimental features.

shell/platform/darwin/ios/framework/Source/FlutterEngine.mm
```diff
@@ -1382,6 +1395,16 @@ - (FlutterDartProject*)project {
   return _dartProject.get();
 }
 
+#pragma mark - Embedder API
+
+- (BOOL)enableEmbedderAPI {
+  return _enableEmbedderAPI;
+}
```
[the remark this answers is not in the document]
cyanglaz: They should. The getters were from the old implementation. I forgot to remove this and the below getters. 
Will remove now.

shell/platform/darwin/ios/framework/Source/FlutterEngine.mm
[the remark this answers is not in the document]
jmagman: I think you still want to keep `assign` (as opposed to `retain` or `copy`) for the primitive? I don't think the readwrite/readonly impacts that.

common/settings.h
```diff
@@ -316,6 +316,11 @@ struct Settings {
   /// If it is not 0 or 1, it must be one of 2, 4, 8, or 16. However, if the
   /// GPU does not support the requested sampling value, MSAA will be disabled.
   uint8_t msaa_samples = 0;
+
+  /// Enable embedder api on the embedder.
+  ///
+  /// This is ignored on the platforms that use the embedder_api by default.
+  bool enable_embedder_api = false;
```
hellohuanlin, line 323: Maybe more self-descriptive naming like `enable_embedder_api_for_ios` (or `enable_embedder_api_for_mobile` if wanna re-use for android)

cyanglaz:
> Can you clarify why do we need both flags and when to use which one?

The info.plist is for running from XCode, the commandline flag is for running from terminal without the need to edit the info.plist.

shell/platform/darwin/ios/framework/Source/FlutterEngine.mm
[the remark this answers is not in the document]
hellohuanlin: iirc `assign/retain/copy` etc only affect how the setters are synthesized. `readonly` won't synthesize the setter at all in the first place.

common/settings.h
cyanglaz: I tried to keep it generic so it could be reused by android.
I also wanted to keep it the same name as the command line flag. And I would like the command line flag to be short for easier typing when testing ;p

shell/platform/darwin/ios/framework/Source/FlutterEngine.mm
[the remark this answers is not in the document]
cyanglaz: I looked at it again. It is in the .mm file. So I will remove the readonly and keep the assign.

PR #40027, opened 2023-03-02
[web] Remember which variant of CanvasKit is being used

mdebbar:
This PR splits the decision making into two variables:

1. `browserSupportsCanvasKitChromium`: This only takes effect in `auto` mode. If it's true, then we try to download the Chromium variant first, and fallback to the full variant. _**(This is currently still hardcoded to false so users don't get the Chromium variant just yet)**_

2. `canvasKitVariant`: After CanvasKit is successfully downloaded, this variable contains the variant that was downloaded. This allows us to control what CanvasKit APIs to use depending on which CanvasKit variant we are using.

Required by https://github.com/flutter/engine/pull/39984

### DIFF
--- a/lib/web_ui/lib/src/engine/browser_detection.dart
+++ b/lib/web_ui/lib/src/engine/browser_detection.dart
@@ -266,3 +266,8 @@ int _detectWebGLVersion() {
   }
   return -1;
 }
+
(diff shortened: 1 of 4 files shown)


eyebrowsoffire (approved):
LGTM!

PR #40031, opened 2023-03-02
Added wide-gamut color support for `ui.Image.toByteData` and `ui.Image.colorSpace`

gaaclarke:
fixes https://github.com/flutter/flutter/issues/121729

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/lib/ui/painting.dart
+++ b/lib/ui/painting.dart
@@ -1568,10 +1568,12 @@ class Paint {
   }
 }
 
-/// The color space in which all of an [Image]'s colors exist.
+/// The color space describes the colors that are available to an [Image].
 ///
-/// This value can help you decide which [ImageByteFormat] to use with
-/// [Image.toByteData] and is the result of [Image.colorSpace].
+/// This value can help decide which [ImageByteFormat] to use with
+/// [Image.toByteData], where images that are in the [extendedSRGB] color space
+/// should use something like [ImageByteFormat.rawExtendedRgba128] so that
+/// colors aren't lost. This is also the result of [Image.colorSpace].
 enum ColorSpace {
   /// The sRGB color space, the defined standard color space for the web.
   sRGB,
--- a/lib/ui/painting.dart
+++ b/lib/ui/painting.dart
@@ -1762,7 +1762,7 @@ class Image {
       case 1:
         return ColorSpace.extendedSRGB;
       default:
-        throw StateError('Unrecognized color space: $colorSpaceValue');
+        throw Exception('Unrecognized color space: $colorSpaceValue');
     }
   }
 
--- a/lib/ui/painting.dart
+++ b/lib/ui/painting.dart
@@ -1574,8 +1574,12 @@ class Paint {
 /// [Image.toByteData], where images that are in the [extendedSRGB] color space
 /// should use something like [ImageByteFormat.rawExtendedRgba128] so that
 /// colors aren't lost. This is also the result of [Image.colorSpace].
+///
+/// See also: https://en.wikipedia.org/wiki/Color_space
 enum ColorSpace {
   /// The sRGB color space, the defined standard color space for the web.
+  ///
+  /// See also: https://en.wikipedia.org/wiki/SRGB
   sRGB,
   /// A color space that is backwards compatible with sRGB but can represent
   /// colors outside of that gamut with values outside of [0..1]. In order to
@@ -1610,7 +1614,7 @@ enum ImageByteFormat {
   /// Raw extended range RGBA format.
   ///
   /// Unencoded bytes, in RGBA row-primary form with straight alpha, 32 bit
-  /// float (IEEE 754) per channel.
+  /// float (IEEE 754 binary32) per channel.
   ///
   /// Example usage:
   ///
@@ -1774,8 +1778,11 @@ class Image {
   /// example, loading a PNG that is in the Display P3 color space will result
   /// in a [ColorSpace.extendedSRGB] image.
   ///
-  /// On certain platforms/rendering backends, wide gamut images will still
-  /// report [ColorSpace.sRGB] if rendering wide gamut colors isn't supported.
+  /// On rendering backends that don't support wide gamut colors (anything but
+  /// iOS impeller), wide gamut images will still report [ColorSpace.sRGB] if
+  /// rendering wide gamut colors isn't supported.
+  // Note: The docstring will become outdated as new platforms support wide
+  // gamut color, please keep it up to date.
   ColorSpace get colorSpace {
     final int colorSpaceValue = _image.colorSpace;
     switch (colorSpaceValue) {
--- a/lib/ui/painting.dart
+++ b/lib/ui/painting.dart
@@ -1573,7 +1573,9 @@ enum ColorSpace {
   /// The sRGB color gamut.
   sRGB,
   /// A color space that is backwards compatible with sRGB but can represent
-  /// colors outside of that gamut with values outside of [0..1];
+  /// colors outside of that gamut with values outside of [0..1]. In order to
+  /// see the extended values a [ImageByteFormat] like
+  /// [ImageByteFormat.rawExtendedRgba128] must be used.
   extendedSRGB,
 }
 
--- a/lib/ui/painting.dart
+++ b/lib/ui/painting.dart
@@ -1568,12 +1568,12 @@ class Paint {
   }
 }
 
-/// The color space that an [Image] uses.
+/// The color space in which all of an [Image]'s colors exist.
 ///
 /// This value can help you decide which [ImageByteFormat] to use with
-/// [Image.toByteData];
+/// [Image.toByteData] and is the result of [Image.colorSpace].
 enum ColorSpace {
-  /// The sRGB color gamut, the defined standard color gamut for the web.
+  /// The sRGB color space, the defined standard color space for the web.
   sRGB,
   /// A color space that is backwards compatible with sRGB but can represent
   /// colors outside of that gamut with values outside of [0..1]. In order to
@@ -1766,7 +1766,12 @@ class Image {
     return _image.toByteData(format: format);
   }
 
-  /// The color space that used by the [Image]'s colors.
+  /// The color space that is used by the [Image]'s colors.
+  ///
+  /// This value is a consequence of how the [Image] has been created.  For
+  /// example, loading a PNG that is in the Display P3 color space will result
+  /// in a [ColorSpace.extendedSRGB] image.
+  ///
   /// On certain platforms/rendering backends, wide gamut images will still
   /// report [ColorSpace.sRGB] if rendering wide gamut colors isn't supported.
   ColorSpace get colorSpace {
--- a/lib/ui/painting.dart
+++ b/lib/ui/painting.dart
@@ -1568,6 +1568,15 @@ class Paint {
   }
 }
 
+/// The color space that an [Image] uses.
+enum ColorSpace {
+  /// The sRGB color gamut.
+  sRGB,
+  /// A color space that is backwards compatible with sRGB but can represent
+  /// colors outside of that gamut with values outside of [0..1];
+  extendedSRGB,
+}
+
 /// The format in which image bytes should be returned when using
 /// [Image.toByteData].
 // We do not expect to add more encoding formats to the ImageByteFormat enum,
@@ -1590,6 +1599,11 @@ enum ImageByteFormat {
   /// Unencoded bytes, in the image's existing format. For example, a grayscale
   /// image may use a single 8-bit channel for each pixel.
   rawUnmodified,
+
+  /// Raw extended range RGBA format.
+  ///
+  /// Unencoded bytes, in RGBA row-primary form with straight alpha, 32 bit float per channel.
+  rawExtendedRgba128,
 
   /// PNG format.
   ///
@@ -1735,6 +1749,19 @@ class Image {
   Future<ByteData?> toByteData({ImageByteFormat format = ImageByteFormat.rawRgba}) {
     assert(!_disposed && !_image._disposed);
     return _image.toByteData(format: format);
+  }
+
+  /// The color space that used by the [Image]'s colors.
+  ColorSpace get colorSpace {
+    final int colorSpaceValue = _image.colorSpace;
+    switch (colorSpaceValue) {
+      case 0:
+        return ColorSpace.sRGB;
+      case 1:
+        return ColorSpace.extendedSRGB;
+      default:
+        throw StateError('Unrecognized color space: $colorSpaceValue');
+    }
   }
 
   /// If asserts are enabled, returns the [StackTrace]s of each open handle from
@@ -1902,6 +1929,9 @@ class _Image extends NativeFieldWrapperClass1 {
   external void _dispose();
 
   final Set<Image> _handles = <Image>{};
+
+  @Native<Int32 Function(Pointer<Void>)>(symbol: 'Image::colorSpace')
+  external int get colorSpace;
 
   @override
   String toString() => '[$width\u00D7$height]';
--- a/lib/ui/painting.dart
+++ b/lib/ui/painting.dart
@@ -1571,13 +1571,18 @@ class Paint {
 /// The color space describes the colors that are available to an [Image].
 ///
 /// This value can help decide which [ImageByteFormat] to use with
-/// [Image.toByteData], where images that are in the [extendedSRGB] color space
+/// [Image.toByteData]. Images that are in the [extendedSRGB] color space
 /// should use something like [ImageByteFormat.rawExtendedRgba128] so that
-/// colors aren't lost. This is also the result of [Image.colorSpace].
+/// colors outside of the sRGB gamut aren't lost.
+///
+/// This is also the result of [Image.colorSpace].
 ///
 /// See also: https://en.wikipedia.org/wiki/Color_space
 enum ColorSpace {
-  /// The sRGB color space, the defined standard color space for the web.
+  /// The sRGB color space.
+  ///
+  /// You may know this as the standard color space for the web or the color
+  /// space of non-wide-gamut Flutter apps.
   ///
   /// See also: https://en.wikipedia.org/wiki/SRGB
   sRGB,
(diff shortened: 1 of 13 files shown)


lib/ui/painting.dart
```diff
@@ -1568,6 +1568,15 @@ class Paint {
...
 
+/// The color space that an [Image] uses.
+enum ColorSpace {
+  /// The sRGB color gamut.
+  sRGB,
+  /// A color space that is backwards compatible with sRGB but can represent
+  /// colors outside of that gamut with values outside of [0..1];
+  extendedSRGB,
+}
+
```
jonahwilliams, line 1579: I think this enum and values probably deserve a lot more explanation. I think @Hixie would probably be interesting in taking a pass over this.

lib/ui/painting.dart
```diff
@@ -1568,6 +1568,15 @@ class Paint {
   }
 }
 
+/// The color space that an [Image] uses.
+enum ColorSpace {
+  /// The sRGB color gamut.
+  sRGB,
+  /// A color space that is backwards compatible with sRGB but can represent
+  /// colors outside of that gamut with values outside of [0..1];
```
jonahwilliams, line 1576: I thought that for "regular" images we gave colors channel values in the range of 0-255. Does the extended srb use 0-1?

lib/ui/painting.dart
```diff
@@ -1737,6 +1751,19 @@ class Image {
...
+  /// The color space that used by the [Image]'s colors.
+  ColorSpace get colorSpace {
+    final int colorSpaceValue = _image.colorSpace;
+    switch (colorSpaceValue) {
+      case 0:
+        return ColorSpace.sRGB;
+      case 1:
+        return ColorSpace.extendedSRGB;
+      default:
+        throw StateError('Unrecognized color space: $colorSpaceValue');
```
jonahwilliams, line 1763: I think since this value can only come from the engine, it would only be a programmer error on our side if it surfaced, rather than a StateError which implies the user used this code incorrectly.

Perhaps just assert?

lib/ui/painting.dart
```diff
@@ -1568,6 +1568,15 @@ class Paint {
   }
 }
 
+/// The color space that an [Image] uses.
+enum ColorSpace {
```
jonahwilliams, line 1572: Consider documenting the support for different values across our various backends

lib/ui/painting.dart
gaaclarke: FYI I'm finishing up a breaking change notice too that I planned on sharing with him for his review.

lib/web_ui/lib/painting.dart
```diff
@@ -351,6 +351,8 @@ abstract class Image {
 
   List<StackTrace>? debugGetOpenHandleStackTraces() => null;
 
+  ColorSpace get colorSpace => ColorSpace.sRGB;
```
yjbanov, line 354: Looks like all concrete subclasses are implementing this getter. I think this means that the getter in the base class can be abstract.

yjbanov (commented):
Is this an impeller-only feature? Does Skia support this? I vaguely recall that it did. If so, we should add a proper CanvasKit implementation for this.

lib/ui/painting.dart
gaaclarke: Yep, you have to use `toByteData(ImageByteFormat.rawExtendedRgba128)` to get normalized color components.  I can add a note here linking that.

gaaclarke: (added a note about ImageByteFormat.rawExtendedRgba128)

lib/ui/painting.dart
gaaclarke: Yea good call.  I moved it to an `Exception`.  An assert could get ripped out, which we don't want here.

lib/ui/painting.dart
gaaclarke: I don't think we want to do that because regardless of the backend the colorSpace property gives valid results, so it's inconsequential what the backend is and discussing it here could confuse and get stale easily.

lib/web_ui/lib/painting.dart
gaaclarke: The reason all of the other places had to implement it themselves is because they are using `implements` instead of `extends`.  I don't know if there was anyone that was using `extends`, but if they were, I wouldn't want them to have to implement this property.

gaaclarke:
> Is this an impeller-only feature? Does Skia support this? I vaguely recall that it did. If so, we should add a proper CanvasKit implementation for this.

Yep, it's on impeller + iOS currently (and behind a switch which I hope to remove).  The skia backend could support it but there are no plans to invest in it.  I'd love to have web support too.  I don't know what that would entail but webkit can already render wide gamut images for what it's worth.

lib/ui/painting.dart
gaaclarke: I added a note to `ui.Image.colorSpace` that wide gamut images may still return sRGB on backends that don't support wide gamut colors.  I think that addresses your concerns here and won't ever get stale.

lib/ui/painting.dart
gaaclarke: Actually I think this should be an Error based on Bob's distinction: 
```
1. Exception: stuff you catch: runtime errors.
2. Error: stuff you don't catch: system and programmatic errors.
```
source: https://groups.google.com/a/dartlang.org/g/misc/c/lx9CXiV3o30/m/s5l_PwpHUGAJ

This isn't something you should catch, but isn't a user error either.  Error seems more apropos.

jonahwilliams: StateError specifically though: "Should be used when this particular object is currently in a state which doesn't support the requested operation, but other similar objects might, or the object itself can later change its state to one which supports the operation."

Which doesn't really make sense to me because if this happens its almost certainly because the engine itself has a bug. Which is why perhaps we should assert and then return a default?

gaaclarke: There isn't a default value though.  We'd have to add a `ImageByteFormat.unknown` then make everyone handle that even though it's presence would be indicative of an error.  The documentation for `StateError` points to `UnsupportedError` "This [Error](https://api.flutter.dev/flutter/dart-core/Error-class.html) is thrown when an instance cannot implement one of the methods in its signature."  I think that sounds right.

jonahwilliams: but again, this is _only_ going to happen if we make a mistake in the engine implementation. End users shouldn't even be aware of it which is why I think it should assert.

The closest error type would be something like InternalError but thats just a java thing.

gaaclarke: Right, asserts are another type of Error though, they just throw an `AssertionError`.  The downside is that they are often removed though so we'd need the `ImageByteFormat.unknown` which essentially makes end-users have to *catch* the engine error, which is what we don't want.  I'm happy to discuss it over a quick VC if you are still on the fence about `UnsupportedError`.

lib/ui/painting.dart
Hixie: In general, if someone is reading the documentation, it's because they had a question. What could that question be?

Maybe here the question is "what is a color space", or "how do i change the color space that an image is using", or "how do i pick which color space to specify". Might be worth answering all of these here.

lib/ui/painting.dart
```diff
@@ -1568,6 +1568,17 @@ class Paint {
...
+  /// The sRGB color gamut.
```
Hixie, line 1573: Similarly here, if someone is looking at the documentation for `ColorSpace.sRGB`, I assume they already know that it means "the sRGB color gamut", so what's the question that led them to looking at the documentation? Maybe it's "what is sRGB", maybe it's "where do i specify this value"? Maybe it's "what range of colors are available for this"?

lib/ui/painting.dart
```diff
@@ -1568,6 +1568,17 @@ class Paint {
...
+  sRGB,
+  /// A color space that is backwards compatible with sRGB but can represent
+  /// colors outside of that gamut with values outside of [0..1]. In order to
+  /// see the extended values a [ImageByteFormat] like
```
Hixie, line 1577: `a [Image` -> `an [Image`

Hixie, line 1577: Mention where the [ImageByteFormat] can be used, since here we're looking at the enum docs and they don't link to anything.

lib/ui/painting.dart
```diff
@@ -1591,6 +1602,11 @@ enum ImageByteFormat {
   /// image may use a single 8-bit channel for each pixel.
   rawUnmodified,
 
+  /// Raw extended range RGBA format.
+  ///
+  /// Unencoded bytes, in RGBA row-primary form with straight alpha, 32 bit float per channel.
```
Hixie, line 1607: does "32 bit float" mean IEEE 754  `binary32`?

lib/ui/painting.dart
```diff
@@ -1737,6 +1753,21 @@ class Image {
     return _image.toByteData(format: format);
   }
 
+  /// The color space that used by the [Image]'s colors. Note that on certain
```
Hixie, line 1756: "note that" violates our style guide.

https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#avoid-empty-prose

Hixie, line 1756: separate the first line (brief description) from the details

https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#dartdoc-specific-requirements

lib/ui/painting.dart
```diff
@@ -1737,6 +1753,21 @@ class Image {
...
+  /// platforms/rendering backends, wide gamut images will still report
+  /// [ColorSpace.sRGB] if rendering wide gamut colors isn't supported.
```
Hixie, line 1758: which platforms/rendering backends? Specificity is the soul of narrative.

Hixie: (in general, when writing documentation, think to yourself "what questions does what i've written raise in the reader's mind?" and then answer those questions as well.)

lib/ui/painting.dart
Hixie: might be worth mentioning the Dart classes you can use to decode this (I assume `Float32List` would work?).

Maybe even include some sample code!

lib/ui/painting.dart
gaaclarke: done

lib/ui/painting.dart
gaaclarke: I'm not sure what else to add, I said it's the standard colorspace of the web so that might help people understand.  It's kind of a complicated topic to insert in a docstring.  They'd be better off googling it beyond that.

lib/ui/painting.dart
gaaclarke: Done, added sample code and specified the float32 format

lib/ui/painting.dart
gaaclarke: removed

gaaclarke: done

lib/ui/painting.dart
gaaclarke: done, added link to `toByteData` to ColorSpace

lib/ui/painting.dart
gaaclarke: I'm worried that if I say, it will become stale.  There's not a good way to make sure that this get updated as new platforms start handling wide gamut images.  Isn't that a concern?

lib/ui/painting.dart
gaaclarke: Okay, i've elaborated a bit more with my last commit.

lib/ui/painting.dart
jonahwilliams: I don't think it matters that much. Ideally no one will ever see this error so I don't think it matters

gaaclarke: Yep, hopefully not.  I switched it to `UnsupportedError` as in not having a skia image or impeller texture is not supported.  I think that makes the most sense.  Thanks.

lib/ui/painting.dart
jonahwilliams: Its kind of a recusive definition though?

I don't think we need a whole wikipedia article, but short explanation of what the thing is and more specific definition of how its used:

"This value can help you decide which [ImageByteFormat] to use with [Image.toByteData] and is the result of [Image.colorSpace]"

This is sort of vague, how do I use it to decide what?

lib/ui/painting.dart
```diff
@@ -1568,6 +1568,20 @@ class Paint {
   }
 }
 
+/// The color space in which all of an [Image]'s colors exist.
+///
+/// This value can help you decide which [ImageByteFormat] to use with
+/// [Image.toByteData] and is the result of [Image.colorSpace].
+enum ColorSpace {
+  /// The sRGB color space, the defined standard color space for the web.
```
jonahwilliams, line 1576: > the defined standard color space for the web.

This is kind of an odd statement, I'm not sure how that helps me to know what it means for flutter. Perhaps describe this in terms of the values the colors can take and/or what encoding to use?

lib/ui/painting.dart
```diff
@@ -1591,6 +1605,21 @@ enum ImageByteFormat {
...
+  /// Unencoded bytes, in RGBA row-primary form with straight alpha, 32 bit
+  /// float (IEEE 754) per channel.
+  ///
+  /// Example usage:
+  ///
+  /// ```dart
+  /// final ByteData data =
+  ///     (await image.toByteData(format: ImageByteFormat.rawExtendedRgba128))!;
+  /// final Float32List floats = Float32List.view(data.buffer);
+  /// print('${floats[0]} ${floats[1]} ${floats[2]} ${floats[3]}');
```
jonahwilliams, line 1619: You should include which colors are RGB and Alpha.

lib/ui/painting.dart
```diff
@@ -1591,6 +1605,21 @@ enum ImageByteFormat {
   /// image may use a single 8-bit channel for each pixel.
   rawUnmodified,
 
+  /// Raw extended range RGBA format.
+  ///
+  /// Unencoded bytes, in RGBA row-primary form with straight alpha, 32 bit
```
jonahwilliams, line 1610: Is straight alpha premultiplied or un-premultiplied (post multiplied)?

jonahwilliams, line 1610: I think this matches the encoding of the dart:ui Color class, right?

lib/ui/painting.dart
gaaclarke: I've reworded it to make it more clear it is a collection of colors.  Good point on the toByteData usage.  I made that more clear.

lib/ui/painting.dart
gaaclarke: It's un-premultiplied, "straight" is the nomenclature we use in the other fields.

gaaclarke: The dart:ui Color class is using integer values for components, this is float32.

lib/ui/painting.dart
gaaclarke: good call, done.

lib/ui/painting.dart
gaaclarke: I don't think this value has any effect on what encoding you should use.  The only way I could describe the values would be to list the frequencies in the gamut.  Did you have a different idea on how we could specify it?  It is kind of weird to list the web standard but it's the only thing I could grasp onto where people would have an existing familiarity with sRGB.

lib/ui/painting.dart
```diff
@@ -1591,6 +1607,21 @@ enum ImageByteFormat {
...
+  /// float (IEEE 754) per channel.
```
Hixie, line 1613: iirc this is `IEEE754 binary32`, specifically?

lib/ui/painting.dart
```diff
@@ -1737,6 +1768,26 @@ class Image {
     return _image.toByteData(format: format);
   }
 
+  /// The color space that is used by the [Image]'s colors.
+  ///
+  /// This value is a consequence of how the [Image] has been created.  For
+  /// example, loading a PNG that is in the Display P3 color space will result
+  /// in a [ColorSpace.extendedSRGB] image.
+  ///
+  /// On certain platforms/rendering backends, wide gamut images will still
```
Hixie, line 1777: whenever i read "on certain foo" i immediately wonder, which foo? how do i find out which foo?

lib/ui/painting.dart
Hixie: might be worth just linking to wikipedia in a "See also" block, too.

lib/ui/painting.dart
Hixie: defined by whom? linking to the relevant standard would be useful for people new to this topic

lib/ui/painting.dart
gaaclarke: Yep, that's right.  There is another 32bit float format in that spec, decimal32.  Updated.

lib/ui/painting.dart
gaaclarke: Citations are provided in the wiki (but it was standardized by the International Electrotechnical Commission (IEC) as IEC 61966-2-1).  Linking the wikipedia page.

lib/ui/painting.dart
gaaclarke: Addressed in another comment, added supported platforms.

lib/ui/painting.dart
gaaclarke: Link added.

lib/ui/painting.dart
```diff
@@ -1568,6 +1568,26 @@ class Paint {
   }
 }
 
+/// The color space describes the colors that are available to an [Image].
+///
+/// This value can help decide which [ImageByteFormat] to use with
+/// [Image.toByteData], where images that are in the [extendedSRGB] color space
+/// should use something like [ImageByteFormat.rawExtendedRgba128] so that
```
jonahwilliams, line 1575:
```suggestion
/// should use something like [ImageByteFormat.rawExtendedRgba128] so that
```

Nit, use stronger phrasing. Consider something like:

"An image in the extendedSRGB color space must use ImageByteFormat.rawExtendedRgba128 to avoid losing colors. Using ImageByteFormat blah blah will cause loss of color information" And define what loss of color information means if its a well defined process

lib/ui/painting.dart
jonahwilliams: The important part isn't whether or not this is the defined color space for the web. As a flutter developer, what does the web's color space have to do with my application? If we don't have a good answer, maybe just something like "this is the default color space for applications" or something like that

jonahwilliams (approved):
LGTM with nits

lib/ui/painting.dart
gaaclarke: Okay, reworded and I think that information would be better to be situated at `toByteData` so I added it there.

lib/ui/painting.dart
gaaclarke: Okay, reworded it and added note about it being the colorspace of non-wide-gamut apps.

PR #40033, opened 2023-03-02
Add platform channel System.exitApplication and System.requestAppExit support

robert-ancell:
This adds support for System.exitApplication and sending and handling the response from System.requestAppExit to the platform channel.

Fixes Linux part of https://github.com/flutter/flutter/issues/30735


robert-ancell:
@gspencergoog I haven't tested this - can you confirm this is using the API as intended?

robert-ancell:
I've started platform plugin tests in https://github.com/flutter/engine/pull/39992 - will land that first to use that framework.

shell/platform/linux/fl_platform_plugin.cc
```diff
@@ -140,6 +150,45 @@ static FlMethodResponse* clipboard_has_strings_async(
...
+      fl_value_get_type(type_value) != FL_VALUE_TYPE_STRING) {
+    return FL_METHOD_RESPONSE(fl_method_error_response_new(
+        kBadArgumentsError, "Missing exit type", nullptr));
+  }
+  const gchar* type = fl_value_get_string(type_value);
+  bool is_cancelable = g_str_equal(type, kExitTypeCancelable);
+
+  bool canceled = false;
+  if (is_cancelable) {
+    // TODO
```
gspencergoog, line 172: For cancelable requests, the TODO code here will need to send a request back to the framework so that any registered listeners for exit events get a chance to respond to the request, so the rest of this function will need to be inside the "reply" callback for that method channel call. 

This bouncing back into the framework is so that if, for instance, a package registers, it can get notified, and so that requests from the OS (like on session end) go through the same path as the framework-originated requests (which should also go through the method call).

I could also short circuit the request in the framework and only tell the engine to exit when it is required to, but I figured fewer different paths was better. And we'd still need almost the same code to handle requests from the OS.

robert-ancell: Oh, I thought the handler for this was in the native code. I'll update this to the desired behaviour.

robert-ancell: Should be good now, please check again!

gspencergoog (commented):
Yes, this all looks right now.

gspencergoog (approved):

shell/platform/linux/fl_platform_plugin.cc
```diff
@@ -140,6 +153,124 @@ static FlMethodResponse* clipboard_has_strings_async(
...
+  if (exit_response == nullptr) {
+    exit_response = g_strdup(kExitResponseCancel);
+  }
+
+  if (g_str_equal(exit_response, kExitResponseExit)) {
+    GApplication* app = g_application_get_default();
+    if (app == nullptr) {
+      g_warning("Unable to quit, no GApplication");
+    } else {
+      g_application_quit(app);
```
gspencergoog, line 209: BTW, I tried this with the Framework code, and even if it responds "cancel" and never executes this line, the application exits once it returns from this callback. I'm confused as to why, since it definitely isn't calling `g_application_quit` in that case (or in the other instance in this file). Is there another reason it might exit?

You can try it out with the framework changes too: patch in https://github.com/flutter/flutter/pull/121378 and run `examples/api/lib/widgets/application_lifecycle/application_lifecycle.0.dart`.  This works with the macOS version, so I'm pretty sure it's sending the right communication.

gspencergoog: Also, it doesn't seem to go through `fl_platform_plugin_request_app_exit` when I click on the window "X" to close it. I would expect the app to get a chance to ask the user a question before exiting in that case.

shell/platform/linux/fl_platform_plugin.cc
```diff
@@ -140,6 +153,124 @@ static FlMethodResponse* clipboard_has_strings_async(
...
+    exit_response = get_exit_response(method_response);
+  }
+  if (exit_response == nullptr) {
+    exit_response = g_strdup(kExitResponseCancel);
+  }
+
+  if (g_str_equal(exit_response, kExitResponseExit)) {
+    GApplication* app = g_application_get_default();
+    if (app == nullptr) {
+      g_warning("Unable to quit, no GApplication");
```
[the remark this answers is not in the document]
robert-ancell: Yes, that seems like reasonable thing to do.

shell/platform/linux/fl_platform_plugin.cc
```diff
@@ -140,6 +153,124 @@ static FlMethodResponse* clipboard_has_strings_async(
...
+    }
+    g_clear_object(&self->exit_application_method_call);
+  }
+}
+
+// Send a request to Flutter to exit the application.
+static void request_app_exit(FlPlatformPlugin* self, const char* type) {
+  g_autoptr(FlValue) args = fl_value_new_map();
+  fl_value_set_string_take(args, kExitTypeKey, fl_value_new_string(type));
+  fl_method_channel_invoke_method(self->channel, kRequestAppExitMethod, args,
```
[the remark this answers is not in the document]
robert-ancell: Oh, I thought it was still supposed to flow back to Flutter in this case. Fixed.

shell/platform/linux/fl_platform_plugin.cc
gspencergoog: Hmm. Seems that it's a SEGV after that function returns, for some reason it's failing here:

```
Thread 1 "flutter_api_sam" received signal SIGSEGV, Segmentation fault.
malloc_consolidate (av=av@entry=0x7f1b83b12c60 <main_arena>) at ./malloc/malloc.c:4749
4749	./malloc/malloc.c: No such file or directory.

#0  malloc_consolidate (av=av@entry=0x7f1b83b12c60 <main_arena>) at ./malloc/malloc.c:4749
#1  0x00007f1b839d76b8 in _int_malloc (av=av@entry=0x7f1b83b12c60 <main_arena>, bytes=bytes@entry=3256) at ./malloc/malloc.c:3961
#2  0x00007f1b839d8799 in __GI___libc_malloc (bytes=3256) at ./malloc/malloc.c:3323
#3  0x00007f1b873eec3d in operator new(unsigned long) (size=3256) at ../../third_party/libcxx/src/new.cpp:67
#4  0x00007f1b87d6ff99 in std::_LIBCPP_ABI_NAMESPACE::make_unique[abi:v15000]<SkRecorder, decltype(nullptr), SkRect>(decltype(nullptr)&&, SkRect&&) (__args=<error reading variable: Attempt to dereference a generic pointer.>, __args=...)
    at ../../third_party/libcxx/include/__memory/unique_ptr.h:714
#5  0x00007f1b87d6f426 in SkPictureRecorder::SkPictureRecorder() (this=0x561d0b6f0a20)
    at ../../third_party/skia/src/core/SkPictureRecorder.cpp:25
#6  0x00007f1b878635bd in std::_LIBCPP_ABI_NAMESPACE::make_unique[abi:v15000]<SkPictureRecorder>() ()
    at ../../third_party/libcxx/include/__memory/unique_ptr.h:714
#7  0x00007f1b87862d7c in flutter::EmbedderExternalView::EmbedderExternalView(SkISize const&, SkMatrix const&, flutter::EmbedderExternalView::ViewIdentifier, std::_LIBCPP_ABI_NAMESPACE::unique_ptr<flutter::EmbeddedViewParams, std::_LIBCPP_ABI_NAMESPACE::default_delete<flutter::EmbeddedViewParams> >)
    (this=0x7f1b68563a70, frame_size=..., surface_transformation=..., view_identifier=..., params=...)
    at ../../flutter/shell/platform/embedder/embedder_external_view.cc:33
#8  0x00007f1b87862cc0 in flutter::EmbedderExternalView::EmbedderExternalView(SkISize const&, SkMatrix const&)
    (this=0x7f1b68563a70, frame_size=..., surface_transformation=...)
    at ../../flutter/shell/platform/embedder/embedder_external_view.cc:21
#9  0x00007f1b87865dad in std::_LIBCPP_ABI_NAMESPACE::make_unique[abi:v15000]<flutter::EmbedderExternalView, SkISize&, SkMatrix&>(SkISize&, SkMatrix&) (__args=..., __args=...) at ../../third_party/libcxx/include/__memory/unique_ptr.h:714
#10 0x00007f1b87864d1c in flutter::EmbedderExternalViewEmbedder::BeginFrame(SkISize, GrDirectContext*, double, fml::RefPtr<fml::RasterThreadMerger>) (this=0x561d0a1d1340, frame_size=..., context=0x561d0a5abc90, device_pixel_ratio=1, raster_thread_merger=...)
    at ../../flutter/shell/platform/embedder/embedder_external_view_embedder.cc:67
#11 0x00007f1b887a4cf3 in flutter::Rasterizer::DrawToSurfaceUnsafe(flutter::FrameTimingsRecorder&, flutter::LayerTree&)
    (this=0x561d0a1f7ae0, frame_timings_recorder=..., layer_tree=...) at ../../flutter/shell/common/rasterizer.cc:500
#12 0x00007f1b887a2fce in flutter::Rasterizer::DrawToSurface(flutter::FrameTimingsRecorder&, flutter::LayerTree&)
    (this=0x561d0a1f7ae0, frame_timings_recorder=..., layer_tree=...) at ../../flutter/shell/common/rasterizer.cc:471
#13 0x00007f1b887a3fcf in flutter::Rasterizer::DoDraw(std::_LIBCPP_ABI_NAMESPACE::unique_ptr<flutter::FrameTimingsRecorder, std::_LI
```

(Maybe I'm just at a bad engine hash?)

shell/platform/linux/fl_platform_plugin.cc
```diff
@@ -140,6 +152,137 @@ static FlMethodResponse* clipboard_has_strings_async(
...
+  if (method_response == nullptr) {
+    if (g_error_matches(error, G_IO_ERROR, G_IO_ERROR_CANCELLED)) {
+      return;
+    }
+    g_warning("Failed to complete System.requestAppExit: %s", error->message);
+  } else {
+    exit_response = get_exit_response(method_response);
+  }
+  if (exit_response == nullptr) {
+    exit_response = g_strdup(kExitResponseCancel);
```
gspencergoog, line 211: This will likely never happen, but you might want to use `kExitResponseExit` for the default here, otherwise a malformed app can't exit.

gspencergoog:
I think not yet.  Robert's still working out a segfault in the engine. Right @robert-ancell ?

robert-ancell:
Yes, I wanted to be sure the crash wasn't due to this change first.

shell/platform/linux/fl_platform_plugin.cc
robert-ancell: Fixed - this is noticeable if you run this against an older Flutter that doesn't have System.requestAppExit support.

shell/platform/linux/fl_platform_plugin.cc
robert-ancell: Turned out to be incorrect reference usage, fixed now.

robert-ancell:
Crash now fixed, will land once the tests complete.

shell/platform/linux/fl_view.cc
```diff
@@ -89,6 +89,15 @@ G_DEFINE_TYPE_WITH_CODE(
                 G_IMPLEMENT_INTERFACE(fl_text_input_view_delegate_get_type(),
                                       fl_view_text_input_delegate_iface_init))
 
+// Signal handler for GtkWidget::delete-event
+static gboolean window_delete_event_cb(GtkWidget* widget,
+                                       GdkEvent* event,
+                                       FlView* self) {
+  fl_platform_plugin_request_app_exit(self->platform_plugin);
+  // Stop the event from propagating.
+  return TRUE;
```
Kingtous, line 98: This line totally blocks the additional listener provided by plugins, such like `window_manager`. :(

gspencergoog: Yes, I agree, it does, but the alternative is returning FALSE, which would not give the app a chance to stop the exit when the last window goes away.

Perhaps this should only make the exit request and return TRUE if there is only one window open. That would still block the plugin listeners on the last window, though.

iciclize: I found the change prevents `*_plugin_dispose(GObject* object)` from being called.

The `some_plugin_dispose` in the generated template will never be called.

```
static void some_plugin_dispose(GObject* object) {
  G_OBJECT_CLASS(some_plugin_parent_class)->dispose(object);
}

static void some_plugin_class_init(SomePluginClass* klass) {
  G_OBJECT_CLASS(klass)->dispose = some_plugin_dispose;
}
```

PR #40060, opened 2023-03-03
[web][felt] Add 'archive' target + 'profile' mode

mdebbar:
When there's a problem in CanvasKit, it's useful to have an easy way to make a profile build. Now passing `--profile` to `felt build` will build everything in profile mode including CanvasKit. The output is located at `wasm_profile` instead of the usual `wasm_release`.

While I'm at it, I also added a target for the `archive`.


eyebrowsoffire (approved):
This is great! Thank you.

PR #40066, opened 2023-03-03
Apply the buffer offset when converting a vertex color Int32List to a Uint32List

jason-simmons:
Fixes https://github.com/flutter/flutter/issues/121557


mdebbar (approved):
Thanks Jason!

PR #40068, opened 2023-03-04
Added float32 support to decodeImageFromPixels

gaaclarke:
fixes https://github.com/flutter/flutter/issues/121730

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/lib/ui/painting/image_decoder_impeller.cc
+++ b/lib/ui/painting/image_decoder_impeller.cc
@@ -86,7 +86,12 @@ ImageDecoderImpeller::ImageDecoderImpeller(
 ImageDecoderImpeller::~ImageDecoderImpeller() = default;
 
 static SkColorType ChooseCompatibleColorType(SkColorType type) {
-  return kRGBA_8888_SkColorType;
+  switch (type) {
+    case kRGBA_F32_SkColorType:
+      return kRGBA_F16_SkColorType;
+    default:
+      return kRGBA_8888_SkColorType;
+  }
 }
 
 static SkAlphaType ChooseCompatibleAlphaType(SkAlphaType type) {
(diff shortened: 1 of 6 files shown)


gaaclarke:
I'm going to move this back to a draft until https://github.com/flutter/engine/pull/40031 lands.  I suspect that as this is written it will create a texture that is in the f32 format.  That can cause things like getByteData to fail since it is looking for f32 or bgr10_xr.  We can add that test after that lands.

gaaclarke:
I've rebased this onto https://github.com/flutter/engine/pull/40031 and https://github.com/flutter/engine/pull/40097, after those land I'll rebase this onto `main` and this should be ready to review.

gaaclarke:
@dnfield the dependent PR has landed so this is ready for review now.

lib/ui/painting/image_decoder_impeller.cc
```diff
@@ -86,7 +86,12 @@ ImageDecoderImpeller::ImageDecoderImpeller(
 ImageDecoderImpeller::~ImageDecoderImpeller() = default;
 
 static SkColorType ChooseCompatibleColorType(SkColorType type) {
-  return kRGBA_8888_SkColorType;
+  switch (type) {
+    case kRGBA_F32_SkColorType:
+      return kRGBA_F16_SkColorType;
```
dnfield, line 91: Why not
```suggestion
    case kRGBA_F32_SkColorType:
    case kRGBA_F16_SkColorType:
      return kRGBA_F16_SkColorType;
```

lib/ui/painting/image_descriptor.cc
```diff
@@ -85,17 +85,21 @@ void ImageDescriptor::initRaw(Dart_Handle descriptor_handle,
...
   switch (pixel_format) {
     case PixelFormat::kRGBA8888:
       color_type = kRGBA_8888_SkColorType;
       break;
     case PixelFormat::kBGRA8888:
       color_type = kBGRA_8888_SkColorType;
       break;
+    case PixelFormat::kRGBAFloat32:
+      color_type = kRGBA_F32_SkColorType;
+      alpha_type = kUnpremul_SkAlphaType;
```
dnfield, line 98: Why are these unpremul but the other ones are premul?

lib/ui/painting/image_decoder_impeller.cc
gaaclarke: There isn't a way to send in kRGBA_F16_SkColorType directly.  If we decide to implement that it will probably be better to do that in its own pr with its own tests.

lib/ui/painting/image_descriptor.cc
gaaclarke: From the Dart API perspective it's not the only one that uses unpremultiplied, `rawStraightRgba` also returns unpremultiplied alpha.  I'm not sure what flow of the code that goes through to achieve it, but I'm maintaining the old behavior here for old flows.  I chose to make this unpremultiplied since that is the most expected result and the benefits of using premultiplied alpha don't exist here since we have to do a conversion anyway because we can't use f32 directly, we have to convert to f16.

dnfield: Consider adding a comment about that.

dnfield (approved):
LGTM

lib/ui/painting/image_descriptor.cc
gaaclarke: I got a bit confused between `toByteData` and `decodeImageFromPixels`.  The input to toByteData was documented as non-premultiplied, but the input to decodeImageFromPixels wasn't.  I've updated all the documentation for `ui.PixelFormat` to be explicit about pixel format.  I also added a comment in this code.  The important thing is I wanted `ImageByteFormat.rawExtendedRgba128` and `PixelFormat.rgbaFloat32` to match so people can easily download, tweak, then reupload.

PR #40073, opened 2023-03-04
[Impeller] Round glyph locations per-run to fix per-glyph jitter

bdero:
Resolves https://github.com/flutter/flutter/issues/121531.

Round glyph positions relative to the text run origin so that individual characters in the run will snap to the pixel grid together. Note that as the atlas changes, there may still be a subpixel/antialiasing jitter, but this should be much less apparent than the full 1 pixel jitter that this patch fixes. 

Before:

https://user-images.githubusercontent.com/919017/222868028-9cc38097-92c8-47e1-9b2b-1bcbb5d82a94.mov

After:

https://user-images.githubusercontent.com/919017/222868013-654b7598-07e7-4ae2-a985-e785747bc7af.mov

### DIFF
--- a/impeller/aiks/aiks_unittests.cc
+++ b/impeller/aiks/aiks_unittests.cc
@@ -4,10 +4,12 @@
 
 #include <array>
 #include <cmath>
+#include <cstdlib>
 #include <iostream>
 #include <memory>
 #include <tuple>
 #include <utility>
+#include <vector>
 
 #include "flutter/testing/testing.h"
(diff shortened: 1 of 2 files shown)


bdero:
The after video actually has zero visible per-glyph jitter in the runs, but the compression of the video compression is making it look a tiny bit jittery.

jonahwilliams (approved):
LGTM!

chinmaygarde (approved):

PR #40091, opened 2023-03-06
[platform_view]fix regression for addSubview when re-ordering

hellohuanlin:
It turns out that in the benchmark test, `BringLayersIntoView` is called every frame, even if view hierarchy stays the same, resulting in unnecessary subview reordering. 

For example, let's say we have subviews array `[v1, v2, v3]`: 

`[v1, v2, v3]` (initial config, v3 is the top) 
=> `addSubview(v1)` brings v1 to top, it becomes `[v2, v3, v1]` 
=> `addSubview(v2)` brings v2 to top, it becomes `[v3, v1, v2]`
=> `addSubview(v3)` brings v3 to top, it becomes `[v1, v2, v3]`

The fix is simply *to check if the view hierarchy stays the same, and avoid manipulating the subviews if not necessary.*

It used to be fine before https://github.com/flutter/engine/pull/39527, because we simply set the zPosition, which is much less costly. However, we should probably investigate whether we can avoid calling `BringLayersIntoView` in the first place. 

## Performance

Right before https://github.com/flutter/engine/pull/39527: 

right before my PR

On https://github.com/flutter/engine/pull/39527: 

on my PR

This fix on top of https://github.com/flutter/engine/pull/39527: 

no-op fix righ tafter my PR

Fixes https://github.com/flutter/flutter/issues/121833

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/shell/platform/darwin/ios/framework/Source/FlutterPlatformViews.mm
+++ b/shell/platform/darwin/ios/framework/Source/FlutterPlatformViews.mm
@@ -750,16 +750,33 @@ void FlutterPlatformViewsController::BringLayersIntoView(LayersMap layer_map) {
   UIView* flutter_view = flutter_view_.get();
   // Clear the `active_composition_order_`, which will be populated down below.
   active_composition_order_.clear();
+  NSMutableArray* desired_platform_subviews = [NSMutableArray array];
   for (size_t i = 0; i < composition_order_.size(); i++) {
     int64_t platform_view_id = composition_order_[i];
     std::vector<std::shared_ptr<FlutterPlatformViewLayer>> layers = layer_map[platform_view_id];
     UIView* platform_view_root = root_views_[platform_view_id].get();
-    // `addSubview` will automatically reorder subview if it is already added.
-    [flutter_view addSubview:platform_view_root];
+    [desired_platform_subviews addObject:platform_view_root];
     for (const std::shared_ptr<FlutterPlatformViewLayer>& layer : layers) {
-      [flutter_view addSubview:layer->overlay_view_wrapper];
+      [desired_platform_subviews addObject:layer->overlay_view_wrapper];
     }
     active_composition_order_.push_back(platform_view_id);
+  }
+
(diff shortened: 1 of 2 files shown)


shell/platform/darwin/ios/framework/Source/FlutterPlatformViews.mm
```diff
@@ -750,17 +750,34 @@ static bool ClipRRectContainsPlatformViewBoundingRect(const SkRRect& clip_rrect,
...
+      [desired_platform_subviews addObject:layer->overlay_view_wrapper];
     }
     active_composition_order_.push_back(platform_view_id);
   }
+
+  NSSet* desired_platform_subviews_set = [NSSet setWithArray:desired_platform_subviews];
+  NSArray* existing_platform_subviews = [flutter_view.subviews
+      filteredArrayUsingPredicate:[NSPredicate predicateWithBlock:^BOOL(id object,
+                                                                        NSDictionary* bindings) {
+        return [desired_platform_subviews_set containsObject:object];
```
cyanglaz, line 769: `BringLayersIntoView` is meant to be called after `RemoveUnusedLayers();`, so I think any view in `existing_platform_subviews` should already be in `desired_platform_subviews_set`? Did you actually see this filter reduces the size of `existing_platform_subviews`?

Maybe instead we can NSAssert it?

hellohuanlin: > Did you actually see this filter reduces the size of existing_platform_subviews?

Yes, and my understanding is that there are non-platform views added somewhere

hellohuanlin: Not sure if it's dummy scroll view on the top status bar that I saw (which enables tap-to-scroll for table views), but that's not important.

cyanglaz: ah ok forgot about those. Yeah makes sense.

cyanglaz (approved):
LGTM

PR #40103, opened 2023-03-07
Revert "Sequester all Skia<->DL interactions into the skia sub-module (#40083)"

jason-simmons:
This reverts commit b053d74305d2268863077195510f5054087276db.


jason-simmons:
This was causing errors in the DanTup_tiler customer test (sorting/isometric golden image test)

See https://ci.chromium.org/ui/p/flutter/builders/try/Mac%20customer_testing/47476/overview

flar (approved):
LGTM

flar:
The original PR was built from about 6 different commits so I can start with bisecting those to narrow down the cause...

jason-simmons:
The sorting/isometric test diff reproduces with the first commit in the PR (https://github.com/flutter/engine/pull/40083/commits/288dfdbf19da28da559115cd8ecef8bcdda8a4ac)

I can get the test to pass by reverting the changes to `lib/ui/painting/canvas.cc` and `lib/ui/painting/paint.cc` within that commit.  But I haven't figured out which change within those files is causing the difference.

I did notice one change in behavior introduced in that commit.  Before PR #40083 `Paint::sync_to` was checking `flags.is_stroked(builder->getStyle())` to decide whether to set stroke properties.  It was doing this even if `flags.applies_style()` was not set.  So the state of the builder prior to calling `Paint::sync_to` could affect what `sync_to` does.

PR #40083 changes this so that the builder's current state is not available to `Paint::paint`.  `Paint::paint` sets stroke properties based only on the paint data passed from Dart.  This seems correct.

But in any case, it does not look like this issue is affecting the sorting/isometric test.

jason-simmons:
I found the issue - in `Canvas::drawImageRect`, the call to `builder()->DrawImageRect` should pass `false` as the last argument.

(It's currently passing the enum `SkCanvas::kFast_SrcRectConstraint`, which is intended for the old `drawImageRect` method and is being implicitly cast to `enforce_src_edges = true`)

flar:
And a quick grep shows a lot more occurences of that flag in the DL code. I'll follow up and see if any of the others are wrong. It's annoying that the compiler allows implicit conversion between a bool and an enum. Grrr...

PR #40114, opened 2023-03-07
Reland "Sequester all Skia<->DL interactions into the skia sub-module"

flar:
The original version of this PR (https://github.com/flutter/engine/pull/40083) ran afoul of some customer tests due to confusion about the last parameter of drawImageRect. It used to use `SkCanvas::SrcRectConstraint`, but was switched to a bool during 40083 in order to remove the minor dependency on Skia types. Unfortunately, since the `SKCanvas` enums are base enums, they willfully implicitly cast back and forth with numeric types like "bool", so some call sites that were using the Sk values were not detected during testing. To make matters worse, `bool(0) == false` represents "don't enforce the src rect edges" whereas `SrcRectConstraint(0) == kStrict" which means "do enforce the edges".

The solution is to make the `DlCanvas` and `DlOpReceiver` versions of `[dD]rawImageRect` take a new `DlCanvas::SrcRectConstraint` which is a type-safe `enum class`. Lots of lines changed and call sites get more verbose, but the type safety is worth it in the long run.


jason-simmons (approved):

flar:
Also, note performance regression related to gradients (likely due to data copying in the ToSk(gradient) code):
https://flutter-flutter-perf.skia.org/e/?begin=1677571681&end=1678478237&keys=X7adea47f12a68fe0a90611307934ee77&requestType=0&xbaroffset=33677

flar:
Also (probably the same measurement reported under 2 different categories):
https://flutter-flutter-perf.skia.org/e/?begin=1677571681&end=1678478237&keys=X7adea47f12a68fe0a90611307934ee77&requestType=0&xbaroffset=33677

PR #40119, opened 2023-03-07
[Android] Fix initial route when app launched by deeplink and not in background

camsim99:
Adds an additional push of the initial route to the framework once the first Flutter frame has been rendered **if** it has not already been pushed.

This will catch the case of an app being launched by a deeplink that was not previously in the background, because when the app is not previously in the background, no call to push the correct initial route is currently being made and we must wait to push it after the `FlutterActivity` has been created, started, and resumed (in terms of the `Activity` lifecycle).

Fixes https://github.com/flutter/flutter/issues/121244.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/shell/platform/android/test/io/flutter/embedding/android/FlutterActivityAndFragmentDelegateTest.java
+++ b/shell/platform/android/test/io/flutter/embedding/android/FlutterActivityAndFragmentDelegateTest.java
@@ -5,12 +5,14 @@ import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertThrows;
+import static org.junit.Assert.assertTrue;
 import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.ArgumentMatchers.isNotNull;
 import static org.mockito.ArgumentMatchers.isNull;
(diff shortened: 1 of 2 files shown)


shell/platform/android/test/io/flutter/embedding/android/FlutterActivityAndFragmentDelegateTest.java
```diff
@@ -1189,6 +1191,36 @@ public void usesFlutterEngineGroup() {
     assertEquals(engineUnderTest, mockFlutterEngine);
   }
 
+  @Test
+  public void pushesRouteInformationAfterFirstFrameIfApplicable() {
```
reidbaker, line 1195: Is it possible to add a test that ensures the route is not pushed the second time? Basically does hasPushedRouteInformation work.

reidbaker (approved):

gaaclarke (changes requested):
I chatted with @camsim99 and one of our customers who was helping us debug this issue.  I think the solution we want here is in https://github.com/flutter/engine/pull/40148.  I don't believe there is a reason to have both PR's.  This fixes the issue but is a bit more distant to the source of the problem, pushing the route instead of starting the engine with the route.  Let me know if there is anything I'm missing.

camsim99:
Since I agree with @gaaclarke and the other fix has been verified, I'm going to go ahead and close this.

PR #40125, opened 2023-03-07
Fix gif frame reuse

dnfield:
Adds a golden test that hopefully works cross platform 😇 

Fixes https://github.com/flutter/flutter/issues/122134

See also #31098


dnfield:
This reintroduces https://github.com/flutter/flutter/issues/61150#issuecomment-679055858

lib/ui/painting/multi_frame_codec.cc
```diff
@@ -115,17 +115,17 @@ sk_sp<DlImage> MultiFrameCodec::State::GetNextFrameImage(
           << "Frame " << nextFrameIndex_ << " depends on frame "
           << requiredFrameIndex
           << " and no required frames are cached. Using blank slate instead.";
+      return nullptr;
     } else if (lastRequiredFrameIndex_ != requiredFrameIndex) {
       FML_DLOG(INFO) << "Required frame " << requiredFrameIndex
                      << " is not cached. Using blank slate instead.";
```
bdero, line 121: So this path ends up getting hit for the problem GIF? It would seem like we're possibly drawing the wrong frame here, and we're just getting lucky that the frame still looks OK while rendering the wrong frame?

It's common for GIFs/WEBPs/APNGs to keep using one frame as the backdrop across the whole animation. That's not really possible with the way we've written multiframecodec today.

Not sure if we should fix it in this PR, but this comment should at least be updated since it's inaccurate with this change. Maybe just drop the "Using blank slate instead" since that doesn't always happen.

lib/ui/painting/multi_frame_codec.cc
```diff
@@ -115,17 +115,17 @@ sk_sp<DlImage> MultiFrameCodec::State::GetNextFrameImage(
           << "Frame " << nextFrameIndex_ << " depends on frame "
           << requiredFrameIndex
           << " and no required frames are cached. Using blank slate instead.";
+      return nullptr;
```
bdero, line 118: It would be more graceful to just draw the frame against the zeroed bitmap in this case instead of failing the frame.

lib/ui/painting/multi_frame_codec.cc
```diff
@@ -115,17 +115,17 @@ sk_sp<DlImage> MultiFrameCodec::State::GetNextFrameImage(
...
-      // starting point.
-      if (lastRequiredFrame_->getPixels() &&
-          CopyToBitmap(&bitmap, lastRequiredFrame_->colorType(),
-                       *lastRequiredFrame_)) {
-        prior_frame_index = requiredFrameIndex;
-      }
+    }
+    // Copy the previous frame's output buffer into the current frame as the
+    // starting point.
+    if (lastRequiredFrame_->getPixels() &&
```
bdero, line 125: I _think_ adding `requiredFrameIndex != SkCodec::kNoFrame` to the conditional would keep both fixes working.

lib/ui/painting/multi_frame_codec.cc
dnfield: It doesn't get hit becaues of the `else if`.

dnfield: I'mg oing to remove it entirely though.

lib/ui/painting/multi_frame_codec.cc
dnfield: Done

lib/ui/painting/multi_frame_codec.cc
dnfield: I did this a bit differently by making the condition below that updates the "keep" frame every time once we first encounter a keep frame.

I think this is fine because Skia's documentation on this parameter says that the last required frame is the _earliest_ frame, and you can use any frame from that one to the last one. See https://github.com/google/skia/blob/7b18d6c5c676abcb274a8dc6b2896c8c2f667a63/include/codec/SkCodec.h#L640

dnfield:
Added a regression test for the webp issue, and have a patch that now correctly handles both the funky gif and the funky webp.

bdero (approved):

PR #40146, opened 2023-03-08
Add null-aware access to semantics instance

htoor3:
We are making calls on a `SemanticsTextEditingStrategy.instance` that might be null in certain cases inside testing environments.  We need to add conditional property access to these instances.

Fixes https://github.com/flutter/flutter/issues/122131

## Pre-launch Checklist

- [x] I signed the [CLA].


htoor3:
@yjbanov @mdebbar should we merge this for now to unblock `samehere` as we think of a better way to ensure these instances are initialized in testing environments?

lib/web_ui/lib/src/engine/semantics/text_field.dart
```diff
@@ -33,14 +33,14 @@ class SemanticsTextEditingStrategy extends DefaultTextEditingStrategy {
...
-      return instance;
+    if (_instance != null && instance?.owner == owner) {
+      return instance!;
     }
     return _instance = SemanticsTextEditingStrategy(owner);
   }
 
   /// The [SemanticsTextEditingStrategy] singleton.
-  static SemanticsTextEditingStrategy get instance => _instance!;
+  static SemanticsTextEditingStrategy? get instance => _instance!;
```
mdebbar, line 43: I would suggest keeping this non-nullable (i.e. remove the `?`).

And callsites should be updated to use the private nullable `_instance`:
```dart
SemanticsTextEditingStrategy._instance?.activate(this);
SemanticsTextEditingStrategy._instance?.deactivate(this);
```

mdebbar (approved):

lib/web_ui/test/engine/semantics/text_field_test.dart
```diff
@@ -43,7 +43,7 @@ void testMain() {
     setUp(() {
       testTextEditing = HybridTextEditing();
       SemanticsTextEditingStrategy.ensureInitialized(testTextEditing);
-      strategy = SemanticsTextEditingStrategy.instance;
+      strategy = SemanticsTextEditingStrategy.instance!;
```
mdebbar, line 46: Not needed anymore.

lib/web_ui/test/engine/semantics/text_field_test.dart
```diff
@@ -446,7 +446,7 @@ void testMain() {
...
```
mdebbar, line 449: same here

mdebbar (approved):
Ship it!

htoor3:
@mdebbar thoughts on whether I should add to this PR the change we discussed a few days ago on this line https://github.com/flutter/engine/blob/89627fda305517674ec68edc3501a8707eaf2d4f/lib/web_ui/lib/src/engine/semantics/semantics.dart#L1789 - where we move that conditional out? 

Are there are situations in which semantics gets enabled before we hit `updateSemantics` and then it goes through that code path when it shouldn't be?

lib/web_ui/lib/src/engine/semantics/text_field.dart
```diff
@@ -460,6 +460,6 @@ class TextField extends RoleManager {
     if (!isIosSafari) {
       editableElement?.remove();
     }
-    SemanticsTextEditingStrategy.instance.deactivate(this);
+    SemanticsTextEditingStrategy._instance?.deactivate(this);
```
yjbanov, line 463: Is it possible to add an engine test that simulates the situation where dispose is called before the strategy is initialized?

mdebbar:
> @mdebbar thoughts on whether I should add to this PR the change we discussed a few days ago

It's up to you, but I would put it in a separate PR so it can be reverted independently in case it breaks something.

lib/web_ui/test/engine/semantics/text_field_test.dart
```diff
@@ -36,6 +36,33 @@ void testMain() {
...
+      semantics()
+        ..debugOverrideTimestampFunction(() => _testTime)
+        ..semanticsEnabled = true;
+    });
+
+    tearDown(() {
+      semantics().semanticsEnabled = false;
+    });
+
+    test('Calling dispose() pre-initialization will not throw an error', () {
```
htoor3, line 50: @yjbanov would something like this work?

yjbanov (approved):
lgtm

PR #40147, opened 2023-03-08
Revert "[engine] move asset mapping copy to background thread"

zanderso:
Reverts flutter/engine#39918

Reverting for framework device lab failure https://ci.chromium.org/ui/p/flutter/builders/prod/Mac_android%20microbenchmarks/9292/overview

### DIFF
--- a/fml/concurrent_message_loop.cc
+++ b/fml/concurrent_message_loop.cc
@@ -170,14 +170,4 @@ void ConcurrentTaskRunner::PostTask(const fml::closure& task) {
   task();
 }
 
-bool ConcurrentMessageLoop::RunsTasksOnCurrentThread() {
-  std::scoped_lock lock(tasks_mutex_);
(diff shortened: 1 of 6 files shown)


jonahwilliams (approved):
LGTM

jason-simmons:
I suspect that this is happening because `ui_task` is being captured into the task posted to the concurrent task runner.  `ui_task` is a shared object created with `fml::MakeCopyable`, and it will be destructed on whatever thread drops the last reference.

`ui_task` captures some `DartPersistentValue` objects that will be deleted when `ui_task` is destructed.  The `DartPersistentValue` destructor will try to enter the associated Dart isolate and invoke Dart APIs.  This will fail if the destructor is running on a background thread but the UI thread has already entered that isolate.

Try capturing the `DartPersistentValue` objects as raw pointers and then rewrapping them in `unique_ptr` objects within the code that is explicitly scheduled on the UI thread.

PR #40150, opened 2023-03-08
Improve error messaging when render target cannot be created

dnfield:
Should help with https://github.com/flutter/flutter/issues/121485


jonahwilliams (approved):
LGTM

PR #40152, opened 2023-03-08
[web] Fix SkParagraphBuilder tests for the chromium variant

mdebbar:
In tests, when we use the `SkParagraphBuilder` APIs directly (bypassing our `CkParagraphBuilder`) then we need to explicitly pass any client ICU data that it needs.


eyebrowsoffire (approved):
LGTM!

PR #40154, opened 2023-03-08
[web] Better way to detect CanvasKit variant

mdebbar:
Partial revert of https://github.com/flutter/engine/pull/40027

For more info, see: https://skia-review.googlesource.com/c/skia/+/653416

The above CL introduces a new CanvasKit API to tell us whether it needs client ICU data or not. This allows us to do a more robust detection of the CanvasKit variant we are using.

~**_BEFORE MERGING:_**~
- ~Make a new CanvasKit release that contains the required changes, and roll to that version.~
- _We are now checking for the existence of the method before invoking it._

Fixes https://github.com/flutter/flutter/issues/121905


eyebrowsoffire (approved):

lib/web_ui/lib/src/engine/browser_detection.dart
```diff
@@ -268,6 +268,4 @@ int _detectWebGLVersion() {
 }
 
 /// Whether the current browser supports the Chromium variant of CanvasKit.
-const bool browserSupportsCanvaskitChromium = false;
-// TODO(mdebbar): Uncomment this to enable real detection of browser support.
-// final bool browserSupportsCanvaskitChromium = domIntl.v8BreakIterator != null;
+final bool browserSupportsCanvaskitChromium = domIntl.v8BreakIterator != null;
```
yjbanov, line 271: If `canvaskit_chromium` also removes the wasm image codecs, then this should also add `&& browserSupportsImageDecoder`. Although in practice, I'm not aware of any Chromium-based browser not supporting image decoders.

yjbanov (approved):

lib/web_ui/lib/src/engine/browser_detection.dart
mdebbar: Agreed.

mdebbar: https://github.com/flutter/flutter/issues/122331

mdebbar:
There's a timeout failure happening in wasm mode. This line is causing a hard wasm failure:
https://github.com/flutter/engine/blob/161f75931834b79ec9f700046c4d3fe748a7a60b/lib/web_ui/lib/src/engine/dom.dart#L2164
because the `js_util.getProperty(...)` patch in dart2wasm has the [wrong signature](https://dart.googlesource.com/sdk/+/fadbd51c83b2d15b3bc39d53614a1fe75b1ef8d0/sdk/lib/_internal/wasm/lib/js_util_patch.dart#83) (it shoud be `Function(Object, Object)` instead of `Function(Object, String)`).

This is being fixed in https://dart-review.googlesource.com/c/sdk/+/288161. (Thanks @joshualitt!). Once that Dart CL lands and rolls into the engine, I'll rebase and try again.

Thanks @eyebrowsoffire for looking into this with me!

PR #40156, opened 2023-03-08
Remove single-window assumption from scenario_app

goderbauer:
Part of https://github.com/flutter/flutter/issues/116929.

### DIFF
--- a/testing/scenario_app/lib/src/animated_color_square.dart
+++ b/testing/scenario_app/lib/src/animated_color_square.dart
@@ -14,8 +14,6 @@ import 'scenario.dart';
 /// that are constantly changing.
 class AnimatedColorSquareScenario extends Scenario {
   /// Creates the AnimatedColorSquare scenario.
-  ///
-  /// The [dispatcher] parameter must not be null.
   AnimatedColorSquareScenario(super.view);
 
   static const double _squareSize = 200;
--- a/testing/scenario_app/lib/src/animated_color_square.dart
+++ b/testing/scenario_app/lib/src/animated_color_square.dart
@@ -16,19 +16,18 @@ class AnimatedColorSquareScenario extends Scenario {
   /// Creates the AnimatedColorSquare scenario.
   ///
   /// The [dispatcher] parameter must not be null.
-  AnimatedColorSquareScenario(PlatformDispatcher dispatcher)
-      : super(dispatcher);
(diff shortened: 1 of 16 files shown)


testing/scenario_app/lib/src/animated_color_square.dart
```diff
@@ -14,21 +14,18 @@ import 'scenario.dart';
...
-  /// The [dispatcher] parameter must not be null.
-  AnimatedColorSquareScenario(PlatformDispatcher dispatcher)
-      : super(dispatcher);
+  AnimatedColorSquareScenario(super.view);
 
   static const double _squareSize = 200;
   /// Used to animate the red value in the color of the square.
   final _NumberSwinger<int> _r = _NumberSwinger<int>(0, 255);
-  _NumberSwinger<double> _top = _NumberSwinger<double>(
+  late _NumberSwinger<double> _top = _NumberSwinger<double>(
```
loic-sharma, line 22: My knowledge of Dart is failing me. Is the `late` necessary here to initialize the field after the constructor?

goderbauer: With `late` this gets initialized when the field is first accessed. That allows us to access other fields (here `view`) from the initializer. Without it, the field is initialized at construction time and accessing other non-static fields would not be allowed.

testing/scenario_app/lib/src/platform_view.dart
```diff
@@ -540,7 +518,7 @@ class MultiPlatformViewBackgroundForegroundScenario extends Scenario
     if (_lastLifecycleState == 'AppLifecycleState.inactive' &&
         message == 'AppLifecycleState.resumed') {
       _nextFrame = _secondFrame;
-      window.scheduleFrame();
+      view.platformDispatcher.scheduleFrame();
```
loic-sharma, line 521: Hm, one day we may want to schedule a frame for a single view instead of all views. Food for thought, not at all relevant to this change :)

loic-sharma (approved):
🚀

testing/scenario_app/lib/src/platform_view.dart
goderbauer: Yeah, this is something that comes up every now and then. Since we want a unified widget tree, frame scheduling is always gonna be global. There is some discussion on that in https://flutter.dev/go/multiple-views in the Background section.

PR #40159, opened 2023-03-09
[Impeller][Compute] Guard subgroups with feature detection

dnfield:
This hasn't been a problem because the tests are always running on hardware that supports subgroups. This should give a better failure if you try to run the test on hardware that does not support subgroups, and will be more important as we get to runtime selection of which rendering strategies to use.


jonahwilliams (approved):
LGTM

PR #40160, opened 2023-03-09
Remove single window assumption from testing/dart

goderbauer:
Part of https://github.com/flutter/flutter/issues/116929.

### DIFF
--- a/testing/dart/observatory/skp_test.dart
+++ b/testing/dart/observatory/skp_test.dart
@@ -25,7 +25,7 @@ void main() {
     );
 
     final Completer<void> completer = Completer<void>();
-    window.onBeginFrame = (Duration timeStamp) {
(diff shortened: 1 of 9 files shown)


testing/dart/observatory/skp_test.dart
```diff
@@ -35,14 +35,14 @@ void main() {
       builder.addPicture(Offset.zero, picture);
       final Scene scene = builder.build();
 
-      window.render(scene);
+      PlatformDispatcher.instance.implicitView!.render(scene);
```
loic-sharma, line 38: The comments on line 40 and lines 42-43 look like test code that was never removed. Consider deleting it.

loic-sharma (approved):

testing/dart/observatory/skp_test.dart
goderbauer: Will do!

PR #40163, opened 2023-03-09
Make the Linux GTK widget focusable

jason-simmons:
Handling of keyboard focus for Flutter text fields appears to have regressed with https://github.com/flutter/engine/pull/39473, which removed the call to `gtk_widget_set_can_focus`

### DIFF
--- a/shell/platform/linux/fl_view.cc
+++ b/shell/platform/linux/fl_view.cc
@@ -667,7 +667,9 @@ static void fl_view_class_init(FlViewClass* klass) {
                                        fl_view_accessible_get_type());
 }
 
-static void fl_view_init(FlView* self) {}
+static void fl_view_init(FlView* self) {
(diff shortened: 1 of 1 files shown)


cbracken (approved):
This lgtm. Thanks for spotting. I'm guessing we didn't have a test for this or it would already have been failing.

It looks like we should be able to test this with `gtk_widget_get_can_focus`.

robert-ancell (approved):
There were a lot of focus related calls that seemed redundant in the code before the simplification, I guess I took out one too many. Confirmed it was broken and this fixes it.

robert-ancell:
Adding a test for `gtk_widget_get_can_focus` would catch a regression in this one particular line, but what we really want is a test that generates a key event and confirms it gets all the way to Flutter. There's a lot of focus / keyboard handling code that could regress.

Hopefully this PR can get a test exemption, but the test proposed by @cbracken should be sufficient if not.

PR #40175, opened 2023-03-09
Change `extends Iterator` to using `implements`

lrhn:
The Dart 3.0 libraries will mark `Iterator` with the `interface` class modifier, which prevents `extends`. It will do so because the class has no implementation to inherit, and is only intended as an interface, which it is now possible to express.

This should unblock relanding https://dart-review.googlesource.com/c/sdk/+/287760
(Also working on disabling the class-modifiers experiment for Flutter `dart:` libraries, which was enabled along with the Dart SDK libraries, until the experiment can be intentionally turned back on.)


lrhn:
There is no test, because the change is impossible to test. It's a refactoring which changes the internal shape of the class hierarchy minimally, but does not change anything that can be detected from Dart code.

mkustermann (approved):

PR #40181, opened 2023-03-09
[web] Use CK's malloced array correctly

mdebbar:
Pass the malloced array to CanvasKit instead of passing the underlying typed array.

### DIFF
--- a/lib/web_ui/test/canvaskit/text_fragmenter_test.dart
+++ b/lib/web_ui/test/canvaskit/text_fragmenter_test.dart
@@ -15,9 +15,17 @@ void main() {
 void testMain() {
   setUpCanvasKitTest();
 
+  late SkUint32List breaks;
+
+  tearDown(() {
+    if (browserSupportsCanvaskitChromium) {
+      free(breaks);
+    }
+  });
+
   group('$fragmentUsingIntlSegmenter', () {
     test('fragments text into words', () {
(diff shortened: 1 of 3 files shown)


lib/web_ui/test/canvaskit/text_fragmenter_test.dart
```diff
@@ -19,23 +17,23 @@ void testMain() {
...
       expect(
-        breaks,
+        breaks.toTypedArray(),
         orderedEquals(<int>[0, 5, 6, 11, 12, 14, 16]),
       );
     });
 
     test('fragments multi-line text into words', () {
-      final Uint32List breaks = fragmentUsingIntlSegmenter(
+      final SkUint32List breaks = fragmentUsingIntlSegmenter(
```
eyebrowsoffire, line 31: In these tests, we may want to actually free these objects afterward. CanvasKit normally will free these for you after you pass them into the APIs, but since they aren't actually being passed into a CanvasKit API, we may be on the hook to free them ourselves. Probably doesn't matter that much for tests, but we are probably technically leaking memory in the CanvasKit module.

eyebrowsoffire (approved):
Looks good! Small thing that you can ignore if you don't care.

lib/web_ui/test/canvaskit/text_fragmenter_test.dart
mdebbar: Good idea!

PR #40183, opened 2023-03-09
Reland: Move asset opening to background thread, fix dart persistent value destruction

jonahwilliams:
Reland of https://github.com/flutter/engine/pull/39918 but with the fixes suggested by @jason-simmons from  https://github.com/flutter/engine/pull/40147


chinmaygarde (approved):

lib/ui/painting/immutable_buffer.cc
```diff
@@ -101,20 +143,28 @@ Dart_Handle ImmutableBuffer::initFromFile(Dart_Handle raw_buffer_handle,
...
+        if (!buffer_callback_ptr) {
+          return;
+        }
+        auto dart_state = buffer_callback_ptr->dart_state().lock();
         if (!dart_state) {
           return;
         }
         tonic::DartState::Scope scope(dart_state);
+
+        std::unique_ptr<tonic::DartPersistentValue> buffer_handle(
```
jason-simmons, line 163: Looked into this, and it's actually ok to destruct the `tonic::DartPersistentValue` outside of a `DartState::Scope`.  The `DartPersistentValue` holds a weak reference to the `DartState` used during construction and will enter that `DartState`'s scope to delete the wrapped value.

So these `unique_ptr`s should be moved to the top of the function in order to ensure that they are not leaked if it returns early.

jason-simmons (approved):

PR #40184, opened 2023-03-09
[Impeller] Render a compute generated stroke

dnfield:
Some minor cleanup around buffer creation.

Adds code to existing test to make the heart cubics actually render.

This shows some problems especially around cusps when you bump up the stroke width, which I'll look at in a separate patch.

image

image
image


jonahwilliams (approved):
LGTM

PR #40186, opened 2023-03-09
Revert "Reland "Sequester all Skia<->DL interactions into the skia sub-module""

CaseyHillers:
Reverts flutter/engine#40114

See b/272191483. This broke a customer test where the opacity now shows undesirable results.


CaseyHillers:
Was already reverted.

PR #40193, opened 2023-03-09
[Impeller] support for foreground shaders on text

jonahwilliams:
Fixes https://github.com/flutter/flutter/issues/120003

FYI @bdero if you have any ideas to make this less ... messy

### Impeller

flutter_08

### Skia

flutter_09


chinmaygarde (approved):

impeller/aiks/canvas.cc
```diff
@@ -380,16 +381,40 @@ void Canvas::DrawTextFrame(const TextFrame& text_frame,
...
+    auto& source = paint.color_source.value();
+    auto color_text_contents = std::make_shared<ColorSourceTextContents>();
+    entity.SetTransformation(GetCurrentTransformation());
+
+    Entity test;
+    auto cvg = text_contents->GetCoverage(test).value();
+    color_text_contents->SetTextPosition(cvg.origin + position);
+
+    text_contents->SetInverseMatrix(
+        Matrix::MakeTranslation(Vector3(-cvg.origin.x, -cvg.origin.y, 0)));
```
bdero, line 402: Rather than creating a new parameter, just amend this matrix to the entity transform?

bdero:
I got to this too late, but I think most of this can be simplified away by just taking a snapshot of the text contents and modulating by the ColorSource with cover geometry. No need for sampling the coverage with a "test" entity, etc.

junaid1460:
is this in 3.10? Doesn't look like it is? For me it just renders color less text

PR #40199, opened 2023-03-09
Impeller: Another quick removal of implicit conversions.

gaaclarke:
Was working with the gradient code and noticed this as well, might as well update it for better performance.

issue: https://github.com/flutter/flutter/issues/122324

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/impeller/compiler/shader_lib/impeller/texture.glsl
+++ b/impeller/compiler/shader_lib/impeller/texture.glsl
@@ -57,12 +57,12 @@ const float kTileModeDecal = 3;
 /// When `t` is between [0 to 1), the original unchanged `t` is always returned.
 float IPFloatTile(float t, float tile_mode) {
   if (tile_mode == kTileModeClamp) {
-    t = clamp(t, 0.0, 1.0);
+    t = clamp(t, 0.0f, 1.0f);
   } else if (tile_mode == kTileModeRepeat) {
     t = fract(t);
   } else if (tile_mode == kTileModeMirror) {
-    float t1 = t - 1;
-    float t2 = t1 - 2 * floor(t1 * 0.5) - 1;
+    float t1 = t - 1.0f;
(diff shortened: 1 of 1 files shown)


jonahwilliams:
According to https://www.khronos.org/opengl/wiki/Data_Type_(GLSL)#Literals

> A numeric literal that uses a decimal is by default of type float. To create a float literal from an integer value, use the suffix f or F as in C/C++.

So i don't think you need the `f` specifier. Does malioc show a diff on this?

gaaclarke:
> According to https://www.khronos.org/opengl/wiki/Data_Type_(GLSL)#Literals
> 
> > A numeric literal that uses a decimal is by default of type float. To create a float literal from an integer value, use the suffix f or F as in C/C++.
> 
> So i don't think you need the `f` specifier. Does malioc show a diff on this?

Nope, there's no difference with the `f`, just force of habit.  The changes are in the int->float in them mirroring branch.

jonahwilliams:
If there was a malioc diff the linux shard would have failed - but it passing indicates to me that we aren't doing any conversions in this code - at least not in the compiled version.

gaaclarke:
Hmm, I find that hard to believe.  CI can't retrieve the output from the malioc diff: https://ci.chromium.org/ui/p/flutter/builders/try/Linux%20Unopt/26307/overview

Shouldn't it only fail if there is a regression?

jonahwilliams:
No, it uses a golden checked in file. So if anything changed it would fail. What happens when you run it locally?

jonahwilliams:
I mean I'm fine with the PR as is, but I don't think we should be spending time worrying about whether we're using decimals everywhere if its not improving performance

gaaclarke:
Yep, looks like the compiler can do the optimization from int literal to float in this case.  I'm fine dropping this until we get the compilation flag.  Since I changed it I was thinking I'm submit it.

PR #40207, opened 2023-03-10
Find the emsdk by relative path so that this works with the monorepo.

eyebrowsoffire:
This ends up basically finding the emsdk directory relative to the gclient checkout, which is different in the monorepo than it is in a engine checkout by itself. So we can find the emsdk directory relative to the script itself.


whesse (approved):
LGTM, don't know if you want another reviewer

PR #24916, opened 2021-03-10
Linux texture support

huanghongxun:
Add support for external textures to the Linux platform shell.

# Related issues
Fixes flutter/flutter#64188.

# Design
For API, I introduce an abstract class `FlTexture` for abstration of different OpenGL textures, allowing any other textures to be supported in the future.
Subclass `FlPixelBufferTexture` support loading textures from given pixel buffers with RGBA format. For other formats, or textures not from pixel buffers, developer may implement their own texture class.

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [Flutter Style Guide] and the [C++, Objective-C, Java style guides].
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I added new tests to check the change I am making or feature I am adding, or Hixie said the PR is test-exempt. See [testing the engine] for instructions on
writing and running engine tests.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I signed the [CLA].
- [x] All existing and new tests are passing.
- [ ] The reviewer has submitted any presubmit flakes in this PR using the [engine presubmit flakes form] before re-triggering the failure.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/flutter/wiki/Tree-hygiene#overview
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[C++, Objective-C, Java style guides]: https://github.com/flutter/engine/blob/master/CONTRIBUTING.md#style
[testing the engine]: https://github.com/flutter/flutter/wiki/Testing-the-engine
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/wiki/Chat
[engine presubmit flakes form]: https://forms.gle/Wc1VyFRYJjQTH6w5A

PR #40208, opened 2023-03-10
Make the context current before accessing GL in MakeSkiaGpuImage

dnfield:
The unconditional `MakeBitmapImage` is no longer necessary on Linux, with or without this context switch change. That fixes https://github.com/flutter/flutter/issues/108835
Might help https://github.com/flutter/flutter/issues/121485

Context switch fixes https://github.com/flutter/flutter/issues/120815. This was figured out by @jason-simmons - without the switch, we might be piggybacking on GDK's GL context/FBO and causing the problems seen in that bug.

I'm not quite sure what to do about the MakeBitmapImage path. It's not entirely safe, for example if you went to background on iOS and have a display_list that contains an SkImage_gpu. We have this problem elsewhere too AFAICT (e.g. in the regular `toImage` path).

~~I'm not sure how to test this. Will try to think on that a bit.~~

### DIFF
--- a/shell/common/BUILD.gn
+++ b/shell/common/BUILD.gn
@@ -328,5 +328,13 @@ if (enable_unittests) {
         deps += [ "//third_party/swiftshader" ]
       }
     }
+
+    if (shell_enable_gl) {
+      deps += [
+        "//third_party/angle:libEGL_static",
+        "//third_party/angle:libGLESv2_static",
+        "//third_party/swiftshader",
+      ]
+    }
   }
 }
--- a/shell/common/fixtures/shell_test.dart
+++ b/shell/common/fixtures/shell_test.dart
@@ -354,6 +354,9 @@ void scene_with_red_box() {
   PlatformDispatcher.instance.scheduleFrame();
 }
 
+@pragma('vm:external-name', 'NativeOnBeforeToImageSync')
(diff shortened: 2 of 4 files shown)


jason-simmons (approved):

dnfield:
I think I know how to test this. Will take a stab at it later today.

dnfield:
Added a test.

It doesn't cover the exact same scenario as the bug, but basically forces the context on the raster (platform) thread to get trampled. If the rasterizer fails to make it curent again the test fails.

alexmercerind:
It seems this broke whole external texture API on Linux: https://github.com/flutter/engine/pull/24916. :pray: 
And... essentially video rendering offered by [`package:media_kit`](https://github.com/alexmercerind/media_kit) on Flutter Linux's `master` channel.

You may try running this minimal application for testing:
https://github.com/lattice0/external_texture_flutter_linux/tree/master/external_texture/example

No GL content draws to the UI.

dayron9110:
@dnfield  so this change was returned?

dnfield:
> @dnfield so this change was returned?

I don't think I understand the question. This patch was not reverted in Flutter, but it looks like it was reverted in a fork of Flutter somewhere else.

PR #40239, opened 2023-03-11
Use plain Uint32List objects with the fragmenter APIs.

eyebrowsoffire:
Let CanvasKit do the malloc/copy into linear memory for us.

Doing malloc ourselves and attempting to use a view into the malloced linear memory fails on dart2wasm for the same reasons we get failures in https://github.com/flutter/flutter/issues/118334

I think this is basically equivalent perf-wise anyway, it's just changing who is doing the work (now CanvasKit does the malloc + copy instead of us).

### DIFF
--- a/lib/web_ui/lib/src/engine/canvaskit/canvaskit_api.dart
+++ b/lib/web_ui/lib/src/engine/canvaskit/canvaskit_api.dart
@@ -1911,12 +1911,12 @@ extension SkParagraphBuilderExtension on SkParagraphBuilder {
   // into a utf16 string.
   String getText() => utf8.decode(getTextUtf8().codeUnits);
 
-  external void setWordsUtf8(SkUint32List words);
-  external void setWordsUtf16(SkUint32List words);
-  external void setGraphemeBreaksUtf8(SkUint32List graphemes);
-  external void setGraphemeBreaksUtf16(SkUint32List graphemes);
-  external void setLineBreaksUtf8(SkUint32List lineBreaks);
-  external void setLineBreaksUtf16(SkUint32List lineBreaks);
+  external void setWordsUtf8(Uint32List words);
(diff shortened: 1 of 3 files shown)


mdebbar (approved):
I guess using malloc is only useful when we are passing the same data multiple times to CK (we may do this in the future if we decide to cache the fragmentation results).

It looks like we are doing the same mistakes with `toSharedSkColor1()` (we are passing the typed array to CK instead of the malloc'd object). Does it fail on dart2wasm too?

eyebrowsoffire:
> It looks like we are doing the same mistakes with `toSharedSkColor1()` (we are passing the typed array to CK instead of the malloc'd object). Does it fail on dart2wasm too?

Hmm, probably. Let me take a look. I'm guessing whatever we do is not covered by a unit test.

eyebrowsoffire:
Actually, it appears the `toSharedSkColor` functions work in dart2wasm, they just don't use the shared buffer. If you look in `_populateSkColor`, we call `toTypedArray()` on the shared buffer, but that ends up creating a copy of the `Float32List`. But after that, the `Float32List` itself is actually passed around, and eventually back into the CanvasKit APIs. Since it's a completely disconnected buffer (not a view of CanvasKit's linear memory), CanvasKit creates its own linear memory buffer and copies the floats into that. So this still does the correct thing, but it undermines the shared buffer optimization basically.

That being said, I actually think the shared buffer optimization might actually be pointless, because for a few fixed sized buffers (like two floats, four floats, etc), CanvasKit itself actually keeps a scratch buffer around for passing to and from APIs. So if we just pass our own buffer of things into CanvasKit, it will basically do the same optimization for us that we're trying to do.

So we might actually be able to take those `toSkSharedColor` calls out completely with no perf hit (in fact it might be faster) but to be honest, I don't really want to dig into that with this change.

PR #40241, opened 2023-03-11
[macOS] Forward mouseDown/Up to view controller

cbracken:
This works around an AppKit bug in which mouseDown/mouseUp events are not correctly forwarded up the responder chain for views nested inside an `NSPopover` if (and only if) the macOS "Reduce Transparency" accessibility setting is enabled in the System Settings.

When the above conditions are satisfied, the nested NSView receives the `mouseDown:`/`mouseUp:` call but if it delegates to the default implementation (implemented in `NSResponder`) `mouseDown:`/`mouseUp:` calls are triggered on containing views (in our case `FlutterViewWrapper`) but not triggered on the view controller and other responders in the responder chain until we an `_NSPopoverWindow` class is hit.

A minimal AppKit-only (non-Flutter) repro shows this bug repros with even a minimal `NSViewController` implementation and an unmodified `NSView`. See: https://github.com/cbracken/PopoverRepro

A radar has been filed with Apple and a copy posted to OpenRadar. See: http://www.openradar.me/FB12050037

In order to work around this bug, we override mouseDown/mouseUp in the topmost containing view of `FlutterView` (in our case, `FlutterViewWrapper`) to have the behaviour documented as the default behaviour in `NSResponder`'s `mouseDown:`/`mouseUp:` documentation. In other words, to simply forward the call to `self.nextResponder`.
See: https://developer.apple.com/documentation/appkit/nsresponder/1524634-mousedown

Because replicating the exact configuration of a `FlutterView` contained in an `NSPopover` and System Settings that have been modified to enable the "Reduce Transparency" setting is difficult and likely error-prone in infra, we instead simulate the bug by testing that even if `NSResponder`'s mouseDown/mouseUp method are swizzled to no-op, these calls are correctly forwarded to the next responder in the chain.

If, in the future Apple does fix this issue, this workaround can be removed once Flutter's minimum supported macOS SDK is at least the version that contains the fix.

Issue: https://github.com/flutter/flutter/issues/115015

## Pre-launch Checklist

- [X] I signed the [CLA].

### DIFF
--- a/shell/platform/darwin/macos/framework/Source/FlutterViewController.mm
+++ b/shell/platform/darwin/macos/framework/Source/FlutterViewController.mm
@@ -278,6 +278,34 @@ void OnKeyboardLayoutChanged(CFNotificationCenterRef center,
   return @[ _flutterView ];
 }
 
+- (void)mouseDown:(NSEvent*)event {
+  // Work around an AppKit bug where mouseDown/mouseUp are not called on the view controller if the
+  // view is the content view of an NSPopover AND macOS's Reduced Transparency accessibility setting
+  // is enabled.
+  //
+  // This simply calls mouseDown on the next responder in the responder chain as the default
+  // implementation on NSResponder is documented to do.
+  //
+  // See: https://github.com/flutter/flutter/issues/115015
+  // See: http://www.openradar.me/FB12050037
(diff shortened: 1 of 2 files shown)


stuartmorgan (approved):
LGTM

PR #40248, opened 2023-03-13
[impeller] implement GetPositionUVBuffer

jonahwilliams:
Implements GetPositionUVBuffer for all geometry classes, moving the computation out of the vertex shader. Effectively, this makes the texture_fill and tiled_texture_fill vertex shaders identical. Also removes some duplicated code from TiledTextureContents to handle vertices.

This will make it easier to perform more cleanups, specifically we should be able to adjust this method a bit more to support a source rect, which will allow us to make Texture contents and TiledTexture contents both use a geometry class. Furhtermore, this will allow us to delete one of the vertex shaders, and replace the tiled_texture fragment shader with a specialize decal shader.

Optimizations that apply to drawImage/drawImageRect will be easier to make if they can be shared with TiledTexture (i.e. color filter forwarding, et cetera).

Fixes https://github.com/flutter/flutter/issues/118553


impeller/entity/geometry.cc
```diff
@@ -441,14 +521,58 @@ GeometryResult StrokePathGeometry::GetPositionBuffer(
...
+  Scalar stroke_width = std::max(stroke_width_, min_size);
+
+  auto& host_buffer = pass.GetTransientsBuffer();
+  auto stroke_builder = CreateSolidStrokeVertices(
+      path_, stroke_width, miter_limit_ * stroke_width_ * 0.5, stroke_cap_,
+      GetJoinProc(stroke_join_), GetCapProc(stroke_cap_),
+      entity.GetTransformation().GetMaxBasisLength());
+
+  VertexBufferBuilder<TextureFillVertexShader::PerVertexData> vertex_builder;
+  stroke_builder.IterateVertices(
```
jonahwilliams, line 562: I took a bit of a shortcut here, because refactoring the stroke computation to handle either path felt a bit like a premature optimization

jonahwilliams:
This has some bugs, from checking playground.

jonahwilliams:
All, the difference is that I used to always treat the origin for vertices as (0, 0), I'll need to double check if it should be consistently zero or the coverage origin. I suspect the former.

bdero (approved):
LGTM

PR #40253, opened 2023-03-13
Use LTO when building the macOS desktop framework

zanderso:


Hixie:
Spoke with @zanderso, he said he's waiting for desktop builds to finish migrating to engine v2 before continuing with this PR.

Hixie:
Spoke with @zanderso, he said this was on his list for Q3.

zanderso:
It turns out that building the desktop frameworks with LTO results in a larger binary.

PR #40254, opened 2023-03-13
Make no_lto in presubmit explicit in .ci.yaml

zanderso:
Currently --no-lto is hardcoded in presubmit builds in the ci_yaml starlark file here https://cs.opensource.google/flutter/infra/+/main:config/lib/ci_yaml/ci_yaml.star;l=398?q=lto&ss=flutter%2Finfra:config%2F.

This PR shifts setting the property into the `.ci.yaml` file. Since in general we want LTO in prod builds, and no LTO in try builds, this PR sets the `no_lto` property to `true` in the `platform_properties`, then for each configuration sets `no_lto` to `false` in their respective `postsubmit_properties`. This would be less confusing if `.ci.yaml` supported a `presubmit_properties` that could be set once-and-for-all in the `platform_properties` cc @keyonghan

### DIFF
--- a/.ci.yaml
+++ b/.ci.yaml
@@ -24,6 +24,7 @@ platform_properties:
       ios_debug: "false"
       ios_profile: "false"
       ios_release: "false"
+      no_lto: "true"
       # CIPD flutter_internal/java/openjdk/$platform
       dependencies: >-
         [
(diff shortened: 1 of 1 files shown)


CaseyHillers:
When this lands, we can remove the starlark code at https://cs.opensource.google/flutter/infra/+/main:config/lib/ci_yaml/ci_yaml.star;l=397-398?q=ci_yaml.star&ss=flutter%2Finfra

keyonghan (approved):
LGTM

PR #40255, opened 2023-03-13
[Impeller] Improve performance / reduce memory usage of path based shadows.

jonahwilliams:
We currently have fast paths for rect, rrect, and circle/oval based shadows. For all other paths, we perform 3 passes - one to render a solid color to a snapshot (at scale) and then two gaussian decal blurs. This is fairly expensive, especially if the ultimate size is large.

Rendering the solid color snapshot is wasteful, instead we could allocate a 1x1 texture with the correct color and use that. This is done by implementing RenderToSnapshot in SolidColorContents.

I did find a particularly large shadow in the new flutter gallery. Its possible this is faster because I broke something, but so far everything seems visual OK.

## Before

image

## After

image

### DIFF
--- a/impeller/entity/contents/solid_color_contents.cc
+++ b/impeller/entity/contents/solid_color_contents.cc
@@ -39,6 +39,53 @@ std::optional<Rect> SolidColorContents::GetCoverage(
   return geometry_->GetCoverage(entity.GetTransformation());
 };
 
+std::optional<Snapshot> SolidColorContents::RenderToSnapshot(
+    const ContentContext& renderer,
+    const Entity& entity,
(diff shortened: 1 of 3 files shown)


jonahwilliams:
hmm, but what if the coverage is wrong. Hmmmmm

jonahwilliams:
this is probably not what I want to do exactly. Not only does it need to be a solid color, but the geometry needs to be such that 100% of the resulting texture would be a solid color. I think we can determine that but I'll need to do more investigation

jonahwilliams:
Ahh yeah, this doesn't work. Oh well

PR #40262, opened 2023-03-13
.ci.yaml: s/flutter_internal/flutter

dnfield:
The flutter_internal version of the package is bad for arm64, whereas the flutter version is not. A recipe change will point the bots to flutter instead of flutter_internal.


yusuf-goog (approved):

dnfield:
https://flutter-review.googlesource.com/c/recipes/+/40320

PR #40285, opened 2023-03-14
Analyze more shaders

zanderso:
This PR adds analysis of shaders that are compiled to SPIR-V 1.3 for the vulkan backend. This allows analyzing shaders that use ssbos and framebuffer fetch.


jonahwilliams (approved):
LGTM!

PR #40290, opened 2023-03-14
Add fl_texture_get_id, so textures can be passed to Flutter

robert-ancell:
Was making use of the texture API and found there wasn't a method to get the ID as required to pass to Flutter.


cbracken (approved):
Much nicer! lgtm!

cbracken:
Looks like `FlTextureRegistrarTest.RegisterTexture` is failing.

cbracken (approved):
re-lgtm!

PR #40292, opened 2023-03-15
[Impeller] Simplify subpass branches; remove unused effect_matrix param

bdero:
Get rid of `Paint::WithFilters` `effect_matrix` parameter which should never be used.

Flatten out/simplify some branches in the subpass entity logic to make following it easier.


jonahwilliams (approved):
LGTM

ColdPaleLight (approved):
LGTM

PR #40309, opened 2023-03-15
[web] Remove image codecs from Canvaskit Chromium

mdebbar:
Size difference of CanvasKit Chromium:
| | Before | After | _Delta_ |
|-|-|-|-|
| **Gzip -9** | 1.489 MB | 1.406 MB | _85 KB_ |
| **Brotli -9** | 1.332 MB | 1.256 MB | _78 KB_ |

Ran the flutter gallery locally with this change. Visited many pages that contain images and text. Everything looks normal.

Closes https://github.com/flutter/flutter/issues/122331


lib/web_ui/lib/src/engine/safe_browser_api.dart
```diff
@@ -217,17 +217,18 @@ const bool _browserImageDecodingEnabled = bool.fromEnvironment(
...
 void debugResetBrowserSupportsImageDecoder() {
-  browserSupportsImageDecoder =
-      _imageDecoderConstructor != null;
+  browserSupportsImageDecoder = _defaultBrowserSupportsImageDecoder;
 }
 
+bool _defaultBrowserSupportsImageDecoder =
+    _browserImageDecodingEnabled &&
+    _imageDecoderConstructor != null &&
+    browserEngine == BrowserEngine.blink;
```
mdebbar, line 230: @yjbanov I made this change to make sure we always reset it to the correct default (it looks like at some point things got out of sync).

yjbanov: If it's not too much trouble, can you please move the `browserEngine == BrowserEngine.blink` into a constant with a TODO and a Github issue, something like this:

```dart
// TODO(yjbanov): $LINK_TO_GITHUB_ISSUE
// Frequently when a browser launches an API that other browsers already support
// there are subtle incompatibilities that may cause apps to crash if we blindly
// adopt the new implementation. This variable prevents us from picking up potentially
// incompatible implementations of ImagdeDecoder API. Instead, when a new browser
// engine launches the API, we'll evaluate it and enable it explicitly.
final bool _kProtectFromImageDecoderIncompatibilities = browserEngine == BrowserEngine.blink;
```

third_party/canvaskit/BUILD.gn
```diff
@@ -20,9 +26,15 @@ group("canvaskit_group") {
 # This toolchain is only to be used by canvaskit_chromium_group below.
 wasm_toolchain("canvaskit_chromium") {
   extra_toolchain_args = {
+    # Remove ICU data.
     skia_use_icu = false
     skia_use_client_icu = true
     skia_icu_bidi_third_party_dir = "//flutter/third_party/canvaskit/icu_bidi"
+
+    # Remove image codecs.
```
yjbanov, line 34: I would instead explain that in Chromium we use ImageDecoder instead of the wasm codecs.

tools/gn
```diff
@@ -617,11 +617,8 @@ def to_gn_wasm_args(args, gn_args):
   gn_args['skia_use_vulkan'] = False
   gn_args['skia_use_webgpu'] = False
   gn_args['skia_use_libheif'] = False
-  gn_args['skia_use_libjpeg_turbo_decode'] = True
   gn_args['skia_use_libjpeg_turbo_encode'] = False
-  gn_args['skia_use_libpng_decode'] = True
   gn_args['skia_use_libpng_encode'] = True
```
yjbanov, line 621: Can you please file an issue to remove this as well? We can implement this by piping the image through a 2D canvas.

yjbanov (approved):

lib/web_ui/lib/src/engine/safe_browser_api.dart
eyebrowsoffire: I think it would be nice if we differentiated between `_browserImageDecodingEnabled` explicitly being set and defaulting to true. That way, if it is explicitly set, it can override the `browserEngine == BrowserEngine.blink`, so that it is easy to test against browsers that have implemented the ImageDecoder API.

mdebbar: @eyebrowsoffire are you talking about overriding for tests or apps?

In tests, we already do override it e.g:
https://github.com/flutter/engine/blob/7a4542c76417441d899ab6b7f28ac9c00ef432dc/lib/web_ui/test/canvaskit/image_golden_test.dart#L64-L79

As for apps, do we really want users to play with this? Whoever wants to test this in different browsers likely already knows how to modify and build the engine, so they can just change the boolean value in the code.

eyebrowsoffire: I meant in apps. If we wanted to test the viability of this in a browser, I'd much rather just be able to pass in a flag and try out a bunch of different apps than to rebuild the engine entirely. If a user stumbles upon the flag and misuses it, it's not the end of the world IMO. Let's just document this as well as we can, and if a user files a bug saying "I turned this flag on and it doesn't work" then we just say "don't do that" and close the bug and move on.

lib/web_ui/lib/src/engine/safe_browser_api.dart
```diff
@@ -217,17 +217,26 @@ const bool _browserImageDecodingEnabled = bool.fromEnvironment(
...
+    _kProtectFromImageDecoderIncompatibilities;
+
+// TODO(yjbanov): https://github.com/flutter/flutter/issues/122761
+// Frequently, when a browser launches an API that other browsers already
+// support, there are subtle incompatibilities that may cause apps to crash if,
+// we blindly adopt the new implementation. This variable prevents us from
+// picking up potentially incompatible implementations of ImagdeDecoder API.
+// Instead, when a new browser engine launches the API, we'll evaluate it and
+// enable it explicitly.
+bool get _kProtectFromImageDecoderIncompatibilities => browserEngine == BrowserEngine.blink;
```
eyebrowsoffire, line 239: Complete nitpick, but I don't like the name of this variable. On its face, the name seems to imply the opposite of what it does. If something called `ProtectFromImageDecoderIncompatibilities` is true, I would think that means that you are protecting from incompatibilities, and if it's false, I would think you would not be protecting from incompatibilities. In general, either way, the name is a bit confusing as a name of a boolean. Might I suggest something like `_kImageDecoderIsStable` or `_kImageDecoderIsBlessed` or something like that?

eyebrowsoffire (approved):
LGTM! One nitpick.

lib/web_ui/lib/src/engine/safe_browser_api.dart
mdebbar: Agreed. I also removed the `k` prefix since these are getters and not consts anymore.

PR #40310, opened 2023-03-15
[web] Migrate DOM shim to JS types.

joshualitt:
This CL is the first in a small set of CLs to migrate the DOM shim to JS types.

### DIFF
--- a/lib/web_ui/lib/src/engine/dom.dart
+++ b/lib/web_ui/lib/src/engine/dom.dart
@@ -3,6 +3,7 @@
 // found in the LICENSE file.
 
 import 'dart:async';
+import 'dart:js_interop';
 import 'dart:math' as math;
 import 'dart:typed_data';
 
@@ -10,18 +11,58 @@ import 'package:js/js.dart';
 import 'package:js/js_util.dart' as js_util;
 import 'package:meta/meta.dart';
 
+import 'browser_detection.dart';
+
 /// This file contains static interop classes for interacting with the DOM and
 /// some helpers. All of the classes in this file are named after their
 /// counterparts in the DOM. To extend any of these classes, simply add an
 /// external method to the appropriate class's extension. To add a new class,
 /// simply name the class after it's counterpart in the DOM and prefix the
 /// class name with `Dom`.
-/// NOTE: After the new static interop DOM API is released in the Dart SDK,
-/// these classes will be replaced by typedefs.
 /// NOTE: Currently, optional parameters do not behave as expected.
 /// For the time being, avoid passing optional parameters directly to JS.
-/// NOTE: Please avoid typing external JS functions with `int`, as passing ints
-/// to and from JS is not supported on all web backends.
+
+// TODO(joshualitt): To make it clearer to users of this shim that this is where
+// the boundary between Dart and JS interop exists, we should expose JS types
+// directly to the engine.
+
+/// Conversions methods to facilitate migrating to JS types.
+///
+/// The existing behavior across the JS interop boundary involves many implicit
+/// conversions. For efficiency reasons, on JS backends we still want those
+/// implicit conversions, but on Wasm backends we need to 'shallowly' convert
+/// these types.
+///
+/// Note: Due to discrepancies between how `null`, `JSNull`, and `JSUndefined`
+/// are currently represented across web backends, these extensions should be
+/// used carefully and only on types that are known to not contains `JSNull` and
+/// `JSUndefined`.
+extension ObjectToJSAnyExtension on Object {
+  JSAny get toJSAnyShallow {
+    if (isWasm)  {
+      return toJSAnyDeep;
+    } else {
+      // TODO(joshualitt): remove this cast when we reify JS types on JS
+      // backends.
+      // ignore: unnecessary_cast
+      return this as JSAny;
+    }
+  }
+
+  JSAny get toJSAnyDeep => js_util.jsify(this) as JSAny;
+}
+
+extension JSAnyToObjectExtension on JSAny {
+  Object get toObjectShallow {
+    if (isWasm) {
+      return toObjectDeep;
+    } else {
+      return this;
+    }
+  }
+
+  Object get toObjectDeep => js_util.dartify(this)!;
+}
 
 @JS()
 @staticInterop
@@ -29,11 +70,22 @@ class DomWindow extends DomEventTarget {}
 
 extension DomWindowExtension on DomWindow {
   external DomConsole get console;
-  external double get devicePixelRatio;
+
+  @JS('devicePixelRatio')
+  external JSNumber get _devicePixelRatio;
+  double get devicePixelRatio => _devicePixelRatio.toDart;
+
   external DomDocument get document;
   external DomHistory get history;
-  external double? get innerHeight;
-  external double? get innerWidth;
+
+  @JS('innerHeight')
+  external JSNumber? get _innerHeight;
+  double? get innerHeight => _innerHeight?.toDart;
+
+  @JS('innerWidth')
+  external JSNumber? get _innerWidth;
+  double? get innerWidth => _innerWidth?.toDart;
+
   external DomLocation get location;
   external DomNavigator get navigator;
   external DomVisualViewport? get visualViewport;
@@ -50,41 +102,84 @@ extension DomWindowExtension on DomWindow {
     );
   }
 
+  @JS('fetch')
+  external JSPromise _fetch1(JSString url);
+  @JS('fetch')
+  external JSPromise _fetch2(JSString url, JSAny headers);
+
   // ignore: non_constant_identifier_names
   external DomURL get URL;
-  external bool dispatchEvent(DomEvent event);
-  external DomMediaQueryList matchMedia(String? query);
-  DomCSSStyleDeclaration getComputedStyle(DomElement elt,
-          [String? pseudoElt]) =>
-      js_util.callMethod(this, 'getComputedStyle', <Object>[
-        elt,
-        if (pseudoElt != null) pseudoElt
-      ]) as DomCSSStyleDeclaration;
+
+  @JS('dispatchEvent')
+  external JSBoolean _dispatchEvent(DomEvent event);
+  bool dispatchEvent(DomEvent event) => _dispatchEvent(event).toDart;
+
+  @JS('matchMedia')
+  external DomMediaQueryList _matchMedia(JSString? query);
+  DomMediaQueryList matchMedia(String? query) => _matchMedia(query?.toJS);
+
+  @JS('getComputedStyle')
+  external DomCSSStyleDeclaration _getComputedStyle1(DomElement elt);
+  @JS('getComputedStyle')
+  external DomCSSStyleDeclaration _getComputedStyle2(
+      DomElement elt, JSString pseudoElt);
+  DomCSSStyleDeclaration getComputedStyle(DomElement elt, [String? pseudoElt]) {
+    if (pseudoElt == null) {
+      return _getComputedStyle1(elt);
+    } else {
+      return _getComputedStyle2(elt, pseudoElt.toJS);
+    }
+  }
+
   external DomScreen? get screen;
-  external double requestAnimationFrame(DomRequestAnimationFrameCallback callback);
+
+  @JS('requestAnimationFrame')
+  external JSNumber _requestAnimationFrame(JSFunction callback);
+  double requestAnimationFrame(DomRequestAnimationFrameCallback callback) =>
+      _requestAnimationFrame(callback.toJS).toDart;
+
+  @JS('postMessage')
+  external JSVoid _postMessage1(JSAny message, JSString targetOrigin);
+  @JS('postMessage')
+  external JSVoid _postMessage2(
+      JSAny message, JSString targetOrigin, JSArray messagePorts);
   void postMessage(Object message, String targetOrigin,
-          [List<DomMessagePort>? messagePorts]) =>
-      js_util.callMethod(this, 'postMessage', <Object?>[
-        message,
-        targetOrigin,
-        if (messagePorts != null) js_util.jsify(messagePorts)
-      ]);
+      [List<DomMessagePort>? messagePorts]) {
+    if (messagePorts == null) {
+      _postMessage1(message.toJSAnyShallow, targetOrigin.toJS);
+    } else {
+      _postMessage2(
+          message.toJSAnyShallow,
+          targetOrigin.toJS,
+          // Cast is necessary so we can call `.toJS` on the right extension.
+          // ignore: unnecessary_cast
+          (messagePorts as List<JSAny>).toJS);
+    }
+  }
 
   /// The Trusted Types API (when available).
   /// See: https://developer.mozilla.org/en-US/docs/Web/API/Trusted_Types_API
   external DomTrustedTypePolicyFactory? get trustedTypes;
 }
 
-typedef DomRequestAnimationFrameCallback = void Function(num highResTime);
+typedef DomRequestAnimationFrameCallback = void Function(JSNumber highResTime);
 
 @JS()
 @staticInterop
 class DomConsole {}
 
 extension DomConsoleExtension on DomConsole {
-  external void warn(Object? arg);
-  external void error(Object? arg);
-  external void debug(Object? arg);
+  @JS('warn')
+  external JSVoid _warn(JSString? arg);
+  void warn(Object? arg) => _warn(arg.toString().toJS);
+
+  @JS('error')
+  external JSVoid _error(JSString? arg);
+  void error(Object? arg) => _error(arg.toString().toJS);
+
+  @JS('debug')
+  external JSVoid _debug(JSString? arg);
+  void debug(Object? arg) => _debug(arg.toString().toJS);
 }
 
 @JS('window')
@@ -102,13 +197,32 @@ class DomNavigator {}
 
 extension DomNavigatorExtension on DomNavigator {
   external DomClipboard? get clipboard;
-  external double? get maxTouchPoints;
-  external String get vendor;
-  external String get language;
-  external String? get platform;
-  external String get userAgent;
-  List<String>? get languages =>
-      js_util.getProperty<List<Object?>?>(this, 'languages')?.cast<String>();
+
+  @JS('maxTouchPoints')
+  external JSNumber? get _maxTouchPoints;
+  double? get maxTouchPoints => _maxTouchPoints?.toDart;
+
+  @JS('vendor')
+  external JSString get _vendor;
+  String get vendor => _vendor.toDart;
+
+  @JS('language')
+  external JSString get _language;
+  String get language => _language.toDart;
+
+  @JS('platform')
+  external JSString? get _platform;
+  String? get platform => _platform?.toDart;
+
+  @JS('userAgent')
+  external JSString get _userAgent;
+  String get userAgent => _userAgent.toDart;
+
+  @JS('languages')
+  external JSArray? get _languages;
+  List<String>? get languages => _languages?.toDart
+      .map<String>((JSAny? any) => (any! as JSString).toDart)
+      .toList();
 }
 
 @JS()
@@ -117,23 +231,54 @@ class DomDocument extends DomNode {}
 
 extension DomDocumentExtension on DomDocument {
   external DomElement? get documentElement;
-  external DomElement? querySelector(String selectors);
+
+  @JS('querySelector')
+  external DomElement? _querySelector(JSString selectors);
+  DomElement? querySelector(String selectors) => _querySelector(selectors.toJS);
+
+  @JS('querySelectorAll')
+  external _DomList _querySelectorAll(JSString selectors);
   Iterable<DomElement> querySelectorAll(String selectors) =>
-      createDomListWrapper<DomElement>(js_util
-          .callMethod<_DomList>(this, 'querySelectorAll', <Object>[selectors]));
-  DomElement createElement(String name, [Object? options]) =>
-      js_util.callMethod(this, 'createElement',
-          <Object>[name, if (options != null) options]) as DomElement;
-  external bool execCommand(String commandId);
+      createDomListWrapper<DomElement>(_querySelectorAll(selectors.toJS));
+
+  @JS('createElement')
+  external DomElement _createElement1(JSString name);
+  @JS('createElement')
+  external DomElement _createElement2(JSString name, JSAny? options);
+  DomElement createElement(String name, [Object? options]) {
+    if (options == null) {
+      return _createElement1(name.toJS);
+    } else {
+      return _createElement2(name.toJS, options.toJSAnyDeep);
+    }
+  }
+
+  @JS('execCommand')
+  external JSBoolean _execCommand(JSString commandId);
+  bool execCommand(String commandId) => _execCommand(commandId.toJS).toDart;
+
   external DomHTMLScriptElement? get currentScript;
-  external DomElement createElementNS(
-      String namespaceURI, String qualifiedName);
-  external DomText createTextNode(String data);
-  external DomEvent createEvent(String eventType);
+
+  @JS('createElementNS')
+  external DomElement _createElementNS(
+      JSString namespaceURI, JSString qualifiedName);
+  DomElement createElementNS(String namespaceURI, String qualifiedName) =>
+      _createElementNS(namespaceURI.toJS, qualifiedName.toJS);
+
+  @JS('createTextNode')
+  external DomText _createTextNode(JSString data);
+  DomText createTextNode(String data) => _createTextNode(data.toJS);
+
+  @JS('createEvent')
+  external DomEvent _createEvent(JSString eventType);
+  DomEvent createEvent(String eventType) => _createEvent(eventType.toJS);
+
   external DomElement? get activeElement;
+
+  @JS('elementFromPoint')
+  external DomElement? _elementFromPoint(JSNumber x, JSNumber y);
   DomElement? elementFromPoint(int x, int y) =>
-      js_util.callMethod<DomElement?>(this, 'elementFromPoint',
-          <Object>[x.toDouble(), y.toDouble()]);
+      _elementFromPoint(x.toJS, y.toJS);
 }
 
 @JS()
@@ -144,13 +289,25 @@ extension DomHTMLDocumentExtension on DomHTMLDocument {
   external DomFontFaceSet? get fonts;
   external DomHTMLHeadElement? get head;
   external DomHTMLBodyElement? get body;
-  external set title(String? value);
-  external String? get title;
+
+  @JS('title')
+  external set _title(JSString? value);
+  set title(String? value) => _title = value?.toJS;
+
+  @JS('title')
+  external JSString? get _title;
+  String? get title => _title?.toDart;
+
+  @JS('getElementsByTagName')
+  external _DomList _getElementsByTagName(JSString tag);
   Iterable<DomElement> getElementsByTagName(String tag) =>
-      createDomListWrapper<DomElement>(js_util
-          .callMethod<_DomList>(this, 'getElementsByTagName', <Object>[tag]));
+      createDomListWrapper<DomElement>(_getElementsByTagName(tag.toJS));
+
   external DomElement? get activeElement;
-  external DomElement? getElementById(String id);
+
+  @JS('getElementById')
+  external DomElement? _getElementById(JSString id);
+  DomElement? getElementById(String id) => _getElementById(id.toJS);
 }
 
 @JS('document')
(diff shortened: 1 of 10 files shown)


lib/web_ui/lib/src/engine/dom.dart
```diff
@@ -1809,14 +2899,18 @@ DomHTMLIFrameElement createDomHTMLIFrameElement() =>
...
-  external void start();
+  @JS('postMessage')
+  external JSVoid _postMessage(JSAny? message);
+  void postMessage(Object? message) => _postMessage(message.toJSAnyShallow);
+
+  external JSVoid start();
 }
 
-@JS()
+@JS('MesageChannel')
```
eyebrowsoffire, line 2909: This is misspelled, which I think is causing the unit test failures

eyebrowsoffire (changes requested):

lib/web_ui/lib/src/engine/dom.dart
joshualitt: Good catch!

joshualitt:
@ditman @eyebrowsoffire @mdebbar , this is basically ready for review. There may be some bugs and back and forth while I do some framework testing, but this CL is large, and we want to land it as quickly as possible so we can unblock other efforts, so it makes sense to start the review ASAP.

These will be the patterns for JS interop we will use until we have inline classes, which will be later this year. That migration will be much smaller than this one as it just involves changing how the classes are defined.

Some things to note:

1. Some places Wasm needs to convert objects, whereas JS backends don't. This logic is hidden behind the 'deep' / 'shallow' conversions in the dom shim. Sorry about the naming scheme. I'm open to alternatives, but I'm not sure there are any good names in this space as the underlying concept we are trying to name(i.e. conversions required on JS backends but not on Wasm backends) is quite vague.
2. Almost all of this migration logic was hidden safely behind the shim. The exceptions are functions. With `JSFunction` we will no longer have the problem where people forget to wrap Dart functions when the flow to JS. However, as you can see in this CL there is some subtlety here. I left the event listeners alone for now, and tried to touch only those functions which get 'thrown over the wall'.
3. I tried to start getting rid of uses of `js_util` but I didn't push this very hard in this CL, with one notable exception. We were using `js_util` to handle optional arguments. I changed these usages to trampolines. However, we've fixed optional arguments to JS functions. Now they have JS semantics(i.e. not passed is `undefined`), which is quite nice. Unfortunately because this logic doesn't work for tearoffs, I figured we should discuss that change and possibly do it in a follow on if we wish. I will push further on getting rid of as many usages of `js_util` as possible in the future, though some usages may be unavoidable.
4. Please let me know if you'd like more comments, or have any questions / concerns.

joshualitt:
@srujzs FYI. Please feel free to leave comments.

lib/web_ui/lib/src/engine/initialization.dart
```diff
@@ -185,7 +186,7 @@ Future<void> initializeEngineServices({
         // milliseconds as a double value, with sub-millisecond information
         // hidden in the fraction. So we first multiply it by 1000 to uncover
         // microsecond precision, and only then convert to `int`.
-        final int highResTimeMicroseconds = (1000 * highResTime).toInt();
+        final int highResTimeMicroseconds = (1000 * highResTime.toDart).toInt();
```
mdebbar, line 189: Does this throw a compile-time error if I forget to do `.toDart`? (I assume yes, but would like to confirm).

lib/web_ui/lib/src/engine/platform_dispatcher.dart
```diff
@@ -934,17 +935,17 @@ class EnginePlatformDispatcher extends ui.PlatformDispatcher {
...
     const String styleAttribute = 'style';
 
-    _fontSizeObserver = createDomMutationObserver(allowInterop(
-        (List<dynamic> mutations, DomMutationObserver _) {
-      for (final dynamic mutation in mutations) {
-        final DomMutationRecord record = mutation as DomMutationRecord;
+    _fontSizeObserver = createDomMutationObserver(
+        (JSArray mutations, DomMutationObserver _) {
+      for (final DomMutationRecord mutation in mutations.toDart.cast<DomMutationRecord>()) {
+        final DomMutationRecord record = mutation;
```
mdebbar, line 941: This is not a big deal (perf-wise) but I would like to understand more.

Is it possible to avoid the `.cast<...>()` on the list and keep the `as`? E.g.
```suggestion
      for (final dynamic mutation in mutations.toDart) {
        final DomMutationRecord record = mutation as DomMutationRecord;
```
Does `.toDart` here return a `List<dynamic>`?

lib/web_ui/lib/src/engine/dom.dart
```diff
@@ -102,13 +193,32 @@ class DomNavigator {}
...
-  external String get vendor;
-  external String get language;
-  external String? get platform;
-  external String get userAgent;
-  List<String>? get languages =>
-      js_util.getProperty<List<Object?>?>(this, 'languages')?.cast<String>();
+
+  @JS('maxTouchPoints')
+  external JSNumber? get _maxTouchPoints;
+  double? get maxTouchPoints => _maxTouchPoints?.toDart;
```
mdebbar, line 199: With this new approach, it looks like we can start using `int` again? (maybe with a `.toInt()` at the end there)

mdebbar (commented):
Overall, the change LGTM if it LGT @eyebrowsoffire. I just have a few questions to understand things better.

lib/web_ui/lib/src/engine/initialization.dart
joshualitt: It will on Dart2Wasm right away, and it will on JS backends after https://dart-review.googlesource.com/c/sdk/+/283742 lands, which will hopefully be soon.

lib/web_ui/lib/src/engine/platform_dispatcher.dart
joshualitt: `toDart` in this case returns `List<JSAny?>`. To be honest, we really don't need the cast list. We can iterate over the `JSAny?` and then cast them individually. Do you prefer that approach?

lib/web_ui/lib/src/engine/dom.dart
joshualitt: Yes, we can start using `int` again now that we are handling the conversions at the boundary.

The rule going forward is 'if the function is external, use JS types`, and beyond that we can do whatever makes the most sense. Once I migrate the framework and `package:test` then this rule will be enforced at compile time in Dart2Wasm, but for the next few weeks we just have to be careful. It won't break anything to violate this rule, it will just make flipping the 'strict mode' flag for Dart2Wasm a bit harder.

We can do a cleanup pass after this to make the engine facing side of the DOM shim more ergonomic.

lib/web_ui/lib/src/engine/dom.dart
```diff
@@ -1739,27 +2807,29 @@ extension DomScreenExtension on DomScreen {
...
-    final Object jsResult = js_util.callMethod<Object>(this, 'lock', <String>[orientation]);
-    return js_util.promiseToFuture(jsResult);
-  }
-  external void unlock();
+  @JS('lock')
+  external JSPromise _lock(JSString orientation);
+  Future<dynamic> lock(String orientation) =>
+      js_util.promiseToFuture(_lock(orientation.toJS));
+
+  external JSVoid unlock();
```
ditman, line 2815: Why is `JSVoid` needed? I thought `void` would be enough? Can you do `unlock().toDart` ?

lib/web_ui/lib/src/engine/dom.dart
```diff
@@ -1980,23 +3078,20 @@ const Set<String> _expectedFilesForTT = <String>{
...
-
-        return null;
-      },
-    ),
-  ),
+      // Validates the given [url].
+      createScriptURL: (JSString url) {
+    final Uri uri = Uri.parse(url.toDart);
+    if (_expectedFilesForTT.contains(uri.pathSegments.last)) {
+      return uri.toString().toJS;
```
ditman, line 3087: Good catch that this needs to return a JS value!

lib/web_ui/lib/src/engine/dom.dart
```diff
@@ -2150,28 +3233,33 @@ class DomIntl {}
...
   ///
   /// See: https://code.google.com/archive/p/v8-i18n/wikis/BreakIterator.wiki
-  external Object? get v8BreakIterator;
+  external JSAny? get v8BreakIterator;
 }
 
 /// Similar to [js_util.callMethod] but allows for providing a non-string method
 /// name.
-T _jsCallMethod<T>(Object o, Object method, [List<Object?> args = const <Object?>[]]) {
+T _jsCallMethod<T>(Object o, Object method,
```
ditman, line 3246: Should this be hoisted to `js_utils`?

lib/web_ui/lib/src/engine/font_change_util.dart
```diff
@@ -23,13 +22,13 @@ FutureOr<void> sendFontChangeMessage() async {
   if (!_fontChangeScheduled) {
     _fontChangeScheduled = true;
     // Batch updates into next animationframe.
-    domWindow.requestAnimationFrame(allowInterop((num _) {
+    domWindow.requestAnimationFrame((_) {
```
ditman, line 25: Not needing allowInterop here is pretty nice, NGL. If you were attempting to make the line between Dart and JS more clear, this is probably not helping :P

lib/web_ui/lib/src/engine/initialization.dart
```diff
@@ -174,7 +175,7 @@ Future<void> initializeEngineServices({
     // fires.
     if (!waitingForAnimation) {
       waitingForAnimation = true;
-      domWindow.requestAnimationFrame(allowInterop((num highResTime) {
+      domWindow.requestAnimationFrame((JSNumber highResTime) {
```
ditman, line 178: In other JS-interop layers, I've tried to prevent the JSNumber from leaking to users by wrapping the user-supplied callback in something that gets called from the js-interop layer after making the transformation?

Something like this maybe?

```
  double requestAnimationFrame(DomRequestAnimationFrameCallback callback) =>
      _requestAnimationFrame((JSNum n) {
        callback(n.toDart);
      }.toJS).toDart;
```

lib/web_ui/lib/src/engine/platform_dispatcher.dart
```diff
@@ -934,17 +935,17 @@ class EnginePlatformDispatcher extends ui.PlatformDispatcher {
   void _addFontSizeObserver() {
     const String styleAttribute = 'style';
 
-    _fontSizeObserver = createDomMutationObserver(allowInterop(
-        (List<dynamic> mutations, DomMutationObserver _) {
-      for (final dynamic mutation in mutations) {
-        final DomMutationRecord record = mutation as DomMutationRecord;
+    _fontSizeObserver = createDomMutationObserver(
+        (JSArray mutations, DomMutationObserver _) {
+      for (final DomMutationRecord mutation in mutations.toDart.cast<DomMutationRecord>()) {
```
ditman, line 940: I think this odd cast here is needed because of a sub-optimal JS-interop layer; I've re-written this recently like so:

```
DomMutationObserver createDomMutationObserver(DomMutationCallbackFn fn) =>
    domCallConstructorString('MutationObserver', <Object>[
      allowInterop(
        (List<dynamic> mutations, DomMutationObserver observer) {
          fn(mutations.cast<DomMutationRecord>(), observer);
        },
      )
    ])! as DomMutationObserver;
```

(same trick as before, really)

ditman (approved):
This is great! If everybody else's happy, and tests pass; I'm happy!

Also I've read in a comment to Mouad that we have rules that are (going to be) automatically enforceable when compiling libraries, so hit it!

lib/web_ui/lib/src/engine/dom.dart
```diff
@@ -50,41 +98,84 @@ extension DomWindowExtension on DomWindow {
...
   external DomScreen? get screen;
-  external double requestAnimationFrame(DomRequestAnimationFrameCallback callback);
+
+  @JS('requestAnimationFrame')
+  external JSNumber _requestAnimationFrame(JSFunction callback);
+  double requestAnimationFrame(DomRequestAnimationFrameCallback callback) =>
+      _requestAnimationFrame(callback.toJS).toDart;
+
+  @JS('postMessage')
+  external JSVoid _postMessage1(JSAny message, JSString targetOrigin);
```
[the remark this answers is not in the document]
joshualitt: Honestly, it was kind of the 'Nth' variation. Do you have a preference that it be the number of arguments? I do not have strong feelings, and I'm happy to change it.

lib/web_ui/lib/src/engine/dom.dart
```diff
@@ -50,41 +98,84 @@ extension DomWindowExtension on DomWindow {
...
+      JSAny message, JSString targetOrigin, JSArray messagePorts);
   void postMessage(Object message, String targetOrigin,
-          [List<DomMessagePort>? messagePorts]) =>
-      js_util.callMethod(this, 'postMessage', <Object?>[
-        message,
-        targetOrigin,
-        if (messagePorts != null) js_util.jsify(messagePorts)
-      ]);
+      [List<DomMessagePort>? messagePorts]) {
+    if (messagePorts == null) {
```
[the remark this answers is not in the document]
joshualitt: Yea, it will be quite a bit faster to use static interop on Dart2Wasm. However, we can switch to using optional arguments, but then tearoffs won't work as expected.

If the consensus is this is okay, then we can do that change in a follow on. Hopefully in the nearish future I can convince everyone to make tearoffs of external functions a static error.

lib/web_ui/lib/src/engine/dom.dart
```diff
@@ -50,41 +98,84 @@ extension DomWindowExtension on DomWindow {
...
+  external JSVoid _warn(JSString? arg);
+  void warn(Object? arg) => _warn(arg.toString().toJS);
+
+  @JS('error')
+  external JSVoid _error(JSString? arg);
+  void error(Object? arg) => _error(arg.toString().toJS);
+
+  @JS('debug')
+  external JSVoid _debug(JSString? arg);
+  void debug(Object? arg) => _debug(arg.toString().toJS);
```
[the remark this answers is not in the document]
joshualitt: I've proposed `JSVarArgs` which would take a `List` literal. Maybe this is something I can push on after inline classes are out the door, and maybe that will be the impetus for adding var args to Dart? One can dream.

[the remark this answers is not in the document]
joshualitt: We can do like N optional arguments in the meantime, but maybe in a follow on?

lib/web_ui/lib/src/engine/dom.dart
```diff
@@ -211,27 +404,50 @@ DomEvent createDomEvent(String type, String name) {
 @JS('ProgressEvent')
 @staticInterop
 class DomProgressEvent extends DomEvent {
-  external factory DomProgressEvent(String type);
+  factory DomProgressEvent(String type) => DomProgressEvent._(type.toJS);
+  external factory DomProgressEvent._(JSString type);
```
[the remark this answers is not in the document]
joshualitt: It uses the `@JS` name for all constructors. It's a neat trick for sure.

lib/web_ui/lib/src/engine/dom.dart
```diff
@@ -3,37 +3,85 @@
...
+      // ignore: unnecessary_cast
+      return this as JSAny;
+    }
+  }
+
+  JSAny get toJSAnyDeep => js_util.jsify(this) as JSAny;
+}
+
+extension JSAnyToObjectExtension on JSAny {
+  Object get toObjectShallow {
```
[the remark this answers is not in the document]
joshualitt: For sanity sake, it does not, because we can't do better than users can do in their own code. Ideally we'd know what types flow to / from where,  and then we could write custom conversion code, perhaps writing our little JS union types? If not, then `dartify` is the answer, but I made `dartify` / `jsify` faster for native types, so it shouldn't be any worse than a giant `if` / `else if` block that we would have to use if we didn't know the JS type we were converting from.

lib/web_ui/lib/src/engine/dom.dart
```diff
@@ -651,13 +1061,17 @@ extension DomCanvasElementExtension on DomCanvasElement {
 class WebGLContext {}
 
 extension WebGLContextExtension on WebGLContext {
-  external int getParameter(int value);
+  @JS('getParameter')
+  external JSNumber _getParameter(JSNumber value);
+  int getParameter(int value) => _getParameter(value.toJS).toDart.toInt();
```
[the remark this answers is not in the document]
joshualitt: The hope is that someday these types won't be so hidden, at least not all of them. We'd consciously choose to let some of these types flow directly into the engine itself. For example, if a `JSNumber` is going to flow right back into JS, then we may as well leave it as a `JSNumber`. Obviously, in the case of a single `JSNumber` this isn't too bad, but then there is `JSArray`...

lib/web_ui/lib/src/engine/dom.dart
```diff
@@ -802,22 +1410,16 @@ Future<HttpFetchResponse> httpFetch(String url) async {
...
+Future<_DomResponse> _rawHttpPost(String url, String data) =>
+    js_util.promiseToFuture<_DomResponse>(domWindow._fetch2(
+        url.toJS,
+        <String, Object?>{
+          'method': 'POST',
+          'headers': <String, Object?>{
+            'Content-Type': 'text/plain',
+          },
+          'body': data,
+        }.toJSAnyDeep));
```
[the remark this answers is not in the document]
joshualitt: hah, I can see how this might be very confusing, because it is confusing :(.

`toJSAnyShallow` only converts a few type, notably primitives, typed data, and `List`s, but not, for example, `Map`s. For those cases we have to go deep.

lib/web_ui/lib/src/engine/dom.dart
```diff
@@ -1189,53 +1813,90 @@ class DomException {
...
-  return domCallConstructorString(
-      'DOMRect', <Object>[left, top, width, height])! as DomRect;
+  return DomRect(left.toJS, top.toJS, width.toJS, height.toJS);
 }
 
-@JS()
+@JS('DOMRect')
 @staticInterop
-class DomRect extends DomRectReadOnly {}
+class DomRect extends DomRectReadOnly {
```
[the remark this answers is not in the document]
joshualitt: lol, yes, yes it should...let me just go ahead and resolve this unless you have strong feelings.

lib/web_ui/lib/src/engine/dom.dart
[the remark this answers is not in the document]
ditman: The first time I looked at syntax like this was when people were poking fun at Scala's [`Function22` Trait](https://www.scala-lang.org/api/2.12.9/scala/Function22.html) (a Function that accepts 22 parameters), and I think I mapped the X in `FunctionX` to "X number of parameters" rather than the Nth variation. I don't have strong feelings at all (as long as we're consistent). Also if we ever wanted to "fix" this, it'd be a fully internal rename, so I don't think it matters that much :P

lib/web_ui/lib/src/engine/dom.dart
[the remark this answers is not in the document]
ditman: > We can do like N optional arguments in the meantime, but maybe in a follow on?

If it entails defining N `console._logN` functions and then calling "the right one" depending on the number of arguments: not even in a follow on, leave this as is :P

If it's `console.log(Object? arg, [Object? arg2, Object? arg3, Object? arg4, Object? arg5])` then yes, I'd like to have like... 10 args :P (or 22!)

lib/web_ui/lib/src/engine/dom.dart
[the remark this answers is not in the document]
ditman: TL;DR: Leave this as is, until somebody gets super offended and does the find-and-replace. It might even be you in a few months! :P

lib/web_ui/lib/src/engine/dom.dart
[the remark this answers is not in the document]
ditman: TIL and my mind is blown.

lib/web_ui/lib/src/engine/dom.dart
```diff
@@ -1116,52 +1721,69 @@ class HttpFetchError implements Exception {
 class _DomResponse {}
 
 extension _DomResponseExtension on _DomResponse {
-  external double get status;
+  @JS('status')
+  external JSNumber get _status;
+  int get status => _status.toDart.toInt();
```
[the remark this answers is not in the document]
ditman: (I've seen a similar question asked by @mdebbar in his review, ssssskip!)

lib/web_ui/lib/src/engine/dom.dart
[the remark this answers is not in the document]
ditman: Aha, to me, "depth" meant... "recursive" transform, rather than a "shallow" transform (which only transforms "the first level"). I guess I associated it with shallow/deep cloning.

I don't have a better name for this, though :/

lib/web_ui/lib/src/engine/dom.dart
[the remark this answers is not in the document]
ditman: Very weak feelings wrt renaming potentially massive amounts of code. Please continue, nothing to see here. 👮

lib/web_ui/lib/src/engine/dom.dart
```diff
@@ -1268,151 +1938,273 @@ DomHTMLTextAreaElement createDomHTMLTextAreaElement() =>
...
+  String get href => _href.toDart;
 }
 
-@JS()
+@JS('PopStateEvent')
 @staticInterop
-class DomPopStateEvent extends DomEvent {}
+class DomPopStateEvent extends DomEvent {
+  external factory DomPopStateEvent.arg1(JSString type);
+  external factory DomPopStateEvent.arg2(JSString type, JSAny initDict);
```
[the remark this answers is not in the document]
ditman: They DO all resolve to the same `@JS('ConstructorName')`! (answered elsewhere, skkkkip!)

lib/web_ui/lib/src/engine/dom.dart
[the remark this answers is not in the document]
joshualitt: I mean the latter, we can use optional args and let the new fancy not passed == JS not passed  logic take care of the details. Anyway, punting for now.

lib/web_ui/lib/src/engine/dom.dart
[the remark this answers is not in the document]
joshualitt: Yes, `JSNumber.toDart`  is a double. We've added `int.toJS` which does the right thing from the Dart side though. Anyway, in all of these cases we can continue to use a trampoline if the `int` quality of the return value is important.

lib/web_ui/lib/src/engine/dom.dart
```diff
@@ -1096,7 +1701,7 @@ class HttpFetchNoPayloadError implements Exception {
 /// report, such as 200, 404, or 500.
 class HttpFetchError implements Exception {
   /// Creates an exception from a failed HTTP request.
-  HttpFetchError(this.url, { required this.requestError });
+  HttpFetchError(this.url, {required this.requestError});
```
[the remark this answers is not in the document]
joshualitt: hah yes, sorry about that. I wasn't about to manually format this file though. Hopefully we can continue to use `dart format` on this file going forward.

lib/web_ui/lib/src/engine/dom.dart
```diff
@@ -1268,151 +1938,273 @@ DomHTMLTextAreaElement createDomHTMLTextAreaElement() =>
...
-      js_util.callMethod(this, 'readText', <Object>[]));
+  @JS('readText')
+  external JSPromise _readText();
+  Future<String> readText() => js_util.promiseToFuture<String>(_readText());
 
-  Future<dynamic> writeText(String data) => js_util
-      .promiseToFuture(js_util.callMethod(this, 'writeText', <Object>[data]));
+  @JS('writeText')
+  external JSPromise _writeText(JSString data);
+  Future<dynamic> writeText(String data) =>
```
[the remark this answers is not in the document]
joshualitt: oof yea, let's definitely clean this up, but not in this CL.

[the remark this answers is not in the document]
joshualitt: to be clear, I'm just really trying to minimize changes that might affect logic. It is a bit scary to return `dynamic` here though, ugh.

lib/web_ui/lib/src/engine/dom.dart
joshualitt: lol, so `JSVoid` is currently just a typedef for `void`. But we all know that JS actually returns `undefined` in all cases. I haven't yet decided if we're going to support this. However, I'm going to make `JSVoid` an actual type that gets erased to `void` in the nearish future. Why? Because `void` is a Dart type, not a JS type. Technically some JS types are Dart types, but only because Dart doesn't give us any other option. Ohhh for a `Top` type someday.

I digress, JS has no notion of `void`. We provide `JSVoid` so users can signal their intent clearly. For the time being there is no `JSVoid.toDart`, nor `void.toJS`. The major benefit of not using `void` is that it allows us to completely disentangle JS types from Dart types, and it ensures that every single type on an external signature is a JS type.

I know this may seem painful now, but it is possible that the existence of `JSVoid` may enable optimization opportunities. For example, maybe it *is* more efficient to actually return `undefined` from JS. Is this true? I have no idea, but if it is, we would need `JSVoid`. Having it in place now ensures we're prepared for a possible future where it is valuable, and we can always revisit the decision down the road if we like. Personally, I think there is an nice symmetry to ensuring every type on an external function is a JS type, but ymmv.

lib/web_ui/lib/src/engine/font_change_util.dart
joshualitt: Well, we still need `allowInterop`, but right now its happening behind the DOM shim. `requestAnimationFrame` isn't a JS function, it's a Dart function that calls `.toJS` on the function passed in. Also, `allowInterop` returns a Dart function so the line was pretty blurry even before this.

Look, honestly, if it were up to me I'd pepper the entire engine codebase with `toJS` / `toDart` and the DOM shim would just be external functions. But, this isn't my call, and honestly its probably easier for the migration to keep things hidden behind the shim when possible. I'm hoping in the future we do expose JS types beyond the shim so that it's clear when looking at code like this where we're calling out to JS.

However, in this particular case, would you prefer we move the `toJS` on Dart functions to the caller? I'm okay with that change.

lib/web_ui/lib/src/engine/initialization.dart
joshualitt: oh we can do that. It does mean an extra closure. Do you prefer that approach?

lib/web_ui/lib/src/engine/platform_dispatcher.dart
joshualitt: See above, I'm okay with this pattern, but it does mean an extra closure. Do you prefer the marginal amount of overhead  and slightly more ergonomics, vs this approach?

lib/web_ui/lib/src/engine/dom.dart
joshualitt: Yes, just have to land https://dart-review.googlesource.com/c/sdk/+/288641. Hopefully sometime early next week.

lib/web_ui/lib/src/engine/initialization.dart
ditman: This particular case is called very often, so the faster, the better (so I guess: the fewer indirections, the better?)

lib/web_ui/lib/src/engine/font_change_util.dart
ditman: Nope, I think this is cool as is. I'm surprised the lint of "add types everywhere" is letting "_" slide without types, but I guess if it doesn't matter, it doesn't matter!

lib/web_ui/lib/src/engine/platform_dispatcher.dart
ditman: I think we can leave this as is now, and eventually make it "nice" in package:web. Weak preference on this one.

lib/web_ui/lib/src/engine/platform_dispatcher.dart
mdebbar: Since this code path isn't performance-critical, I don't have a preference either way.

But for hot paths, I would prefer to avoid the `.cast<...>()` call because it creates an unnecessary additional iterator.

lib/web_ui/test/text/font_collection_test.dart
```diff
@@ -165,9 +165,9 @@ void testMain() {
         await fontManager.downloadAllFonts();
         fontManager.registerDownloadedFonts();
         domDocument.fonts!
-            .forEach(allowInterop((DomFontFace f, DomFontFace f2, DomFontFaceSet s) {
+            .forEach((DomFontFace f, DomFontFace f2, DomFontFaceSet s) {
```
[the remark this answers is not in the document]
srujzs: Woops, the large diff didn't load for me. Ignore.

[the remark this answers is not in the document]
joshualitt: Hah yea, sorry. I moved the `toJS` internal to the shim.

The really wonderful super nice thing about the direction we are going is that soon we won't even have to ask questions like this because Dart functions won't even be able to flow to JS unless someone calls `toJS` *somewhere* in the codebase. First we have to get to strict mode...

JS types++

lib/web_ui/lib/src/engine/platform_dispatcher.dart
joshualitt: Yea, I'm pretty torn on this TBH. On the one hand it's very hard to imagine the cost of the additional wrapper object could ever be significant, and it is certainly very ergonomic. But on the other hand, sometimes premature optimization is the only way to prevent 'death by a thousand cuts' from a performance perspective. Every code base is different.

It almost certainly does not matter in this case, but I've changed this to the iterate and down cast approach in order to document that pattern in the engine code base.

lib/web_ui/lib/src/engine/dom.dart
```diff
@@ -2231,52 +3323,66 @@ class DomSegment {}
...
   }
 
-  return js_util.callConstructor<DomSegmenter>(
-    segmenterConstructor,
-    <Object?>[
-      <String>[],
-      js_util.jsify(<String, String>{'granularity': granularity}),
-    ],
+  return DomSegmenter(
+    <JSAny?>[].toJS,
```
eyebrowsoffire, line 3348: This can also be `JSUndefined`, if that's possible here. I switched it to an empty array previously because there was no way for me to pass `JSUndefined` into the constructor without it round-tripping through and becoming null. If `JSUndefined` would be `undefined` all the way down, I think it would be better to pass that than an empty array.

eyebrowsoffire (approved):
Looks good with a minor nitpick. I also wanted to share a few non-blocking thoughts on how this is shaping up.

1) One thing that bothers me a little bit is `JSNumber.toDart`, which secretly means "double". I think a `toDart` function is perfectly fine for things where their translation is immediately obvious, like `JSString -> String`. But for things that are actually ambiguous like `JSNumber`, I sort of wish it had `toInt` and `toDouble` but no `toDart`. I understand that `toDart` is actually on `Object` (I think?) so maybe this is not that tractable, but I thought I'd share my general feeling about this.
2) Definitely not necessary for this PR or anything like that, but I actually feel like a lot of the `toDart` and `toJS` calls should actually go to the call sites, not in `dom.dart`. `dom.dart` is more or less a direct representation of the browser API itself, and as such, if you're using the APIs exposed in here, you are very much aware that you are interacting with JavaScript. Therefore, I believe it is on the layer above to decide if and how to do that marshaling.

There are many situations in which we might be taking a `JSString` directly from a browser API and passing it back into another browser API, which doesn't necessarily need a translation. With this setup, we are always doing conversions back and forth even if we don't need it. This combined with the fact that the profiles of wasm startup are heavily bogged down in JS<->Dart string conversions, I think we should really start thinking about doing another pass at some point that makes all the APIs in `dom.dart` deal with JS objects purely, and modify the call sites. That being said, this PR is probably the simpler and more iterative way to do it, so I think that is a bridge to cross another day.

lib/web_ui/lib/src/engine/dom.dart
joshualitt: Okay, `JSUndefined` isn't reified yet unfortunately. This is for compatibility reasons with how JS backends handle `null` / `undefined`. Let me add a comment to revisit this when we reify `JSUndefined` on Dart2Wasm.

PR #40311, opened 2023-03-15
Pass parameters by value to `StreamHandlerError`

yaakovschectman:
`StreamHandlerError` previously took references to its parameters, which means if the error persists longer than its calling scope, it may point to nonexistent values and crash the program. Change the constructor to take the string parameters as value rather than reference, and take its details as a `unique_ptr` rather than a raw pointer. https://github.com/flutter/flutter/issues/101682#issuecomment-1095311213 suggests that existing projects only ever pass `nullptr` to this parameter, so changing to a `unique_ptr` will not break them.

The `details` field is potentially passed to `EncodeErrorEnvelopeInternal`. Our implementations of this abstract method in the standard and json codecs copy the value pointed to by the parameter, so passing the raw pointer behind the unique pointer should not lead to dangling pointers.

https://github.com/flutter/flutter/issues/101682

## Pre-launch Checklist

- [ ] I signed the [CLA].


shell/platform/common/client_wrapper/include/flutter/event_channel.h
```diff
@@ -106,7 +106,7 @@ class EventChannel {
             if (error) {
               result = codec->EncodeErrorEnvelope(error->error_code,
                                                   error->error_message,
-                                                  error->error_details);
+                                                  error->error_details.get());
```
yaakovschectman, line 109: This method points to an abstract method that we implement in the standard and json codecs, both of which copy the value pointed to by the parameter.

loic-sharma (approved):
LGTM but please also get an approval from Stuart

shell/platform/common/client_wrapper/event_channel_unittests.cc
```diff
@@ -274,4 +274,22 @@ TEST(EventChannelTest, HandlerOutlivesEventChannel) {
...
+TEST(EventChannelTest, StreamHandlerErrorPassByValue) {
+  std::unique_ptr<StreamHandlerError<>> error = nullptr;
+
+  {
+    std::string code = "Code";
+    std::string msg = "Message";
+    std::unique_ptr<EncodableValue> details =
+        std::make_unique<EncodableValue>("Details");
+    error =
+        std::make_unique<StreamHandlerError<>>(code, msg, std::move(details));
```
stuartmorgan, line 286: Let's add a second test that does a `(code, msg, nullptr)` so that we're absolutely certain that we aren't breaking that existing-in-the-wild pattern.

stuartmorgan (approved):
LGTM with one added test.

> https://github.com/flutter/flutter/issues/101682#issuecomment-1095311213 suggests that existing projects only ever pass `nullptr` to this parameter, so changing to a `unique_ptr` will not break them.

Fingers crossed that this is still true. (If there's significant unexpected fallout we may have to revert and figure out another plan.)

cbracken (approved):
Nice! Thanks for fixing!

PR #40322, opened 2023-03-15
Print the malioc_diff.py command to update for new shaders

zanderso:


jonahwilliams (approved):
LGTM!

impeller/tools/malioc_diff.py
```diff
@@ -285,14 +285,20 @@ def main(argv):
...
-          .format(filename)
-      )
+      new_shader = True
+      print('Shader {} is new.'.format(filename))
       changed = True
 
+  if new_shader:
+    print(
+        'There are new shaders. The golden file must be updated with:\n\n'
+        '$ ./flutter/impeller/tools/malioc_diff.py --before {} --after {} --update'
```
gaaclarke, line 298: I believe there is a precondition that `android_debug_unopt` is built with the `--malioc-path` path for this script to work which is worth mentioning.

gaaclarke: Or maybe it's better to put that into the script itself.

zanderso: Good call. Updated.

gaaclarke (approved):

PR #40323, opened 2023-03-15
[macOS] Rename FlutterViewController.id to viewId

cbracken:
`id` is a type in Objective-C similar to dynamic in Dart, and while it *can* be used as an identifier, it's bad practice to do so and results in unidiomatic code. `viewId` also more clearly indicates the purpose of this identifier.

Also replaces the use of `@dynamic` (which is meant to indicate properties whose implementation is either added at runtime or stashed away in a superclass) with `@synthesize`, which is an idiomatic way to declare that an ivar `_foo` should be synthesized for property `foo`.

While this is renaming a public property and thus constitutes a breaking change, multi-view support is still under development and thus, practically speaking, this should not be a breaking change.

## Pre-launch Checklist

- [X] I signed the [CLA].

### DIFF
--- a/shell/platform/darwin/macos/framework/Headers/FlutterViewController.h
+++ b/shell/platform/darwin/macos/framework/Headers/FlutterViewController.h
@@ -61,7 +61,7 @@ FLUTTER_DARWIN_EXPORT
  * If the view controller is unattached (see FlutterViewController#attached),
  * reading this property throws an assertion.
  */
-@property(nonatomic, readonly) uint64_t id;
+@property(nonatomic, readonly) uint64_t viewId;
 
 /**
  * The style of mouse tracking to use for the view. Defaults to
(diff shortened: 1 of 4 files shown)


goderbauer (approved):
LGTM

cbracken:
Note - I will send a followup that deals with the other `@dynamic` properties momentarily but wanted to keep this one separate since it makes an API change.

It looks like the `id` property was added post-review in flutter/engine#38981. Please make sure you get a re-review if adding new API.

dkwingsmt:
Do we need the `@synthesize` at all, since ObjC should be synthesizing them automatically?

cbracken:
> Edit: Does ObjC auto-synthesize readonly properties?

You got it -- nope; we've provided an implementation of the accessor for viewId, so it's not synthesized by default.

PR #40328, opened 2023-03-15
replace some ._() constructors with class modifiers

goderbauer:
These constructors were only there to prevent instantiation and extension. We now have class modifiers for that!

**To be submitted after https://github.com/flutter/engine/pull/40318.** (presubmits will fail until it is merged)


Hixie:

albertodev01:
Why `IsolateNameServer` and `PluginRegister` are `abstract final` and not just `final`?

goderbauer:
> 
Why IsolateNameServer and PluginUtilities are abstract final and not just final?

Because they also may not be constructed: https://github.com/dart-lang/language/blob/master/accepted/future-releases/class-modifiers/feature-specification.md#syntax

cbracken (approved):

goderbauer:
> LGTM, but looks like presubmit is failing. Looks like needs some experiment flags enabled.

Yes! (see also PR description) it is waiting for https://github.com/flutter/engine/pull/40318 :)

goderbauer:
In addition to that it turned out that the api_check tool was also hard-coding an outdated dart version to do its internal analysis. I bumped that up. PTAL at the last commit, @cbracken. Thanks!

mdebbar (approved):

goderbauer:
I had to update the analyzer used by the `api_conform_test.dart` to make it know about class modifiers. @mdebbar Can you take another look at this?

lib/web_ui/pubspec.yaml
```diff
@@ -18,7 +18,6 @@ dependencies:
     path: ../../third_party/web_test_fonts
 
 dev_dependencies:
-  analyzer: 5.2.0
```
mdebbar, line 21: Not sure why this is being removed.

mdebbar (approved):

lib/web_ui/pubspec.yaml
goderbauer: This dependency is unused.

mdebbar: What about the IDE? Which analyzer does it use when I open the `lib/web_ui` folder? For some reason, I always assumed the IDE used the `analyzer` dependency version, but I'm not 100% sure.

goderbauer: It uses the analyzer of the dart SDK configured in the IDE. This dependency is used for ‘package:analyzer/…’ imports of which there don’t seem to be any within web_ui.

mdebbar: Thanks for clarifying!

PR #40330, opened 2023-03-15
Skwasm scene

eyebrowsoffire:
This implements `Scene` and `SceneBuilder` in Skwasm, which lets us have golden tests and also can actually run a flutter app with skwasm, albeit missing quite a few rendering capabilities including text rendering. The subsequent changes to flutter tool that will allow compilation against the skwasm renderer will be following soon after this lands.


lib/web_ui/dev/build.dart
```diff
@@ -114,6 +128,8 @@ class GnPipelineStep extends ProcessStep {
 
   String get runtimeModeFlag {
     switch (runtimeMode) {
```
kevmoo, line 130: This could/should just be `runitemMode.name` I think - https://api.dart.dev/stable/2.19.6/dart-core/EnumName/name.html

lib/web_ui/dev/build.dart
```diff
@@ -45,6 +46,12 @@ class BuildCommand extends Command<bool> with ArgUtils<bool> {
           'output will be located at "out/wasm_profile".\nThis only applies to '
           'the wasm build. The host build is always built in release mode.',
     );
+    argParser.addFlag(
```
kevmoo, line 49: Do we have logic to barf if both `debug` and `profile` are set?

eyebrowsoffire: You're right, I should probably add that.

lib/web_ui/dev/build.dart
eyebrowsoffire: Yep, good call.

lib/web_ui/dev/build.dart
eyebrowsoffire: Nah, you were right, they are mutually exclusive. It should just barf.

lib/web_ui/dev/build.dart
```diff
@@ -170,6 +183,8 @@ class NinjaPipelineStep extends ProcessStep {
       return environment.hostDebugUnoptDir.path;
     }
     switch (runtimeMode) {
+      case RuntimeMode.debug:
```
kevmoo, line 186: Next time you're bored – this could be a switch "expression" 😄

lib/web_ui/dev/utils.dart
```diff
@@ -328,6 +334,32 @@ mixin ArgUtils<T> on Command<T> {
...
+    } else if (isDebug) {
+      return RuntimeMode.debug;
+    } else {
+      return RuntimeMode.release;
+    }
+  }
+}
+
+io.Directory getBuildDirectoryForRuntimeMode(RuntimeMode runtimeMode) {
+  switch (runtimeMode) {
```
kevmoo, line 355: you can use the new switch expression feature here!

=> switch (runtimeMode) { RuntimeMode.debug => envi.wasmdebugOUt, ...}

kevmoo: that should be same same. Weird!

eyebrowsoffire: Ah, I had the syntax slightly wrong. It worked when I copied your example more closely. Thanks, this is slick!

lib/web_ui/lib/src/engine/renderer.dart
```diff
@@ -213,5 +213,5 @@ abstract class Renderer {
 
   ui.ParagraphBuilder createParagraphBuilder(ui.ParagraphStyle style);
 
-  void renderScene(ui.Scene scene);
+  Future<void> renderScene(ui.Scene scene);
```
mdebbar, line 216: This is being called from a synchronous method:
https://github.com/flutter/engine/blob/2e5cdeade5cc15e52589099f571347987b5fb3ba/lib/web_ui/lib/src/engine/platform_dispatcher.dart#L708-L710

I don't think we should change the signature of `Renderer.renderScene`. Since `skwasm` is the only one that needs the asynchronicity, let's only change `SkwasmRenderer.renderScene`'s signature? And I would argue that even `SkwasmRenderer.renderScene` doesn't need to be asynchronous. Nothing is awaiting it, so effectively, we are firing the future and forgetting about it.

One concern that I have here though is race conditions. E.g. what if the previous `renderScene` call takes a while and we receive a new one? Should we somehow force these calls to go in sequence (by pushing them into a queue for example)? Again, if this is something we plan to handle later, let's add a TODO.

mdebbar (commented):
This is looking good!

I have some comments for the dart code, and I would like someone else to review the C++ code.

lib/web_ui/dev/test_runner.dart
```diff
@@ -26,7 +26,7 @@ class TestCommand extends Command<bool> with ArgUtils<bool> {
   TestCommand() {
     argParser
       ..addFlag(
-        'debug',
+        'debug-browser',
```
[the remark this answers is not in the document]
eyebrowsoffire: Yeah, that is probably a better name. One detail is that this option also causes us to run in a window instead of headless, but that probably is fine to just leave implied.

lib/web_ui/lib/src/engine/renderer.dart
eyebrowsoffire: It is fine for this to be called and not awaited from a synchronous method in a "fire and forget" manner, and that's what is still going on right now. Making this method async at the renderer level only makes things a bit more expressive and flexible. The existing renderers which do not asynchronously render can return an immediately resolved future.

The main impetus behind doing this is because the new golden screenshot capabilities in the golden tests need to wait for the rendering to be actually complete before attempting a screenshot. So really, this future is only being awaited in tests.

As for your concern in the last paragraph, I have a lot of the same thoughts. That's sort of the reason why I'd like to expose this as an async method. In situations where the render thread is slower than our target FPS, we may want to keep track of pending futures and stop actually submitting new frames if it starts getting backed up. Otherwise, if we continue pushing more frames, the render thread will get continually more backed up, when it's better to to just drop frames. We can't do this with our current "fire-and-forget" method, but making the method async opens us up to doing this in the future.

lib/web_ui/lib/src/engine/skwasm/skwasm_impl/renderer.dart
```diff
@@ -154,12 +201,13 @@ class SkwasmRenderer implements Renderer {
...
-  FlutterFontCollection get fontCollection => throw UnimplementedError('fontCollection not yet implemented');
-
   @override
   FutureOr<void> initialize() {
-    throw UnimplementedError('initialize not yet implemented');
+    sceneElement = createDomCanvasElement();
+    sceneElement.id = 'flt-scene';
+    domDocument.body!.appendChild(sceneElement);
+    surface = SkwasmSurface('#flt-scene');
+    domDocument.body!.removeChild(sceneElement);
```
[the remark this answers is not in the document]
eyebrowsoffire: Yeah, maybe I should just add a todo. This is just a very rough initial implementation to just get things working in the most basic case. We'll also have to handle multiple surfaces once we start introducing platform views, so this embedding is very much temporary.

lib/web_ui/lib/src/engine/skwasm/skwasm_impl/surface.dart
```diff
@@ -2,25 +2,58 @@
...
+    final SkwasmSurface surface = SkwasmSurface._fromHandle(surfaceHandle);
+    surface._initialize();
+    return surface;
   }
 
-  Surface._fromHandle(this._handle);
+  SkwasmSurface._fromHandle(this._handle);
   final SurfaceHandle _handle;
+  OnRenderCallbackHandle _callbackHandle = nullptr;
+  final Map<int, Completer<void>> _pendingCallbacks = <int, Completer<void>>{};
```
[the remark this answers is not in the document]
eyebrowsoffire: Hmm how about `_pendingRenders` maybe?

lib/web_ui/lib/src/engine/vector_math.dart
```diff
@@ -1014,6 +1016,31 @@ class Matrix4 {
...
+      vector[0] = point.dx;
+      vector[1] = point.dy;
+      transform2(vector);
+      minX = minX == null ? vector[0] : math.min(minX, vector[0]);
+      minY = minY == null ? vector[1] : math.min(minY, vector[1]);
+      maxX = maxX == null ? vector[0] : math.max(maxX, vector[0]);
+      maxY = maxY == null ? vector[1] : math.max(maxY, vector[1]);
+    }
+    return ui.Rect.fromLTRB(minX!, minY!, maxX!, maxY!);
+  }
```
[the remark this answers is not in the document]
eyebrowsoffire: So I went ahead and did some microbenchmarking comparing the two functions. As it turns out, in dart2js, my method is only very slightly slower, about 2.5% in my measurements, but in dart2wasm my method is more than twice as fast:
```
[chrome-dart2js-html-engine] Running...
00:00 +0: loading bench_transform_rect_test.dart                                                         
[CHROME STDERR]:
[CHROME STDERR]:DevTools listening on ws://127.0.0.1:12345/devtools/browser/207d349b-2b04-4cba-881e-23f0afc4d755
00:02 +0: benchmarks Benchmark transformRect                                                             
Counter = 1000000 elapsed time = 2197ms
00:05 +1: benchmarks Benchmark matrix.transformRect                                                      
Counter = 1000000 elapsed time = 2252ms
00:05 +2: All tests passed!                                                                              
[chrome-dart2js-html-engine] All tests passed!
[chrome-dart2wasm-html-engine] Running...
00:00 +0: loading bench_transform_rect_test.dart                                                         
[CHROME STDERR]:
[CHROME STDERR]:DevTools listening on ws://127.0.0.1:12345/devtools/browser/c52f7c4a-2c93-4e7a-84a1-b0a9dbb5c146
00:06 +0: benchmarks Benchmark transformRect                                                             
Counter = 1000000 elapsed time = 5606ms
00:08 +1: benchmarks Benchmark matrix.transformRect                                                      
Counter = 1000000 elapsed time = 2493ms
00:08 +2: All tests passed!                                                                              
[chrome-dart2wasm-html-engine] All tests passed!
```
I think maybe we should switch over to my method. I might also be able to make an additional change to try to get that 2.5% to go away on the JS target, let me see.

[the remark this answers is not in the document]
eyebrowsoffire: This is what the benchmark looks like btw:
```
    test('Benchmark transformRect', () {
      final Matrix4 transform = Matrix4.rotationZ(math.pi / 3.0);
      const Rect rect = Rect.fromLTRB(-100, -100, 100, 100);
      final Stopwatch stopwatch = Stopwatch()..start();
      int counter = 0;
      for (int i = 0; i < 1000000; i++) {
        final Rect transformedRect = transformRect(transform, rect);
        if (transformedRect.contains(Offset.zero)) {
          counter++;
        }
      }
      stopwatch.stop();
      print('Counter = $counter elapsed time = ${stopwatch.elapsedMilliseconds}ms');
    });
```
And then copy-pasta'd except with `matrix.transformRect` put in place of `transformRect`

lib/web_ui/dev/test_runner.dart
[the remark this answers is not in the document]
mdebbar: Yep, that's how it works in the flutter tool too.

lib/web_ui/lib/src/engine/vector_math.dart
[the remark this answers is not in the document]
eyebrowsoffire: I made some other minor changes to narrow the gap slightly in the JS case to about 1-1.5%.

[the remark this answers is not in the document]
eyebrowsoffire: So, as it turns out, these are actually slightly different, as the existing one handles perspective transforms and mine does not. I think I'm going to go ahead and roll things back and use the existing implementation, and hope that binaryen can optimize the wasm side enough to make it acceptable. If it doesn't, I can always hand-tune the wasm side later, if needed.

lib/web_ui/lib/src/engine/renderer.dart
hterkelsen: Agreed with Mouad, this should be synchronous for consistency with the mobile Flutter engine and our public facing API. The Skwasm renderer can have a way to `await` for the scene to actually be rendered which can be used in tests. As for the problem with the render thread getting backed up, this is already something that can happen in Flutter mobile and I think they just drop everything but the latest frame.

eyebrowsoffire: Is there a suggested alternative? This `Renderer.renderScene` API is web-specific, so there is nothing in the mobile Flutter engine to keep it consistent with, and I'm also not changing the public facing API (`PlatformDispatcher.render` is still a synchronous method). Basically the reason I want to return a Future here is:
* To allow us to wait for the render to complete in the tests.
* If we do want to "drop everything but the latest frame" like the native engine does, the UI thread still needs a way to tell whether we are actually getting backed up.
A `Future` returned directly from the method seems like the most natural way to achieve this, but if there is something different you all have in mind I'm open to suggestions.

eyebrowsoffire: Maybe I could have `Renderer.renderScene` return `FutureOr<void>`, so that the subclasses of `Renderer` that aren't asynchronous can still just return `void`?

mdebbar: One way is to only make `SkwasmRenderer.renderScene` asynchronous, and leave others synchronous:
```dart
abstract class Renderer {
  void renderScene();
}

class OtherRenderer implements Renderer {
  @override
  void renderScene() {
    print('foo');
  }
}

class SkwasmRenderer implements Renderer {
  // `Future<void Function()` can override `void Function()`.
  @override
  Future<void> renderScene() async {
    await Future.delayed(Duration(milliseconds: 2));
    print('foo');
  }
}
```
Another way is to keep `SkwasmRenderer.renderScene` synchronous too, and add a separate async method:
```dart
abstract class Renderer {
  void renderScene();
}

class OtherRenderer implements Renderer {
  @override
  void renderScene() {
    print('foo');
  }
}

class SkwasmRenderer implements Renderer {
  @override
  void renderScene() {
    // Fire and forget.
    renderSceneAndWait();
  }
  
  @visibleForTesting
  Future<void> renderSceneAndWait() async {
    await Future.delayed(Duration(milliseconds: 2));
    print('foo');
  }
}
```

Both of these approaches will be annoying if you want a test that runs against all renderers:
```dart
for (final Renderer renderer in renderers) {
  // In order to access the `renderSceneAndWait` method, you need to do:
  if (renderer is SkwasmRenderer) {
    await renderer.renderSceneAndWait();
  } else {
    renderer.renderScene();
  }
}
```

eyebrowsoffire: I've already switched over to `FutureOr<void>`, which allows some renderers to be sync and others to be async. I don't understand why we would want to add the complexity of having a layer that does "fire and forget" within the skwasm renderer, when we could simply "fire and forget" from the `PlatformDispatcher`, which is what we are already doing. It's worth mentioning that my change as written makes no change to production behavior, and does not modify the public API. I don't really understand the objection, and why we would want to add the complexity of separate `renderScene` vs `renderSceneAndWait` APIs and class introspection when the behavior can be sufficiently encapsulated generically with a single method.

mdebbar: I should've clarified my motivation for keeping things as `sync` as possible. It's because I want future maintainers to be aware that this is used as a sync API and it's better to stay that way. And if they really need the asynchronicity, then they can make the deliberate decision of converting it to async.

If we make all of them async today, I'm afraid it becomes too easy in the future to add async stuff in there without thinking about the consequences. I want to conversion to async to be a deliberate decision based on a real need (like in the case of skwasm).

eyebrowsoffire: Does the `FutureOr<void>` change not address your issue? `SkwasmRenderer`'s render will return `Future<void>`, but the other renderers still return `void`. So it would require us to explicitly change a renderer's return value to make it async

mdebbar: It doesn't address the issue. The developer making change to `renderScene` will think it's okay to add an `await` in there because the signature already "supports" async. I want them to have the extra burden of changing the signature, which will make them look around and check whether it's fine to convert it to async or not.

That said, I don't feel strongly about it, and maybe this is all just me doing premature design optimizations :)

mdebbar (approved):

mdebbar:
I approved the PR but I would love someone more qualified than me to review the C++ changes.

lib/web_ui/lib/src/engine/renderer.dart
eyebrowsoffire: Just to clarify, if a developer is editing the `renderScene` function on one of the existing renderers (whose `renderScene` method is synchronous), and they want to add an await, they *will* have the extra burden of changing the signature, since the signature of the existing renderers is staying `void renderScene(...)`. They would have to explicitly change it to `Future<void> renderScene(...)`. The abstract `Renderer` class has the signature `FutureOr<void> renderScene(...)`, but the actual classes themselves either explicitly return `void` or explicitly return `Future<void>`

hterkelsen (approved):
LGTM

lib/web_ui/lib/src/engine/renderer.dart
mdebbar: Ahh okay got it now. I misunderstood your previous comment, thought you were making subclasses `FutureOr<void>` too.

Ok, this actually addresses my concerns, thanks!

PR #40332, opened 2023-03-15
Build iOS and macOS Obj-C with -fapplication-extension flag

jmagman:
First-pass on allowing engine to be built for app extensions without unsafe APIs.  This would allow the engine to be built twice, once for apps and again for app extensions.  Note the build rule tweaks are only to allow testing, this would require much more work to output two versions of the framework.

See [design doc](https://docs.google.com/document/d/1WloBrAYpqNKOhEHytTqJEHshcXxHflLUGKvULKRHbuU/edit#)

> An alternative would be to build two versions of the Flutter engines–one safe to be used in app extensions and built with with the -fapplication-extension flag and with APPLICATION_EXTENSION_API_ONLY=1 added to the defines, and the other identical to the one that exists today.  Both versions would be built and published as artifacts with the same engine infrastructure and rolled into g3.  Apps would continue to use the existing engine, and app extensions would instead link on and embed the new safe variant.  This would double the Flutter engine size in the final app bundle (at time of writing 9.1 MB in an uncompressed release app).  This alternative would be faster to implement than splitting the engines, and could be a medium-term stopgap for customers willing to sacrifice app size for extension features.
> 
> App extension unsafe code would be compiled out of the new safe framework behind conditional APPLICATION_EXTENSION_API_ONLY checks and NS_EXTENSION_UNAVAILABLE_IOS annotations.  Some APIs like the FlutterPluginAppLifeCycleDelegate would be disallowed in the new safe framework, potentially impacting what plugins or plugin features can be used in extensions.

### DIFF
--- a/common/config.gni
+++ b/common/config.gni
@@ -57,6 +57,7 @@ if (is_ios || is_mac) {
   flutter_cflags_objc = [
     "-Werror=overriding-method-mismatch",
     "-Werror=undeclared-selector",
+    "-fapplication-extension",
   ]
   flutter_cflags_objcc = flutter_cflags_objc
   flutter_cflags_objc_arc = flutter_cflags_objc + [ "-fobjc-arc" ]
--- a/shell/platform/darwin/ios/BUILD.gn
+++ b/shell/platform/darwin/ios/BUILD.gn
@@ -44,7 +44,10 @@ source_set("flutter_framework_source_arc") {
   cflags_objc = flutter_cflags_objc_arc
   cflags_objcc = flutter_cflags_objcc_arc
 
-  defines = [ "FLUTTER_FRAMEWORK=1" ]
+  defines = [
+    "FLUTTER_FRAMEWORK=1",
+    "APPLICATION_EXTENSION_API_ONLY=1",
+  ]
   allow_circular_includes_from = [ ":flutter_framework_source" ]
   deps = [
     ":flutter_framework_source",
@@ -153,7 +156,10 @@ source_set("flutter_framework_source") {
 
   sources += _flutter_framework_headers
 
-  defines = [ "FLUTTER_FRAMEWORK=1" ]
+  defines = [
+    "FLUTTER_FRAMEWORK=1",
+    "APPLICATION_EXTENSION_API_ONLY=1",
+  ]
 
   if (shell_enable_metal) {
     sources += [
--- a/shell/platform/darwin/ios/BUILD.gn
+++ b/shell/platform/darwin/ios/BUILD.gn
@@ -326,7 +326,10 @@ shared_library("create_flutter_framework_dylib") {
 
   output_name = "Flutter"
 
-  ldflags = [ "-Wl,-install_name,@rpath/Flutter.framework/Flutter" ]
+  ldflags = [
+    "-Wl,-install_name,@rpath/Flutter.framework/Flutter",
+    "-fapplication-extension",
+  ]
 
   public = _flutter_framework_headers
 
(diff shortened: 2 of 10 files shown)


shell/platform/darwin/ios/framework/Source/FlutterViewController.mm
```diff
@@ -814,9 +814,14 @@ - (void)viewDidAppear:(BOOL)animated {
   if ([_engine.get() viewController] == self) {
     [self onUserSettingsChanged:nil];
     [self onAccessibilityStatusChanged:nil];
+
+#if !APPLICATION_EXTENSION_API_ONLY
     if (UIApplication.sharedApplication.applicationState == UIApplicationStateActive) {
+#endif
       [[_engine.get() lifecycleChannel] sendMessage:@"AppLifecycleState.resumed"];
```
xster, line 821: Is it possible to end up sending this if the whole process hosting the extension itself was in the background, thus UIApplication.sharedApplication.applicationState != UIApplicationStateActive if we had access to the sharedApplication? If we sent in "resumed" without checking, might be end up in an engine state where we continue to issue GPU commands while in the background and crash?

jmagman: Seems like a race that I bet could happen in the app extension https://github.com/flutter/engine/pull/23175

shell/platform/darwin/ios/framework/Source/FlutterViewController.mm
```diff
@@ -1804,6 +1814,7 @@ - (void)onOrientationPreferencesUpdated:(NSNotification*)notification {
 }
 
 - (void)performOrientationUpdate:(UIInterfaceOrientationMask)new_preferences {
+#if !APPLICATION_EXTENSION_API_ONLY
```
jmagman, line 1817: Might be able to fall back to changing `UIWindowSceneGeometryPreferencesIOS` on `self.viewIfLoaded.window.windowScene` instead, if that works.

common/config.gni
```diff
@@ -57,6 +57,7 @@ if (is_ios || is_mac) {
   flutter_cflags_objc = [
     "-Werror=overriding-method-mismatch",
     "-Werror=undeclared-selector",
+    "-fapplication-extension",
```
jmagman, line 60: This would be set by a buildroot flag so the engine can be built with or without app extension safety. I put it here for testing.

shell/platform/darwin/ios/BUILD.gn
```diff
@@ -44,7 +44,10 @@ source_set("flutter_framework_source_arc") {
   cflags_objc = flutter_cflags_objc_arc
   cflags_objcc = flutter_cflags_objcc_arc
 
-  defines = [ "FLUTTER_FRAMEWORK=1" ]
+  defines = [
+    "FLUTTER_FRAMEWORK=1",
+    "APPLICATION_EXTENSION_API_ONLY=1",
```
jmagman, line 49: Same, this would be set by a buildroot flag.

shell/platform/darwin/ios/BUILD.gn
```diff
@@ -320,7 +326,10 @@ shared_library("create_flutter_framework_dylib") {
 
   output_name = "Flutter"
 
-  ldflags = [ "-Wl,-install_name,@rpath/Flutter.framework/Flutter" ]
+  ldflags = [
+    "-Wl,-install_name,@rpath/Flutter.framework/Flutter",
+    "-fapplication-extension",
```
jmagman, line 331: Same, behind a new flag, etc.

jmagman:
Closing in favor of https://github.com/flutter/engine/pull/40972

PR #40338, opened 2023-03-16
Expose  iOS plist and AndroidManifest setting to disable partial repaint

sanlittt:
expose a unified setting for iOS plist and Android Manifest to disable partial repaint
Android
`<meta-data
            android:name="io.flutter.embedding.android.DisableParticalRepaint"
            android:value="true" />`
iOS
`  <key>FLTDisablePartialRepaint</key>
  <true/>`

We find screen flickering both on ios and android platform,
try to enable application Settings to see if disabling partial repaint helps.

As existing only supports ios Settings:https://github.com/flutter/engine/pull/34328

### DIFF
--- a/common/settings.h
+++ b/common/settings.h
@@ -211,6 +211,10 @@ struct Settings {
   // Enable the Impeller renderer on supported platforms. Ignored if Impeller is
   // not supported on the platform.
   bool enable_impeller = false;
+    
+  // Disable the partical repaint on supported platforms. Ignored if partical repaint is
+  // not supported on the platform.
+  bool disable_partical_repaint = false;
 
   // Data set by platform-specific embedders for use in font initialization.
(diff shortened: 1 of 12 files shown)


gmackall:
Adding @reidbaker to review the android part of this, @jmagman would you be able to find a reviewer for the iOS side?

chinmaygarde (changes requested):
We should likely just consider backing out the change as there are a lot of patches from folks wanting to disable this. If we can't guarantee correctness, we should back out partial repaint.

dnfield:
We actually intend to remove the flag for iOS. It was only added temporarily to debug something. Closing this as wontfix - especially since it sounds like we're just going to remove Android support on this anywya (https://github.com/flutter/flutter/issues/123353)

PR #40355, opened 2023-03-16
Output web test artifacts to out directory.

eyebrowsoffire:


lib/web_ui/dev/environment.dart
```diff
@@ -150,8 +150,8 @@ class Environment {
   ///
   /// This is where compiled output goes.
   io.Directory get webUiBuildDir => io.Directory(pathlib.join(
-        webUiRootDir.path,
-        'build',
+        outDir.path,
```
godofredoc, line 153: Can we copy to both locations as a transitional step to not break the current tests?

lib/web_ui/dev/licenses.dart
```diff
@@ -79,13 +79,18 @@ class LicensesCommand extends Command<bool> {
   }
 
   List<io.File> _flatListSourceFiles(io.Directory directory) {
+    // This is the old path that tests used to be built into. Ignore anything
+    // within this path.
+    final String legacyBuildPath = path.join(environment.webUiRootDir.path, 'build');
```
mdebbar, line 84: Should we also add the legacy path to the `felt clean` command?
https://github.com/flutter/engine/blob/b2720b7e9887ff0c000d43f5323c9b4995df7831/lib/web_ui/dev/clean.dart#L40-L49

I personally rarely use `felt clean` so I don't have a strong opinion here.

lib/web_ui/dev/environment.dart
eyebrowsoffire: I don't think we'll have to do that. AFAIK the current tests will not be broken by this.

mdebbar (approved):

lib/web_ui/dev/licenses.dart
eyebrowsoffire: Yeah good idea

Hixie:
test-exempt: is tests

PR #40361, opened 2023-03-16
Handle error more gracefully, add test for same

gspencergoog:
## Description

This fixes the application exit handler so that it deals more gracefully with errors returned to it from the framework.

In this case, since the framework doesn't yet have support for requesting an application exit, it was returning an error condition, which the previous code tried to convert to an NSDictionary (and failed, causing a crash).

Now it will log the error and quit as requested if an error or unrecognized object type is returned.

## Related Issues
 - https://github.com/flutter/flutter/issues/122813

## Tests
 - Added a test for the error condition.

### DIFF
--- a/shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
+++ b/shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
@@ -231,7 +231,6 @@ constexpr char kTextPlainFormat[] = "text/plain";
                      return;
                    }
                    NSDictionary* replyArgs = (NSDictionary*)decoded_reply;
-                   NSLog(@"dictionary: %@", replyArgs);
                    if ([replyArgs[@"response"] isEqual:@"exit"]) {
                      _terminator(sender);
                    }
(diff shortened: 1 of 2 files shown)


cbracken (approved):

shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
```diff
@@ -215,9 +215,24 @@ - (void)requestApplicationTermination:(id)sender
...
+                     return;
+                   }
+                   if (![decoded_reply isKindOfClass:[NSDictionary class]]) {
+                     NSLog(@"Call to System.requestAppExit returned an unexpected object: %@",
+                           decoded_reply);
+                     _terminator(sender);
+                     return;
+                   }
+                   NSDictionary* replyArgs = (NSDictionary*)decoded_reply;
+                   NSLog(@"dictionary: %@", replyArgs);
```
cbracken, line 234: Delete this line.

gspencergoog: Done!

PR #40370, opened 2023-03-17
[Impeller] dont tessellate rectangles in TextureContents

jonahwilliams:
Doesn't make a huge difference but helps reduce noise for tracking work the tessellator is doing.


bdero (approved):
LGTM

PR #40373, opened 2023-03-17
Wrap the iOS platform message handler in an autorelease pool block

jason-simmons:
With background platform isolate channels, the handler can be invoked on threads that are created by Dart and do not already have an autorelease pool.

Fixes https://github.com/flutter/flutter/issues/122792


gaaclarke (approved):
I think there should be runtime warnings getting printed out if you are adding objects to the autorelease pool but there isn't one present.

If you want to add a test for this it wouldn't be that hard.  In `platform_message_handler_ios_test.mm` you could make an NSObject that takes in a pointer to a bool and sets it's value in its dealloc.  Then you can assert that dealloc was called.

PR #40386, opened 2023-03-17
Update analyzer for api_conform_test

goderbauer:
Follow-up to https://github.com/flutter/engine/pull/40328.

Just now a new analyzer version was released on pub that has the 3.0 experiments enabled by default (see also https://github.com/dart-lang/sdk/issues/51760). With that version we don't have to manually manage experiment flags in the api_conform_test.


mdebbar (approved):
Can't wait to try new Dart features!

PR #40387, opened 2023-03-17
Revert "[Impeller] mark decoded images as optimized for GPU access"

jonahwilliams:
Reverts flutter/engine#40356

This change regressed raster thread metrics, see: https://github.com/flutter/flutter/issues/122897

Image upload itself is done from a background thread, so it can't block directly. Unfortunately, that likely means that the optimizeForGpuAccess is causing extra work that the raster thead is blocked on.


jonahwilliams:
Alternatively, I could try swapping the command order so we do optimizeForGpuAccess and then mipmap generation. Shot in the dark, but might be worth trying

dnfield (approved):

PR #40394, opened 2023-03-17
Post 3.0 lint sync

goderbauer:


goderbauer:
Blocked until the pre-built Dart SDK has been bumped past `3.0.0-325.0.dev`.

goderbauer:
Should be unblocked by Dart v3.0.0-366.0.dev rolling into the engine.

lib/web_ui/lib/src/engine/text/word_breaker.dart
```diff
@@ -5,11 +5,11 @@
 import '../util.dart';
 import 'word_break_properties.dart';
 
-class _FindBreakDirection {
-  const _FindBreakDirection({required this.step});
+enum _FindBreakDirection {
```
ditman, line 8: Was this automatically detected by a lint? wow!

lib/web_ui/test/canvaskit/semantics_test.dart
```diff
@@ -3,6 +3,7 @@
 // found in the LICENSE file.
 
 @TestOn('chrome || safari || firefox')
+library;
```
ditman, line 6: Why are these empty `library;` needed? Is it because of the `@TestOn`?

ditman (commented):
```diff
- break;
```

RSLGTM, if Dan and CI like it too.

The web_ui bits look good to me. Most of them are removing `break;` from switches. I've left a couple of questions.

Heads up @eyebrowsoffire, potentially big conflicts incoming with your test reorg PR?

lib/web_ui/lib/src/engine/text/word_breaker.dart
goderbauer: Yes!

lib/web_ui/test/canvaskit/semantics_test.dart
goderbauer: Yeah, previously, the TestOn was technically annotating the next import statement.

analysis_options.yaml
```diff
@@ -168,10 +168,10 @@ linter:
     - provide_deprecation_message
     - public_member_api_docs # DIFFERENT FROM FLUTTER/FLUTTER
     - recursive_getters
-    # - require_trailing_commas # blocked on https://github.com/dart-lang/sdk/issues/47441
+    # - require_trailing_commas # would be nice, but requires a lot of manual work: 10,000+ code locations would need to be reformatted by hand after bulk fix is applied
```
dnfield, line 171: What does `dart format --line-length=120` look like?

Should not be handled in this PR but might work well enough.

dnfield: Maybe splitting the web and VM implementations would help too, because I think at some point web used dartfmt.

goderbauer: It's probably much more realistic to enable this for the engine than it is for the framework, it will need some additional investigation, though.

PR #40395, opened 2023-03-17
Rename FontCollection to FlutterFontCollection

jason-simmons:
This will avoid name conflicts with the FontCollection API that will be added to CanvasKit.
(see https://skia-review.googlesource.com/c/skia/+/657270)


yjbanov (approved):
lgtm

PR #40396, opened 2023-03-17
drawTextBlob should not be compatible with opacity inheritance

flar:
Technically we've never been able to trust drawTextBlob to be non-overlapping, but in practice it has nearly always been OK with inheriting opacity as most text strings don't have overlapping glyphs, and also in practice most text is drawn over a background fill so it is already involved in an incompatible DisplayList.

Recently, as we try to fix our opacity inheritance implementation code, though, some tests that draw text can have a lot of off-by-1 errors as various paths through the engine start to become capable of distributing the opacity to single-text-blob DisplayList objects, pointing out that there are real world cases where there are differences in rendering text with or without inherited opacity.

So, it is time to bite the bullet and declare that text blobs are incompatible with group opacity optimizations.


jonahwilliams (approved):
LGTM

PR #40398, opened 2023-03-17
Use analyzer from dart source everywhere

goderbauer:
In https://github.com/flutter/engine/pull/40328 it tripped me up that the analyzer used by `tools/api_check` behaved differently than the analyzer used by `web_sdk`. The reason for that mystery was: `tools/api_check` uses the analyzer directly from the dart sources, while `web_sdk` used an (outdated) version from pub. The version from pub will always be slightly behind, which is why `web_sdk` sometimes requires fiddling with experiment flags to get it to behave as expected (see https://github.com/flutter/engine/pull/40386). That's annoying and having to deal with different analyzer versions is confusing. This change migrates `web_sdk` to always use the up-to-date analyzer from the dart sources directly, just like `tools/api_check` does:

https://github.com/flutter/engine/blob/fc57995fe137d883016a6c115bf45710883cd77c/tools/api_check/pubspec.yaml#L40-L82

### DIFF
--- a/web_sdk/pubspec.yaml
+++ b/web_sdk/pubspec.yaml
@@ -6,9 +6,49 @@ environment:
 
 dependencies:
   args: 2.3.1
-  path: 1.8.2
+  path: any # see dependency_overrides
 
 dev_dependencies:
-  analyzer: 5.8.0
+  # Using the bundled analyzer (see dependency_overrides) to always be
+  # up-to-date instead of a version from pub, which may not have the latest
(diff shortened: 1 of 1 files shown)


goderbauer:
@mdebbar Can you take another look at this? Sorry with the churn on this one, but I think this approach will lessen future pain when it comes to upgrading to new dart versions.

mdebbar (approved):

PR #40399, opened 2023-03-17
Multiview painting: Studio

dkwingsmt:

## Pre-launch Checklist

- [ ] I signed the [CLA].


lib/ui/window.dart
```diff
@@ -289,10 +289,10 @@ class FlutterView {
   ///   scheduling of frames.
   /// * [RendererBinding], the Flutter framework class which manages layout and
   ///   painting.
-  void render(Scene scene) => _render(scene);
+  void render(Scene scene) => _render(viewId == 0 ? 0 : viewId.hashCode, scene);
```
loic-sharma, line 292: Why `viewId.hashCode`? Should this be `viewId as int`?
cene);

dkwingsmt: Good point, it was a panic move because I didn't know how to convert Object to int...

flow/surface.h
```diff
@@ -35,22 +35,13 @@ class Surface {
 
   virtual bool IsValid() = 0;
 
-  virtual std::unique_ptr<SurfaceFrame> AcquireFrame(const SkISize& size) = 0;
+  virtual std::unique_ptr<SurfaceFrame> AcquireFrame(int64_t view_id,
```
loic-sharma, line 38: Could you explain why this has a `view_id` parameter? We will have separate surfaces for each view right?

flow/surface.h
```diff
@@ -35,22 +35,13 @@ class Surface {
...
+                                                     const SkISize& size) = 0;
 
   virtual SkMatrix GetRootTransformation() const = 0;
 
   virtual GrDirectContext* GetContext() = 0;
 
-  virtual std::unique_ptr<GLContextResult> MakeRenderContextCurrent();
-
-  virtual bool ClearRenderContext();
```
loic-sharma, line 46: Could you explain why you removed the context methods from the surface?

dkwingsmt: Because they don't seem needed :)  I was actually testing whether they were, and after I removed these methods it actually compiles.

shell/common/rasterizer.cc
```diff
@@ -125,51 +124,89 @@ void Rasterizer::DisableThreadMergerIfNeeded() {
...
-    return;
+int Rasterizer::DrawLastLayerTree(
+    std::unique_ptr<FrameTimingsRecorder> frame_timings_recorder,
+    bool enable_leaf_layer_tracing) {
+  if (!studio_) {
+    return 0;
+  }
+  int success_count = 0;
+  bool should_resubmit_frame = false;
+  for (auto& record_pair : surfaces_) {
```
loic-sharma, line 184: Should this be able to render the last layer tree for just a single surface?

From my understanding, `DrawLastLayerTree` is used in scenarios where an external texture was updated but the framework's layer tree is otherwise unchanged. For example, if a video player needs to update the surface on each video frame even though the Flutter app itself hasn't rendered a frame. We shouldn't re-draw a second window that doesn't contain the video.

shell/common/rasterizer.cc
```diff
@@ -185,17 +222,19 @@ RasterStatus Rasterizer::Draw(
...
+  DoDrawResult draw_result;
   LayerTreePipeline::Consumer consumer =
-      [&](std::unique_ptr<LayerTreeItem> item) {
+      [this, &draw_result,
+       &discard_callback](std::unique_ptr<LayerTreeItem> item) {
         std::shared_ptr<LayerTree> layer_tree = std::move(item->layer_tree);
         std::unique_ptr<FrameTimingsRecorder> frame_timings_recorder =
             std::move(item->frame_timings_recorder);
+        int64_t view_id = item->view_id;
         if (discard_callback(*layer_tree.get())) {
```
loic-sharma, line 233: I believe this discard callback is used when a window was resized. We discard the scene if its dimensions don't match the window's new size. We'll need to update this callback as each surface may have different dimensions.

flow/surface.h
dkwingsmt: Yeah you are right.

Initially I added the view_id to `SurfaceFrame::Submit`, so that surfaces and surface frames can be "view agnostic" until the last second so that they feel more "pure", but it turns out that we want frame buffers to be tied with a view ID so that buffers can be invalidated more easily. I moved `view_id` earlier to `AcquireFrame`, but on a second thought we should just straight up tie surfaces with views anyway.

shell/common/rasterizer.cc
dkwingsmt: If there are two videos in the app on two different windows, then we'll need the last-frame information for both views.

shell/common/rasterizer.cc
dkwingsmt: Sounds reasonable, I'll update it! Thank you.

flow/surface.h
dkwingsmt: Although I'm wondering if we can assume that all surfaces for a view are of the same size, so that if we invalidate the frame buffer for one surface of a view, then we can also invalidate all other frame buffers of that view.

shell/common/rasterizer.cc
loic-sharma: Right, but do we need to rasterize all windows if only one window has a video player?

flow/surface.h
loic-sharma: > I'm wondering if we can assume that all surfaces for a view are of the same size

Are you including the macOS embedder's surfaces here? If so, I don't think so: the engine may have one layer for a button and another layer for the app's background. From my understanding, these would translate into macOS embedder surfaces with different dimensions.

shell/common/rasterizer.cc
dkwingsmt: I think you have a valid point, but let me rephrase your question.

In the case that multiple views have a video player each, we should probably allow `DrawLastLayerTree` to only draw the views that need a resubmission, i.e. whose video player triggered this vsync - instead of resubmitting all views on all vsyncs.

However, this feature is tricky. We need to allow VSyncs to carry information on the views that triggered it, so that the animator can send it back to the rasterizer.

Also, this feature can be done as a separate issue after the main changes, since it affects performance, not functionality.

PR #40401, opened 2023-03-17
Migrate SVG to JS types

joshualitt:


joshualitt:
@eyebrowsoffire ptal. @ditman, @mdebbar FYI, please feel free to review if you like.

lib/web_ui/test/initialization_test.dart
```diff
@@ -10,7 +11,8 @@ import 'package:ui/src/engine.dart' as engine;
 import 'package:ui/ui.dart' as ui;
 
 @JS('_flutter')
-external set loader(Object? loader);
+external set _loader(JSAny? loader);
+set loader(Object? l) => _loader = l?.toJSAnyShallow;
```
ditman, line 15: Took me a while to figure out why this was a setter (it's a test 🤦)

ditman (approved):
LGTM if @eyebrowsoffire likes it.

lib/web_ui/lib/src/engine/view_embedder/hot_restart_cache_handler.dart
```diff
@@ -11,10 +13,14 @@ import '../dom.dart';
...
 /// transition appear smooth.
 @JS('window.__flutterState')
-external List<Object?>? get hotRestartStore;
+external JSArray? get _hotRestartStore;
+List<Object?>? get hotRestartStore =>
+    _hotRestartStore?.toObjectShallow as List<Object>?;
 
 @JS('window.__flutterState')
-external set hotRestartStore(List<Object?>? nodes);
+external set _hotRestartStore(JSArray? nodes);
```
eyebrowsoffire, line 21: I think this one is a prime candidate for eliminating all the `toDart` conversion stuff and using JS types directly. It's localized to just this file, so it should be a pretty straightforward change. Any chance you might actually switch this over Joshua? I think it would be really nice as a small example we can reference for places where we directly use the JS objects themselve.

eyebrowsoffire (approved):
Looks great! I have one comment... it's not necessarily blocking but it would be really nice if you had the time to do it.

lib/web_ui/lib/src/engine/view_embedder/hot_restart_cache_handler.dart
ditman: Ah, great call! IIRC this JSArray is never used from JS actually, so the shallower you can make this conversion, the better :P

joshualitt: Actually, I'm just going to revert this change. Until we reify JS types on JS backends, it's just too awkward to use the JS types past the shim layer.

The one notable exception to this policy is `JSFunction`, because we really shouldn't be calling JS functions from Dart anyways.

Anyway, RE: JSArray, let's circle back after reification and I can write some canonical example code, for now we'll just convert these behind the scenes.

lib/web_ui/test/initialization_test.dart
mdebbar: I may be misunderstanding something here, but isn't this redundant with the `js_util.jsify(...)` calls below?
```dart
set loader(Object? l) => _loader = l?.toJSAnyShallow;

// in tests:
loader = js_util.jsify(<String, Object>{
  'loader': <String, Object>{
    'didCreateEngineInitializer': allowInterop(() { print('not mocked'); }),
  },
});
```

joshualitt: Ahh yea, good catch, it is redundant. I'm not actually trying to investigate usage patterns too deeply at this phase of the migration, mostly to derisk the conversion. That said, this is a good candidate for cleanup when we try to push JS types deeper into the Web engine. Thankfully this is a test, and `jsify` now has fast paths for types it doesn't need to convert.

PR #40403, opened 2023-03-17
[Impeller] Remove Skia dependency from the impeller/image interface

bdero:
Allows building alternative `CompressedImage` backends that don't depend on Skia.


bdero:
Actually this is only used by the playground, just gonna move it to the backend.

chinmaygarde (approved):

PR #40409, opened 2023-03-17
[web] Use records for zoned callbacks

mdebbar:
Exploring the usage of records in some parts of our code base.

### DIFF
--- a/lib/web_ui/lib/src/engine/platform_dispatcher.dart
+++ b/lib/web_ui/lib/src/engine/platform_dispatcher.dart
@@ -118,20 +118,17 @@ class EnginePlatformDispatcher extends ui.PlatformDispatcher {
   /// Receives all events related to platform configuration changes.
   @override
   ui.VoidCallback? get onPlatformConfigurationChanged =>
-      _onPlatformConfigurationChanged;
-  ui.VoidCallback? _onPlatformConfigurationChanged;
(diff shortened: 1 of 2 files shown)


lib/web_ui/lib/src/engine/platform_dispatcher.dart
```diff
@@ -1245,68 +1207,58 @@ bool _handleWebTestEnd2EndMessage(MethodCodec codec, ByteData? data) {
   return false;
 }
 
-/// Invokes [callback] inside the given [zone].
-void invoke(void Function()? callback, Zone? zone) {
+/// Invokes the zoned [callback].
+void invoke(ui.ZonedCallback<ui.VoidCallback>? callback) {
```
mdebbar, line 1211: We could also make `invoke`, `invoke1` and `invoke2` extension methods on `ZonedCallback`. E.g:
```dart
typedef ZonedCallback<T extends Function> = ({Zone zone, T fn});

extension ZonedCallbackExtension on ZonedCallback {
  void invoke() {
    if (identical(zone, Zone.current)) {
      callback();
    } else {
      zone.runGuarded(fn);
    }
  }

  // ...
}
```
then callsites can simply do:
```dart
_onDrawFrame?.invoke();
_onPlatformMessage?.invoke3(...);
```

PR #40412, opened 2023-03-17
[rotation_distortion]send interpolated viewport metrics to address rotation distortion

hellohuanlin:
Send interpolated viewport metrics to make the rotation smoother. Result is pretty good, as show below.  

There are still some minor amount of distortion. There are 2 main reasons: 

1. Flutter drawing happens on the UI thread, which is not iOS main thread. The delay between issuing engine command till the drawing happens is indeterministic and could vary depending on how busy/idle the app is during the rotation. This amount of distortion is unavoidable, and unlikely solvable. 
2. We use linear interpolation for simplicity, but iOS window rotation is likely non-linear, with ease-in ease-out effect. Consider using non-linear interpolation if this becomes a problem. 

Currently we are calling framework from engine per frame. A potential performance improvement can be just sending the rotation duration to the framework, and have the framework to do the interpolation by itself. Though this will require a whole lot more refactoring on both engine and framework. 

https://user-images.githubusercontent.com/41930132/226756067-b5d99590-5565-47e3-bf37-56ca71e763b6.mp4

Fixes https://github.com/flutter/flutter/issues/16322

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/shell/platform/darwin/ios/framework/Source/FlutterViewController.mm
+++ b/shell/platform/darwin/ios/framework/Source/FlutterViewController.mm
@@ -98,9 +98,9 @@ typedef struct MouseState {
 @property(nonatomic, assign) BOOL isPresentingViewControllerAnimating;
 
 /**
- * Whether the device is rotating.
+ * Whether it is interpolating viewport metrics for rotation transition.
  */
-@property(nonatomic, assign) BOOL isDuringRotationTransition;
+@property(nonatomic, assign) BOOL isDuringViewportMetricsInterpolationForRotation;
 
 /**
  * The timer for sending interpolated viewport metrics during rotation.
@@ -918,16 +918,18 @@ static void SendFakeTouchEvent(FlutterEngine* engine,
   // the FlutterViewController to make view controller life cycle methods no-op, and insert
   // this mock into the responder chain.
   CGFloat scale = [UIScreen mainScreen].scale;
-  _isDuringRotationTransition = YES;
+  _isDuringViewportMetricsInterpolationForRotation = YES;
 
   CGSize oldSize = self.view.bounds.size;
   UIEdgeInsets oldPadding = self.view.safeAreaInsets;
 
   __block double rotationProgress = 0;
-  // Invalidate the timer to avoid race condition when a new rotation starts before the previous rotation's timer ends.
-  // The `viewWillTransitionToSize` itself is guaranteed to be called after the previous rotation is complete. However, there can still be race condition because:
-  // 1. the transition duration may not be divisible by `kRotationViewportMetricsUpdateInterval`, resulting in 1 additional frame.
-  // 2. there can still be rounding errors when accumulating the progress which is normalized from 0 to 1.
+  // Invalidate the timer to avoid race condition when a new rotation starts before the previous
+  // rotation's timer ends. The `viewWillTransitionToSize` itself is guaranteed to be called after
+  // the previous rotation is complete. However, there can still be race condition because:
+  // 1. the transition duration may not be divisible by `kRotationViewportMetricsUpdateInterval`, \
+  // resulting in 1 additional frame.
+  // 2. there can still be rounding errors when accumulating the progress which is normalized.
   // 3. NSTimer is backed by the run loop, which is not accurate timing.
   if ([_rotationTimer isValid]) {
     [_rotationTimer invalidate];
@@ -950,7 +952,7 @@ static void SendFakeTouchEvent(FlutterEngine* engine,
 
                                  // End of rotation. Invalidate the timer.
                                  if (rotationProgress == 1) {
-                                   _isDuringRotationTransition = NO;
+                                   _isDuringViewportMetricsInterpolationForRotation = NO;
                                    [timer invalidate];
                                  }
                                }];
@@ -1393,8 +1395,8 @@ static flutter::PointerData::DeviceKind DeviceKindFromTouchType(UITouch* touch) 
 #pragma mark - Handle view resizing
 
 - (void)updateViewportMetricsIfNeeded:(BOOL)forRotation {
-  // update viewport metrics only if `_isDuringRotationTransition` matches `forRotation`.
-  if (_isDuringRotationTransition != forRotation) {
+  // update only if `_isDuringViewportMetricsInterpolationForRotation` matches `forRotation`.
+  if (_isDuringViewportMetricsInterpolationForRotation != forRotation) {
     return;
   }
   if ([_engine.get() viewController] == self) {
--- a/shell/platform/darwin/ios/framework/Source/FlutterViewController.mm
+++ b/shell/platform/darwin/ios/framework/Source/FlutterViewController.mm
@@ -48,6 +48,8 @@ NSNotificationName const FlutterViewControllerShowHomeIndicator =
  * Compute the interpolated value under linear interpolation.
  */
 CGFloat FLTLinearInterpolatedValue(double progress, CGFloat from, CGFloat to, CGFloat scale) {
+  // TODO(hellohuanlin): consider non-linear interpolation to further reduce rotation distortion.
+  // See: https://github.com/flutter/flutter/issues/123248
   NSCAssert(progress >= 0 && progress <= 1, @"progress must be between 0 and 1");
   return (from * (1 - progress) + to * progress) * scale;
 }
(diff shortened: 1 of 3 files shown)


shell/platform/darwin/ios/framework/Source/FlutterViewController.mm
```diff
@@ -44,6 +44,38 @@
...
+/**
+ * Interpolate the viewport metrics for smoother rotation transition.
+ */
+void FLTInterpolateViewportMetrics(flutter::ViewportMetrics& viewportMetrics,
+                                   double rotationProgress,
+                                   CGSize fromSize,
+                                   UIEdgeInsets fromPadding,
+                                   CGSize toSize,
+                                   UIEdgeInsets toPadding) {
+  CGFloat scale = [UIScreen mainScreen].scale;
```
jmagman, line 66: Suggest checking if the view is loaded before doing this:
```suggestion
  UIScreen* mainScreen = self.mainScreenIfViewLoaded;
  if (mainScreen == nil) {
    return;
  }
  CGFloat scale = mainScreen.scale;
```

shell/platform/darwin/ios/framework/Source/FlutterViewController.mm
```diff
@@ -44,6 +44,38 @@
...
+ */
+CGFloat FLTLinearInterpolatedValue(double progress, CGFloat from, CGFloat to, CGFloat scale) {
+  NSCAssert(progress >= 0 && progress <= 1, @"progress must be between 0 and 1");
+  return (from * (1 - progress) + to * progress) * scale;
+}
+
+/**
+ * Interpolate the viewport metrics for smoother rotation transition.
+ */
+void FLTInterpolateViewportMetrics(flutter::ViewportMetrics& viewportMetrics,
```
jmagman, line 60: Instead of passing in a viewportMetrics reference, can this instead be a method on the view controller that directly accesses `_viewportMetrics`?  Or was that too difficult to test?

shell/platform/darwin/ios/framework/Source/FlutterViewController.mm
```diff
@@ -843,6 +882,62 @@ - (void)viewDidDisappear:(BOOL)animated {
...
+                                 UIEdgeInsets newPadding = self.view.safeAreaInsets;
+
+                                 FLTInterpolateViewportMetrics(_viewportMetrics, rotationProgress,
+                                                               oldSize, oldPadding, newSize,
+                                                               newPadding);
+                                 [self updateViewportMetricsIfNeeded:YES];
+
+                                 // End of rotation. Invalidate the timer.
+                                 if (rotationProgress == 1) {
+                                   _isDuringRotationTransition = NO;
```
jmagman, line 935: Is this potentially racy, can `-viewWillTransitionToSize` be called before the timer is invalidated/the rotation is complete? Can the rotation be reversed halfway through?  Maybe this should instead be tracked with a `NSConditionLock`?  Or the timer object be cached on the view controller and invalidated if `-viewWillTransitionToSize` is called a second time before it's complete?

shell/platform/darwin/ios/framework/Source/FlutterViewController.mm
hellohuanlin: Yes I had issue with testing - `OCMock/NSInvocation` only supports populating NSObject references, and not structs. And I do not want to expose `viewportMetrics` property as it's anti-pattern. So I made this pure function for easier testing.

shell/platform/darwin/ios/framework/Source/FlutterViewController.mm
```diff
@@ -63,6 +97,11 @@ @interface FlutterViewController () <FlutterBinaryMessenger,
 @property(nonatomic, assign) BOOL isHomeIndicatorHidden;
 @property(nonatomic, assign) BOOL isPresentingViewControllerAnimating;
 
+/**
+ * Whether the device is rotating.
```
cyanglaz, line 101: nits: by looking at the property name and document, one might get confused thinking this bool indicates whether the physical device is rotating.

shell/platform/darwin/ios/framework/Source/FlutterViewController.mm
```diff
@@ -843,6 +882,62 @@ - (void)viewDidDisappear:(BOOL)animated {
...
+
+  _isDuringRotationTransition = YES;
+
+  CGSize oldSize = self.view.bounds.size;
+  UIEdgeInsets oldPadding = self.view.safeAreaInsets;
+
+  __block double rotationProgress = 0;
+  // Timer is retained by the run loop, and will be released after invalidated.
+  [NSTimer
+      scheduledTimerWithTimeInterval:kRotationViewportMetricsUpdateInterval
```
cyanglaz, line 918: Maybe replace  kRotationViewportMetricsUpdateInterval with `1/[DisplayLinkManager displayRefreshRate]` for pro-motion devices.

hellohuanlin: Probably not necessary during rotation since the device is moving. I've even tried 30fps - not too bad either

shell/platform/darwin/ios/framework/Source/FlutterViewController.mm
hellohuanlin: Just did some research - it turned out to be more involved and likely requires non-trivial refactoring to make it work in unit test, as explained [here](https://github.com/flutter/engine/pull/40412/files#diff-365b13daa5168d36d5824f220bc472a4a90f4685761947f8daf9dc76e7f8d962R915-R919).

shell/platform/darwin/ios/framework/Source/FlutterViewController.mm
cyanglaz: i can test with promotion device

hellohuanlin: Feel free to pull the branch, or can wait for me to fix my laptop's engine build (hopefully this week).

hellohuanlin: Actually, let me see if I can expense one.

hellohuanlin:
Update: @cyanglaz and I found a performance issue for complex projects like flutter gallery

shell/platform/darwin/ios/framework/Source/FlutterViewController.mm
```diff
@@ -843,6 +887,77 @@ - (void)viewDidDisappear:(BOOL)animated {
...
+  // that is hard to mock since they take/return structs. An alternative approach is to partial mock
+  // the FlutterViewController to make view controller life cycle methods no-op, and insert
+  // this mock into the responder chain.
+  CGFloat scale = [UIScreen mainScreen].scale;
+  _isDuringViewportMetricsInterpolationForRotation = YES;
+
+  CGSize oldSize = self.view.bounds.size;
+  UIEdgeInsets oldPadding = self.view.safeAreaInsets;
+
+  __block double rotationProgress = 0;
```
jmagman, line 926: Alternatively use `NSProgress`, this is fine though.

shell/platform/darwin/ios/framework/Source/FlutterViewController.mm
```diff
@@ -1299,7 +1419,7 @@ - (void)viewDidLayoutSubviews {
   _viewportMetrics.physical_height = viewBounds.size.height * scale;
 
   [self updateViewportPadding];
-  [self updateViewportMetrics];
+  [self updateViewportMetricsIfNeeded:NO];
```
jmagman, line 1422: At the call sites it's not clear what the boolean does (missing named parameters in Objective-C 😞).  How about `-updateViewportMetricsIfNeeded` and `-updateViewportMetricsIfNeededForRotation`?

hellohuanlin:
To share an update: I tried fine-tuning the interpolation count to address the perf issue on complex UI, and found `1/20` to `1/30` second (rather than `1/60` second) update interval worked for flutter gallery. However, this resulted in very "un-smooth" transition which is easily noticeable. 

There is also a good news - Fun story, when i was playing basketball yesterday, a new solution came to my mind, and I call it **delayed swap**: 

In the originally behavior, we swap the width/height immediately before the rotation, resulting in roughly ~4x distortion in the beginning of the rotation. With "delayed swap" solution, we swap the width/height right in the middle of the rotation (i.e. delay the swap for half of the transition duration). This solution is much easier than both "size interpolation" and "snapshot" solution. 

If I am imagining it right, this new "delayed swap" solution should hopefully give us the same result as the "snapshot" solution (reducing ~4x distortion to ~2x), but without the drawback of "snapshot" solution (i.e. it should work well with dynamic content like animation or video). 🤞

PR #40413, opened 2023-03-18
[Impeller] Add anonymous contents

bdero:
This makes it easy for filters and friends to return the subpass callbacks as deferred contents. Also did some drive by cleanup of the Contents interface.


jonahwilliams (approved):
LGTM!

PR #40434, opened 2023-03-19
add final class modifiers to fix dart roll

goderbauer:
Following the lead of https://github.com/dart-lang/sdk/commit/1755f8909264330a207a0e0e44d31f5fec9c489a, this adds the `final` class modifier to all subclasses of `Struct`, `Union`, `Opaque`, and `AbiSpecificInteger` to unblock the dart roll (https://github.com/flutter/engine/pull/40426).


dcharkes (approved):
lgtm.

None of the subtypes of `Opaque` or `Struct` may be sub typed again, so `final` is indeed the right keyword here.

zanderso (approved):

PR #40440, opened 2023-03-19
[web] Fix leaking listener in DOM shim.

joshualitt:


joshualitt:
@eyebrowsoffire ptal.

joshualitt:
Also, FYI there will be a better pattern for functions after we've landed the CL to reify JS types on JS backends. In particular, I think we can make factories that take Dart types and return subtypes of `JSFunction`, which should be a bit safer than just passing around `JSFunction`. Or maybe we can come up with some other pattern, but unfortunately for now raw `JSFunction` is the best we can do because `JSFunction` on JS backends is just a typedef of `Function` and we can't subtype `Function`. We might be able to change the typedef if we feel it is urgent to explore a different model in the short term, but hopefully the reification CL lands soonish.

eyebrowsoffire (approved):
Honestly I like this better than the previous way. I prefer the code that actually sets up and defines the listener to deal with the explicit dart/JS translation and hold onto the JS object. If the code is just holding on to the same dart function, and expecting `toJS` to always give it the same JS function object back, I think that is wrong. To me, `toJS` implies probably *converting* the object, which means that the calling code should more or less assume that it could be returning a new value every time.

eyebrowsoffire (changes requested):
Actually hang on. Don't we have to do this same thing for all the other listeners? `DomEventListener`, e.g.?

joshualitt:
> Actually hang on. Don't we have to do this same thing for all the other listeners? `DomEventListener`, e.g.?

Thankfully, not yet. I didn't change the signature of `EventTarget.addEventListener` yet, that one still takes a Dart function. This is for `DomMediaQueryList.addListener`.

I have a follow on CL where I change `EventTarget.addEventListener` to a similar pattern.

eyebrowsoffire (approved):
Cool, sounds good.

PR #40460, opened 2023-03-20
[Impeller] change default sampler descriptor to use nearest mip level and remove kNone

jonahwilliams:
By default we create mipmaps for every decoded image. Choosing the nearest mip level is the only reasonable default.

### DIFF
--- a/impeller/renderer/backend/gles/sampler_gles.cc
+++ b/impeller/renderer/backend/gles/sampler_gles.cc
@@ -72,10 +72,12 @@ bool SamplerGLES::ConfigureBoundTexture(const TextureGLES& texture,
   if (!target.has_value()) {
     return false;
   }
+  bool has_mips = texture.GetTextureDescriptor().mip_count > 1;
 
   const auto& desc = GetDescriptor();
-  gl.TexParameteri(target.value(), GL_TEXTURE_MIN_FILTER,
-                   ToParam(desc.min_filter, desc.mip_filter));
+  gl.TexParameteri(
+      target.value(), GL_TEXTURE_MIN_FILTER,
+      ToParam(desc.min_filter, has_mips ? desc.mip_filter : std::nullopt));
   gl.TexParameteri(target.value(), GL_TEXTURE_MAG_FILTER,
                    ToParam(desc.mag_filter));
   gl.TexParameteri(target.value(), GL_TEXTURE_WRAP_S,
--- a/impeller/renderer/backend/gles/sampler_gles.cc
+++ b/impeller/renderer/backend/gles/sampler_gles.cc
@@ -19,8 +19,18 @@ bool SamplerGLES::IsValid() const {
   return true;
 }
 
-static GLint ToParam(MinMagFilter minmag_filter, MipFilter mip_filter) {
-  switch (mip_filter) {
+static GLint ToParam(MinMagFilter minmag_filter,
+                     std::optional<MipFilter> mip_filter = std::nullopt) {
+  if (!mip_filter.has_value()) {
+    switch (minmag_filter) {
+      case MinMagFilter::kNearest:
+        return GL_NEAREST;
+      case MinMagFilter::kLinear:
+        return GL_LINEAR;
+    }
+  }
+
+  switch (mip_filter.value()) {
     case MipFilter::kNearest:
       switch (minmag_filter) {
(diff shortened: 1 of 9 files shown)


chinmaygarde (approved):
This is the more accurate default. If there is only one mip level, then the nearest mip index should be zero. I can't think of any other case where we'd need to have mip levels but always sample from zero. Makes me wonder why @bdero preferred trilinear filtering here https://github.com/flutter/engine/pull/40425/files though. So he may have additional context.

chinmaygarde:
Ok, we should remove MipFilter::kNone altogether. It is [not supported in Vulkan](https://github.com/flutter/engine/blob/b4336781d5b8b285beb231074057b228eb50750b/impeller/renderer/backend/vulkan/formats_vk.h#L228) either.

impeller/renderer/backend/gles/sampler_gles.cc
```diff
@@ -74,7 +67,7 @@ bool SamplerGLES::ConfigureBoundTexture(const TextureGLES& texture,
   gl.TexParameteri(target.value(), GL_TEXTURE_MIN_FILTER,
                    ToParam(desc.min_filter, desc.mip_filter));
   gl.TexParameteri(target.value(), GL_TEXTURE_MAG_FILTER,
-                   ToParam(desc.mag_filter, MipFilter::kNone));
+                   ToParam(desc.mag_filter, MipFilter::kNearest));
```
jonahwilliams, line 70: @chinmaygarde what is this supposed to be? This line seems sus before

impeller/renderer/backend/gles/sampler_gles.cc
```diff
@@ -21,13 +21,6 @@ bool SamplerGLES::IsValid() const {
 
 static GLint ToParam(MinMagFilter minmag_filter, MipFilter mip_filter) {
   switch (mip_filter) {
-    case MipFilter::kNone:
```
bdero, line 24: I think drawing with `GL_[NEAREST|LINEAR]_MIPMAP_[NEAREST|LINEAR]` sampler will fail in GLES if the mip level is 0. We'll need to set either `GL_NEAREST` or `GL_LINEAR` when the mip count is 1.

bdero: Please verify that this doesn't break the GLES playgrounds if landing.

impeller/renderer/backend/gles/sampler_gles.cc
bdero: I'm not sure why I changed this when landing the blit pass... This might be debugging leftovers. I'd try changing it back and seeing if any of the GLES playgrounds don't draw anything once [this](https://github.com/flutter/engine/pull/40460#discussion_r1142695900) is addressed.

bdero: Just reread this and realized `kNone` is actually correct here because of https://github.com/flutter/engine/pull/40460#discussion_r1142695900

impeller/renderer/backend/gles/sampler_gles.cc
jonahwilliams: we can retain this behavior for now without the enum variant. Make the mip_filter optional and treat nullopt as if it were none.

bdero: This still needs to be addressed for the min filter. Right now, if a texture has a mip count of 1 (as all subpass and filter-created textures do), it will get a min filter that includes mipmap sampling. This will fail to render on some drivers: 
image

bdero: This still needs to be addressed for the min filter. Right now, if a texture has a mip count of 1 (as all subpass and filter-created textures do), it will get a min filter that includes mipmap sampling. This will fail to render on some drivers: 
image

jonahwilliams: ahh yeah, got it

bdero: The test above is a capture of `Play/DisplayListTest.CanDrawBackdropFilter/OpenGLES`, which was rendering anything a couple of days ago prior to this fix: https://github.com/flutter/engine/pull/40425/files

bdero (approved):
LGTM!

PR #40462, opened 2023-03-20
Rename setters for lifecycle state

gspencergoog:
## Description

This renames `PlatformConfiguration::UpdateLifecycleState` and `RuntimeController::SetLifecycleState` to `PlatformConfiguration::UpdateInitialLifecycleState` and `RuntimeController::SetInitialLifecycleState`, respectively, and some associated private functions.

This is because calling this function really only updates the initial state that the binding loads when it initializes, it doesn't continuously update the lifecycle state.

## Tests
 - No tests changed. This is a rename of the API, and currently doesn't appear to have any direct test dependencies.

**One note:** These appear to both be deprecated APIs. Perhaps it would be better to just remove them?  The only thing is I'm not sure what the deprecation message means by "persistent isolate data", and what the replacement code would be.

### DIFF
--- a/lib/ui/hooks.dart
+++ b/lib/ui/hooks.dart
@@ -67,8 +67,8 @@ void _updateUserSettingsData(String jsonData) {
 }
 
 @pragma('vm:entry-point')
-void _updateLifecycleState(String state) {
-  PlatformDispatcher.instance._updateLifecycleState(state);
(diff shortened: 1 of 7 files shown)


jason-simmons (approved):

Hixie:

PR #40472, opened 2023-03-20
[web] scenelets

yjbanov:
Implement the concept of "scenelet" described in https://flutter.dev/go/flutter-web-scenelets.

Does not include "PointerCaptureRegion" and "Non-painting platform view", as those things have already been implemented. Does not include sceneletization of the semantics tree. That part requires dedicated design and implementation work.

Key concepts:

* `scenelet/scene.dart` provides a new `SceneBuilder` implementation that slices a scene recording into a flat list of `Scenelet` objects. This builder reorders invisible platform views to minimize the number of scenelets necessary to express a scene.
* `scenelet/scenelet.dart` provides the `Scenelet` class which expresses zero or more platform views optionally overlaid by a "render tree".
* A render tree, represented by the `SceneletRenderTree` class, provides "pure Flutter renderable content", which is content renderable into a single canvas. A render tree contains layers, pictures, and textures, but does not contain platform views.
* `scenelet/properties.dart` provides stateless classes that attach layer properties to scene and scenelet layers.
* `scenelet/compositor.dart` renders a `SceneletScene` into a given host element. The current implementation provides a toy renderer good enough for debugging, but not good for production usage. The next step is to integrate the compositor with a real renderer, such as Skwasm (and depending on the Skwasm timeline, perhaps CanvasKit too).
* `scenelet/renderer.dart` provides a `Renderer` implementation that uses the `SceneletSceneBuilder` for constructing scenelet-backed scenes. The current implementation temporarily uses the HTML renderer's rendering primitives until we have enough of Skwasm to integrate a real renderer.


Hixie:
@yjbanov said this was still on his radar as of May 16th.

Hixie:
@yjbanov said this was still on his radar as of Aug 1st, but that this PR may eventually be superseded by another.

yjbanov:
Closed it because it turned out to be easier to implement the same features in CanvasKit and Skwasm directly. Since platform view compositor has to look into `Renderer`-specific objects it is cleaner to have the `Renderer` implementation to own it. The downside is we have to implement it more than once, but hopefully long-term our renderer story will converge so we can delete the duplicates.

PR #40478, opened 2023-03-21
[Linux] fix: make textures thread-safe on linux

Kingtous:
We've encountered some memory corruption when developing `RustDesk`, which is an open-source remote desktop built with Flutter, and it uses the texture API provided by Flutter to provide the remote screen to clients.

Both Windows and macOS work well, except for Linux. Sometimes random crashes will show and print **different** stack traces in these conditions:

- after registering a new texture
- preparing a RGBA data and `markAvaliable`

The runtime often complains `double free or corruption`.
One of the stack traces using lldb-10:
image

We checked the implementation of calling textures API on Linux several times, but have no luck. 

We've noticed that on Linux, textures are organized by `g_hash_table`, which is not the thread-safe hashtable and we should ensure synchronization on different threads. Also, we noticed the textures map on Windows already has the synchronization (for example https://github.com/flutter/engine/blob/2c342a8af94ee9ef19bbb222ac003c3340b9be00/shell/platform/windows/flutter_windows_texture_registrar.cc#L87-L90). So the implementation on Linux lacks this synchronization on textures management.

We add the synchronization by declaring a `GMutex`, which is a mutex provided by GTK, in this PR. It can make `g_hash_table` thread-safe in different threads(ui, raster, etc).

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/shell/platform/linux/fl_texture_registrar.cc
+++ b/shell/platform/linux/fl_texture_registrar.cc
@@ -207,6 +207,18 @@ fl_texture_registrar_unregister_texture(FlTextureRegistrar* self,
                                                                   texture);
 }
 
+gboolean fl_texture_registrar_get_textures_locked(
+    FlTextureRegistrar* registrar) {
+  g_return_val_if_fail(FL_IS_TEXTURE_REGISTRAR_IMPL(registrar), FALSE);
+  FlTextureRegistrarImpl* self = FL_TEXTURE_REGISTRAR_IMPL(registrar);
+  gboolean locked = g_mutex_trylock(&self->textures_mutex);
+  if (locked) {
+    g_mutex_unlock(&self->textures_mutex);
+    return FALSE;
+  }
+  return TRUE;
+}
+
 FlTextureRegistrar* fl_texture_registrar_new(FlEngine* engine) {
   FlTextureRegistrarImpl* self = FL_TEXTURE_REGISTRAR_IMPL(
       g_object_new(fl_texture_registrar_impl_get_type(), nullptr));
--- a/shell/platform/linux/fl_texture_registrar.cc
+++ b/shell/platform/linux/fl_texture_registrar.cc
@@ -208,13 +208,6 @@ fl_texture_registrar_unregister_texture(FlTextureRegistrar* self,
                                                                   texture);
 }
 
-guint fl_texture_registrar_get_textures_table_size(
-    FlTextureRegistrar* registrar) {
-  g_return_val_if_fail(FL_IS_TEXTURE_REGISTRAR_IMPL(registrar), 0);
-  FlTextureRegistrarImpl* self = FL_TEXTURE_REGISTRAR_IMPL(registrar);
-  return g_hash_table_size(self->textures);
-}
-
 FlTextureRegistrar* fl_texture_registrar_new(FlEngine* engine) {
   FlTextureRegistrarImpl* self = FL_TEXTURE_REGISTRAR_IMPL(
       g_object_new(fl_texture_registrar_impl_get_type(), nullptr));
--- a/shell/platform/linux/fl_texture_registrar.cc
+++ b/shell/platform/linux/fl_texture_registrar.cc
@@ -34,6 +34,9 @@ struct _FlTextureRegistrarImpl {
   // plugins.  The keys are directly stored int64s. The values are stored
   // pointer to #FlTexture.  This table is freed by the responder.
   GHashTable* textures;
+
+  // The mutex guard to make `textures` thread-safe.
+  GMutex textures_mutex;
 };
 
 static void fl_texture_registrar_impl_iface_init(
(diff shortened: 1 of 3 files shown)


shell/platform/linux/fl_texture_registrar.cc
```diff
@@ -201,5 +229,8 @@ FlTextureRegistrar* fl_texture_registrar_new(FlEngine* engine) {
   self->engine = engine;
   g_object_weak_ref(G_OBJECT(engine), engine_weak_notify_cb, self);
 
+  // Initialize the mutex for textures.
+  g_mutex_init(&self->textures_mutex);
```
robert-ancell, line 233: This should be in `fl_texture_registrar_impl_init`.

shell/platform/linux/fl_texture_registrar.cc
```diff
@@ -57,21 +60,27 @@ static void engine_weak_notify_cb(gpointer user_data,
   self->engine = nullptr;
 
   // Unregister any textures.
+  g_mutex_lock(&self->textures_mutex);
   g_autoptr(GHashTable) textures = self->textures;
   self->textures = g_hash_table_new_full(g_direct_hash, g_direct_equal, nullptr,
                                          g_object_unref);
   g_hash_table_remove_all(textures);
+  g_mutex_unlock(&self->textures_mutex);
+  g_mutex_clear(&self->textures_mutex);
```
robert-ancell, line 69: The mutex shouldn't be cleared here - it will be done in dispose. It may still be accessed before that.

shell/platform/linux/fl_texture_registrar_test.cc
```diff
@@ -104,3 +104,24 @@ TEST(FlTextureRegistrarTest, MarkTextureFrameAvailable) {
   EXPECT_TRUE(
       fl_texture_registrar_mark_texture_frame_available(registrar, texture));
 }
+
+// Test that mutex should be unlocked after calling register_texture.
+TEST(FlTextureRegistrarTest, RegistrarRegisterTextureMutex) {
```
robert-ancell, line 109: I would drop this test - the mutex is not an external detail of this object. To do this test we've added a method that's only used for testing. @cbracken WDYT.

robert-ancell (changes requested):
Change looks good, thanks! Just some details.

shell/platform/linux/fl_texture_registrar_test.cc
Kingtous: I've added a function called `fl_texture_registrar_get_textures_locked` in this PR to test whether the mutex is locked or not. I don't know whether it can be the method that's only used for testing.

robert-ancell: That's the one I'm referring to. The only way this test is going to fail is if the clear mutex unlock code is removed, which seems like an unlikely thing to occur. I think a better test would be to access the object with multiple threads as that's the failure case we have now and don't want to regress to. The test wouldn't always fail but it might fail sometimes which would indicate a regression.

shell/platform/linux/fl_texture_registrar.cc
```diff
@@ -191,6 +208,13 @@ fl_texture_registrar_unregister_texture(FlTextureRegistrar* self,
                                                                   texture);
 }
 
+guint fl_texture_registrar_get_textures_table_size(
+    FlTextureRegistrar* registrar) {
+  g_return_val_if_fail(FL_IS_TEXTURE_REGISTRAR_IMPL(registrar), 0);
+  FlTextureRegistrarImpl* self = FL_TEXTURE_REGISTRAR_IMPL(registrar);
+  return g_hash_table_size(self->textures);
```
robert-ancell, line 215: Probably should use the mutex here for consistency.

shell/platform/linux/fl_texture_registrar_private.h
```diff
@@ -33,6 +33,18 @@ FlTextureRegistrar* fl_texture_registrar_new(FlEngine* engine);
 FlTexture* fl_texture_registrar_lookup_texture(FlTextureRegistrar* registrar,
                                                int64_t texture_id);
 
+/**
+ * fl_texture_registrar_get_textures_locked:
```
robert-ancell, line 37: Doesn't match the new function name.

shell/platform/linux/fl_texture_registrar_test.cc
```diff
@@ -104,3 +114,22 @@ TEST(FlTextureRegistrarTest, MarkTextureFrameAvailable) {
...
+
+  for (uint64_t t = 0; t < kThreadCount; t++) {
+    EXPECT_EQ(pthread_create(&threads[t], NULL, add_mock_texture_to_registrar,
+                             (void*)registrar),
+              0);
+  }
+  for (uint64_t t = 0; t < kThreadCount; t++) {
+    pthread_join(threads[t], NULL);
+  };
+  EXPECT_EQ(fl_texture_registrar_get_textures_table_size(registrar),
```
robert-ancell, line 133: You could just use `fl_texture_registrar_lookup_texture` to look up each texture - they will be numbered 1-N. That will avoid adding `fl_texture_registrar_get_textures_table_size`.

robert-ancell (changes requested):

shell/platform/linux/fl_texture_registrar_test.cc
Kingtous: Quite a good way! I will use it later.

Kingtous: I removed `fl_texture_registrar_get_textures_table_size` and use `fl_texture_registrar_lookup_texture` to check each texture now. :)

robert-ancell (approved):
Changes look great now, thanks!

PR #40493, opened 2023-03-21
Listen to `WM_CLOSE` message on Windows to allow framework to cancel exit

yaakovschectman:
Register the platform handler to receive top level window messages and allow the framework to cancel an exit. When the user closes the application window, `WM_CLOSE` is sent to the top level window, which triggers a request from the engine to the framework, which it can pass to any registered observers before returning the result to the engine.

Addresses windows portion of https://github.com/flutter/flutter/issues/30735
Design doc: https://flutter.dev/go/application-lifecycle-events

## Pre-launch Checklist

- [ ] I signed the [CLA].


shell/platform/windows/platform_handler.cc
```diff
@@ -447,4 +448,14 @@ void PlatformHandler::HandleMethodCall(
   }
 }
 
+// static
+bool PlatformHandler::WindowProcCallback(HWND hwnd, UINT msg, WPARAM wpar, LPARAM lpar, void* user_data, LRESULT* result) {
+  PlatformHandler* that = static_cast<PlatformHandler*>(user_data);
+  if (msg == WM_CLOSE) {
```
gspencergoog, line 454: This blocks the window from closing, and assumes that closing the window is meant to close the app. What happens when we eventually have multiple windows, won't we have to check to make sure this is the last window before asking? Also, is there a case where the developer might want the window to actually close, but then keep running the app until they are ready to exit?

gspencergoog:
> This PR has no tests yet. If you are aware of how to test the behavior of closing the top-level window from within the engine, please inform me.

In the test, could you intercept the WM_QUIT message in the message loop so that when `PostQuitMessage` is called, it has no effect until you want to exit the test app?

I think you could also write this as an integration test that logs when it is about to quit, and run the app in a separate process and see if it quits. We might not have the infrastructure set up for that, though, so I'm not sure how big a lift that is.

shell/platform/windows/platform_handler.cc
yaakovschectman: It would require a check if it is the last window, but I do not think that check is possible to write yet.

shell/platform/windows/platform_handler.cc
```diff
@@ -227,6 +227,7 @@ PlatformHandler::PlatformHandler(
       return std::make_unique<ScopedClipboard>();
     };
   }
+  FlutterDesktopPluginRegistrarRegisterTopLevelWindowProcDelegate(engine_->GetRegistrar(), PlatformHandler::WindowProcCallback, this);
```
loic-sharma, line 230: Instead of using the plugin API, could we instead use `WindowProcDelegateManager::RegisterTopLevelWindowProcDelegate` directly?

loic-sharma, line 230: Plugins should be able to override the Windows embedder's behavior. In other words, plugins' top-level window handlers should run before the Windows embedder's top-level window handlers. Please add a test to cover this behavior.

yaakovschectman:
@gspencergoog The issue I see is that the window that receives `WM_CLOSE` in a flutter application is not part of the engine; it's part of the runner application from the framework side. So there is the question of how/where to pass the message in the first place.

shell/platform/windows/platform_handler.cc
loic-sharma, line 230: I'm not sure I like having Windows messaging handling in `PlatformHandler`. Personally, I'd prefer if `PlatformHandler` focused on just platform channel messaging (receiving and sending messages to the framework).

Could we move this logic to a new class whose top-level window message handler is registered by `FlutterWindowsEngine`?

shell/platform/windows/platform_handler.cc
gspencergoog: Could you use a Windows API to find how many open windows an app has? I think that even when we can enumerate Flutter-controlled windows in an app, we won't want to actually exit the app unless this is the last window, including windows that aren't Flutter controlled.

shell/platform/windows/platform_handler.cc
yaakovschectman: That makes sense. Though do we have a way of sending a window message to a top level window for our tests? I'm not aware of anywhere we already do something like this.

gspencergoog: It doesn't change your valid point, and I realize that's just an example, but saving data on shutdown is an anti-pattern: there are plenty of ways for an app to close without sending a WM_CLOSE that can be handled, so we don't really want to encourage people to do that (it's the reason we eliminated the `exiting` state from the app state design doc).

I think a better example would be so that a plugin could abort the close if it wanted to.

loic-sharma:
> If you are aware of how to test the behavior of closing the top-level window from within the engine, please inform me.

Can you craft Windows messages, similar to how we craft [Windows messages to the Flutter view](https://github.com/flutter/engine/blob/87ddd910f1ad02bdd8dc0ac31897ec0ebab36a0b/shell/platform/windows/window_unittests.cc#L42)? You can use this logic to send top-level windows messages to the Windows embedder:

https://github.com/flutter/engine/blob/87ddd910f1ad02bdd8dc0ac31897ec0ebab36a0b/shell/platform/windows/flutter_windows.cc#L117-L120

shell/platform/windows/window_top_level_message_handler.cc
```diff
@@ -0,0 +1,74 @@
...
+}
+
+static BOOL CALLBACK WindowEnumCallback(HWND hwnd, LPARAM user_data) {
+  int64_t& count = *static_cast<int64_t*>(reinterpret_cast<void*>(user_data));
+  count++;
+  return true;
+}
+
+bool WindowTopLevelMessageHandler::IsLastWindowOfProcess() {
+  // TODO(schectman): This currently is not working. Test apps have 3 "windows" to their process.
```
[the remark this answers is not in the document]
gspencergoog: Oh, I see. Yeah, we'd have to only count top level windows.

shell/platform/windows/window_top_level_message_handler.cc
```diff
@@ -0,0 +1,74 @@
...
+    case WM_CLOSE:
+      if (IsLastWindowOfProcess()) {
+        engine_.RequestApplicationQuit(kExitTypeCancelable, wpar);
+      }
+      return true;
+  }
+  return false;
+}
+
+static BOOL CALLBACK WindowEnumCallback(HWND hwnd, LPARAM user_data) {
```
gspencergoog, line 37: Could you call `GetParent` on the window here and only count the ones that return NULL?

shell/platform/windows/fixtures/main.dart
```diff
@@ -86,6 +87,42 @@ void alertPlatformChannel() async {
...
+    ByteData responseData = ByteData.sublistView(Uint8List.fromList(utf8.encode(jsonString)));
+    callback(responseData);
+    closed.complete(data);
+  });
+  await closed.future;
+
+  // From here down, nothing should be called, because the application will have already been closed.
+  final Completer<ByteData?> exited = Completer<ByteData?>();
+  const String jsonString = '{"method":"System.exitApplication","args":{"type":"required","exitCode":0}}';
+  ui.PlatformDispatcher.instance.sendPlatformMessage('flutter/platform', ByteData.sublistView(Uint8List.fromList(utf8.encode(jsonString))), (ByteData? reply) {exited.complete(reply);});
```
loic-sharma, line 104: Please break this up into multiple lines for readability

shell/platform/windows/flutter_windows_engine.cc
```diff
@@ -203,6 +204,7 @@ FlutterWindowsEngine::FlutterWindowsEngine(
       std::make_unique<FlutterWindowsTextureRegistrar>(this, gl_procs_);
   surface_manager_ = AngleSurfaceManager::Create();
   window_proc_delegate_manager_ = std::make_unique<WindowProcDelegateManager>();
+  window_proc_delegate_manager_->RegisterTopLevelWindowProcDelegate(TopLevelWindowProc, static_cast<void*>(this));
```
loic-sharma, line 207: Consider making this a lambda like the `kAccessibilityChannelName` message callback above. It's a little bit less verbose than having a separate method.

shell/platform/windows/flutter_windows_engine.cc
```diff
@@ -751,4 +753,24 @@ void FlutterWindowsEngine::HandleAccessibilityMessage(
...
 
+// static
+bool FlutterWindowsEngine::TopLevelWindowProc(HWND hwnd, UINT msg, WPARAM wpar, LPARAM lpar, void* user_data, LRESULT* result) {
+  if (user_data == nullptr) {
+    return false;
+  }
+  FlutterWindowsEngine* that = static_cast<FlutterWindowsEngine*>(user_data);
+  if (!that->top_level_handler_) {
+    return false;
+  }
```
loic-sharma, line 764: Should the null checks be assertions instead? Are there expected scenarios where they would be null?

shell/platform/windows/window_top_level_message_handler.h
```diff
@@ -0,0 +1,33 @@
...
+
+#include <Windows.h>
+
+namespace flutter {
+
+class FlutterWindowsEngine;
+
+class WindowTopLevelMessageHandler {
+ public:
+  WindowTopLevelMessageHandler(FlutterWindowsEngine& engine);
```
loic-sharma, line 18: Could this accept a `PlatformHandler` instead of the engine? That would allow us to remove the `RequestApplicationQuit` from the engine

shell/platform/windows/window_top_level_message_handler.h
```diff
@@ -0,0 +1,33 @@
...
+
+#include <cstdint>
+
+#include <Windows.h>
+
+namespace flutter {
+
+class FlutterWindowsEngine;
+
+class WindowTopLevelMessageHandler {
```
loic-sharma, line 16: This name doesn't capture that this not only responds to Windows messages but also posts them.

shell/platform/windows/window_top_level_message_handler.h
loic-sharma: Alternatively, I would consider making this class agnostic to Flutter. You could introduce a `WindowsLifecycleDelegate` with methods like `OnExitRequested`. The `FlutterWindowsEngine` could implement this lifecycle delegate and call the `PlatformHandler::RequestAppExit`.

shell/platform/windows/window_top_level_message_handler.cc
```diff
@@ -0,0 +1,74 @@
...
+#include <TlHelp32.h>
+#include <Windows.h>
+#include <WinUser.h>
+#include <tchar.h>
+
+#include "flutter/shell/platform/windows/flutter_windows_engine.h"
+
+namespace flutter {
+
+static constexpr char kExitTypeCancelable[] = "cancelable";
```
loic-sharma, line 16: Could we move this string out of this class? Ideally all strings for Flutter framework messages are contained in the `PlatformHandler`. Perhaps introduce an enum or boolean?

shell/platform/windows/window_top_level_message_handler.h
yaakovschectman: The former can be done so long as we expose the engine's platform handler, basically replacing `RequestApplicationQuit` with a getter for it. In the case of the latter, what is the advantage of making it agnostic to Flutter?

shell/platform/windows/fixtures/main.dart
```diff
@@ -86,6 +87,42 @@ void alertPlatformChannel() async {
   ui.PlatformDispatcher.instance.sendPlatformMessage('flutter/accessibility', byteData, (ByteData? _){});
 }
 
+@pragma('vm:entry-point')
+void exitTestExit() async {
+  final Completer<ByteData?> closed = Completer<ByteData?>();
+  ui.channelBuffers.setListener('flutter/platform', (ByteData? data, ui.PlatformMessageResponseCallback callback) async {
+    const String jsonString = '[{"response":"exit"}]';
```
cbracken, line 94: Optional: rather than hand-writing JSON, consider doing something like:
```dart
final String jsonString = json.encode(<String, dynamic>{"response": "exit"});
```

Mostly because that allows the compiler to catch Dart syntax errors and json.encode to catch JSON syntax errors.

shell/platform/windows/fixtures/main.dart
```diff
@@ -86,6 +87,42 @@ void alertPlatformChannel() async {
...
+    ByteData responseData = ByteData.sublistView(Uint8List.fromList(utf8.encode(jsonString)));
+    callback(responseData);
+    closed.complete(data);
+  });
+  await closed.future;
+
+  // From here down, nothing should be called, because the application will have already been closed.
+  final Completer<ByteData?> exited = Completer<ByteData?>();
+  const String jsonString = '{"method":"System.exitApplication","args":{"type":"required","exitCode":0}}';
```
cbracken, line 103: Similarly here, using `json.encode` to encode a map would allow the map to be laid out the way it's structured:
```dart
final String jsonString = json.encode(<String, dynamic>{
    "method": "System.exitApplication",
    "args": <String, dynamic>{
      "type": "required",
      "exitCode": 0,
    }
  };

File: shell/platform/windows/windows_lifecycle_manager.h [hunk 13]
```diff
@@ -0,0 +1,33 @@
...
+
+#include <Windows.h>
+
+namespace flutter {
+
+class FlutterWindowsEngine;
+
+class WindowsLifecycleManager {
+ public:
+  WindowsLifecycleManager(FlutterWindowsEngine& engine);
```
cbracken, line 18: Since we've got a reference cycle between `FlutterWindowsEngine` and `WindowsLifecycleManager` this should be a weak pointer (just a raw `FlutterWindowsEngine*` pointer) to indicate that to the reader.

shell/platform/windows/windows_lifecycle_manager.cc
```diff
@@ -0,0 +1,84 @@
...
+      }
+      return true;
+  }
+  return false;
+}
+
+static BOOL CALLBACK WindowEnumCallback(HWND hwnd, LPARAM user_data) {
+  HWND parent = GetParent(hwnd);
+  if (parent == NULL) {
+    int64_t& count = *static_cast<int64_t*>(reinterpret_cast<void*>(user_data));
```
cbracken, line 44: Just `reinterpret_cast<int64_t*>(user_data)` directly.

shell/platform/windows/windows_lifecycle_manager.cc
```diff
@@ -0,0 +1,84 @@
...
+  if (!Thread32First(thread_snapshot, &thread)) {
+    DWORD error_num = GetLastError();
+    char msg[256];
+    FormatMessageA(FORMAT_MESSAGE_FROM_SYSTEM | FORMAT_MESSAGE_IGNORE_INSERTS,
+                   NULL, error_num, MAKELANGID(LANG_NEUTRAL, SUBLANG_DEFAULT),
+                   msg, 256, nullptr);
+    FML_LOG(ERROR) << "Failed to get thread(" << error_num << "): " << msg;
+    CloseHandle(thread_snapshot);
+    return true;
+  }
```
cbracken, line 69: I'd be tempted to extract this code to a private method like `GetFirstThreadForProcess` for readability. Could probably be static since I don't think you're using any fields of this struct here.

cbracken: Ah I see looking at this, you need to keep `thread_snapshot` alive while you poke through the windows list for the thread.

You could probably extract a tiny class like `ThreadSnapshot` with a method like `std::optional<THREADENTRY32> GetFirstThread()` that returns the first thread and have `ThreadSnapshot` close the snapshot handle in its destructor, which has the nice advantage of ensuring that that happens on return without having to manually clean up before each path out of this function.

yaakovschectman: I don't think I can put all of this into a separate function, as we still need to `CloseHandle(thread_snapshot)` after enumerating all threads, but getting the first thread can be extracted.

shell/platform/windows/windows_lifecycle_manager.cc
```diff
@@ -0,0 +1,84 @@
...
+
+  int num_windows = 0;
+  do {
+    if (thread.th32OwnerProcessID == pid) {
+      EnumThreadWindows(
+          thread.th32ThreadID, WindowEnumCallback,
+          reinterpret_cast<LPARAM>(static_cast<void*>(&num_windows)));
+    }
+  } while (Thread32Next(thread_snapshot, &thread));
```
cbracken: If you broke out a private static function:
```cc
/// Returns the number of top-level 
static int64_t WindowCountForThread(LPTHREADENTRY32 thread) {
  int64_t num_windows = 0;
  EnumThreadWindows(
    thread.th32ThreadID,
    [](LPARAM lparam) {
      HWND parent = GetParent(hwnd);
      if (parent == nullptr) {
        int64_t* count = reinterpret_cast<int64_t*>(lparam);
        (*count)++;
      }
    },
    reinterpret_cast<LPARAM>(&num_windows)  // not sure if the cast is necessary
  );
  return num_windows;
}
```

Then code could then just be:
```cc
for (auto LPTHREADENTRY32 thread : snapshot.GetThreads()) {
  if (WindowCountForThread(thread) > 0) {
    return true;
  }
}
return false;
```

which is slightly more efficient since it bails out at the first top-level window it finds.

shell/platform/windows/windows_lifecycle_manager.h
```diff
@@ -0,0 +1,33 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef FLUTTER_SHELL_PLATFORM_WINDOWS_WINDOW_TOP_LEVEL_MESSAGE_HANDLER_H_
+#define FLUTTER_SHELL_PLATFORM_WINDOWS_WINDOW_TOP_LEVEL_MESSAGE_HANDLER_H_
+
+#include <cstdint>
+
+#include <Windows.h>
```
cbracken, line 10:
```suggestion
#include <Windows.h>

#include <cstdint>
```

shell/platform/windows/flutter_windows_engine.h
```diff
@@ -254,6 +255,12 @@ class FlutterWindowsEngine {
   // Updates accessibility, e.g. switch to high contrast mode
   void UpdateAccessibilityFeatures(FlutterAccessibilityFeature flags);
 
+  // Called when the application quits in response to a quit request.
+  void OnQuit(int64_t exit_code);
```
cbracken, line 259: Is this the right type? 

Looking at [ExitProcess](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-exitprocess?redirectedfrom=MSDN), exit codes are [UINT](https://learn.microsoft.com/en-us/windows/win32/winprog/windows-data-types#uint) which are an unsigned [INT](https://learn.microsoft.com/en-us/windows/win32/winprog/windows-data-types#int)
which claims to be a 32-bit integer.

shell/platform/windows/platform_handler.h
```diff
@@ -33,6 +34,10 @@ class PlatformHandler {
 
   virtual ~PlatformHandler();
 
+  // Send a request to the framework to test if a cancelable exit request
+  // should be canceled or honored.
+  virtual void RequestAppExit(const std::string& exit_type, int64_t exit_code);
```
cbracken, line 39: Is the exit type an enum on the Dart side? If so, we should probably decode the incoming lifecycle message to an enum and use that in these methods?

shell/platform/windows/windows_lifecycle_manager.h
```diff
@@ -0,0 +1,33 @@
...
+
+namespace flutter {
+
+class FlutterWindowsEngine;
+
+class WindowsLifecycleManager {
```
cbracken, line 16: Add a doc comment for the class.

While [this bit of the style guide](https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#documentation-dartdocs-javadocs-etc) is somewhat Dart-specific, I think the overall recommendations for the sort of content to include are pretty good.

shell/platform/windows/windows_lifecycle_manager.cc
```diff
@@ -0,0 +1,84 @@
...
+#include "flutter/shell/platform/windows/flutter_windows_engine.h"
+
+namespace flutter {
+
+WindowsLifecycleManager::WindowsLifecycleManager(FlutterWindowsEngine& engine)
+    : engine_(engine) {}
+
+WindowsLifecycleManager::~WindowsLifecycleManager() {}
+
+void WindowsLifecycleManager::Quit(int64_t exit_code) {
```
cbracken, line 21: Since this method isn't modifying any class state, consider making it `const`.

shell/platform/windows/windows_lifecycle_manager.cc
cbracken: See my note on doing this in the destructor, which gives a nice RAII pattern and avoids the need for the close call in user code altogether.

shell/platform/windows/platform_handler.h
```diff
@@ -22,10 +22,14 @@ namespace flutter {
 class FlutterWindowsEngine;
 class ScopedClipboardInterface;
 
+enum ExitType {
```
cbracken, line 25:
```suggestion
enum class ExitType {
```
In order to avoid the implicit type conversion issues present in traditional enums.

shell/platform/windows/windows_lifecycle_manager.cc
```diff
@@ -88,6 +88,18 @@ class ThreadSnapshot {
...
 };
 
+static int64_t NumWindowsOfThread(const THREADENTRY32& thread) {
+  int64_t num_windows = 0;
+  EnumThreadWindows(thread.th32ThreadID, [](HWND hwnd, LPARAM lparam){
+    int64_t* windows_ptr = reinterpret_cast<int64_t*>(lparam);
+    if (GetParent(hwnd) == NULL) {
+      (*windows_ptr)++;
+    }
+    return *windows_ptr <= 1 ? 1 : 0; // Must return BOOL (i.e. int), not bool.
```
cbracken, line 98:
```suggestion
    return *windows_ptr <= 1 ? TRUE : FALSE;
```

shell/platform/windows/windows_lifecycle_manager.cc
```diff
@@ -0,0 +1,126 @@
...
+  ThreadSnapshot() {
+    thread_snapshot_ = CreateToolhelp32Snapshot(TH32CS_SNAPTHREAD, 0);
+  }
+  ~ThreadSnapshot() {
+    if (thread_snapshot_ != INVALID_HANDLE_VALUE) {
+      CloseHandle(thread_snapshot_);
+    }
+  }
+
+  std::optional<THREADENTRY32> GetFirstThread() {
```
cbracken, line 61: Updated the suggestion a bit later but another way to do this that would avoid the need for the get first, get next pair (and the optional) would be to have this return a std::vector<THREADENTRY32> of all the threads. 

What you've got is great -- I leave it entirely up to you as to which you think is the more readable option.

cbracken (approved):
This looks good to me! @gspencergoog & @loic-sharma for a second pass.

shell/platform/windows/platform_handler.h
```diff
@@ -22,6 +22,11 @@ namespace flutter {
 class FlutterWindowsEngine;
 class ScopedClipboardInterface;
 
+enum class ExitType {
```
loic-sharma, line 25: Please add a comment that this must be kept in sync with `platform_handler.cc`'s `exit_type_names`

shell/platform/windows/platform_handler.cc
```diff
@@ -205,6 +203,18 @@ int ScopedClipboard::SetString(const std::wstring string) {
 
 }  // namespace
 
+static const char* exit_type_names[] = {"required", "cancelable"};
```
loic-sharma, line 206: Please add a comment that this must be kept in sync with `platform_handler.h`'s `ExitType`

loic-sharma, line 206:
```suggestion
static const char* kExitTypeNames[] = {"required", "cancelable"};
```

loic-sharma: Alternatively, turn this into a switch on the enum and inline that into `RequestAppExit`

shell/platform/windows/platform_handler.cc
```diff
@@ -372,30 +382,31 @@ void PlatformHandler::SystemExitApplication(
...
+      kExitTypeKey, std::string(exit_type_names[static_cast<int>(exit_type)]),
+      args->GetAllocator());
   channel_->InvokeMethod(kRequestAppExitMethod, std::move(args),
                          std::move(callback));
 }
 
 void PlatformHandler::RequestAppExitSuccess(const rapidjson::Document* result,
-                                            int64_t exit_code) {
+                                            UINT exit_code) {
   const std::string& exit_type = result[0][kExitResponseKey].GetString();
```
loic-sharma, line 402: We cannot assume that documents are well-formed as that can cause crashes if that assumption is incorrect. We should ignore malformed messages but log an error.

shell/platform/windows/platform_handler.cc
```diff
@@ -405,8 +416,9 @@ void PlatformHandler::HandleMethodCall(
   if (method.compare(kExitApplicationMethod) == 0) {
```
loic-sharma, line 416: We cannot assume that the JSON is well-formed as that can cause crashes. Please use logic similar to `Clipboard.setData` to validate the arguments: https://github.com/flutter/engine/blob/686552f9444af6b169fbe522fc1987d99be88586/shell/platform/windows/platform_handler.cc#L429-L438

shell/platform/windows/windows_lifecycle_manager.cc
```diff
@@ -0,0 +1,128 @@
...
+
+bool WindowsLifecycleManager::WindowProc(HWND hwnd,
+                                         UINT msg,
+                                         WPARAM wpar,
+                                         LPARAM lpar,
+                                         LRESULT* result) {
+  switch (msg) {
+    case WM_CLOSE:
+      if (IsLastWindowOfProcess()) {
+        engine_->RequestApplicationQuit(ExitType::cancelable, wpar);
```
loic-sharma, line 33: Is using `wParam` correct? The win32 [docs](https://learn.microsoft.com/en-us/windows/win32/winmsg/wm-close) state this parameter is not used. Should we use `0` instead?

shell/platform/windows/windows_lifecycle_manager.cc
```diff
@@ -0,0 +1,128 @@
...
+      }
+      return true;
+  }
+  return false;
+}
+
+static BOOL CALLBACK WindowEnumCallback(HWND hwnd, LPARAM user_data) {
```
loic-sharma, line 40: Maybe I'm missing something, but this code looks unused.

shell/platform/windows/windows_lifecycle_manager.cc
```diff
@@ -0,0 +1,128 @@
...
+      return false;
+    }
+    return Thread32Next(thread_snapshot_, &thread);
+  }
+
+ private:
+  HANDLE thread_snapshot_;
+};
+
+static int64_t NumWindowsOfThread(const THREADENTRY32& thread) {
```
loic-sharma, line 90:
```suggestion
static int64_t NumWindowsForThread(const THREADENTRY32& thread) {
```

shell/platform/windows/windows_lifecycle_manager.cc
```diff
@@ -0,0 +1,128 @@
...
+    int64_t& count = *reinterpret_cast<int64_t*>(user_data);
+    count++;
+  }
+  return true;
+}
+
+class ThreadSnapshot {
+ public:
+  ThreadSnapshot() {
+    thread_snapshot_ = CreateToolhelp32Snapshot(TH32CS_SNAPTHREAD, 0);
```
loic-sharma, line 52: Chris suggested prefixing global win32 API calls with `::`. That would include `CreateToolhelp32Snapshot`, `CloseHandle`, `Thread32First`, etc :)

shell/platform/windows/flutter_windows_engine.h
```diff
@@ -395,6 +402,9 @@ class FlutterWindowsEngine {
   // Wrapper providing Windows registry access.
   std::unique_ptr<WindowsRegistry> windows_registry_;
 
+  // Handler for top level window messages.
```
loic-sharma, line 405: Consider something like:

```suggestion
  // Listens and notifies the framework of app lifecycle events like shutdown.
```

I don't feel strongly, feel free to keep it as is

shell/platform/windows/testing/engine_modifier.h
```diff
@@ -66,6 +66,10 @@ class EngineModifier {
   // restart. This resets the keyboard's state if it exists.
   void Restart() { engine_->OnPreEngineRestart(); }
 
+  void SetTopLevelHandler(std::unique_ptr<WindowsLifecycleManager>&& handler) {
```
loic-sharma, line 69:
```suggestion
  void SetLifecycleManager(std::unique_ptr<WindowsLifecycleManager>&& handler) {
```

shell/platform/windows/windows_lifecycle_manager.cc
yaakovschectman: I think I confused it with `WM_QUIT`.

shell/platform/windows/platform_handler.cc
```diff
@@ -205,6 +206,20 @@ int ScopedClipboard::SetString(const std::wstring string) {
 
 }  // namespace
 
+// Indicates whether an exit request may be canceled by the framework.
+// These values must be kept in sync with ExitType in platform_handler.h
+static const char* kExitTypeNames[] = {"required", "cancelable"};
```
gspencergoog, line 211: Maybe move this closer to (or inside of) `RequestAppExit`, since that's the only place it is used.

You could also define the contents of the array in terms of `kExitTypeCancelable` and `kExitTypeRequired`, since they're constexprs. You might also be able to make this a constexpr. It would be one less thing to keep in sync.

yaakovschectman: I can use the constants in the array, but it does not appear that this can be set to a `constexpr`.

gspencergoog (approved):

Thanks Yaakov!

shell/platform/windows/windows_lifecycle_manager.cc
loic-sharma: I also slightly prefer returning a `std::vector` but agree both options are good :)

loic-sharma (approved):
Great work! Don't forget to address the `wParam` exit code issue before merging :)

PR #40497, opened 2023-03-21
[impeller] disable opacity peephole due to stencil issues

jonahwilliams:
This causes rendering issues on wonderous that I missed. I think we're screwing up the stencil depth checks when collapsing


bdero (approved):
LGTM.

Stencil coverage is one of the most annoying things to debug at the moment and I've been working on some tooling to help with this. Right now I usually end up doing a combination of print debugging and conditional breaks. :/

PR #40506, opened 2023-03-21
[Impeller] Refactor Cubic/Quad tests to make sure all threads reach barrier()

dnfield:
Also does some refactoring to move code under test to a shared file for future use in the more generalized path polyline generator.

This is another part of addressing https://github.com/flutter/flutter/issues/122828. 

The last piece left is to split invocations up into chunks that fit into the `threadExecutionWidth`.

I don't have a test with this change, but it will be tested when we re-enable the tests from the linked issue on Intel machines. Ideally we'd have some kind of linting or fuzzing for this, but there's no good existing solution for that right now.

### DIFF
--- a/impeller/compiler/shader_lib/impeller/path.glsl
+++ b/impeller/compiler/shader_lib/impeller/path.glsl
@@ -28,9 +28,18 @@ struct CubicData {
   vec2 p2;
 };
 
-struct Position {
-  uint index;
-  uint count;
+struct QuadDecomposition {
+  float a0;
+  float a2;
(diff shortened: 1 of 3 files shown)


jonahwilliams (approved):
🤔 RSLGTM

dnfield:
Probably should've added in the description - if some thread fails to reach `barrier()`, the shader is just hanging on Intel.

zanderso:
Testing is something that we should discuss after the next branch cut and before the GPU compute code goes live.

PR #40508, opened 2023-03-22
(Reland) "[gn + codesign] codesign metadata for FlutterMacOS.framework.zip"

XilaiZhang:
The initial PR https://github.com/flutter/engine/pull/35707 was reverted due to lint check failures. Lint check is now fixed and added to presubmit runs when .py files are modified.

This is the reland of the initial PR to add codesign metadata for FlutterMacOS.framework.zip


christopherfujino (approved):
LGTM

PR #40509, opened 2023-03-22
[Impeller] Take subgroup size into account when prefix summing

dnfield:
Fixes https://github.com/flutter/flutter/issues/122828

On Intel hardware, the subgroup size is coming out as 16 for the quad shader in particular. There are, however, 26 quads to compute. The test works fine on M1 hardware where the subgroup size stays at 32 for this shader, but fails on Intel.

This change makes both the test shaders safe against smaller threadgroup sizes.

### DIFF
--- a/impeller/compiler/shader_lib/impeller/path.glsl
+++ b/impeller/compiler/shader_lib/impeller/path.glsl
@@ -28,9 +28,18 @@ struct CubicData {
   vec2 p2;
 };
 
-struct Position {
-  uint index;
-  uint count;
+struct QuadDecomposition {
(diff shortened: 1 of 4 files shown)


jonahwilliams (approved):
RSLGTM

dnfield:
A fun note to add: I was scratching my head for a bit debugging this problem because if I simplified the quad shader even a little, it jumped back up to a subgroup size of 32 on the intel mac and everything looked good. I finally thought to check the size specifically when I noticed that the `count` value was consistently smaller than expected once the shader got more complex.

PR #40516, opened 2023-03-22
[Impeller] Give ComputeSubgroupTest unique name to avoid running tests twice

dnfield:
Currently all the `ComputeTest` tests are run twice because the macro gets called twice with the same name. I missed this when factoring the subgroup related tests into their own TU a ways back.

This also makes it easier to filter so you run only the subgroup tests or you run all tests except subgroup tests.


jonahwilliams (approved):
LGTM

PR #40535, opened 2023-03-22
Read the FLTEnableImpeller flag from the right bundle

zanderso:
This PR causes the `FLTEnableImpeller` flag to be read first out of the application bundle, and next out of the main bundle.

This is not exactly what is described in https://github.com/flutter/engine/blob/e159e35fb29ca1ca29581d4f1c5124f4abd3b29d/shell/platform/darwin/ios/framework/Source/FlutterDartProject.mm#L82 since the command line is still ignored, but I can try that in a follow-up PR.

Fixes https://github.com/flutter/flutter/issues/123046

### DIFF
--- a/shell/platform/darwin/ios/framework/Source/FlutterDartProject.mm
+++ b/shell/platform/darwin/ios/framework/Source/FlutterDartProject.mm
@@ -206,8 +206,12 @@ flutter::Settings FLTDefaultSettingsForBundle(NSBundle* bundle) {
   BOOL enableWideGamut = nsEnableWideGamut ? nsEnableWideGamut.boolValue : NO;
   settings.enable_wide_gamut = enableWideGamut;
 
-  // Whether to enable Impeller.
-  NSNumber* enableImpeller = [mainBundle objectForInfoDictionaryKey:@"FLTEnableImpeller"];
+  // Whether to enable Impeller. First, look in the app bundle.
+  NSNumber* enableImpeller = [bundle objectForInfoDictionaryKey:@"FLTEnableImpeller"];
+  if (enableImpeller == nil) {
+    // If it isn't in the app bundle, look in the main bundle.
+    enableImpeller = [mainBundle objectForInfoDictionaryKey:@"FLTEnableImpeller"];
+  }
   // Change the default only if the option is present.
   if (enableImpeller != nil) {
     settings.enable_impeller = enableImpeller.boolValue;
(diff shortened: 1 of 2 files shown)


shell/platform/darwin/ios/framework/Source/FlutterDartProject.mm
```diff
@@ -207,7 +207,7 @@
   settings.enable_wide_gamut = enableWideGamut;
 
   // Whether to enable Impeller.
-  NSNumber* enableImpeller = [mainBundle objectForInfoDictionaryKey:@"FLTEnableImpeller"];
+  NSNumber* enableImpeller = [bundle objectForInfoDictionaryKey:@"FLTEnableImpeller"];
```
jmagman, line 210: This needs to check both, right?  Otherwise we will break customers who followed our instructions to add the flag to their app's Info.plist.

jmagman: Something like this:
```suggestion
  NSNumber* enableImpellerMainBundle = [mainBundle objectForInfoDictionaryKey:@"FLTEnableImpeller"];
  NSNumber* enableImpeller =
      enableImpellerMainBundle != nil
          ? enableImpellerMainBundle
          : [bundle objectForInfoDictionaryKey:@"FLTEnableImpeller"];
```

zanderso: Thanks! Based on the comment above, I look in the app bundle first.

shell/platform/darwin/ios/framework/Source/FlutterDartProjectTest.mm
```diff
@@ -73,15 +73,54 @@ - (void)testFLTFrameworkBundleInternalWhenBundleIsPresent {
   XCTAssertNotNil(found);
 }
 
+- (void)testDisableImpellerSettingIsCorrectlyParsed {
+  id mockMainBundle = OCMPartialMock([NSBundle mainBundle]);
+  OCMStub([mockMainBundle objectForInfoDictionaryKey:@"FLTEnableImpeller"]).andReturn(@"NO");
+
+  auto settings = FLTDefaultSettingsForBundle();
+  // Check settings.enable_impeller value is same as the value defined in Info.plist.
+  XCTAssertEqual(settings.enable_impeller, NO);
```
jmagman, line 82: `XCTAssertTrue` and `XCTAssertFalse` are a little more idiomatic, but it's a minor nit.
```suggestion
  XCTAssertFalse(settings.enable_impeller);
```

jmagman (approved):
LGTM, thank you!

shell/platform/darwin/ios/framework/Source/FlutterDartProjectTest.mm
zanderso: I suspect I'll be touching this file again soon, and I'll keep this in mind for a cleanup then.

PR #40540, opened 2023-03-22
[Impeller] Share stencil coverage stack between subpasses to support pass collapsing

bdero:
Also re-enables the opacity peephole optimization by reverting https://github.com/flutter/engine/pull/40497


jonahwilliams (approved):
LGTM!

PR #40542, opened 2023-03-22
Adding app lifecycle notification for macOS and Linux, add `hidden` state.

gspencergoog:
## Description

This adds app lifecycle notification for macOS and Linux, and adds a new `hidden` state to the `AppLifecycleState` enum.

To be functional, this needs a corresponding framework change: https://github.com/flutter/flutter/pull/123274

## Related Issues
 - https://github.com/flutter/flutter/issues/30735

## Tests
 - Added tests for state changes.


shell/platform/common/application_lifecycle.h
gspencergoog, line 1: Would it be better to go back to an enum here?  I had it as an enum, but that required a couple of functions to convert to/from strings, and I figured that just making them strings would be more efficient in both code and CPU.

lib/ui/platform_dispatcher.dart
```diff
@@ -1679,13 +1679,11 @@ class FrameTiming {
   }
 }
 
-/// States that an application can be in.
+/// States that an application can be in once it is running.
```
cbracken, line 1682: > once it is running.

I wonder if it's worth stating the initial state near the top here. Something like "Applications start in the detached state and transition to ..."

shell/platform/common/application_lifecycle.h
gspencergoog: Yeah, that's basically what I had before.  Okay, I'll switch back.

cbracken: One other benefit of doing that is there's an enums test somewhere we can use that's a somewhat shady approximation of the if-this-then-that tooling we have internally:
https://github.com/flutter/engine/blob/main/tools/api_check/test/apicheck_test.dart

shell/platform/darwin/macos/framework/Source/FlutterAppLifecycleDelegate.mm
```diff
@@ -0,0 +1,130 @@
...
+- (void)removeDelegate:(NSObject<FlutterAppLifecycleDelegate>*)delegate {
+  NSUInteger index = [[_delegates allObjects] indexOfObject:delegate];
+  if (index >= 0) {
+    [_delegates removePointerAtIndex:index];
+  }
+}
+
+// This isn't done via performSelector because that can cause leaks due to the
+// selector not being known. Using a macro to avoid mismatch errors between the
+// notification and the selector.
```
cbracken, line 97: I'm assuming the leaks in question would be if the method for the selector allocates and returns an object? It looks like there's a bit of discussion of this in the performSelector docs.

https://developer.apple.com/documentation/objectivec/1418956-nsobject/1418867-performselector

cbracken (approved):
lgtm - with a few quick comments.

shell/platform/darwin/macos/framework/Source/FlutterAppLifecycleDelegate.mm
gspencergoog: Yes, and the compiler was generating warnings when I tried it, so I just switched to this.

shell/platform/common/app_lifecycle_state.h
```diff
@@ -0,0 +1,93 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef FLUTTER_SHELL_PLATFORM_COMMON_APP_LIFECYCLE_STATE_H_
+#define FLUTTER_SHELL_PLATFORM_COMMON_APP_LIFECYCLE_STATE_H_
+
+#include <cassert>
+#include <map>
+#include <string>
```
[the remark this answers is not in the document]
gspencergoog: Well, they were, but no longer.  Removed.

shell/platform/darwin/macos/framework/Source/FlutterAppLifecycleDelegate.mm
```diff
@@ -0,0 +1,130 @@
...
+
+static BOOL IsPowerOfTwo(NSUInteger x) {
+  return x != 0 && (x & (x - 1)) == 0;
+}
+
+- (void)addDelegate:(NSObject<FlutterAppLifecycleDelegate>*)delegate {
+  [_delegates addPointer:(__bridge void*)delegate];
+  if (IsPowerOfTwo([_delegates count])) {
+    [_delegates compact];
+  }
```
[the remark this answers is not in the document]
gspencergoog: This is what the [iOS engine already uses](https://github.com/flutter/engine/blob/eed12f36f5956a37fb4ba13f1b670a5d41c37edb/shell/platform/darwin/ios/framework/Source/FlutterPluginAppLifeCycleDelegate.mm#L101) for its lifecycle delegate registration, so I was mainly just keeping them similar.

I agree that it's probably overkill, but also probably harmless.  The naive implementation would just not call compact at all, instead of on powers of two, so it's not that different/complex, and potentially saves a little memory.  I'm happy to just nuke it, though, if you think it's adding too much complexity.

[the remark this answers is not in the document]
cbracken: Thanks for the additional context - agreed it's best to keep them consistent!

PR #40551, opened 2023-03-23
[impeller] dont apply opacity peephole on image/color filters

jonahwilliams:
Fixes the rendering issues observed on the wide gamut test app (wide gamut was fine, blur was lost).


bdero (approved):
LGTM

PR #40567, opened 2023-03-23
`SemanticsFlag`/`SemanticsAction` cleanup (part 3)

bernaferrari:
These two classes should be an enum, because they work as enum, but they are older than the enhanced enum, so this wasn't possible before. Now it is.

This is kind of breaking because Flutter uses `semanticsFlag.values.values`, but I searched on code search and couldn't find a repo that uses that, so it seems really internal. 
We could do in these ways:
- Add `deserialize`/`indexAt`/`[]` on them, migrate Flutter, convert to enum;
- Convert to enum and land the Flutter change at the same time (is this possible?);
- Don't do anything, lol

### DIFF
--- a/lib/ui/platform_dispatcher.dart
+++ b/lib/ui/platform_dispatcher.dart
@@ -1155,7 +1155,7 @@ class PlatformDispatcher {
       onSemanticsAction,
       _onSemanticsActionZone,
       nodeId,
-      SemanticsAction.values[action]!,
+      SemanticsAction.fromAction(action)!,
       args,
     );
   }
--- a/lib/ui/platform_dispatcher.dart
+++ b/lib/ui/platform_dispatcher.dart
@@ -1155,7 +1155,7 @@ class PlatformDispatcher {
       onSemanticsAction,
       _onSemanticsActionZone,
       nodeId,
-      SemanticsAction.fromAction(action)!,
+      SemanticsAction.fromIndex(action)!,
       args,
     );
   }
--- a/lib/ui/semantics.dart
+++ b/lib/ui/semantics.dart
@@ -233,7 +233,7 @@ class SemanticsAction {
 
   static List<SemanticsAction> get values => _kActionById.values.toList(growable: false);
 
-  static SemanticsAction? fromAction(int index) => _kActionById[index];
+  static SemanticsAction? fromIndex(int index) => _kActionById[index];
 
   /// Temporary API until [values] return a list.
   /// https://github.com/flutter/flutter/issues/123346
@@ -571,7 +571,7 @@ class SemanticsFlag {
 
   static List<SemanticsFlag> get values => _kFlagById.values.toList(growable: false);
 
-  static SemanticsFlag? fromFlag(int index) => _kFlagById[index];
+  static SemanticsFlag? fromIndex(int index) => _kFlagById[index];
 
   /// Temporary API until [values] return a list.
   /// https://github.com/flutter/flutter/issues/123346
--- a/lib/ui/semantics.dart
+++ b/lib/ui/semantics.dart
@@ -233,6 +233,8 @@ class SemanticsAction {
 
   static List<SemanticsAction> get values => _kActionById.values.toList(growable: false);
 
+  static SemanticsAction? fromAction(int index) => _kActionById[index];
+
   /// Temporary API until [values] return a list.
   /// https://github.com/flutter/flutter/issues/123346
   @Deprecated('This getter is temporary and will be removed shortly.')
@@ -568,6 +570,8 @@ class SemanticsFlag {
   };
 
   static List<SemanticsFlag> get values => _kFlagById.values.toList(growable: false);
+
+  static SemanticsFlag? fromFlag(int index) => _kFlagById[index];
 
   /// Temporary API until [values] return a list.
   /// https://github.com/flutter/flutter/issues/123346
--- a/lib/ui/semantics.dart
+++ b/lib/ui/semantics.dart
@@ -13,7 +13,15 @@ part of dart.ui;
 /// See also:
 ///   - file://./../../lib/ui/semantics/semantics_node.h
 class SemanticsAction {
-  const SemanticsAction._(this.index);
+  const SemanticsAction._(this.index, this.name);
+
+  /// The numerical value for this action.
+  ///
+  /// Each action has one bit set in this bit field.
+  final int index;
+
+  /// A human-readable name for this flag, used for debugging purposes.
+  final String name;
 
   static const int _kTapIndex = 1 << 0;
   static const int _kLongPressIndex = 1 << 1;
@@ -41,62 +49,57 @@ class SemanticsAction {
   // numSemanticsActions value in testing/dart/semantics_test.dart, or tests
   // will fail.
 
-  /// The numerical value for this action.
-  ///
-  /// Each action has one bit set in this bit field.
-  final int index;
-
   /// The equivalent of a user briefly tapping the screen with the finger
   /// without moving it.
-  static const SemanticsAction tap = SemanticsAction._(_kTapIndex);
+  static const SemanticsAction tap = SemanticsAction._(_kTapIndex, 'tap');
 
   /// The equivalent of a user pressing and holding the screen with the finger
   /// for a few seconds without moving it.
-  static const SemanticsAction longPress = SemanticsAction._(_kLongPressIndex);
+  static const SemanticsAction longPress = SemanticsAction._(_kLongPressIndex, 'longPress');
 
   /// The equivalent of a user moving their finger across the screen from right
   /// to left.
   ///
   /// This action should be recognized by controls that are horizontally
   /// scrollable.
-  static const SemanticsAction scrollLeft = SemanticsAction._(_kScrollLeftIndex);
+  static const SemanticsAction scrollLeft = SemanticsAction._(_kScrollLeftIndex, 'scrollLeft');
 
   /// The equivalent of a user moving their finger across the screen from left
   /// to right.
   ///
   /// This action should be recognized by controls that are horizontally
   /// scrollable.
-  static const SemanticsAction scrollRight = SemanticsAction._(_kScrollRightIndex);
+  static const SemanticsAction scrollRight = SemanticsAction._(_kScrollRightIndex, 'scrollRight');
 
   /// The equivalent of a user moving their finger across the screen from
   /// bottom to top.
   ///
   /// This action should be recognized by controls that are vertically
   /// scrollable.
-  static const SemanticsAction scrollUp = SemanticsAction._(_kScrollUpIndex);
+  static const SemanticsAction scrollUp = SemanticsAction._(_kScrollUpIndex, 'scrollUp');
 
   /// The equivalent of a user moving their finger across the screen from top
   /// to bottom.
   ///
   /// This action should be recognized by controls that are vertically
   /// scrollable.
-  static const SemanticsAction scrollDown = SemanticsAction._(_kScrollDownIndex);
+  static const SemanticsAction scrollDown = SemanticsAction._(_kScrollDownIndex, 'scrollDown');
 
   /// A request to increase the value represented by the semantics node.
   ///
   /// For example, this action might be recognized by a slider control.
-  static const SemanticsAction increase = SemanticsAction._(_kIncreaseIndex);
+  static const SemanticsAction increase = SemanticsAction._(_kIncreaseIndex, 'increase');
 
   /// A request to decrease the value represented by the semantics node.
   ///
   /// For example, this action might be recognized by a slider control.
-  static const SemanticsAction decrease = SemanticsAction._(_kDecreaseIndex);
+  static const SemanticsAction decrease = SemanticsAction._(_kDecreaseIndex, 'decrease');
 
   /// A request to fully show the semantics node on screen.
   ///
   /// For example, this action might be send to a node in a scrollable list that
   /// is partially off screen to bring it on screen.
-  static const SemanticsAction showOnScreen = SemanticsAction._(_kShowOnScreenIndex);
+  static const SemanticsAction showOnScreen = SemanticsAction._(_kShowOnScreenIndex, 'showOnScreen');
 
   /// Move the cursor forward by one character.
   ///
@@ -104,7 +107,7 @@ class SemanticsAction {
   ///
   /// The action includes a boolean argument, which indicates whether the cursor
   /// movement should extend (or start) a selection.
-  static const SemanticsAction moveCursorForwardByCharacter = SemanticsAction._(_kMoveCursorForwardByCharacterIndex);
+  static const SemanticsAction moveCursorForwardByCharacter = SemanticsAction._(_kMoveCursorForwardByCharacterIndex, 'moveCursorForwardByCharacter');
 
   /// Move the cursor backward by one character.
   ///
@@ -112,7 +115,7 @@ class SemanticsAction {
   ///
   /// The action includes a boolean argument, which indicates whether the cursor
   /// movement should extend (or start) a selection.
-  static const SemanticsAction moveCursorBackwardByCharacter = SemanticsAction._(_kMoveCursorBackwardByCharacterIndex);
+  static const SemanticsAction moveCursorBackwardByCharacter = SemanticsAction._(_kMoveCursorBackwardByCharacterIndex, 'moveCursorBackwardByCharacter');
 
   /// Replaces the current text in the text field.
   ///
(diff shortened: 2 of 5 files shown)


kevmoo:
@yjbanov ?

bernaferrari:
Bot tagged as web because there are two SemanticsFlag, one on the web and one everywhere else. I don't know who is responsible, lol. This could fall under accessibility. But Greg likes to review my PRs, if he is on engine and Yegor doesn't fit, you can also tag him.

chunhtai:
It looks like ci is failing, can you fix it?

bernaferrari:
Yes, but I need your help. Getting gsync to work is too much for me. Can you help in Discord? I don't know how to get api_check to work.

chunhtai:
based on some of the error message, you may need to do a migration
```dart
Try correcting the name to the name of an existing getter, or defining a getter or field named 'values'.
      for (final SemanticsAction action in SemanticsAction.values.values)
                                                                  ^^^^^^
lib/src/semantics/semantics.dart:683:61: Error: The getter 'values' isn't defined for the class 'List<SemanticsFlag>'.
 - 'List' is from 'dart:core'.
 - 'SemanticsFlag' is from 'dart:ui'.
Try correcting the name to the name of an existing getter, or defining a getter or field named 'values'.
      for (final SemanticsFlag flag in SemanticsFlag.values.values)
                                                            ^^^^^^
lib/src/semantics/semantics.dart:2758:53: Error: The getter 'values' isn't defined for the class 'List<SemanticsFlag>'.
 - 'List' is from 'dart:core'.
 - 'SemanticsFlag' is from 'dart:ui'.
Try correcting the name to the name of an existing getter, or defining a getter or field named 'values'.
    final List<String> flags = SemanticsFlag.values.values.where((SemanticsFlag flag) => hasFlag(flag)).map((SemanticsFlag flag) => flag.toString().substring('SemanticsFlag.'.length)).toList();
                                                    ^^^^^^
```

bernaferrari:
Oh, this one is easy. I was seeing an entirely different error. Hold on, 2 minutes.

bernaferrari:
Do I make the migration in this PR? Or do I make one in Flutter and tie them togehter? I thought the error message was something else. This one is easy. I have it here: https://github.com/flutter/flutter/pull/123329

bernaferrari:
Blocked by https://github.com/flutter/engine/pull/40571 :)

bernaferrari:
I just rebased on top of master.

Do you think `SemanticsAction.deserialize(flag)` is a good replacement for `SemanticsAction.values[flag]`? AndroidSemanticsAction used "deserialize", which was my inspiration.

goderbauer:
Looks like some checks are failing

lib/ui/platform_dispatcher.dart
```diff
@@ -1155,7 +1155,7 @@ class PlatformDispatcher {
       onSemanticsAction,
       _onSemanticsActionZone,
       nodeId,
-      SemanticsAction.values[action]!,
+      SemanticsAction.deserialize(action)!,
```
goderbauer, line 1158: Since the getter is called `value` maybe we should just call this `SemanticsAction.fromValue`.

goderbauer: Alternatively, we could keep the old name `index` for the property name and call this `SemanticsAction.fromIndex`.

lib/ui/semantics.dart
```diff
@@ -181,29 +148,62 @@ class SemanticsAction {
...
   /// movement should extend (or start) a selection.
-  static const SemanticsAction moveCursorBackwardByWord = SemanticsAction._(_kMoveCursorBackwardByWordIndex);
+  moveCursorBackwardByWord(_kMoveCursorBackwardByWordIndex);
+
+  const SemanticsAction(this.value);
+
+  /// The numerical value for this action.
+  ///
+  /// Each action has one bit set in this bit field.
+  final int value;
```
goderbauer, line 174: Is there a good reason to change the name of this property from `index` to `value`?

lib/ui/semantics.dart
```diff
@@ -574,12 +486,54 @@ class SemanticsFlag {
...
   /// navigate to the next page when reaching the end of the current one.
-  static const SemanticsFlag hasImplicitScrolling = SemanticsFlag._(_kHasImplicitScrollingIndex);
+  hasImplicitScrolling(_kHasImplicitScrollingIndex);
+
+  const SemanticsFlag(this.value);
+
+  /// The numerical value for this flag.
+  ///
+  /// Each flag has one bit set in this bit field.
+  final int value;
```
goderbauer, line 496: same question here: value vs. index

lib/ui/semantics.dart
bernaferrari: enum has its own index, so you can't have two index with the same name

goderbauer: Oh, right! Makes sense.

lib/ui/semantics.dart
```diff
@@ -298,49 +251,8 @@ class SemanticsAction {
...
-  // value in testing/dart/semantics_test.dart, or tests will fail. Also,
-  // please update the Flag enum in
-  // flutter/shell/platform/android/io/flutter/view/AccessibilityBridge.java,
-  // and the SemanticsFlag class in lib/web_ui/lib/semantics.dart. If the new flag
-  // affects the visibility of a [SemanticsNode] to accessibility services,
-  // `flutter_test/controller.dart#SemanticsController._importantFlags`
-  // must be updated as well.
-
+enum SemanticsFlag {
+
```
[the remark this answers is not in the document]
bernaferrari: Thanks!

goderbauer (approved):
LGTM

lib/ui/semantics.dart
```diff
@@ -228,67 +228,20 @@ class SemanticsAction {
     _kSetTextIndex: setText,
   };
 
+  /// Creates a new [SemanticsAction] from an integer `value`.
+  ///
+  /// Returns `null` if the id is not a known action.
+  static SemanticsAction? fromValue(int value) => _kActionById[value];
```
chunhtai, line 234: I am not super familiar with this but could this just be `SemanticsAction(value)`?

bernaferrari: No, from [] maybe. Could be potentially confusing, but I need to check.

bernaferrari:
One test fixed. The next step is this. Does anyone know what might be happening? Is there a way `Dart_GetNativeIntegerArgument` is getting the wrong value because it is used to class instead of enum? I don't know who to even ask for help, who writes c++ on the engine?

image

tools/api_check/lib/apicheck.dart
```diff
@@ -52,6 +52,47 @@ List<String> getDartClassFields({
...
+  final SomeParsedUnitResult result = session.getParsedUnit(sourcePath);
+  if (result is! ParsedUnitResult) {
+    return <String>[];
+  }
+
+  // Locate all fields matching the expression in the class.
+  final RegExp fieldExp = RegExp(r'_k(\w*)Index');
+  final List<String> fields = <String>[];
+  for (final CompilationUnitMember unitMember in result.unit.declarations) {
+    if (unitMember is EnumDeclaration && unitMember.name.lexeme == className) {
```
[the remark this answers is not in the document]
bernaferrari: Yes! I'll fix this.

tools/api_check/test/apicheck_test.dart
```diff
@@ -105,14 +105,17 @@ void checkApiConsistency(String flutterRoot) {
 
   test('SemanticsFlag enums match', () {
     // Dart values: _kFooBarIndex = 1 << N.
-    final List<String> uiFields = getDartClassFields(
+    final List<String> uiFields = getDartEnumFields(
       sourcePath: path.join(flutterRoot, 'lib', 'ui', 'semantics.dart'),
       className: 'SemanticsFlag',
     );
-    final List<String> webuiFields = getDartClassFields(
+    print('uiFields are $uiFields');
```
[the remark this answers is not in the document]
bernaferrari: Ops. Done!

tools/api_check/lib/apicheck.dart
```diff
@@ -35,7 +35,7 @@ List<String> getDartClassFields({
   final RegExp fieldExp = RegExp(r'_k(\w*)Index');
   final List<String> fields = <String>[];
   for (final CompilationUnitMember unitMember in result.unit.declarations) {
-    if (unitMember is ClassDeclaration && unitMember.name.lexeme == className) {
+    if ((unitMember is ClassDeclaration || unitMember is EnumDeclaration) && unitMember.name.lexeme == className) {
```
goderbauer, line 38: Meh, looks like `unitMember` is not properly type promoted anymore, so `unitMember.name` doesn't resolve. I guess we'd have to do something like

```suggestion
    if ((unitMember is ClassDeclaration || unitMember is EnumDeclaration) && (unitMember as NamedCompilationUnitMember).name.lexeme == className) {
```

bernaferrari: uhhhhh, thanks!! I was still trying to see what the error is on the logs, and trying to get the engine to compile again here!

bernaferrari: Now it is something else. I'm going to cry. Testing here.

tools/api_check/lib/apicheck.dart
```diff
@@ -35,7 +35,7 @@ List<String> getDartClassFields({
   final RegExp fieldExp = RegExp(r'_k(\w*)Index');
   final List<String> fields = <String>[];
   for (final CompilationUnitMember unitMember in result.unit.declarations) {
-    if (unitMember is ClassDeclaration && unitMember.name.lexeme == className) {
+    if ((unitMember is ClassDeclaration || unitMember is EnumDeclaration) && (unitMember as NamedCompilationUnitMember).name.lexeme == className) {
       for (final ClassMember classMember in unitMember.members) {
```
goderbauer, line 39: Looks like deduplicating this is not as easy as I thought as there is no common base class of ClassDeclaration and EnumDeclaration that defines the `.members` here. We may have to come up with a different approach to de-duplicating this.

bernaferrari: If it doesn't work now... lol

tools/api_check/lib/apicheck.dart
```diff
@@ -52,6 +53,16 @@ List<String> getDartClassFields({
   return fields;
 }
 
+NodeList<ClassMember>? _getMembers(String className, CompilationUnitMember unitMember) {
```
goderbauer, line 56: suggestion for simplification: type this as non-nullable and make it return an empty list instead of null. Then you can also remove the null check on line 39.

bernaferrari: TIL NodeList can be a []. Hold on!

tools/api_check/lib/apicheck.dart
```diff
@@ -35,8 +35,9 @@ List<String> getDartClassFields({
   final RegExp fieldExp = RegExp(r'_k(\w*)Index');
   final List<String> fields = <String>[];
   for (final CompilationUnitMember unitMember in result.unit.declarations) {
-    if ((unitMember is ClassDeclaration || unitMember is EnumDeclaration) && unitMember.name.lexeme == className) {
-      for (final ClassMember classMember in unitMember.members) {
+    final NodeList<ClassMember> members = _getMembers(className, unitMember);
```
goderbauer, line 38: if you don't like my other suggestion, you'll have to type this as nullable:

```suggestion
   final NodeList<ClassMember>? members = _getMembers(className, unitMember);
```

bernaferrari: oh, it was nullable but I added the if as a precaution. Uhhhhhh. Thanks for your patience.

bernaferrari: I'm not that bad, but half my changes I see this and think I got it right: 
image

tools/api_check/lib/apicheck.dart
goderbauer: Oh, instead of NodeList you could also just use List since `NodeList` implements `List`. You may actually have to do that to do the empty list thing.

tools/api_check/lib/apicheck.dart
```diff
@@ -53,14 +51,14 @@ List<String> getDartClassFields({
   return fields;
 }
 
-NodeList<ClassMember>? _getMembers(String className, CompilationUnitMember unitMember) {
+NodeList<ClassMember> _getMembers(String className, CompilationUnitMember unitMember) {
```
goderbauer, line 54: This likely will have to be typed `List<ClassMember>`. Same for the variable in line 38. I think after that, we got this.

tools/api_check/lib/apicheck.dart
bernaferrari: It turns out it couldn't... lol

tools/api_check/lib/apicheck.dart
bernaferrari: This is the hardest live programming I've done. Got it.

goderbauer: :) Thanks for sticking with it. I think we got it now. Let's see what the checks say!

bernaferrari: It passed. Finally!

goderbauer:
Hm, somewhere this is still failing. I wonder if there is another place where we accidentally use `index` instead of value....

goderbauer:
Yeah, I think that is it: Things like this here

https://github.com/flutter/flutter/blob/73bd97852978a4778f51dae0c48bc0c08d674edb/packages/flutter/test/cupertino/button_test.dart#L291

are now doing the wrong thing because they use `Enum.index` instead of what we now call `value`. There may not be a good way to migrate this since I don't think we can override `Enum.index`.

bernaferrari:
Ahhhhhhhh... We kind of opened a pandora box, each day something new appears lol
I think I found a test that might fail (just commited), but that's a lot worse.

I guess... hmm.. add `value` to engine, replace index with `value` in Flutter and then make the migration?

goderbauer:
What about user code that currently uses `.index`? They will be broken in the worst possible way: everything still appears to be valid and compiles, but then it may fail in hard-to-debug ways at runtime. I am actually more leaning towards not doing this clean-up now.

bernaferrari:
I agree. But...are there users using SemanticsAction? I searched code search and found a [single usage](https://github.com/search?type=code&auto_enroll=true&q=SemanticsAction), which the person copied Flutter bindings for Go. I doesn't seem to have that big of a user-base. But we could deprecate index, move to value, then migrate, or just don't do it.

goderbauer:
You could:

a) add `value` to the SemanticsAction class (not the enum)
b) migrate the framework 
c) remove `index` from the SemanticsAction class and see if any of our customer tests (or google testing) break. If nothing breaks, we can continue with this migration and change the class to an enum. If things break, we may have to undo all that.

It's upto you if you're up for doing all that extra work :)

bernaferrari:
Well, we went until here, we can get there (as long as your patience allows)... `value` is 1 LOC. Let's try and see.

bernaferrari:
After a lot of discussion and nice ideas, this will be the end result. It is not an enum, and might probably never be, but it is almost there.

Do you prefer `from`, `fromAction`, `fromIndex` or `deserialize` for the `=> values[index]` getter? Somewhere else in Flutter uses deserialize, but I don't like. `from` or `fromAction` sounds better because `SemanticsAction.fromAction(action)` is more readable than `SemanticsAction.fromIndex(action)`. But anything is fine.

Thanks for all the help!

lib/ui/semantics.dart
```diff
@@ -608,75 +569,22 @@ class SemanticsFlag {
     _kIsCheckStateMixedIndex: isCheckStateMixed,
   };
 
+  static List<SemanticsFlag> get values => _kFlagById.values.toList(growable: false);
+
+  static SemanticsFlag? fromFlag(int index) => _kFlagById[index];
```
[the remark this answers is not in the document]
bernaferrari: Thanks. Done!

lib/ui/semantics.dart
```diff
@@ -228,67 +231,22 @@ class SemanticsAction {
     _kSetTextIndex: setText,
   };
 
+  static List<SemanticsAction> get values => _kActionById.values.toList(growable: false);
+
+  static SemanticsAction? fromAction(int index) => _kActionById[index];
```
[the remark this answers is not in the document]
bernaferrari: Thanks. Done!

goderbauer (approved):
LGTM

chunhtai (approved):
LGTM

bernaferrari:
Thanks, everybody for this long journey. Feel free to add the "autosubmit".

PR #40578, opened 2023-03-23
[embedder] embedder external view adopt DisplayListEmbedderViewSlice

cyanglaz:
Adopt `DisplayListEmbedderViewSlice` in EmbedderExternalView

Fixes https://github.com/flutter/flutter/issues/123349

## Pre-launch Checklist

- [x] I signed the [CLA].


shell/platform/embedder/embedder_external_view_dl_op_spy.h
```diff
@@ -0,0 +1,136 @@
...
+#include "flutter/display_list/dl_op_receiver.h"
+
+namespace flutter {
+
+//------------------------------------------------------------------------------
+/// Receives to drawing commands of a DisplayListBuilder.
+///
+/// This is used to determine whether anything will be drawn
+/// on the canvas.
+class EmbedderExternalViewDlOpReceiver final : public virtual DlOpReceiver {
```
flar, line 17: Have this inherit from the various "Ignore" helpers in dl_receiver_utils then you don't have to create all of the empty overrides.

You can even do that for the attributes and then override the ones you want to handle and leave the rest to be ignored by the helper.

shell/platform/embedder/embedder_external_view_dl_op_spy.cc
```diff
@@ -0,0 +1,159 @@
...
+                                                 const DlColor colors[],
+                                                 int count,
+                                                 DlBlendMode mode,
+                                                 DlImageSampling sampling,
+                                                 const SkRect* cull_rect,
+                                                 bool render_with_attributes) {}
+void EmbedderExternalViewDlOpReceiver::drawDisplayList(
+    const sk_sp<DisplayList> display_list,
+    SkScalar opacity) {
+  FML_DLOG(ERROR) << "drawDisplayList ";
```
flar, line 147: For this method, I would do something like:

```
  Something like [auto remembered_transparency_info = current_transparency_info_;]
  Something like [current_transparency_info_ = defaults;]
  display_list->Dispatch(this);
  Something like [current_transparency_info_ = remembered_transparency_info]
```

Basically, play back the DL with the following caveats:
- Your local attributes should be unaffected by the actions of the DL
- The DL should be played back assuming default attributes
- Your local attributes when everything is done should be the same state as before you recursed.

That is why I have the "save/restore" lines for whatever "current_transparency_info" might be.

flar: Another option would be:

```
  EmbedderExternalViewDlOpReceiver sub_receiver;
  display_list->Dispatch(sub_receiver);
  this->did_draw_ = this->did_draw_ || sub_receiver.did_draw_;
```

flar: And, of course, if "did_draw" is already true, there is no need to recurse in this call as it can't add any new info for your eventual reported answer...

shell/platform/embedder/embedder_external_view.cc
```diff
@@ -59,7 +62,10 @@ bool EmbedderExternalView::HasPlatformView() const {
 }
 
 bool EmbedderExternalView::HasEngineRenderedContents() const {
-  return canvas_spy_->DidDrawIntoCanvas();
+  slice_->end_recording();
```
flar, line 65: Just be sure that this is not called before the rendering is actually done...

shell/platform/embedder/embedder_external_view.cc
```diff
@@ -95,11 +101,15 @@ bool EmbedderExternalView::Render(const EmbedderRenderTarget& render_target) {
...
   }
-
-  canvas->setMatrix(surface_transformation_);
-  canvas->clear(SK_ColorTRANSPARENT);
-  canvas->drawPicture(picture);
-  canvas->flush();
+  auto dl_canvas = std::make_unique<DlSkCanvasAdapter>(canvas);
+  int restore_count = dl_canvas->GetSaveCount();
+  dl_canvas->SetTransform(surface_transformation_);
+  dl_canvas->Clear(SK_ColorTRANSPARENT);
```
flar, line 107: Technically, while SK_Color constants are automatically interchangeable with DlColor instances (they both just boil down to a uint32_t), It would be better to use `DlColor::kTransparent()` here to continue the theme of non-skia use while talking to Dl objects.

shell/platform/embedder/embedder_external_view.cc
```diff
@@ -95,11 +101,15 @@ bool EmbedderExternalView::Render(const EmbedderRenderTarget& render_target) {
...
+  slice_->render_into(dl_canvas.get());
+  dl_canvas->RestoreToCount(restore_count);
+
+  // canvas->drawPicture(picture);
+  // canvas->flush();
```
flar, line 112: `dl_canvas.Flush()`

cyanglaz:
@flar This is ready to review!

display_list/BUILD.gn
```diff
@@ -123,6 +126,7 @@ if (enable_unittests) {
       ":display_list_fixtures",
       "//flutter/display_list/testing:display_list_testing",
       "//flutter/testing",
+      "//third_party/skia",
```
flar, line 129: Why is this needed? I'm guessing if the spy stuff gets moved out of here into the embedder source directories then it doesn't need to be added here, but I would have thought that you could implement an OpReceiver without having to add this dependency...?

display_list/dl_op_spy.cc
```diff
@@ -0,0 +1,171 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "flutter/display_list/dl_op_spy.h"
+
+namespace flutter {
+
+bool DlOpSpy::DidDraw() {
```
flar, line 9: I've been using lower case for getters which I think is compliant with the style guidelines - `did_draw()`...

display_list/dl_op_spy.cc
```diff
@@ -0,0 +1,171 @@
...
+
+void DlOpSpy::setColor(DlColor color) {
+  if (color.isTransparent()) {
+    dirty_ = false;
+  } else {
+    dirty_ = true;
+  }
+}
+void DlOpSpy::setColorSource(const DlColorSource* source) {
+  const DlColorColorSource* color_source = source->asColor();
```
flar, line 21: What if source is null?

display_list/dl_op_spy.cc
```diff
@@ -0,0 +1,171 @@
...
+  if (did_draw_) {
+    return;
+  }
+  did_draw_ = dirty_;
+}
+void DlOpSpy::drawLine(const SkPoint& p0, const SkPoint& p1) {
+  if (did_draw_) {
+    return;
+  }
+  did_draw_ = dirty_ && p0 != p1;
```
flar, line 44: p0 == p1 draws something

display_list/dl_op_spy.cc
```diff
@@ -0,0 +1,171 @@
...
+}
+void DlOpSpy::drawRect(const SkRect& rect) {
+  if (did_draw_) {
+    return;
+  }
+  did_draw_ = dirty_ && !rect.isEmpty();
```
flar, line 50: I'd have to check, but I think an empty rect will draw something in stroke mode just like a zero length line. Maybe there is a check somewhere that avoids drawing an empty rect even in stroke mode, but I'd have to search to make sure.

display_list/dl_op_spy.cc
```diff
@@ -0,0 +1,171 @@
...
+  if (did_draw_) {
+    return;
+  }
+  did_draw_ = dirty_ && !outer.isEmpty() && outer != inner;
+}
+void DlOpSpy::drawPath(const SkPath& path) {
+  if (did_draw_) {
+    return;
+  }
+  did_draw_ = dirty_ && !path.isEmpty();
```
flar, line 80: Double check if path.isEmpty really means there is no geometry. If it means the bounds are empty, then we might end up stroking something in stroke mode.

display_list/dl_op_spy.cc
```diff
@@ -0,0 +1,171 @@
...
+}
+void DlOpSpy::drawShadow(const SkPath& path,
+                         const DlColor color,
+                         const SkScalar elevation,
+                         bool transparent_occluder,
+                         SkScalar dpr) {
+  if (did_draw_) {
+    return;
+  }
+  did_draw_ = !color.isTransparent() && elevation > 0 && !path.isEmpty();
```
flar, line 168: I'm not sure about elevation == 0. I think it might also depend on transparent_occluder as well.

shell/platform/embedder/embedder_external_view.cc
```diff
@@ -59,7 +58,9 @@ bool EmbedderExternalView::HasPlatformView() const {
 }
 
 bool EmbedderExternalView::HasEngineRenderedContents() const {
-  return canvas_spy_->DidDrawIntoCanvas();
+  TryEndRecording();
+  slice_->dispatch(*dl_op_spy_);
+  return dl_op_spy_->DidDraw() && !slice_->is_empty();
```
flar, line 63: If this method is called more than once it will repeat the op_spy operation. Also, it will re-use the op_spy object which means the second answer might not be correct. I would change this so that it only does the spy operation once and remembers the answer.

Also, you don't need to permanently store an op_spy object, just create one on the fly when you need to compute the answer and remove the dl_op_spy field from the object.

flar (changes requested):

display_list/BUILD.gn
cyanglaz: This is for the unittest, to generate a concrete DlImage for testing.

ci/licenses_golden/licenses_flutter
```diff
@@ -730,6 +730,8 @@ ORIGIN: ../../../flutter/display_list/dl_op_receiver.cc + ../../../flutter/LICEN
 ORIGIN: ../../../flutter/display_list/dl_op_receiver.h + ../../../flutter/LICENSE
 ORIGIN: ../../../flutter/display_list/dl_op_records.cc + ../../../flutter/LICENSE
 ORIGIN: ../../../flutter/display_list/dl_op_records.h + ../../../flutter/LICENSE
+ORIGIN: ../../../flutter/display_list/dl_op_spy.cc + ../../../flutter/LICENSE
```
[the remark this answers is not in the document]
cyanglaz: I initially put it in the platform/embedder/ but saw `canvas_spy` is in /common, so I thought maybe this class should be in a common place. 

Do you have a preference where it should be?

display_list/dl_op_spy.cc
cyanglaz: What does it draw? A single pixel? Do you know if there is an existing test or document for this we can check?

display_list/dl_op_spy.cc
cyanglaz: what does `transparent_occluder` do?

display_list/dl_op_spy.h
```diff
@@ -0,0 +1,92 @@
...
+
+//------------------------------------------------------------------------------
+/// Receives to drawing commands of a DisplayListBuilder.
+///
+/// This is used to determine whether anything will be drawn
+/// on the canvas.
+class DlOpSpy final : public IgnoreAttributeDispatchHelper,
+                      public IgnoreClipDispatchHelper,
+                      public IgnoreTransformDispatchHelper,
+                      public IgnoreDrawDispatchHelper {
```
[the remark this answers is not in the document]
cyanglaz: There are still calls in ignoreDraw: save, savelayer, restore.

cyanglaz:
@flar Updated per review comments and offline discussions. PTAL.

flar (approved):
Minor edits suggested, but the basic concept is good.

flar (approved):
Reiterating LGTM after review updates...

PR #40579, opened 2023-03-23
[impeller] runtime effect doesnt accept opacity

jonahwilliams:
Noticed while bisecting other issues.

### DIFF
--- a/impeller/entity/contents/runtime_effect_contents.cc
+++ b/impeller/entity/contents/runtime_effect_contents.cc
@@ -37,6 +37,10 @@ void RuntimeEffectContents::SetTextureInputs(
   texture_inputs_ = std::move(texture_inputs);
 }
 
+bool RuntimeEffectContents::CanAcceptOpacity(const Entity& entity) const {
+  return false;
+}
+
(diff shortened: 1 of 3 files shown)


dnfield (approved):

impeller/entity/contents/runtime_effect_contents.cc
```diff
@@ -37,6 +37,10 @@ void RuntimeEffectContents::SetTextureInputs(
   texture_inputs_ = std::move(texture_inputs);
 }
 
+bool RuntimeEffectContents::CanAcceptOpacity(const Entity& entity) const {
+  return false;
```
dnfield, line 41: Should this just default to returning false on the base class?

jonahwilliams: It does, but not on color source. I'm somewhat indifferent, we're probably not going to add any more content types.

jonahwilliams: Contents (false) -> ColorSource (true) -> RuntimeEffectContents(false).

Its sort of an odd hierarchy for sure

PR #40582, opened 2023-03-23
Return an empty JavaScript object as a replacement for exports/module in patchCanvasKitModule

jason-simmons:
Fixes https://github.com/flutter/flutter/issues/123219


yjbanov (approved):
lgtm

PR #40597, opened 2023-03-24
[web] Add `js_types` sources to build.

joshualitt:


joshualitt:
@eyebrowsoffire ptal, I'm actually not sure if this is the right fix, still trying to figure out a way to test this, but it seems promising.

eyebrowsoffire (approved):
LGTM

PR #40604, opened 2023-03-24
Ensure all golden filenames end in .png.

eyebrowsoffire:
This fixes https://github.com/flutter/flutter/issues/121906

Skia seems to silently ignore files without an extension, so we should double-check the extension before sending the screenshot.


yjbanov (approved):
lgtm

mdebbar (commented):
LGTM

PR #40608, opened 2023-03-24
`ui_web` library

eyebrowsoffire:
This introduces a public `ui_web` library that contains web-specific flutter APIs. The first thing to incorporate into this is `UrlStrategy`, which will allow us to replace a lot of the magic JS interop handshaking between framework and engine into a formalized API.


lib/web_ui/lib/src/engine/navigation/history.dart
```diff
@@ -4,18 +4,18 @@
 
 import 'package:meta/meta.dart';
 import 'package:ui/ui.dart' as ui;
+import '../../../ui_web/src/ui_web.dart' as ui_web;
```
mdebbar, line 7: How hard would it be to make this similar to the `package:ui/ui.dart` above? (i.e. `package:ui_web/ui_web.dart` or something like that). I guess it's just a matter of rewriting it in the sdk_rewriter?

lib/web_ui/lib/src/engine/navigation/history.dart
```diff
@@ -100,7 +100,7 @@ abstract class BrowserHistory {
   ///
   /// Subclasses should send appropriate system messages to update the flutter
   /// applications accordingly.
-  void onPopState(covariant DomPopStateEvent event);
+  void onPopState(Object? state);
```
mdebbar, line 103: I'm not sure I follow the reasoning for this change. Could you clarify?

Below, the override is still defined as `void Function(covariant DomPopStateEvent)`.

lib/web_ui/lib/src/engine/navigation/history.dart
```diff
@@ -183,15 +183,15 @@ class MultiEntriesBrowserHistory extends BrowserHistory {
   }
 
   @override
-  void onPopState(covariant DomPopStateEvent event) {
+  void onPopState(covariant DomPopStateEvent state) {
```
mdebbar, line 186: Can we keep it the name `event` or make it `Object? state`?

I have a feeling you were trying to change `DomPopStateEvent event ` ==> `Object? state` but forgot to finish the change, or reverted it partially.

lib/web_ui/lib/ui_web/src/ui_web.dart
```diff
@@ -0,0 +1,8 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+/// This library defines the web-specific additions that go along with web:ui
```
mdebbar, line 5:
```suggestion
/// This library defines the web-specific additions that go along with dart:ui
```

web_sdk/BUILD.gn
```diff
@@ -20,6 +20,7 @@ web_engine_libraries = [
   ":skwasm_stub_library",
   ":web_engine_library",
   ":web_ui_library",
+  ":web_ui_ui_web",
```
mdebbar, line 23: 😂

web_sdk/BUILD.gn
```diff
@@ -77,6 +78,11 @@ template("sdk_rewriter") {
       "--stamp",
       rebase_path(stamp_location, root_build_dir),
     ]
+
+    if (defined(invoker.is_public) && invoker.is_public) {
+      args += [ "--public" ]
+    }
```
mdebbar, line 84: It looks like this flag only affects the library name `$libraryName` vs `_$libraryName`. Could we just change library names (e.g. `library_name = "_engine"`) instead of introducing a new flag?

lib/web_ui/lib/src/engine/navigation/url_strategy.dart
```diff
@@ -5,60 +5,12 @@
 import 'dart:async';
 
 import 'package:ui/ui.dart' as ui;
+import 'package:ui/ui_web/src/ui_web.dart' as ui_web;
 
 import '../dom.dart';
 import '../safe_browser_api.dart';
 import 'js_url_strategy.dart';
```
mdebbar, line 12: My understanding is this will go away after we migrate `flutter_web_plugins` to use the Dart API directly, correct?

mdebbar (commented):
LGTM!

I don't have strong opinions on the library name, so I'll save you the bikeshedding from my end :)

lib/web_ui/lib/src/engine/navigation/history.dart
eyebrowsoffire: This just gets erased by the rewriter, and in the top-level library file it says `import 'dart:ui_web' as ui_web;` so for the original source files we can import it any way we like as long as it is imported under the name `ui_web`. I can try to find a cleaner way to do this import.

lib/web_ui/lib/src/engine/navigation/url_strategy.dart
eyebrowsoffire: Yep!

web_sdk/BUILD.gn
eyebrowsoffire: A man, a plan, a canal, panama

web_sdk/BUILD.gn
eyebrowsoffire: Not without doing a ton of changes across the codebase that actually change the original source files to all be underscore prefixed (and I think there is probably a reason it is this way, having to do with how package imports work vs library imports). The difference is that the original source files all treat it as `engine` and the transformed files treat it as `_engine`. It's not just a matter of changing the library name, because there is actually an "input library name" and "output library name". In the private case, that means adding an underscore, in the public case they are the same. So I can either have a `public` flag that tells us whether to add the underscore or not, or I can have an `inputLibraryName` and `outputLibraryName` or something like that that is specified by each transformation, but that seems too weird and generic and easy to mess up.

lib/web_ui/lib/src/engine/navigation/history.dart
eyebrowsoffire: Hmmm... you're right, I kind of did a half-measure here. I think I can clean this up a bit.

lib/web_ui/lib/src/engine/navigation/history.dart
eyebrowsoffire: For reasons that I don't really want to write a massive essay about right now (but would be happy to talk about more offline if you want) I don't want any of the surface area of this API to be JS interop objects. It gets complicated with `package:web` and some of the future stuff there.

lib/web_ui/lib/src/engine/navigation/history.dart
mdebbar: On second look, I think `package:ui/src/ui_web.dart` should work already?

lib/web_ui/lib/src/engine/navigation/history.dart
eyebrowsoffire: So what I ended up doing here is to basically unwrap the JS event in `JSUrlStrategy` before it actually gets to the darty `UrlStrategy` object, and so the pop state listener just takes the state itself. I think this is a bit better. For stuff on the framework side that uses actual JS interop to do the pop listener thing, they can just unwrap it there as well. The path through JSUrlStrategy remains unchanged, so I think this is non-breaking for the framework (but I'll find out as soon as I start testing against the framework side).

lib/web_ui/lib/src/engine/navigation/url_strategy.dart
```diff
@@ -150,15 +102,17 @@ class HashUrlStrategy extends UrlStrategy {
...
+class CustomUrlStrategy extends ui_web.UrlStrategy {
   /// Wraps the [delegate] in a [CustomUrlStrategy] instance.
   CustomUrlStrategy.fromJs(this.delegate);
 
   final JsUrlStrategy delegate;
 
   @override
-  ui.VoidCallback addPopStateListener(DomEventListener fn) =>
-      delegate.addPopStateListener(allowInterop(fn));
+  ui.VoidCallback addPopStateListener(ui_web.PopStateListener fn) =>
```
mdebbar, line 112: `HashUrlStrategy` (above) also needs a similar change. I'm surprised the analyzer isn't screaming at you since the method has a different signature now.

eyebrowsoffire: Good call!

yjbanov (approved):
lgtm

lib/web_ui/lib/ui_web/src/ui_web.dart
```diff
@@ -0,0 +1,8 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+/// This library defines the web-specific additions that go along with dart:ui
+library ui_web;
```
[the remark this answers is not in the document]
eyebrowsoffire: I definitely don't want a nameless library though. This is a library that the framework (and maybe eventually users) will use by name.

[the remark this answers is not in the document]
kevmoo: See also: https://github.com/flutter/engine/pull/40669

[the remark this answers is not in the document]
eyebrowsoffire: The sdk_rewriter will not be able to rewrite this file properly without this statement.

kevmoo:
That's okay. I don't think it's a huge deal

eyebrowsoffire:
I'm going to wait until after the branch cut to merge this.

PR #40613, opened 2023-03-24
Reland "[web] Migrate canvaskit_api to JS types."

joshualitt:
First patch is reland


joshualitt:
@eyebrowsoffire ptal. The bug was in `transform`. I audited the other methods and didn't see any other obvious mistakes like this, but then again most of the other functions with many arguments have a mixture of types so this sort of error would be less likely.

The changes to `module` and `exports` didn't seem to show up anywhere, but I'm reasonably sure we need them.

eyebrowsoffire (approved):
LGTM

PR #40617, opened 2023-03-25
[web] remove obsolete object caches; simplify native object management

yjbanov:
Remove obsolete object caches and introduce a simpler way to manage native objects:

* Remove the unused `SynchronousSkiaObjectCache`.
* Introduce new library `native_memory.dart` that's smaller and simpler than `skia_object_cache.dart`.
* Introduce two types of native object references:
  * `UniqueRef` a reference with a unique Dart object owner.
  * `CountedRef` a ref-counted reference with multiple Dart object owners.
* All native references use GC (via `FinalizationRegistry`) as a back-up.
* The new library removes everything related to object resurrection that was needed only in browsers that didn't support `FinalizationRegistry`. All browsers support it now.
* Remove the ad hoc `SkParagraph` cache that predates the introduction of `Paragraph.dispose`.
* Rewrite `CkParagraph` in terms of `UniqueRef`.
* Rewrite `CkImage` in terms of `CountedRef`; delete `SkiaObjectBox`.

This PR does not migrate all objects from the old `skia_object_cache.dart` to `native_memory.dart`. That would be too big of a change. The migration can be done in multiple smaller PRs.

This also removes a few unnecessary relayouts observed in https://github.com/flutter/flutter/issues/120921, but not all of them (more details in https://github.com/flutter/flutter/issues/120921#issuecomment-1481958762)


lib/web_ui/lib/src/engine/canvaskit/native_memory.dart
```diff
@@ -0,0 +1,206 @@
...
+///
+/// To prevent memory leaks, the underlying C++ object is deleted by the GC if
+/// it wasn't previously disposed of explicitly.
+class UniqueRef<T extends Object> {
+  UniqueRef(Object owner, T nativeObject, this._debugOwnerLabel) {
+    _nativeObject = nativeObject;
+    if (Instrumentation.enabled) {
+      Instrumentation.instance.incrementCounter('$_debugOwnerLabel Created');
+    }
+    _finalizationRegistry.register(owner, UniqueRef);
```
hterkelsen, line 33: should this be `finalizationRegistry.register(owner, this)`?

hterkelsen (changes requested):
We should add a test for when the finalization registry disposes GC'ed objects to test the line I pointed out in my review

lib/web_ui/lib/src/engine/canvaskit/native_memory.dart
yjbanov: Done. PTAL

yjbanov:
> We should add a test for when the finalization registry disposes GC'ed objects to test the line I pointed out in my review

Done. PTAL.

hterkelsen (approved):
LGTM

PR #40632, opened 2023-03-25
[Impeller] use half precision in blending, blur, texture shaders

jonahwilliams:
Depends on https://github.com/flutter/engine/pull/40590 and https://github.com/flutter/engine/pull/40616

Converts blending, blur, and texture filling shaders to use half precision when targeting metal. I've been testing the performance using "Profile with induced GPU performance state". In wonderous the menu blur over the carousel has roughly the following numbers:

```
	Half 	ToT
Max	19.61	23.64
Med	7.86	9.33
Min	5.69	6.48
```

We can of course profile more consistently, but I suspect this is a generally solid, if marginal, improvement for metal. This will have no effect on gles/vulkan backend as these platforms are already using relaxed precision.


jonahwilliams:
Breaking this down into smaller changes

PR #40636, opened 2023-03-26
[Impeller] Document tricky bits in EntityPass, make positioning less confusing

bdero:
Documenting to make stuff less confusing. TBH, I keep having to remind myself about why some of this stuff is here. 😅


bdero:
This also changes the behavior of the "position" parameter to be the local position of the current pass.

chinmaygarde (approved):

PR #40685, opened 2023-03-27
Revert "[web] Add `js_types` sources to build."

joshualitt:
Reverts flutter/engine#40597

May have broken the build.


bdero (approved):

bdero:
This was a speculative revert to fix a breakage in the Engine roll: https://github.com/flutter/flutter/pull/123562

Follow-up bug to catch these failures earlier is here: https://github.com/flutter/flutter/issues/123567

We found that https://github.com/flutter/engine/pull/40394 was the cause of the issue in the end, so this should be safe to reland.

PR #40686, opened 2023-03-27
Revert "Return an empty JavaScript object as a replacement for exports/module in patchCanvasKitModule"

bdero:
Reverts flutter/engine#40582

To fix possible breakage of engine roll.


jonahwilliams (approved):
LGTM

bdero:
FYI @jason-simmons: FYI, we speculatively reverted this along with some other PRs because the tree was getting so far behind. We found that https://github.com/flutter/engine/pull/40394 was the cause of the issue, so this should be safe to reland.

PR #40697, opened 2023-03-28
Support the GN no-LTO flag for Wasm builds

jason-simmons:


yjbanov (approved):

jason-simmons:
Abandoning this in favor of the approach used in https://github.com/flutter/engine/pull/40733

PR #40698, opened 2023-03-28
Reland "Return an empty JavaScript object as a replacement for exports/module in patchCanvasKitModule (#40582)"

jason-simmons:


yjbanov (approved):
LGTM. cc @hterkelsen

PR #40703, opened 2023-03-28
[iOS] Send connectionClosed message when resignFirstResponder to ensure framework focus state is correct.

luckysmg:
Fix:
- https://github.com/flutter/flutter/issues/123523

### Before this patch:
see in linked issue.

### After this patch:

iPhone | iPad
---|---
 | 

## Pre-launch Checklist

- [x] I signed the [CLA].


cyanglaz (approved):
LGTM % nits

cyanglaz:
> Blocked by:

https://github.com/flutter/flutter/issues/123523

Did you mean to link to a framework PR instead?

shell/platform/darwin/ios/framework/Source/FlutterTextInputDelegate.h
```diff
@@ -59,7 +59,8 @@ typedef NS_ENUM(NSInteger, FlutterFloatingCursorDragState) {
     insertTextPlaceholderWithSize:(CGSize)size
                        withClient:(int)client;
 - (void)flutterTextInputView:(FlutterTextInputView*)textInputView removeTextPlaceholder:(int)client;
-- (void)flutterTextInputViewDidResignFirstResponder:(FlutterTextInputView*)textInputView;
+- (void)flutterTextInputViewDidResignFirstResponder:(FlutterTextInputView*)textInputView
```
jmagman, line 62: Agree with @cyanglaz here, it should be `flutterTextInputView:didResignFirstResponderWithClient` or something similar, that would also match the other methods in this file.

jmagman (changes requested):

LongCatIsLooong:
There's already this message: https://master-api.flutter.dev/flutter/services/TextInputClient/connectionClosed.html

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
```diff
@@ -1043,7 +1043,8 @@ - (BOOL)canBecomeFirstResponder {
 - (BOOL)resignFirstResponder {
   BOOL success = [super resignFirstResponder];
   if (success) {
-    [self.textInputDelegate flutterTextInputViewDidResignFirstResponder:self];
+    [self.textInputDelegate flutterTextInputViewDidResignFirstResponder:self
```
LongCatIsLooong, line 1046: nit: Looks like this one will be called when either the Flutter framework or UIKit tells the input view to give up its first responder status. It seems redundant for the former since there's no need to inform the Flutter framework when it is the one that initiated the focus change.

shell/platform/darwin/ios/framework/Source/FlutterTextInputDelegate.h
hellohuanlin: nit: maybe `didResignFirstResponderWithTextInputClient` since it's unclear what a client is.

luckysmg:
> There's already this message: https://master-api.flutter.dev/flutter/services/TextInputClient/connectionClosed.html

Oh.thx for reminding me that. I will try invoke that method to unfocus in framework, you can also take a look that framework pr😄

shell/platform/darwin/ios/framework/Source/FlutterEngineTest.mm
```diff
@@ -13,9 +13,15 @@
 #import "flutter/shell/platform/darwin/ios/framework/Source/FlutterBinaryMessengerRelay.h"
 #import "flutter/shell/platform/darwin/ios/framework/Source/FlutterDartProject_Internal.h"
 #import "flutter/shell/platform/darwin/ios/framework/Source/FlutterEngine_Test.h"
+#import "flutter/shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.h"
 
 FLUTTER_ASSERT_ARC
 
+@interface FlutterEngine ()
+- (void)flutterTextInputViewDidResignFirstResponder:(FlutterTextInputView*)textInputView
```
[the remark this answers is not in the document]
luckysmg: In framework side, will use client to identify the msg. So I think the client arg is required.

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
luckysmg: Yes. When framework close the keyboard first, this will send a msg to framework and will handle return by 

https://github.com/flutter/flutter/blob/659ba386f6d045690f9783c79fe5d320fb7847a1/packages/flutter/lib/src/services/text_input.dart#L1788

And only when platform starts to unfocus, the `_currentConnection!._client.connectionClosed();` will be called. 

So I think this cost can be acceptable?

shell/platform/darwin/ios/framework/Source/FlutterEngineTest.mm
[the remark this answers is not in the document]
luckysmg: Done.

jmagman (approved):
This LGTM, would like @LongCatIsLooong to confirm concerns were addressed.

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
LongCatIsLooong: Yeah that makes sense the old connection is already torn down by the framework so messages associated with the old client id will be thrown away.

LongCatIsLooong (approved):

luckysmg:
Thx everyone reviewing this.  Applying label for merging O(∩_∩)O

PR #40707, opened 2023-03-28
[Impeller] Gaussian blur: Add alpha mask specialization

bdero:
Adds a second bit to the gaussian blur specializations to get rid of the second texture sampling for the most common blur case.


jonahwilliams (approved):
LGTM

PR #40717, opened 2023-03-28
Default the CanvasKit base URL to local artifacts.

hterkelsen:
Reverts flutter/engine#40700

A fix for the g3 bug b/275455289 is out for review. We should be able to safely reland the original change.


yjbanov:
Renamed the PR according to https://github.com/flutter/flutter/wiki/Tree-hygiene#avoid-revert-revert-revert-revert-fix-foo-commit-messages

yjbanov (approved):
lgtm

itsjustkevin (approved):
Thank you for the quick fix!  

Is there a possible test that could cover this case in the future?

hterkelsen:
> Thank you for the quick fix!
> 
> Is there a possible test that could cover this case in the future?

Good idea. I'll see if I can add one to my PR in google3

PR #40719, opened 2023-03-28
[Impeller] support half-precision vertex data for iOS

jonahwilliams:
Work towards https://github.com/flutter/flutter/issues/123501

Ammends the compiler to support VertexData that contains half precision data. If utilized, this would imply automatic conversion when appending to a VBB subclass. Writing directly to the host buffer without conditionally specifying a float16 type would lead to incorrect runtime behavior.

While our vertex shaders are generally much faster than our fragment shaders, it can still be advantageous to reduce precision as it reduces the amount of data copied per frame. Still, needs more investigation on ergonomic ways to copy the data without accidentally introducing runtime regressions


jonahwilliams:
Given that we're having some issues with performance/precision when changing these I'm going to icebox this for now

PR #40720, opened 2023-03-28
Handle removed shaders more gracefully in malioc_diff.py

zanderso:


bdero:
Instead of doing partial diffs on CI, would it make sense to just generate a blessed file and literally run `diff ./impeller/tools/malioc.json [after.json]`?

zanderso:
> Instead of doing partial diffs on CI, would it make sense to just generate a blessed file and literally run `diff ./impeller/tools/malioc.json [after.json]`?

Yeah, there are a couple of different directions we could go. One thought I had was that when there is a change in performance in an existing shader, then having the prettier message is useful. But if that's not actually the case, then we can tear out a bunch of code. OTOH, if it is useful, then I'd go even further in the same direction and pull the human readable display names and verbose descriptions out of the malioc json output as well.

bdero:
Having a readable interpretation seems valuable to me. Perhaps we could do both and get the best of both worlds. The primary value in printing the full diff for me would be the ability to just copy it from CI and apply it using a patch tool in cases where I forgot (or can't) run malioc locally.

PR #40732, opened 2023-03-28
Add an option to malioc_diff.py to print a unified diff

zanderso:
This PR adds a `--print-diff` option to `malioc_diff.py` to make it print a unified diff when it finds a difference. After this lands, I'll send a recipe CL to make CI pass the new flag.

### DIFF
--- a/impeller/tools/malioc_diff.py
+++ b/impeller/tools/malioc_diff.py
@@ -4,6 +4,7 @@
 # found in the LICENSE file.
 
 import argparse
+import difflib
 import json
 import os
 import sys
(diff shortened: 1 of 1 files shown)


impeller/tools/malioc_diff.py
```diff
@@ -303,6 +317,23 @@ def main(argv):
...
+          before_json, sort_keys=True, indent=2
+      ).splitlines(keepends=True)
+      after_lines = json.dumps(
+          after_json, sort_keys=True, indent=2
+      ).splitlines(keepends=True)
+      before_path = os.path.relpath(
+          os.path.abspath(args.before), start=SRC_ROOT
+      )
+      diff = difflib.unified_diff(
+          before_lines, after_lines, fromfile=before_path
```
bdero, line 331: Nice, didn't know difflib was a thing

bdero (approved):
LGTM

PR #40733, opened 2023-03-28
Disable LTO in builds of CanvasKit to reduce binary size

jason-simmons:


hterkelsen (approved):
LGTM

third_party/canvaskit/BUILD.gn
```diff
@@ -15,6 +15,9 @@ wasm_toolchain("canvaskit") {
     skia_use_libjpeg_turbo_decode = true
     skia_use_libpng_decode = true
     skia_use_libwebp_decode = true
+
+    # Disable LTO.
+    enable_lto = false
```
mdebbar, line 20: If this is going to be the same for all CanvasKit builds, you can also add it to [`tools/gn`](https://github.com/flutter/engine/blob/main/tools/gn#L607). But this is good too.

mdebbar (approved):

third_party/canvaskit/BUILD.gn
jason-simmons: Adding it to `tools/gn` would conflict with skwasm, which must always be built with LTO enabled

kevmoo:
@jason-simmons – we're are super confident that

1) this is a solid size win (how much?)
2) there are ~zero perf regressions?

jason-simmons:
After disabling LTO I saw a 13% reduction in the size of `canvaskit.wasm` in my local `wasm_release` build.  I did not notice significant performance differences in local runs of `web_benchmarks_canvaskit` when disabling LTO.

I can wait a day before landing this PR in order to get more runs of the benchmarks on the Skia dashboard with `-O3` and LTO.  After that we can check if the dashboard reports any change from disabling LTO.

jason-simmons:
Also, the builds of CanvasKit that we had previously obtained from Skia were not built with LTO.

PR #40735, opened 2023-03-28
[Impeller] migrate texture fill shaders to half precision.

jonahwilliams:


dnfield (approved):

impeller/compiler/shader_lib/impeller/texture.glsl
```diff
@@ -92,6 +93,26 @@ vec4 IPSampleWithTileMode(sampler2D tex,
   return texture(tex, coords);
 }
 
+const float16_t kTileModeDecalHf = 3.0hf;
+
+/// Sample a texture, emulating a specific tile mode.
+///
+/// This is useful for Impeller graphics backend that don't have native support
+/// for Decal.
+f16vec4 IPSampleWithTileModeHf(f16sampler2D tex,
```
bdero, line 102: Maybe `IPHalfSampleWithTileMode` to follow the pattern of putting the type info near the beginning? Not sure the best route to go for this, but this is public and available to Flutter users as well. We should probably document some kind of convention for these at some point.

bdero (approved):
LGTM

impeller/compiler/shader_lib/impeller/texture.glsl
jonahwilliams: I thought we only exported runtime effect?

jonahwilliams: (Renaming done)

jonahwilliams: At a certain point though, we'll probably drop this and only have a half precision verison.

PR #40705, opened 2023-03-28
Revert "[canvaskit] Use FontCollection API from CanvasKit to mitigate caching issues"

hterkelsen:
Reverts flutter/engine#40673

Causing breakages here: https://github.com/flutter/flutter/pull/123582

PR #40740, opened 2023-03-29
Reland "[canvaskit] Use FontCollection API from CanvasKit to mitigate caching issues"

hterkelsen:
Reverts flutter/engine#40705

The original failed framework tests because the local CanvasKit change was reverted right before it was submitted.


yjbanov (approved):
lgtm

PR #40743, opened 2023-03-29
[draft] Embedder api background message

cyanglaz:


Hixie:
@cyanglaz Is this still on your radar?

PR #40748, opened 2023-03-29
[Impeller] Use NSAutoReleasePool for Metal playgrounds.

dnfield:
Fixes https://github.com/flutter/flutter/issues/117339

I judged this better than having a separate ObjC++ implementation for macOS playgrounds.


chinmaygarde (approved):
Oh, makes sense! Could this also explain the [leak of the command buffers](https://github.com/flutter/engine/commit/6a274fe3515dda0e80aca9c01e7f9fdd40455f1d)? The buffers hold onto resources and that could explain the massive leaks. Perhaps we can back out that commit now.

I also see a [scoped NSObject for autorelease pools in fml](https://github.com/flutter/engine/blob/1947b95c506078192584e929112216d5fce79c79/fml/platform/darwin/scoped_nsobject.h#L153). But the implementation is missing and the comment mentions a missing alternative. In a separate patch, we should clean that up move these utilities to FML.

Either way, lgtm. This is a massive improvement to the playgrounds experience.

bdero (approved):
LGTM

dnfield:
The command buffer leak is a separate issue - applying this patch doesn't seem to help it (they're actual leaks, whereas this is just flushing the autorelease pool on each turn of the loop)

PR #40752, opened 2023-03-29
[Impeller] migrate Gaussian shaders to half precision.

jonahwilliams:


chinmaygarde (approved):

jonahwilliams:
Had to update malioc diff because I forgot to opt gaussian.glsl into relaxed precision before!

PR #40762, opened 2023-03-29
[Impeller] remove advanced blend special casing

jonahwilliams:
From my understanding with the branching we're essentially always sampling from the source sampler anyway. Can we special case this and fill in the foreground with a host visible texture.


jonahwilliams:
From the malioc diff doesn't seem worthwhile

PR #40764, opened 2023-03-29
Create a shared pointer mechanism for DisplayList objects

flar:
Currently DL has been using std::shared_ptr which has some drawbacks including the inability to get a shared copy of yourself or move an already allocated pointer into a shared_ptr which is useful in several cases. There is a "shared_from_this" mechanism that std provides, but it comes with some drawbacks.

In some places "sk_sp" is being used and that does have the ability to create shared managed pointers from raw pointers, but will become an unnecessary dependency on the Skia source base as we migrate to a neutral graphics front end.

This PR creates `class DlShareable` which is inherited by all shareable classes and a `dl_shared<T>` and `dl_weak_shared<T>` set of templates for holding counted references (both strong and weak) to any `DlShareable` object. It has a couple of improvements over the `sk_sp` mechanism that work well for our use case, most importantly that the developers never need to manipulate the ref counts manually for any operation (gone are the `sk_ref_sp` et al template functions).


jonahwilliams:
> There is a "shared_from_this" mechanism that std provides, but it comes with some drawbacks.

What are the drawbacks of this approach? Isn't this more or less what we'd need this for?

display_list/dl_shared.h
```diff
@@ -0,0 +1,508 @@
...
+
+#ifndef FLUTTER_DISPLAY_LIST_DL_SHARED_H_
+#define FLUTTER_DISPLAY_LIST_DL_SHARED_H_
+
+#include <cstddef>
+#include <cstdint>
+
+#include "fml/logging.h"
+
+// #define DL_SHAREABLE_STATS
```
dnfield, line 13: Commented out code

display_list/dl_shared.h
```diff
@@ -0,0 +1,508 @@
...
+/// ----------------------------------------------------------------------
+/// A template class to manage weak shared pointers to objects which
+/// implement a compatible API, such as the DlShareable class defined
+/// below.
+///
+/// Assigning a pointer into a |dl_weak_shared| instance will automatically
+/// add a weak reference to the object which does not indicate an active
+/// interest in maintaining the object, but will nonetheless prevent its
+/// destruction until all of both weak and strong shared references are
+/// destroyed or replaced.
```
dnfield, line 160: I might be misreading this but this doesn't seem to be what a weak pointer is.

IOW, if I have this code:

```cpp
auto strong = std::make_shared<int>(1);
std::weak_ptr<int> weak= strong;
strong.reset();
```

I'd expect `weak` to be `nullptr` now. This description and implementation leads me to believe that's not the case.

display_list/dl_shared.h
```diff
@@ -0,0 +1,508 @@
...
+///
+/// A dl_weak_shared instance is not meant to be accessed from multiple
+/// threads, though separate dl_weak_shared instances in different threads
+/// may all point to the same underlying shareable object as long as its
+/// reference management methods are themselves thread-safe (as they are
+/// in the standard implementation of |DlShareable|.
```
dnfield, line 166: This has caused a large number of headaches in `fml::WeakPtr`. I would like to not reproduce them.

`fml::WeakPtr` is not thread safe, `std::weak_ptr` is. People never remember which thread they're allowed to use `fml::WeakPtr` on and it causes problems, especially when the code being written only indirectly accesses `fml::WeakPtr`s. We added debugging checks but the support for them vary across platforms (so on some platforms you don't get much helpful information when you violate the rules, but on others you do, and even then sometimes that information isn't quite sensible).

dnfield: Put more strongly: Please do not create a new weak pointer type that is not threadsafe.

display_list/dl_shared.h
flar: That's how the weak version of sk_sp is defined.

dnfield:
> move an already allocated pointer into a shared_ptr which is useful in several cases. 

I'd like to hear more about this. Is the problem that we have a lot of code that does this and it'd be tedious to refactor, or that this is a pattern that we really want to adopt?

Without understanding the context, I'm inclined to say that `std::shared_ptr` has the right idea here and `SkRefCnt` is weird.

display_list/dl_shared.h
flar: dl_shared and sk_sp aren't full objects, they are templated wrappers around a pointer that ensure that proper ref-counting happens to the object.

flar:
> > move an already allocated pointer into a shared_ptr which is useful in several cases.
> 
> I'd like to hear more about this. Is the problem that we have a lot of code that does this and it'd be tedious to refactor, or that this is a pattern that we really want to adopt?
> 
> Without understanding the context, I'm inclined to say that `std::shared_ptr` has the right idea here and `SkRefCnt` is weird.

Consider `Foo::MakeModifiedFoo(...modification attributes...)`

The result may be "those attributes don't change anything about me, the caller can just reuse my instance". If this was called via std::shared_ptr, the method would have to `return std::make_shared(me)` which constructs an entirely new object. This happens a lot in the filters and such. Some cases include "makeWithLocalMatrix" where the ImageFilter ignores the matrix and so can just return "itself", or Compose(outer_filter, inner_filter) when one of the children is null, it can just return the other child. Sometimes the caller isn't aware that they have a null filter because they just went through the trouble of calling a factory, but they didn't realize that the factory decided that the operation they described was a NOP and so it returned a nullptr.

I was aware of these cases when I originally decided to just use `std::shared_ptr` but when I looked around and saw so many calls to the `shared()` method (which tells an attribute to reconstruct itself inside a brand new shared ptr) I realized I was just creating a recipe to duplicate and copy memory everywhere.

flar:
> > There is a "shared_from_this" mechanism that std provides, but it comes with some drawbacks.
> 
> What are the drawbacks of this approach? Isn't this more or less what we'd need this for?

There was a long internal thread that I should turn into a document. It looks like there were concerns raised but then we found workarounds for pretty much all of them. To summarize:

Depending on how you expose constructors `shared_from_this` could dereference a missing pointer. It relies on the original creation of the object having been created as a shared_ptr and if that doesn't happen then it won't be set up right. The recommendations are typically to always use factory methods as a result. There are also some caveats to use it with placement new which we do use, but there are guidelines on how to make that work.

Having said that, it looks like my "best practices" suggestions for dl_shared will be to always use factory methods. With dl_shared, there is nothing stopping you from doing `new ShareableFooThingy()` yourself and it will work with dl_shared, but stack allocation can create problems. Stack allocation for shared_from_this situations cannot be done either. So, we have:

`dl_shared`
- size of a reference: 1xptr
- added to the object itself: 1xrefcnt (or 2xrefcnt if weak is supported)
- factory constructors: recommended
- new: supported if you follow guidelines
- stack alloc: supported if you never share it, but strongly un-recommended

`std::shared_ptr` with support for `shared_from_this`
- size of a reference: 2xptr
- added to the object itself: 1xweak_shared == ?2xptr?
- factory constructors: recommended
- new: not allowed
- stack alloc: not allowed

So, they are very close if we insist on private constructors and only public factories. dl_shared is a little more efficient in terms of memory usage.

display_list/dl_shared.cc
```diff
@@ -0,0 +1,28 @@
...
+namespace flutter {
+
+// Normally exists just to ensure that the header can be cleanly imported.
+
+#ifdef DL_SHAREABLE_STATS
+
+int64_t DlShareable::total_made_ = 0;
+int64_t DlShareable::total_disposed_ = 0;
+void DlShareable::report_shareable_counts() {
+  FML_DCHECK(total_disposed_ <= total_made_);
```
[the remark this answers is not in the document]
flar: It's mainly debug code enabled with that #define and then seeing if the number of shared objects goes to zero over the life of running a complicated app. I suppose I should make them atomic to be safe.

display_list/dl_shared.h
flar: One of the issues with a truly weak pointer is that you need some way to make the weak references disappear when the object is collected - and that creates threading issues for the pointers themselves.

With `sk_sp` and `dl_shared`, the weakness is not imposed by sabotaging the pointer from somewhere else when the object is destructed, so the wrapper doesn't ever need to interact with other threads.

One possible use case that is supported by dl_weak_shared as compared to a more proactive weak ptr mechanism is a cache of precomputed data. If you frequently see requests for a particularly complicated set of data, you can hold that in a cache with purely weak pointers. When a new request comes in, you can quickly check if you already have the answer in the cache and reuse it. You can't do that if the weak pointers are sabotaged when the primary object goes away.

flar: A down-side of this kind of weak pointer, though, is that you need to periodically check if you are holding on to stale data.

dnfield:
> dl_shared is a little more efficient in terms of memory usage

It saves `sizeof(uintptr_t)` per instance, and instance counts for these things in a normal Flutter program are typically going to be in the 1000s. 

IMHO `std::shared_from_this` forbidding things we recommend you don't do is a feature.

dnfield:
> If this was called via std::shared_ptr, the method would have to return std::make_shared(me) which constructs an entirely new object

Why can't the method return the same `std::shared_ptr` that was passed in? Or return a new shared_ptr instance pointing to the same memory?

display_list/dl_shared.h
jason-simmons: This could potentially be useful for caches.  The cache can hold a weak reference to the object and later scan the entries for objects that have no remaining strong references and therefore can be evicted.

I agree that the "weak reference" naming may mislead users into thinking that this behaves like `std::weak_ptr` where a weak reference will not prevent the object from being destructed.

dnfield: We already have caches like this using `std::weak_ptr` which have the extra advantage of not needing to wait for the scan operation to free memory in the cache (Although you still need to make sure to let go of the weak_ptr object itself), e.g. in https://github.com/flutter/engine/blob/293cab1a7dccdbbdf0303a00095b86b2f62b8850/common/graphics/texture.h#L106. [Moving to std::weak_ptr](https://github.com/flutter/engine/commit/e0625641f3fb677544d9585b6a32b46a58f962c1) resulted in a pretty marked performance benefit in the case of the volatile path tracker.

flar: There are different uses of these in caches. If it is the key in the cache, then you'd want it to automatically go away when the strong references are gone. If it is the value, then it might have a strong reference while that value is being used, but otherwise may live most of its useful life with only a weak reference from the cache waiting to be re-strengthened when you encounter its key again.

display_list/dl_shared.h
flar: fml::WeakPtr specifically says it cannot be handed to another thread, but must be recreated in the new thread with a new WeakFactory.

With dl_shared, you are responsible for the data and ops of your object being thread-safe. DlShareable adds some fields to your object, but it implements those in a thread-safe manner. the dl_shared() really are just syntactic wrappers so the only thing you would need to worry about is sharing them as a ref, as in `const dl_shared<Foo>&`. That's pretty efficient within a thread (and, in fact, you could just share the pointer and be equally as efficient). If you share that between threads, then neither thread should try to swap in a new referent, but either thread could still do refcount operations.

flar: We don't currently use the weak version of this. I included it as a proof of concept. I could delete it and nothing would get broken right now.

It would be easy to add back if we like this concept.

Or, we could also create some sort of auto-self-destruct weak pointers as well? Or can a traditional weak_ptr wrap a dl_shared<Foo>?

dnfield: There's no periodic checking - you call `weak_ptr.lock()` to turn it into a shared pointer if possible, or null. While you have that shared instance you now have a strong reference. Once you let go of that shared instance the weak pointer serves as your only weak reference again.

dl_shared won't play nicely with the STL types - any more than SkRefCnt or fml::RefCnt do. It's a bit of a headache.

The documentation on fml::WeakPtr is referring to actually accessing the underlying data, e.g. `fml_weak.get()` can only be done on the thread that the factory is on - but you can use instances of fml::WeakPtrs on various threads to eventually get to the right thread to use the `.get()` call on again.

display_list/dl_shared.h
dnfield: While you're using the cache entry you call `.lock()` on the weak reference to find out if you can get a strong reference to it or not.

flar:
> > If this was called via std::shared_ptr, the method would have to return std::make_shared(me) which constructs an entirely new object
> 
> Why can't the method return the same `std::shared_ptr` that was passed in? Or return a new shared_ptr instance pointing to the same memory?

You can return the shared_ptr you were handed if it is an argument to a method.

You can't return `shared_ptr<this>` because your shared wrapper was dereferenced away when the method was called. So, `shared_ptr<Foo>->doSomething()` never sees the shared_ptr to itself. That's where the fragile magic of shared_from_this comes in.

flar:
> > If this was called via std::shared_ptr, the method would have to return std::make_shared(me) which constructs an entirely new object
> 
> Why can't the method return the same `std::shared_ptr` that was passed in? Or return a new shared_ptr instance pointing to the same memory?

You can't have multiple shared_ptr instances pointing to the same memory. They own the instance. To get a new shared_ptr you have to clone from the original shared_ptr or create a whole new instance to wrap with the new shared_ptr.

display_list/dl_shared.h
flar: weak_ptr.lock() can only become a shared ptr if the underlying object is still alive.

flar:
> > dl_shared is a little more efficient in terms of memory usage
> 
> It saves `sizeof(uintptr_t)` per instance, and instance counts for these things in a normal Flutter program are typically going to be in the 1000s.

If you support `shared_from_this` it also costs you a weak ptr of some sort per object. dl_shared gets that functionality for free.

flar:
I'm going to make a prototype variant that uses `shared_ptr` and `shared_from_this` instead of dl_shared and see how it compares.

I've already confirmed that dl_shared<Foo> is 8-bytes (on a 64-bit arch) and shared_ptr<Foo> is 16.
DlShareable is 16 bytes with a vtable and both strong and weak counts.
enable_shared_from_this is 24 bytes.

Minor size differences, but I also need to look at how much code gets executed to pass them around. dl_shared is essentially free to pass since it is just a conceptual wrapper around a raw pointer.

But first, I'm going to just try replacing DlShareable with enable_shared_from_this and dl_shared with shared_ptr and see if the code still provides the same mechanisms and ease of use.

display_list/dl_shared.h
flar: It looks like Sk weak pointers are a sort of hybrid. On the surface, they work like I implemented `dl_weak_shared` in that the weak counts keep the object from being deleted. However, when the count goes to "all weak", they do call a virtual method that you could implement to dump your aggregate data so that while your base object remains alive, at least it doesn't hold on to any large resources when it is in the weak state. That is totally up to the object that inherits from the `SkWeakRefCnt` base class, though.

flar:
The `std::shared_ptr` using `shared_from_this` variant is https://github.com/flutter/engine/pull/40918

dnfield:
> I've detailed the issue in this godbolt example. Towards the bottom you can find a couple of lines commented out. If you restore them then the example compiles and runs, but gets a memory corruption due to double-freeing the multi-shared foo object.
> 
> https://godbolt.org/z/sarrjKG89

Doesn't this violate the design guidelines in the reference for enabled_shared_from_this? My reading of https://en.cppreference.com/w/cpp/memory/enable_shared_from_this says that any class that extends enable_shared_from_this should have no public constructors.

flar:
> > I've detailed the issue in this godbolt example. Towards the bottom you can find a couple of lines commented out. If you restore them then the example compiles and runs, but gets a memory corruption due to double-freeing the multi-shared foo object.
> > https://godbolt.org/z/sarrjKG89
> 
> Doesn't this violate the design guidelines in the reference for enabled_shared_from_this? My reading of https://en.cppreference.com/w/cpp/memory/enable_shared_from_this says that any class that extends enable_shared_from_this should have no public constructors.

The case where it happened in the wild had no public constructors. I can rewrite the example without a public constructor, but that isn't the problem here. The problem is that shared_ptr() has a public constructor and we are not in control of that. If the only way to create a shared_ptr was via "make_shared" where the shared_ptr mechanism controlled the lifetime of the embedded object there would be no issue. But, since shared_ptr supports either using make_shared or just doing the new yourself and constructing a shared_ptr from it, it has no protection against constructing a second shared_ptr from an existing pointer.

The problem identified is inherent to shared_ptr and cannot be prevented by the code using it.

flar:
Note that because of the non-public constructor, in the previous example you need to use the shared_ptr constructor on a "new Foo()" because make_shared cannot access private constructors. The difference between make_shared and the shared_ptr constructor is that the control block can only be inlined with the object in the make_shared case.

flar:
I beefed up the stats in the dl_shared module and ran the gallery benchmark on top of it on a Pixel 6a. There were about 800 frames benchmarked, but then the test continues running the app through its paces after that for many more transitions, so I'd guess there may have been a couple thousand frames at least by the time it exited.

- Total DlShareable objects created was only around 80 meaning we rarely create these objects and reuse them a lot. Per-object memory bloat should be negligible.
- Total times we placed one of these into a DL was only around 100 meaning that when we used these, we tended to keep their DL around so placement doesn't happen often. The total number of times we dispatched one of these objects to, say, the Impeller renderer was around 2500 times which means it was only 1 or 2 per frame if we assume the benchmark rendered 1000 frames in total, but I'm pretty sure there were more total frames than that.
- Total inc/decrefs was around 3600 which again is around a couple per frame, so the cost of doing the inc/decref should also be negligible.

The question is, how typical is the gallery transitions test (not new gallery, the regular one)?

display_list/dl_shared.h
flar: > I might be misreading this but this doesn't seem to be what a weak pointer is.
> 
> IOW, if I have this code:
> 
> ```c++
> auto strong = std::make_shared<int>(1);
> std::weak_ptr<int> weak= strong;
> strong.reset();
> ```
> 
> I'd expect `weak` to be `nullptr` now. This description and implementation leads me to believe that's not the case.

The weak pointer there would behave as if it were null at that point, but the object would not have been freed. That is the default behavior of the `dl_weak_shared` as I prototyped it, but I'm not planning to go with that prototype anyway.

Basically, `make_shared` will allocate the std control block inline with the object. When the strong reference goes away, the control block must survive which means the control block cannot be freed which means the object memory cannot be freed either.

The `dl_weak_shared` that I originally wrote, just as an exercise in prototyping, also had a similar behavior in that `strengthen()` would return `nullptr` unless you passed an argument that indicated whether you wanted to strengthen it even if it was a zombie. The object wouldn't be freed until all references disappeared - similar to the memory allocation behavior of your example.

Potential differences are that I haven't looked into when the `std` mechanism would call the destructor method on the object, but since it tends to use the default deleter, it may not be until it is ready to free the memory which means that it will wait until the last strong or weak reference is dropped. `sk_sp` at least provides a virtual method you can override to "clean up" the object when it enters the zombie state even if the memory itself is not going away.

flar: > Potential differences are that I haven't looked into when the `std` mechanism would call the destructor method on the object

It looks like the object destructor will be called whether the shared pointer was constructed using the manual constructor or via make_shared. This means that the only memory that is held after all strong references go away is the base allocation for the object body itself.

flar:
This work has been deprioritized in favor of other work. I'll reopen the PR when it becomes more relevant.

PR #40765, opened 2023-03-29
Revert "Revert "Listen to `WM_CLOSE` message on Windows to allow framework to…"

yaakovschectman:
Reverts flutter/engine#40739

Ignore this for now if you see it. I need to greatly rework how messages are handled before we can consider re-integrating this.


yaakovschectman:
Will need to create another branch, I think.

PR #40770, opened 2023-03-29
[Impeller] backfilled golden tests from playground tests in aiks_unittests.

gaaclarke:
issue: https://github.com/flutter/flutter/issues/123790

## Pre-launch Checklist

- [x] I signed the [CLA].


gaaclarke:
I've approved all changes since they look roughly right (modulo https://github.com/flutter/flutter/issues/123789).  It's worth just getting these landed and iterating on them as we want.

impeller/aiks/BUILD.gn
```diff
@@ -31,17 +31,43 @@ impeller_component("aiks") {
...
+    "//flutter/impeller/golden_tests:golden_playground_test",
+    "//flutter/impeller/playground:playground_test",
+    "//flutter/testing:testing_lib",
+  ]
+}
+
+impeller_component("aiks_unittests_golden") {
+  testonly = true
+
+  defines = [ "IMPELLER_GOLDEN_TESTS" ]
```
dnfield, line 62: nit: consider making a GN template that just defines this rather than defining it in each invocation of `impeller_component` that wants it.

impeller/golden_tests/golden_playground_test_mac.cc
```diff
@@ -0,0 +1,139 @@
...
+          "impeller_Play_AiksTest_GradientStrokesRenderCorrectly_Metal" ||
+      test_name ==
+          "impeller_Play_AiksTest_"
+          "CoverageOriginShouldBeAccountedForInSubpasses_Metal" ||
+      test_name == "impeller_Play_AiksTest_SceneColorSource_Metal" ||
+      test_name == "impeller_Play_AiksTest_CanRenderLinearGradient_Metal") {
+    GTEST_SKIP_(
+        "GoldenPlaygroundTest doesn't support interactive playground tests "
+        "yet.");
+  }
```
dnfield, line 92: It'd be nice if we could somehow check if the test actually used `ImGui` or not. This will be brittle, but it's probably just temporary right?

gaaclarke: I looked into it, the problem is that we have to call `GTEST_SKIP` inside of `SetUp` so there's no way (at runtime) to know if a test is going to use ImGui in `SetUp` =T

I think we should probably abolish any ImGui tests since they can't be automated.

gaaclarke:
I'm going to rebase this to capture brandon's latest fix which should change a golden.

impeller/golden_tests/golden_playground_test_mac.cc
dnfield: Alternatively we move them into some other target that makes it clear they'r enot automated tests.

gaaclarke: Yea I was thinking about something similar, maybe making the target something like `impeller_playground`.  I'm not sure if we want to just move everything over to goldens though.  In which case there would be no need for that.  I'm still trying to feel out the problem.

PR #40771, opened 2023-03-29
Fix autofill for android

bparrishMines:
Fixes https://github.com/flutter/flutter/issues/92165.

Temporary fix for plugin: https://github.com/flutter/packages/pull/3588

## Pre-launch Checklist

- [x] I signed the [CLA].


shell/platform/android/test/io/flutter/embedding/android/FlutterViewTest.java
```diff
@@ -92,6 +92,15 @@ public void attachToFlutterEngine_alertsPlatformViews() {
     verify(platformViewsController, times(1)).attachToView(flutterView);
   }
 
+  @Test
+  public void flutterView_importantForAutofillDoesNotExcludeDescendants() {
```
reidbaker, line 96: Is it possible to test not that the value is set but instead that a child view can have autofill? 
This feels a bit like a change detector test.

reidbaker: Does this test need to indicate that it should only run for android o and above?

LongCatIsLooong (approved):
LGTM

reidbaker (approved):
@bparrishMines approved independent of a different method of testing.

shell/platform/android/test/io/flutter/embedding/android/FlutterViewTest.java
bparrishMines: > Is it possible to test not that the value is set but instead that a child view can have autofill?
This feels a bit like a change detector test.

I wasn't able to find a way to test if child views receive autofills automatically. I tried to manually call [autofill](https://developer.android.com/reference/android/view/View#autofill(android.view.autofill.AutofillValue)) with no luck. And the documentation only provides an example to test it manually: https://developer.android.com/guide/topics/text/autofill-optimize.

> Does this test need to indicate that it should only run for android o and above?

I believe the method is only callable on versions O and above. Without the version check, we would get lint warnings that the method would cause crashes on lower versions.

PR #40772, opened 2023-03-30
[Impeller] Fail GN when building an enabled Impeller backend is impossible.

bdero:
When `--enable-impeller-*` args write `impeller_enable_*` args, and so `declare_args` in `impeller.gni` are ignored and it'll always attempt to build enabled backends even if they will obviously fail.

I ran into some confusion when accidentally flipping on Vulkan for iOS: https://gist.github.com/bdero/fa11a596ece56055d50e5afecfd4d03d

### DIFF
--- a/impeller/renderer/backend/BUILD.gn
+++ b/impeller/renderer/backend/BUILD.gn
@@ -8,14 +8,17 @@ group("backend") {
   public_deps = []
 
   if (impeller_enable_metal) {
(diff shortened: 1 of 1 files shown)


bdero:
An alternative would be to just change the names and change the `impeller.gni` logic to gracefully disable things when they won't work.

chinmaygarde (approved):

PR #40787, opened 2023-03-30
[Impeller] Fallback to graphics queue is transfer queue is not available

dnfield:
On some devices (for example an Android Pixel 4), a dedicated transfer queue is not available. This should not be considered a fatal error - instead we should use the graphics queue for transfer.

Fixes https://github.com/flutter/flutter/issues/123791


chinmaygarde (approved):
We don't use the transfer queue at all right now. Just FYI. I think we'll end up needing one eventually though.

dnfield:
> We don't use the transfer queue at all right now. Just FYI. I think we'll end up needing one eventually though.

I noticed that but wasn't sure if I was just missing something.

PR #40798, opened 2023-03-30
[web] use callConstructor for FinalizationRegistry due to bug in dart2js

yjbanov:
Use `js_util.callConstructor` to initialize `FinalizationRegistry` to make it compatible with `dart2js`.

Fixes b/276167870


hterkelsen (approved):
LGTM

PR #40801, opened 2023-03-30
[Impeller] move everything needed by the code gen template to core

dnfield:
This makes it so the code_gen_template stuff doesn't have to refer to anything in `renderer`, which Google's internal build system is complaining about even though GN is not (maybe because of the `nogncheck` lines on those includes in the template).

Moves the following from renderer to core:

- `Sampler` and `SamplerDescriptor`
- shader_types.h

Creates a new type in core called `ResourceBinder` which is now implemented by both `ComputeCommand` and `Command`.


impeller/core/resource_binder.cc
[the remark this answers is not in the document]
dnfield: vertex_buffer_builder.cc was meant to be deleted.

But this follows the style of other TUs in this repository (particularyl in impeller)

impeller/core/resource_binder.h
```diff
@@ -0,0 +1,52 @@
...
+#include "impeller/core/sampler.h"
+#include "impeller/core/shader_types.h"
+#include "impeller/core/texture.h"
+
+namespace impeller {
+
+//------------------------------------------------------------------------------
+/// @brief      An interface for binding resources. This is implemented by
+///             |Command| and |ComputeCommand| to make CPU resources available
+///             to the GPU.
```
bdero, line 25: It'd be more accurate to say it makes GPU resources (via handles) available to a given command's pipeline, since you can bind device private resources that never had a host buffer too (like we do with subpass render targets).

impeller/core/resource_binder.h
```diff
@@ -0,0 +1,52 @@
...
+///
+struct ResourceBinder {
+  virtual ~ResourceBinder() = default;
```
bdero, line 28: I'm not strongly opposed to adding this pure virtual interface, but it may cause some confusion/friction down the road for whoever ends up adding the next command type.

We're sort of lucky that the bindings match up between raster and compute commands right now, but if we add support for ray trace commands, the bindings won't perfectly match up anymore, and so we'll probably wind up needing to refactor anything that comes to rely on this interface.

dnfield: I think in that case we'd make a `RayTraceCommand` _not_ implement `ResourceBinder` and instead implement some other type from core.

impeller/core/resource_binder.h
dnfield: Since this is causing friction for the Flutter-at-Google team I'm going to merge this and follow up with a doc fix.

impeller/core/resource_binder.h
dnfield: Alternatively, in that scenario we could make `ResourceBinder` not be pure virtual.

bdero: I'm not sure I understand. What is the purpose of `ResourceBinder`? If it's not being used as a generic interface for commands, what is it useful for?

dnfield: https://github.com/flutter/engine/pull/40812#issuecomment-1491174918

PR #40802, opened 2023-03-30
Re-introduce WM_CLOSE listener, new quit protocol posts a second message to next handler

yaakovschectman:
With this approach, when the last window of an app is closed, the engine sends a message to the framework to check if it is allowed to exit. If it may, the windows lifecycle manager synthesizes a second `WM_CLOSE` message that it will ignore, and so the next registered top level window proc delegate, if any, will be able to process the message. If none do so, the message will be handled by the default window proc, so the app will be able to close.

I was not able to get a full system tray example application running to test this, but I could get an application that stays open when its window is closed and can be seen as a system tray icon as long as it is running, albeit the icon was non-functional. As this repro app still exhibited this behavior when using this engine build, I am reasonably confident in concluding that applications that want to be able to run headless when their windows close will function properly.

Addresses https://github.com/flutter/flutter/issues/123654.

## Pre-launch Checklist

- [ ] I signed the [CLA].


shell/platform/windows/windows_lifecycle_manager.h
```diff
@@ -0,0 +1,40 @@
...
+  virtual void Quit(UINT exit_code, HWND window);
+
+  bool WindowProc(HWND hwnd, UINT msg, WPARAM w, LPARAM l, LRESULT* result);
+
+ private:
+  bool IsLastWindowOfProcess();
+
+  FlutterWindowsEngine* engine_;
+
+  std::map<HWND, int> sent_msgs_;
```
gspencergoog, line 35: Maybe `sent_close_messages_`?

shell/platform/windows/platform_handler.cc
```diff
@@ -205,6 +207,16 @@ int ScopedClipboard::SetString(const std::wstring string) {
 
 }  // namespace
 
+static ExitType StringToExitType(const std::string& string) {
```
gspencergoog, line 210:
```suggestion
static ExitType StringToAppExitType(const std::string& string) {
```

gspencergoog (approved):

cbracken (approved):
Thanks for reworking this!

loic-sharma:
@yaakovschectman Should we ask our partner team if they'd like us to wait until after they ship to merge this? My understanding is that they're shipping their app soon.

shell/platform/windows/windows_lifecycle_manager.h
```diff
@@ -0,0 +1,40 @@
...
+/// A manager for lifecycle events of the top-level window.
+///
+/// Currently handles the following events:
+/// WM_CLOSE
+class WindowsLifecycleManager {
+ public:
+  WindowsLifecycleManager(FlutterWindowsEngine* engine);
+  virtual ~WindowsLifecycleManager();
+
+  virtual void Quit(UINT exit_code, HWND window);
```
loic-sharma, line 26: Nit: `HWND` is generally the first parameter for win32 APIs, consider swapping

loic-sharma, line 26: Could you add a quick comment on `Quit` and `WindowProc`?

shell/platform/windows/windows_lifecycle_manager.cc
```diff
@@ -0,0 +1,134 @@
...
+    : engine_(engine) {}
+
+WindowsLifecycleManager::~WindowsLifecycleManager() {}
+
+void WindowsLifecycleManager::Quit(UINT exit_code, HWND hwnd) {
+  if (hwnd == nullptr) {
+    ::PostQuitMessage(exit_code);
+  } else {
+    sent_close_messages_[hwnd]++;
+    PostMessage(hwnd, WM_CLOSE, 0, 0);
```
loic-sharma, line 26: We should maintain the `wparam` and `lparam` values when we redispatch (you can do evil tricks and shove state in these unused values). Here's how the keyboard does it:

https://github.com/flutter/engine/blob/252f6cfd0335918d4147ba63e52b5800c7254478/shell/platform/windows/keyboard_manager.cc#L118-L120

shell/platform/windows/windows_lifecycle_manager.cc
```diff
@@ -0,0 +1,134 @@
...
+WindowsLifecycleManager::WindowsLifecycleManager(FlutterWindowsEngine* engine)
+    : engine_(engine) {}
+
+WindowsLifecycleManager::~WindowsLifecycleManager() {}
+
+void WindowsLifecycleManager::Quit(UINT exit_code, HWND hwnd) {
+  if (hwnd == nullptr) {
+    ::PostQuitMessage(exit_code);
+  } else {
+    sent_close_messages_[hwnd]++;
```
loic-sharma, line 25: They keyboard logs an error if there are too many pending redispatches messages as that indicates redispatching is broken. Should we do that as well?

shell/platform/windows/platform_handler.h
```diff
@@ -33,6 +40,14 @@ class PlatformHandler {
 
   virtual ~PlatformHandler();
 
+  // String values used for encoding/decoding exit requests.
+  static constexpr char kExitTypeCancelable[] = "cancelable";
+  static constexpr char kExitTypeRequired[] = "required";
+
+  // Send a request to the framework to test if a cancelable exit request
+  // should be canceled or honored.
+  virtual void RequestAppExit(AppExitType exit_type, UINT exit_code, HWND hwnd);
```
loic-sharma, line 49: Consider updating the comment to mention that `HWND` is the window that initiated the app exit.

shell/platform/windows/windows_lifecycle_manager.cc
```diff
@@ -0,0 +1,134 @@
...
+}
+
+bool WindowsLifecycleManager::WindowProc(HWND hwnd,
+                                         UINT msg,
+                                         WPARAM wpar,
+                                         LPARAM lpar,
+                                         LRESULT* result) {
+  switch (msg) {
+    case WM_CLOSE:
+      auto itr = sent_close_messages_.find(hwnd);
```
loic-sharma, line 37: Please add a comment here about why this is necessary. (Framework shutdown is async, we consume the initial `WM_CLOSE` and redispatch it, so need to ignore redispatched `WM_CLOSE`s).

shell/platform/windows/windows_lifecycle_manager.cc
```diff
@@ -0,0 +1,134 @@
...
+          sent_close_messages_.erase(itr);
+        } else {
+          sent_close_messages_[hwnd]--;
+        }
+        return false;
+      }
+      if (IsLastWindowOfProcess()) {
+        engine_->RequestApplicationQuit(AppExitType::cancelable, 0, hwnd);
+      }
+      return true;
```
loic-sharma, line 49: Is this right? Should we leave the `WM_CLOSE` message unhandled if this isn't the last window of the process?

shell/platform/windows/platform_handler.cc
```diff
@@ -354,48 +366,65 @@ void PlatformHandler::SystemSoundPlay(
...
-    int64_t exit_code,
+    AppExitType exit_type,
+    UINT exit_code,
     std::unique_ptr<MethodResult<rapidjson::Document>> result) {
   rapidjson::Document result_doc;
   result_doc.SetObject();
-  if (exit_type.compare(kExitTypeRequired) == 0) {
-    QuitApplication(exit_code);
+  if (exit_type == AppExitType::required) {
+    QuitApplication(exit_code, nullptr);
```
loic-sharma, line 375: Oh wow, I didn't realize `HWND` was optional until very late in the review. This is not at all obvious and will result in crashes if folks miss this. Please make all these `HWND`s `std::optional<HWND>`s

shell/platform/windows/flutter_windows_engine_unittests.cc
```diff
@@ -619,5 +620,142 @@ TEST_F(FlutterWindowsEngineTest, AlertPlatformMessage) {
...
+    finished = exit_code == 0;
+  });
+  EXPECT_CALL(*handler, Quit).Times(1);
+  modifier.SetLifecycleManager(std::move(handler));
+
+  auto binary_messenger =
+      std::make_unique<BinaryMessengerImpl>(engine->messenger());
+  binary_messenger->SetMessageHandler(
+      "flutter/platform", [&did_call](const uint8_t* message,
+                                      size_t message_size, BinaryReply reply) {
```
loic-sharma, line 653: Can we check the message's content?

shell/platform/windows/windows_lifecycle_manager.cc
yaakovschectman: For the keyboard, it should be impossible to get consecutive keyup msgs without a keydown in between, right? But I don't think there's any such guarantee for windows, e.g. if the close button is clicked the same instant as alt+f4 is pressed, so I am not so sure this should be an error state.

shell/platform/windows/flutter_windows_engine_unittests.cc
yaakovschectman: We can, albeit perhaps a bit clunkily. What do you think we should check it for?

shell/platform/windows/flutter_windows_engine_unittests.cc
```diff
@@ -619,5 +620,142 @@ TEST_F(FlutterWindowsEngineTest, AlertPlatformMessage) {
...
+        did_call = true;
+        char response[] = "";
+        reply(reinterpret_cast<uint8_t*>(response), 0);
+      });
+
+  engine->Run();
+
+  // This delegate will be registered after the lifecycle manager, so it will be
+  // called only when it returns false. This should be called on the second,
```
loic-sharma, line 734: `when it returns false` is a bit unclear, consider something like `when the lifecycle manager doesn't handle a message`

shell/platform/windows/windows_lifecycle_manager.cc
loic-sharma: The error is to catch a different scenario: we're redispatching messages but not receiving them. An unbalance would result us in dropping `WM_CLOSE` messages unnecessarily.

shell/platform/windows/flutter_windows_engine_unittests.cc
loic-sharma: The concern would be that we accidentally regress the messages' contents.

Why would it be clunky? Can you not cast it to a `std::string` and then use `EXPECT_EQ`? For example:

https://github.com/flutter/engine/blob/252f6cfd0335918d4147ba63e52b5800c7254478/shell/platform/windows/platform_handler_unittests.cc#L113

https://github.com/flutter/engine/blob/252f6cfd0335918d4147ba63e52b5800c7254478/shell/platform/windows/platform_handler_unittests.cc#L200-L202

shell/platform/windows/windows_lifecycle_manager.cc
yaakovschectman: What would be the condition that you are concerned may arise in this case, that the value held by the map exceeds 1? If so, could we not need to redispatch 2 messages at a time?

shell/platform/windows/flutter_windows_engine_unittests.cc
yaakovschectman: Yes, I was just referring to expecting the json message object to be equal to the string. I think we can know the order, etc., ahead of time, but it might not be obvious, since multiple valid strings would represent the correct message.

yaakovschectman: `The plain [HashMap](https://api.dart.dev/be/180791/dart-collection/HashMap-class.html) is unordered (no order is guaranteed),`
This leads be to believe that the `jsonString` created at https://github.com/flutter/engine/pull/40802/files/ec79b21b9a93211115228da555d0d02a5c227287#diff-0ec6bbe0e6c9b89b68745c51c5e6beda1692354554e97b6c1789bb4614ae8fdfR103 and below does not have a guaranteed order, so testing for equality with a string literal would not be as simple as testing the list example you provided, which is ordered.

loic-sharma: Hm that's a fair concern. We have many tests that assume the order is stable. If Dart ever changes how hash map's order is defined, our tests that use string equality checks on JSON will break and we'll need to fix them.

Regardless, we should verify the message is expected. Feel free to choose whichever approach you prefer :)

cbracken: Dart's default Map class is not a HashMap but a LinkedHashMap, which guarantees order.

shell/platform/windows/windows_lifecycle_manager.cc
loic-sharma: If we redispatch `WM_CLOSE` messages we expect they'll be handled shortly thereafter. It would be unexpected for us to have many redispatched `WM_CLOSE` messages in flight. This may indicate that our redispatch logic is broken. However, I don't feel strongly about asserting for this scenario.

shell/platform/windows/windows_lifecycle_manager.cc
```diff
@@ -0,0 +1,151 @@
...
+}
+
+bool WindowsLifecycleManager::WindowProc(HWND hwnd,
+                                         UINT msg,
+                                         WPARAM wpar,
+                                         LPARAM lpar,
+                                         LRESULT* result) {
+  switch (msg) {
+    // When WM_CLOSE is received from the final window of an application, we
+    // send a request to the framework to see if it is allowed to close. If it
```
loic-sharma, line 48: Consider:

```suggestion
    // send a request to the framework to see if the app should exit. If it
```

shell/platform/windows/windows_lifecycle_manager.h
```diff
@@ -0,0 +1,57 @@
...
+  virtual ~WindowsLifecycleManager();
+
+  // Called when the engine is notified it should quit, e.g. by an application
+  // call to `exitApplication`. When window is std::nullopt, this quits the
+  // application. Otherwise, it holds the HWND of the window that initiated the
+  // request, and exit_code is unused.
+  virtual void Quit(std::optional<HWND> window,
+                    std::optional<WPARAM> wparam,
+                    std::optional<LPARAM> lparam,
+                    UINT exit_code);
```
loic-sharma, line 34: What do you think of having two separate APIs here?

1. Close window
2. Quit app

Apologies for not suggesting this earlier, didn't think of this until now :)

shell/platform/windows/flutter_windows_engine.h
```diff
@@ -254,6 +255,19 @@ class FlutterWindowsEngine {
...
+  // Called when the application quits in response to a quit request.
+  void OnQuit(std::optional<HWND> hwnd,
+              std::optional<WPARAM> wparam,
+              std::optional<LPARAM> lparam,
+              UINT exit_code);
+
+  // Called when a WM_CLOSE message is received.
+  void RequestApplicationQuit(std::optional<HWND> hwnd,
+                              std::optional<WPARAM> wparam,
+                              std::optional<LPARAM> lparam,
```
loic-sharma, line 267: This API is only called when we receive a `WM_CLOSE` right? If so, we will always have `hwnd`, `wparam`, and `lparam` values and these don't need to be `std::optional`. Please let me know if I'm missing something!

shell/platform/windows/windows_lifecycle_manager.cc
```diff
@@ -0,0 +1,151 @@
...
+
+void WindowsLifecycleManager::Quit(std::optional<HWND> hwnd,
+                                   std::optional<WPARAM> wparam,
+                                   std::optional<LPARAM> lparam,
+                                   UINT exit_code) {
+  if (!hwnd.has_value()) {
+    ::PostQuitMessage(exit_code);
+  } else {
+    BASE_CHECK(wparam.has_value() && lparam.has_value());
+    sent_close_messages_[*hwnd]++;
```
loic-sharma, line 29: We should only ignore `WM_CLOSE` messages whose `wparam` and `lparam` matches what we posted.

shell/platform/windows/flutter_windows_engine.h
yaakovschectman: This is also called when `exitApplication` is called on the app side

shell/platform/windows/windows_lifecycle_manager.h
yaakovschectman: We currently intend for the two to both be controlled by the `didRequestAppExit` framework-side callback, which is why they are handled by the same engine-side method as well.

shell/platform/windows/flutter_windows_engine.h
loic-sharma: Gotcha. Please update the comment as it currently indicates it's only for `WM_CLOSE`

cbracken (approved):

PR #40804, opened 2023-03-30
[Impeller] Adds color wheel golden tests.

gaaclarke:
This gets rid of the interactive playground test for a finite set of values we are interested in.  I'm not sure if this captures all the cases we would be interested in capturing but it at least hits all the blend modes.

issue: https://github.com/flutter/flutter/issues/123790

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/impeller/aiks/aiks_playground.cc
+++ b/impeller/aiks/aiks_playground.cc
@@ -44,7 +44,8 @@ bool AiksPlayground::OpenPlaygroundHere(AiksPlaygroundCallback callback) {
 }
 
 bool AiksPlayground::OpenPlaygroundHere(
-    const std::function<void()>& update_imgui,
+    void* state,
+    const std::function<void(void*)>& update_imgui,
     const PictureCallback& callback) {
   if (!switches_.enable_playground) {
     return true;
@@ -57,17 +58,17 @@ bool AiksPlayground::OpenPlaygroundHere(
   }
 
   return Playground::OpenPlaygroundHere(
-      [&renderer, &callback,
-       update_imgui](RenderTarget& render_target) -> bool {
+      [&renderer, &callback, update_imgui,
+       state](RenderTarget& render_target) -> bool {
         static bool wireframe = false;
         if (ImGui::IsKeyPressed(ImGuiKey_Z)) {
           wireframe = !wireframe;
           renderer.GetContentContext().SetWireframe(wireframe);
         }
 
-        update_imgui();
+        update_imgui(state);
 
-        std::optional<Picture> picture = callback(renderer);
+        std::optional<Picture> picture = callback(state, renderer);
         if (!picture.has_value()) {
           return false;
         }
--- a/impeller/aiks/aiks_playground.cc
+++ b/impeller/aiks/aiks_playground.cc
@@ -43,10 +43,10 @@ bool AiksPlayground::OpenPlaygroundHere(AiksPlaygroundCallback callback) {
       });
 }
 
-bool AiksPlayground::OpenPlaygroundHere(
+bool AiksPlayground::OpenPlaygroundHereImpl(
     void* state,
-    const std::function<void(void*)>& update_imgui,
-    const PictureCallback& callback) {
+    const UpdateCallback<void>& update_imgui,
+    const PictureCallback<void>& callback) {
   if (!switches_.enable_playground) {
     return true;
   }
--- a/impeller/aiks/aiks_playground.cc
+++ b/impeller/aiks/aiks_playground.cc
@@ -43,8 +43,9 @@ bool AiksPlayground::OpenPlaygroundHere(AiksPlaygroundCallback callback) {
       });
 }
 
-bool AiksPlayground::OpenPlaygroundHere(std::function<void()> update_imgui,
-                                        PictureCallback callback) {
+bool AiksPlayground::OpenPlaygroundHere(
+    const std::function<void()>& update_imgui,
+    const PictureCallback& callback) {
   if (!switches_.enable_playground) {
     return true;
   }
--- a/impeller/aiks/aiks_playground.cc
+++ b/impeller/aiks/aiks_playground.cc
@@ -43,4 +43,36 @@ bool AiksPlayground::OpenPlaygroundHere(AiksPlaygroundCallback callback) {
       });
 }
 
+bool AiksPlayground::OpenPlaygroundHere(std::function<void()> update_imgui,
+                                        PictureCallback callback) {
+  if (!switches_.enable_playground) {
+    return true;
+  }
+
+  AiksContext renderer(GetContext());
+
+  if (!renderer.IsValid()) {
+    return false;
+  }
+
+  return Playground::OpenPlaygroundHere(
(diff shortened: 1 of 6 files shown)


bdero:
I'd prefer to keep the original playground with ImGui intact so that it can be used to interactive debugging toy. As a starting point for blending goldens, it seems fine to copy/fanout colorwheel as a starting point, and then just allow them to diverge because they are different tools that serve a different purposes.

Also, it would make sense to perhaps label static image tests differently (like start with "Golden/" instead of "Play/") to distinguish between the use cases.

gaaclarke:
> I'd prefer to keep the original playground with ImGui intact so that it can be used to interactive debugging toy. As a starting point for blending goldens, it seems fine to copy/fanout colorwheel, and then just allow them to diverge because they are different tools that serve different purposes.

Okay, I'll bring it back we can have a bigger discussion about where we want to go in the future as a team later.  I don't have a good recommendation yet anyway, I have to think about it.  The one thing I want to make sure is that we don't have diverging tests.  Every interactive test should have a golden companion.  I think I can make that a little easier to happen by reworking `OpenPlaygroundHere` a bit.  Let me give it a shot.

gaaclarke:
@bdero I've created an alternative OpenPlaygroundHere declaration that segregates `imgui` calls.  This is a bit cleaner and the benefit is that the first frame of interactive tests turn into golden images.  I can migrate the rest of the tests in a followup pr if we like this.

gaaclarke:
fyi @dnfield here is a way we can get rid of this skipped tests.  By moving to this new version of the `OpenPlaygroundHere()` the interactive tests will also be golden image tests where the first rendered frame will be the golden.

bdero:
> The one thing I want to make sure is that we don't have diverging tests. 

Up until now we've been sort of abusing gtest for convenience, but some of these interactive playgrounds (like ColorWheel) are debugging tools, not tests. We definitely have some tests in here that are designed to basically be manual goldens, but stuff like `AiksTest.ColorWheel` and `EntityTest.GaussianBlurFilter` aren't.

> Every interactive test should have a golden companion.

Why?

gaaclarke:
> > The one thing I want to make sure is that we don't have diverging tests.
> 
> Up until now we've been sort of abusing gtest for convenience, but some of these interactive playgrounds (like ColorWheel) are debugging tools, not tests. We definitely have some tests in here that are designed to basically be manual goldens, but stuff like `AiksTest.ColorWheel` and `EntityTest.GaussianBlurFilter` aren't.

Right, they are debugging tools where you render something then inspect the output visually.  Golden tests are the same thing, just automated and with the limitation that they aren't animated.  Being able to animate things can useful for debugging so where it's useful we can maintain those.  Check out this latest iteration, your tool still exists.
 
> > Every interactive test should have a golden companion.
> 
> Why?

Because every interactive test represents something we've decided is important which we should automate.  An interactive test without a golden tests would be a gap in our test coverage.

gaaclarke:
I updated the title so it isn't confusing about what is happening.   Nothing is being removed, only added.

bdero:
> > > Every interactive test should have a golden companion.
> > 
> > 
> > Why?
> 
> Because every interactive test represents something we've decided is important which we should automate. An interactive test without a golden tests would be a gap in our test coverage.

Not to say that converting some of these playgrounds is bad -- I think converting simple ones is a good idea and I'm super excited to have goldens for the blends. But I have some concerns about using the complicated interactive ones as goldens, especially if they're refactored into a form where their code is shared.

I'll enumerate a subset of my concerns below, but perhaps we could just make super simple glanceable goldens instead? Like there's no need to draw the full color wheel and retain support for all of the variants that aren't being used. This whole thing can be replaced with 1 golden test with 10 lines of code that does `DrawPaint` to fill in the background, and then does `SaveLayer({.blend_mode=...}); DrawRect(...); Restore()` 29 times for the different blend modes next to each other with a 50% alpha color. This will catch nearly all regressions and probably never flake 🤞. No need to switch on the blur enum either, just count up from zero to `Entity::kLastAdvancedBlendMode`.

1. Playgrounds aren't just used for visual verification of still images, they're used for frame debugging. Some interactive playgrounds have parameters that don't result in a visual difference, but tease `EntityPass` into behaving in certain ways that make debugging easier.
2. The likelihood that these interactive playgrounds will be moved out of the test codebase altogether at some point is very high. We've been talking on and off about building frame inspection tools into all of these interactive playgrounds. We want to have nice things like a quake console that can be used to jump between playgrounds and configure Impeller's internals, or have the ability to open multiple playgrounds in different tabs and dock them side-by-side. gtest and our current playground lifecycle isn't fit host for this purpose.
3. Complicated goldens are more likely to flake, especially those with complex tessellation (curvy surfaces that we triangulate with `libtess`) or complicated transforms applied on the GPU.
4. Looking into a failed golden's code can be a lot quicker an easier if we just keep automated tests straight forward and simple. The interactive playgrounds were not designed to be automated tests, and so they're often very generic (or even kind of eccentric, like in this case) rather than straight forward and simple.

As an aside: Some playgrounds have features constructed for profiling purposes too. For example, we've used `ColorWheel` to compare draw call throughput between different backends (that's why there's a caching flag). If we went and made a performance test suite, I would also be opposed to sharing code between `ColorWheel` and those tests for the same reasons. It's trivial to make simpler, specialized tests that are easier to understand and better fit for automated testing use cases.

bdero:
We talked about this over GVC earlier today, and decided to move forward with not converting ColorWheel for the time being.

There are a bunch of interactive Aiks playgrounds that are great candidates for converting into goldens. These interactive ColorSource-related playgrounds are noisy and repetitive, and if we want to retain their interactivity at some point, we'd be better off consolidating them into 1 or 2 interactive tools.
```
AiksTest.CanRenderLinearGradient
AiksTest.CanRenderLinearGradientWithOverlappingStops
AiksTest.CanRenderLinearGradientManyColors
AiksTest.CanRenderLinearGradientWayManyColors
AiksTest.CanRenderSweepGradient
AiksTest.CanRenderSweepGradientManyColors
AiksTest.CanPictureConvertToImage
AiksTest.CanRenderTiledTexture
```

For these, we can just remove ImGui altogether and avoid the need for a new `OpenPlaygroundHere` overload.

Coming up with good state variations to generate goldens for these requires knowhow about what variations are valuable/non-redundant (a problem which still exists if we try to retain ImGui, of course). A nice approach would be to just land one state variation for each, and then someone can come through later and add more.

gaaclarke:
Here's a PR that is just ripping out imgui for tests: https://github.com/flutter/engine/pull/40835/files

It looks almost like what keeping imgui would have looked like had we migrated them with this pattern.  The difference is the imgui update lambda is removed and there is no struct that represents the data used for the tests (instead the data is just reduced to one variable).  It took more effort only because I took some missteps.  Now I think I nailed the pattern to do it just as quickly.

cc @zanderso

gaaclarke:
Unnecessary after https://github.com/flutter/engine/pull/40879. I'd still like to see something that is a bit easier to parse and verify.  This isn't necessarily it though.  Copying some documentation of the blend modes would be a better idea.

PR #40806, opened 2023-03-30
[Impeller] Declare gradient shader inputs as high precision to avoid overflows.

jason-simmons:


jonahwilliams (approved):
LGTM

jonahwilliams:
You'll need to apply the malioc diff output to the file impeller/tools/malioc.json

jonahwilliams:
going to autosub as I have PRs that also want to update malioc diff

PR #40811, opened 2023-03-30
Revert "[Impeller] move everything needed by the code gen template to core"

dnfield:
Reverts flutter/engine#40801

Broke the build.

PR #40810, opened 2023-03-30
doc fix on ResourceBinder

dnfield:
See https://github.com/flutter/engine/pull/40801#discussion_r1153871844

@bdero

PR #40812, opened 2023-03-31
[Impeller] Reland move more files to core.

dnfield:
Relands https://github.com/flutter/engine/pull/40801
See https://github.com/flutter/engine/pull/40811
Includes changes in https://github.com/flutter/engine/pull/40810

This got reverted due to an unhappy merge missing an include in a modified file. Fixed that up and it should build now.


bdero (approved):

zanderso (commented):
There was an unanswered question from @bdero on the original PR. Was that discussed out-of-band?

impeller/core/resource_binder.cc
```diff
@@ -0,0 +1,11 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "impeller/core/resource_binder.h"
+
+namespace impeller {
+
+//
```
zanderso, line 9: What's this?

dnfield: This is how all the other otherwise header-only TUs are handled in impeller

dnfield:
Due to github magic I missed that question. The answer (I'll add it there too), is that `ResourceBinder`'s real purpose is to make sure that the generated code can depend exclusively on core and not on renderer, because renderer has shaders in it now.

If we add some other command type that's radically different from `ComputeCommand` and `Command`, we can add a different type that's like `ResourceBinder` but more suited to that new (currently hypothetical) command.

PR #40817, opened 2023-03-31
Clean up gn script goma logic a bit

zanderso:


chinmaygarde (approved):

whesse:
There was a big effort in Dart to stop using goma from the HOME dir and always use it from the depot_tools cipd checkout, following a request from the depot_tools team, a few years ago.

None of our CI uses goma from the HOME dir, or installs it there.  I don't think removing support for the depot_tools location is a good idea, especially if support for the HOME location as a default is left in.

This is currently breaking the tip-of-tree builds of Flutter on our benchmarking system, but I think we will fix it by explicitly disabling goma use.

@athomas

tools/gn
```diff
@@ -208,52 +208,59 @@ def get_target_cpu(args):
...
   elif args.goma and os.path.exists(goma_home_dir):
-    gn_args['use_goma'] = True
-    gn_args['goma_dir'] = goma_home_dir
-  elif args.goma and os.path.exists(depot_tools_bin_dir):
-    gn_args['use_goma'] = True
-    gn_args['goma_dir'] = depot_tools_bin_dir
+    goma_gn_args['use_goma'] = True
+    goma_gn_args['goma_dir'] = goma_home_dir
+  elif args.goma:
+    raise Exception(
```
whesse, line 233: Previously, there was a fallback to non-goma builds if GOMA could not be found, rather than a hard failure.
This change to a hard failure may break some CI systems.

whesse:
The specification of the --depot-tools argument to this script says that it is an alternative path that is searched for gomacc.  So that documentation should be changed as well.

whesse:
We have added --no-goma where needed.  There is no problem with this change sticking, if it is intended to remove the fallback to no Goma.  I'm surprised it removes the fallback to depot-tools Goma, because that is what Dart uses, and has removed support for Goma in $HOME, but that may be best if Flutter uses only Fuchsia's Goma with a different server hardcoded.

PR #40822, opened 2023-03-31
Revert "Re-enable LTO and roll buildroot to enable `sINLINING_LIMIT`"

hterkelsen:
Reverts flutter/engine#40808

Regresses benchmarks: https://flutter-flutter-perf.skia.org/e/?numCommits=100&queries=test%3Dweb_benchmarks_canvaskit

### DIFF
--- a/DEPS
+++ b/DEPS
@@ -239,7 +239,7 @@ allowed_hosts = [
 ]
 
 deps = {
-  'src': 'https://github.com/flutter/buildroot.git' + '@' + '37659b25708d805ff77e1a7849b73f2598b2e68d',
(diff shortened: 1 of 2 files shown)


yjbanov (approved):

eyebrowsoffire (changes requested):
We might need to revert the buildroot and move the deps forward to the revert

hterkelsen:
> We might need to revert the buildroot and move the deps forward to the revert

Yes, we should revert the -sINLINING_LIMIT PR in buildroot as well

eyebrowsoffire (approved):
LGTM

PR #40824, opened 2023-03-31
[Impeller] Adds the ability to specify a golden threshold

gaaclarke:
reverts https://github.com/flutter/engine/pull/40818

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/impeller/aiks/aiks_playground.cc
+++ b/impeller/aiks/aiks_playground.cc
@@ -14,8 +14,7 @@ AiksPlayground::AiksPlayground() = default;
 
 AiksPlayground::~AiksPlayground() = default;
 
-bool AiksPlayground::OpenPlaygroundHere(const Picture& picture,
-                                        double max_diff_pixels_percent) {
+bool AiksPlayground::OpenPlaygroundHere(const Picture& picture) {
   return OpenPlaygroundHere(
       [&picture](AiksContext& renderer, RenderTarget& render_target) -> bool {
         return renderer.Render(picture, render_target);
--- a/impeller/aiks/aiks_playground.cc
+++ b/impeller/aiks/aiks_playground.cc
@@ -15,7 +15,7 @@ AiksPlayground::AiksPlayground() = default;
 AiksPlayground::~AiksPlayground() = default;
 
 bool AiksPlayground::OpenPlaygroundHere(const Picture& picture,
-                                        double threshold) {
+                                        double max_diff_pixels_percent) {
   return OpenPlaygroundHere(
       [&picture](AiksContext& renderer, RenderTarget& render_target) -> bool {
         return renderer.Render(picture, render_target);
--- a/impeller/aiks/aiks_playground.cc
+++ b/impeller/aiks/aiks_playground.cc
@@ -14,7 +14,8 @@ AiksPlayground::AiksPlayground() = default;
 
 AiksPlayground::~AiksPlayground() = default;
 
-bool AiksPlayground::OpenPlaygroundHere(const Picture& picture) {
+bool AiksPlayground::OpenPlaygroundHere(const Picture& picture,
+                                        double threshold) {
   return OpenPlaygroundHere(
       [&picture](AiksContext& renderer, RenderTarget& render_target) -> bool {
         return renderer.Render(picture, render_target);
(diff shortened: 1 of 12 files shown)


gaaclarke:
fyi @Piinks @chinmaygarde

chinmaygarde (changes requested):
We don't want thresholds, we want to follow the Skia GPU teams guidelines on having the Gold results be tackled out of band.

gaaclarke:
> We don't want thresholds, we want to follow the Skia GPU teams guidelines on having the Gold results be tackled out of band.

I thought those things are mutually exclusive?  No matter how failures are triaged we don't want this test to be tripping up all the time, right?

chinmaygarde:
> No matter how failures are triaged we don't want this test to be tripping up all the time, right?

The understanding is that each gold test has multiple valid variants. As long as a triager confirms that two variants are sufficiently alike, the test will pass unless an entirely new variant is detected.

gaaclarke:
> The understanding is that each gold test has multiple valid variants. As long as a triager confirms that two variants are sufficiently alike, the test will pass unless an entirely new variant is detected.

Ahh yea, I've heard mention of such a feature.  Who can help use with that?

mdebbar:
The way the fuzzy matcher works in Gold is by doing two checks:

1. How many pixels are different between baseline image and test image? This is controlled by the `differentPixelsRate` parameter that you are using in this PR. So you are good here.

2. How much is each of these pixels allowed to differ from its corresponding pixel in the baseline image? This is controlled by the [`pixelColorDelta`](https://github.com/flutter/engine/blob/ea07ee9ec3318b5214f8786187d91f1d37ea3b1a/testing/skia_gold_client/lib/skia_gold_client.dart#LL194C9-L194C24) parameter. The default is 0 and you are overriding this in your PR.

What's happening is your image is passing the 1st condition but it's failing on the 2nd.

gaaclarke:
@chinmaygarde there was a lot of talk about this in different channels.  I'll try to summarize it:
1) Skia Gold already can work against multiple goldens
1) We want to make failures not block the roll (outlined in https://github.com/flutter/flutter/issues/123979)
1) The number of accepted golden variations is limited to around the number of devices running

I think we still want a tiny bit of fuzzing in addition to the other mechanisms.  This PR makes it so all the tests pass if less than 1% of pixels are different by less than 4 color component deltas.  (I increased the color delta for the rotated text to 40).  PTAL, let me know if you want to discuss this further and we can try to suss everything out.

chinmaygarde (commented):
> I think we still want a tiny bit of fuzzing in addition to the other mechanisms...

I think that is totally reasonable. But I'd rather not have the tests get into the business of determining the deltas like in the current version of the patch.

If there is a global fuzz value you are comfortable with, let's set it in the test harness. If not, we can leave it out altogether. I'd rather not have to think about how to figure out `maxDiffPixelsPercent` and `max_color_delta` for each screenshot. That is cognitive overhead we don't want to add every time we write a test. Especially if the process is going to be semi-manual anyway.

gaaclarke:
> If there is a global fuzz value you are comfortable with, let's set it in the test harness.

Okay, I switched it up so the whole test_runner has the same values across the whole tests.

chinmaygarde (approved):

gaaclarke:
I'm a bit confused about what Skia gold is doing.  The color delta is clearly [set to 8](https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8784801639580552833/+/u/test:_Impeller_golden_Tests_for_host_release/stdout) but the skia gold report has it set to zero:

Screenshot 2023-04-03 at 3 47 54 PM

Let's land this and keep an eye on it.  It may be some latent state or inability to change the fuzzy threshold after the fact, we'll see.

Piinks:
This does not appear to be working as intended. The correct config for fuzzy matching is being received by gold, but it looks like this image is just flaky. @camsim99 came across it in #40924 and looking at the digest, we can see the image has a different color dot for every commit. This means it is not producing a consistent image, and is blocking PRs from landing.

https://flutter-engine-gold.skia.org/search?issue=40924&crs=github&patchsets=3&corpus=flutter-engine

Screenshot 2023-04-05 at 4 48 25 PM

image

PR #40827, opened 2023-03-31
Add ui_web to embedder.yaml so that the analyzer knows about it.

eyebrowsoffire:
We need this for the analyzer to recognize `ui_web`.


yjbanov (approved):
lgtm

PR #40832, opened 2023-03-31
[Impeller] convert filters to half precision

jonahwilliams:

### DIFF
--- a/impeller/entity/shaders/morphology_filter.frag
+++ b/impeller/entity/shaders/morphology_filter.frag
@@ -8,29 +8,35 @@
 
 // These values must correspond to the order of the items in the
 // 'FilterContents::MorphType' enum class.
-const float kMorphTypeDilate = 0;
-const float kMorphTypeErode = 1;
+const float16_t kMorphTypeDilate = 0.0hf;
+const float16_t kMorphTypeErode = 1.0hf;
 
-uniform sampler2D texture_sampler;
+uniform f16sampler2D texture_sampler;
 
 uniform FragInfo {
-  vec2 texture_size;
-  vec2 direction;
-  float radius;
-  float morph_type;
+  f16vec2 uv_offset;
+  float16_t radius;
+  float16_t morph_type;
 }
 frag_info;
 
-in vec2 v_texture_coords;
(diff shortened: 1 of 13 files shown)


impeller/entity/shaders/morphology_filter.frag
```diff
@@ -8,29 +8,35 @@
...
-    vec2 texture_coords = v_texture_coords + uv_offset * i;
-    vec4 color;
-    color = IPSampleDecal(texture_sampler, texture_coords);
+  f16vec4 result =
+      frag_info.morph_type == kMorphTypeDilate ? f16vec4(0.0) : f16vec4(1.0);
+  for (float16_t i = -frag_info.radius; i <= frag_info.radius; i++) {
+    vec2 texture_coords = v_texture_coords + frag_info.uv_offset * i;
+
+// gles 2.0 is the only backend without native decal support.
+#ifdef IMPELLER_TARGET_OPENGLES
```
jonahwilliams, line 34: ifdefing this out to avoid defining a new variant and since vulkan and metal support decal sampling unconditionally

impeller/tools/malioc.json
```diff
@@ -1699,16 +1699,16 @@
...
         "Position": {
-          "fp16_arithmetic": 100,
+          "fp16_arithmetic": 0,
           "has_stack_spilling": false,
           "performance": {
             "longest_path_bound_pipelines": [
               "load_store"
             ],
             "longest_path_cycles": [
-              0.0625,
```
zanderso, line 1709: Here and below, most of these numbers appear to be going up. Is that a concern?

zanderso (approved):
rslgtm w/ question.

impeller/tools/malioc.json
jonahwilliams: These are vertex shaders, so generally its OK as most of our vertex shaders are incredibly fast. We do need to increase precision of input positions as these can overflow more easily.

jonahwilliams: Additionally, these half float conversions only apply to iOS. gles and vulkan have been mostly using relaxed precision for many months

PR #40833, opened 2023-03-31
[web] Fix canvasKitVariant test

mdebbar:
Right now we only check for `v8BreakIterator` and we don't check `ImageDecoder` support. So some tests need to be updated.


eyebrowsoffire (approved):
LGTM

PR #40834, opened 2023-03-31
Fail the run_suite_step if any test failures are detected.

eyebrowsoffire:
We need to throw if the `run_suite_step` encounters any test failures.


lib/web_ui/dev/steps/run_suite_step.dart
```diff
@@ -118,6 +118,7 @@ class RunSuiteStep implements PipelineStep {
     if (io.exitCode != 0) {
       print('[${suite.name.ansiCyan}] ${'Some tests failed.'.ansiRed}');
       io.exitCode = 0;
+      throw ToolExit('Some unit tests failed in suite ${suite.name.ansiCyan}.');
```
yjbanov, line 121: Do we need the previous line? It would be good to have a comment explaining why. It's not obvious why exitCode is set to 0 just before throwing an error, which presumably will result in a non-zero exitCode again.

yjbanov (approved):

lib/web_ui/dev/steps/run_suite_step.dart
eyebrowsoffire: It's basically because we're calling "main" of the test runner, which modifies the exit code. We aggregate all the results and then at the end do our own logic on how to exit the program. I'll add a comment.

PR #40839, opened 2023-03-31
[Impeller] take advantage of native decal sampling, blend cleanups

jonahwilliams:
* In advanced blend shaders, use native decal sampling where available.
* In framebuffer blend shaders always use native decal sampling.
* Adjust blend.frag to do texture flip computation in vertex shader like other samplers, remove tessellation usage in favor of Rect geometry.


zanderso (approved):
lgtm

PR #40846, opened 2023-04-01
Include `ui_web` API files in sky_engine.

eyebrowsoffire:
This is necessary in order for the analyzer to function properly. The sources in sky_engine are basically a superset of what is needed for native and web, so that the analyzer can validate either code path.


yjbanov (approved):
lgtm

PR #40863, opened 2023-04-02
Convert `AccessibilityFeatures` to enum

bernaferrari:
Kind of surprising, but part of https://github.com/flutter/engine/pull/40567.

There is a test there that applies to this, and even though I can duplicate the test and fix for SemanticsAction, if this is used as an enum, it can be an enum.

Right? Let's see if tests pass.


bernaferrari:
This is confusing, theoretically it could be possible, but the class is weird. It is used as an enum, but its parents (EngineAccessibilityFeaturesBuilder) are used "correctly". Yet, `_updateAccessibilityFeatures` still expects the value to be passed in a correct way. My test didn't work, I'll duplicate the method in the other branch.

PR #40871, opened 2023-04-02
[Impeller] reduce gaussian sampling by 2x

jonahwilliams:
Uses technique outlined in https://rastergrid.com/blog/2010/09/efficient-gaussian-blur-with-linear-sampling to reduce number of texture samples by 2x. Rather than sampling each texture element with nearest neighbor, sample a lerp'd weighted value that should be identical to the previous measure.

* Iteration now works from 0 to radius instead of -radius to radius. This allows reuse of the gaussian integral computation since the function is symmetric. Also solves problem where we'd sometimes chop off one end of the iteration (because we started at some value like -9.33 -> 8.33 (because rounding took us up to 9.4 something)

From local testing on an iPhone 13, this is about ~50% faster on the backdropfilter stress test, though no major difference on the animated backdrop filter benchmark as the blur is already fast enough to not cause frame drops.

Most test apps look identical? Malioc results are much worse, but I suspect that it is not really analyzing the number of loops and instead just looking at the content per loop. Because I'm doing more computations per iteration (But fewer iterations) it looks worse.

### DIFF
--- a/impeller/entity/shaders/gaussian_blur/gaussian_blur.glsl
+++ b/impeller/entity/shaders/gaussian_blur/gaussian_blur.glsl
@@ -60,14 +60,28 @@ void main() {
   f16vec4 total_color = f16vec4(0.0hf);
   float16_t gaussian_integral = 0.0hf;
 
-  for (float16_t i = -blur_info.blur_radius; i <= blur_info.blur_radius; i++) {
-    float16_t gaussian = IPGaussian(i, blur_info.blur_sigma);
-    gaussian_integral += gaussian;
-    total_color +=
-        gaussian * Sample(texture_sampler,  // sampler
-                          v_texture_coords + blur_info.blur_uv_offset *
-                                                 i  // texture coordinates
-                   );
+  for (float16_t i = 0.0hf; i <= blur_info.blur_radius; i += 2.0hf) {
+    float16_t w1 = IPGaussian(i, blur_info.blur_sigma);
+    float16_t w2 = IPGaussian(i + 1.0hf, blur_info.blur_sigma);
+    float16_t gaussian = w1 + w2;
+
+    f16vec2 offset_1 = blur_info.blur_uv_offset * i;
+    f16vec2 offset_2 = offset_1 + blur_info.blur_uv_offset;
+    vec2 pos_c1 = v_texture_coords + offset_1;
+    vec2 pos_c2 = v_texture_coords + offset_2;
+    vec2 neg_c1 = v_texture_coords - offset_1;
+    vec2 neg_c2 = v_texture_coords - offset_2;
+
+    vec2 pos_coord = (w1 * pos_c1 + w2 * pos_c2) / gaussian;
+    vec2 neg_coord = (w1 * neg_c1 + w2 * neg_c2) / gaussian;
+
+    gaussian_integral += (gaussian + gaussian);
+    total_color += gaussian * Sample(texture_sampler,  // sampler
+                                     pos_coord         // texture coordinates
+                              );
+    total_color += gaussian * Sample(texture_sampler,  // sampler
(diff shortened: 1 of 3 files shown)


jonahwilliams:
This looks about the same as Impeller on ToT, though both look different from Skia:

Using the macrobenchmark test app

## Android ##

### Impeller ToT

impeller

### Impeller  patch

impeller_patch

### Skia

skia

## iOS ##

### Impeller ToT

impeller_ios

### Impeller  patch

impeller_patch_ios

### Skia

skia_ios

bdero (approved):

impeller/entity/shaders/gaussian_blur/gaussian_blur.glsl
```diff
@@ -60,14 +60,28 @@ void main() {
...
+    vec2 pos_coord = (w1 * pos_c1 + w2 * pos_c2) / gaussian;
+    vec2 neg_coord = (w1 * neg_c1 + w2 * neg_c2) / gaussian;
+
+    gaussian_integral += (gaussian + gaussian);
+    total_color += gaussian * Sample(texture_sampler,  // sampler
+                                     pos_coord         // texture coordinates
+                              );
+    total_color += gaussian * Sample(texture_sampler,  // sampler
+                                     neg_coord         // texture coordinates
+                              );
```
bdero, line 84: Leveraging linear sampling is a great idea. :)

Instead of fanning out the two sides here, I think doing something like `for (float16_t i = round(-blur_info.blur_radius); i <= blur_info.blur_radius; i += 2.hf)` would be the same?

bdero (approved):

impeller/entity/shaders/gaussian_blur/gaussian_blur.glsl
jonahwilliams: should we just round in the contents so blur_radius is always integral?

bdero: Yeah rounding on the CPU seems good

jonahwilliams: Done

jonahwilliams: oops, where did it go...

jonahwilliams: oh nvm, its there

jonahwilliams:
This improved the performance of the backdrop filter blur, good job!

https://flutter-flutter-perf.skia.org/e/?begin=1680558012&end=1680624018&keys=Xdfd0c86066c942f5c30596155b976f36&selected=commit%3D34164%26name%3D%252Carch%253Dintel%252Cbranch%253Dmaster%252Cconfig%253Ddefault%252Cdevice_type%253DiPhone_11%252Cdevice_version%253Dnone%252Chost_type%253Dmac%252Csub_result%253Daverage_frame_rasterizer_time_millis%252Ctest%253Dbackdrop_filter_perf_ios__timeline_summary%252C&xbaroffset=34164

e33

jonahwilliams:
The regular animated blur benchmark also has reduced GPU utilization on this commit:

https://flutter-flutter-perf.skia.org/e/?begin=1680558012&end=1680624018&keys=X10518134e88c8dd3d4ee2f8bbc6784a4&selected=commit%3D34164%26name%3D%252Carch%253Dintel%252Cbranch%253Dmaster%252Cconfig%253Ddefault%252Cdevice_type%253DiPhone_11%252Cdevice_version%253Dnone%252Chost_type%253Dmac%252Csub_result%253Daverage_gpu_usage%252Ctest%253Danimated_blur_backdrop_filter_perf_ios__timeline_summary%252C&xbaroffset=34164

PR #40879, opened 2023-04-03
[Impeller] Add subpass blend goldens

bdero:
Adds a checked macro "loop" for the blend modes in the style of `FOR_EACH_DISPLAY_LIST_OP` + a quick golden for each blend mode that runs the `EntityPass` subpass blending paths.

We should also add Aiks tests to verify the filters in a more comprehensive way (including the blend filter). I can do this in a follow-up. The BlendFilter actually uses different shaders than SaveLayer for the advanced blends on Metal -- the two code paths are wildly different, and we've had serious regressions for both cases.

https://user-images.githubusercontent.com/919017/229463806-c3f20baf-f379-4d14-999b-aab3aec188d7.mov

### DIFF
--- a/impeller/aiks/aiks_unittests.cc
+++ b/impeller/aiks/aiks_unittests.cc
@@ -1287,6 +1287,8 @@ TEST_P(AiksTest, PaintBlendModeIsRespected) {
   ASSERT_TRUE(OpenPlaygroundHere(canvas.EndRecordingAsPicture()));
 }
 
+#define BLEND_MODE_TUPLE(blend_mode) {#blend_mode, BlendMode::k##blend_mode},
+
 TEST_P(AiksTest, ColorWheel) {
   // Compare with https://fiddle.skia.org/c/@BlendModes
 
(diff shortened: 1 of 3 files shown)


impeller/aiks/aiks_unittests.cc
```diff
@@ -1939,5 +1910,22 @@ TEST_P(AiksTest, DrawPaintAbsorbsClears) {
   ASSERT_EQ(picture.pass->GetClearColor(), Color::CornflowerBlue());
 }
 
+Picture BlendModeSaveLayerTest(BlendMode blend_mode) {
```
gaaclarke, line 1913: Add this to an anonymous namespace to avoid collisions.

impeller/geometry/color.cc
```diff
@@ -13,6 +13,28 @@
 
 namespace impeller {
 
+#define _IMPELLER_ASSERT_BLEND_MODE(blend_mode)                            \
+  auto enum_##blend_mode = static_cast<std::underlying_type_t<BlendMode>>( \
+      BlendMode::k##blend_mode);                                           \
+  if (i != enum_##blend_mode) {                                            \
```
gaaclarke, line 19: I'd turn `i` into a parameter so that the macro isn't making assumptions about the context it's used in.

gaaclarke (commented):
The code looks good to me modulo a couple cleanup comments.

Do you think these will be easy to parse these results by human reviewers?  The documentation I saw for the blend modes uses pngs of parrots.  Do you think maybe we should just copy the documentation's example?

Here's some documentation from skia that could be used that is a bit more simple too: https://skia.org/docs/user/api/skblendmode_overview/

The idea being that it would be easy to look at the documentation to see the difference between what it should be and what is being rendered.

impeller/aiks/aiks_unittests.cc
bdero: Done.

impeller/geometry/color.cc
bdero: Hoping to keep this param out of the macro input of `IMPELLER_FOR_EACH_BLEND_MODE` if that's cool, since it shouldn't be needed for actual usage. Open to alternative ideas for verifying the macro matches the enum.

bdero:
> Do you think these will be easy to parse these results by human reviewers?

Yeah I think so. Assuming what's being rendered is currently correct, someone introducing a regression will be able to compare with the previous correct golden.

bdero:
I do like the idea of using the w3c duck. ;) Maybe I can follow-up with that?

Also made a quick issue about improving our blend docs to add inline examples: https://github.com/flutter/flutter/issues/124048

impeller/geometry/color.cc
gaaclarke: It's not a huge deal but here's how you could do it without macros that implicitly capture their context.  We don't actually care about the asserted value, just that they are all accounted for, they all exist and there is no duplicates:

```c++
#define _IMPELLER_ASSERT_BLEND_MODE(blend_mode) \
  auto enum_##blend_mode = static_cast<std::underlying_type_t<BlendMode>>( \
      BlendMode::k##blend_mode);                                           \
  (void)enum_##blend_mode;
  
#define _IMPELLER_PLUS_ONE(blend_mode) + 1

static constexpr inline bool ValidateBlendModes() {
  IMPELLER_FOR_EACH_BLEND_MODE(_IMPELLER_ASSERT_BLEND_MODE);
  int tally = 0  IMPELLER_FOR_EACH_BLEND_MODE(_IMPELLER_PLUS_ONE);
  if (tally - 1 !=
      static_cast<std::underlying_type_t<BlendMode>>(BlendMode::kLast)) {
    return false;
  }
  return true;
}
```

gaaclarke (approved):
lgtm I think we could make the tests easier to parse since looking the goldens I can't easily tell if the are correct.  This will at least make sure we don't change from the current logic accidentally.

impeller/geometry/color.cc
bdero: This is close, but it doesn't quite prevent duplicates. If kSrcOver was accidentally replaced with a second kSrc (thereby dropping kSrcOver), I think this check would pass?

gaaclarke: nope, you'd get a redefinition error when compiling, `enum_SrcOver` would be declared twice.

bdero: ah, nice

PR #40880, opened 2023-04-03
[Impeller] Build fixes for impeller-cmake

bdero:
These are patches in @iamralpht's Aiks fork of impeller-cmake.


iamralpht:
Wow, thanks for putting these up, @bdero!

PR #40886, opened 2023-04-03
[Impeller] reduce advanced blend subpass count for single input with foreground color

jonahwilliams:
Work towards https://github.com/flutter/flutter/issues/124025

This makes Advanced Blends applied via ColorFiltered take one fewer subpass, and Advanced Blends applied directly to contents (via Paint) take no extra subpasses at all.

To test this, I used the flutter codelabs nextgen UI app. Though this uses a ColorFiltered with a pipeline blend, I changed to using ColorDodge to experiment with the costs of the advanced blend filters. While this reduces the number of subpasses significantly, it only has a minor impact on overall performance.

### Before
Screenshot 2023-04-03 at 9 56 15 AM

### After

Screenshot 2023-04-03 at 9 52 07 AM

### DIFF
--- a/impeller/entity/contents/filters/blend_filter_contents.cc
+++ b/impeller/entity/contents/filters/blend_filter_contents.cc
@@ -380,7 +380,7 @@ std::optional<Entity> BlendFilterContents::RenderFilter(
       auto contents = std::make_shared<AdvancedForegroundBlendContents>();
       contents->SetBlendMode(blend_mode_);
       contents->SetCoverage(coverage);
-      contents->SetSrcInput(std::move(inputs[0]));
+      contents->SetSrcInput(inputs[0]);
       contents->SetForegroundColor(foreground_color_.value());
       Entity entity;
       entity.SetTransformation(Matrix::MakeTranslation(coverage.origin));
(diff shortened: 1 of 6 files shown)


impeller/entity/contents/filters/blend_filter_contents.cc
```diff
@@ -373,6 +374,20 @@ std::optional<Entity> BlendFilterContents::RenderFilter(
...
+        potential_alpha >= 1.0 - kEhCloseEnough) {
+      auto contents = std::make_shared<AdvancedForegroundBlendContents>();
+      contents->SetBlendMode(blend_mode_);
+      contents->SetCoverage(coverage);
+      contents->SetSrcInput(inputs[0]);
+      contents->SetForegroundColor(foreground_color_.value());
+      Entity entity;
+      entity.SetTransformation(Matrix::MakeTranslation(coverage.origin));
+      entity.SetContents(std::move(contents));
+      return entity;
```
bdero, line 388: To prevent the need for adding new contents that don't really have a use outside of filters (and for just keeping all the logic local), you could also use an `AnonymousContents` here, which I added to help make returning specialized contents easier when optimizing filters like this.
It just takes a render method and coverage, no questions asked. :)

Up to you if you want to do that refactor though.

bdero (approved):
LGTM!

impeller/entity/contents/filters/blend_filter_contents.cc
jonahwilliams: I do intend to use this a bit more though, for some of the follow ups in https://github.com/flutter/flutter/issues/124025 the bypass the color filter entirely.

PR #40899, opened 2023-04-03
Change back to using the memory address of a texture as its ID.

robert-ancell:
This stops a regression in code that uses the original method.

Once all new code uses fl_texture_get_id, we could re-enable the more predictable incremental texture IDs.

Fixes https://github.com/flutter/flutter/issues/124009


cbracken (approved):
Thanks!

PR #40902, opened 2023-04-03
[Impeller] Respect enable-impeller command line setting over info.plist setting

dnfield:
This way developers get what they ask for regardless of what they ship to their users.

I think we should probably do this for all settings, but will hold off on that and filed https://github.com/flutter/flutter/issues/124049 to track.


zanderso (approved):
LGTM

PR #40903, opened 2023-04-04
Exit early on invalid gn CPU argument combos for simulators

dnfield:
This isn't really foolproof, but would catch some simple mistakes.

I mistakenly tried to build an arm64 simulator build by doing `./flutter/tools/gn --simulator --unopt --mac-cpu=arm64`. It generated the expected out directory (out/ios_debug_sim_unopt_arm64), but did not set the gn arguments for the simulator build and resulted in linker errors down the line.

Also deletes an otherwise unused function.

fixes https://github.com/flutter/flutter/issues/124067

### DIFF
--- a/tools/gn
+++ b/tools/gn
@@ -1036,12 +1036,14 @@ def validate_args(args):
   if args.simulator:
     if args.mac_cpu != 'x64':
       print(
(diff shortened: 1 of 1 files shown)


jmagman (approved):
LGTM

PR #40912, opened 2023-04-04
[Impeller] Snap glyph positions to screen space pixels and map UVs correctly

bdero:
Resolves https://github.com/flutter/flutter/issues/121887.

Hoo boy, this one took longer than I'd like to admit to figure out. Improvements:

1. Properly snap the vertex positions to the screen space pixel grid by keeping track of screen space.
2. Glyph positions are converted to screen space and rounded; glyph sizes are converted to "local" (no translation) screen space and rounded up.
3. UVs are properly mapped to the atlas with a half texel offset so that we sample in the middle of texels.
4. Only use these rounding behaviors when the transform is translation-scale (which is when we nearest sample). In particular, 2 only works for affine matrices.

I think we can come through and remove more CPU cycles here, but wanted to land this since it finally fixes the problem where we skip sampling rows/columns in glyphs.

Before:

https://user-images.githubusercontent.com/919017/229825405-288cddf7-e162-4dca-bc71-6825e024a1ee.mov

After:

https://user-images.githubusercontent.com/919017/229825375-7f242dde-df43-4bc3-a7f5-70625675ed83.mov


jonahwilliams (approved):
LGTM!

chinmaygarde:
This golden issue seems related 
Screenshot 2023-04-04 at 1 34 17 PM

bdero:
Once the text gets far enough away it seems to flip to the other side and start coming inward. I suspect something is overflowing in our matrix utils on the CPU. Looking into it.
image

bdero:
Figured it out. I was doing a vec2 transform instead of a vec4 transform, and so we were resetting the projective coord prior to the final transform to clip space.

PR #40914, opened 2023-04-04
Revert "[Impeller] reduce advanced blend subpass count for single input with foreground color"

jonahwilliams:
Reverts flutter/engine#40886

This renders incorrectly when applied directly to contents instead of as a filter, essentially due to some transorm handling.

The code only works when the snapshot has an identity transform, which is only true as a subpass


jonahwilliams:
Applied to subpass: Correct

flutter_04

Applied directly to contents: Incorrect

flutter_05

zanderso (approved):

jonahwilliams:
I think these checks are stuck for unrelated reasons. Since this is a revert I'm just going to manually merge.

PR #40917, opened 2023-04-04
[Impeller] Fix coverage values when a stroked path is rotated

dnfield:
Fixes https://github.com/flutter/flutter/issues/124131

If the matrix has negative values in it, we were deflating the coverage rect.


bdero (approved):
LGTM!

dnfield:
@gaaclarke - it looks like the playground test draws against a black background, but the golden test against a white one...

gaaclarke:
> @gaaclarke - it looks like the playground test draws against a black background, but the golden test against a white one...

It's just how transparent is interpreted between the viewers.

PR #40918, opened 2023-04-04
Dl std shared from this

flar:
This is an experiment to use `std::shared_ptr` and `std::enable_shared_from_this` as an alternative to an `sk_sp`-like mechanism as demonstrated in https://github.com/flutter/engine/pull/40764

This PR was created from that other PR by deleting the `dl_shared` source files and replacing the references with `shared_ptr` so the differences between the 2 experiments can be viewed simply by looking at the recent commit on this PR titled "remove dl_shared in favor of std::shared_from_this" (it is likely the last commit, but I may add more commits on top of that as issues are discussed in review)

All functionality is achievable with the `std` mechanism. The caveats discovered while creating this version of the change include:

- Both mechanisms can suffer from holding on to a raw pointer past its lifetime if the developer doesn't match the use of that raw pointer with an owner (either its own shared_ptr, or assuming that a caller will keep the object alive for the extent of a method call).
- `shared_ptr` can create a "multiple controlling owners" issue where improper use can cause the object to be managed by unrelated `shared_ptr` instances
  - See https://godbolt.org/z/sarrjKG89 for an example based on an accidental mis-use that occurred while this PR was being created
  - `sk_sp` and `dl_shared` do not have this issue since the object itself manages its own controlling lifetime and the shared wrappers delegate the lifetime management to the object so that they don't even need to know about each other
- The storage cost of `sk_sp` and `dl_shared`:
  - The wrapper is a single pointer
  - The wrapper is simple enough to be passed like an ordinary pointer
  - The object bloat is 1 or 2 ref counts depending on whether a weak count is desired
- The storage cost of `std::shared_ptr`:
  - The wrapper is 2 pointers and one of them points to a control block that holds a pointer for `get()` and a ref count
  - Since the wrapper is non-trivial, it must be passed around indirectly as would be a complex object
  - There is no object bloat without `std::enable_from_this`, but either the control block or a second block used to wrap the pointer to the object itself may contain things like pointers to Delete and Copy operators. If `make_shared` is used, then that extra data is allocated along-side the storage for the object itself as if it were a subclass. If the `shared_ptr()` constructor is used on an allocated object, then the object pointer contains the extra data along with an additional pointer to the object.
  - Inheriting from `shared_from_this` to achieve similar functionality as `sk_sp` and `dl_shared` bloats the object by an additional `std::weak_ptr`


camsim99:
FYI this has no reviewers and some of the tests aren't passing. Should this be a draft?

flar:
This work has been deprioritized in favor of other work. I'll reopen the PR when it becomes more relevant.

PR #40929, opened 2023-04-04
[macOS] Handle termination in FlutterAppDelegate

cbracken:
Moves application termination handling to FlutterAppDelegate. Previously, we required macOS applications using Flutter to ensure their main application class was FlutterApplication. Instead, we now do all handling in FlutterAppDelegate and FlutterEngine.

There are two termination workflows to consider:
* Termination requested from framework side: In this case, then engine receives a `System.exitApplication` method call, and starts the app termination workflow via `[FlutterEngine requestApplicationTermination:exitType:result]`.
* Termination requested from macOS (e.g. Cmd-Q): In this case, `FlutterAppDelegate`'s `applicationShouldTerminate:` handler is invoked by AppKit, and the delegate starts the app termination workflow via `[FlutterEngine requestApplicationTermination:exitType:result]`.

In either case, at this point, if the request is not cancellable, the app immediately exits. If it is cancellable, the embedder sends a `System.requestAppExit` method channel invocation to the framework, which responds with either `exit` or `cancel`. In the case of `exit` we immediately exit, otherwise we do nothing and the app continues running.

This is a minor refactoring of the original approach we took in: https://github.com/flutter/engine/pull/39836

This does not remove the FlutterApplication class, since the framework migration from NSApplication to FlutterApplication still depends on it. A followup patch with replace the migration with a reverse migration will land, then FlutterApplication will be removed.

Issue: https://github.com/flutter/flutter/issues/30735

No new tests since this refactors existing behaviour while retaining the same app semantics as before.

## Pre-launch Checklist

- [X] I signed the [CLA].


shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
```diff
@@ -176,15 +174,11 @@ - (instancetype)initWithEngine:(FlutterEngine*)engine
   _terminator = terminator ? terminator : ^(id sender) {
     // Default to actually terminating the application. The terminator exists to
     // allow tests to override it so that an actual exit doesn't occur.
-    FlutterApplication* flutterApp = [FlutterApplication sharedApplication];
-    if (flutterApp && [flutterApp respondsToSelector:@selector(terminateApplication:)]) {
-      [[FlutterApplication sharedApplication] terminateApplication:sender];
-    } else if (flutterApp) {
-      [flutterApp terminate:sender];
-    }
+    NSApplication* flutterApp = [NSApplication sharedApplication];
```
gspencergoog, line 177: Is there any difference between `[NSApplication sharedApplication]` and `NSApp`?  Is one better than the other style-wise?

gspencergoog (approved):

shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
cbracken: There's no difference; NSApp is just a global variable containing `[NSApplication sharedApplication]`. Not sure if there's ever a style reason to prefer one over the other. @stuartmorgan might know.

PR #40937, opened 2023-04-05
Revert "[web] remove obsolete object caches; simplify native object management"

CaseyHillers:
Reverts flutter/engine#40894

This is breaking Google Testing. See b/277004090

PR #40938, opened 2023-04-05
[web] fix SkFinalizationRegistry for dart2js (attempt 4)

yjbanov:
(this is attempt 4; details below)

Remove obsolete object caches and introduce a simpler way to manage native objects:

* Remove the unused `SynchronousSkiaObjectCache`.
* Introduce new library `native_memory.dart` that's smaller and simpler than `skia_object_cache.dart`.
* Introduce two types of native object references:
  * `UniqueRef` a reference with a unique Dart object owner.
  * `CountedRef` a ref-counted reference with multiple Dart object owners.
* All native references use GC (via `FinalizationRegistry`) as a back-up.
* The new library removes everything related to object resurrection that was needed only in browsers that didn't support `FinalizationRegistry`. All browsers support it now.
* Remove the ad hoc `SkParagraph` cache that predates the introduction of `Paragraph.dispose`.
* Rewrite `CkParagraph` in terms of `UniqueRef`.
* Rewrite `CkImage` in terms of `CountedRef`; delete `SkiaObjectBox`.

This PR does not migrate all objects from the old `skia_object_cache.dart` to `native_memory.dart`. That would be too big of a change. The migration can be done in multiple smaller PRs.

This also removes a few unnecessary relayouts observed in https://github.com/flutter/flutter/issues/120921, but not all of them (more details in https://github.com/flutter/flutter/issues/120921#issuecomment-1481958762)

## About attempt 4

More info about the revert of attempt 3 in https://github.com/flutter/engine/pull/40937.

In this attempt I check that the browser supports `FinalizationRegistry` before registering the object. This will allow the code to run in older browsers, but there will be no protection from memory leaks when the app fails to dispose of the respective objects.

## Benchmarks

Now that this landed in flutter/flutter I have some benchmark numbers from the devicelab. The `text_out_of_picture_bounds` benchmark dropped by 3-4x (lower is better):

Screenshot 2023-04-04 at 6 13 06 PM

The repro provided in https://github.com/flutter/flutter/issues/123204 dropped from 110ms/frame to 10ms/frame.


lib/web_ui/lib/src/engine/canvaskit/native_memory.dart
```diff
@@ -0,0 +1,218 @@
...
+import 'dart:js_interop';
+import 'package:meta/meta.dart';
+
+import '../../engine.dart' show Instrumentation;
+import '../util.dart';
+import 'canvaskit_api.dart';
+
+/// Collects native objects that weren't explicitly disposed of using
+/// [UniqueRef.dispose] or [CountedRef.unref].
+SkObjectFinalizationRegistry _finalizationRegistry = createSkObjectFinalizationRegistry(
```
eyebrowsoffire, line 14: I mentioned this offline, but perhaps it would be a bit less fragile to do the following:
* Make this variable nullable
* Guard the construction of this object by `browserSupportsFinalizationRegistry`, otherwise return null
* Have any uses of this object (i.e. `register` below) do a null check instead of `browserSupportsFinalizationRegistry`
That feels a little less fragile, because right now if someone accidentally touches this variable without checking `browserSupportsFinalizationRegistry` first they could cause things to explode. A nullable variable would let the static type system save you from accidentally doing that.

eyebrowsoffire (changes requested):

eyebrowsoffire (changes requested):
Also I wonder if it would be worthwhile to emit a warning to console if `FinalizationRegistry` is not supported?

ditman:
> Also I wonder if it would be worthwhile to emit a warning to console if FinalizationRegistry is not supported?

I think this is a warning worth adding, especially because FinalizationRegistry is fairly recent on iOS Safari (14.5 according to MDN).

ditman (approved):
RSLGTM, good luck! (Please create an issue with whatever fast-follows you want to do on top of this PR, like Jackson's null-safety suggestion or adding the warning)

eyebrowsoffire (approved):
My comments are not blocking, we can always follow up if needed.

lib/web_ui/lib/src/engine/canvaskit/native_memory.dart
yjbanov: Happy to fix it as a follow-up. For now there's no risk of this variable being touched by accident, because it's private. But in general, I like the idea of explicit lazy initialization.

yjbanov:
> Also I wonder if it would be worthwhile to emit a warning to console if FinalizationRegistry is not supported?

That's a good idea. I'll do it as a follow-up. I realized that we should have similar warnings when `FinalizationRegistry` detects objects that were not disposed of, which I can add as a feature of `UniqueRef` and `CountedRef`,

yjbanov:
Flutter Gold stuff is a know Impeller flakines, so merging manually.

PR #40941, opened 2023-04-05
[fuchsia] Rename params and add comments to FlatlandConnection

uysalere:
This CL renames params and adds comments to FlatlandConnection to clarify behavior.

Bug: fxb/123838

### DIFF
--- a/shell/platform/fuchsia/flutter/flatland_connection.cc
+++ b/shell/platform/fuchsia/flutter/flatland_connection.cc
@@ -56,7 +56,7 @@ void FlatlandConnection::Present() {
   if (present_credits_ > 0) {
     DoPresent();
   } else {
-    present_pending_ = true;
+    present_waiting_for_credit_ = true;
   }
 }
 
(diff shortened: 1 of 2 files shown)


shell/platform/fuchsia/flutter/flatland_connection.cc
```diff
@@ -86,26 +86,26 @@ void FlatlandConnection::DoPresent() {
 
 // This method is called from the UI thread.
 void FlatlandConnection::AwaitVsync(FireCallbackCallback callback) {
+  const fml::TimePoint now = fml::TimePoint::Now();
```
arbreng, line 89: Could you add a comment explaining why you moved this here?

Since it's outside of the mutex, we could wait for some time inside of the mutex lock before executing the code below.  At that point the "now" variable wouldnt be "now" anymore - it would represent some time in the past.

Was that your intention here?

arbreng (approved):
LGTM after either explaining the moved "now" variable or moving it back

PR #40946, opened 2023-04-05
Update Dart pub package revision.

lrhn:
Changes needed to land Dart SDK CL.

A small update to Pub needed to allow some `interface` modifiers to be added to the Dart SDK libraries.

Needs to be landed (and rolled to mono-repo) before SDK CL can land.

### DIFF
--- a/DEPS
+++ b/DEPS
@@ -63,7 +63,7 @@ vars = {
   'dart_perfetto_rev': 'b8da07095979310818f0efde2ef3c69ea70d62c5',
   'dart_protobuf_gn_rev': 'f872f05cb0378eef9a7a2609076929f0f35b4141',
   'dart_protobuf_rev': 'b90a4c44d56075eb23ee38ae56ce0febfd7f280f',
-  'dart_pub_rev': '196e89e5716d64b2954021e6ac9bb5a20d975fc6',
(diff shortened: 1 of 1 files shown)


devoncarew:
Here's the diff from dart-lang/pub for posterity: https://github.com/dart-lang/pub/compare/196e89e5716d64b2954021e6ac9bb5a20d975fc6..f0613dc67423976f1c13cc871e6ddf6db17793b0 .

devoncarew (approved):

zanderso (changes requested):
What is the reason for updating this manually? Normally this would be updated in the PR that the autoroller creates when rolling the Dart SDK.

lrhn:
So the process would be to (get the Pub change landed and)
make a Dart CL to change the Pub rev in the Dart DEPS file and land that.
Then get that rolled into Flutter.
Then get that rolled into the internal mono-repo. 
Then I'd be able to land the Dart CL depending on it.

lrhn:
My (very recent) experience is that a change in the Dart SDK which depends on a change in the Flutter SDK, can get rolled into G3 independently of Flutter, break the old version of Flutter in G3, and get summarily reverted.
I'd prefer to avoid the extra cycle of revert and reland.

zanderso:
Closing as presumed stale. Feel free to reopen and rebase if I'm mistaken.

PR #40949, opened 2023-04-05
[Impeller] Sample the gaussian function once per texture sample

bdero:
Follow-up to remove the second gaussian sample, which isn't really improving quality.

For both before and after, I don't think we're sampling from the center of texels, and this is causing textures with high frequency info to "ripple" a bit when slowly increasing the blur radius (another issue might be that our texture downsizing is, again, too aggressive too early). I'm currently working out the best way to fix this and may follow-up with something.

### DIFF
--- a/impeller/entity/shaders/gaussian_blur/gaussian_blur.glsl
+++ b/impeller/entity/shaders/gaussian_blur/gaussian_blur.glsl
@@ -62,21 +62,13 @@ void main() {
 
   for (float16_t i = -blur_info.blur_radius; i <= blur_info.blur_radius;
        i += 2.0hf) {
-    float16_t w1 = IPGaussian(i, blur_info.blur_sigma);
-    float16_t w2 = IPGaussian(i + 1.0hf, blur_info.blur_sigma);
-    float16_t gaussian = w1 + w2;
-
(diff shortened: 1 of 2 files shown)


jonahwilliams (approved):
LGTM.

I have seen the rippling, but thats in both the before and after versions. But my change assumes that we were sampling the from the center before. Should we just back that out entirely?

bdero:
Yeah just like with the text UV mapping, I don't think we were doing it properly to begin with. I tested out before/after your change and it doesn't seem to make the problem any worse than it already was, so I think we should keep it.

jonahwilliams:
SGTM, ship it

jonahwilliams:
Continued drop in GPU utilization from this patch:

https://flutter-flutter-perf.skia.org/e/?begin=1680818463&end=1680882758&keys=Xf6333a58002b9f41439f0cd9572dadd1&selected=commit%3D34245%26name%3D%252Carch%253Dintel%252Cbranch%253Dmaster%252Cconfig%253Ddefault%252Cdevice_type%253DiPhone_11%252Cdevice_version%253Dnone%252Chost_type%253Dmac%252Csub_result%253Daverage_gpu_usage%252Ctest%253Danimated_blur_backdrop_filter_perf_ios__timeline_summary%252C&xbaroffset=34245

PR #40951, opened 2023-04-05
Copy canvaskit files directly into `flutter_web_sdk`

eyebrowsoffire:
Since the flutter tool now always looks inside the `flutter_web_sdk` artifact for canvaskit/skwasm files, we should copy them to their expected location inside the flutter_web_sdk, so that `--local-web-sdk` works correctly when targeting a local build.

### DIFF
--- a/third_party/canvaskit/BUILD.gn
+++ b/third_party/canvaskit/BUILD.gn
@@ -21,9 +21,15 @@ wasm_toolchain("canvaskit") {
   }
 }
 
-group("canvaskit_group") {
+copy("canvaskit_group") {
   visibility = [ "//flutter/web_sdk:*" ]
   public_deps = [ "//third_party/skia/modules/canvaskit(:canvaskit)" ]
+
+  sources = [
+    "$root_out_dir/canvaskit/canvaskit.js",
+    "$root_out_dir/canvaskit/canvaskit.wasm",
+  ]
+  outputs = [ "$root_out_dir/flutter_web_sdk/canvaskit/{{source_file_part}}" ]
 }
 
 # This toolchain is only to be used by canvaskit_chromium_group below.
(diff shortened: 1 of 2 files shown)


hterkelsen (approved):
LGTM

yjbanov (approved):
lgtm

mdebbar (approved):

alexmarkov:
This change broke Dart/Flutter HEAD-HEAD-HEAD bot.

It caused the following GN errors during "gn --full-dart-sdk --no-prebuilt-dart-sdk" step:

```
Generating GN files in: out/host_debug
ERROR Input to target not generated by a dependency.
The file:
  //out/host_debug/skwasm.wasm
is listed as an input or source for the target:
  //flutter/third_party/canvaskit:skwasm_group
but no targets in the build generate that file.

ERROR Input to target not generated by a dependency.
The file:
  //out/host_debug/skwasm.worker.js
is listed as an input or source for the target:
  //flutter/third_party/canvaskit:skwasm_group
but no targets in the build generate that file.

If you have generated inputs, there needs to be a dependency path between the
two targets in addition to just listing the files. For indirect dependencies,
the intermediate ones must be public_deps. data_deps don't count since they're
only runtime dependencies. If you think a dependency chain exists, it might be
because the chain is private. Try "gn path" to analyze.

2 generated input errors found.
```

https://logs.chromium.org/logs/dart/buildbucket/cr-buildbucket/8784588539471655729/+/u/gn_--full-dart-sdk_--no-prebuilt-dart-sdk/stdout

https://ci.chromium.org/p/dart/builders/ci.sandbox/flutter-engine-linux/19512

eyebrowsoffire:
@alexmarkov Fix for that issue is here: https://github.com/flutter/engine/pull/40967

PR #40954, opened 2023-04-05
[Impeller] adjusted the rotated text test to have an opaque background and account for screen scale

gaaclarke:
issue: https://github.com/flutter/flutter/issues/124277

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/impeller/aiks/aiks_unittests.cc
+++ b/impeller/aiks/aiks_unittests.cc
@@ -1210,11 +1210,18 @@ TEST_P(AiksTest, CanRenderTextOutsideBoundaries) {
 
 TEST_P(AiksTest, TextRotated) {
   Canvas canvas;
-  canvas.Transform(Matrix(0.5, -0.3, 0, -0.002,  //
-                          0, 1, 0, 0,            //
-                          0, 0, 0.3, 0,          //
-                          100, 100, 0, 1.3));
+  canvas.Scale(GetContentScale());
 
+  Paint paint;
+  paint.color = Color(0.1, 0.1, 0.1, 1.0);
+  canvas.DrawRect(
+      Rect::MakeLTRB(0, 0, GetWindowSize().width, GetWindowSize().height),
+      paint);
+
+  canvas.Transform(Matrix(0.25, -0.3, 0, -0.002,  //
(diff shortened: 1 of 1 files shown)


bdero (approved):
LGTM!

impeller/aiks/aiks_unittests.cc
```diff
@@ -1210,11 +1210,18 @@ TEST_P(AiksTest, CanRenderTextOutsideBoundaries) {
...
-                          0, 1, 0, 0,            //
-                          0, 0, 0.3, 0,          //
-                          100, 100, 0, 1.3));
+  canvas.Scale(GetContentScale());
 
+  Paint paint;
+  paint.color = Color(0.1, 0.1, 0.1, 1.0);
+  canvas.DrawRect(
+      Rect::MakeLTRB(0, 0, GetWindowSize().width, GetWindowSize().height),
+      paint);
```
bdero, line 1219: nit: `DrawPaint` would also work here.

gaaclarke: Good call.  I guess we should probably add docstrings.  It didn't occur to me that's what it would do, but makes sense.

dnfield:
I'm seeing this flake still. I think we need to have something in gold that accounts for the GPU model.

gaaclarke:
> I'm seeing this flake still. I think we need to have something in gold that accounts for the GPU model.

Is this flaking or getting confused since it is so different than the old goldens?  It seems like there should be a way to reset the test but I couldn't find a way.

gaaclarke:
It's flaking, here is an example: https://flutter-engine-gold.skia.org/search?corpus=flutter-engine&include_ignored=false&left_filter=name%3Dimpeller_Play_AiksTest_TextRotated_Metal&max_rgba=0&min_rgba=0&negative=true&not_at_head=false&positive=true&reference_image_required=false&right_filter=&sort=descending&untriaged=true

gaaclarke:
I've added an ignore rule for that test: https://flutter-engine-gold.skia.org/ignores

I *think* this will allow us to observe all the variations and decide what to do.

PR #40955, opened 2023-04-05
[Impeller] Support for image decode requests when the GPU sync switch is disabled

dnfield:
This feels a bit hacking (the `mutable` texture, doing the upload as soon as `impeller_texture` is requested with no guard, no tests), but it works locally to fix https://github.com/flutter/flutter/issues/124269

I'd like some feedback/sanity check before I go ahead and try to write tests. I'm also going to be unavailable for the rest of today, so if someone wants to move this forward sooner than later I'm fine with someone else taking over.

### DIFF
--- a/impeller/display_list/display_list_image_impeller_raster.cc
+++ b/impeller/display_list/display_list_image_impeller_raster.cc
@@ -1,0 +1,167 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "impeller/display_list/display_list_image_impeller_raster.h"
+
+#include "flutter/fml/trace_event.h"
+#include "flutter/impeller/core/allocator.h"
+#include "flutter/impeller/core/texture.h"
+#include "flutter/impeller/renderer/command_buffer.h"
+#include "impeller/aiks/aiks_context.h"
+#include "impeller/entity/contents/filters/filter_contents.h"
+
+namespace impeller {
+
+static std::optional<impeller::PixelFormat> ToPixelFormat(SkColorType type) {
+  switch (type) {
+    case kRGBA_8888_SkColorType:
+      return impeller::PixelFormat::kR8G8B8A8UNormInt;
+    case kRGBA_F16_SkColorType:
+      return impeller::PixelFormat::kR16G16B16A16Float;
+    case kBGR_101010x_XR_SkColorType:
+      return impeller::PixelFormat::kB10G10R10XR;
+    default:
+      return std::nullopt;
+  }
+  return std::nullopt;
+}
+
+sk_sp<DlImageImpellerRaster> DlImageImpellerRaster::Make(
+    std::weak_ptr<Context> context,
+    std::shared_ptr<SkBitmap> bitmap,
+    OwningContext owning_context) {
+  if (!bitmap) {
+    return nullptr;
+  }
+  return sk_sp<DlImageImpellerRaster>(new DlImageImpellerRaster(
+      std::move(context), std::move(bitmap), owning_context));
+}
+
+DlImageImpellerRaster::DlImageImpellerRaster(std::weak_ptr<Context> context,
+                                             std::shared_ptr<SkBitmap> bitmap,
+                                             OwningContext owning_context)
+    : context_(std::move(context)),
+      bitmap_(std::move(bitmap)),
+      owning_context_(owning_context) {}
+
+// static
+std::shared_ptr<Texture> DlImageImpellerRaster::UploadTextureToShared(
+    std::shared_ptr<Context> context,
+    std::shared_ptr<SkBitmap> bitmap,
+    bool create_mips) {
+  TRACE_EVENT0("impeller", __FUNCTION__);
+  if (!context || !bitmap) {
+    return nullptr;
+  }
+  const auto image_info = bitmap->info();
+  const auto pixel_format = ToPixelFormat(image_info.colorType());
+  if (!pixel_format) {
+    FML_DLOG(ERROR) << "Pixel format unsupported by Impeller.";
+    return nullptr;
+  }
+
+  impeller::TextureDescriptor texture_descriptor;
+  texture_descriptor.storage_mode = impeller::StorageMode::kHostVisible;
+  texture_descriptor.format = pixel_format.value();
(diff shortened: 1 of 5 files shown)


impeller/display_list/display_list_image_impeller_raster.cc
```diff
@@ -0,0 +1,167 @@
...
+      return impeller::PixelFormat::kR8G8B8A8UNormInt;
+    case kRGBA_F16_SkColorType:
+      return impeller::PixelFormat::kR16G16B16A16Float;
+    case kBGR_101010x_XR_SkColorType:
+      return impeller::PixelFormat::kB10G10R10XR;
+    default:
+      return std::nullopt;
+  }
+  return std::nullopt;
+}
```
gaaclarke, line 28: This is copied from another file, can we share this logic?

impeller/display_list/display_list_image_impeller_raster.cc
```diff
@@ -0,0 +1,167 @@
...
+      context->GetResourceAllocator()->CreateTexture(texture_descriptor);
+  if (!texture) {
+    FML_DLOG(ERROR) << "Could not create Impeller texture.";
+    return nullptr;
+  }
+
+  auto mapping = std::make_shared<fml::NonOwnedMapping>(
+      reinterpret_cast<const uint8_t*>(bitmap->getAddr(0, 0)),  // data
+      texture_descriptor.GetByteSizeOfBaseMipLevel(),           // size
+      [bitmap](auto, auto) mutable { bitmap.reset(); }          // proc
```
gaaclarke, line 81: nit: bitmap = std::move(bitmap)

lib/ui/painting/image_decoder_impeller.cc
```diff
@@ -180,7 +181,9 @@ std::optional<DecompressResult> ImageDecoderImpeller::DecompressTexture(
 
   auto bitmap = std::make_shared<SkBitmap>();
   bitmap->setInfo(image_info);
-  auto bitmap_allocator = std::make_shared<ImpellerAllocator>(allocator);
+  auto bitmap_allocator = allocator
+                              ? std::make_shared<ImpellerAllocator>(allocator)
+                              : std::shared_ptr<ImpellerAllocator>(nullptr);
```
gaaclarke, line 186:
```suggestion
                              : nullptr;
```

lib/ui/painting/image_decoder_impeller.cc
```diff
@@ -227,7 +233,9 @@ std::optional<DecompressResult> ImageDecoderImpeller::DecompressTexture(
   const auto scaled_image_info = image_info.makeDimensions(target_size);
 
   auto scaled_bitmap = std::make_shared<SkBitmap>();
-  auto scaled_allocator = std::make_shared<ImpellerAllocator>(allocator);
+  auto scaled_allocator = allocator
+                              ? std::make_shared<ImpellerAllocator>(allocator)
+                              : std::shared_ptr<ImpellerAllocator>(nullptr);
```
gaaclarke, line 238:
```suggestion
                              : nullptr;
```

impeller/display_list/display_list_image_impeller_raster.h
```diff
@@ -0,0 +1,67 @@
...
+#include "flutter/fml/macros.h"
+#include "impeller/core/texture.h"
+#include "impeller/renderer/context.h"
+#include "third_party/skia/include/core/SkBitmap.h"
+
+namespace impeller {
+
+class AiksContext;
+
+class DlImageImpellerRaster final : public flutter::DlImage {
```
gaaclarke, line 17: Add class docstring.

lib/ui/painting/image_decoder_impeller.cc
```diff
@@ -424,6 +375,11 @@ void ImageDecoderImpeller::Decode(fml::RefPtr<ImageDescriptor> descriptor,
         auto upload_texture_and_invoke_result = [result, context,
                                                  bitmap_result =
                                                      bitmap_result.value()]() {
+          if (!bitmap_result.device_buffer) {
+            result(impeller::DlImageImpellerRaster::Make(
```
gaaclarke, line 379: This should have a comment explaining this branch.

impeller/display_list/display_list_image_impeller_raster.cc
```diff
@@ -0,0 +1,167 @@
...
+
+// |DlImage|
+std::shared_ptr<impeller::Texture> DlImageImpellerRaster::impeller_texture()
+    const {
+  if (!texture_) {
+    auto context = context_.lock();
+    if (!context) {
+      return nullptr;
+    }
+    texture_ = UploadTextureToShared(context, bitmap_,
```
gaaclarke, line 133: So this is where the magic happens right?  Should this have a syncswitch too?

gaaclarke (changes requested):
This needs a test =)

gaaclarke:
Sorry, I just read your description.

Did we verify that we can't actually create the texture with metal while backgrounded?  Like make both branches of the sync switch upload the texture?  It is possible that certain calls, like making textures, may just work in metal with shared vram / ram.

Is the idea that all the code that would access the texture would already be guarded by the sync_switch higher up in the stack trace?  Can we add an assert to make sure that's the case?

impeller/display_list/display_list_image_impeller_raster.cc
dnfield: Yeah, I meant to make this shared.

dnfield:
> Sorry, I just read your description.
> 
> Did we verify that we can't actually create the texture with metal while backgrounded? Like make both branches of the sync switch upload the texture? It is possible that certain calls, like making textures, may just work in metal with shared vram / ram.
> 
> (I guess even if that works we still need to account for opengl in impeller where it doesn't work.)
> 
> Is the idea that all the code that would access the texture would already be guarded by the sync_switch higher up in the stack trace? Can we add an assert to make sure that's the case?

I'm not really clear on this. We're not allowed to encode commands. A shared texture doesn't require any command encoding from what I can tell. So maybe the shared path is just ok as is?

dnfield:
There's only command buffer creation/usage if mipmapping is created, which I think is fine to skip for this case. I'll try an alternative patch for that, which should be a bit simpler.

dnfield:
I think https://github.com/flutter/engine/pull/40963 is a better appraoch.

PR #40956, opened 2023-04-05
[web] Fix trivial type issue.

joshualitt:


joshualitt:
@eyebrowsoffire ptal, just a trivial fix to the type here.

osa1:
The changed line here causes problems with https://dart-review.googlesource.com/c/sdk/+/292881, right?

Assuming the CL is correct, how did this line work on other web backends before? Shouldn't this line cause runtime type errors in other backends that implement runtime type tests correctly?

joshualitt:
I think what might be happening here is that `_jsElements` is initialized with `<Object>[]`, and JS backends are able to pass `List` by reference to JS. Interestingly, we could probably do something like this on Dart2Wasm as well, because I don't think this `List` is used in JS at all, it's just being stored in JS temporarily. That is for a future CL though.

eyebrowsoffire (approved):

Hixie:
test-exempt: current code doesn't pass tests

PR #40959, opened 2023-04-06
[Impeller] partial repaint for Impeller/iOS.

jonahwilliams:
Implements partial repaint for Impeller.

Fixes https://github.com/flutter/flutter/issues/124526

The new code that manages the damage regions is more or less a copy paste from the existing Skia implementation. Compared to Skia, there are a few differences:

Normally Impeller wants to use the drawable as the resolve texture for the root MSAA pass. Unfortunately this will unconditonally clear that texture. Thus to do a partial repaint, we have to allocate a separate texture to resolve to and then blit into the drawable.

The blit seems to take about 500ns for a full screen on an iPhone 13. That implies that partial repaint is likely not worth doing if the screen is significantly changed. Thus I've added code in compositor_context.cc that computes the percentage of width or height that is part of the dirty rect. Above a threshold of (abitrarily chosen) 70%, we just render as normal. This should mean there is only a very minor hit from performing the diff on screens that are highly changed.

The other special case, is that sometimes we get damage rects that are empty - that is the drawable is already completely up to date with what we want to render. IN that case I shortcircuit all of the impeller code and just present immediately. I previously tried returning without a present but this resulted in Xcode reporting dropped frames. One caveat here is that if you use the XCode frame debugger and attempt to capture a frame where we early present, then it will claim it couldn't capture any command buffers (because we didn't create any).

To facilitate all of this, I added some additonal plumbing so that the impeller surface can get the clip rect from the submit info. Additionally, rather than using a clip rect impeller will translate and then shrink the root surface texture. This reduces memory usage compared to just clippling.


jonahwilliams:
image changes seem unrelated from canvaskit changes?

impeller/renderer/blit_pass.cc
```diff
@@ -64,12 +64,12 @@ bool BlitPass::AddCopy(std::shared_ptr<Texture> source,
     return true;  // Nothing to blit.
   }
 
-  // Clip the destination image.
-  source_region = source_region->Intersection(
-      IRect(-destination_origin, destination->GetSize()));
-  if (!source_region.has_value()) {
-    return true;  // Nothing to blit.
-  }
+  // // Clip the destination image.
```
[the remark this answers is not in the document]
jonahwilliams: This didn't work as expected if the source region isn't the entire size of the source texture. Since I changed the impeller impl to use a translate instead of just a clip we don't hit this case anymore.

shell/common/rasterizer.cc
```diff
@@ -605,6 +605,7 @@ RasterStatus Rasterizer::DrawToSurfaceUnsafe(
     if (damage) {
       submit_info.frame_damage = damage->GetFrameDamage();
       submit_info.buffer_damage = damage->GetBufferDamage();
+      submit_info.clip_rect = damage->GetClipRect();
```
knopp, line 608: `buffer_damage` contains the area of buffer that has been written to, which should be identical with the clip rect. Does this need to be passed separately?

jonahwilliams: I wonder if I got this mixed up due to a separate math problem and assumed they were different. Let me double check that.

jonahwilliams: Yup, got things mixed. Removed clip rect in favor of buffer damage. Thanks!

flow/compositor_context.cc
```diff
@@ -130,6 +129,12 @@ RasterStatus CompositorContext::ScopedFrame::Raster(
           ? frame_damage->ComputeClipRect(layer_tree, !ignore_raster_cache)
           : std::nullopt;
 
+  if (aiks_context_ &&
+      !ShouldPerformPartialRepaint(clip_rect, layer_tree.frame_size())) {
+    clip_rect = std::nullopt;
+    frame_damage->Reset();
```
knopp, line 135: This needs a null check for `frame_damage`, otherwise crashes with `FLTDisablePartialRepaint`.

knopp (changes requested):

flow/compositor_context.h
```diff
@@ -91,15 +91,20 @@ class FrameDamage {
 
   // See Damage::buffer_damage.
   std::optional<SkIRect> GetBufferDamage() {
-    return damage_ ? std::make_optional(damage_->buffer_damage) : std::nullopt;
+    return (damage_ && !ignore_damage_)
+               ? std::make_optional(damage_->buffer_damage)
+               : std::nullopt;
   }
 
+  void Reset() { ignore_damage_ = true; }
```
knopp, line 99: I'm wondering if this could simply be `void Reset() { damage_ = std::nullopt; }`

knopp (approved):
Other than possibly removing `ignore_damage_` in favour of setting the `damage_` field to `nullopt` this looks good to me. Thanks!

flow/compositor_context.h
jonahwilliams: Interestingly if I do this rendering breaks. I wonder if the logic for computing damage requires all frames to have up to date damage calculations? Or if perhaps the std::nullopt signal is incorrect for frame damage and it should instead be treated as if the entire screen were dirty?

knopp: I see. It breaks because in that case `frame_damage` is not properly reported in submit info which breaks tracking of accumulated damage in other two buffers. So the area subsequently repainted in back buffers is smaller than it should be be causing glitches. This is visible in the material progress indicator example when sliding in/out the menu:

https://user-images.githubusercontent.com/96958/234208847-4dd190e9-ede0-44c1-b167-0f370d2929dc.mov

Possible solution would be something like this https://github.com/knopp/engine/commit/30c3d61e960e35b2d3028944bf152991e4a68761 . Except *this doesn't work* because it causes `ShouldPerformPartialRepaint` always return `false` in subsequent repaints effectively disabling partial repaint completely.

So I'd keep it as it is, maybe add a comment that we need to preserve `frame_damage` to correctly track accumulated damage for other buffers while at same time reset `buffer_damage` to disable partial repaint for current buffer.

jonahwilliams: Makes sense, thanks for the investigation here.  I added a comment, but I think it makes sense to keep frame damage because the damage info is still correct a subsequent frame could use this and decide to do a partial repaint.

impeller/renderer/backend/metal/surface_mtl.mm
```diff
@@ -143,6 +193,27 @@
...
   [command_buffer waitUntilScheduled];
+
+  // If there is no partial repaint, then immediately present without a blit
+  // pass.
+  if (!requires_blit_) {
+    [drawable_ present];
+    return true;
+  }
+
+  auto blit_command_buffer = context->CreateCommandBuffer();
```
jonahwilliams, line 204: @bdero , if we are doing a final blit for partial repaint, I think I could just hop on the command buffer above and use that? I don't think it actually needs to be empty right, its just a way to ensure that everything else has been scheduled

jonahwilliams: I tried this out a bit but ran into some memory corruption issues. not investigating, just moved the blit before the final command buffer creation for now.

flow/compositor_context.cc
```diff
@@ -164,9 +181,37 @@ RasterStatus CompositorContext::ScopedFrame::Raster(
...
+    std::optional<SkRect> clip_rect,
+    bool ignore_raster_cache) {
+  if (canvas() && clip_rect) {
+    canvas()->Translate(-clip_rect->x(), -clip_rect->y());
+  }
+
+  layer_tree.Paint(*this, ignore_raster_cache);
+}
+
+constexpr float kImpellerRepaintRatio = 0.7f;
```
dnfield, line 200: Consider adding a comment about why this number and how to figure out if changing it is helping or hurting things.

impeller/renderer/backend/metal/surface_mtl.mm
```diff
@@ -16,10 +17,10 @@
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wunguarded-availability-new"
 
-std::unique_ptr<SurfaceMTL> SurfaceMTL::WrapCurrentMetalLayerDrawable(
+id<CAMetalDrawable> SurfaceMTL::GetMetalDrawableAndValidate(
     const std::shared_ptr<Context>& context,
     CAMetalLayer* layer) {
-  TRACE_EVENT0("impeller", "SurfaceMTL::WrapCurrentMetalLayerDrawable");
+  TRACE_EVENT0("impeller", "SurfaceMTL::GetMetalDrawableAndValidate");
```
dnfield, line 23: Are both of these events (this and the one int he new method) helpful?

dnfield (approved):

flow/compositor_context.cc
jonahwilliams: Done.

impeller/renderer/backend/metal/surface_mtl.mm
jonahwilliams: Probably not!, I removed one

PR #40963, opened 2023-04-06
[Impeller] Make multi-frame image decodes work even if the GPU switch is disabled

dnfield:
Fixes https://github.com/flutter/flutter/issues/124269

Manually verified this is safe on an iPhone 11.

We're not creating/encoding command buffers in this path so it's ok. AFAICT the test I added would fail if we started doing that because it doesn't provide any real command buffer interfaces. Most of the code here is related to tests.


lib/ui/painting/multi_frame_codec.cc
```diff
@@ -145,16 +145,11 @@ sk_sp<DlImage> MultiFrameCodec::State::GetNextFrameImage(
...
-    sk_sp<DlImage> result;
-    // impeller, transfer to DlImageImpeller
-    gpu_disable_sync_switch->Execute(fml::SyncSwitch::Handlers().SetIfFalse(
-        [&result, &bitmap, &impeller_context_] {
-          result = ImageDecoderImpeller::UploadTextureToShared(
-              impeller_context_, std::make_shared<SkBitmap>(bitmap),
-              /*create_mips=*/false);
-        }));
-
-    return result;
```
gaaclarke, line 157: It would be nice to verify that this is fine for android opengl es as well.  This may only be an issue for iOS opengl es.

gaaclarke: (I think on android we don't even set the switch since it wasn't an issue)

dnfield: Android allows background GPU utilization

dnfield:
Blocked on https://github.com/flutter/flutter/issues/124332

PR #40970, opened 2023-04-06
[Fuchsia] Implement thread priorities.

eieio:
Implement thread priorities using the Fuchsia media profile provider API to enable system / product configuration of host threads.

This change helps to address b/275114261.

### DIFF
--- a/shell/platform/fuchsia/flutter/engine.cc
+++ b/shell/platform/fuchsia/flutter/engine.cc
@@ -5,8 +5,11 @@
 #include "engine.h"
 
 #include <fuchsia/accessibility/semantics/cpp/fidl.h>
+#include <fuchsia/media/cpp/fidl.h>
 #include <fuchsia/ui/scenic/cpp/fidl.h>
 #include <lib/async/cpp/task.h>
+#include <lib/zx/thread.h>
+#include <zircon/rights.h>
 #include <zircon/status.h>
 #include <zircon/types.h>
 #include <memory>
@@ -21,6 +24,7 @@
 #include "flutter/shell/common/rasterizer.h"
 #include "flutter/shell/common/run_configuration.h"
 #include "flutter/shell/common/serialization_callbacks.h"
+#include "flutter/shell/common/thread_host.h"
 #include "third_party/skia/include/core/SkPicture.h"
 #include "third_party/skia/include/core/SkSerialProcs.h"
 #include "third_party/skia/include/ports/SkFontMgr_fuchsia.h"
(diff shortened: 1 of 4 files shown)


eieio:
This PR primarily changes `Engine::CreateThreadHost`, which is tested. However, the changes are only functional on Fuchsia in an environment configured to prioritize the Flutter threads. IIUC, the automated Flutter tests do not provide such an environment. We have adequate coverage in our Fuchsia product tests.

chinmaygarde (commented):
@eieio Can you suggest a reviewer? If not, I can RSLGTM.

studgeek (commented):
+1 based on general approach which we have used in many places for Fuchsia. Needs Flutter-specific review and approval.

studgeek:
> @eieio Can you suggest a reviewer? If not, I can RSLGTM.

LGTM very welcome. This is a pretty urgent change for a release blocking bug. Thanks!

eieio:
> @eieio Can you suggest a reviewer? If not, I can RSLGTM.

@arbreng provided the guidance internally on where this code should live, but he's out until next week.

Given that and the time constraints and LGTM would be appreciated.

shell/platform/fuchsia/flutter/engine.cc
```diff
@@ -21,6 +24,7 @@
 #include "flutter/shell/common/rasterizer.h"
 #include "flutter/shell/common/run_configuration.h"
 #include "flutter/shell/common/serialization_callbacks.h"
+#include "shell/common/thread_host.h"
```
arbreng, line 27: Nit: "flutter/shell/common/thread_host.h"

chinmaygarde (approved):

shell/platform/fuchsia/flutter/engine.cc
```diff
@@ -54,14 +58,122 @@ std::unique_ptr<flutter::PlatformMessage> MakeLocalizationPlatformMessage(
...
+           flutter::ThreadHost::Type::RASTER, name_prefix),
+       fml::Thread::ThreadPriority::RASTER});
+  thread_host_config.SetUIConfig(
+      {flutter::ThreadHost::ThreadHostConfig::MakeThreadName(
+           flutter::ThreadHost::Type::UI, name_prefix),
+       fml::Thread::ThreadPriority::DISPLAY});
+  thread_host_config.SetIOConfig(
+      {flutter::ThreadHost::ThreadHostConfig::MakeThreadName(
+           flutter::ThreadHost::Type::IO, name_prefix),
+       fml::Thread::ThreadPriority::DISPLAY});
```
arbreng, line 174: IO thread should be either ThreadPriority::NORMAL or ThreadPriority::BACKGROUND; it interacts with the disk.

Under the old logic, it was NORMAL (the default).  Probably wise to keep that to avoid unexpected regressions in asset loading performance.

arbreng (approved):

shell/platform/fuchsia/flutter/engine.cc
eieio: Looking at the codebase it looks like this abstraction won't be used for any ephemeral threads. Is that correct?

If so, then using NORMAL makes sense, since we can assign the corresponding role to whatever scheduling parameters make sense for the product requirements. If NORMAL were to be assigned to every new thread by default then we would have a problem, since IO latency matters on some products, and the IO thread would have to compete with every other random thread for CPU time.

shell/platform/fuchsia/flutter/engine.cc
eieio: Done!

shell/platform/fuchsia/flutter/engine.cc
eieio: Changed the NORMAL based on this and offline conversation.

eieio:
@flutter-dashboard: I don't believe you. This change only affects thread priorities and in any tests here they should be a no-op entirely.

eieio:
Updated based on offline conversation about the platform thread being important to input latency. This final change should be good to go!

arbreng (approved):

eieio:
Thanks, everyone!

PR #40973, opened 2023-04-06
[Impeller] convert most filters to utilize entities instead of subpasses

jonahwilliams:
Fixes https://github.com/flutter/flutter/issues/124175

Converts most filter entities to return an anon contents instead of an entity from a snapshot. This avoids potentially large texture fills and should make rendering this filters marginally more efficient.

The general technique I've used here and in https://github.com/flutter/engine/pull/40927 is to entirely ignore the snapshot uvs and instead just use the exact coordinates. This seems to always be correct, where the snapshot uvs can change based on whether we are applying this to an entity pass or to an individual contents.

The shaders needed to be updated to support taking positions separate from texture coordinates as well. 

In a future patch, we could simplify these by making most filters use the same vertex shader.


jonahwilliams:
Border mask blur is a little funky right now, still working on it.

jonahwilliams:
ahh, I think the issue was actually the blend mode? changing to SrcOver makes it render as expected.

jonahwilliams:
I suppose the question is - in general - do I need to change all of the Src blend modes to be SrcOver to correctly render into parents? I think for the filters I was only looking at cases where the filtered contents were already fully opaque, but maybe I'm overthinking it

jonahwilliams:
Will incorporate changes from blend filter PR

jonahwilliams (commented):
Reverted changes to morphology filter, since I forgot it was a 2-stage filter like gaussian blur which needs a slightly different approach

bdero (approved):
LGTM

PR #40975, opened 2023-04-06
[Impeller] Fix issues in path polyline generation.

dnfield:
Bumps up some size assumptions to accomodate larger paths.
Fixes bugs in the polyline shader that were highlighted by
processing larger more complicated paths.

Adds a test for a larger more complex path. Adds a playground for
testing arbitrary SVG-encoded paths.

Also re-organizes some Skia-to-Impeller utilities into their own TU instead of as scattered static methods in various TUs (@flar @gaaclarke fyi)

Next steps for strokes:

- Handle multiple contours
- Handle non-default caps/joins.

However, with this patch I think I have enough to do some more interesting experiments around stroke rendering for real.


flar:
My only suggestion, and this isn't an issue with the fix itself, is that perhaps the isolation of the conversion functions could be submitted as a separate PR so that the actual changes needed to fix this problem are easier to examine in isolation. (And I'll admit that I've been guilty of combining sweeping reorg changes with a bug fix myself.)

dnfield:
@flar I opened #40997 and will look to merge that one first if it LGTY.

bdero (approved):
LGTM

chinmaygarde (approved):

PR #40984, opened 2023-04-06
[codesign] add codesign white list to engine binary

XilaiZhang:
certain engine binaries should be exempted from mac code signing. Embedding a whitelist.txt file inside the engine binary will mark this artifact as exempted from code signing.

context: https://github.com/flutter/flutter/issues/124322


XilaiZhang:
FlutterMacOS.dSYM.zip and FlutterEmbedder.framework.zip are now verified to have whitelist.txt embedded to be exempted from code signing. 

FlutterMacOS.dSYM.zip is verified with local run of generator script.
FlutterEmbedder.framework.zip is verified through the experimental build at [google cloud bucket](https://pantheon.corp.google.com/storage/browser/_details/flutter_archives_v2/flutter_infra_release/flutter/experimental/c522a4579b1788a04f6a2fd7ed5084b4f78df371/darwin-x64/FlutterEmbedder.framework.zip;tab=live_object?authuser=0&project=flutter-infra)

embedder.zip wasn't verified with local ninja build, since I couldn't get rid of the error with `../../flutter/impeller/renderer/backend/metal/allocator_mtl.mm:86:35: error: use of undeclared identifier 'MTLGPUFamilyApple8'` even with gclient sync. But since the cloud build succeeded I think we are good to go.

godofredoc (commented):
Let's use this PR only to add the signing metadata to flutter embedder. I'll implement skipping  code signing the recipes side to avoid having to add whitelist files to all the mac and ios artifacts.

XilaiZhang:
> Let's use this PR only to add the signing metadata to flutter embedder. I'll implement skipping code signing the recipes side to avoid having to add whitelist files to all the mac and ios artifacts.

thanks for explaining! Based on Jenn's answer in https://github.com/flutter/flutter/issues/124322#issuecomment-1499407354 it seems like flutter embedder should be skipped too. Should we just abandon this PR and https://github.com/flutter/cocoon/pull/2594?

godofredoc (commented):
Let's update this PR to only include the embedder metadata.

zanderso (changes requested):
Please use "allowlist" and not "whitelist".

XilaiZhang:
> Please use "allowlist" and not "whitelist".

Sure, converted to draft for now. Once we figure out which files in embedder should be codesigned, will make corresponding changes, change to allowlist, and republish this pr

XilaiZhang:
The skipping logic is now done on the recipe

PR #40997, opened 2023-04-07
Move Skia conversion utilities to own TU

dnfield:
These utilities are useful in tests and it would otherwise be tempting to duplicate the logic of them in other TUs. It also helps make it a little clearer which parts of Impeller are talking to Skia types without DL interfaces right now, for better or worse.

This is split from https://github.com/flutter/engine/pull/40975 and has a test in there.


lib/ui/BUILD.gn
```diff
@@ -193,7 +193,10 @@ source_set("ui") {
       "painting/image_encoding_impeller.h",
     ]
 
-    deps += [ "//flutter/impeller" ]
+    deps += [
+      "//flutter/impeller",
+      "//flutter/impeller/display_list",
```
gaaclarke, line 198: I think we should turn the skia_conversions to its own target instead of making ui have to depend on all of display_list.  Otherwise looks good to me.

gaaclarke (dismissed):

lib/ui/BUILD.gn
dnfield: Done.

flar (approved):
LGTM

flar:
Filed https://github.com/flutter/flutter/issues/124455 that I noticed while looking this over. I'll take care of it as a separate bookkeeping PR when these go in as it has no impact on functionality or performance.

flar:
@gaaclarke it looks like this PR is blocked by your review even though @dnfield addressed the issues.

PR #41001, opened 2023-04-07
[Impeller] playground upload to device private on macOS

jonahwilliams:
Generally this is fine to do regardless of whether or not we're on UMA . However I never finished https://github.com/flutter/flutter/issues/123468 so it will only work on metal.

I don't think we have a way to runtime check if metal, so I added an ifdef for macos. Im not sure if we have vulkan on macos tests on CI, so lets find out.

Fixes https://github.com/flutter/flutter/issues/124428

### DIFF
--- a/impeller/playground/playground.cc
+++ b/impeller/playground/playground.cc
@@ -400,6 +400,54 @@ std::shared_ptr<Texture> CreateTextureForDecompressedImage(
     const std::shared_ptr<Context>& context,
     DecompressedImage& decompressed_image,
     bool enable_mipmapping) {
+  // TODO(https://github.com/flutter/flutter/issues/123468): copying buffers to
+  // textures is not implemented for GLES/Vulkan.
+#if FML_OS_MACOSX
+  impeller::TextureDescriptor texture_descriptor;
+  texture_descriptor.storage_mode = impeller::StorageMode::kDevicePrivate;
+  texture_descriptor.format = PixelFormat::kR8G8B8A8UNormInt;
+  texture_descriptor.size = decompressed_image.GetSize();
+  texture_descriptor.mip_count =
+      enable_mipmapping ? decompressed_image.GetSize().MipCount() : 1u;
+
+  auto dest_texture =
+      context->GetResourceAllocator()->CreateTexture(texture_descriptor);
+  if (!dest_texture) {
+    FML_DLOG(ERROR) << "Could not create Impeller texture.";
+    return nullptr;
+  }
+
+  auto buffer = context->GetResourceAllocator()->CreateBufferWithCopy(
+      *decompressed_image.GetAllocation().get());
+
+  dest_texture->SetLabel(
+      impeller::SPrintF("ui.Image(%p)", dest_texture.get()).c_str());
+
+  auto command_buffer = context->CreateCommandBuffer();
+  if (!command_buffer) {
+    FML_DLOG(ERROR) << "Could not create command buffer for mipmap generation.";
+    return nullptr;
+  }
+  command_buffer->SetLabel("Mipmap Command Buffer");
+
+  auto blit_pass = command_buffer->CreateBlitPass();
+  if (!blit_pass) {
+    FML_DLOG(ERROR) << "Could not create blit pass for mipmap generation.";
+    return nullptr;
+  }
+  blit_pass->SetLabel("Mipmap Blit Pass");
+  blit_pass->AddCopy(buffer->AsBufferView(), dest_texture);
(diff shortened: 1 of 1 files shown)


impeller/playground/playground.cc
```diff
@@ -400,6 +400,54 @@ std::shared_ptr<Texture> CreateTextureForDecompressedImage(
     const std::shared_ptr<Context>& context,
     DecompressedImage& decompressed_image,
     bool enable_mipmapping) {
+  // TODO(https://github.com/flutter/flutter/issues/123468): copying buffers to
+  // textures is not implemented for GLES/Vulkan.
```
gaaclarke, line 404: The comment doesn't seem to align with the `#if`.  The if blocks a platform, the comment talks about backends.

gaaclarke (approved):
LGTM.  Just a note that what you are doing here to copy the texture is probably something that happens elsewhere and/or is worth capturing for easy use elsewhere.

It would be nice to verify that this unblocks https://github.com/flutter/engine/pull/40998.

impeller/playground/playground.cc
jonahwilliams: we don't have an ifdef for metal as far as I know. Technically macOS is only metal, except that we can run vulkan locally so this change breaks that without a modification to this code.

We can also fix this by finishing the blit pass implementation and removing the check altogether.

jonahwilliams: Looks like this did break something, but hard to tell from CI logs.

gaaclarke:
Close, we get a new error now:
```
[INFO:test_timeout_listener.cc(76)] Test timeout of 120 seconds per test case will be enforced.
2023-04-07 16:14:10.433 impeller_unittests[28742:155517] Metal API Validation Enabled
-[MTLDebugBlitCommandEncoder generateMipmapsForTexture:]:1109: failed assertion `Generate Mipmaps For Texture Validation
[tex mipmapLevelCount](1) must be > 1.
'
[ERROR:flutter/fml/backtrace.cc(108)] Caught signal SIGABRT during program execution.
Frame 0: 0x7ff80b0f40cb err
Frame 1: 0x7ff813c2e8a4 MTLGetEnvCase<>()
Frame 2: 0x7ff813c18c05 MTLReportFailure
Frame 3: 0x7ff813c11378 _MTLMessageContextEnd
Frame 4: 0x7ff80b99a281 -[MTLDebugBlitCommandEncoder generateMipmapsForTexture:]
Frame 5: 0x10482b753 impeller::TextureMTL::GenerateMipmap()
Frame 6: 0x104814bee impeller::BlitPassMTL::EncodeCommands()
Frame 7: 0x1048148b0 impeller::BlitPassMTL::EncodeCommands()
Frame 8: 0x103f289e0 impeller::Playground::CreateTextureForMapping()
Frame 9: 0x103f28c84 impeller::Playground::CreateTextureForFixture()
Frame 10: 0x103eebeb6 impeller::testing::EntityTest_FilterCoverageRespectsCropRect_Test::TestBody()
Frame 11: 0x1049cc2ac testing::Test::Run()
Frame 12: 0x1049cd265 testing::TestInfo::Run()
Frame 13: 0x1049cdf75 testing::TestSuite::Run()
Frame 14: 0x1049dcb0d testing::internal::UnitTestImpl::RunAllTests()
Frame 15: 0x1049dc574 testing::UnitTest::Run()
Frame 16: 0x104093f5e main
Frame 17: 0x11094a52e start
```

gaaclarke:
Here's the run: https://ci.chromium.org/ui/p/flutter/builders/try/Mac%20Engine%20Drone/154281/overview

impeller/playground/playground.cc
```diff
@@ -400,6 +400,54 @@ std::shared_ptr<Texture> CreateTextureForDecompressedImage(
...
+  command_buffer->SetLabel("Mipmap Command Buffer");
+
+  auto blit_pass = command_buffer->CreateBlitPass();
+  if (!blit_pass) {
+    FML_DLOG(ERROR) << "Could not create blit pass for mipmap generation.";
+    return nullptr;
+  }
+  blit_pass->SetLabel("Mipmap Blit Pass");
+  blit_pass->AddCopy(buffer->AsBufferView(), dest_texture);
+  if (texture_descriptor.size.MipCount() > 1) {
```
jonahwilliams, line 440: Need to fix this

gaaclarke:
Oops looks like thats a different function making that fixture than `CreateTextureForDecompressedImage`

jonahwilliams:
Had to fix enable_mipmapping. Hopefully we're just using that one function for all playground tests...

PR #41033, opened 2023-04-10
[Impeller] add goldens for filter optimizations.

jonahwilliams:
Adds a series of goldens that apply a filter to a translated and rotated rectangle to verify that the math in https://github.com/flutter/engine/pull/40973 and https://github.com/flutter/engine/pull/40927 are correct


dnfield (approved):
LGTM

PR #41052, opened 2023-04-10
[Impeller] Obtain the subpass texture after calling OnRender

jason-simmons:
The texture may change during rendering due to operations such as EntityPassTarget::Flip


bdero (approved):
Oh yup, that's a bad one. Was this breaking Wondrous on Vulkan? LGTM.

PR #41054, opened 2023-04-10
Improve Wasm Debugging.

eyebrowsoffire:
This PR makes some improvements to the debugging experience for built wasm modules (CanvasKit and Skwasm):
* We now make sure to copy over the .map files to the artifacts directory for the `host.dart.js` files as well as any `.wasm` outputs, if they are produced.
* We serve source files so that devtools can find them
* We also support DWARF instead of sourcemaps if `--dwarf` is passed to felt on the command line.
  * This requires installing a Chrome extension. I have set up a flow for this, as well as a caching scheme so that the develper only need to install the extension once.
This requires some buildroot changes as well, PR for that is here: https://github.com/flutter/buildroot/pull/710

One thing to note is that the sourcemap strategy doesn't currently work for CanvasKit. This is because the CanvasKit target specifies the `--source-map-base` to be the location inside node_modules for npm clients. I will follow up with the Skia team to try to change this behavior so that the sourcemap strategy works. However, DWARF debugging works for both CanvasKit and Skwasm.


lib/web_ui/dev/build.dart
```diff
@@ -46,6 +47,11 @@ class BuildCommand extends Command<bool> with ArgUtils<bool> {
           'output will be located at "out/wasm_debug".\nThis only applies to '
           'the wasm build. The host build is always built in release mode.',
     );
+    argParser.addFlag(
+      'dwarf',
+      help: 'Embed dwarf debugging info into the output wasm modules. This is'
```
yjbanov, line 52: nit: s/dwarf/DWARF/ (but keep the small case flag name)

yjbanov, line 52: nit2: add one space after "is" so it's not glued to "only" on the next line.

lib/web_ui/dev/build.dart
```diff
@@ -62,11 +68,16 @@ class BuildCommand extends Command<bool> with ArgUtils<bool> {
 
   @override
   FutureOr<bool> run() async {
+    final bool embedDwarf = boolArg('dwarf');
+    if (embedDwarf && runtimeMode != RuntimeMode.debug) {
+      throw ToolExit('Embedding dwarf data requires debug runtime mode.');
```
yjbanov, line 73: DWARF

lib/web_ui/dev/test_platform.dart
```diff
@@ -96,6 +96,7 @@ class BrowserPlatform extends PlatformPlugin {
         //  * /test/alarm_clock_test.dart
         //  * /lib/src/engine/alarm_clock.dart
         .add(createStaticHandler(env.environment.webUiRootDir.path))
+        .add(_createSourceHandler())
```
yjbanov, line 99: Both `_createSourceHandler` and the previous handler attempt to handle `.dart` sources. Can't tell which one wins when.

lib/web_ui/dev/test_runner.dart
```diff
@@ -82,8 +82,11 @@ class TestCommand extends Command<bool> with ArgUtils<bool> {
       )
       ..addFlag(
         'debug',
-        help:
-            'Use artifacts from the debug build instead of release.'
+        help: 'Use artifacts from the debug build instead of release.'
+      )
+      ..addFlag(
+        'dwarf',
+        help: 'Debug wasm modules using embedded dwarf data.'
```
yjbanov, line 89: DWARF

tools/gn
```diff
@@ -761,6 +761,13 @@ def parse_args(args):
   parser.add_argument(
       '--wasm', dest='target_os', action='store_const', const='wasm'
   )
+  parser.add_argument(
+      '--wasm-use-dwarf',
+      action='store_true',
+      default=False,
+      help='Embed dwarf debugging info in the output module instead of using '
```
yjbanov, line 768: DWARF

yjbanov (approved):

DEPS
```diff
@@ -239,7 +239,10 @@ allowed_hosts = [
 ]
 
 deps = {
-  'src': 'https://github.com/flutter/buildroot.git' + '@' + 'e18c162be6327beb90fc885e81db8fadceafb94f',
+  # TODO(jacksongardner): This targets a commit that is not on the master branch.
+  # This needs to be removed before this PR lands, but this will allow us to verify
+  # these changes in CI.
+  'src': 'https://github.com/flutter/buildroot.git' + '@' + '5adf0d1951587debc612831a285d765d8603dfbb',
```
mdebbar, line 245: Just a reminder

mdebbar (changes requested):

third_party/canvaskit/icu_bidi/BUILD.gn
```diff
@@ -47,6 +47,7 @@ static_library("icu_bidi") {
     "//third_party/icu/source/common/ustrtrns.cpp",
     "//third_party/icu/source/common/utf_impl.cpp",
     "//third_party/icu/source/common/utrie2.cpp",
+    "//third_party/icu/source/common/utypes.cpp",
```
mdebbar, line 50: Mind explaining what these changes mean? I think I understand adding `utypes.cpp`. It's used in debug mode to show error messages.

What about `static_library` => `source_set`?

eyebrowsoffire: `utypes.cpp` was because of a compile error I was getting when debugging, for the `u_errorName_skia` symbol. Changing from `static_library` to `source_set` was just something that is a holdover from when I was troubleshooting this issue. I think I can change it back.

lib/web_ui/dev/test_platform.dart
eyebrowsoffire: Turns out the first one is subsumed by the new one, so I removed the old one.

eyebrowsoffire: Actually, we still need that static handler, because it serves specific files that the tests use, but it isn't for sources anymore, so I updated the comment.

mdebbar (approved):

PR #41056, opened 2023-04-10
Implement shadows in the skwasm renderer.

eyebrowsoffire:
Add shadow rendering support to the skwasm renderer and write some unit tests.


lib/web_ui/lib/src/engine/skwasm/skwasm_impl/canvas.dart
```diff
@@ -255,12 +255,19 @@ class SkwasmCanvas implements ui.Canvas {
...
   @override
   void drawShadow(
-    ui.Path path,
+    ui.Path uiPath,
     ui.Color color,
     double elevation,
     bool transparentOccluder,
   ) {
-    // TODO(jacksongardner): implement this
+    final SkwasmPath path = uiPath as SkwasmPath;
```
yjbanov, line 263: pro tip: you can do `uiPath as SkwasmPath;` to promote the original `uiPath` variable to `SkwasmPath`. There's not need to rename the variable.

lib/web_ui/skwasm/canvas.cpp
```diff
@@ -7,8 +7,22 @@
...
 #include "wrappers.h"
 
+#include "third_party/skia/include/core/SkPoint3.h"
+#include "third_party/skia/include/utils/SkShadowUtils.h"
+
 using namespace Skwasm;
 
+namespace {
+// These numbers have been chosen empirically to give a result closest to the
+// material spec.
```
yjbanov, line 17: Do we need to keep these constants in sync with the mobile engine? If so, a link to the source of the mobile numbers could be useful.

yjbanov (approved):
lgtm

mdebbar (approved):

PR #41082, opened 2023-04-11
null check added to avoid NPE while calling FlutterView.detachFromFlutterEngine()

bastionkid:
Issue fixed by this PR: https://github.com/flutter/flutter/issues/110138

This PR fixes an issue which is causing app crash when User tries to navigate to a new instance **FlutterFragment** (whose old instance is already present in the fragment backstack) in an Activity (which was restored to saved state after being killed in background due to memory pressure).

Detailed case to reproduce the crash and identify its' root cause:

Setup: Let's say we've an Activity1 which has a bottom nav bar with 3 tabs. Each of this 3 tabs are **FlutterFragment** i.e. Fragment1, Fragment2 & Fragment3 and all of them are using separate **FlutterEngine** but all of them will be cached. e.g. Multiple instances of **Fragment1** will be going to use same cached **FlutterEngine1**.
1. When User opens the app, Fragment1 gets added to fragment backstack
2. Then User navigates to Fragment2 (gets added to backstack as well)
3. Then User navigates to Fragment3 (gets added to backstack as well)
4. Then User puts the app in background. Due to memory pressure OS/platform kills the Activity1 and all 3 FlutterFragments while the app is in background.
5. Then after sometime User tries to bring the app to foreground from the app stack. Since Activity1 was killed by the OS/platform the app process will try to restore the Activity1 in the same state it was before it got killed. This leads to all 3 fragments present in backstack to get instantiated and then the **onAttach()** gets called for all 3, but only Fragment3 gets **onCreateView()** lifecycle event as it was the top most visible Fragment before the FragmentManager saved the state and app went into background. All 3 FlutterFragment goes through following function calls.

FlutterFragment.onAttach() -> FlutterActivityAndFragmentDelegate.onAttach(). There is a one-to-one mapping between **FlutterFragment <-> FlutterActivityAndFragmentDelegate**

FlutterActivityAndFragmentDelegate.onAttach() -> FlutterEngineConnectionRegistry.attachToActivity(). There is a one-to-one mapping between **FlutterEngine <-> FlutterEngineConnectionRegistry**. _**NOTE**: THIS IS VERY IMPORTANT POINT TO KEEP IN MIND TO UNDERSTAND THE ROOT CAAUSE OF THIS CRASH._

Since all the 3 **FlutterFragment** were just instantiated on activity restore **exclusiveActivity** will be null and exclusiveActivity will be assigned the host **FlutterActivityAndFragmentDelegate**.

6. Then FlutterFragment.onCreateView() will be called only for Fragment3 and it will be visible without an issue as its' state gets restored properly.
7. Then if User tries to navigate to Fragment2 via instantiating new instance of it (this means that now there will be two instances of Fragment2 in the backstack), then there will be crash as it'll go through following function calls.

FlutterFragment.onAttach() -> FlutterActivityAndFragmentDelegate.onAttach().

FlutterActivityAndFragmentDelegate.onAttach() -> FlutterEngineConnectionRegistry.attachToActivity().

THIS IS WHERE THE CRASH STARTS. Since this is the second instance of Fragment2 and both instances are going to use the same cached **FlutterEngine** and hence same **FlutterEngineConnectionRegistry**, this time around **exclusiveActivity** will be non-null as it was assigned during step 5. And since exclusiveActivity will be be non null it'll try to detach from the old **ExclusiveAppComponent** via calling **exclusiveActivity.detachFromFlutterEngine()**.

FlutterActivityAndFragmentDelegate.detachFromFlutterEngine() -> FlutterFragment.detachFromFlutterEngine()

FlutterFragment.detachFromFlutterEngine() -> FlutterActivityAndFragmentDelegate.onDestroyView(). This ideally should not be called if the hosts'  **FlutterFragments.onCreateView()** was not called in the first place. Also since the previous author has added **// Redundant calls are ok.** comment, I'm guessing that this is just a fallback cleanup.

THIS IS WHERE THE CRASH HAPPENS. FlutterActivityAndFragmentDelegate.onDestroyView() -> FlutterView.detachFromFlutterEngine(). Since the lifecycle of older instance of this FlutterFragment2 was capped at onAttach(), it's **FlutterView** property will be null and calling FlutterView.detachFromFlutterEngine() will throw NPE.

## Pre-launch Checklist

- [x] I signed the [CLA].


shell/platform/android/io/flutter/embedding/android/FlutterActivityAndFragmentDelegate.java
```diff
@@ -678,8 +678,11 @@ void onDestroyView() {
       flutterView.getViewTreeObserver().removeOnPreDrawListener(activePreDrawListener);
       activePreDrawListener = null;
     }
-    flutterView.detachFromFlutterEngine();
-    flutterView.removeOnFirstFrameRenderedListener(flutterUiDisplayListener);
+
+    if (flutterView != null) {
```
reidbaker, line 682: Can you add a shortened version of your explanation as a comment here? 
Perhaps something like 
```suggestion
    // flutterView can be null in instances where a delegate.onDestroyView is called without onCreateView being called. See https://github.com/flutter/engine/pull/41082 for more detail. 
    if (flutterView != null) {
```

reidbaker:
Thank you for such a well documented and researched patch!

We are normally pretty strict about not accepting pull requests without tests. That said, I am not sure what of our testing infrastructure should be used to recreate the reproduction conditions you lay out. I have an open question out in [hackers-engine on discord](https://discord.com/channels/608014603317936148/608021010377080866/1096469092896080103). 

If we decide it is ok to take this pr without a test we will need approval from Hixie. See https://github.com/flutter/flutter/wiki/Tree-hygiene#tests for more information on our policies.

shell/platform/android/io/flutter/embedding/android/FlutterActivityAndFragmentDelegate.java
bastionkid: Your comment looks concise and pretty good to me so I'll take the liberty to use the same. 🙂

bastionkid:
> Thank you for such a well documented and researched patch!
> 
> We are normally pretty strict about not accepting pull requests without tests. That said, I am not sure what of our testing infrastructure should be used to recreate the reproduction conditions you lay out. I have an open question out in [hackers-engine on discord](https://discord.com/channels/608014603317936148/608021010377080866/1096469092896080103).
> 
> If we decide it is ok to take this pr without a test we will need approval from Hixie. See https://github.com/flutter/flutter/wiki/Tree-hygiene#tests for more information on our policies.

I checked your Discord message and looks like there is one 👍 and not any replies yet. I guess I just need to wait for people to get some free time to look at the PR then right?

reidbaker (approved):

shell/platform/android/io/flutter/embedding/android/FlutterActivityAndFragmentDelegate.java
```diff
@@ -678,8 +678,13 @@ void onDestroyView() {
       flutterView.getViewTreeObserver().removeOnPreDrawListener(activePreDrawListener);
       activePreDrawListener = null;
     }
-    flutterView.detachFromFlutterEngine();
-    flutterView.removeOnFirstFrameRenderedListener(flutterUiDisplayListener);
+
+    // flutterView can be null in instances where a delegate.onDestroyView is called without
+    // onCreateView being called. See https://github.com/flutter/engine/pull/41082 for more detail.
```
Hixie, line 683: nit: in general the code should stand in isolation; there's no guarantee that the issue database will last as long as the code comments. so if there is more useful data to include it would be good to just describe it all here rather than linking to the issue.

Hixie:
is calling onDestroyView without calling onCreateView considered a in-spec?

reidbaker:
The next flutter contributor to approve please add auto-submit.

gmackall (approved):
LGTM. Thanks for the investigation and patch!

dnfield (commented):
@bastionkid - would you consider adding a test, like the ones in https://github.com/flutter/engine/blob/eadb540b8d1b3cac7458c1b7bba8761a96cdb2a5/shell/platform/android/test/io/flutter/embedding/android/FlutterActivityAndFragmentDelegateTest.java for this?

This code should be pretty easily testable if I'm reading it right.

shell/platform/android/io/flutter/embedding/android/FlutterActivityAndFragmentDelegate.java
Hixie: In general if we think more documentation is needed, then I would recommend putting it in a README.md, or, at the extreme, in a wiki page. We shouldn't consider PR descriptions to be canonical documentation sources.

PR #41088, opened 2023-04-11
[Impeller] Absorb root pass DrawPaints for most blend modes

bdero:
We can be a lot more aggressive with absorbing DrawPaints by just cheaply blending them together on the CPU. We're missing some blend modes at the moment but can slot those in later.


bdero:
Found two new bugs:
- We weren't premultiplying the clear color (fixed)
- `Color::BlendColor` is not handling alpha correctly for all of the advanced blends (updated https://github.com/flutter/flutter/issues/124623)

PR #41094, opened 2023-04-11
Determine lifecycle by looking at window focus also

gspencergoog:
## Description

This incorporates additional signal from `Activity.onWindowFocusChanged` to help decide if the application is `resumed` or `inactive`.

When the user pulls down the notification shade or opens the app switcher in iOS, then iOS sends a notification to the application that it no longer has input focus (is no longer "active" in Apple terminology). 

However, Android (at least on a Pixel) doesn't send `onPause` and `onResume` events for these things, as one might expect. Instead, this PR changes things so that we listen to `Activity.onWindowFocusChanged` and see if any of the windows still have focus.

If it doesn't have focus, then the lifecycle switches to `inactive` (even if `onPause` hasn't been called), and if it does have focus (and `onResume` hasn't been called) then we should go to `resumed`.

State changes are determined and deduped in the `LifecycleChannel` class.

Here's the old state table:

| Android State | Flutter state |
| ------------- | ------------- |
| Resumed  | resumed |
| Paused  | inactive |
| Stopped | paused |
| Detached | detached |

Here's the new state table:

| Android State | Window focused | Flutter state |
| ------------- | ------------- | ------------- |
| Resumed  | true  | resumed |
| Resumed  | false  | _inactive_ * |
| Paused  | true  | inactive |
| Paused  | false  | inactive |
| Stopped  | true  | paused |
| Stopped | false  | paused |
| Detached | true  | detached |
| Detached | false  | detached |

* = This is the relevant change in this PR.

("Window focused" means one or more windows managed by Flutter are focused)

The `inactive` state is for when the application is running and visible, but doesn't have the input focus.  An example where this currently happens are when a phone call is in progress on top of the app, or on some OEMs when going into the app switcher (I've tested on Realme and it does that, at least).  With the PR, it will also go into `inactive` when the app has lost input focus, but is still in the Android `onResume` state. This means that on phones that don't pause the app when they go into the app switcher or the notification window shade (Pixel, others), the app will go into `inactive` when it didn't before. If developers weren't doing anything special in the `inactive` state before, then this PR will have no change for them. If they were, they will go into that state more often (but more consistently across OEMs).

## Related Issues

  - Fixes  https://github.com/flutter/flutter/issues/124591

## Tests

  - Added unit tests for handling `onWindowFocusChanged`.

### DIFF
--- a/lib/ui/platform_dispatcher.dart
+++ b/lib/ui/platform_dispatcher.dart
@@ -1642,31 +1642,61 @@ class FrameTiming {
 /// States that an application can be in.
 ///
 /// The values below describe notifications from the operating system.
-/// Applications should not expect to always receive all possible
-/// notifications. For example, if the users pulls out the battery from the
-/// device, no notification will be sent before the application is suddenly
-/// terminated, along with the rest of the operating system.
+/// Applications should not expect to always receive all possible notifications.
+/// For example, if the users pulls out the battery from the device, no
+/// notification will be sent before the application is suddenly terminated,
+/// along with the rest of the operating system.
+///
+/// For historical and name collision reasons, Flutter's application state names
+/// do not correspond one to one with the state names on all platforms. On
+/// Android, for instance, when the OS calls
+/// [`Activity.onPause`](https://developer.android.com/reference/android/app/Activity#onPause()),
+/// Flutter will enter the [inactive] state, but when Android calls
+/// [`Activity.onStop`](https://developer.android.com/reference/android/app/Activity#onStop()),
+/// Flutter enters the [paused] state. See the individual state's documentation
+/// for descriptions of what they mean on each platform.
 ///
 /// See also:
 ///
-///  * [WidgetsBindingObserver], for a mechanism to observe the lifecycle state
-///    from the widgets layer.
+/// * [WidgetsBindingObserver], for a mechanism to observe the lifecycle state
+///   from the widgets layer.
+/// * iOS's [IOKit activity lifecycle](https://developer.apple.com/documentation/uikit/app_and_environment/managing_your_app_s_life_cycle?language=objc) documentation.
+/// * Android's [activity lifecycle](https://developer.android.com/guide/components/activities/activity-lifecycle) documentation.
+/// * macOS's [AppKit activity lifecycle](https://developer.apple.com/documentation/appkit/nsapplicationdelegate?language=objc) documentation.
 enum AppLifecycleState {
-  /// The application is visible and responding to user input.
+  /// The application is visible and responsive to user input.
+  ///
+  /// On Android, this state corresponds to the Flutter host view having focus
+  /// ([`Activity.onWindowFocusChanged`](https://developer.android.com/reference/android/app/Activity#onWindowFocusChanged(boolean))
+  /// was called with true) while in Android's "resumed" state. It is possible
+  /// for the Flutter app to be in the [inactive] state while still being in
+  /// Android's
+  /// ["onResume"](https://developer.android.com/guide/components/activities/activity-lifecycle)
+  /// state if the app has lost focus
+  /// ([`Activity.onWindowFocusChanged`](https://developer.android.com/reference/android/app/Activity#onWindowFocusChanged(boolean))
+  /// was called with false), but hasn't had
+  /// [`Activity.onPause`](https://developer.android.com/reference/android/app/Activity#onPause())
+  /// called on it.
   resumed,
 
   /// The application is in an inactive state and is not receiving user input.
   ///
   /// On iOS, this state corresponds to an app or the Flutter host view running
-  /// in the foreground inactive state. Apps transition to this state when in
-  /// a phone call, responding to a TouchID request, when entering the app
+  /// in the foreground inactive state. Apps transition to this state when in a
+  /// phone call, responding to a TouchID request, when entering the app
   /// switcher or the control center, or when the UIViewController hosting the
   /// Flutter app is transitioning.
   ///
-  /// On Android, this corresponds to an app or the Flutter host view running
-  /// in the foreground inactive state.  Apps transition to this state when
-  /// another activity is focused, such as a split-screen app, a phone call,
-  /// a picture-in-picture app, a system dialog, or another view.
+  /// On Android, this corresponds to an app or the Flutter host view running in
+  /// Android's paused state (i.e.
+  /// [`Activity.onPause`](https://developer.android.com/reference/android/app/Activity#onPause())
+  /// has been called), or in Android's "resumed" state (i.e.
+  /// [`Activity.onResume`](https://developer.android.com/reference/android/app/Activity#onResume())
+  /// has been called) but it has lost window focus. Examples of when apps
+  /// transition to this state include when the app is partially obscured or
+  /// another activity is focused, such as: a split-screen app, a phone call, a
+  /// picture-in-picture app, a system dialog, another view, when the
+  /// notification window shade is down, or the application switcher is visible.
   ///
   /// Apps in this state should assume that they may be [paused] at any time.
   inactive,
--- a/shell/platform/android/io/flutter/embedding/android/FlutterActivityAndFragmentDelegate.java
+++ b/shell/platform/android/io/flutter/embedding/android/FlutterActivityAndFragmentDelegate.java
@@ -581,7 +581,7 @@ import java.util.List;
   void onResume() {
     Log.v(TAG, "onResume()");
     ensureAlive();
-    if (host.shouldDispatchAppLifecycleState()) {
+    if (host.shouldDispatchAppLifecycleState() && flutterEngine != null) {
       flutterEngine.getLifecycleChannel().appIsResumed();
     }
   }
@@ -629,7 +629,7 @@ import java.util.List;
   void onPause() {
     Log.v(TAG, "onPause()");
     ensureAlive();
-    if (host.shouldDispatchAppLifecycleState()) {
+    if (host.shouldDispatchAppLifecycleState() && flutterEngine != null) {
       flutterEngine.getLifecycleChannel().appIsInactive();
     }
   }
@@ -652,7 +652,7 @@ import java.util.List;
     Log.v(TAG, "onStop()");
     ensureAlive();
 
-    if (host.shouldDispatchAppLifecycleState()) {
+    if (host.shouldDispatchAppLifecycleState() && flutterEngine != null) {
       flutterEngine.getLifecycleChannel().appIsPaused();
     }
 
@@ -763,7 +763,7 @@ import java.util.List;
       platformPlugin = null;
     }
 
-    if (host.shouldDispatchAppLifecycleState()) {
+    if (host.shouldDispatchAppLifecycleState() && flutterEngine != null) {
       flutterEngine.getLifecycleChannel().appIsDetached();
     }
 
@@ -895,6 +895,27 @@ import java.util.List;
       flutterEngine.getActivityControlSurface().onUserLeaveHint();
     } else {
       Log.w(TAG, "onUserLeaveHint() invoked before FlutterFragment was attached to an Activity.");
+    }
+  }
+
+  /**
+   * Invoke this from {@code Activity#onWindowFocusChanged()}.
+   *
+   * <p>A {@code Fragment} host must have its containing {@code Activity} forward this call so that
+   * the {@code Fragment} can then invoke this method.
+   */
+  void onWindowFocusChanged(boolean hasFocus) {
+    ensureAlive();
+    Log.v(TAG, "Received onWindowFocusChanged: " + (hasFocus ? "true" : "false"));
+    if (host.shouldDispatchAppLifecycleState() && flutterEngine != null) {
+      // TODO(gspencergoog): Once we have support for multiple windows/views,
+      // this code will need to consult the list of windows/views to determine if
+      // any windows in the app are focused and call the appropriate function.
+      if (hasFocus) {
+        flutterEngine.getLifecycleChannel().aWindowIsFocused();
+      } else {
+        flutterEngine.getLifecycleChannel().noWindowsAreFocused();
+      }
     }
   }
 
--- a/shell/platform/android/io/flutter/embedding/android/FlutterFragment.java
+++ b/shell/platform/android/io/flutter/embedding/android/FlutterFragment.java
@@ -8,13 +8,16 @@ import android.app.Activity;
 import android.content.ComponentCallbacks2;
 import android.content.Context;
 import android.content.Intent;
+import android.os.Build;
 import android.os.Bundle;
 import android.view.LayoutInflater;
 import android.view.View;
 import android.view.ViewGroup;
+import android.view.ViewTreeObserver.OnWindowFocusChangeListener;
 import androidx.activity.OnBackPressedCallback;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
+import androidx.annotation.RequiresApi;
 import androidx.annotation.VisibleForTesting;
 import androidx.fragment.app.Fragment;
 import androidx.fragment.app.FragmentActivity;
@@ -166,6 +169,19 @@ public class FlutterFragment extends Fragment
    */
   protected static final String ARG_SHOULD_AUTOMATICALLY_HANDLE_ON_BACK_PRESSED =
       "should_automatically_handle_on_back_pressed";
+
+  @RequiresApi(18)
+  private final OnWindowFocusChangeListener onWindowFocusChangeListener =
+      Build.VERSION.SDK_INT >= 18
+          ? new OnWindowFocusChangeListener() {
+            @Override
+            public void onWindowFocusChanged(boolean hasFocus) {
+              if (stillAttachedForEvent("onWindowFocusChanged")) {
+                delegate.onWindowFocusChanged(hasFocus);
+              }
+            }
+          }
+          : null;
 
   /**
    * Creates a {@code FlutterFragment} with a default configuration.
@@ -1110,8 +1126,22 @@ public class FlutterFragment extends Fragment
   }
 
   @Override
+  public void onViewCreated(View view, Bundle savedInstanceState) {
+    super.onViewCreated(view, savedInstanceState);
+    if (Build.VERSION.SDK_INT >= 18) {
+      view.getViewTreeObserver().addOnWindowFocusChangeListener(onWindowFocusChangeListener);
+    }
+  }
+
+  @Override
   public void onDestroyView() {
     super.onDestroyView();
+    if (Build.VERSION.SDK_INT >= 18) {
+      // onWindowFocusChangeListener is API 18+ only.
+      requireView()
+          .getViewTreeObserver()
+          .removeOnWindowFocusChangeListener(onWindowFocusChangeListener);
+    }
     if (stillAttachedForEvent("onDestroyView")) {
       delegate.onDestroyView();
     }
--- a/shell/platform/android/io/flutter/embedding/engine/systemchannels/LifecycleChannel.java
+++ b/shell/platform/android/io/flutter/embedding/engine/systemchannels/LifecycleChannel.java
@@ -5,39 +5,106 @@
 package io.flutter.embedding.engine.systemchannels;
 
 import androidx.annotation.NonNull;
+import androidx.annotation.VisibleForTesting;
 import io.flutter.Log;
 import io.flutter.embedding.engine.dart.DartExecutor;
 import io.flutter.plugin.common.BasicMessageChannel;
 import io.flutter.plugin.common.StringCodec;
 
-/** TODO(mattcarroll): fill in javadoc for LifecycleChannel. */
+/**
+ * A {@link BasicMessageChannel} that communicates lifecycle events to the framework.
+ *
+ * <p>The activity listens to the Android lifecycle events, in addition to the focus events for
+ * windows, and this channel combines that information to decide if the application is the inactive,
+ * resumed, paused, or detached state.
+ */
 public class LifecycleChannel {
   private static final String TAG = "LifecycleChannel";
+  private static final String CHANNEL_NAME = "flutter/lifecycle";
 
-  @NonNull public final BasicMessageChannel<String> channel;
+  // These should stay in sync with the AppLifecycleState enum in the framework.
+  private static final String RESUMED = "AppLifecycleState.resumed";
+  private static final String INACTIVE = "AppLifecycleState.inactive";
+  private static final String PAUSED = "AppLifecycleState.paused";
+  private static final String DETACHED = "AppLifecycleState.detached";
+
+  private String lastAndroidState = "";
+  private String lastFlutterState = "";
+  private boolean lastFocus = false;
+
+  @NonNull private final BasicMessageChannel<String> channel;
 
   public LifecycleChannel(@NonNull DartExecutor dartExecutor) {
-    this.channel =
-        new BasicMessageChannel<>(dartExecutor, "flutter/lifecycle", StringCodec.INSTANCE);
+    this(new BasicMessageChannel<String>(dartExecutor, CHANNEL_NAME, StringCodec.INSTANCE));
   }
 
-  public void appIsInactive() {
-    Log.v(TAG, "Sending AppLifecycleState.inactive message.");
-    channel.send("AppLifecycleState.inactive");
+  @VisibleForTesting
+  public LifecycleChannel(@NonNull BasicMessageChannel<String> channel) {
+    this.channel = channel;
+  }
+
+  // Here's the state table this implements:
+  //
+  // | Android State | Window focused | Flutter state |
+  // |---------------|----------------|---------------|
+  // | Resumed       |     true       |    resumed    |
+  // | Resumed       |     false      |    inactive   |
+  // | Paused        |     true       |    inactive   |
+  // | Paused        |     false      |    inactive   |
+  // | Stopped       |     true       |    paused     |
+  // | Stopped       |     false      |    paused     |
+  // | Detached      |     true       |    detached   |
+  // | Detached      |     false      |    detached   |
+  private void sendState(String state, boolean hasFocus) {
+    if (lastAndroidState == state && hasFocus == lastFocus) {
+      // No inputs changed, so Flutter state could not have changed.
+      return;
+    }
+    String newState;
+    if (state == RESUMED) {
+      // Focus is only taken into account when the Android state is "Resumed".
+      // In all other states, focus is ignored, because we can't know what order
+      // Android lifecycle notifications and window focus notifications events
+      // will arrive in, and those states don't send input events anyhow.
+      newState = hasFocus ? RESUMED : INACTIVE;
+    } else {
+      newState = state;
+    }
+    // Keep the last reported values for future updates.
+    lastAndroidState = state;
+    lastFocus = hasFocus;
+    if (newState == lastFlutterState) {
+      // No change in the resulting Flutter state, so don't report anything.
+      return;
+    }
+    Log.v(TAG, "Sending " + newState + " message.");
+    channel.send(newState);
+    lastFlutterState = newState;
+  }
+
+  // Called if at least one window in the app has focus.
+  public void aWindowIsFocused() {
+    sendState(lastAndroidState, true);
+  }
+
+  // Called if no windows in the app have focus.
+  public void noWindowsAreFocused() {
+    sendState(lastAndroidState, false);
   }
 
   public void appIsResumed() {
-    Log.v(TAG, "Sending AppLifecycleState.resumed message.");
-    channel.send("AppLifecycleState.resumed");
+    sendState(RESUMED, lastFocus);
+  }
+
+  public void appIsInactive() {
+    sendState(INACTIVE, lastFocus);
   }
 
   public void appIsPaused() {
-    Log.v(TAG, "Sending AppLifecycleState.paused message.");
-    channel.send("AppLifecycleState.paused");
+    sendState(PAUSED, lastFocus);
   }
 
   public void appIsDetached() {
-    Log.v(TAG, "Sending AppLifecycleState.detached message.");
-    channel.send("AppLifecycleState.detached");
+    sendState(DETACHED, lastFocus);
   }
 }
--- a/shell/platform/android/io/flutter/plugin/common/PluginRegistry.java
+++ b/shell/platform/android/io/flutter/plugin/common/PluginRegistry.java
@@ -311,6 +311,23 @@ public interface PluginRegistry {
 
     /**
      * Adds a callback allowing the plugin to take part in handling incoming calls to {@link
+     * Activity#onWindowFocusChanged(boolean)}.
+     *
+     * <p>This registrar is for Flutter's v1 embedding. To listen for leave hints in the v2
+     * embedding, use {@link
+     * ActivityPluginBinding#addOnWindowFocusChangedListener(PluginRegistry.WindowFocusChangedListener)}.
+     *
+     * <p>For instructions on migrating a plugin from Flutter's v1 Android embedding to v2, visit
+     * http://flutter.dev/go/android-plugin-migration
+     *
+     * @param listener a {@link WindowFocusChangedListener} callback.
+     * @return this {@link Registrar}.
+     */
+    @NonNull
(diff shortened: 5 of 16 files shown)


shell/platform/android/io/flutter/embedding/android/FlutterActivityAndFragmentDelegate.java
```diff
@@ -898,6 +906,28 @@ void onUserLeaveHint() {
...
+    Log.v(TAG, "Received onWindowFocusChanged: " + (hasFocus ? "true" : "false"));
+    if (host.shouldDispatchAppLifecycleState() && flutterEngine != null) {
+      // This may notify in a redundant fashion with onResume and onPause. The
+      // lifecycle channel will deduplicate, but since the Android documentation
+      // says that we shouldn't rely on any particular order between lifecycle
+      // events and onWindowFocusChanged, we notify in both cases.
+      if (hasFocus) {
+        flutterEngine.getLifecycleChannel().appIsResumed();
+      } else {
+        flutterEngine.getLifecycleChannel().appIsInactive();
```
reidbaker, line 926: I think this is really the core of the review. Do we want to overload inactive to mean either paused or no focus? 

My gut based on the bug is yes but I am worried about breaking existing apps assumptions. 

@gnprice Open to your thoughts here.

shell/platform/android/io/flutter/embedding/android/FlutterActivityAndFragmentDelegate.java
```diff
@@ -898,6 +906,28 @@ void onUserLeaveHint() {
...
 
+  /**
+   * Invoke this from {@code Activity#onWindowFocusChanged()}.
+   *
+   * <p>A {@code Fragment} host must have its containing {@code Activity} forward this call so that
+   * the {@code Fragment} can then invoke this method.
+   */
+  void onWindowFocusChanged(boolean hasFocus) {
+    ensureAlive();
+    Log.v(TAG, "Received onWindowFocusChanged: " + (hasFocus ? "true" : "false"));
```
reidbaker, line 917: This was probably useful for debugging but I think the flutter style is to not log unless something is actionable.

reidbaker (commented):
This code looks great but it will need some tests. Specifically tests to make sure that we are forwarding the onWindowFocusChanged events. I pinged another contributor for a second opinion about the potential risks of having a second item set inactive. 


My final thought is about ensuring that android activities forward the onWindowFocusChanged call along to flutter fragments. Do you know if we have the ability to write custom linters to warn the caller if they dont forward events to us?

shell/platform/android/io/flutter/embedding/android/FlutterActivityAndFragmentDelegate.java
gspencergoog: Ok.  I was following the pattern in most of the other "on..." functions in this file to send a verbose log entry when they are called.  Should we remove the other log messages too?

shell/platform/android/io/flutter/embedding/android/FlutterActivityAndFragmentDelegate.java
gnprice: That reasoning seems right to me.

I think it'd be good to amend that "On Android," paragraph in the [AppLifecycleState.inactive] doc a bit with this change, to make it better align with the terminology in the Android docs. I don't know what "the foreground inactive state" means in an Android context; but if it said "is paused, or is resumed but loses window focus", then that'd make it clear how to map it to the Android lifecycle methods.

Perhaps also add on [AppLifecycleState.resumed] a note that that state no longer agrees with the Android SDK's "resumed" concept, something like

> On Android, this state corresponds to an app or the Flutter host view having focus while in an Android "resumed" state.

That wording is perhaps still confusing and can probably be improved. But I think some confusion is inherent once we choose to have the state we call "resumed" not be the same on Android as the state Android calls "resumed", so we should try to be explicit about it.

shell/platform/android/io/flutter/embedding/android/FlutterActivityAndFragmentDelegate.java
```diff
@@ -898,6 +906,28 @@ void onUserLeaveHint() {
...
+    if (host.shouldDispatchAppLifecycleState() && flutterEngine != null) {
+      // This may notify in a redundant fashion with onResume and onPause. The
+      // lifecycle channel will deduplicate, but since the Android documentation
+      // says that we shouldn't rely on any particular order between lifecycle
+      // events and onWindowFocusChanged, we notify in both cases.
+      if (hasFocus) {
+        flutterEngine.getLifecycleChannel().appIsResumed();
```
gnprice, line 924: From that disclaimer in the [`onWindowFocusChanged`](https://developer.android.com/reference/android/app/Activity#onWindowFocusChanged(boolean)) doc about "any particular order", it seems possible that we could get `onWindowFocusChanged(true)` before we get `onResume()`, while we're still in paused state.

If that does happen, it'd probably be best to hold off on `appIsResumed`. That way we preserve an invariant that when we say `AppLifecycleState.resumed` on Android, it always means at least that the app or Flutter host view is in a resumed state.

shell/platform/android/io/flutter/plugin/common/PluginRegistry.java
```diff
@@ -388,6 +405,14 @@ interface UserLeaveHintListener {
     void onUserLeaveHint();
   }
 
+  /**
+   * Delegate interface for handling user leave hints on behalf of the main {@link
```
gnprice, line 409:
```suggestion
   * Delegate interface for handling window focus changes on behalf of the main {@link
```

gnprice (commented):
Seems good to me. Comments below.

> * That the `onWindowFocusChanged` will report false, even when one of the windows in the activity is focused (I'm not sure yet what it does when you switch between one window and another in a multi-window app).

Yeah, I'd expect that if you have two activities visible in multi-window mode and you switch focus from one to the other, then the system would call `onWindowFocusChanged(false)` on the one activity and `onWindowFocusChanged(true)` on the other.

That should be pretty similar to what happens if you switch between activities in the app switcher, which I believe will call `onStop` on one activity and `onResume` et al. on the other. So if that causes confusion in Flutter, then it's probably a confusion that already exists.

shell/platform/android/io/flutter/embedding/android/FlutterActivityAndFragmentDelegate.java
gspencergoog: Yes, it's unfortunate that we have all these overloaded terms that mean different things on different platforms. For instance, that on Android we go into the `paused` state in `onStop`, and into the `inactive` state in `onPause` is particularly confusing. I'm not sure there's a cure for that, other than breaking all the apps (which we've elected not to do).

(More examples in this [design doc](https://docs.google.com/document/d/1FL5Fgk7EzMNOpJL0GZJt-lfWtGt2igHuC2ZznItS2l8/edit?resourcekey=0-IZNfpSVPkMbigQLLXoJcfw#bookmark=id.dtrwzax9ec0t).)

Added more explicit comments to the enum docs. Also, I have another doc update in this [draft PR that adds more app lifecycle support](https://github.com/flutter/engine/pull/40542/files#diff-57d6953e215d0e5dd7260ee60b665c812aa730566ef0b30f7ff1e3d661143585).

gspencergoog:
I removed the state in the `LifecycleChannel` (the `private String lastState` variable), since I think it's too easy to get out of sync with the real state of the Android Activity, or with the state in the framework, since an add-to-app activity could start handling the lifecycle states, or could send messages to the framework to change the state on its own. Sure, not common things, but they could happen.

Sending extra state updates in the case where we move from inactive to resumed or vice versa are unfortunate, but I'd rather have one source of truth for the state of the app and just deduplicate them.

Off to writing tests now.

shell/platform/android/io/flutter/embedding/android/FlutterActivityAndFragmentDelegate.java
gnprice: Great, those added comments look good to me.

> I'm not sure there's a cure for that, other than breaking all the apps (which we've elected not to do).

Yeah. Given the differences between platforms, even if we found the best possible set of names unconstrained by compatibility with existing apps and used those, it'd probably still be fairly confusing. So I think the best we can do is just try to be explicit in the docs about what they do mean in each platform's terms.

gspencergoog:
> Specifically tests to make sure that we are forwarding the onWindowFocusChanged events. I pinged another contributor for a second opinion about the potential risks of having a second item set inactive.

What should those tests look like?  I've added a unit test to verify that we do the right thing when we receive `onWindowFocusChanged`, but I don't know how to write something to test that a fragment receives the events when we also handle them. Are there already tests like these for fragments that I could look at as an example?

> Do you know if we have the ability to write custom linters to warn the caller if they don't forward events to us?

No, I don' t know the answer to that, maybe @gnprice knows?

shell/platform/android/io/flutter/embedding/android/FlutterActivityAndFragmentDelegate.java
reidbaker: Feel free to leave it alone if the log message was intentional.

shell/platform/android/io/flutter/embedding/android/FlutterActivityAndFragmentDelegate.java
reidbaker: This comment might indicate that we need to keep track on onPaused and only hold the resumed state onWindowFocusedChangedI(true) if we got an onPause. 

What makes this difficult to reason about it it is a path where many customers will have an assumption that is  unlikely to be tested and those together can lead to really frustrated users.

lib/ui/platform_dispatcher.dart
```diff
@@ -1642,31 +1642,53 @@ class FrameTiming {
...
+/// notification will be sent before the application is suddenly terminated,
+/// along with the rest of the operating system.
+///
+/// For historical and name collision reasons, Flutter's application state names
+/// do not correspond 1:1 with the state names on all platforms. On Android, for
+/// instance, when the OS calls
+/// [`Activity.onPause`](https://developer.android.com/reference/android/app/Activity#onPause()),
+/// Flutter will enter the `inactive` state, but when Android calls
+/// [`Activity.onStop`](https://developer.android.com/reference/android/app/Activity#onStop()),
+/// Flutter enters the `paused` state. See the individual state documenation for
```
math1man, line 1656: nit: typo in the word "documentation"

shell/platform/android/io/flutter/embedding/android/FlutterActivityAndFragmentDelegate.java
```diff
@@ -898,6 +906,27 @@ void onUserLeaveHint() {
...
+    ensureAlive();
+    if (host.shouldDispatchAppLifecycleState() && flutterEngine != null) {
+      // This may notify in a redundant fashion with onResume and onPause. The
+      // lifecycle channel will deduplicate, but since the Android documentation
+      // says that we shouldn't rely on any particular order between lifecycle
+      // events and onWindowFocusChanged, we notify in both cases.
+      if (hasFocus) {
+        flutterEngine.getLifecycleChannel().appIsResumed();
+      } else {
+        flutterEngine.getLifecycleChannel().appIsInactive();
```
math1man, line 925: Is there any chance that this method could get called while the activity is stopped (after `onStop()`)? If so, this would break the `paused` lifecycle state. Even more so than the comment above about checking if the activity is Android-"resumed", I think it's probably necessary to check here that the activity is at least in the Android-"started" state.

shell/platform/android/io/flutter/embedding/android/FlutterFragment.java
```diff
@@ -1235,6 +1235,13 @@ public void onUserLeaveHint() {
     }
   }
 
+  @ActivityCallThrough
```
math1man, line 1238: Rather than requiring an explicit activity call through, you could use an [OnWindowFocusChangeListener](https://developer.android.com/reference/android/view/ViewTreeObserver.OnWindowFocusChangeListener) registered in `onViewCreated` and unregistered in `onDestroyView`. Something like

```java
private final OnWindowFocusChangeListener onWindowFocusChangeListener = 
    new OnWindowFocusChangeListener() {
      @Override
      public void onWindowFocusChanged(boolean hasFocus) {
        if (stillAttachedForEvent("onWindowFocusChanged")) {
          delegate.onWindowFocusChanged(hasFocus);
        }
      }
    }

@Override
public void onViewCreated(View view, Bundle savedInstanceState) {
  super.onViewCreated(view, savedInstanceState);
  view.getViewTreeObserver().addOnWindowFocusChangeListener(onWindowFocusChangeListener);
}

@Override
public void onDestroyView() {
  super.onDestroyView();
  requireView().getViewTreeObserver()
      .removeOnWindowFocusChangeListener(onWindowFocusChangeListener);
}
```
This assumes a minimum SDK version of 18+, so if Flutter supports older than that this may not be an option. And if you don't go this pattern, you need to update `FlutterFragmentActivity` to do the required call through.

lib/ui/platform_dispatcher.dart
```diff
@@ -1642,31 +1642,53 @@ class FrameTiming {
...
 ///
-///  * [WidgetsBindingObserver], for a mechanism to observe the lifecycle state
-///    from the widgets layer.
+/// * [WidgetsBindingObserver], for a mechanism to observe the lifecycle state
+///   from the widgets layer.
 enum AppLifecycleState {
-  /// The application is visible and responding to user input.
+  /// The application is visible and responsive to user input.
+  ///
+  /// On Android, this state corresponds to an app or the Flutter host view
```
xster, line 1666: not sure the `or` is apt here. If you consider the first part of the or, 'an app having focus while in Android's "resumed" state' could be true most of the time while this enum isn't `resumed` in hybrid apps

lib/ui/platform_dispatcher.dart
```diff
@@ -1642,31 +1642,53 @@ class FrameTiming {
...
+  /// having focus while in Android's "resumed" state. It is possible for the
+  /// Flutter app to be in the [inactive] state while still being in Android's
+  /// "resumed" state if the app has lost focus, but hasn't had
```
xster, line 1669: this is first paragraph where the word 'focus' is used in this file, and you're already saying here that it !=== to "resumed" which is a better known concept. Let's be very technically specific on what "focus" means and add links

lib/ui/platform_dispatcher.dart
math1man: It's also worth noting that "focus" here refers to the window, and a single application may have more than one window, thus implying part of the application has focus while other parts don't. I think even something as simple as displaying a native Android dialog (which a plugin could do) would trigger `onWindowFocusChanged`.

lib/ui/platform_dispatcher.dart
```diff
@@ -1642,31 +1642,53 @@ class FrameTiming {
...
+  /// On Android, this corresponds to an app or the Flutter host view running in
+  /// Android's paused state (i.e.
+  /// [`Activity.onPause`](https://developer.android.com/reference/android/app/Activity#onPause())
+  /// has been called), or in Android's "resumed" state (i.e.
+  /// [`Activity.onResume`](https://developer.android.com/reference/android/app/Activity#onResume())
+  /// has been called) but it has lost window focus. Examples of when apps
+  /// transition to this state include when the app is partially obscured or
+  /// another activity is focused, such as: a split-screen app, a phone call, a
+  /// picture-in-picture app, a system dialog, another view, when the
+  /// notification window shade is down, or the application switcher is visible.
```
xster, line 1691: are we sure about this for all OEMs? The Android docs aren't very clear about it

shell/platform/android/io/flutter/embedding/android/FlutterActivityAndFragmentDelegate.java
```diff
@@ -898,6 +906,27 @@ void onUserLeaveHint() {
...
+  /**
+   * Invoke this from {@code Activity#onWindowFocusChanged()}.
+   *
+   * <p>A {@code Fragment} host must have its containing {@code Activity} forward this call so that
+   * the {@code Fragment} can then invoke this method.
+   */
+  void onWindowFocusChanged(boolean hasFocus) {
+    ensureAlive();
+    if (host.shouldDispatchAppLifecycleState() && flutterEngine != null) {
+      // This may notify in a redundant fashion with onResume and onPause. The
```
xster, line 918: what's the sequence of the enum's traversal now with this change and without in the most basic case?

lib/ui/platform_dispatcher.dart
gspencergoog: I haven't tested all OEMs, I'm just assuming that they function similarly to PIxel when it comes to the window shade and app switcher. Keeping focus on the app in either of those two cases would be pretty confusing, I would think, but I suppose it could be done. What subset of OEMs would we need to test?

shell/platform/android/io/flutter/embedding/android/FlutterActivityAndFragmentDelegate.java
gspencergoog: The sequence is the same, there just might be redundant notifications when the app loses focus and then is made inactive, or is resumed and given focus.

lib/ui/platform_dispatcher.dart
math1man: Make sure to also test multi-window: https://developer.android.com/guide/topics/large-screens/multi-window-support#lifecycle

lib/ui/platform_dispatcher.dart
gspencergoog: What are the possible values it could be set to in the hybrid case where the app has focus and is in the Android resumed state?

I am assuming that if "onResume" has been called by Android on the activity, and the Flutter activity has focus, then the only Flutter state this should be in is "resumed". If the Flutter activity doesn't have focus, then it should be in "inactive".

shell/platform/android/io/flutter/embedding/android/FlutterActivityAndFragmentDelegate.java
gspencergoog: The thing I'm worried about is having more than one source for the state Flutter thinks the app is in.  If I keep information in this class, or in the `LifecycleChannel` class, so that I know if the activity is already resumed, or in a particular state, then also hold the state in the Dart code in the framework, then it's possible for them to get out of sync. If I send all the state changes to the framework and deduplicate there, then I only have one place where the state is stored, and it can't be out of sync.

shell/platform/android/io/flutter/embedding/android/FlutterActivityAndFragmentDelegate.java
gspencergoog: I can't say for sure, but it seems unlikely given Android's state machine. Any app that is stopped wouldn't be receiving any focus changes (or much of anything else event related, I would think), and it should go through "onPause" first, which would seem to me to be where Android should tell any app that it has lost focus. So there might be redundant "inactive" state notifications, but we wouldn't get an `onWindowFocusChanged(true)` until right before or right after `onResume` was called again (order depending on how Android sends its events), and at that time it would be warranted.

However, as you can probably tell, I'm not positive of any of that. These sorts of questions are why I need help here from the Android experts.

math1man: The documentation I read said you shouldn't rely on the ordering of `onWindowFocusChanged()` and other lifecycle methods, which I assume also includes `onStop()`. This makes me suspect it could be possible for the methods to be called in quick succession in order of `onPause()`-`onStop()`-`onWindowFocusChanged(false)`.

shell/platform/android/io/flutter/embedding/android/FlutterActivityAndFragmentDelegate.java
math1man: That's not entirely true though. The Android framework also defines a source-of-truth for both Android lifecycle state and window focus state, which could already conceivably be out of sync with the Dart state. Unfortunately, vanilla Android does not expose either of these states as directly introspectable (AndroidX provides an API for introspecting the lifecycle state). It doesn't seem to me like tracking the Android state locally is any worse, unless the framework majorly violates its own contracts.

shell/platform/android/io/flutter/embedding/android/FlutterActivityAndFragmentDelegate.java
gspencergoog: Hmm.  I read that too, which is why I was doing deduplication instead of trying to rely on the last reported state to decide. Sigh.  OK, if it can be that far out of sync, then I guess I have no choice but to keep track of both the Android state and the focus state locally, and send the appropriate state when they meet the right criteria, and just hope that they don't get out of sync with the state that the framework thinks it is in. That should only be possible when other code besides the `LifecycleChannel` sends messages to the `flutter/lifecycle` channel, so hopefully it's pretty rare.

I added a state table to the PR description to help clarify what states would be in force when.

shell/platform/android/io/flutter/embedding/android/FlutterActivityAndFragmentDelegate.java
gspencergoog: Okay, good point.  No, the framework just takes what it is told on the `flutter/lifecycle` channel as gospel.  It doesn't conditionally do anything else. If I plugin were to talk to it on that channel, it would be confused, but I supposed you get what you ask for there.

shell/platform/android/io/flutter/embedding/android/FlutterFragment.java
gspencergoog: Flutter doesn't support 18.0 or below, so I think we're fine.

https://docs.flutter.dev/reference/supported-platforms

gspencergoog: > And if you don't go this pattern, you need to update FlutterFragmentActivity to do the required call through.

I already did that in this PR. 

What's the advantage of doing it the way you propose?

lib/ui/platform_dispatcher.dart
gspencergoog: I don't know how to make a multi-window app where the Activity has multiple windows in it where one is a Flutter window. I'll look for an example.

I did test it with the split screen functionality (two activities up at the same time), and that works correctly. Clicking on one of them focuses the input, and sets the Flutter state to resumed, and clicking away sets it to inactive: it used to ignore that and not send anything.

gspencergoog:
Okay, I've updated it so that `LifecycleChannel` keeps track of the focus state and emits the appropriate state to the framework for the combination of Android lifecycle state and focus state.

shell/platform/android/io/flutter/embedding/android/FlutterFragment.java
math1man: The advantage is you don't need the activity call through. The logic is entirely self-contained to the fragment. Activity call through APIs are error-prone and difficult to work with, because a user could very easily forget to invoke them, and there isn't always an easy way to call through from the activity (such as a FlutterFragment nested within X other fragments). In a perfect world, there wouldn't be any methods in FlutterFragment requiring an activity call through.

Also, you need to update `shell/platform/android/io/flutter/embedding/android/FlutterFragmentActivity.java`. There's two of them, and the other one that you updated is (as I understand) deprecated.

reidbaker: This a much better way to avoid the issue I was trying to prevent with a linter. Host apps are less likley to mess our their own state by failing to pass calls through. 

Context: https://github.com/flutter/engine/pull/41094#issuecomment-1518314588

shell/platform/android/io/flutter/embedding/android/FlutterFragment.java
```diff
@@ -1109,12 +1120,21 @@ public void onStop() {
...
+
   @Override
   public void onDestroyView() {
     super.onDestroyView();
     if (stillAttachedForEvent("onDestroyView")) {
       delegate.onDestroyView();
     }
+    requireView()
+        .getViewTreeObserver()
+        .removeOnWindowFocusChangeListener(onWindowFocusChangeListener);
```
math1man, line 1137: You may wan to call this before calling `delegate.onDestroyView()`. I don't know if it matters though.

lib/ui/platform_dispatcher.dart
gnprice: I believe that in Android multi-window mode (which includes split-screen mode), each of the windows has to be a different Activity.

Typically this means the windows come from unrelated apps (and [the doc](https://developer.android.com/guide/topics/large-screens/multi-window-support) repeatedly refers to "two apps", "another app", etc.), because most apps only ever show one activity at a time. But I just tried a quick demo with one of the few apps I use that cheerfully makes multiple activities at a time, namely Google Sheets: open the app, open one sheet, go back to the list of sheets, open a second sheet, put the two sheets in split-screen mode. And split-screen mode worked great for those two activities from the same app.

So after testing split-screen mode where one of the activities is a Flutter window, the next more complex edge case would be split-screen mode where both of the activities are Flutter windows from the same app. (I'm not sure if that's currently possible, though, or if it's still in the future — I know you've been working on Flutter multi-window support.)

shell/platform/android/io/flutter/embedding/android/FlutterFragment.java
gspencergoog: Good point. Moved.

shell/platform/android/io/flutter/embedding/engine/systemchannels/LifecycleChannel.java
```diff
@@ -5,39 +5,95 @@
...
+    }
+    channel.send(newState);
+    lastFlutterState = newState;
+  }
+
+  public void aWindowIsFocused() {
+    sendState(lastAndroidState, true);
+  }
+
+  public void noWindowsAreFocused() {
```
gnprice, line 80: In particular, when switching from one activity to another in split-screen mode, I wonder if the `onWindowFocusChanged(false)` on the old activity comes before or after the `onWindowFocusChanged(true)` on the new one (or if both orderings are possible.)

If the `false` always comes before the `true`, then these method names are accurate. But if the `true` can come before the `false`, then on that belated `false` we'll call `noWindowsAreFocused` when in fact one window is focused, and set the Flutter state to "inactive".

gnprice (commented):
I like this state-table-based approach!

shell/platform/android/io/flutter/embedding/engine/systemchannels/LifecycleChannel.java
math1man: Two activities using the same flutter engine is likely to not work very well or at all, because there's no way to differentiate the Flutter UI between the two activities. And if the activities have different flutter engines, they should also have different LifecycleChannels. It's likely not an edge case you'd need to worry about.

gspencergoog: The intent of the "inactive" state in the framework is how I named the functions. I suspect that once we do support multiple windows in the same app (it is coming, but probably coming to mobile platforms last), there will have to be some work done to give the right information to these functions where we check on focus change and see if greater than zero windows have focus before calling one function or the other. Then the difference in order will only mean that either we notify of loss and then immediate regain of focus, or we don't notify at all, since there were never zero windows focused.

lib/ui/platform_dispatcher.dart
gspencergoog: Okay, PTAL, I think I addressed this.

shell/platform/android/io/flutter/embedding/engine/systemchannels/LifecycleChannel.java
gnprice: > I suspect that once we do support multiple windows in the same app (it is coming, but probably coming to mobile platforms last), there will have to be some work done to give the right information to these functions where we check on focus change and see if greater than zero windows have focus before calling one function or the other.

Cool, that plan sounds good to me.

I guess my suggestion then would just be to either name the functions in the singular to reflect the current assumptions, or add a comment here explaining how the names reflect a future that's planned but hasn't yet arrived (and therefore doesn't yet call for logic to handle).

gspencergoog: I'll add a comment: I don't want to have to change the API later, and they're still accurate names even if the max number of windows is one.

lib/ui/platform_dispatcher.dart
gspencergoog: > I'm picking on the doc wording, not on the implementation.

Sure, I got that, and a totally fine thing to do! I want to get it right.

>  In hybrid apps where 95% of the app isn't Flutter, you're not going to be AppLifecycleState.resumed if "the app has focused while in Android's resumed state". The Flutter view might not even be attached to the FlutterEngine.

If the Flutter view isn't attached to the engine, wouldn't it be in Android's "Detached" state? In that state, it doesn't matter what the focus is, the Flutter state will be `AppLifecycleState.detached`. Only in Android's "Resumed" state am I making any changes to how things work. Take a look at the state table I put in the PR description and tell me if that helps explain it.

lib/ui/platform_dispatcher.dart
gspencergoog: Okay, I tried the Realme (I couldn't find an Oppo), and although it's app switcher takes it all the way to `paused` (Android's `onStop` state), unlike the Pixel one (which just notifies of focus loss, and leaves the app in the "onResume" Android state), it does correct notification. I suspect it does this because it doesn't provide a live view of the app in the app switcher, just a snapshot. The changes also work fine in split screen mode, and with the window shade.

shell/platform/android/io/flutter/embedding/android/FlutterFragment.java
```diff
@@ -1109,9 +1125,23 @@ public void onStop() {
     }
   }
 
+  @Override
+  @TargetApi(18)
+  @RequiresApi(18)
+  public void onViewCreated(View view, Bundle savedInstanceState) {
+    super.onViewCreated(view, savedInstanceState);
+    view.getViewTreeObserver().addOnWindowFocusChangeListener(onWindowFocusChangeListener);
```
math1man, line 1133: You also need to guard this by an API level check. Adding `@RequiresApi(18)` to an Android lifecycle method is a bad idea. But I don't think this is the correct solution. You said Flutter only supports Android 19+, so if you're running into API level issues, you likely just need to change some `minSdkVersion` values wherever you're hitting the issue. Probably this [AndroidManifest](https://github.com/flutter/engine/blob/main/shell/platform/android/AndroidManifest.xml) and possibly also in some test code.

gspencergoog: Yes, it seems that while we do only officially support only 19+, the migration of our build environment is still in progress, so we need the min version checks until all the `minSdkVersion`s have been bumped.

See https://github.com/flutter/flutter/pull/125515

math1man: In that case, you should not put `@RequiresApi` on `onViewCreated`, and instead add an explicit API level check as you did on `onDestroyView`. You may also want to manually verify that the field declaration using `@RequiresApi` doesn't break anything on Android 17 or below. My instinct is that it shouldn't as long as the field is otherwise unreferenced, but I'm not 100% positive.

math1man: Just confirmed with an Android framework expert that the field will definitely be problematic. You should be able to solve that by initializing the value like 
```
Build.VERSION.SDK_INT >= 18 ? new OnWindowFocusChangeListener() { ... } : null
```

gspencergoog: Should I continue to use `@TargetApi(18)` for `onViewCreated`, now that I've got the version check? Or is it redundant now that the `onViewCreated` doesn't have `@RequiresApi`? There's no warning to suppress.

math1man: The advice I've heard from AndroidX/Android Framework teams is never use `@TargetApi` at all. But you shouldn't apply either to `onViewCreated` since you've applied a runtime internal version check.

gspencergoog:
Okay, I've added tests for the `LifecycleChannel`, and I've done manual testing on some devices. Are there any other tests I should add?

Otherwise, I think this is ready for another look.

lib/ui/platform_dispatcher.dart
reidbaker, line 1: Great documentation!

reidbaker (approved):

PR #41098, opened 2023-04-12
[Impeller] add a generic porter duff blend foreground shader.

jonahwilliams:
The pipeline blend component of https://github.com/flutter/flutter/issues/124025 . This should also land alongside https://github.com/flutter/flutter/pull/124640 which was necessary before we had completely optimized the image filters for texture inputs.

Fixes https://github.com/flutter/flutter/issues/121650
Fixes https://github.com/flutter/flutter/issues/124025


impeller/entity/shaders/blending/porter_duff_blend.frag
```diff
@@ -0,0 +1,74 @@
...
+    frag_color = IPBlendSourceATop(frag_info.color, dst_color);
+  } else if (frag_info.operation == kDestinationATop) {
+    frag_color = IPBlendDestinationATop(frag_info.color, dst_color);
+  } else if (frag_info.operation == kXor) {
+    frag_color = IPBlendXOR(frag_info.color, dst_color);
+  } else if (frag_info.operation == kPlus) {
+    frag_color = IPBlendPlus(frag_info.color, dst_color);
+  } else {
+    frag_color = IPBlendModulate(frag_info.color, dst_color);
+  }
```
bdero, line 73: Rather than enumerating these on the GPU, most of these can be parameterized with 4 coefficients:
```glsl
frag_color =
    src * (frag_info.src_coeff + dst.a * frag_info.src_coeff_dst_alpha) +
    dst * (frag_info.dst_coeff + src.a * frag_info.dst_coeff_src_alpha);
```

And then the coefficients can be pulled from a LUT on the CPU without enumeration or hashing:

```cpp
{src_coeff, src_coeff_dst_alpha, dst_coeff, dst_coeff_src_alpha}

{0,  0,  0,  0}, // Clear
{1,  0,  0,  0}, // Source
{0,  0,  1,  0}, // Destination
{1,  0,  1, -1}, // SourceOver
{1, -1,  1,  0}, // DestinationOver
{0,  1,  0,  0}, // SourceIn
{0,  0,  0,  1}, // DestinationIn
{1, -1,  0,  0}, // SourceOut
{0,  0,  1, -1}, // DestinationOut
{0,  1,  1, -1}, // SourceATop
{1, -1,  0,  1}, // DestinationATop
{1, -1,  1, -1}, // Xor
{1,  0,  1,  0}, // Plus
```

I think we can also cover `Modulate` and `Screen` here by adding one more dst coefficient to multiply the source color in:

```glsl
frag_color =
    src * (frag_info.src_coeff + dst.a * frag_info.src_coeff_dst_alpha) +
    dst * (frag_info.dst_coeff + src.a * frag_info.dst_coeff_src_alpha +
                                 src * frag_info.dst_coeff_src_color);
```

...and then the full table with all the blends becomes:

```cpp
{src_coeff, src_coeff_dst_alpha, dst_coeff, dst_coeff_src_alpha, dst_coeff_src_color}

{0,  0,  0,  0,  0}, // Clear
{1,  0,  0,  0,  0}, // Source
{0,  0,  1,  0,  0}, // Destination
{1,  0,  1, -1,  0}, // SourceOver
{1, -1,  1,  0,  0}, // DestinationOver
{0,  1,  0,  0,  0}, // SourceIn
{0,  0,  0,  1,  0}, // DestinationIn
{1, -1,  0,  0,  0}, // SourceOut
{0,  0,  1, -1,  0}, // DestinationOut
{0,  1,  1, -1,  0}, // SourceATop
{1, -1,  0,  1,  0}, // DestinationATop
{1, -1,  1, -1,  0}, // Xor
{1,  0,  1,  0,  0}, // Plus
{0,  0,  0,  0,  1}, // Modulate
{0,  0,  1,  0, -1}, // Screen
```

+cc @flar in case he has a moment to check my math :)

jonahwilliams: Seems like a more technically sound approach, I'll make this change.

jonahwilliams: Done, lines up with the current behavior using a 0.5 alpha red and blue:

ToT

flutter_01

Patched:

flutter_03

bdero (approved):
LGTM!

jonahwilliams:
Goldens revealed that the check to convert the foreground blend back into a snapshot wasn't quite working correctly so I took it out. now we only apply the foreground if we can absorb opacity (which should generally be always)

PR #41100, opened 2023-04-12
[Impeller] Keep a reference to BlitPassGLES in the EncodeCommands lambda to prevent use-after-free

jason-simmons:
Without this the BlitPassGLES and its command list can be destructed before the reactor executes the lambda.


dnfield (approved):
Lgtm

bdero (approved):
LGTM

jason-simmons:
`RenderPassGLES::OnEncodeCommands` is making a copy of its command list and does not have the object lifetime issue.

But we might get a performance improvement by doing something like this in `RenderPassGLES` to avoid the copy.

PR #41136, opened 2023-04-12
[web:canvaskit] move path API to UniqueRef

yjbanov:
Moves all path API (`Path`, `PathMetrics`) from `ManagedSkiaObject` to `UniqueRef`, and clean up all the dead resurrection logic since all browsers support `FinalizationRegistry`.

This is a follow-up to https://github.com/flutter/engine/pull/40862


hterkelsen (approved):
LGTM

PR #41138, opened 2023-04-12
Revert "Add support for double tap action from Apple Pencil 2"

LouiseHsu:
Reverts flutter/engine#39637

Since this feature https://github.com/flutter/flutter/issues/73172 has been scrapped for now, previous engine work should be reverted. Refer to [this design doc](https://docs.google.com/document/d/1r4P5r-jGt2Sjqro3ldCU2axUiHTpu3yhIycnI94OKQw/edit ) for more discussion and details.

### DIFF
--- a/lib/ui/platform_dispatcher.dart
+++ b/lib/ui/platform_dispatcher.dart
@@ -392,7 +392,7 @@ class PlatformDispatcher {
   //  * pointer_data.cc
   //  * pointer.dart
   //  * AndroidTouchProcessor.java
-  static const int _kPointerDataFieldCount = 36;
+  static const int _kPointerDataFieldCount = 35;
 
   static PointerDataPacket _unpackPointerDataPacket(ByteData packet) {
     const int kStride = Int64List.bytesPerElement;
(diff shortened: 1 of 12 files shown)


jmagman (approved):

hellohuanlin (approved):
Thx for cleaning this up!

zanderso:
https://github.com/flutter/engine/pull/39637 is included in the 3.10 beta. Does this revert need to be CP'd to the release branch?

jmagman:
> #39637 is included in the 3.10 beta. Does this revert need to be CP'd to the release branch?

I don't think so, `preferredStylusAuxiliaryAction` was never hooked up to anything in the framework, so it's unused in the engine but it's not hurting anything.  @LouiseHsu @hellohuanlin can check me if I'm missing something though.

hellohuanlin:
@LouiseHsu did you land the framework or engine change first? When reverting, it's probably safer to do the opposite order. (I'm a bit lost in the "revert of revert" in one of your PR, but I can help to look into if you don't remember either)

jmagman:
> @LouiseHsu did you land the framework or engine change first? When reverting, it's probably safer to do the opposite order. (I'm a bit lost in the "revert of revert" in one of your PR, but I can help to look into if you don't remember either)

I think the only framework change that made it in was https://github.com/flutter/flutter/pull/120731 which wouldn't need to be reverted in any particular order.

PR #41141, opened 2023-04-12
Add builders for arm64 on macOS unopt

dnfield:
Fixes https://github.com/flutter/flutter/issues/122534

Adds a descritiption for building the host_debug_unopt_arm64 and ios_sim_unopt_arm64 to the mac arm64 bots.

Does not run host tests yet, but will be easy to do that after https://github.com/flutter/flutter/issues/103386 is resolved.

@cbracken fyi

### DIFF
--- a/ci/builders/mac_unopt.json
+++ b/ci/builders/mac_unopt.json
@@ -165,7 +165,7 @@
         {
             "archives": [
                 {
-                    "base_path": "out/ios_debug_sim/zip_archives/",
+                    "base_path": "out/ios_debug_sim_arm64/zip_archives/",
                     "type": "gcs",
                     "include_paths": [
                     ],
@@ -195,7 +195,7 @@
                 "--simulator-cpu",
                 "arm64"
             ],
-            "name": "ios_debug_sim",
+            "name": "ios_debug_sim_arm64",
             "ninja": {
                 "config": "ios_debug_sim_arm64",
                 "targets": [
--- a/ci/builders/mac_unopt.json
+++ b/ci/builders/mac_unopt.json
@@ -146,21 +146,7 @@
                 "targets": [
                 ]
             },
-            "tests": [
-                {
-                    "language": "python3",
-                    "name": "Host Tests for host_debug_unopt_arm64",
-                    "parameters": [
-                        "--variant",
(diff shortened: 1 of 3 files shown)


ci/builders/mac_unopt.json
```diff
@@ -111,6 +111,109 @@
...
+            ]
+        },
+        {
+            "archives": [
+                {
+                    "base_path": "out/host_debug_unopt_arm64/zip_archives/",
+                    "type": "gcs",
+                    "include_paths": [
+                    ],
+                    "name": "host_debug_unopt_arm64"
```
godofredoc, line 123: Is the only purpose of this build to validate it builds correctly?

ci/builders/mac_unopt.json
```diff
@@ -111,6 +111,109 @@
...
+                }
+            ],
+            "drone_dimensions": [
+                "device_type=none",
+                "os=Mac-12",
+                "cpu=arm64"
+            ],
+            "gclient_custom_vars": {
```
godofredoc, line 131: gclient_variables?

jmagman (approved):
cc @keyonghan this will add more arm builders that potentially could be xcode-cache corrupted if we try to update just the arm bots to test out your named cache work. However we can turn these off temporarily if that's a problem, if needed.

LGTM once @godofredoc is happy.

godofredoc:
Adding my +2 with nits because I'll be ooo the rest of the day.

godofredoc (approved):
LGTM with nits

ci/builders/mac_unopt.json
dnfield: For now, until we can run tests.

ci/builders/mac_unopt.json
dnfield: This was copy-pasted from the other ones.

ci/builders/mac_unopt.json
dnfield: I should have said:

We're not ready to run host unit tests on CI for reasons in https://github.com/flutter/flutter/issues/103386. It seems close.

This is running tests on iOS simulator though.

PR #41144, opened 2023-04-13
Implement Gradient Shaders and Fragment Shaders in Skwasm

eyebrowsoffire:
This implements gradient shaders (linear, radial, conical, and sweep) as well as custom fragment shaders in Skwasm.

Something is still up with the ShaderMask layer though, it doesn't quite render correctly on Skwasm.

Edit: fixed and it's rendering consistently with CanvasKit now.


lib/web_ui/lib/src/engine/skwasm/skwasm_impl/raw/raw_memory.dart
```diff
@@ -132,6 +132,16 @@ class StackScope {
...
+  Pointer<Float> convertDoublesToNative(Iterable<double> values) {
+    final Pointer<Float> pointer = allocFloatArray(values.length);
+    int i = 0;
+    for (final double value in values) {
+      pointer[i] = value;
+      i++;
+    }
+    return pointer;
+  }
+
```
jonahwilliams, line 144:
```suggestion
  Pointer<Float> convertDoublesToNative(List<double> values) {
    final Pointer<Float> pointer = allocFloatArray(values.length);
    for (int i = 0; i < values.length; i++) {
      pointer[i] = value;
    }
    return pointer;
  }

```

jonahwilliams (approved):
LGTM with some nits

PR #41148, opened 2023-04-13
[Impeller] Fix the cmake example build.

zanderso:
To fit in better with the `engine_v2` context, this PR shifts the `cmake` step for the `impeller-cmake-example` to run within the `flutter/tools/gn` script. That way, the recipe will ensure that `GOMA_DIR` is defined, goma is running, etc.. Since we use `cmake` to generate a ninja build for `impeller-cmake-example`, the engine_v2 `ninja` step can build it, and that feature of the python script is no longer needed.

Successful run is here https://ci.chromium.org/ui/p/flutter/builders/try/Mac%20impeller-cmake-example/12/overview

### DIFF
--- a/.ci.yaml
+++ b/.ci.yaml
@@ -450,12 +450,11 @@ targets:
     timeout: 60
     properties:
       release_build: "true"
-      config_name: impeller_cmake_build_test
+      config_name: mac_impeller_cmake_example
       $flutter/osx_sdk : >-
         { "sdk_version": "14a5294e" }
       dependencies: >-
         [
-          {"dependency": "cmake", "version": "build_id:8784715802296535313"},
           {"dependency": "jazzy", "version": "0.14.1"}
         ]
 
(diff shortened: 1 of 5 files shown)


.ci.yaml
```diff
@@ -445,17 +445,16 @@ targets:
         ]
 
   - name: Mac impeller-cmake-example
-    bringup: true
+    #bringup: true
```
zanderso, line 448: I will uncomment this before landing. It's uncommented to force it to run in the presubmit checks.

.ci.yaml
```diff
@@ -445,17 +445,16 @@ targets:
...
     timeout: 60
     properties:
       release_build: "true"
-      config_name: impeller_cmake_build_test
+      config_name: mac_impeller_cmake_example
       $flutter/osx_sdk : >-
         { "sdk_version": "14a5294e" }
       dependencies: >-
         [
-          {"dependency": "cmake", "version": "build_id:8784715802296535313"},
```
zanderso, line 458: Pulling this in via the DEPS file instead since it's just a CIPD package.

chinmaygarde (approved):

bdero (approved):
LGTM

PR #41159, opened 2023-04-13
Do not use Macmini7,1 for host tests

dnfield:
fixes https://github.com/flutter/flutter/issues/124678

those machines do not support compute subgroups. They make up only a relatively small fraction of the pool.

This does not restrict the arm machines, as they're already restricted by the cpu architecture dimension.

### DIFF
--- a/ci/builders/mac_host_engine.json
+++ b/ci/builders/mac_host_engine.json
@@ -16,7 +16,8 @@
             "drone_dimensions": [
                 "device_type=none",
                 "os=Mac-12",
-                "cpu=x86"
+                "cpu=x86",
+                "mac_model=Macmini8,1"
             ],
             "gclient_custom_vars": {
                 "download_android_deps": false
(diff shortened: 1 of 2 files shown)


zanderso (approved):
I wish we could add comments to `.json` files to put a link to the issue explaining why the restriction is there!

keyonghan:
Seems the change is not picked up by the pre-submit build. https://ci.chromium.org/ui/p/flutter/builders/try/Mac%20Engine%20Drone/160302/overview still uses the old `drone_dimension`. Do you need a bug? /cc @godofredoc

dnfield:
> Seems the change is not picked up by the pre-submit build. https://ci.chromium.org/ui/p/flutter/builders/try/Mac%20Engine%20Drone/160302/overview still uses the old `drone_dimension`. Do you need a bug? /cc @godofredoc

That one is allowed to run on the old ones - I only updated the dimensions for the drones that run `Host tests for host*`

keyonghan:
> Seems the change is not picked up by the pre-submit build. https://ci.chromium.org/ui/p/flutter/builders/try/Mac%20Engine%20Drone/160302/overview still uses the old `drone_dimension`. Do you need a bug? /cc @godofredoc

I took this back. The change takes effect right away. Here is the task where the dimension is enforced: https://chromium-swarm.appspot.com/task?id=619400c517759a10&o=true&w=true

keyonghan (approved):
LGTM

PR #41163, opened 2023-04-13
[web] Expose PlatformLocation and HashUrlStrategy through ui_web

mdebbar:
1. Expose `PlatformLocation` and its subclass `BrowserPlatformLocation` through the new `ui_web`.
2. Expose `HashUrlStrategy` too. It's useful for users who want to extend and customize it instead of building their own from scratch.
3. `ui_web/url_strategy.dart` => `ui_web/navigation/url_strategy.dart`.


eyebrowsoffire (approved):
Awesome! Looks great!

lib/web_ui/lib/ui_web/src/ui_web/navigation/platform_location.dart
```diff
@@ -0,0 +1,126 @@
...
+
+/// Encapsulates all calls to DOM apis, which allows the [UrlStrategy] classes
+/// to be platform agnostic and testable.
+///
+/// For convenience, the [PlatformLocation] class can be used by implementations
+/// of [UrlStrategy] to interact with DOM apis like pushState, popState, etc.
+abstract class PlatformLocation {
+  /// Abstract const constructor. This constructor enables subclasses to provide
+  /// const constructors so that they can be used in const expressions.
+  const PlatformLocation();
```
yjbanov, line 17: It looks like this class is meant to be implemented rather than extended, so we shouldn't need any constructor at all. Also, if we're allowed to use Dart 3 features, maybe we can make it `interface class` (or whatever the new syntax is for interfaces).

lib/web_ui/lib/ui_web/src/ui_web/navigation/url_strategy.dart
```diff
@@ -0,0 +1,155 @@
...
+
+import 'dart:async';
+
+import 'package:ui/src/engine.dart';
+import 'package:ui/ui.dart' as ui;
+
+import 'platform_location.dart';
+
+
+set urlStrategy(UrlStrategy? strategy) => customUrlStrategy = strategy;
```
yjbanov, line 13: This needs API docs.

lib/web_ui/lib/ui_web/src/ui_web/navigation/url_strategy.dart
```diff
@@ -0,0 +1,155 @@
...
+
+typedef PopStateListener = void Function(Object? state);
```
yjbanov, line 15: API docs.

lib/web_ui/lib/ui_web/src/ui_web/navigation/url_strategy.dart
```diff
@@ -0,0 +1,155 @@
...
+
+/// Represents and reads route state from the browser's URL.
+///
+/// By default, the [HashUrlStrategy] subclass is used if the app doesn't
+/// specify one.
+abstract class UrlStrategy {
+  /// Abstract const constructor. This constructor enables subclasses to provide
+  /// const constructors so that they can be used in const expressions.
+  const UrlStrategy();
```
yjbanov, line 24: Same here. All methods are abstract, which tells me this class can be implemented. So we don't need a constructor. The implementation can still be const.

lib/web_ui/lib/ui_web/src/ui_web/navigation/url_strategy.dart
```diff
@@ -0,0 +1,155 @@
...
+/// import 'package:flutter_web_plugins/flutter_web_plugins.dart';
+///
+/// // Somewhere before calling `runApp()` do:
+/// setUrlStrategy(const HashUrlStrategy());
+/// ```
+class HashUrlStrategy extends UrlStrategy {
+  /// Creates an instance of [HashUrlStrategy].
+  ///
+  /// The [PlatformLocation] parameter is useful for testing to mock out browser
+  /// interations.
```
yjbanov, line 82: typo: "interactions"

PR #41059, opened 2023-04-11
[Impeller] [Android] Refactor the Android context/surface implementation to work more like Skia.

dnfield:
Noticed this while working on  https://github.com/flutter/flutter/issues/124181

In Impeller right now, Android creates a context for each surface - but in Skia, Android creates one context and shares it across surfaces. For the multiple-engine scenario this matters. This also makes it possible for Android to provide the impeller::Context independently of surface creation.

PR #41168, opened 2023-04-13
[Impeller] Allow image rasterization/decoding before/without surface acquisition

dnfield:
This is the same as #41059 with the Android stuff reverted.

This change provides an `impeller::Context` to the rasterizer so that it can help with snapshotting even if the surface has not been setup yet. It also makes the image decoding logic use that context directly instead of trying to acquire it through a surface, which the rasterizer may or may not have when the image decode is requested.

I originally (incorrectly) believed I'd have to implement the Android side for any of it to work. I think we should still land the Android changes, but this will make for a smaller cherry pick and just leave Android broken (the other patch didn't fix Android with GLES anyway).

Fixes https://github.com/flutter/flutter/issues/124181


shell/common/rasterizer.cc
```diff
@@ -52,6 +52,10 @@ fml::TaskRunnerAffineWeakPtr<SnapshotDelegate> Rasterizer::GetSnapshotDelegate()
     const {
   return weak_factory_.GetWeakPtr();
 }
+void Rasterizer::SetImpellerContext(
```
loic-sharma, line 55:
```suggestion

void Rasterizer::SetImpellerContext(
```

shell/common/shell.cc
```diff
@@ -186,30 +186,33 @@ std::unique_ptr<Shell> Shell::CreateShellOnPlatformThread(
                     !settings.skia_deterministic_rendering_on_cpu),
                 is_gpu_disabled));
 
+  // Create the platform view on the platform thread (this thread).
+  auto platform_view = on_create_platform_view(*shell.get());
```
loic-sharma, line 190: Just to double-check, will `Shell::OnPlatformViewCreated` still work as expected and call `Rasterizer::Setup` with the created platform view?

dnfield: Yes

dnfield: Basically we're dealing with an artifact of targeting OpenGL first - in GL you need a surface to do rendering, but in Metal and Vulkan you don't. Even in GL we can just create pbuffer surfaces to do the rendering when we need to.

dnfield:
Failures appear unrelated to this patch, updating branch...

dnfield:
Failures on the mac_host_engine are unrelated to this pr, see https://github.com/flutter/engine/pull/41179

PR #41172, opened 2023-04-13
[web:canvaskit] move shaders to UniqueRef; fix minor memory leaks

yjbanov:
Move all classes implementing the `Shader` interface from `ManagedSkiaObject` to `UniqueRef`.

Fix two (probably minor) memory leaks:

* `CkImageShader` was leaking intermediate instances of `SkShader` because those were neither attached to a `FinalizationRegistry`, nor disposed of explicitly through the `dispose()` method.
* `CkFragmentShader` was leaking `SkShader` instances because it never called `CkFragmentInstance.dispose`. Additionally, the `CkFragmentInstance` class isn't really necessary, because `UniqueRef` is fully sufficient to manage instances of `SkShader` objects for fragment shaders.

### DIFF
--- a/lib/web_ui/lib/src/engine/canvaskit/painting.dart
+++ b/lib/web_ui/lib/src/engine/canvaskit/painting.dart
@@ -4,15 +4,17 @@
 
 import 'dart:typed_data';
 
-import 'package:ui/src/engine/shader_data.dart';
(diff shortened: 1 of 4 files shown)


hterkelsen (approved):
LGTM

PR #41129, opened 2023-04-12
[Impeller] Check coverage after screen clip

bdero:
Resolves https://github.com/flutter/flutter/issues/124658.

Need to add error messages around all failed scenarios in EntityPass since we don't have a recovery and the whole frame fails to render. Will CP this.

Before: Failed rendering.

After:

image

PR #41178, opened 2023-04-13
[Impeller] Print a log on all non-recoverable failure paths in EntityPass

bdero:
This eliminates a lot of surface area when debugging non-recoverable frame failures such as https://github.com/flutter/engine/pull/41129. Every failure case should be covered with at least one log, but without recursive log spam.


chinmaygarde (approved):

bdero:
I think the unrelated golden issue is on the base branch and fixed at HEAD, so ignoring.

PR #41179, opened 2023-04-13
Macmini8,1 for another target that runs host tests

dnfield:
This target also runs impeller_unittests that aren't safe to run right now on Macmini7,1

I missed this in https://github.com/flutter/engine/pull/41159


zanderso (approved):

dnfield:
@CaseyHillers should these be CP'd? I'm not sure if the release branch is using the engine_v2 recipes, but the flakiness will show up there too since we have tests that always fail on the Intel Iris 5100 GPUs.

PR #41183, opened 2023-04-14
Run clang tidy builds on arm Macs

jmagman:
Intel Mac capacity is limited.  Swap the two Mac clang-tidy builders to arm bots.  The clang-tidy linter should run with arm natively.

### Mac Host clang-tidy
Prod x64: https://ci.chromium.org/p/flutter/builders/prod/Mac%20Host%20clang-tidy/3318 Execution 35 mins 2 secs
This PR arm: https://ci.chromium.org/p/flutter/builders/try/Mac%20Host%20clang-tidy/8087 Execution 36 mins 30 secs (this hit an Xcode cache miss which added 4 minutes)

### Mac iOS clang-tidy
Prod x64: https://ci.chromium.org/p/flutter/builders/prod/Mac%20iOS%20clang-tidy/3303 Execution 11 mins 57 secs
This PR arm: https://ci.chromium.org/p/flutter/builders/try/Mac%20iOS%20clang-tidy/8087 Execution 18 mins 10 secs (this hit an Xcode cache miss which added 5 minutes)

### DIFF
--- a/.ci.yaml
+++ b/.ci.yaml
@@ -383,7 +383,7 @@ targets:
     recipe: engine/engine_lint
     properties:
       add_recipes_cq: "true"
-      cores: "12"
+      cpu: arm64
       lint_host: "true"
       lint_ios: "false"
     timeout: 75
@@ -405,6 +405,7 @@ targets:
     recipe: engine/engine_lint
     properties:
       add_recipes_cq: "true"
+      cpu: arm64
       lint_host: "false"
       lint_ios: "true"
     timeout: 75
--- a/.ci.yaml
+++ b/.ci.yaml
@@ -383,6 +383,7 @@ targets:
     recipe: engine/engine_lint
     properties:
       add_recipes_cq: "true"
+      cores: "8"
       cpu: arm64
       lint_host: "true"
       lint_ios: "false"


.ci.yaml
```diff
@@ -383,7 +383,7 @@ targets:
     recipe: engine/engine_lint
     properties:
       add_recipes_cq: "true"
-      cores: "12"
```
jmagman, line 386: The arm bots have 8 cores but should be faster.

.ci.yaml
```diff
@@ -383,7 +383,8 @@ targets:
...
+      cores: "8"
```
jmagman, line 386: This dimension should be removed, but when I do the try falls back to 12 in the upstreamed config and fails with no resource https://chromium-swarm.appspot.com/task?id=619612e614a28f10&o=true&w=true.

The ones in prod are also 8 core so this should still run there.  I'll remove it when the config syncs after merging.

zanderso (approved):

.ci.yaml
keyonghan: Yeah, all arms are with 8 cores. Why not removing the cores from this PR?

jmagman: > Yeah, all arms are with 8 cores. Why not removing the cores from this PR?

When I remove it the build fails with resource unavailable: when I do the try falls back to 12 in the upstreamed config and fails with no resource https://chromium-swarm.appspot.com/task?id=619612e614a28f10&o=true&w=true.

I assumed it's not passing in "no cores", it's falling back to the synced config that says 12?  I was thinking I would merge this and then try to remove the cores part in a future PR.

jmagman: I'm going to merge this and then remove the cores dimension in the next PR.

PR #36143, opened 2022-09-14
feat: add custom cursor interface on windows

Kingtous:
This PR adds an additional interface to set custom cursor on windows, which fixes issues mentioned below.

Currently,  Customizing `MouseCursor` on windows is not working, the customed mouse will restore to arrow once it moves. See https://github.com/flutter/flutter/issues/31952#issuecomment-1056115014 
https://github.com/flutter/flutter/issues/99484
for details. Variable `current_cursor_` in `cursor handler` will not update when we set our custom cursors, and we cannot change this field to prevent the cursor from restoring due to lacking of interface.

The reason is that cursor handler on windows has no interface to change current cursor to custom cursor, only support setting windows self-contained icons.  

Changes:
- add new function handled by method channel of `SystemCursors.mousecursors`, which can change `current_cursor_`, make possible to handle custom cursor on windows correctly.

Screenshot_20220914_183105

We have successfully tested this interface in https://github.com/Kingtous/rustdesk_flutter_custom_cursor/blob/8f3be4ac68b4fb20d05b8d30df4c13d8fd77f9c4/lib/flutter_custom_cursor.dart#L113, which works perfectly on windows.

*If you had to change anything in the [flutter/tests] repo, include a link to the migration guide as per the [breaking change policy].*

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [Flutter Style Guide] and the [C++, Objective-C, Java style guides].
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I added new tests to check the change I am making or feature I am adding, or Hixie said the PR is test-exempt. See [testing the engine] for instructions on
writing and running engine tests.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I signed the [CLA].
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/flutter/wiki/Tree-hygiene#overview
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[C++, Objective-C, Java style guides]: https://github.com/flutter/engine/blob/main/CONTRIBUTING.md#style
[...]

PR #41186, opened 2023-04-14
feat: add custom cursor capabilities to web platform. Addresses #31952, #89351 and #99484

timmaffett:
This change allows the web platform to implement custom mouse cursors.
It simply allows `image-set`, `-webkit-image-set` and `url` css commands to be set from the cursor's `kind` attribute.
This allows the kind to hold device pixel ratio supporting custom cursor images.

There is a web live example using this engine change [here](https://timmaffett.github.io/custom_mouse_cursor/).

I have tried to create a package with a clean unified interface to custom cursors for windows, macOS, linux and web.
[https://pub.dev/packages/custom_mouse_cursor](https://pub.dev/packages/custom_mouse_cursor)

The package implements complete device pixel ratio responsive cursors that change automatically to support higher DPR screens and allows specifying DPR assets or multiple images for a cursor at different DPRs, etc.
It also allows creating cursors from icons.

It allows us to get closer to what's need for https://github.com/flutter/flutter/issues/31952  and https://github.com/flutter/flutter/issues/89351 and https://github.com/flutter/flutter/issues/99484.
I have tried to implement this in a way that could possibly be considered for core flutter engine support of custom mouse cursors.

For windows this uses @Kingtous 's PR https://github.com/flutter/engine/pull/36143 additions.
At the macos and linux advantage of @imiskolee  and @Kingtous 's work in their lower level flutter_custom_cursor plugin.
(These directly mirror the platform channel methods in https://github.com/flutter/engine/pull/36143  ).

## Pre-launch Checklist

- [X] I signed the [CLA].


kevmoo:
LOVE the example app running! Cool!

kevmoo:
is there anyway to have a test here?

timmaffett:
i had just started reviewing the existing tests for cursors in the engine and will be adding tests for these changes.   I am away for the weekend, so it may be monday.

dkwingsmt:
This is cool! Let me know once you've added some tests. :)

kevmoo:
@timmaffett – no need to keep merging in main, as long as there are no conflicts. It just wastes CPU time.

timmaffett:
@kevmoo - ok - I will avoid unnecessary syncs.

@kevmoo, @dkwingsmt - the tests have finally been added.   The test first tests all of the existing functionality (all built in system cursors flutter uses) and then tests the new functionality allowing url/-webkit-image-set css styles.

timmaffett:
I updated the branch like the bot asked.

@dkwingsmt  @kevmoo  Is there a way to get the 'needs tests' tag removed now that tests have been added ?

Like I mentioned above, the test I have added now test all of the pre-existing functionality of `_mapKindToCssValue()` as well as the changes that this PR makes to allow custom cursors.

kevmoo:
@timmaffett – done!

timmaffett:
@dkwingsmt Just wanted to ping you once again and remind you that the tests have been added and see if we could move forward on this ?

dkwingsmt:
Sorry, this PR slipped through my mind and I failed to address it. And the tests look great!

But now that I look at it, I think it might not be a good idea to use the "system cursor" method to activate the url cursors, since they're simply not "system".

I think it'd be better if we create a new method to handle this. We can either make it platform-independent, such as 
```
method: "activateUrlCursor"
payload: { "url": "data:image/png;base64[...]==", "pixelRatio": 2.0 }
```

Or a platform-dependent one,
```
method: "web/activateUrlCursor"
payload: { "command": "url(data:image/png;base64[...]==)" }
```

I'm currently leaning toward the platform-specific API because we haven't made up our mind how a unified custom cursor API should look like. What do you think? Also cc @mdebbar

dkwingsmt:
I've read the [custom_mouse_cursor](https://pub.dev/packages/custom_mouse_cursor) plugin, and I greatly appreciate the effort that you and others put on this! I understand it's a feature of high demand. But I still have quite some questions with your reply.

> BUT on all other platforms activateSystemCursor is the way you set the cursor

The plugin's implementation actually kind of contradicts with what you said. In the plugin, desktop platforms already use `setCustomCursor` to set custom cursors, and [it's a special case for the Web](https://github.com/timmaffett/custom_mouse_cursor/blob/975694bfa8d076da03290ed39326b5e2274f2871/lib/custom_mouse_cursor.dart#L1845) to use `activateSystemCursor`. Doesn't it make your code a bit cleaner if all you have to call is `activeCustomCursor`?

> the reason for this is that your define a new system cursor and give it a name (on linux,macos and windows) - and that becomes a new 'system cursor' - you then use activateSystemCursor() to activate it.

When I came up with the method name, the name _system_ cursors comes from the fact that these cursors come with the operating system and are available without extra assets or plugins to be shipped with the app. The term "cursor" alone means the mouse cursor. Since the cursors you're implementing in your plugin are bitmaps loaded from assets, I think they definitely are not system cursors. But it's fine. As I'll say later, what matters the most is not how we define this term, but the readability of APIs.

Also, all other desktop platforms are using the pattern of "first register with a key, then activate with the key". Yet Web is trying to send the entire data URL on every activation, which goes against your statement.

> In practice users would interact with custom cursors how the https://pub.dev/packages/custom_mouse_cursor package defines. This api entry point is not really targeted at end users..

This is the toughest part: Everything added to Flutter is always targeted at end users, and that's I can't just say "sure let's land it since it works" despite I really wish I could. Once we add a new way to call this method channel, Flutter will document it and assume everything might call it. In fact, Flutter even allows people to not use our bindings, but call `ui.PlatformDispatcher` and method channels directly, and we make great effort to ensure those users get reasonably designed APIs and are not broken. Imagine you're calling the method channel directly and you're reading the doc for `activateSystemCursor` and found out that, "Oh the `key` argument can be one of the predefined values but also on Web it can be a CSS expression" - I wouldn't call it ideal.

> So.. with that in mind activateSystemCursor really is the platform independent way that cursors are set. 

With your design proposal, `activateSystemCursor` is not platform independent, because the CSS expression format is only available to Web. 

> We could add a new entry point - but that would be equivalent to if had css decided that you set system cursors with `cursor: pointer;`

I actually agree to this part to some extent. At the end of the day, in terms of functionality, it doesn't matter if we use the same method name of different names. If it's somehow parsable, it will work. Why make it complicated?

But that's why designing APIs is not a matter of functionality, but a matter of readability. Just like how program is for human to read, not for computers. Splitting features of different kinds into different names helps understanding how each is used, in which conditions each API is supported, etc. 

To summarize, ideally, all APIs should be platform independent. Users write Flutter apps to expect they don't have to care about the resulting platform. In reality, this is sometimes impossible. Flutter's style guide requires [avoiding the lowest common denominator](https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#avoid-the-lowest-common-denominator), which also applies to this method channel API.

I understand the frustration of having to go back and forth with adding this API for your plugin. As I said, I really wish I could just give it a green light. In fact, the reason I didn't work on this feature was how complicated and how platform dependent the feature is to design a good enough API. But I do want to land it. I welcome more discussion on how users use this API, since you definitely have more experience than me for actually using it in your apps. We can also chat over [Discord](https://github.com/flutter/flutter/wiki/Chat) which can be more instant.

Hixie:
@timmaffett Thanks for your contributions so far! Is this still something you are interesting in working on and landing? (If not that's totally fine, we can park the PR and let someone else take over when they have the time!)

Hixie:
@timmaffett awesome. I recommend discussing with @dkwingsmt either here or on our [Discord](https://github.com/flutter/flutter/wiki/Chat) to figure out next steps then!

goderbauer:
@timmaffett Just checking in to see if you were able to connect with the person suggested above and if you're still planning on continue working in this.

timmaffett:
@goderbauer I have not had a chance to connect with @dkwingsmt but I will put this on my whiteboard to get to it. 
It has crossed my mind several times but I have been slammed with work.
@dkwingsmt I will make time to go over your review and come up with a proposal to address your comments and concerns.

dkwingsmt:
I have to sincerely apologize for leaving this PR stale but I've just been afraid of reviewing this. I don't have the bandwidth to implement a full API for cross platform custom cursors, but it's also really hard to agree with a platform specific API at the risk of API breakage if we're able to transit to a cross platform one some day. I know you've worked hard on this and honestly, if we agree that Flutter can accept this platform specific API, I don't remember having any problems with the implementation. But approving it is a different responsibility I'm reluctant to take, especially since I can't see myself permitted to prioritize the cursors as my main project in the foreseeable future. I encourage you should seek some other reviewers with the Team Member role, since I don't see myself capable of reviewing it due to review fatigue. Again, thanks for your contribution so far, and I understand whether you continue to push it or not.

Hixie:
Looking at the code, it's hard for me to tell if this is the right approach. The [documentation for the message channel](https://main-api.flutter.dev/flutter/services/SystemChannels/mouseCursor-constant.html) doesn't say what `kind` means, e.g. whether it's platform-specific or whether there's a standard API we're following on all platforms.

I think before we can land something like this we should document what the message channel API is.

christopherfujino:
from critical issue triage: @timmaffett are you still working on this one?

Piinks:
Thank you for your contribution. I'm going to close this PR for now since there are outstanding comments, just to get this off our PR review queue. Please don't hesitate to re-open or submit a new PR if you have the time to address the review comments. Thanks!

PR #41202, opened 2023-04-14
Web - Fix selection jump on Chrome for Android

bleroux:
## Description

This PR fixes cursor jump on Chrome for Android when the user taps in a multiline `TextField`.

Using the following code sample:

<details><summary>Code sample</summary>

```dart

void main() => runApp(const MyApp());

  const MyApp({super.key});

  @override
    return const MaterialApp(
      title: 'Text Field Focus',
      home: MyCustomForm(),
    );
  }
}

// Define a custom Form widget.
class MyCustomForm extends StatelessWidget {
  const MyCustomForm({super.key});

  @override
    return Scaffold(
      appBar: AppBar(
        title: const Text('Text Field Focus'),
      ),
      backgroundColor: Colors.amber,
      body: Padding(
        padding: const EdgeInsets.all(16.0),
        child: TextField(
          decoration: const InputDecoration(
            fillColor: Colors.white,
            filled: true
          ),
          autofocus: true,
          maxLines: 3,
          controller: TextEditingController(text: '1\n2\n3\n4\n'),
        ),
      ),// This trailing comma makes auto-formatting nicer for build methods.
    );
  }
}

```
</details>

On a mobile browser, once the page is loaded, tap after the number 3: 

- Before this PR: the TextField content is automaticaly scrolled and the selection is set after number 1.

https://user-images.githubusercontent.com/840911/232051413-b913f890-6cb1-4c60-92d0-7a3bf74cc688.mov

## Implementation

A multiline `TextField` relies on an HTML `<textarea>` elements. When a tap occurs the selection should be updated from Flutter not by the HTML element itself. 
This PR prevents mouse events on Chrome for Android. Those events conflicts with Flutter selection changes.
Previously, mouse events were only prevented on desktop but they are also emitted on mobile, see https://bugs.chromium.org/p/chromium/issues/detail?id=119216#c11.

## Related Issue

Related to https://github.com/flutter/flutter/issues/124483 (partial fix because the issue is also reproducible on iOS/Safari).

## Tests

Adds 1 test.

### DIFF
--- a/lib/web_ui/test/engine/text_editing_test.dart
+++ b/lib/web_ui/test/engine/text_editing_test.dart
@@ -1943,8 +1943,9 @@ Future<void> testMain() async {
       checkTextAreaEditingState(textarea, '1\n2\n3\n4\n', 8, 8);
 
       // 'mousedown' event should be prevented.
-      final bool prevented = !textarea.dispatchEvent(createDomEvent('Event', 'mousedown'));
-      expect(prevented, true);
+      final DomEvent event = createDomEvent('Event', 'mousedown');
+      textarea.dispatchEvent(event);
+      expect(event.defaultPrevented, isTrue);
 
       hideKeyboard();
     });
--- a/lib/web_ui/test/engine/text_editing_test.dart
+++ b/lib/web_ui/test/engine/text_editing_test.dart
@@ -2377,7 +2377,6 @@ Future<void> testMain() async {
       expect(textArea.selectionEnd, 2);
     });
 
-
     test('Configure input element editing state for a flipped base and extent',
         () {
       final DomHTMLInputElement input =
--- a/lib/web_ui/test/engine/text_editing_test.dart
+++ b/lib/web_ui/test/engine/text_editing_test.dart
@@ -1896,6 +1896,59 @@ Future<void> testMain() async {
       hideKeyboard();
     });
 
+    test('prevent mouse events on Android', () {
+      // Regression test for https://github.com/flutter/flutter/issues/124483.
+      debugOperatingSystemOverride = OperatingSystem.android;
+      debugBrowserEngineOverride = BrowserEngine.blink;
+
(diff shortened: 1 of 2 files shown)


lib/web_ui/test/engine/text_editing_test.dart
```diff
@@ -2324,6 +2377,7 @@ Future<void> testMain() async {
       expect(textArea.selectionEnd, 2);
     });
 
+
```
kevmoo, line 2380: extra new line

kevmoo:
Thanks for doing the tweaks here!

bleroux:
@mdebbar @kevmoo If you have some bandwidth, thank you to review this PR.

mdebbar (approved):
The change looks good to me, thanks for the contribution!

But I don't think it solves the linked issue. It only hides the symptom. Can this be done in `IOSTextEditingStrategy` too?

bleroux:
@mdebbar Thank you for the review!
 
> But I don't think it solves the linked issue. It only hides the symptom. 

Do you have in mind another way to fix this? Because it might help to fix the iOS side.

>Can this be done in `IOSTextEditingStrategy` too?

This solution is not applicable on the iOS side because `IOSTextEditingStrategy` relies on DOM `click`event to tackle tap/longPress behaviors: 

https://github.com/flutter/engine/blob/40136ed00bc901b9f7c1752fe6315a69b329a429/lib/web_ui/lib/src/engine/text_editing/text_editing.dart#L1595-L1625

mdebbar:
> Do you have in mind another way to fix this? Because it might help to fix the iOS side.

Well, that's a tricky issue that we've had for a while (see https://github.com/flutter/flutter/issues/99918 and https://github.com/flutter/flutter/issues/70841 for example). The only solution for it is to overhaul our text editing system (discussed in https://github.com/flutter/flutter/issues/120613).

PR #41203, opened 2023-04-14
Do not run mac unit tests on macmini7,1

zanderso:
Impeller tests are hanging on Intel Haswell Iris Graphics 5100 (8086:0a2e)

### DIFF
--- a/.ci.yaml
+++ b/.ci.yaml
@@ -338,6 +338,7 @@ targets:
         {"download_emsdk": true}
       add_recipes_cq: "true"
       build_host: "true"
+      mac_model: "Macmini8,1"
     timeout: 75
 
   - name: Mac mac_android_aot_engine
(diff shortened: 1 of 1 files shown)


zanderso:
Mac Host Engine on https://github.com/flutter/engine/pull/41194 has failed 3 or 4 times in impeller_unittests on the 5100 GPU.

.ci.yaml
```diff
@@ -338,6 +338,7 @@ targets:
         {"download_emsdk": true}
       add_recipes_cq: "true"
       build_host: "true"
+      mac_model: "Macmini8,1"
```
dnfield, line 341: I think this has to go into the json files where we specify the dimensions. Does this work?

dnfield (approved):
LGTM but I'm not sure if this actually work - it shouldn't hurt anything though.

CaseyHillers (approved):

.ci.yaml
CaseyHillers: https://cs.opensource.google/flutter/cocoon/+/main:app_dart/lib/src/model/ci_yaml/target.dart;l=40

This does get run, but it'll only change the orchestrator vm. This is useful as the spawned tasks should inherit the properties from the orchestrator.

PR #41207, opened 2023-04-14
[web] Don't run goldctl init more than once

mdebbar:
From the logs reported in https://github.com/flutter/flutter/issues/124864 I noticed we are making multiple calls to `goldctl init` which could be causing some race conditions with the `goldctl imgtest add` calls.

This PR makes sure we only call `goldctl init` once.


testing/skia_gold_client/lib/skia_gold_client.dart
```diff
@@ -495,5 +481,12 @@ class SkiaGoldClient {
   }
 }
 
+Future<void>? _oneResult;
```
eyebrowsoffire, line 484: I don't think we can make this a global, it needs to be per `SkiaGoldClient`. We have separate gold clients per suite now.

eyebrowsoffire (changes requested):

testing/skia_gold_client/lib/skia_gold_client.dart
mdebbar: Done.

Nice catch! I've been contemplating making it per client instead of global but didn't find a good reason to do that. Totally forgot that we run multiple suites at the same time now.

testing/skia_gold_client/lib/skia_gold_client.dart
```diff
@@ -55,9 +55,12 @@ class SkiaGoldClient {
   String get _keysPath => path.join(workDirectory.path, 'keys.json');
   String get _failuresPath => path.join(workDirectory.path, 'failures.json');
 
-  /// Indicates whether the `goldctl` tool has been initialized for the current
-  /// test context.
-  bool _isInitialized = false;
+  Future<void>? _initResult;
+  Future<void> _initOnce(Future<void> Function() callback) async {
```
eyebrowsoffire, line 59: nit: the function body doesn't need to be marked as async, as you're not actually using `await` semantics or anything here, you're just directly returning something that is already a future.

eyebrowsoffire (approved):
LGTM! One small unimportant nit

testing/skia_gold_client/lib/skia_gold_client.dart
mdebbar: Done

PR #40746, opened 2023-03-29
[Android] Send connectionClosed message when keyboard becomes invisible to ensure framework focus state is correct.

ksballetba:
Fix:
- https://github.com/flutter/flutter/issues/123523

### Before this patch:

https://user-images.githubusercontent.com/30322203/228413196-29c57bb0-3220-495b-9e73-f58777de440f.mp4

### After this patch:

https://user-images.githubusercontent.com/30322203/228413249-fc06f49d-6579-4476-9788-90f12a53b8c3.mp4

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [Flutter Style Guide] and the [C++, Objective-C, Java style guides].
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I added new tests to check the change I am making or feature I am adding, or Hixie said the PR is test-exempt. See [testing the engine] for instructions on writing and running engine tests.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I signed the [CLA].
- [x] All existing and new tests are passing.

PR #41211, opened 2023-04-14
Revert "[Android] Send connectionClosed message when keyboard becomes invisible to ensure framework focus state is correct."

CaseyHillers:
Reverts flutter/engine#40746

Googler bug: b/278174021

Failing on

```
shell/platform/android/io/flutter/plugin/editing/TextInputPlugin.java:239: Error: This method should only be accessed from tests or within private scope [VisibleForTests]
      imeSyncCallback.remove();
```

### DIFF
--- a/shell/platform/android/io/flutter/embedding/engine/systemchannels/TextInputChannel.java
+++ b/shell/platform/android/io/flutter/embedding/engine/systemchannels/TextInputChannel.java
@@ -367,14 +367,6 @@ public class TextInputChannel {
         "TextInputClient.performPrivateCommand", Arrays.asList(inputClientId, json));
   }
 
-  /** Instructs Flutter to execute a "onConnectionClosed" action. */
-  public void onConnectionClosed(int inputClientId) {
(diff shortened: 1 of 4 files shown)


CaseyHillers:
\cc @ksballetba as FYI, this appears to have a lint issue where test code is being used. I'm sending a revert to start the presubmits while I investigate if there's any lints we need to opt of.

zanderso (approved):

reidbaker (approved):

CaseyHillers:
@reidbaker do you know if there's any lints we need to add to the engine to prevent issues like this from landing?

reidbaker:
> @reidbaker do you know if there's any lints we need to add to the engine to prevent issues like this from landing?

Not off hand but if you file a bug against me we can get it looked at.

ksballetba:
> \cc @ksballetba as FYI, this appears to have a lint issue where test code is being used. I'm sending a revert to start the presubmits while I investigate if there's any lints we need to opt of.

It looks like I didn't change this code, this lint problem exists before my code change.
And I found same lint issue in TextInputPlugin.

image

PR #38261, opened 2022-12-13
Run Mac Host clang-tidy on 12 cores

jmagman:
Prod bots are all 12 cores:
https://chromium-swarm.appspot.com/botlist?c=id&c=task&c=cores&c=os&c=status&d=asc&f=pool%3Aluci.flutter.prod&f=os%3AMac-12&f=cpu%3Ax86&f=device_type%3Anone&k=cores&s=id

Try bots are mostly 12 cores:
https://chromium-swarm.appspot.com/botlist?c=id&c=task&c=cores&c=os&c=status&d=asc&f=os%3AMac-12&f=cpu%3Ax86&f=device_type%3Anone&f=pool%3Aluci.flutter.try&k=pool&s=id

Staging is 50-50 (would be used if this bot ever marked flaky):
https://chromium-swarm.appspot.com/botlist?c=id&c=task&c=cores&c=os&c=status&d=asc&f=os%3AMac-12&f=cpu%3Ax86&f=device_type%3Anone&f=pool%3Aluci.flutter.staging&k=pool&s=id

Fixes https://github.com/flutter/flutter/issues/116931

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [Flutter Style Guide] and the [C++, Objective-C, Java style guides].
- [x] I listed at least one issue that this PR fixes in the description above.
- [ ] I added new tests to check the change I am making or feature I am adding, or Hixie said the PR is test-exempt. See [testing the engine] for instructions on writing and running engine tests.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I signed the [CLA].
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/flutter/wiki/Tree-hygiene#overview
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[C++, Objective-C, Java style guides]: https://github.com/flutter/engine/blob/main/CONTRIBUTING.md#style
[testing the engine]: https://github.com/flutter/flutter/wiki/Testing-the-engine
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[...]

PR #41214, opened 2023-04-14
Remove cores dimension from Mac Host clang-tidy

jmagman:
Follow up to #41183.  Remove the `cores` dimensions so this builder can run on any arm machine which currently all have 8 cores but there's no reason to specify now that the arch is arm.

Introduced in #38261 to avoid 4-core Intel machines.


keyonghan (approved):
LGTM

zanderso (approved):

PR #41222, opened 2023-04-14
Remove jazzy and Xcode properties from impeller-cmake-example builder

jmagman:
This builder doesn't use or install jazzy or Xcode, remove the properties from the config.

Passing without these properties: https://ci.chromium.org/p/flutter/builders/try/Mac%20impeller-cmake-example/14

### DIFF
--- a/.ci.yaml
+++ b/.ci.yaml
@@ -448,7 +448,6 @@ targets:
         ]
 
   - name: Mac impeller-cmake-example
-    bringup: true
     recipe: engine_v2/engine_v2
     timeout: 60
     properties:
--- a/.ci.yaml
+++ b/.ci.yaml
@@ -448,6 +448,7 @@ targets:
         ]
 
   - name: Mac impeller-cmake-example
+    bringup: true
     recipe: engine_v2/engine_v2
     timeout: 60
     properties:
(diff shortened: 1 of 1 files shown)


.ci.yaml
```diff
@@ -455,12 +455,6 @@ targets:
       release_build: "true"
       cpu: arm64
       config_name: mac_impeller_cmake_example
-      $flutter/osx_sdk : >-
-        { "sdk_version": "14a5294e" }
```
zanderso, line 459: It uses the macOS SDK: https://github.com/flutter/engine/blob/main/ci/builders/mac_impeller_cmake_example.json#L17. Will that still be there? You can try this change in CI by temporarily commenting out `bringup: true`.

jmagman: Hm, the builders inherit dimensions from their orchestrator?  That's unexpected...  
It doesn't seem to be inheriting the dependencies. https://ci.chromium.org/p/flutter/builders/staging/Mac%20Staging%20Engine%20Drone/87661

This is the first time I've played with the v2 stuff.
Trying with `bringup` to see what happens.

zanderso: I'm not sure. I say let's try this change in presubmit, and if it works, let's ship it.

jmagman: It worked: https://ci.chromium.org/p/flutter/builders/try/Mac%20impeller-cmake-example/14

zanderso (approved):

PR #41231, opened 2023-04-15
[Windows] Don't block raster thread until v-blank

loic-sharma:
Currently the Windows raster thread blocks until the v-blank. As a result, rasterization takes ~17ms on Windows.

### Results

I hovered on and off the counter app's button repeatedly.

#### Before 
Rasterization takes ~17ms consistently:

Pasted image 20230414180144

#### After
After this change, rasterization takes less than 1ms consistently:
Pasted image 20230414180011

## Pre-launch Checklist

- [x] I signed the [CLA].


shell/platform/windows/flutter_windows_view.cc
```diff
@@ -81,8 +81,8 @@ uint32_t FlutterWindowsView::GetFrameBufferId(size_t width, size_t height) {
   if (resize_target_width_ == width && resize_target_height_ == height) {
     // Platform thread is blocked for the entire duration until the
     // resize_status_ is set to kDone.
-    engine_->surface_manager()->ResizeSurface(GetRenderTarget(), width, height);
-    engine_->surface_manager()->MakeCurrent();
```
loic-sharma, line 85: This line is now unneeded as resizing the window makes the surface current to update the wait interval.

shell/platform/windows/windows_lifecycle_manager.cc
```diff
@@ -48,7 +48,7 @@ bool WindowsLifecycleManager::WindowProc(HWND hwnd,
     // send a request to the framework to see if the app should exit. If it
     // is, we re-dispatch a new WM_CLOSE message. In order to allow the new
     // message to reach other delegates, we ignore it here.
-    case WM_CLOSE:
+    case WM_CLOSE: {
```
loic-sharma, line 51: The braces were added as the C++ compiler complained about potential jumping to uninitialized values

shell/platform/windows/angle_surface_manager.cc
```diff
@@ -300,6 +305,24 @@ EGLSurface AngleSurfaceManager::CreateSurfaceFromHandle(
...
 
+void AngleSurfaceManager::SetVSyncEnabled(bool enabled) {
+  if (eglMakeCurrent(egl_display_, render_surface_, render_surface_,
+                     egl_context_) != EGL_TRUE) {
+    LogEglError("Unable to make surface current to update the swap interval");
+    return;
+  }
+
+  // OpenGL swap intervals can be used to prevent screen tearing.
+  // If enabled, the raster thread blocks until the v-blank.
```
yaakovschectman, line 316: Does this mean that we will still block until the v-blank when this parameter is true, i.e. when composition is not enabled? And if so, is that what we intend?

loic-sharma: Yes and yes. Screen tearing is possible if a Windows 7 user does not have DWM composition enabled. The app should synchronize with the vsync to prevent screen tearing.

yaakovschectman (approved):
LGTM then

loic-sharma:
This change is blocked on this Dart bug: https://github.com/dart-lang/sdk/issues/52071

MiinoY:
@loic-sharma 
Hello! This commit is very helpful to my issue! Thanks!
We have met the similar problem and this commit fix it. But i'm wondering that why Android doesn't have the same issue?  On Android, I have forcibly set the `eglSwapInterval` to 1, but the SwapBuffers operation on the raster thread is still very fast(1~3ms), and the same case on Windows without this commit, the SwapBuffers time will reach 16ms.
Do you know any reasons for this difference? looking forward to your reply! Thank you!

PR #41237, opened 2023-04-15
focus SkiaGPUObject wrappers on only those objects that need the protection

flar:
With this PR we no longer need to hold DisplayLists in GPUObject wrappers and they can be disposed instantly instead of queueing on the Unref thread.

This will definitely be a win for Impeller as none of the objects used in a frame now require queueing, but the performance impact on apps running on top of skia is less clear if they depend on a lot of images inside their DisplayLists that still need to be queued to be freed. After getting further in the work, it looks like only decoded images need to use the protected DlImage wrappers and most of those should survive many frames before they are disposed. That should hopefully leave very few unrefs happening per frame.

~There are 3 unit tests in `shell_unittests.cc` and `embedder_metal_unittests.mm` that are now GSKIP'd as they now invoke code that needs a fully initialized UIDartState in order to protect their images. I will look into fixing the tests and/or making the code they invoke provide protection without relying on UIDartState.~ (This looks to be fixed in the latest commit by simply not creating DlImageGPUs all over the source base and simply catching only those that end up in UI data structures. There is actually existing code in one of the modules that feeds ui.Image with an answer to wrap the image in a DlImageGPU if it has a skia image anyway, so most of these additional uses of DlImageGPU that were having trouble getting the Skia unref queue just didn't need it anyway.)


flar:
It looks like there are more unit tests that run into the same problem.

I'm thinking the right fix for that is to find a way for the code in `shell/` to get access to the SkiaUnrefQueue and make its own SkiaGPUObject wrappers without having to rely on the UIDartState code in lib/ui...

flar:
The use of SkiaGPUObject wrappers is now limited to the following files that need to wrap an image for use in a UI thread:

```
% git grep skia_gpu_object     
ci/licenses_golden/excluded_files:../../../flutter/flow/skia_gpu_object_unittests.cc
ci/licenses_golden/licenses_flutter:ORIGIN: ../../../flutter/flow/skia_gpu_object.h + ../../../flutter/LICENSE
ci/licenses_golden/licenses_flutter:FILE: ../../../flutter/flow/skia_gpu_object.h
flow/BUILD.gn:    "skia_gpu_object.h",
flow/BUILD.gn:      "skia_gpu_object_unittests.cc",
flow/skia_gpu_object_unittests.cc:#include "flutter/flow/skia_gpu_object.h"
lib/ui/io_manager.h:#include "flutter/flow/skia_gpu_object.h"
lib/ui/painting/display_list_image_gpu.h:#include "flutter/flow/skia_gpu_object.h"
shell/common/shell_io_manager.h:#include "flutter/flow/skia_gpu_object.h"
```

```
% git grep SkiaGPUObject  
flow/skia_gpu_object.h:class SkiaGPUObject {
flow/skia_gpu_object.h:  SkiaGPUObject() = default;
flow/skia_gpu_object.h:  SkiaGPUObject(sk_sp<SkiaObjectType> object, fml::RefPtr<SkiaUnrefQueue> queue)
flow/skia_gpu_object.h:  SkiaGPUObject(SkiaGPUObject&&) = default;
flow/skia_gpu_object.h:  ~SkiaGPUObject() { reset(); }
flow/skia_gpu_object.h:  SkiaGPUObject& operator=(SkiaGPUObject&&) = default;
flow/skia_gpu_object.h:  FML_DISALLOW_COPY_AND_ASSIGN(SkiaGPUObject);
flow/skia_gpu_object_unittests.cc:    SkiaGPUObject<TestSkObject> sk_object(std::move(object), unref_queue());
flow/skia_gpu_object_unittests.cc:    // Verify that the default SkiaGPUObject dtor queues and unref.
flow/skia_gpu_object_unittests.cc:  SkiaGPUObject<TestSkObject> sk_object(
flow/skia_gpu_object_unittests.cc:  SkiaGPUObject<TestSkObject> sk_object(
lib/ui/painting/display_list_image_gpu.cc:sk_sp<DlImageGPU> DlImageGPU::Make(SkiaGPUObject<SkImage> image) {
lib/ui/painting/display_list_image_gpu.cc:DlImageGPU::DlImageGPU(SkiaGPUObject<SkImage> image)
lib/ui/painting/display_list_image_gpu.h:  static sk_sp<DlImageGPU> Make(SkiaGPUObject<SkImage> image);
lib/ui/painting/display_list_image_gpu.h:  SkiaGPUObject<SkImage> image_;
lib/ui/painting/display_list_image_gpu.h:  explicit DlImageGPU(SkiaGPUObject<SkImage> image);
lib/ui/painting/image_decoder_skia.cc:static SkiaGPUObject<SkImage> UploadRasterImage(
lib/ui/painting/image_decoder_skia.cc:  SkiaGPUObject<SkImage> result;
lib/ui/painting/image_decoder_skia.cc:          SkiaGPUObject<SkImage> image, fml::tracing::TraceFlow flow) {
```

flar:
And only a couple of files need to use the enqueueing DlImageGPU as seen here:

```
% git grep DlImageGPU
lib/ui/painting/display_list_image_gpu.cc: ...
lib/ui/painting/display_list_image_gpu.h: ...
lib/ui/painting/image_decoder_skia.cc:              callback(DlImageGPU::Make(std::move(image)));
lib/ui/painting/multi_frame_codec.cc:  return DlImageGPU::Make({skImage, std::move(unref_queue)});
lib/ui/painting/picture.cc:              DlImageGPU::Make({image->skia_image(), std::move(unref_queue)});
shell/common/snapshot_controller.h:  // be converted to a DlImageGPU if it is to be handed back to the UI
shell/common/snapshot_controller_skia.cc:  // It is up to the caller to create a DlImageGPU version of this image
```

Note that the last 2 references outside of lib/ui are just comments that those lines of code are not responsible for protecting the images (and, in fact, the code in `picture.cc` that you see here is doing just that on their behalf).

PR #41259, opened 2023-04-16
[web:canvaskit] clean up the rest of skia_object_cache usages

yjbanov:
This should go after https://github.com/flutter/engine/pull/41230.

Removes the remaining usages of `skia_object_cache.dart` and deletes it.

Also fixes https://github.com/flutter/flutter/issues/86632


hterkelsen (approved):
Woohoo!

PR #41201, opened 2023-04-14
[iOS] Fix random jitter in keyboard animation and correct spring animation target time.

luckysmg:
### Fix
- https://github.com/flutter/flutter/issues/120555

## Locally test for reviewers

flutter framework ref: `master`

<details>

<summary>Demo code</summary>

```dart

void main() {
  runApp(const MyApp());
}


class MyApp extends StatelessWidget {
  const MyApp({Key? key}) : super(key: key);

  @override
  Widget build(BuildContext context) {
    return MaterialApp(
      title: 'Flutter Demo',
      theme: ThemeData(
        colorScheme: ColorScheme.fromSeed(seedColor: Colors.deepPurple),
        useMaterial3: true,
      ),
      home: DemoScreen(),
    );
  }
}

class DemoScreen extends StatefulWidget {
  const DemoScreen({Key? key}) : super(key: key);

  @override
  State<DemoScreen> createState() => _DemoScreenState();
}

class _DemoScreenState extends State<DemoScreen> with TickerProviderStateMixin {
  final TextEditingController _messageController = TextEditingController();
  final FocusNode _focusNode = FocusNode();
  late final _messageNotifier = ValueNotifier(_messageController.text.isEmpty);
  bool hasFocus = false;

  @override
  Widget build(BuildContext context) {
    return GestureDetector(
      onTap: () {
        _focusNode.unfocus();
      },
      child: Scaffold(
        appBar: AppBar(
          title: const Text('Chat screen'),
        ),
        body: Column(
          children: <Widget>[
            const Spacer(),
            const Text(
              'keyboard animation',
              style: TextStyle(fontSize: 30),
            ),
            const Spacer(),
            Container(
              padding: EdgeInsets.only(
                bottom: MediaQuery.paddingOf(context).bottom + 10,
                top: 6.0,
                left: 15.0,
                right: 15.0,
              ),
              constraints: const BoxConstraints(
                minHeight: 36,
              ),
              child: Row(
                crossAxisAlignment: CrossAxisAlignment.end,
                children: [
                  Expanded(
                    child: DecoratedBox(
                      decoration: BoxDecoration(
                        color: Colors.white,
                        borderRadius: BorderRadius.circular(19),
                        border: Border.all(
                          width: 0.5,
                          color: const Color(0xffCFCFCF),
                        ),
                      ),
                      child: Row(
                        crossAxisAlignment: CrossAxisAlignment.end,
                        children: [
                          Container(
[...]

PR #41268, opened 2023-04-17
[iOS] Fix jitter keyboard animation by updating metrics value in onFrameRasterized callback.

luckysmg:

## Pre-launch Checklist

- [ ] I signed the [CLA].

### DIFF
--- a/shell/platform/darwin/ios/framework/Source/FlutterViewController.mm
+++ b/shell/platform/darwin/ios/framework/Source/FlutterViewController.mm
@@ -1609,15 +1609,16 @@ static flutter::PointerData::DeviceKind DeviceKindFromTouchType(UITouch* touch) 
         flutterViewController.get()->_viewportMetrics.physical_view_inset_bottom =
             flutterViewController.get()
                 .keyboardAnimationView.layer.presentationLayer.frame.origin.y;
-        [flutterViewController updateViewportMetrics];
+        [flutterViewController keyboardAnimationUpdateViewportMetricsInFrameRasterizedCallback];
       }
     } else {
-      fml::TimeDelta timeElapsed = recorder.get()->GetVsyncTargetTime() -
+      fml::TimeDelta frameInterval = recorder->GetVsyncTargetTime() - recorder->GetVsyncStartTime();
+      fml::TimeDelta timeElapsed = recorder.get()->GetVsyncTargetTime() + frameInterval -
                                    flutterViewController.get().keyboardAnimationStartTime;
 
       flutterViewController.get()->_viewportMetrics.physical_view_inset_bottom =
           [[flutterViewController keyboardSpringAnimation] curveFunction:timeElapsed.ToSecondsF()];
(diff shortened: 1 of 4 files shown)


luckysmg:
@cyanglaz Re: https://github.com/flutter/engine/pull/41201#discussion_r1167109844

From the debugging this change does help with reducing jitter. But I m not sure current patch is correct and match the thought you mentioned in comment. Would you mind giving me some advice? ^_^

shell/platform/darwin/ios/framework/Source/FlutterViewController.mm
```diff
@@ -1630,6 +1631,26 @@ - (void)setupKeyboardAnimationVsyncClient {
...
+        [(FlutterViewController*)weakSelf.get() retain]);
+    if (!flutterViewController) {
+      return;
+    }
+    [flutterViewController updateViewportMetrics];
+  };
+
+  flutter::Shell& shell = [_engine.get() shell];
+  fml::closure onFrameRasterizedCallback = [&shell, platformTask] {
+    shell.GetTaskRunners().GetPlatformTaskRunner()->PostTask(platformTask);
```
cyanglaz, line 1649: I believe this also needs to be synchronized. The rasterizer needs to wait until the platform task to finish before starting next frame.

shell/platform/darwin/ios/framework/Source/FlutterViewController.mm
```diff
@@ -1630,6 +1631,26 @@ - (void)setupKeyboardAnimationVsyncClient {
...
+  };
+  shell.AddOnFrameRasterizedCallback(onFrameRasterizedCallback);
```
cyanglaz, line 1651: I think add the callback here might be a frame late? 
What I was thinking was the callback will be some sort of delegate method and runs every frame. The delegate then decide if the method body is a no-op (when there is no keyboard animation), or synchronously runs the keyboard animation on platform thread.

luckysmg:
I made a patch to add callback in `Shell` (Still have some issues but can test at least)

https://user-images.githubusercontent.com/49340347/232720936-295b9277-5970-4f86-86c5-79bee415d26d.MP4

Seems has some delay with system's keyboard. Maybe we should still use vsync to track value so that we can get the correct presentation time by adding one frame interval time.

As for https://github.com/flutter/engine/pull/41201
I think it is safe just keep the delay less than one vsync time. Maybe Just < 8ms on 120HZ devices is ok.... Just keep the task run after the vsync process callback and before the next vsync callback...

delfme:
Was the master updated meantime the fix is landed? Asking coz I updated to latest master and now keyboard has a bit different behaviour. I now see a delay (instead of the previous instant bump to end of run). Random jitters issue remained the same. Tested on iphone 13 pro ios 16.3.1

https://user-images.githubusercontent.com/53510751/236608004-15dce0d6-7e36-49e2-93a8-f017e3837248.MOV

PR #41279, opened 2023-04-17
Reland  "Migrate mac_host_engine to engine v2 builds."

godofredoc:
Reverts flutter/engine#41233

The issue this PR was reverted was not replicable. On top of that we added extra precautions to ensure global generators are using left over binaries from out folder. https://flutter-review.googlesource.com/c/recipes/+/41960

### DIFF
--- a/.ci.yaml
+++ b/.ci.yaml
@@ -332,6 +332,7 @@ targets:
     timeout: 60
 
   - name: Mac Host Engine
+    bringup: true
     recipe: engine/engine
     properties:
       gclient_variables: >-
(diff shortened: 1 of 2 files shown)


zanderso:
> Reverts #41233
> 
> The issue this PR was reverted was not replicable. On top of that we added extra precautions to ensure global generators are using left over binaries from out folder. https://flutter-review.googlesource.com/c/recipes/+/41960

Assume you mean "are *not* using left over binaries".

zanderso (approved):

PR #41285, opened 2023-04-18
[Impeller] Remove glyph pixel rounding during text frame conversion

bdero:
A tale of glyph alignment problems:

- This rounding was originally introduced to solve: https://github.com/flutter/flutter/issues/121531. This was to align all of the glyphs to the pixel grid during text frame conversion.
- However, doing so introduced: https://github.com/flutter/flutter/issues/124310.
- When working on fixing https://github.com/flutter/flutter/issues/121887, I reworked the UV handling/rounding in text_contents in a way that normalizes the glyph rounding per-textframe, which actually solves both problems.

..and so we can just get rid of this early rounding now.

Before:
Screenshot 2023-04-17 at 6 36 20 PM

After:
Screenshot 2023-04-17 at 6 33 09 PM

All of the glyphs in a text frame still snap together:

https://user-images.githubusercontent.com/919017/232648307-cc298fd9-603d-4cd6-a927-c44834225303.mov


jonahwilliams (approved):
LGTM

PR #41290, opened 2023-04-18
[Impeller] faster glyph atlas generation by removing data copies

gaaclarke:
fixes https://github.com/flutter/flutter/issues/124996

This is the best I could come up with easily.  It doesn't give as good of a result as I'd hoped.  I measured it with a simple microbenchmark derived from one of the unit tests and it gave a 6% decrease in time ( 241.314us vs 257.626us) on macos release builds (run with rosetta).

Improvements:
1) Removed the copying of the std::set to an std::vector
1) Uses references instead of copying FontGlyphPairs out of collections in a few places
1) Holds new glyphs as a vector of references to the set instead of copying all the FontGlyphPairs
1) Deletes more lines of code than it adds

## the benchmark

```diff
diff --git a/impeller/typographer/typographer_unittests.cc b/impeller/typographer/typographer_unittests.cc
index 01a11d494c..1b99afa699 100644
--- a/impeller/typographer/typographer_unittests.cc
+++ b/impeller/typographer/typographer_unittests.cc
@@ -2,6 +2,7 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+#include <chrono>
 #include "flutter/testing/testing.h"
 #include "impeller/playground/playground_test.h"
 #include "impeller/typographer/backends/skia/text_frame_skia.h"
@@ -149,23 +150,29 @@ TEST_P(TypographerTest, GlyphAtlasWithLotsOfdUniqueGlyphSize) {
       sk_font);
   ASSERT_TRUE(blob);
 
-  TextFrame frame;
-  size_t count = 0;
-  TextRenderContext::FrameIterator iterator = [&]() -> const TextFrame* {
-    if (count < 8) {
-      count++;
-      frame = TextFrameFromTextBlob(blob, 0.6 * count);
-      return &frame;
-    }
-    return nullptr;
-  };
-  auto atlas = context->CreateGlyphAtlas(GlyphAtlas::Type::kAlphaBitmap,
-                                         atlas_context, iterator);
-  ASSERT_NE(atlas, nullptr);
-  ASSERT_NE(atlas->GetTexture(), nullptr);
-
-  ASSERT_EQ(atlas->GetTexture()->GetSize().width * 2,
-            atlas->GetTexture()->GetSize().height);
+  auto beg = std::chrono::high_resolution_clock::now();
+  int count = 10000;
+  for (int i = 0; i < count; ++i) {
+    TextFrame frame;
+    size_t count = 0;
+    TextRenderContext::FrameIterator iterator = [&]() -> const TextFrame* {
+      if (count < 8) {
+        count++;
+        frame = TextFrameFromTextBlob(blob, 0.6 * count);
+        return &frame;
+      }
+      return nullptr;
+    };
+    auto atlas = context->CreateGlyphAtlas(GlyphAtlas::Type::kAlphaBitmap,
+                                           atlas_context, iterator);
+    ASSERT_NE(atlas, nullptr);
+    ASSERT_NE(atlas->GetTexture(), nullptr);
+    ASSERT_EQ(atlas->GetTexture()->GetSize().width * 2,
+              atlas->GetTexture()->GetSize().height);
+  }
+  auto end = std::chrono::high_resolution_clock::now();
+  auto duration = std::chrono::duration_cast<std::chrono::microseconds>(end - beg);
+  FML_LOG(ERROR) << "Elapsed Time: " << static_cast<double>(duration.count())/count << "us";
 }

```

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/impeller/typographer/backends/skia/text_render_context_skia.cc
+++ b/impeller/typographer/backends/skia/text_render_context_skia.cc
@@ -22,7 +22,8 @@
 
 namespace impeller {
 
-using FontGlyphPairRefVector = std::vector<std::reference_wrapper<const FontGlyphPair>>;
+using FontGlyphPairRefVector =
+    std::vector<std::reference_wrapper<const FontGlyphPair>>;
 
 std::unique_ptr<TextRenderContext> TextRenderContext::Create(
     std::shared_ptr<Context> context) {
--- a/impeller/typographer/backends/skia/text_render_context_skia.cc
+++ b/impeller/typographer/backends/skia/text_render_context_skia.cc
@@ -22,6 +22,8 @@
 
 namespace impeller {
 
+using FontGlyphPairRefVector = std::vector<std::reference_wrapper<const FontGlyphPair>>;
+
 std::unique_ptr<TextRenderContext> TextRenderContext::Create(
     std::shared_ptr<Context> context) {
   // There is only one backend today.
@@ -95,7 +97,7 @@ static size_t PairsFitInAtlasOfSize(
 
 static bool CanAppendToExistingAtlas(
     const std::shared_ptr<GlyphAtlas>& atlas,
-    const FontGlyphPair::Vector& extra_pairs,
+    const FontGlyphPairRefVector& extra_pairs,
     std::vector<Rect>& glyph_positions,
     ISize atlas_size,
     const std::shared_ptr<GrRectanizer>& rect_packer) {
@@ -318,7 +320,7 @@ static void DrawGlyph(SkCanvas* canvas,
 
 static bool UpdateAtlasBitmap(const GlyphAtlas& atlas,
                               const std::shared_ptr<SkBitmap>& bitmap,
-                              const FontGlyphPair::Vector& new_pairs) {
+                              const FontGlyphPairRefVector& new_pairs) {
   TRACE_EVENT0("impeller", __FUNCTION__);
   FML_DCHECK(bitmap != nullptr);
 
@@ -333,7 +335,7 @@ static bool UpdateAtlasBitmap(const GlyphAtlas& atlas,
 
   bool has_color = atlas.GetType() == GlyphAtlas::Type::kColorBitmap;
 
-  for (const auto& pair : new_pairs) {
+  for (const FontGlyphPair& pair : new_pairs) {
     auto pos = atlas.FindFontGlyphBounds(pair);
     if (!pos.has_value()) {
       continue;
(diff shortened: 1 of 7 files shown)


impeller/typographer/backends/skia/text_render_context_skia.cc
```diff
@@ -141,8 +134,9 @@ static bool CanAppendToExistingAtlas(
   return true;
 }
 
-static ISize OptimumAtlasSizeForFontGlyphPairs(
-    const FontGlyphPair::Vector& pairs,
+namespace {
```
jonahwilliams, line 137: what does putting in a namespace like this do?

impeller/typographer/font.h
```diff
@@ -50,7 +50,7 @@ class Font : public Comparable<Font> {
 
   Font(std::shared_ptr<Typeface> typeface, Metrics metrics);
 
-  ~Font();
+  virtual ~Font();
```
jonahwilliams, line 53: also question here: why do we need a virtual destructor?

impeller/typographer/backends/skia/text_render_context_skia.cc
gaaclarke: It should have no functional difference in this case.  I was playing around with making that function templated and looks like I left that change in there by happenstance.

impeller/typographer/font.h
gaaclarke: Yep, because we have virtual methods (via inheritance).

jonahwilliams: but this isn't being implemented anywhere

gaaclarke: You are right.  It's my understanding that this should be fine.  The reason I added it was that when using `std::shared_ptr<Font>` it complained about it not being virtual.  I'm not sure why.  I would have only expected that to be a problem if we are calling delete on a `Comparable<Font>*`.  I'll remove this.

gaaclarke: done

impeller/typographer/backends/skia/text_render_context_skia.cc
```diff
@@ -81,8 +73,9 @@ static size_t PairsFitInAtlasOfSize(
   glyph_positions.clear();
   glyph_positions.reserve(pairs.size());
 
-  for (size_t i = 0; i < pairs.size(); i++) {
-    const auto& pair = pairs[i];
+  size_t i = 0;
+  for (auto it = pairs.begin(); it != pairs.end(); ++i, ++it) {
+    const auto& pair = *it;
```
jonahwilliams, line 78: I don't really love this. Does it make a significant improvement on the generated code?

jonahwilliams (approved):
LGTM with nits

impeller/typographer/backends/skia/text_render_context_skia.cc
gaaclarke: It was required because you can't index into `std::unordered_set`.  We could use a foreach loop but it isn't really cleaner imo.

Previously we were copying all of the data in the `std::unordered_set` to make a `std::vector` just so we could cleanly index into the data.  But since the order of iteration of an `std::unordered_set` is deterministic, we don't really need to copy the data just to index into it.

jonahwilliams: makes sense

PR #41315, opened 2023-04-18
[Impeller] Provide an opaque interface for color sources in Aiks

bdero:
Part of https://github.com/flutter/flutter/issues/125017.

Pulls ColorSource construction out of the dispatcher and into Aiks.

Cleans up a few design warts we've accumulated over time:
- Removes special exceptions for solid colors in `Paint`.
- Makes `SolidColorContents` a `ColorSourceContents`.
- Removes paint ref capture for kImage.

FYI @iamralpht

### DIFF
--- a/impeller/aiks/color_source.h
+++ b/impeller/aiks/color_source.h
@@ -1,0 +1,100 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#pragma once
+
+#include <functional>
+#include <memory>
+#include <vector>
+
+#include "flutter/fml/macros.h"
+#include "impeller/entity/contents/runtime_effect_contents.h"
+#include "impeller/entity/entity.h"
+#include "impeller/geometry/color.h"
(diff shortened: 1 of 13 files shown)


jonahwilliams (approved):
LGTM!

impeller/aiks/color_source.h
```diff
@@ -0,0 +1,100 @@
...
+ public:
+  enum class Type {
+    kColor,
+    kImage,
+    kLinearGradient,
+    kRadialGradient,
+    kConicalGradient,
+    kSweepGradient,
+    kRuntimeEffect,
+    kScene,
```
zanderso, line 34: Not quite following where the Scene stuff needs to be #ifdef'd out vs. in. Why is this in the clear here, but Scene code under `IMPELLER_ENABLE_3D` in others?

bdero: Currently we always build Scene as part of Impeller, but completely eliminate it from the DisplayList + UI unless `IMPELLER_ENABLE_3D` is on to make accidental exposure impossible. Maybe we should rename it to e.g. `FLUTTER_ENABLE_3D` some time. 😄

PR #41320, opened 2023-04-19
[Impeller] Add buffer-to-texture blit capability check; fix GL+VK playgrounds on macOS

bdero:
GL/VK playgrounds that load image fixtures are currently broken on macOS.


jonahwilliams (approved):
LGTM

PR #41325, opened 2023-04-19
[Impeller] Comprehensively label snapshots

bdero:
Makes it much easier to work out where things are going wrong in filter chains.

For example, when capturing the app in https://github.com/flutter/flutter/issues/124956, the chain of RenderPasses gets labeled as follows:
1. Contents to ForegroundPorterDuffBlend Filter Snapshot
2. Contents to GaussianBlur Filter Snapshot
3. Directional Gaussian Blur Filter
4. Directional Gaussian Blur Filter

...and so it's easy to see that the filter chain is ForegroundPorterDuffBlend->DirectionalGaussianBlur->DirectionalGaussianBlur, and the former two are triggering a non-collapsed snapshot of their inputs.


chinmaygarde (approved):
Awesome.

PR #41332, opened 2023-04-19
[Impeller] Fix transform regressions for chained filters

bdero:
Resolves https://github.com/flutter/flutter/issues/124956.

Fixes some recent transform issues that mainly apply to chained filters (like when both a color + image filter has been set).

- For all filters, the returned entity should just be an identity matrix because the snapshot transform gets fully absorbed.
- For all filters, given entity's transform should be applied in the coverage proc (although this currently makes no difference because of the way coverage is queried when snapshotting).
- For filters drawing with coverage + transforming UVs, the transform handed to the contents of the deferred entity needs to be applied. Otherwise the entity won't get repositioned correctly while snapshotting.

New golden: `Play/AiksTest.TranslucentSaveLayerWithColorFilterAndImageFilterDrawsCorrectly/Metal`

```bash
autoninja -C ../out/host_debug_unopt/ -j 1000 flutter/impeller:impeller_unittests && ../out/host_debug_unopt/impeller_unittests --timeout=0 --gtest_filter="Play/AiksTest.TranslucentSaveLayerWithColorFilterAndImageFilterDrawsCorrectly/Metal" --enable_playground
```

Before:
Screenshot 2023-04-19 at 12 03 57 AM
Screenshot 2023-04-19 at 3 02 06 AM

After:
Screenshot 2023-04-19 at 12 00 34 AM
Screenshot 2023-04-19 at 2 59 56 AM

### DIFF
--- a/impeller/aiks/aiks_unittests.cc
+++ b/impeller/aiks/aiks_unittests.cc
@@ -1955,7 +1955,7 @@ TEST_P(AiksTest, TranslucentSaveLayerWithBlendColorFilterDrawsCorrectly) {
   ASSERT_TRUE(OpenPlaygroundHere(canvas.EndRecordingAsPicture()));
 }
 
-TEST_P(AiksTest, TranslucentSaveLayerWithBlendImageFilterAndDrawsCorrectly) {
+TEST_P(AiksTest, TranslucentSaveLayerWithBlendImageFilterDrawsCorrectly) {
   Canvas canvas;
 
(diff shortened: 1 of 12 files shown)


chinmaygarde (approved):

jonahwilliams (approved):
LGTM 😅

bdero:
The regressing changes (https://github.com/flutter/engine/pull/41098 and https://github.com/flutter/engine/pull/40973) landed after the April 6th branch cut, so this doesn't need a 3.10 cherry-pick:

PR #41347, opened 2023-04-19
[impeller] fixed flake where blit wasn't tracking its texture

gaaclarke:
fixes https://github.com/flutter/flutter/issues/125147

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/impeller/renderer/backend/vulkan/blit_command_vk.cc
+++ b/impeller/renderer/backend/vulkan/blit_command_vk.cc
@@ -29,6 +29,10 @@ bool BlitCopyTextureToTextureCommandVK::Encode(
   const auto& src = TextureVK::Cast(*source);
   const auto& dst = TextureVK::Cast(*destination);
 
+  if (!encoder.Track(source) || !encoder.Track(destination)) {
+    return false;
+  }
+
   LayoutTransition src_tran;
   src_tran.cmd_buffer = cmd_buffer;
   src_tran.new_layout = vk::ImageLayout::eTransferSrcOptimal;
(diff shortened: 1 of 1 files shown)


gaaclarke:
@chinmaygarde Let me know if you have any ideas on how to make this less likely to happen in the future.  It seems like an easy mistake to make and it's difficult to debug.  Nothing about the failure communicated to me that the command buffer was referencing a texture that doesn't exist.

I mentioned in the chat but maybe we should make people requesting the command buffer to state their tracked resources at the same time so that they have to explicitly omit tracking a resource.  That may help.

gaaclarke:
By extension, shouldn't [`BlitCopyTextureToTextureCommandVK::Encode`](https://github.com/gaaclarke/engine/blob/64b2241a4ccf677e9d8fb61705c79e3be17cf488/impeller/renderer/backend/vulkan/blit_command_vk.cc#L25) also track its textures?

gaaclarke:
I expanded this to the other blit commands.  Every 10th run of the unit tests can pass now, but there seems to be another case where we aren't tracking something we should be.

gaaclarke:
@chinmaygarde can you please give this a review.  This makes sense to me but I want to double check that it does for you too.  After we decide that I can look into adding tests and we can think about how to maybe change the API to make this harder to happen or easier to debug.

gaaclarke:
I'm going to put this up for review, it is technically covered by playground tests but I plan on investigating creating a more extensive testing platform that will allow us to assert this problem directly.

chinmaygarde (approved):

PR #41351, opened 2023-04-19
[impeller] convert src over to src for solid color

jonahwilliams:
Related to https://github.com/flutter/flutter/issues/114402

At least in the flutter gallery test app, with https://github.com/flutter/gallery/pull/909 reverted - this significantly improves the performance in the Xcode frame debugger. If we wanted to do this for more color sources or filter effects, we need to do much more work to track texture opacity, but this seems like a reasonable and low cost first step.

### Before

image

### After

image

### DIFF
--- a/impeller/entity/contents/solid_color_contents.cc
+++ b/impeller/entity/contents/solid_color_contents.cc
@@ -46,6 +46,11 @@ bool SolidColorContents::ShouldRender(
   return Contents::ShouldRender(entity, stencil_coverage);
 }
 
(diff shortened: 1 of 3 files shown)


bdero (approved):
LGTM. I'm working on a comprehensive patch for this. We need to do the coercion earlier in order to be compatible with https://github.com/flutter/flutter/issues/114402, but cool with landing this in the meantime.

jonahwilliams:
Sounds good!

PR #41374, opened 2023-04-20
[Impeller] Use a device buffer for SkBitmap allocation, use Linear texture on Metal backend.

jonahwilliams:
This reduces the cost of uploading the glyph atlas, as we can exploit metal's capabilies to create a texture from a device buffer and share the underlying memory with a linear texture. This also means that subsequent updates don't require uploads or copies either.

From scrolling through wonderous, this shaves off about 0.2 ms (0.6 ms -> 0.4 ms) of fresh atlas construction.


impeller/typographer/backends/skia/text_render_context_skia.h
```diff
@@ -6,9 +6,30 @@
...
+  // |Allocator|
+  bool allocPixelRef(SkBitmap* bitmap) override;
+
+  std::optional<std::shared_ptr<DeviceBuffer>> GetDeviceBuffer() const;
+
+ private:
+  std::shared_ptr<impeller::Allocator> allocator_;
+  std::optional<std::shared_ptr<DeviceBuffer>> buffer_;
+};
+
```
jonahwilliams, line 32: This is an exact copy of the implementation used in image decoding.

jonahwilliams: I'd like to put it somewhere shared but I'm not quite sure where that would be.

jonahwilliams:
For non metal platforms this may still be faster as it will use the device allocator instead of malloc, but IIRC it should have the same number of copies and uploads.

impeller/typographer/backends/skia/text_render_context_skia.h
jonahwilliams: ignoring the skia dependency, it would just go in core. But with the skia dep I have no idea. I could create a new impeller-skia interop target?

impeller/renderer/backend/vulkan/capabilities_vk.cc
```diff
@@ -348,10 +348,16 @@ bool CapabilitiesVK::SupportsReadFromOnscreenTexture() const {
...
 }
 
+// |Capabilities|
 bool CapabilitiesVK::SupportsDecalTileMode() const {
   return true;
 }
 
+// |Capabilities|
+bool CapabilitiesVK::SupportsSharedDeviceBufferTextureMemory() const {
+  return false;
```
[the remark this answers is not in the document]
jonahwilliams: I did some research but I think is metal only

impeller/typographer/backends/skia/text_render_context_skia.cc
```diff
@@ -502,15 +506,18 @@ std::shared_ptr<GlyphAtlas> TextRenderContextSkia::CreateGlyphAtlas(
...
       return nullptr;
     }
 
     // ---------------------------------------------------------------------------
     // Step 6: Update the existing texture with the updated bitmap.
+    //         This is only necessary on backends that don't support creating
+    //         a texture that shares memory with the underlying device buffer.
     // ---------------------------------------------------------------------------
-    if (!UpdateGlyphTextureAtlas(bitmap, last_atlas->GetTexture())) {
+    if (!capabilities->SupportsSharedDeviceBufferTextureMemory() &&
```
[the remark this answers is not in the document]
jonahwilliams: Technically this and below where we construct a texture from the device buffer, that required a second copy before but now its "free" on iOS/macOS

impeller/typographer/backends/skia/text_render_context_skia.h
```diff
@@ -6,9 +6,30 @@
...
 #include "flutter/fml/macros.h"
 #include "impeller/typographer/text_render_context.h"
+#include "third_party/skia/include/core/SkBitmap.h"
 
 namespace impeller {
 
+class DeviceBuffer;
+class Allocator;
+
+class FontImpellerAllocator : public SkBitmap::Allocator {
```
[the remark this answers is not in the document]
jonahwilliams: still working on this, want to combine it with the identical class in image_decode_impeller but haven't thought of the right target yet.

impeller/typographer/backends/skia/text_render_context_skia.cc
```diff
@@ -552,11 +559,12 @@ std::shared_ptr<GlyphAtlas> TextRenderContextSkia::CreateGlyphAtlas(
   // ---------------------------------------------------------------------------
   // Step 7: Draw font-glyph pairs in the correct spot in the atlas.
   // ---------------------------------------------------------------------------
-  auto bitmap = CreateAtlasBitmap(*glyph_atlas, atlas_size);
+  auto [bitmap, device_buffer] = CreateAtlasBitmap(
```
gaaclarke, line 562: These are shadowing the declarations in 509, right?

gaaclarke (approved):
lgtm

impeller/typographer/backends/skia/text_render_context_skia.cc
jonahwilliams: No, those are in a different block

impeller/typographer/backends/skia/text_render_context_skia.h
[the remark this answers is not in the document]
jonahwilliams: Added a doc comment here. Haven't found a home yet but I think this can still be landed

impeller/typographer/backends/skia/text_render_context_skia.cc
```diff
@@ -139,7 +141,9 @@ ISize OptimumAtlasSizeForFontGlyphPairs(
     const FontGlyphPair::Set& pairs,
     std::vector<Rect>& glyph_positions,
     const std::shared_ptr<GlyphAtlasContext>& atlas_context) {
-  static constexpr auto kMinAtlasSize = 8u;
+  // This size needs to be above the minimum required aligment for linear
+  // textures. This is 256 for older intel macs and decreases on iOS devices.
+  static constexpr auto kMinAtlasSize = 256u;
```
dnfield, line 146: nit: Look this up based on caps?

impeller/typographer/backends/skia/text_render_context_skia.cc
```diff
@@ -421,32 +430,27 @@ static std::shared_ptr<Texture> UploadGlyphTextureAtlas(
...
+  // a linear texture and instead must blit to a new texture.
   if (pixmap.rowBytes() * pixmap.height() !=
       texture_descriptor.GetByteSizeOfBaseMipLevel()) {
     return nullptr;
   }
 
-  auto texture = allocator->CreateTexture(texture_descriptor);
+  FML_DCHECK(allocator.MinimumBytesPerRow(format) <= pixmap.rowBytes());
+  auto texture = device_buffer->AsTexture(allocator, texture_descriptor,
+                                          texture_descriptor.GetBytesPerRow());
```
dnfield, line 442: Does this punish non-metal implementations right now?

I think it doesn't because before we did the copy explicitly in this code, and now it gets done implicitly if needed there, right?

impeller/typographer/backends/skia/text_render_context_skia.h
```diff
@@ -6,9 +6,37 @@
...
 #include "impeller/typographer/text_render_context.h"
+#include "third_party/skia/include/core/SkBitmap.h"
 
 namespace impeller {
 
+class DeviceBuffer;
+class Allocator;
+
+/// @brief An implementation of an SkBitmap allocator that deferrs allocation to
+///        an impeller allocator. This allows usage of Skia software rendering
```
dnfield, line 17:
```suggestion
///        an Impeller allocator. This allows usage of Skia software rendering
```

impeller/typographer/backends/skia/text_render_context_skia.h
```diff
@@ -6,9 +6,37 @@
...
+///        to write to a host buffer or linear texture without an extra copy.
+///
+///        This class is an exact copy of the implementation in
+///        image_decode_impeller.cc due to the lack of a reasonable library
+///        that could be shared.
```
dnfield, line 22: Right now we have `display_list:skia_conversions`, `aiks`, and `image` and this that know about Skia.

It seems like this could probably go into `image`, and we could add a dep to `impeller/image` to both `lib/ui` and `impeller/typographer`. Doesn't have to be done in this patch but would be good as a follow up if not done here.

impeller/typographer/backends/skia/text_render_context_skia.h
dnfield: I don't think we should add a dependency on skia to core but we _could_. I think `impeller/image` is probably the right spot for this.

dnfield (approved):
LGTM with nits

impeller/typographer/backends/skia/text_render_context_skia.cc
jonahwilliams: Yup!

impeller/typographer/backends/skia/text_render_context_skia.cc
jonahwilliams: I think it might actually be better to keep a larger minimum atlas size since we reuse the texture across frames now, but only if the size is unchanged. The memory hit is fairly minor

dnfield: Yes, but what happens if we run on some backend where the minimum size is even larger?

jonahwilliams: it only matters for metal

jonahwilliams: I suppose its possible that a future iOS or macOS versions requires a larger minimum alignment. I'll update this to compute the exact minimum alignment.

That said, separately I think we should use a much larger minimum texture size

dnfield (approved):

PR #41386, opened 2023-04-21
[Impeller] Fix SupportsReadFromOnscreenTexture capability check

bdero:
I noticed we still weren't going down the blit-less path on iOS, and it turned out that the wrong capability was getting assigned in the builder.


jonahwilliams (approved):
LGTM

PR #41391, opened 2023-04-21
[Impeller] Fix division by zero for transparent shadows

bdero:
A division by zero happens if the shadow color is fully transparent, and the NaNs are getting assigned to the RGB values of the paint color being handed to the rrect shadow draw. This doesn't cause a problem when wide gamut is off because NaN output ends up being interpreted as zero (thus making the shadow output fully transparent black, which happens to be the expected and correct result). However, when a wide gamut attachment is present, the NaN output ends up being interpreted as a negative value.

Reproduction app:
```dart

void main() => runApp(const GeneralDialogApp());

class EvilPainter extends CustomPainter {
  @override
    final Rect rect = Offset.zero & size;
    canvas.drawPaint(Paint()..color = Colors.white);
    canvas.saveLayer(null, Paint()..blendMode = BlendMode.srcOver);
    canvas.drawShadow(Path()..addRect(Rect.fromLTRB(100, 100, 300, 300)),
        Colors.black54, 15, false);
    canvas.drawShadow(Path()..addRect(Rect.fromLTRB(100, 100, 300, 300)),
        Colors.black54, 15, false);
    canvas.drawShadow(Path()..addRect(Rect.fromLTRB(100, 100, 300, 300)),
        Colors.transparent, 15, false);
    canvas.restore();
  }

  @override
  bool shouldRepaint(EvilPainter oldDelegate) => false;
  @override
  bool shouldRebuildSemantics(EvilPainter oldDelegate) => false;
}

class GeneralDialogApp extends StatelessWidget {
  const GeneralDialogApp({super.key});

  @override
    return MaterialApp(
      restorationScopeId: 'app',
      home: CustomPaint(painter: EvilPainter()),
    );
  }
}
```

Before:
image

After:
image


zanderso (approved):
Nice find!

zanderso:
Skia gold seems confused. I'm going to go ahead and land this.

PR #41414, opened 2023-04-22
[Impeller] dont use waitUntilScheduled on iOS

jonahwilliams:
IIRC this was perfectly safe, we needed to disable it for macOS and iOS simulator. Unfortunately `FML_OS_MACOSX` is defined on all apple platforms so this was still running.


jonahwilliams:
actually just noticed a weird issue running this locally... will need to investigate.

PR #41425, opened 2023-04-22
Double zip FlutterMacOS.dSYM.zip.

godofredoc:
The flutter tool is expecting FlutterMacOS.dSYM.zip to be double zipped.

Bug: https://github.com/flutter/flutter/issues/124911

## Pre-launch Checklist

- [X] I signed the [CLA].

### DIFF
--- a/sky/tools/create_macos_framework.py
+++ b/sky/tools/create_macos_framework.py
@@ -164,14 +164,17 @@ def process_framework(dst, args, fat_framework, fat_framework_binary):
   # Zip FlutterMacOS.framework.
   if args.zip:
     filepath_with_entitlements = ''
-    filepath_without_entitlements = 'FlutterMacOS.framework/Versions/A/FlutterMacOS'
+
+    framework_dst = os.path.join(dst, 'FlutterMacOS.framework')
+    filepath_without_entitlements = 'Versions/A/FlutterMacOS'
 
     embed_codesign_configuration(
-        os.path.join(dst, 'entitlements.txt'), filepath_with_entitlements
+        os.path.join(framework_dst, 'entitlements.txt'),
+        filepath_with_entitlements
     )
 
     embed_codesign_configuration(
-        os.path.join(dst, 'without_entitlements.txt'),
+        os.path.join(framework_dst, 'without_entitlements.txt'),
         filepath_without_entitlements
     )
     subprocess.check_call([
@@ -179,11 +182,11 @@ def process_framework(dst, args, fat_framework, fat_framework_binary):
         '-r',
         '-y',
         'FlutterMacOS.framework.zip',
-        'FlutterMacOS.framework',
+        '.',
         'entitlements.txt',
         'without_entitlements.txt',
     ],
-                          cwd=dst)
+                          cwd=framework_dst)
     # Double zip to make it consistent with legacy artifacts.
     # TODO(fujino): remove this once https://github.com/flutter/flutter/issues/125067 is resolved
     subprocess.check_call([
@@ -192,9 +195,9 @@ def process_framework(dst, args, fat_framework, fat_framework_binary):
         'FlutterMacOS.framework_.zip',
         'FlutterMacOS.framework.zip',
     ],
-                          cwd=dst)
+                          cwd=framework_dst)
     # Use doubled zipped file.
-    final_src_path = os.path.join(dst, 'FlutterMacOS.framework_.zip')
+    final_src_path = os.path.join(framework_dst, 'FlutterMacOS.framework_.zip')
     final_dst_path = os.path.join(dst, 'FlutterMacOS.framework.zip')
     shutil.move(final_src_path, final_dst_path)
 
--- a/sky/tools/create_macos_framework.py
+++ b/sky/tools/create_macos_framework.py
@@ -166,7 +166,8 @@ def process_framework(dst, args, fat_framework, fat_framework_binary):
     filepath_with_entitlements = ''
 
     framework_dst = os.path.join(dst, 'FlutterMacOS.framework')
-    filepath_without_entitlements = 'Versions/A/FlutterMacOS'
+    # TODO(xilaizhang): Remove the zip file from the path when outer zip is removed.
+    filepath_without_entitlements = 'FlutterMacOS.framework.zip/Versions/A/FlutterMacOS'
 
     embed_codesign_configuration(
         os.path.join(framework_dst, 'entitlements.txt'),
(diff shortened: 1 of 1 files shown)


godofredoc:
\cc @XilaiZhang this will impact codesigning in two different parts: identifying if the file requires signing and also on the signing tool detecting if it needs to unzip twice.

XilaiZhang:
> \cc @XilaiZhang this will impact codesigning in two different parts: identifying if the file requires signing and also on the signing tool detecting if it needs to unzip twice.

thanks for the info! Yeah I think the current topology covers this case.

based on the insights from https://github.com/flutter/flutter/issues/124322#issuecomment-1499409762 we would skip signing this file since it is debug symbols and not executables. On the signing tool side, this file will be skipped since if no signing metadata is found upon the first unzip, the file will be skipped.

sky/tools/create_macos_framework.py
```diff
@@ -152,26 +164,29 @@ def process_framework(dst, args, fat_framework, fat_framework_binary):
...
     )
     subprocess.check_call([
         'zip',
         '-r',
         '-y',
         'FlutterMacOS.framework.zip',
-        'FlutterMacOS.framework',
+        '.',
         'entitlements.txt',
         'without_entitlements.txt',
```
XilaiZhang, line 187: Right we would remove the metadata from the inner layer of the zip, and move them to the outer layer of the zip, so we will remove L186-187.

sky/tools/create_macos_framework.py
```diff
@@ -180,9 +195,9 @@ def process_framework(dst, args, fat_framework, fat_framework_binary):
         'FlutterMacOS.framework_.zip',
         'FlutterMacOS.framework.zip',
     ],
```
XilaiZhang, line 197: yeah we would add codesign metadata here in the outer layer of the zip, something similar to  
         'zip',
        '-y',
        'FlutterMacOS.framework_.zip',
        'FlutterMacOS.framework.zip',
        'entitlements.txt',
        'without_entitlements.txt',

So adding  'entitlements.txt', 'without_entitlements.txt', on L197 and L198

sky/tools/create_macos_framework.py
```diff
@@ -152,26 +164,29 @@ def process_framework(dst, args, fat_framework, fat_framework_binary):
   # Zip FlutterMacOS.framework.
   if args.zip:
     filepath_with_entitlements = ''
-    filepath_without_entitlements = 'FlutterMacOS.framework/Versions/A/FlutterMacOS'
+
+    framework_dst = os.path.join(dst, 'FlutterMacOS.framework')
+    filepath_without_entitlements = 'Versions/A/FlutterMacOS'
```
XilaiZhang, line 169: The metadata path will now be `FlutterMacOS.framework.zip/Versions/A/FlutterMacOS`
so L169 will be like `filepath_without_entitlements = FlutterMacOS.framework.zip/Versions/A/FlutterMacOS`

XilaiZhang (commented):
Thanks for explaining! I added the changes needed for codesign metadata in comments.

I tested the correctness of the new change by downloading the artifacts built from bab6f0e811596da498e70372c3fc91ca03ebb44c engine v2 bucket, and run codesign locally on it.

sky/tools/create_macos_framework.py
godofredoc: duplicated the metadata to simplify removing the outer zip later.

sky/tools/create_macos_framework.py
godofredoc: ack

sky/tools/create_macos_framework.py
godofredoc: ack

godofredoc: Actually adding this seems overcomplicated. As a workaround I'll check the file name explicitly in the recipes with a TODO to remove it after the outer zip is deleted.

sky/tools/create_macos_framework.py
godofredoc: That will complicate the logic in several places. As the outer zipped is expected to be removed ~2weeks from now I'll hard code the name in recipes for now.

godofredoc:
> > \cc @XilaiZhang this will impact codesigning in two different parts: identifying if the file requires signing and also on the signing tool detecting if it needs to unzip twice.
> 
> thanks for the info! Yeah I think the current topology covers this case.
> 
> based on the insights from [flutter/flutter#124322 (comment)](https://github.com/flutter/flutter/issues/124322#issuecomment-1499409762) we would skip signing this file since it is debug symbols and not executables. On the signing tool side, this file will be skipped since if no signing metadata is found upon the first unzip, the file will be skipped.

This is covered at the recipes level: https://flutter-review.googlesource.com/c/recipes/+/42560

godofredoc:
> Thanks for explaining! I added the changes needed for codesign metadata in comments.
> 
> I tested the correctness of the new change by downloading the artifacts built from [bab6f0e](https://github.com/flutter/engine/commit/bab6f0e811596da498e70372c3fc91ca03ebb44c) engine v2 bucket, and run codesign locally on it.

ACK

XilaiZhang:
> > > \cc @XilaiZhang this will impact codesigning in two different parts: identifying if the file requires signing and also on the signing tool detecting if it needs to unzip twice.
> > 
> > 
> > thanks for the info! Yeah I think the current topology covers this case.
> > based on the insights from [flutter/flutter#124322 (comment)](https://github.com/flutter/flutter/issues/124322#issuecomment-1499409762) we would skip signing this file since it is debug symbols and not executables. On the signing tool side, this file will be skipped since if no signing metadata is found upon the first unzip, the file will be skipped.
> 
> This is covered at the recipes level: https://flutter-review.googlesource.com/c/recipes/+/42560

Umm I left a comment in recipe. Sorry if I misunderstood, umm I think in the recipe we are requiring the zip to be codesigned, but we are supplying empty configs. My understanding is that the codesign tool might detect the binary in the zip and report a failure due to the empty config. Or maybe we are planning to add more commits in this pr

godofredoc:
> > > > \cc @XilaiZhang this will impact codesigning in two different parts: identifying if the file requires signing and also on the signing tool detecting if it needs to unzip twice.
> > > 
> > > 
> > > thanks for the info! Yeah I think the current topology covers this case.
> > > based on the insights from [flutter/flutter#124322 (comment)](https://github.com/flutter/flutter/issues/124322#issuecomment-1499409762) we would skip signing this file since it is debug symbols and not executables. On the signing tool side, this file will be skipped since if no signing metadata is found upon the first unzip, the file will be skipped.
> > 
> > 
> > This is covered at the recipes level: https://flutter-review.googlesource.com/c/recipes/+/42560
> 
> Umm I left a comment in recipe. Sorry if I misunderstood, umm I think in the recipe we are requiring the zip to be codesigned, but we are supplying empty configs. My understanding is that the codesign tool might detect the binary in the zip and report a failure due to the empty config. Or maybe we are planning to add more commits in this pr

Details:

* The double zip behavior is temporary, ~couple of weeks until it is fixed.
* Having the inner zip with the correct structure(metadata with correct paths) would have been the easiest way to remove the outer zip, preventing additional disruption when the outer zip is removed.

godofredoc:
@XilaiZhang implemented your suggestions and added to TODOs to follow up when removing the outer zip folders.

@christopherfujino @jmagman  any concerns about adding the double zip logic to the dsym artifact?

jmagman:
> any concerns about adding the double zip logic to the dsym artifact?

I don't think the tool knows anything about the dSYM artifact, I don't think it downloads it at all.  Users manually download it https://github.com/flutter/flutter/wiki/Crashes (and macOS isn't mentioned).  Was that artifact actually causing issues somewhere?  (I know the .framework was https://github.com/flutter/flutter/issues/124911#issuecomment-1513263342).

godofredoc:
> > any concerns about adding the double zip logic to the dsym artifact?
> 
> I don't think the tool knows anything about the dSYM artifact, I don't think it downloads it at all. Users manually download it https://github.com/flutter/flutter/wiki/Crashes (and macOS isn't mentioned). Was that artifact actually causing issues somewhere? (I know the .framework was [flutter/flutter#124911 (comment)](https://github.com/flutter/flutter/issues/124911#issuecomment-1513263342)).

It was not causing any issues but when checking all the artifacts I found the legacy dsym artifact is double zipped. I'm trying to get all the artifacts generated by engine v2 aligned with the legacy artifacts to avoid new disruptions during the migration.

XilaiZhang (approved):
LGTM! sorry for blocking v2 migration

As a side note, we could keep the metadata for both the inner zip and outer zip (have them both present), and remove the outer metadata when the outer zip is removed. This would work because the codesign tool only looks at metadata at the outer most layer.

PR #41466, opened 2023-04-24
Revert "[impeller] convert src over to src for solid color"

jonahwilliams:
Reverts flutter/engine#41351


jonahwilliams:
This caused an unexpected regression. Reverting until we investigate the cause

jason-simmons (approved):

PR #41488, opened 2023-04-25
Download and use the goma client from cipd

zanderso:
In https://github.com/flutter/flutter/issues/125361 we discovered that a new version of clang can require updating goma in order for goma to work properly. This PR adds a dependency on CIPD goma to the DEPS file so that we can update it for use in local builds when needed. Since CI does its own management of the goma client and the compiler proxy, and since goma can only be used by Googlers, the DEPS file download is guarded behind the `use_cipd_goma` gclient var. To use it one would update the engine `.gclient` file to be something like:
```
~/flutter/engine/src $ cat ../.gclient
solutions = [
  {
    "managed": False,
    "name": "src/flutter",
    "url": "git@github.com:zanderso/engine.git",
    "custom_deps": {},
    "custom_vars": {
      "use_cipd_goma": True,
    },
    "deps_file": "DEPS",
    "safesync_url": "",
  },
]
```

I'll update the wiki with these instructions after this PR lands.

### DIFF
--- a/DEPS
+++ b/DEPS
@@ -39,7 +39,13 @@ vars = {
   # The list of revisions for these tools comes from Fuchsia, here:
   # https://fuchsia.googlesource.com/integration/+/HEAD/toolchain
   # If there are problems with the toolchain, contact fuchsia-toolchain@.
-  'clang_version': 'git_revision:e8cfbfd05a951b85f80156dffc8eeecb34c7271c',
+  'clang_version': 'git_revision:20d06c833d833ef6b2d0f519cc4a7998d49a2803',
+
+  # The goma version and the clang version can be tightly coupled. If goma
+  # stops working on a clang roll, this may need to be updated using the value
+  # from the 'integration' tag of
+  # https://chrome-infra-packages.appspot.com/p/fuchsia/third_party/goma/client
+  'goma_version': ' git_revision:41b3bcb64014144a844153fd5588c36411fffb56',
 
   # When updating the Dart revision, ensure that all entries that are
   # dependencies of Dart are also updated to match the entries in the
@@ -98,7 +104,12 @@ vars = {
   "checkout_llvm": False,
 
   # Setup Git hooks by default.
-  "setup_githooks": True,
+  'setup_githooks': True,
+
+  # When this is true, the goma client will be downloaded from cipd, and
+  # the engine build will prefer to use this client over a client that is
+  # specified by GOMA_DIR, or installed in the default goma install location.
+  'use_cipd_goma': False,
 
   # This is not downloaded be default because it increases the
   # `gclient sync` time by between 1 and 3 minutes. This option is enabled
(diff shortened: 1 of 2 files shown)


tools/gn
```diff
@@ -237,6 +266,9 @@ def setup_goma(args):
         os.getenv('FLUTTER_GOMA_CREATE_XCODE_SYMLINKS', '0') == '1'):
       goma_gn_args['create_xcode_symlinks'] = True
 
+  if goma_gn_args['use_goma'] and args.goma_ensure_start:
+    goma_ensure_start(goma_gn_args['goma_dir'])
```
godofredoc, line 270: Should this call be gated by an env variable to execute only on local builds? starting goma this way for CI builds will remove all the visibility from the goma recipe module and can potentially cause issues difficult to triage.

tools/gn
```diff
@@ -1070,6 +1102,19 @@ def parse_args(args):
       help='Do not run GN. Instead configure the Impeller cmake example build.',
   )
 
+  parser.add_argument(
```
godofredoc, line 1105: Are we giving too much power to this script? by the name it seems it was originally designed as a gn wrapper but now it is doing gn, cmake, and goma.

zanderso: By design, this is the build setup step that the engine_v2 recipe calls, so if we want to run something before we do a ninja build, this is where it has to go, right?

tools/gn
zanderso: If we're un-forking the goma recipe module, then this should be gated to be only for local builds, but if we can't un-fork the recipe module, then driving goma from this script is one way we could get rid of it, and avoid blocking clang rolls going forward.

godofredoc: I'll send the CL to prepare the goma module to unfork it when there is an arm64 cipd package

DEPS
```diff
@@ -965,6 +1010,36 @@ hooks = [
       'src/flutter/tools/activate_emsdk.py',
     ]
   },
+  {
```
godofredoc, line 1013: Will this work for external users with no access to goma?

zanderso:
Added stronger gating on the gclient var, and shifted `goma_ctl.py ensure_start` to a gclient runhook (also gated by the gclient var) to avoid running on every `gn` invocation.

DEPS
zanderso: No, but this runhook will only run if the `use_cipd_goma` is explicitly set to true.

DEPS
```diff
@@ -41,6 +41,12 @@ vars = {
   # If there are problems with the toolchain, contact fuchsia-toolchain@.
   'clang_version': 'git_revision:20d06c833d833ef6b2d0f519cc4a7998d49a2803',
 
+  # The goma version and the clang version can be tightly coupled. If goma
+  # stops working on a clang roll, this may need to be updated using the value
+  # from the 'integration' tag of
+  # https://chrome-infra-packages.appspot.com/p/fuchsia/third_party/goma/client
+  'goma_version': ' git_revision:41b3bcb64014144a844153fd5588c36411fffb56',
```
dnfield, line 48: not `integration`?

Are we updating the clang roller to update this?

dnfield:
Making the compiler proxy start as a gclient hook rather than the GN script is an improvement

DEPS
zanderso: Yeah that's a good idea. We can ask the Skia autoroller folks if that's possible.

zanderso: Not `integration` because we should probably avoid unpinned dependencies.

PR #41501, opened 2023-04-26
[Impeller] iOS/macOS: Only wait for command scheduling prior to present (redux)

bdero:
Reverts https://github.com/flutter/engine/pull/40895.
Resolves https://github.com/flutter/flutter/issues/120399 (again).

A bunch of frames get pumped on the main thread _without a transaction_ just before unmerging occurs (I don't know why this happens), and so checking the current thread to determine whether we need to present with a transaction or not isn't sufficient. In the prior fix, after the unmerge, the raster thread would hang for one second while waiting for the next drawable to get freed up (happens on the second raster thread frame post-unmerge), and then subsequent presents would just do nothing for a while, but eventually recover.

`presentsWithTransaction` works whether the `CATransaction` stack is empty or not, and so the only difference here is that `presentsWithTransaction` is always turned on and `presentDrawable` is always avoided (otherwise it tries to present too early and nothing renders when platform views are present).

### DIFF
--- a/impeller/renderer/backend/metal/command_buffer_mtl.mm
+++ b/impeller/renderer/backend/metal/command_buffer_mtl.mm
@@ -167,27 +167,6 @@ bool CommandBufferMTL::OnSubmitCommands(CompletionCallback callback) {
 
   [buffer_ commit];
 
-#if (FML_OS_MACOSX || FML_OS_IOS_SIMULATOR)
-  // We're using waitUntilScheduled on macOS and iOS simulator to force a hard
-  // barrier between the execution of different command buffers. This forces all
-  // renderable texture access to be synchronous (i.e. a write from a previous
-  // command buffer will not get scheduled to happen at the same time as a read
-  // in a future command buffer).
-  //
-  // Metal hazard tracks shared memory resources by default, and we don't need
-  // to do any additional work to synchronize access to MTLTextures and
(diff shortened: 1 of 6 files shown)


bdero:
FYI @chinmaygarde, this returned to the original plan of forcing one `waitUntilScheduled` on the fly using a command buffer prior to present.

jonahwilliams:
This PR results in the same issue I ran into when I just removed the waitUntilScheduled - the artifact discovery popup on wonderous doesn't show up.

jonahwilliams:
(From local testing)

bdero:
> This PR results in the same issue I ran into when I just removed the waitUntilScheduled - the artifact discovery popup on wonderous doesn't show up.

Hmm, this behavior seems to be happening for me at HEAD, without these changes?

jonahwilliams:
Uh oh. I must be on an older commit in flutter/flutter, guess its time to bisect

jonahwilliams:
I looked in the range of commits from 6d798393a1 to 122c3b38 and the only thing that seemed relevant was the change to SetSupportsReadFromOnscreenTexture. I confirmed that setting that back to false fixes the artifact popup.

bdero:
image

bdero:
Fixed the artifact discovery problem here: https://github.com/flutter/engine/pull/41509

PR #41521, opened 2023-04-26
[platform_view] Only dispose view when it is removed from the composition order

cyanglaz:
DisposeView happens every frame before laying out PlatformViews, with the order specified in `composition_order_`. Because `view_to_dispose_` is determined on UI thread and `composition_order_` is determined on the platform thread, there could be a race if the dart code on the UI thread runs faster than the rasterizer on the Platform thread, causing a view in `view_to_dispose_` is still in the `composition_order_`.

This PR delays the views in the `composition_order_` being disposed and wait for the next frame to dispose them. 

fixes https://github.com/flutter/flutter/issues/125223

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/shell/platform/darwin/ios/framework/Source/FlutterPlatformViews.mm
+++ b/shell/platform/darwin/ios/framework/Source/FlutterPlatformViews.mm
@@ -869,10 +869,10 @@ void FlutterPlatformViewsController::DisposeViews() {
 
   std::unordered_set<int64_t> views_to_composite(composition_order_.begin(),
                                                  composition_order_.end());
-  std::unordered_set<int64_t> views_to_dispose_next_frame;
+  std::unordered_set<int64_t> views_to_delay_dispose;
   for (int64_t viewId : views_to_dispose_) {
     if (views_to_composite.count(viewId)) {
-      views_to_dispose_next_frame.insert(viewId);
+      views_to_delay_dispose.insert(viewId);
       continue;
     }
     UIView* root_view = root_views_[viewId].get();
@@ -883,10 +883,9 @@ void FlutterPlatformViewsController::DisposeViews() {
     current_composition_params_.erase(viewId);
     clip_count_.erase(viewId);
     views_to_recomposite_.erase(viewId);
-    slices_.erase(viewId);
   }
 
-  views_to_dispose_ = std::move(views_to_dispose_next_frame);
+  views_to_dispose_ = std::move(views_to_delay_dispose);
 }
 
 void FlutterPlatformViewsController::BeginCATransaction() {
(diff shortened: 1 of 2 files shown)


cyanglaz:
@saintmac-google Do you mind to test this patch again :) thank you

shell/platform/darwin/ios/framework/Source/FlutterPlatformViews.mm
```diff
@@ -876,8 +883,10 @@ static bool ClipRRectContainsPlatformViewBoundingRect(const SkRRect& clip_rrect,
     current_composition_params_.erase(viewId);
     clip_count_.erase(viewId);
     views_to_recomposite_.erase(viewId);
+    slices_.erase(viewId);
   }
-  views_to_dispose_.clear();
+
+  views_to_dispose_ = views_to_dispose_next_frame;
```
saintmac-google, line 889: this could be std::move(views_to_dispose_next_frame)

saintmac-google (commented):
tested on our app, it works (also with std::move)

cyanglaz:
Friendly ping @hellohuanlin  :)

shell/platform/darwin/ios/framework/Source/FlutterPlatformViews.mm
```diff
@@ -867,7 +867,14 @@ static bool ClipRRectContainsPlatformViewBoundingRect(const SkRRect& clip_rrect,
 
   FML_DCHECK([[NSThread currentThread] isMainThread]);
```
hellohuanlin, line 868: thx for the chat just now. just to check if i understand it correctly (let me know if i miss anything): 

Dart layer sends separate messages (info about `views_to_dispose` and `composition_order`) to the engine. The `composition_order` contains what should be rendered on the current frame, and `views_to_dispose` contains views that needs to be disposed. We need separate messages because a platform view could be just hidden and not disposed, so it won't show up in `composition_order`. 

The crash happens when `views_to_dispose` message is delivered earlier than the new `composition_order` (due to the async nature of the message channel). So when rendering the current frame, a platform view is already disposed, but it still exists in `composition_order`. The fix here is simply to delay the `views_to_dispose` if it's still part of `composition_order`, and try again next frame. Alternative fix is to remove the view from `composition_order`, but we decided that `composition_order` should be the source of truth of the current frame.

hellohuanlin (approved):

shell/platform/darwin/ios/framework/Source/FlutterPlatformViews.mm
cyanglaz: Yes.

minseok-jeong-gn:
Is this only for iOS?

Can you guys check below issue too?
>>> https://github.com/flutter/flutter/issues/112542

I think in android side, app crash still happening.

Only iOS side after flutter stable version 3.13.x upgrade, app doesn't crash.

But in android still happening..

Please help me to examine the above issue...

PR #41525, opened 2023-04-26
[Impeller] Coerce opaque ColorSourceContents to Source

bdero:
This coercion happens when Entities are appended to a pass, which is the moment where we will need to perform depth sorting.

### DIFF
--- a/impeller/aiks/aiks_unittests.cc
+++ b/impeller/aiks/aiks_unittests.cc
@@ -2136,5 +2136,32 @@ TEST_P(AiksTest, CanRenderOffscreenCheckerboard) {
   ASSERT_TRUE(OpenPlaygroundHere(canvas.EndRecordingAsPicture()));
 }
 
+TEST_P(AiksTest, OpaqueEntitiesGetCoercedToSource) {
+  Canvas canvas;
+  canvas.Scale(Vector2(1.618, 1.618));
+  canvas.DrawCircle(Point(), 10,
+                    {
+                        .color = Color::CornflowerBlue(),
+                        .blend_mode = BlendMode::kSourceOver,
+                    });
+  Picture picture = canvas.EndRecordingAsPicture();
+
+  // Extract the SolidColorSource.
+  Entity entity;
+  std::shared_ptr<SolidColorContents> contents;
+  picture.pass->IterateAllEntities([&e = entity, &contents](Entity& entity) {
+    if (ScalarNearlyEqual(entity.GetTransformation().GetScale().x, 1.618f)) {
+      e = entity;
+      contents =
+          std::static_pointer_cast<SolidColorContents>(entity.GetContents());
+      return false;
+    }
+    return true;
+  });
+
+  ASSERT_TRUE(contents->IsOpaque());
+  ASSERT_EQ(entity.GetBlendMode(), BlendMode::kSource);
+}
+
 }  // namespace testing
 }  // namespace impeller
--- a/impeller/aiks/paint_pass_delegate.cc
+++ b/impeller/aiks/paint_pass_delegate.cc
@@ -7,6 +7,7 @@
 #include "impeller/entity/contents/contents.h"
 #include "impeller/entity/contents/texture_contents.h"
 #include "impeller/entity/entity_pass.h"
+#include "impeller/geometry/color.h"
 #include "impeller/geometry/path_builder.h"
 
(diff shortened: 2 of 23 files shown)


impeller/entity/entity_pass.cc
```diff
@@ -48,6 +48,11 @@ void EntityPass::SetDelegate(std::unique_ptr<EntityPassDelegate> delegate) {
 }
 
 void EntityPass::AddEntity(Entity entity) {
+  if (entity.GetBlendMode() == BlendMode::kSourceOver &&
+      entity.GetContents()->IsOpaque()) {
+    entity.SetBlendMode(BlendMode::kSource);
+  }
```
jonahwilliams, line 54: Does this happen before or after we do the opacity peephole check?

impeller/aiks/aiks_unittests.cc
```diff
@@ -2136,5 +2136,32 @@ TEST_P(AiksTest, CanRenderOffscreenCheckerboard) {
...
+                        .color = Color::CornflowerBlue(),
+                        .blend_mode = BlendMode::kSourceOver,
+                    });
+  Picture picture = canvas.EndRecordingAsPicture();
+
+  // Extract the SolidColorSource.
+  Entity entity;
+  std::shared_ptr<SolidColorContents> contents;
+  picture.pass->IterateAllEntities([&e = entity, &contents](Entity& entity) {
+    if (ScalarNearlyEqual(entity.GetTransformation().GetScale().x, 1.618f)) {
```
bdero, line 2153: I've been occasionally using this pattern as a sort of poor man's predicate test for inspecting what Aiks + EntityPass is doing, but I don't think it's a sustainable solution in the longterm. @chinmaygarde recall you writing a doc around this kind of thing many moons ago, but I can't seem to find it...

impeller/entity/entity_pass.cc
bdero: This happens before (when Aiks pushes an Entity into the pass).

Hmm, it looks like we weren't handling blend modes in the opacity peephole. I added some restrictions and made it so it's able to unfold things with the Source blend mode. There's a new utility where the `Entity` handles `SetInheritedOpacity` with the blend mode in mind. Working on test cases for this at the moment.

impeller/aiks/paint_pass_delegate.cc
```diff
@@ -72,7 +73,8 @@ bool OpacityPeepholePassDelegate::CanElide() {
 // |EntityPassDelgate|
 bool OpacityPeepholePassDelegate::CanCollapseIntoParentPass(
     EntityPass* entity_pass) {
-  if (paint_.color.alpha <= 0.0 || paint_.color.alpha >= 1.0 ||
+  if (paint_.blend_mode != BlendMode::kSourceOver ||
```
jonahwilliams, line 76: This is handled in https://github.com/flutter/engine/blob/main/impeller/aiks/canvas.cc#L390

impeller/entity/entity_pass.cc
bdero: Like if the pass has kLuminosity as its blend mode and the child entities have kHue, it'll incorrectly collapse the pass.

bdero: Ah nvm about the pass. I think the entities are still not being checked right, though.

jonahwilliams:
Seems reasonable to me, though I suppose its inevitable we're going to have to be careful about the ordering of these various optimizations. It would be good to note for future that we can only re-order the entities after doing the opacity peephole as that can change the opacity and blend mode too

impeller/aiks/paint_pass_delegate.cc
bdero: Alrighty, removed this and threw in a comment.

jonahwilliams (approved):
LGTM

bdero:
My initial plan is to just have all EntityPass [Elements](https://github.com/flutter/engine/blob/fbef3281e0041412c4b5abd29682f8a137a6cf14/impeller/entity/entity_pass.h#L28-L37) treated as default/translucent and defer solving this problem.

I think we'll want to come up with a longer term plan here for pass collapsing that involves some precomputation. Maybe wrangle prioritizing other easy collapsible situations if we find they're common enough, too. Like collapsing if the pass is SourceOver+Opaque and all of its children are SourceOver/Source.

PR #41533, opened 2023-04-26
Revert "[Impeller] Use a device buffer for SkBitmap allocation, use Linear texture on Metal backend."

jonahwilliams:
Reverts flutter/engine#41374

Breaks on Simulators!


jonahwilliams:
Two changes are needed:

1. device_buffer->AsTexture should do the right thing on simulator, by falling back performing a copy
2.SupportsSharedDeviceBufferTextureMemory should be false so that we do an update if the buffer changes

fluttergithubbot (approved):

zanderso (commented):
LGTM

dnfield:
I think this is the same reason my previous attempt at something sort of like this had to be reverted lol

PR #41543, opened 2023-04-27
[Impeller] allow shader read for root resolve texture

jonahwilliams:
This fixes a validation error I see from Xcode running the backdrop filter benchmarks


bdero (approved):
LGTM

PR #35608, opened 2022-08-22
Include TextureLayer in contents of toImageSync

jonahwilliams:
See revert https://github.com/flutter/engine/pull/35587

Fixes NPE in GetGrContext

PR #41559, opened 2023-04-27
Remove single window assumption from SceneBuilder

dkwingsmt:
This PR fixes https://github.com/flutter/flutter/issues/112202. `Scene` is no longer tied to views, but receives method arguments for view properties, clearing the path to multiview Flutter.

`Scene` no longer creates a `LayerTree` on construction, but only part of the config. When a output method is called (`toImage`, `toImageSync`, or `takeLayerTree`), the `LayerTree` is constructed on the spot. In this way, all `LayerTree`s return to being `unique_ptr`s instead of `shared_ptr`s, reverting part of https://github.com/flutter/engine/pull/35608 . And `Scene` no longer needs to disfunction after one `takeLayerTree`, since the layer tree config is never really taken away.

~~The `dart:ui` methods `Scene.toImage` and `Scene.toImageSync` receive a new parameter `pixelRatio` each. This parameter does not change the output dimension but only controls the deprecated physical shape models (`PhysicalShapeLayer`). We should also modify the framework to pass in the `devicePixelRatio` argument.~~

The `toImage` and `toImageSync` keeps the legacy behavior by fetching pixel ratio from window 0, or defaults to `2.0f`.

## Pre-launch Checklist

- [x] I signed the [CLA].


dkwingsmt:
@jonahwilliams This PR should be ready for review, can you take a look?

jonahwilliams:
> The dart:ui methods Scene.toImage and Scene.toImageSync receive a new parameter pixelRatio each. This parameter does not change the output dimension but only controls the deprecated physical shape models (PhysicalShapeLayer). We should also modify the framework to pass in the devicePixelRatio argument.

Please don't add parameters we'd need to support indefinitely for functionality we're going to remove in < 1 year. I would instead  default the physical shape layer to use a 2.0 DPR.

jonahwilliams:
Here is a draft PR to delete PML: https://github.com/flutter/engine/pull/41593

I'll see if this is landable in the near term.

dkwingsmt:
@jonahwilliams I've removed the extra parameter. Now they fetch the pixel ratio from window 0 with a fallback of 2.0f.

runtime/runtime_controller.cc
```diff
@@ -316,7 +316,17 @@ void RuntimeController::ScheduleFrame() {
 
 // |PlatformConfigurationClient|
 void RuntimeController::Render(Scene* scene) {
-  client_.Render(scene->takeLayerTree());
+  // TODO(dkwingsmt): Currently only supports a single window.
+  int64_t view_id = 0ll;
```
loic-sharma, line 320: Consider adding a constant for the implicit view ID: 

https://github.com/flutter/engine/blob/1ae848ce6b556ee471eca27d9eb8be33a829ba90/lib/ui/window/platform_configuration.cc#LL25C1-L25C1

lib/ui/compositing/scene.cc
```diff
@@ -126,15 +126,35 @@ void Scene::RasterizeToImage(uint32_t width,
...
+                                                 uint32_t height,
+                                                 float pixel_ratio) {
+  if (!valid()) {
+    return nullptr;
+  }
+  return std::make_unique<LayerTree>(layer_tree_config_,
+                                     SkISize::Make(width, height), pixel_ratio);
+}
+
+float Scene::defaultViewPixelRatio() {
```
loic-sharma, line 152: This adds a new single window assumption. Can we avoid this? Otherwise, please add a TODO comment

dkwingsmt: I've added a paragraph to explain it, and an issue to track the TODO.

runtime/runtime_controller.cc
dkwingsmt: Done.

dkwingsmt:
Also cc @goderbauer

jonahwilliams:
Looks like I missed once place , Canvas.cc actually depends on the device pixel ratio in Canvas::drawShadow:

```
  // Not using SafeNarrow because DPR will always be a relatively small number.
  SkScalar dpr = static_cast<float>(UIDartState::Current()
                                        ->platform_configuration()
                                        ->get_window(0)
                                        ->viewport_metrics()
                                        .device_pixel_ratio);

```

  We'll need to add support for providing the device pixel ratio on the framework side if this is used for real

jonahwilliams:
This is more or less just multiplied with the provided elevation:

https://github.com/flutter/engine/blob/main/impeller/display_list/dl_dispatcher.cc#L1113

The impeller code is based on the Skia code. So technically we could make this work today by multiplying elevation by DPR in the framework, but that would break folks using the canvas interface

dkwingsmt:
@jonahwilliams I've created an issue for `Canvas`: https://github.com/flutter/flutter/issues/125780

lib/ui/compositing/scene.cc
```diff
@@ -23,6 +23,8 @@
 
 namespace flutter {
 
+static constexpr float kFallbackPixelRatio = 2.0f;
```
goderbauer, line 26: This should probably have a comment answering questions like: Why is this 2.0? When do we fallback to this? Is this just meant to support legacy behavior and new logic should not use this constant? Or when should new logic use this constant?

goderbauer: (I see that this is kinda explained on defaultViewPixelRatio, might be worthwhile to link to that here)

lib/ui/compositing/scene.h
```diff
@@ -44,16 +46,36 @@ class Scene : public RefCountedDartWrappable<Scene> {
...
+  // method aims to keep the legacy behavior in single-window Flutter, which
+  // feeds the toImage and toImageSync with the pixel ratio of the only window.
   //
-  // No longer valid after calling `takeLayerTree`.
-  std::shared_ptr<flutter::LayerTree> layer_tree_;
+  // TODO(dkwingsmt): If PhysicalShapeLayer has been removed as well as
+  // {Preroll,Paint}Context.frame_device_pixel_ratio, remove this method and its
+  // related logic.
+  // https://github.com/flutter/flutter/issues/125720
+  static float defaultViewPixelRatio();
```
goderbauer, line 78: Can we give this a name that clearly indicated that this is legacy behavior and new code should not depend on this? Maybe we can replace `default` with `legacy` or something like that... The "default" should be to _not_ assume a specific view or DPR.

runtime/runtime_controller.cc
```diff
@@ -20,6 +20,8 @@
 
 namespace flutter {
 
+const uint64_t kFlutterDefaultViewId = 0llu;
```
goderbauer, line 23: Should this be "implicitView" instead of default view? We are trying to remove the annotation that flutter has a default view. In some modes, it may run with an implicitly created view, though...

goderbauer (commented):
Overall this makes sense to me, deferring to the engine experts for details.

PR #41563, opened 2023-04-28
[Impeller] Manage the onscreen stencil in EntityPass

bdero:
Moves onscreen stencil management from the surface to EntityPass for Metal and Vulkan. EntityPass has enough context to set up the stencil optimally.

Also forces us to fail loudly when the supplied stencil is set up in a way that can't be used for rendering the pass on GLES.

- On Vulkan, when we're blitting, we don't need to set up a onscreen stencil attachment at all.
- And on both Vulkan and Metal, the onscreen stencil can be transient when no pass reads are necessary (fixes the benchmark regressions introduced by https://github.com/flutter/engine/pull/41509).


jonahwilliams (approved):
LGTM

PR #41567, opened 2023-04-28
Revert "Reland: [Impeller] Use a device buffer for SkBitmap allocation, use Linear texture on Metal backend. "

bdero:
Reverts flutter/engine#41538

Playground atlas text is broken for the Vulkan + OpenGLES backends with this change. Maybe something's not getting zero initialized?

Before revert:

Screenshot 2023-04-27 at 7 56 34 PM

After revert:

Screenshot 2023-04-27 at 7 57 47 PM


jonahwilliams (approved):
LGTM!

bdero:
I think we're getting close to being able to capture Vulkan goldens now that most of the flakes are gone, so this should be an easy catch on CI soon (FYI @gaaclarke).

PR #41585, opened 2023-04-28
fuchsia: do not read version_history.json

maruel:
Otherwise this causes GN to read JSON files that are irrelevant.


zanderso (approved):
rslgtm

maruel:
https://ci.chromium.org/b/8782561224204679553 seems to be a flake?

zanderso:
Landing over the `Mac clang-tidy` flake (https://github.com/flutter/flutter/issues/119750), and the golden tests that will not start as a result of that flake. This change could not affect `Mac clang-tidy` results or goldens.

PR #41593, opened 2023-04-28
Remove physical model layer

jonahwilliams:
Removes the physical model layer and associated engine code. This was already deprecated and removed in the framework. By removing it in the engine, we can also remove the need for layer tree diff/paint/preroll to have the device pixel ratio. This will simplify some of the multi-view work

Fixes https://github.com/flutter/flutter/issues/125720


dkwingsmt:
Related: https://github.com/flutter/flutter/issues/125720

dkwingsmt:
Shall we also delete `{Preroll,Paint}Context.frame_device_pixel_ratio`? (This is the actual goal.)

PR #41600, opened 2023-04-28
[Impeller] Remove destructive blend mode check from Contents::ShouldRender

bdero:
Resolves https://github.com/flutter/flutter/issues/125717.
The "destructive" check is only relevant for turning off SaveLayer shrinkwrapping (used for the `cover_whole_screen_` property in EntityPass), and shouldn't be used to avoid culling entities.


jonahwilliams (approved):
LGTM

PR #41606, opened 2023-04-29
[Impeller] take advantage of DisplayList culling

flar:
Switching the calls to dispatch into an Impeller Dispatcher to use a cull rect to enable pre-culling of the out-of-bounds ops.

This change showed an improvement of around 2x on the rendering performance of the non-intersecting platform view benchmark, but that was measured without the recent changes to the destructive blend modes in Impeller renderer.

### DIFF
--- a/impeller/aiks/canvas.cc
+++ b/impeller/aiks/canvas.cc
@@ -23,7 +23,7 @@
 namespace impeller {
 
 Canvas::Canvas() {
-  Initialize(Rect::MakeLTRB(-1e9, -1e9, 1e9, 1e9));
+  Initialize(std::nullopt);
 }
 
 Canvas::Canvas(Rect cull_rect) {
@@ -37,11 +37,11 @@ Canvas::Canvas(IRect cull_rect) {
 
 Canvas::~Canvas() = default;
 
-void Canvas::Initialize(Rect cull_rect) {
+void Canvas::Initialize(std::optional<Rect> cull_rect) {
   initial_cull_rect_ = cull_rect;
   base_pass_ = std::make_unique<EntityPass>();
   current_pass_ = base_pass_.get();
-  xformation_stack_.emplace_back(CanvasStackEntry{.clip_bounds = cull_rect});
+  xformation_stack_.emplace_back(CanvasStackEntry{.cull_rect = cull_rect});
   lazy_glyph_atlas_ = std::make_shared<LazyGlyphAtlas>();
   FML_DCHECK(GetSaveCount() == 1u);
   FML_DCHECK(base_pass_->GetSubpassesDepth() == 1u);
@@ -64,7 +64,7 @@ void Canvas::Save(
     std::optional<EntityPass::BackdropFilterProc> backdrop_filter) {
   auto entry = CanvasStackEntry{};
   entry.xformation = xformation_stack_.back().xformation;
-  entry.clip_bounds = xformation_stack_.back().clip_bounds;
+  entry.cull_rect = xformation_stack_.back().cull_rect;
   entry.stencil_depth = xformation_stack_.back().stencil_depth;
   if (create_subpass) {
     entry.is_subpass = true;
@@ -120,9 +120,13 @@ const Matrix& Canvas::GetCurrentTransformation() const {
   return xformation_stack_.back().xformation;
 }
 
-const Rect Canvas::GetCurrentLocalClipBounds() const {
-  Matrix inverse = xformation_stack_.back().xformation.Invert();
-  return xformation_stack_.back().clip_bounds.TransformBounds(inverse);
+const std::optional<Rect> Canvas::GetCurrentLocalCullingBounds() const {
+  auto cull_rect = xformation_stack_.back().cull_rect;
+  if (cull_rect.has_value()) {
+    Matrix inverse = xformation_stack_.back().xformation.Invert();
+    cull_rect = cull_rect.value().TransformBounds(inverse);
+  }
+  return cull_rect;
 }
 
 void Canvas::Translate(const Vector3& offset) {
@@ -273,22 +277,61 @@ void Canvas::DrawCircle(Point center, Scalar radius, const Paint& paint) {
 }
 
 void Canvas::ClipPath(const Path& path, Entity::ClipOperation clip_op) {
-  ClipGeometry(Geometry::MakeFillPath(path), clip_op, path.GetBoundingBox());
+  ClipGeometry(Geometry::MakeFillPath(path), clip_op);
+  if (clip_op == Entity::ClipOperation::kIntersect) {
+    auto bounds = path.GetBoundingBox();
+    if (bounds.has_value()) {
+      IntersectCulling(bounds.value());
+    }
+  }
 }
 
 void Canvas::ClipRect(const Rect& rect, Entity::ClipOperation clip_op) {
-  ClipGeometry(Geometry::MakeRect(rect), clip_op, rect);
+  ClipGeometry(Geometry::MakeRect(rect), clip_op);
+  switch (clip_op) {
+    case Entity::ClipOperation::kIntersect:
+      IntersectCulling(rect);
+      break;
+    case Entity::ClipOperation::kDifference:
+      SubtractCulling(rect);
+      break;
+  }
 }
 
 void Canvas::ClipRRect(const Rect& rect,
                        Scalar corner_radius,
                        Entity::ClipOperation clip_op) {
-  ClipGeometry(Geometry::MakeRRect(rect, corner_radius), clip_op, rect);
+  ClipGeometry(Geometry::MakeRRect(rect, corner_radius), clip_op);
+  switch (clip_op) {
+    case Entity::ClipOperation::kIntersect:
+      IntersectCulling(rect);
+      break;
+    case Entity::ClipOperation::kDifference:
+      if (corner_radius <= 0) {
+        SubtractCulling(rect);
+      } else {
+        // We subtract the inner "tall" and "wide" rectangle pieces
+        // that fit inside the corners which cover the greatest area
+        // without involving the curved corners
+        // Since this is a subtract operation, we can subtract each
+        // rectangle piece individually without fear of interference.
+        if (corner_radius * 2 < rect.size.width) {
+          SubtractCulling(Rect::MakeLTRB(
+              rect.GetLeft() + corner_radius, rect.GetTop(),
+              rect.GetRight() - corner_radius, rect.GetBottom()));
+        }
+        if (corner_radius * 2 < rect.size.height) {
+          SubtractCulling(Rect::MakeLTRB(
+              rect.GetLeft(), rect.GetTop() + corner_radius,  //
+              rect.GetRight(), rect.GetBottom() - corner_radius));
+        }
+      }
+      break;
+  }
 }
 
 void Canvas::ClipGeometry(std::unique_ptr<Geometry> geometry,
-                          Entity::ClipOperation clip_op,
-                          std::optional<Rect> geometry_bounds) {
+                          Entity::ClipOperation clip_op) {
   auto contents = std::make_shared<ClipContents>();
   contents->SetGeometry(std::move(geometry));
   contents->SetClipOperation(clip_op);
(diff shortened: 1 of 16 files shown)


flar:
A/B Comparison of Framework ToT against a local engine that includes both https://github.com/flutter/flutter/issues/125717 and this PR:

```
                   Score                     Average A (noise) Average B (noise) Speed-up
average_frame_build_time_millis                   0.56 (0.00%)      0.46 (0.00%)  1.22x  
worst_frame_build_time_millis                     3.04 (0.00%)      4.26 (0.00%)  0.71x  
90th_percentile_frame_build_time_millis           0.96 (0.00%)      0.65 (0.00%)  1.48x  
99th_percentile_frame_build_time_millis           2.93 (0.00%)      2.94 (0.00%)  1.00x  
average_frame_rasterizer_time_millis             55.05 (0.00%)     25.20 (0.00%)  2.18x  
worst_frame_rasterizer_time_millis              105.66 (0.00%)     62.40 (0.00%)  1.69x  
90th_percentile_frame_rasterizer_time_millis     72.62 (0.00%)     31.63 (0.00%)  2.30x  
99th_percentile_frame_rasterizer_time_millis    104.08 (0.00%)     54.28 (0.00%)  1.92x  
```

shell/gpu/gpu_surface_metal_impeller.mm
```diff
@@ -118,8 +118,13 @@
...
 
-        impeller::DlDispatcher impeller_dispatcher;
-        display_list->Dispatch(impeller_dispatcher);
+        if (!buffer_damage.has_value()) {
+          auto size = surface->GetSize();
+          buffer_damage = SkIRect::MakeWH(size.width, size.height);
+          clip_rect = impeller::IRect::MakeXYWH(0, 0, size.width, size.height);
+        }
+        impeller::DlDispatcher impeller_dispatcher(clip_rect.value());
+        display_list->Dispatch(impeller_dispatcher, buffer_damage.value());
```
jonahwilliams, line 127: I implemented this a bit differently for impeller, rather than adding a clip - i added a translate and then shrunk the size of the root render pass, which causes impeller to cull.

translate: https://github.com/flutter/engine/blob/main/flow/compositor_context.cc#L189

Shrink: https://github.com/flutter/engine/blob/main/impeller/renderer/backend/metal/surface_mtl.mm#L57

I'm not sure if this clip would already do the right thing though.

flar: > which causes impeller to cull.

Causes it to cull where?

impeller::Dispatcher::DrawRect immediately calls Canvas::DrawRect without any tests.
Canvas::DrawRect (sometimes does something different for strokes and mask blurs) immediately creates and adds an entity without any other checks.

Where is the culling done if that rectangle was so far outside the clip that it isn't visible? The Dispatch method on DL can do this kind of bounds culling and then the impeller::Dispatcher never even sees the op...

impeller/aiks/canvas.cc
```diff
@@ -109,6 +120,11 @@ const Matrix& Canvas::GetCurrentTransformation() const {
   return xformation_stack_.back().xformation;
 }
 
+const Rect Canvas::GetCurrentLocalClipBounds() const {
+  Matrix inverse = xformation_stack_.back().xformation.Invert();
+  return xformation_stack_.back().clip_bounds.TransformBounds(inverse);
+}
```
bdero, line 126: Might be good to add a regular test (doesn't use `OpenPlaygroundHere`) in `aiks_unittests` which uses this method to verify the intersection.

bdero:
Impeller already tracks the clip bounds in EntityPass at render time, but this change will prevent lots of unnecessary API traffic from hitting Impeller in the first place when platform views are being used, which seems like a good idea to me.

I've been strongly considering pulling some of the culling behavior out into Aiks anyhow so that we can throw away culled things earlier, and so I'm not opposed to tracking a screen space coverage rectangle for the clip in the transform stack.

impeller/aiks/canvas.cc
```diff
@@ -23,15 +23,25 @@
 namespace impeller {
 
 Canvas::Canvas() {
-  Initialize();
+  Initialize(Rect::MakeLTRB(-1e9, -1e9, 1e9, 1e9));
```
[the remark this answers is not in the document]
flar: How well does MakeMaximum deal with being transformed?

impeller/aiks/canvas.cc
flar: Is there a way I can have a test that creates a wide ranging DL, then dispatches it to an impeller Dispatcher with a cull_rect (and one that does that with a sub-DisplayList as well), and then checks that there are only Entities for the ops that should have made it through the culling?

impeller/entity/entity_pass.h
```diff
@@ -236,6 +236,7 @@ class EntityPass {
 
 struct CanvasStackEntry {
   Matrix xformation;
+  Rect clip_bounds;
```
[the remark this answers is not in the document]
flar: One question, does "coverage" imply exact coverage? This would be a conservative estimate of just the bounds of the device clip.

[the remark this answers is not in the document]
flar: I moved this struct to canvas.h since it's not used from anywhere else, and how do you feel about "cull_rect" for a name since it is only used for culling, not for clipping?

impeller/aiks/canvas.cc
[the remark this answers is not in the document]
flar: Rather than worry about specifying a huge clip that might encourage degenerate math, I changed the cull_rect to an std::optional.

impeller/aiks/canvas.cc
flar: I created a new `canvas_unittests` file and wrote a whole suite of tests which found several bugs in Rect::CutOut().

flar:
Another run of the Impeller non-intersecting platform view benchmark on the latest commit produced similar results to before:

```
                   Score                     Average A (noise) Average B (noise) Speed-up
average_frame_build_time_millis                   0.56 (0.00%)      0.48 (0.00%)  1.16x  
worst_frame_build_time_millis                     3.30 (0.00%)      4.72 (0.00%)  0.70x  
90th_percentile_frame_build_time_millis           0.93 (0.00%)      0.88 (0.00%)  1.06x  
99th_percentile_frame_build_time_millis           3.11 (0.00%)      3.61 (0.00%)  0.86x  
average_frame_rasterizer_time_millis             56.10 (0.00%)     25.27 (0.00%)  2.22x  
worst_frame_rasterizer_time_millis              105.56 (0.00%)     61.11 (0.00%)  1.73x  
90th_percentile_frame_rasterizer_time_millis     73.32 (0.00%)     31.10 (0.00%)  2.36x  
99th_percentile_frame_rasterizer_time_millis    103.67 (0.00%)     54.79 (0.00%)  1.89x  
```

shell/gpu/gpu_surface_metal_impeller.mm
flar: I'm not sure this code here can apply a translation. If there is no damage rect, the cull rect is the size of the surface so it is already sized correctly and 0-based. If there is a damage rect, then the remainder of the surface must remain intact so we need to render the culled ops "in situ" with respect to origin and surface size.

While I still don't think I see all of the relevance of your comment, I don't think we can do translations and resizes of surfaces here because of the nature of this code and the damage rect...?

impeller/aiks/canvas.cc
flar: With respect to making sure that the Entities and Passes are culled by the cull rect, we already have tests of the DisplayList RTree mechanism that make sure that the culled `Dispatch` variant correctly omits ops in the playback, and with the new tests in `canvas_unittests.cc` that check that Impeller's `Canvas` manages the cull rect according to the desired behavior - is there anything left to test here?

jonahwilliams:
I tried running this on the Flutter gallery (the old one in flutter/flutter). On the text fields demo, the blinking cursor no longer shows up correctly and the console is filled with messages of the form:

```
dispatched with culling rect on DL with no rtree
```

shell/gpu/gpu_surface_metal_impeller.mm
flar: Ding ding! With your later comment that the blinking cursor was missing I think I finally got what you are saying. Unlike the other platforms that reuse the original framebuffer surface to render the partial damage inside a clip in the middle of the surface, Impeller renders the damage to a reduced size surface and then blits it into place?

In that case then the Impeller cull rect should always be set to just the rect(size) value with no origin, but the Dispatch call should be called with the damage rect since it is in the coordinate space of the original frame?

jonahwilliams: Makes sense to me

flar:
> I tried running this on the Flutter gallery (the old one in flutter/flutter). On the text fields demo, the blinking cursor no longer shows up correctly and the console is filled with messages of the form:
> 
> ```
> dispatched with culling rect on DL with no rtree
> ```

I'll have to see where we are creating these DLs without an rtree, but the blinking cursor is likely because I didn't understand what you were saying above about how Impeller manages partial repaints. I'm curious if my latest reply there indicates if I'm on the same page.

impeller/aiks/canvas.cc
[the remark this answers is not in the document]
bdero: SGTM

impeller/entity/entity_pass.h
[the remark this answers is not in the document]
bdero: Nah, "coverage" is always assumed to be a best effort minimum in Impeller. "cull_rect" SGTM.

impeller/aiks/canvas.cc
bdero: I'd say this level of test is beyond the call of action. 😆

shell/gpu/gpu_surface_metal_impeller.mm
flar: Looking again, it looks like the translate happens before the frame DisplayList is generated, so both rects should be 0-origin-based?

And, doing all of that fixes the blinking cursor problem...

bdero (approved):
LGTM! (Modulo the blinking cursor issue mentioned by Jonah)

flar:
The latest version seems to fix both the blinking cursor and "DL with no RTree" problems.

impeller/aiks/canvas.cc
flar: Well, given that it found errors in an existing geom method, isn't it just "at the call of duty"?

bdero: We already cover rect methods in geometry_unittests.cc. It sounds like I probably missed a case there.

flar:
I have been going through these results. The impact on the non-intersecting platform view benchmark was less than I measured on my iPhone 6 by quite a lot.

I did not expect any impact on any benchmarks that didn't use Impeller, though. All of the changes were in code specific to Impeller, so the massive improvement on the backdrop perf benchmark was unexpected. I can see that it would benefit from the culling, though, as the animation is not over the backdrop entries so with proper culling they will not be attempted on each frame. But, that would only be impacted by this PR if the benchmark was run using Impeller...?

jonahwilliams:
I think the  much earlier culling is driving most of those improvements. Since the dirty region itself is fairly small (just the animating rect). Arguably we should turn of partial repaint for those benchmarks...


For the non-intersecting platform views, there is not partial repaint so it seems like the additional culling may not be significant once we fixed Impeller to not encode those extra instructions? They're just draw rects so its fairly cheap.

I think the opportunity for that benchmark is in recognizing that the overlay surfaces are empty. To take advantage of that, culling isn't sufficient - we need to skip the entire surface.

flar:
> I think the much earlier culling is driving most of those improvements. Since the dirty region itself is fairly small (just the animating rect). Arguably we should turn of partial repaint for those benchmarks...

I need to go back and check out the backdrop perf, but I think it was originally testing the caching of BDFLayers and that was before partial repaints. The current situation makes an excellent test of partial repaints, though. If we "fixed" the benchmark to continue rendering the BDFLayers by using a larger animating rect (or one above and one below), we'd then be measuring Impeller's lack of layer caching.

flar:
> For the non-intersecting platform views, there is not partial repaint so it seems like the additional culling may not be significant once we fixed Impeller to not encode those extra instructions? They're just draw rects so its fairly cheap.

Actually, isn't there partial repaint, but it isn't very significant? The scrolling list is the only part changing so the header would not need to be redrawn...?

Either way, there is culling because before we didn't cull to anything and now we cull at least to the size of the surface which skips all of the data that is outside of the surface. I haven't looked closely at what is generated by the benchmark, but depending on the "overdraw" of the scrolling list, tons of non-visible rects may have been rendered on each frame.

jonahwilliams:
Partial repaint is disabled with platform views. Also neither backend raster cached backdrop filters

flar:
> Partial repaint is disabled with platform views.

Good point. So the gain there is the granularity of the culling.

> Also neither backend raster cached backdrop filters

When the benchmark was written it was to measure the performance of backdrop filters in order to track the effects of caching them. There was a prototype, but getting it right ended up being harder than expected and so the work was abandoned, but the benchmark remains. At the time we decided that it was good to keep it to measure the performance of that operation even if our short-term plans didn't pan out as we might have other ways of improving it.

PR #41634, opened 2023-05-01
Forward fix for roll of Dart SDK to version with new checked-in SDK

whesse:
The checked-in SDK downloaded by Dart and engine checkouts from cipd has changed.
The cipd package no longer includes dart-sdk at the start of all the paths of its files, so /dart-sdk is added to the path of the cipd download location.

The script that automatically copies DEPS changes from the Dart SDK to engine does not handle the download location changing, and drops that entry.

This change copies the changed download location and cipd reference from Dart SDK DEPS to engine. A roll of the Dart SDK to a version that does not include this update will fail by removing the changed line, so the next successful roll after this is landed must include Dart SDK commit 88aeaa4fec387ce6993b0bfe54d0d33c41d4554a.

Link to the Dart SDK commit is https://dart.googlesource.com/sdk.git/+/88aeaa4fec387ce6993b0bfe54d0d33c41d4554a


whesse:
I didn't realize that rebasing the PR started all the checks again - they already ran and finished on the first commit.

PR #41642, opened 2023-05-01
Clone the parent Android engine's asset provider into a spawned engine

jason-simmons:
Typically the AndroidShellHolder's apk_asset_provider_ is set by AndroidShellHolder::Launch when the DartExecutor runs the engine's entrypoint.

But if the engine was started by Spawn, then the apk_asset_provider_ was not being set.  This would cause a crash if the resulting engine was later used to spawn another engine.

Fixes https://github.com/flutter/flutter/issues/122364


shell/platform/android/android_shell_holder.cc
reidbaker, line 1: Does this need a test?

reidbaker (approved):
Rubber stamp.

Hixie:
Do we test the Spawn codepaths in general?

jason-simmons:
Added a test for this in the Android scenario_app

reidbaker (approved):

PR #41646, opened 2023-05-01
Run flutter license script with dart `--intrepret_irregexp` runtime option

aam:
For extremely large RegExp's that are used in flutter license script, running interpreter-based implementation is faster than running generated unoptimized code. For normal-sized RegExp's optimized code is expected to be faster.

Local testing shows 5x speed up on full license script run (311s vs 1430s).

Should help with https://github.com/flutter/flutter/issues/118193


ci/licenses.sh
```diff
@@ -72,10 +72,10 @@ function run_tests() (
 # Runs in a subshell.
 function collect_licenses() (
   cd "$SRC_DIR/flutter/tools/licenses"
-  dart --enable-asserts lib/main.dart         \
-    --src ../../..                            \
-    --out ../../../out/license_script_output  \
-    --golden ../../ci/licenses_golden \
+  dart --enable-asserts --interpret_irregexp lib/main.dart \
```
zanderso, line 75: Awesome. Please add a comment here that explains why the flag is needed and include the url for the GitHub issue. Thanks!

zanderso (approved):

PR #41656, opened 2023-05-02
Populate fallback fonts from CIPD in tests

eyebrowsoffire:
This fixes https://github.com/flutter/flutter/issues/85793

The script that generates the font fallback data also rolls a package to CIPD with these files and updates the DEPS to download those files from CIPD. This makes sure that these tests are hermetic.


mdebbar (approved):
Thanks for refactoring CIPD utilities too!

zanderso (commented):
Would you mind throwing some documentation under https://github.com/flutter/engine/tree/main/tools/cipd explaining how the CIPD package is created and uploaded? Thanks!

lib/web_ui/lib/src/engine/canvaskit/font_fallbacks.dart
```diff
@@ -5,21 +5,23 @@
...
-import 'noto_font.dart';
+import 'package:ui/src/engine.dart';
 
 /// Global static font fallback data.
 class FontFallbackData {
+
+  factory FontFallbackData() =>
+    FontFallbackData._(getFallbackFontData(configuration.useColorEmoji));
+
+  FontFallbackData._(List<NotoFont> fallbackFonts) :
```
mdebbar, line 16:
```suggestion
  FontFallbackData._(this.fallbackFonts) :
```

lib/web_ui/lib/src/engine/canvaskit/font_fallbacks.dart
```diff
@@ -39,10 +41,20 @@ class FontFallbackData {
   /// Code units which are known to be covered by at least one fallback font.
   final Set<int> knownCoveredCodeUnits = <int>{};
 
+  final List<NotoFont> fallbackFonts = getFallbackFontData(configuration.useColorEmoji);
```
mdebbar, line 44:
```suggestion
  final List<NotoFont> fallbackFonts;
```

zanderso (approved):

mdebbar (approved):

lib/web_ui/lib/src/engine/canvaskit/font_fallbacks.dart
eyebrowsoffire: Good catch! Thanks.

lib/web_ui/test/canvaskit/fallback_fonts_golden_test.dart
```diff
@@ -430,7 +428,10 @@ void testMain() {
         }
       }
     });
-  }, skip: isSafari);
+  },
+  skip: isSafari,
+  // These tests can take a long time, particularly when compiled to wasm.
+  timeout: const Timeout.factor(4));
```
Hixie, line 434: changing the timeout violates our style guide

DEPS
```diff
@@ -913,6 +913,16 @@ deps = {
     ],
     'condition': 'download_impeller_cmake_example and host_os == "mac"',
     'dep_type': 'cipd',
+  },
+
+  'src/third_party/flutter_font_fallbacks': {
```
Hixie, line 918: we might want to call these something that more clearly indicates their provenance (Google Fonts) and use (testing only), e.g. `google_fonts_for_unit_tests` or something

PR #41669, opened 2023-05-02
[fuchsia] Stop calling FIDL from Dart in Flutter integration tests

ianloic:
As we move the Dart FIDL bindings we need to stop calling FIDL methods
from Dart in the Flutter repo. This only affects some integration tests.

This moves the FIDL calls into C++ and uses platform messages to make
the functionality available to the Dart tests.


shell/platform/fuchsia/flutter/tests/integration/touch-input/embedding-flutter-view/lib/embedding-flutter-view.dart
```diff
@@ -10,7 +10,6 @@ import 'dart:ui';
 import 'package:args/args.dart';
 import 'package:fidl_fuchsia_ui_app/fidl_async.dart';
```
arbreng, line 11: How are you going to handle "fuchsia.ui.app" and "fuchsia.ui.views"?

I think we could do:
* "_launchGfxChildView" connects to the incoming ViewProvider service, so it can be made into a platform message.
* The ViewHolderToken and ViewRef/ViewRefControl types can all be encoded as integers in the platform message (cast the zx_handle_t to an integer).  This is already what we do for the "View.create" message on the platform views channel so there is precedent. When the associated dart EventPair/etc objects die, they will call zx_handle_close but then ignore the error that gets returned.

shell/platform/fuchsia/flutter/platform_view.cc
```diff
@@ -140,6 +143,47 @@ PlatformView::PlatformView(
       std::move(pointerinjector_registry), std::move(view_ref_clone),
       is_flatland);
 
+  if (svc) {
```
arbreng, line 146: Add a comment here indicating this is only for tests

arbreng (approved):

shell/platform/fuchsia/flutter/platform_view.cc
```diff
@@ -673,4 +720,143 @@ bool PlatformView::HandleFuchsiaShaderWarmupChannelPlatformMessage(
   return true;
 }
 
+namespace {
```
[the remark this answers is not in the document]
ianloic: There wasn't a `namespace{}` block at the top but I moved this block up to the top so now there is.

shell/platform/fuchsia/flutter/tests/integration/touch-input/embedding-flutter-view/lib/embedding-flutter-view.dart
ianloic: I'm going to update this PR with those changes too. Basically I expose a native function over platform messages that mints a child view (either gfx or flatland based on an argument) and returns the view id as an integer.

arbreng (approved):

PR #41681, opened 2023-05-03
[web] dialog a11y fixes

yjbanov:
Fixes two issues in dialog accessibility:

* Fixes https://github.com/flutter/flutter/issues/45207 by setting `role="dialog"` on nodes that have both `scopesRoute` and `namesRoute` set. There's no guarantee that this combination of flags is an actual dialog, but it's close enough, and it makes the screen reader announce the appearance of the dialog. Note that `scopesRoute` alone is not sufficient, because Flutter uses it for overlays that are not semantically dialogs, such as dismiss barriers.
* Fixes an issue with focus management, where focus fails to transfer to background content after the dialog is dismissed. This happened because `EngineSemanticsOwner._semanticsTree` retained descendants of parents that were removed. This is benign in many cases. However, for focus this is problematic because the HTML element can go away and come back (losing focus along the way), but its corresponding `SemanticsObject` is never marked as "dirty" and fails to update and request focus.

I'm hoping this is sufficient to fix b/251839784 as well. Will work with the relevant team to find out.

### DIFF
--- a/lib/web_ui/lib/src/engine/semantics/dialog.dart
+++ b/lib/web_ui/lib/src/engine/semantics/dialog.dart
@@ -20,8 +20,8 @@ class Dialog extends RoleManager {
 
   @override
   void update() {
+    final String? label = semanticsObject.label;
     assert(() {
-      final String? label = semanticsObject.label;
       if (label == null || label.trim().isEmpty) {
         printWarning(
           'Semantic node ${semanticsObject.id} was assigned dialog role, but '
@@ -32,7 +32,7 @@ class Dialog extends RoleManager {
       }
       return true;
     }());
-    semanticsObject.element.setAttribute('aria-label', semanticsObject.label ?? '');
+    semanticsObject.element.setAttribute('aria-label', label ?? '');
     semanticsObject.setAriaRole('dialog', true);
   }
 }
--- a/lib/web_ui/lib/src/engine/semantics/dialog.dart
+++ b/lib/web_ui/lib/src/engine/semantics/dialog.dart
@@ -1,0 +1,35 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import '../dom.dart';
+import '../semantics.dart';
+import '../util.dart';
+
+class Dialog extends RoleManager {
+  Dialog(SemanticsObject semanticsObject) : super(Role.dialog, semanticsObject);
+
+  @override
+  void dispose() {
+    semanticsObject.element.removeAttribute('aria-label');
+    semanticsObject.clearAriaRole();
+  }
+
+  @override
+  void update() {
+    assert(() {
+      final String? label = semanticsObject.label;
+      if (label == null || label.trim().isEmpty) {
+        printWarning(
+          'Semantic node ${semanticsObject.id} was assigned dialog role, but '
+          'is missing a label. A dialog should contain a label so that a '
+          'screen reader can communicate to the user that a dialog appeared '
+          'and a user action is requested.'
+        );
+      }
+      return true;
+    }());
+    semanticsObject.element.setAttribute('aria-label', semanticsObject.label ?? '');
+    semanticsObject.setAriaRole('dialog', true);
+  }
+}
--- a/lib/web_ui/lib/src/engine/semantics/dialog.dart
+++ b/lib/web_ui/lib/src/engine/semantics/dialog.dart
@@ -6,6 +6,9 @@ import '../dom.dart';
 import '../semantics.dart';
 import '../util.dart';
 
+/// Provides accessibility for dialogs.
+///
+/// See also [Role.dialog].
 class Dialog extends RoleManager {
   Dialog(SemanticsObject semanticsObject) : super(Role.dialog, semanticsObject);
 
--- a/lib/web_ui/lib/src/engine/semantics/semantics.dart
+++ b/lib/web_ui/lib/src/engine/semantics/semantics.dart
@@ -17,6 +17,7 @@ import '../platform_dispatcher.dart';
 import '../util.dart';
 import '../vector_math.dart';
 import 'checkable.dart';
+import 'dialog.dart';
 import 'image.dart';
 import 'incrementable.dart';
 import 'label_and_value.dart';
@@ -356,6 +357,18 @@ enum Role {
   /// with this role, they will be able to get the assistive technology's
   /// attention right away.
   liveRegion,
+
+  /// Adds the "dialog" ARIA role to the node.
+  ///
+  /// This corresponds to a semantics node that has both `scopesRoute` and
+  /// `namesRoute` bits set. While in Flutter a named route is not necessarily a
+  /// dialog, this is the closest analog on the web.
+  ///
+  /// Why is `scopesRoute` alone not sufficient? Because Flutter can create
+  /// routes that are not logically dialogs and there's nothing interesting to
+  /// announce to the user. For example, a modal barrier has `scopesRoute` set
+  /// but marking it as a dialog would be wrong.
+  dialog,
 }
 
 /// A function that creates a [RoleManager] for a [SemanticsObject].
@@ -370,6 +383,7 @@ final Map<Role, RoleManagerFactory> _roleFactories = <Role, RoleManagerFactory>{
   Role.checkable: (SemanticsObject object) => Checkable(object),
   Role.image: (SemanticsObject object) => ImageRoleManager(object),
   Role.liveRegion: (SemanticsObject object) => LiveRegion(object),
+  Role.dialog: (SemanticsObject object) => Dialog(object),
 };
 
 /// Provides the functionality associated with the role of the given
@@ -845,6 +859,15 @@ class SemanticsObject {
       !hasAction(ui.SemanticsAction.tap) &&
       !hasFlag(ui.SemanticsFlag.isButton);
 
+  /// Whether this node should be treated as an ARIA dialog.
+  ///
+  /// See also [Role.dialog].
+  bool get isDialog {
+    final bool scopesRoute = hasFlag(ui.SemanticsFlag.scopesRoute);
+    final bool namesRoute = hasFlag(ui.SemanticsFlag.namesRoute);
+    return scopesRoute && namesRoute;
+  }
+
   /// Whether this object carry enabled/disabled state (and if so whether it is
   /// enabled).
   ///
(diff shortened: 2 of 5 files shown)


jonahwilliams:
Technically the ways these two flags are supposed to work does not require them to be on the same node. All pages, which I suppose probably includes modal dialogs, introduce a scopes route flag. Then walking the tree is depth first order until namesRoute flag determines the announced name

yjbanov:
@jonahwilliams That's interesting. It sounds like a `scopesRoute` widget can be a `role="dialog"` that points to the `namesRoute` using `aria-labelledby` ([Example](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/dialog_role#labeling)). What do you think?

I can add it in a follow-up PR as the case of the two being on the same node vs separate nodes are mutually exclusive and use different ARIA properties.

jonahwilliams:
That sounds like it would work

lib/web_ui/lib/src/engine/semantics/semantics.dart
```diff
@@ -1393,6 +1420,16 @@ class SemanticsObject {
     }
   }
 
+  /// Recursively visits the tree rooted at `this` node in depth-first fashion.
+  ///
+  /// Calls the [callback] for `this` node, then for all of its descendants.
+  void visitDepthFirst(void Function(SemanticsObject) callback) {
+    callback(this);
```
ditman, line 1427: Should `callback` return a `bool` so you abort the the drilling into children if `callback(this)` returns `false`?

ditman (approved):
If CI is happy, I can't find anything egregious either! LGTM!

lib/web_ui/lib/src/engine/semantics/semantics.dart
```diff
@@ -1513,20 +1552,30 @@ class EngineSemanticsOwner {
...
-        object.element.remove();
-      } else {
-        assert(object._parent == parent);
-        assert(object.element.parentNode == parent._childContainerElement);
+    for (final SemanticsObject detachmentRoot in _detachments) {
+      // A detached node may or may not have some of its descendants reattached
+      // elsewhere. Walk the descendant tree and find all descendants that were
+      // reattached to a parent. Those descendants need to be removed.
+      final List<SemanticsObject> removals = <SemanticsObject>[];
+      detachmentRoot.visitDepthFirst((SemanticsObject node) {
```
mdebbar, line 1560: This is traversing the entire tree including all descendants of detached nodes. Do you think it's worth making a short circuit to avoid traversing children of detached nodes?

It can be done by returning a boolean from the callback to indicate whether you want to continue down that subtree or not.

mdebbar (approved):

lib/web_ui/lib/src/engine/semantics/dialog.dart
```diff
@@ -0,0 +1,38 @@
...
+        printWarning(
+          'Semantic node ${semanticsObject.id} was assigned dialog role, but '
+          'is missing a label. A dialog should contain a label so that a '
+          'screen reader can communicate to the user that a dialog appeared '
+          'and a user action is requested.'
+        );
+      }
+      return true;
+    }());
+    semanticsObject.element.setAttribute('aria-label', semanticsObject.label ?? '');
```
htoor3, line 35: Do you think we should we have a more descriptive fallback `aria-label` other than an empty string if no label is provided? Or do we generally defer to the dev via warnings like on line 26?

lib/web_ui/test/engine/semantics/semantics_test.dart
```diff
@@ -2086,6 +2175,73 @@ void _testGroup() {
...
+    );
+
+    semantics().updateSemantics(builder.build());
+    expect(
+      warnings,
+      <String>[
+        'Semantic node 0 was assigned dialog role, but is missing a label. A dialog should contain a label so that a screen reader can communicate to the user that a dialog appeared and a user action is requested.',
+      ],
+    );
+
```
htoor3, line 2240: nit: we can also assert that the dialog `aria-role` was applied here despite the warning

htoor3 (approved):
lgtm

lib/web_ui/lib/src/engine/semantics/semantics.dart
```diff
@@ -1828,7 +1877,23 @@ class EngineSemanticsOwner {
...
+      root.visitDepthFirst((SemanticsObject child) {
+        liveIds.add(child.id);
+      });
+      if (!_semanticsTree.keys.every(liveIds.contains)) {
+        throw AssertionError(
+          'The semantics node map is inconsistent:\n'
+          '  Nodes in tree: [${liveIds.join(', ')}]\n'
+          '  Nodes in map : [${_semanticsTree.keys.join(', ')}]'
+        );
+      }
```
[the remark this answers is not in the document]
yjbanov: Because I was under the false impression that assertion error expressions are evaluated eagerly, and this one's really expensive. However, I was proven wrong.

Done.

lib/web_ui/lib/src/engine/semantics/dialog.dart
```diff
@@ -0,0 +1,38 @@
...
+  @override
+  void dispose() {
+    semanticsObject.element.removeAttribute('aria-label');
+    semanticsObject.clearAriaRole();
+  }
+
+  @override
+  void update() {
+    assert(() {
+      final String? label = semanticsObject.label;
```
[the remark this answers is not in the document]
yjbanov: Done.

lib/web_ui/lib/src/engine/semantics/semantics.dart
```diff
@@ -1241,7 +1264,11 @@ class SemanticsObject {
   /// Detects the roles that this semantics object corresponds to and manages
   /// the lifecycles of [SemanticsObjectRole] objects.
   void _updateRoles() {
```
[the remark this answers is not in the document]
yjbanov: It can't be the responsibility of each role manager, because each node may have multiple roles and which ones are assigned depends on adjacent flag values. It's easier to have this logic all in one place. Right now this method is <30 LoC, which is quite manageable.

However, you made me think! I wonder if we could assign a "primary" role, which would be 1:1 with each object, and then the primary role manager can add extra "secondary" roles. One thing that doesn't work well with the current design is it's unclear who is responsible for setting the `role` attribute. Any role manager can set it, which can conflict with another role manager. But in a primary/secondary model only the primary role manager would be responsible for it.

Filed https://github.com/flutter/flutter/issues/126384

lib/web_ui/lib/src/engine/semantics/semantics.dart
```diff
@@ -1484,17 +1521,19 @@ class EngineSemanticsOwner {
...
   /// Declares that the [SemanticsObject] with the given [id] was detached from
   /// its current parent object.
   ///
   /// The object will be detached permanently unless it is reattached via the
   /// [_attachObject] method.
-  void _detachObject(int? id) {
+  void _detachObject(int id) {
     assert(_semanticsTree.containsKey(id));
     final SemanticsObject? object = _semanticsTree[id];
-    _detachments.add(object);
```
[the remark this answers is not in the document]
yjbanov: Done.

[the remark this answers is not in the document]
yjbanov: Didn't add a message since this is an internal assertion. Developers shouldn't ever see it.

lib/web_ui/lib/src/engine/semantics/semantics.dart
```diff
@@ -1513,20 +1552,30 @@ class EngineSemanticsOwner {
...
+    for (final SemanticsObject detachmentRoot in _detachments) {
+      // A detached node may or may not have some of its descendants reattached
+      // elsewhere. Walk the descendant tree and find all descendants that were
+      // reattached to a parent. Those descendants need to be removed.
+      final List<SemanticsObject> removals = <SemanticsObject>[];
+      detachmentRoot.visitDepthFirst((SemanticsObject node) {
+        final SemanticsObject? parent = _attachments[node.id];
+        if (parent == null) {
+          // Was not reparented and is removed permanently from the tree.
+          removals.add(node);
```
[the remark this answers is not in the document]
yjbanov: Yep! That's one of the main issues that I'm fixing. When a parent is removed, we didn't remove children from the map. So next time children were readded (after the the dialog is dismissed), we started with old children containing wrong state information. We have to remove them from the map so next time we create fresh new objects for them.

lib/web_ui/lib/src/engine/semantics/semantics.dart
yjbanov: See https://github.com/flutter/engine/pull/41681#discussion_r1189263649

lib/web_ui/lib/src/engine/semantics/semantics.dart
```diff
@@ -1484,17 +1521,19 @@ class EngineSemanticsOwner {
...
   /// The object will be detached permanently unless it is reattached via the
   /// [_attachObject] method.
-  void _detachObject(int? id) {
+  void _detachObject(int id) {
     assert(_semanticsTree.containsKey(id));
     final SemanticsObject? object = _semanticsTree[id];
-    _detachments.add(object);
+    if (object != null) {
+      _detachments.add(object);
+    }
```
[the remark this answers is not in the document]
yjbanov: I'm playing it safe. My worry is that we can get something inconsistent from the framework. Cecause child nodes are supplied as `Uint8List` of child IDs, and their nodes separately in their own `SemanticsUpdate` objects, the consistency is merely a hand-shake. The API does not enforce it. An assertion should be effective at keeping most bugs away, but we don't want to crash in production if one sneaked in.

lib/web_ui/lib/src/engine/semantics/semantics.dart
yjbanov: See https://github.com/flutter/engine/pull/41681#discussion_r1189263649

lib/web_ui/lib/src/engine/semantics/dialog.dart
yjbanov: We can't because labels are part of the UI, which needs to be internationalized. Error messages and warnings in Flutter, OTOH, are all in English. So I think a warning and missing label is probably a better fallback.

lib/web_ui/test/engine/semantics/semantics_test.dart
yjbanov: Good catch! Done.

PR #41685, opened 2023-05-03
Add a `Display` API to dart:ui that reports the physical size, DPR, and refresh rate of the main display

dnfield:
Fixes https://github.com/flutter/flutter/issues/123307 - for Android, iOS, and Web, for the main display only (https://github.com/flutter/flutter/issues/125938 tracks supporting multiple displays, https://github.com/flutter/flutter/issues/125939 for desktop).

Desktop will need to be implemented for this, but given priority for a couple of our customers targetting foldable devices on Android I'm inclined to get this in before desktop can be finished.

The main concern for this right now is that on some Android foldable devices, setting a preferred orientation will cause letterboxing and the `MediaQuery` will _never_ get the full screen size when unfolded. This causes apps to think the screen is smaller than it is, as they've mainly been using `MediaQueryData.size` to figure this out. Android's recommendation is to not set a preferred orientation, and if you must to use the new method introduced in `ViewUtil.java` to calculate the maximal window size.


dnfield:
I'm going to try to add some tests for this tomorrow.

dnfield:
This now has test coverage for VM Dart and Android.

@cyanglaz or @jmagman  - I think this gets some coverage from scenario_app on iOS, but I can't think of a good way to unit test the changes to the iOS embedding more particularly. Open to suggestions.

cyanglaz:
> This now has test coverage for VM Dart and Android.
> 
> @cyanglaz or @jmagman - I think this gets some coverage from scenario_app on iOS, but I can't think of a good way to unit test the changes to the iOS embedding more particularly. Open to suggestions.

You can prob test if `updateDisplays` is called during `setViewController` and `setupShell`. Similar to how you did it for android.

dnfield:
> > This now has test coverage for VM Dart and Android.
> > @cyanglaz or @jmagman - I think this gets some coverage from scenario_app on iOS, but I can't think of a good way to unit test the changes to the iOS embedding more particularly. Open to suggestions.
> 
> You can prob test if `updateDisplays` is called during `setViewController` and `setupShell`. Similar to how you did it for android.

How can I do that on the object itself though? Ther'es some weird ObjC way to spy on a method of a real class right? I can't remember.

lib/ui/hooks.dart
```diff
@@ -4,6 +4,28 @@
 
 part of dart.ui;
 
+@pragma('vm:entry-point')
+void _updateDisplays(
+  List<int> ids,
+  List<double> sizes,
+  List<double> devicePixelRatios,
+  List<double> refreshRates,
+) {
```
loic-sharma, line 13: Consider adding an assert that all parameters' lengths are expected

lib/ui/hooks.dart
```diff
@@ -4,6 +4,28 @@
 
 part of dart.ui;
 
+@pragma('vm:entry-point')
+void _updateDisplays(
+  List<int> ids,
+  List<double> sizes,
```
loic-sharma, line 10: What do you think of using different parameters for widths/heights? It makes the size logic a bit simpler. I don't feel strongly about this

lib/ui/window/platform_configuration.h
```diff
@@ -268,10 +269,15 @@ class PlatformConfiguration final {
   void DidCreateIsolate();
 
   //----------------------------------------------------------------------------
-  /// @brief      Update the specified locale data in the framework.
+  /// @brief      Update the specified display data in the framework.
   ///
-  /// @deprecated The persistent isolate data must be used for this purpose
-  ///             instead.
+  /// @param[in]  locale_data  The locale data. This should consist of groups of
+  ///             4 strings, each group representing a single locale.
```
loic-sharma, line 275: Don't forget to update this comment.

shell/platform/embedder/embedder.cc
```diff
@@ -2998,17 +2998,12 @@ FlutterEngineResult FlutterEngineNotifyDisplayUpdate(
...
-        if (embedder_displays[i].single_display) {
-          displays.push_back(std::make_unique<flutter::Display>(
-              embedder_displays[i].refresh_rate));
-        } else {
-          displays.push_back(std::make_unique<flutter::Display>(
-              embedder_displays[i].display_id,
-              embedder_displays[i].refresh_rate));
-        }
+        displays.push_back(std::make_unique<flutter::Display>(
+            embedder_displays[i].display_id, embedder_displays[i].refresh_rate,
```
loic-sharma, line 3002: FYI, this should use the `struct_size` to iterate through the array dynamically: https://github.com/flutter/flutter/issues/121352

I don't think it's strictly necessary for this change as you haven't added any new members to `FlutterEngineDisplay`... but please double-check me on that.

lib/ui/window.dart
```diff
@@ -67,6 +98,12 @@ class FlutterView {
     return platformDispatcher._viewConfigurations[viewId]!;
   }
 
+  /// The [Display] this view is drawn in.
+  Display get display {
+    assert(platformDispatcher._displays.containsKey(_viewConfiguration.displayId));
```
loic-sharma, line 103: Will this crash if the embedder doesn't provide display information? (I'm not sure Windows, macOS, or Linux provide display information)

dnfield: Right now they provide a display that is not quite valid. But yes, the assert would fire if the embedder failed to provide a display.

lib/ui/window/platform_configuration.h
dnfield: Done

lib/ui/window.dart
loic-sharma: Should we make this property nullable and force the caller to have some fallback logic if display information isn't available?

shell/platform/embedder/embedder.cc
dnfield: No changes to FlutterEngineDisplay yet. That'll be in the resolution to https://github.com/flutter/flutter/issues/125939

lib/ui/hooks.dart
dnfield: Done

lib/ui/hooks.dart
dnfield: Done

shell/platform/embedder/embedder.cc
```diff
@@ -2998,17 +2998,12 @@ FlutterEngineResult FlutterEngineNotifyDisplayUpdate(
...
+            // TODO(dnfield): Supply real values.
```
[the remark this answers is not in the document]
dnfield: Added issue link.

shell/platform/embedder/embedder.cc
loic-sharma: Sounds good, marking this as resolved

dnfield:
@cyanglaz I figured out what you were suggesting, I think, and added a test. I've also made a couple small adjustments to make some other tests that don't quite treat the FlutterEngine the way a normal application would do not crash.

lib/ui/window.dart
dnfield: I think we should just implement it. Making it nullable now and non-nullable later is harder once people start using it (it results in compilation errors, which make the migration in google's monorepo harder). I don't expect it to take long to implement the desktop portion of this, I just don't want to delay this patch further and it's already fairly large.

lib/ui/platform_dispatcher.dart
```diff
@@ -149,6 +149,14 @@ class PlatformDispatcher {
     _onPlatformConfigurationChangedZone = Zone.current;
   }
 
+  /// The current list of displays.
+  ///
+  /// If any of their configurations change, [onMetricsChanged] will be called.
+  ///
```
goderbauer, line 155: Maybe document some of the current platform-specific limitations? Presumably, this is just empty for desktops right now?

lib/ui/window.dart
goderbauer: So, what does this currently return for unimplemented platforms?

Have we verified that this is implementable on all desktop platforms we want to support? Wayland comes to mind as a platform that may not be able to tell us what display we are on (I haven't verified this, this is just from what I have heard). If that turns out to be true, what would we expect this getter to do? For that case, it may be nicer if this can just be null.

lib/ui/window.dart
```diff
@@ -67,6 +98,12 @@ class FlutterView {
...
+    return platformDispatcher._displays[_viewConfiguration.displayId]!;
+  }
+
   /// The number of device pixels for each logical pixel for the screen this
```
goderbauer, line 107: There is a little bit of duplicated information available now as dpr as also available via the display getter now. It's probably not an issue, but a little strange? Presumably, both will always have the same value?

dnfield: DPR really belongs on the display. I'm half tempted to deprecate it on View, but that seems like it'd create a lot of unnecessary churn.

But yes, DPR on the view will always match DPR from the display.

lib/ui/window.dart
dnfield: Right now it's returning -1 for the DPR, width, and height, and a real refresh rate/display ID.

lib/ui/platform_dispatcher.dart
dnfield: I added some docs.

lib/ui/window.dart
dnfield: On Linux, we can indeed get the width and height of at least the display the application is rendering to, and probably all displays. Even if Wayland doesn't have an API for this, you have to tell Wayland to connect to a particular display adapter, and it will be possible to query the information about the display size from that adapter.  GTK also has an API for this.

dnfield:
There still seem to be a couple failures in the iOS testing, will look into that more today.

lib/ui/platform_dispatcher.dart
goderbauer: The same caveat should likely also be documented on the FlutterView.display property?

lib/ui/window.dart
dnfield: Sorry, I meant GDK :) What the Linux embedding uses for windowing etc.

dnfield:
All my remaining failures were the flaky tests that got disabled in 41740 - this is ready for review

lib/ui/platform_dispatcher.dart
```diff
@@ -149,6 +149,26 @@ class PlatformDispatcher {
...
+  /// Platforms may limit what information is available to the application with
+  /// regard to secondary displays and/or displays that do not have an active
+  /// application window.
+  ///
+  /// Presently, on Android and Web this collection will only contain the
+  /// display that the current window is on. On iOS, it will only contains the
+  /// main display on the phone or tablet. On Desktop, it will contain only
+  /// a main display with a valid refresh rate but invalid size and device
+  /// pixel ratio values.
```
reidbaker, line 166: Is it possible to make these values missing or null instead of invalid (which I assume means incorrect).

lib/ui/window.dart
reidbaker: You might want to add to the dart doc to DPR on view that says window is the prefered way if you indeed think that is the case.

reidbaker (approved):

lib/ui/platform_dispatcher.dart
dnfield: I don't want to make them null, since it will be harder to migrate them away from nullable later if people start treating them as nullable (e.g. they add `!` all over and the compiler refuses to compiile things when we switch it).

The plan is to implement them for desktop sooner than later.

lib/ui/window.dart
dnfield: You need the DPR if you want to figure out logical pixel sizes. I'll document that they're the same.

dnfield: Added some docs.

PR #41694, opened 2023-05-03
[Web+Linux] Fix pressing Meta keys throws

bleroux:
## Description

This PR fixes Meta keys throwing exception on Chrome Linux.

The assertions throws because the DOM event sent when Meta keys is pressed is not coherent when Meta is the only pressed key.

## Related Issue

Fixes https://github.com/flutter/flutter/issues/125672

## Tests

Adds 1 test.

### DIFF
--- a/lib/web_ui/lib/src/engine/raw_keyboard.dart
+++ b/lib/web_ui/lib/src/engine/raw_keyboard.dart
@@ -6,6 +6,7 @@ import 'dart:async';
 import 'dart:typed_data';
 
 import '../engine.dart'  show registerHotRestartListener;
+import 'browser_detection.dart';
 import 'dom.dart';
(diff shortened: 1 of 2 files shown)


dkwingsmt:
Generally looks good. Can we limit this behavior on Linux? Generally I'd like the exception rules to be as strict as possible unless unnecessary. (Also if this rule is too loose, doesn't it make non-Linux Web app generate fewer events?

bleroux:
> Can we limit this behavior on Linux? Generally I'd like the exception rules to be as strict as possible unless unnecessary. 

Sure, that makes sense, I updated the PR to limit this behavior to Linux.

dkwingsmt (approved):
LGTM, thanks for taking care of it!

PR #41699, opened 2023-05-03
[web] Improve null safety for color->css

mdebbar:
Turning `colorToCssString` into an extension method on `ui.Color` improves null safety and developer ergonomics.

### DIFF
--- a/lib/web_ui/lib/src/engine/util.dart
+++ b/lib/web_ui/lib/src/engine/util.dart
@@ -339,20 +339,15 @@ bool rectContainsOther(ui.Rect rect, ui.Rect other) {
       rect.bottom >= other.bottom;
 }
 
-/// Converts color to a css compatible attribute value.
-String? colorToCssString(ui.Color? color) {
-  if (color == null) {
-    return null;
+extension CssColor on ui.Color {
+  /// Converts color to a css compatible attribute value.
+  String toCssString() {
+    return colorValueToCssString(value);
   }
-  final int value = color.value;
-  return colorValueToCssString(value);
 }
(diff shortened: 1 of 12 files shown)


yjbanov (approved):

lib/web_ui/lib/src/engine/util.dart
```diff
@@ -339,20 +339,15 @@ bool rectContainsOther(ui.Rect rect, ui.Rect other) {
...
-  final int value = color.value;
-  return colorValueToCssString(value);
 }
 
 // Converts a color value (as an int) into a CSS-compatible value.
-String? colorValueToCssString(int? value) {
-  if (value == null) {
-    return null;
-  }
+String colorValueToCssString(int value) {
```
ditman, line 350: Should this be made private? Or do we still have usages of this method where value is a `0xFFFABADA`?

ditman (commented):
Looks good to me!

lib/web_ui/lib/src/engine/util.dart
mdebbar: This is still used in other files when we don't have a `ui.Color` instance but we have an `int` representation of the color. Examples:
https://github.com/flutter/engine/blob/4d507067285936553780967e3fda4dabd8258fc3/lib/web_ui/lib/src/engine/html/bitmap_canvas.dart#L575
https://github.com/flutter/engine/blob/9322c14c1d59a3c957f34f3f25c9f15b3c5a3016/lib/web_ui/lib/src/engine/canvas_pool.dart#L1000

PR #41700, opened 2023-05-03
[web] Remove unused function parameter

mdebbar:
The `hasAutofillOverlay` and `cssSelectorPrefix` parameters aren't really used in any useful way. Let's remove them.

### DIFF
--- a/lib/web_ui/lib/src/engine/global_styles.dart
+++ b/lib/web_ui/lib/src/engine/global_styles.dart
@@ -9,8 +9,6 @@ import 'text_editing/text_editing.dart';
 // Applies the required global CSS to an incoming [DomCSSStyleSheet] `sheet`.
 void applyGlobalCssRulesToSheet(
   DomHTMLStyleElement styleElement, {
-  required bool hasAutofillOverlay,
-  String cssSelectorPrefix = '',
   required String defaultCssFont,
 }) {
   // TODO(web): use more efficient CSS selectors; descendant selectors are slow.
@@ -24,7 +22,7 @@ void applyGlobalCssRulesToSheet(
   //
   // Fixes #115216 by ensuring that our parameters only affect the flt-scene-host children.
   sheet.insertRule('''
-    $cssSelectorPrefix flt-scene-host {
+    flt-scene-host {
       color: red;
       font: $defaultCssFont;
     }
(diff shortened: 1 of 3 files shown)


lib/web_ui/lib/src/engine/global_styles.dart
```diff
@@ -9,8 +9,6 @@ import 'text_editing/text_editing.dart';
 // Applies the required global CSS to an incoming [DomCSSStyleSheet] `sheet`.
 void applyGlobalCssRulesToSheet(
   DomHTMLStyleElement styleElement, {
-  required bool hasAutofillOverlay,
-  String cssSelectorPrefix = '',
```
mdebbar, line 13: Well, turns out the `cssSelectorPrefix` param is used by the new text editing host node:
https://github.com/flutter/engine/blob/d2dce25cc686189ceff73c9e02d7409ddb87aca7/lib/web_ui/lib/src/engine/embedder.dart#L437

I have 2 questions:
1. Is that necessary because it's outside of the shadow dom?
2. Since the styles are scoped to `flutter-view`, should we attach it somewhere global so that it applies to all `flutter-view`s?


cc @htoor3 @ditman

htoor3: Isn't it more necessary now that it's out of the shadowDOM? I thought one of the big gains we got from the shadowDOM was that the styles don't leak and affect things outside of the app's context.  Now that inputs are outside of the shadowDOM, the selector prefix makes sure we're only applying these styles to flutter app inputs. 

Option 2 sounds like a good idea.  Also having multiple `flutter-view`s means multiple `<style>` elements get created unnecessarily currently.  Having one style sheet at the top level would probably be a cleaner approach.

mdebbar: > Isn't it more necessary now that it's out of the shadowDOM? I thought one of the big gains we got from the shadowDOM was that the styles don't leak and affect things outside of the app's context. Now that inputs are outside of the shadowDOM, the selector prefix makes sure we're only applying these styles to flutter app inputs.

Alright, cool, that's what I thought too. So you are saying "yes" to my first question :)


> Option 2 sounds like a good idea. Also having multiple `flutter-view`s means multiple `<style>` elements get created unnecessarily currently. Having one style sheet at the top level would probably be a cleaner approach.

Yep. I'll leave it as is for now. I'm happy to revisit and do this when we start figuring things out for multi-view.

mdebbar: Done. I put the `cssSelectorPrefix` parameter back. Now the PR is only removing the `hasAutofillOverlay` parameter.

htoor3 (approved):
lgtm

PR #41702, opened 2023-05-03
Reland: "Determine lifecycle by looking at window focus also" (#41094)

gspencergoog:
## Description

This reverts commit 9183bff56244bf8d9700fb1625bb99d8849f31c4 to re-land #41094 because the Google test failures have been fixed. There are no changes to the original PR, since the fixes were in the Google code.

### DIFF
--- a/lib/ui/platform_dispatcher.dart
+++ b/lib/ui/platform_dispatcher.dart
@@ -1642,31 +1642,61 @@ class FrameTiming {
 /// States that an application can be in.
 ///
 /// The values below describe notifications from the operating system.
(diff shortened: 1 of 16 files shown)


yaakovschectman (approved):
Looks fine, but not what I'm most familiar with. Maybe it would be a good idea to request some of the original PR's reviewers as well

gspencergoog:
Okay, sure.  Mainly this is a rubber-stamp because nothing changed from the original PR.

reidbaker (approved):

PR #41711, opened 2023-05-03
[Impeller] Remove the DrawPaint->Clear optimization

bdero:
Fixes https://github.com/flutter/flutter/issues/125906.
Fixes https://github.com/flutter/flutter/issues/125908.

Going to add this back once I'm able to work out why the clear colors aren't working on GL/Vulkan. Stepping though with a debugger for both GLES and Vulkan, the clear colors seems to be getting piped through as expected. But the frame captures reveal that they're actually not.


jonahwilliams (approved):
LGTM RIP

PR #41720, opened 2023-05-04
Make run_tests.py assert that the ios test dylib is at least as new as libFlutter.dylib

dnfield:
This would have helped me figure out why I wasn't getting the same failures as on CI more quickly.

### DIFF
--- a/testing/run_tests.py
+++ b/testing/run_tests.py
@@ -629,6 +629,14 @@ def ensure_ios_tests_are_built(ios_out_dir):
   assert os.path.exists(tmp_out_dir
                        ) and os.path.exists(ios_test_lib), final_message
 
+  ios_test_lib_time = os.path.getmtime(ios_test_lib)
+  flutter_dylib = os.path.join(tmp_out_dir, 'libFlutter.dylib')
+  flutter_dylib_time = os.path.getmtime(flutter_dylib)
+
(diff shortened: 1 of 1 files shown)


jmagman (commented):
What if there were no changes to the libFlutter but I'm iterating on the tests themselves?

dnfield:
> What if there were no changes to the libFlutter but I'm iterating on the tests themselves?

Then the test dylib will be newer and this will pass.

testing/run_tests.py
```diff
@@ -629,6 +629,15 @@ def ensure_ios_tests_are_built(ios_out_dir):
   assert os.path.exists(tmp_out_dir
                        ) and os.path.exists(ios_test_lib), final_message
 
+  ios_test_lib_time = os.path.getmtime(ios_test_lib)
+  flutter_dylib = os.path.join(tmp_out_dir, 'libFlutter.dylib')
+  flutter_dylib_time = os.path.getmtime(flutter_dylib)
+
+  final_message = '%s is older than %s. Please run the following commands: \n%s' % (
```
zanderso, line 636: `autoninja` is deprecated. Could you also change line 625 to be `'ninja -C ...'` instead?

zanderso (approved):

testing/run_tests.py
dnfield: done

PR #41744, opened 2023-05-04
[Impeller] check if path is Rect/RRect/Oval

jonahwilliams:
This avoids tessellation if the application draws a Rect or RREct with Path.addRect or Path.addRREct.

### DIFF
--- a/impeller/display_list/dl_dispatcher.cc
+++ b/impeller/display_list/dl_dispatcher.cc
@@ -895,11 +895,15 @@ void DlDispatcher::drawDRRect(const SkRRect& outer, const SkRRect& inner) {
 void DlDispatcher::drawPath(const SkPath& path) {
   SkRect rect;
   SkRRect rrect;
+  SkRect oval;
   if (path.isRect(&rect)) {
     canvas_.DrawRect(skia_conversions::ToRect(rect), paint_);
   } else if (path.isRRect(&rrect) && rrect.isSimple()) {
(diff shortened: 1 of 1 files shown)


impeller/display_list/dl_dispatcher.cc
```diff
@@ -893,7 +893,16 @@ void DlDispatcher::drawDRRect(const SkRRect& outer, const SkRRect& inner) {
...
 void DlDispatcher::drawPath(const SkPath& path) {
-  canvas_.DrawPath(skia_conversions::ToPath(path), paint_);
+  SkRect rect;
+  SkRRect rrect;
+  if (path.isRect(&rect)) {
+    canvas_.DrawRect(skia_conversions::ToRect(rect), paint_);
+  } else if (path.isRRect(&rrect) && rrect.isSimple()) {
+    canvas_.DrawRRect(skia_conversions::ToRect(rrect.rect()),
+                      rrect.getSimpleRadii().fX, paint_);
+  } else {
```
bdero, line 903: DrawCircle would also push us into the rrect fast path if the color source is solid and there's a mask blur.

bdero (approved):

impeller/display_list/dl_dispatcher.cc
jonahwilliams: Done

jonahwilliams:
Its really only a performance optimization. I expect that we'll see _something_ in our benchmarks change with this...

PR #41748, opened 2023-05-04
[Impeller] introduces DeviceHolder to avoid accessing a dead Device

gaaclarke:
An alternative to https://github.com/flutter/engine/pull/41527

This introduces a new protocol called `DeviceHolder` whose only purpose is to return a pointer to a device.  That way people can hold onto weak_ptr's to the `DeviceHolder` and know to abort further work once the Device is deleted (like loading pipelines on a background thread).  This PR also changes the return type of ContextVK::GetDevice from returning a reference to a pointer as a hope to better communicate what is happening and discourage copying of the vk::Device.

I would actually recommend abolishing holding onto instance variables of the vk::Device since it isn't clear that that is just a raw pointer.  That's sort of how we got into this problem.

fixes https://github.com/flutter/flutter/issues/125522
fixes https://github.com/flutter/flutter/issues/125519
~~fixes https://github.com/flutter/flutter/issues/125571~~

Testing: Fixes flake in existing tests.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/impeller/renderer/backend/vulkan/command_pool_vk.cc
+++ b/impeller/renderer/backend/vulkan/command_pool_vk.cc
@@ -15,10 +15,6 @@
 namespace impeller {
 
 using CommandPoolMap = std::map<uint64_t, std::shared_ptr<CommandPoolVK>>;
-
-// TODO(https://github.com/flutter/flutter/issues/125571): This is storing tons
-// of CommandPoolVK's in the test runner since many contexts are created on the
-// same thread. We need to come up with a different way to clean these up.
 FML_THREAD_LOCAL fml::ThreadLocalUniquePtr<CommandPoolMap> tls_command_pool;
 
 static Mutex g_all_pools_mutex;
@@ -52,6 +48,9 @@ std::shared_ptr<CommandPoolVK> CommandPoolVK::GetThreadLocal(
 }
 
 void CommandPoolVK::ClearAllPools(const ContextVK* context) {
+  if (tls_command_pool.get()) {
+    tls_command_pool.get()->erase(context->GetHash());
+  }
   Lock pool_lock(g_all_pools_mutex);
   if (auto found = g_all_pools.find(context); found != g_all_pools.end()) {
     for (auto& weak_pool : found->second) {
--- a/impeller/renderer/backend/vulkan/command_pool_vk.cc
+++ b/impeller/renderer/backend/vulkan/command_pool_vk.cc
@@ -74,7 +74,7 @@ CommandPoolVK::CommandPoolVK(const ContextVK* context)
 
   pool_info.queueFamilyIndex = context->GetGraphicsQueue()->GetIndex().family;
   pool_info.flags = vk::CommandPoolCreateFlagBits::eTransient;
-  auto pool = context->GetDevice().createCommandPoolUnique(pool_info);
+  auto pool = (*context->GetDevice())->createCommandPoolUnique(pool_info);
   if (pool.result != vk::Result::eSuccess) {
     return;
   }
@@ -113,7 +113,7 @@ vk::UniqueCommandBuffer CommandPoolVK::CreateGraphicsCommandBuffer() {
   alloc_info.commandPool = graphics_pool_.get();
   alloc_info.commandBufferCount = 1u;
   alloc_info.level = vk::CommandBufferLevel::ePrimary;
-  auto [result, buffers] = device_.allocateCommandBuffersUnique(alloc_info);
+  auto [result, buffers] = (*device_)->allocateCommandBuffersUnique(alloc_info);
   if (result != vk::Result::eSuccess) {
     return {};
   }
--- a/impeller/renderer/backend/vulkan/command_pool_vk.cc
+++ b/impeller/renderer/backend/vulkan/command_pool_vk.cc
@@ -74,7 +74,7 @@ CommandPoolVK::CommandPoolVK(const ContextVK* context)
 
   pool_info.queueFamilyIndex = context->GetGraphicsQueue()->GetIndex().family;
   pool_info.flags = vk::CommandPoolCreateFlagBits::eTransient;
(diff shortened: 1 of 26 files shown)


impeller/renderer/backend/vulkan/device_holder.h
```diff
@@ -0,0 +1,17 @@
...
+#pragma once
+
+#include "impeller/renderer/backend/vulkan/vk.h"
+
+namespace impeller {
+
+class DeviceHolder {
+ public:
+  virtual ~DeviceHolder() = default;
+  virtual const vk::Device* GetDevice() const = 0;
```
jason-simmons, line 14: `DeviceHolder::GetDevice()` should return a `vk::Device` (instead of a `vk::Device*`).  And code that uses a `vk::Device` should pass instances (or const references) rather than pointers.

Passing handle wrappers like `vk::Device` by value is consistent with how the vulkan-hpp API itself uses these objects.

IIUC the intent of this change was to make the C++ type reflect the ownership semantics.  But developers of this subsystem are going to need to understand the behavior of the Khronos Vulkan C++ wrappers and that a `vk::Device` instance is equivalent to a Vulkan device handle.

The `const vk::Device*` inaccurately implies that the `vk::Device` is nullable, and creates the potential for someone to be holding onto a `vk::Device*` pointing to a C++ object that has been deleted.

impeller/renderer/backend/vulkan/pipeline_cache_vk.h
```diff
@@ -8,14 +8,16 @@
...
 #include "impeller/renderer/backend/vulkan/vk.h"
 
 namespace impeller {
 
 class PipelineCacheVK {
  public:
   explicit PipelineCacheVK(std::shared_ptr<const Capabilities> caps,
-                           vk::Device device,
+                           std::weak_ptr<DeviceHolder> device_holder,
+                           const vk::Device* device,
```
jason-simmons, line 20: Remove the `device` - it's redundant with the `device_holder`

impeller/renderer/backend/vulkan/command_pool_vk.cc
```diff
@@ -52,6 +48,9 @@ std::shared_ptr<CommandPoolVK> CommandPoolVK::GetThreadLocal(
 }
 
 void CommandPoolVK::ClearAllPools(const ContextVK* context) {
+  if (tls_command_pool.get()) {
```
jason-simmons, line 51: Remove this - it's unrelated to the rest of this PR

gaaclarke: Done, i'll file a new PR.  It was related in that the question was "will we be able to solve the leak once we go down this route, since we know we can with the alternative route".  So the answer is yes.

impeller/renderer/backend/vulkan/pipeline_cache_vk.h
gaaclarke: It's not redundant unfortunately.  The way we initialize the `ContextVK` the `DeviceHolder` is not valid at the time we call `PipelineCacheVK()`.  It's because in the ContextVK constructor all data is held on the stack and only if everything is successful are the ivar's written to and `is_valid_` is set to true.

Our options were:
1) Send in the Device and avoid calling into the DeviceHolder
1) Stop using the Device in the PipelineCacheVK constructor
1) Write the Device to the ivar of ContextVK but make a RAII wrapper that unsets it if we don't get to `is_valid_ = true`.

I went for the first option.  I can add a comment to make it more clear.  The second option is probably a good design but is a risky change (I'd rather do that as a followup if we wanted to).  The 3rd option seemed to have just as much baggage as "don't use the DeviceHolder if you have the Device".

impeller/renderer/backend/vulkan/device_holder.h
gaaclarke: The reason I went with this approach is that the whole reason we got into this mess is that people were copying the `vk::Device` and not realize what they actually had is a raw pointer that may be invalid.  By passing around a `std::weak_ptr<DeviceHolder>` which returns a pointer to the device, my hope is that this would make that error much less likely.

I can make DeviceHolder return a `const vk::Device&` but that is going to make it so much easier for someone to copy the `vk::Device` without realizing the semantics of it since `vk::Device` has a copy constructor.

Alternatively, I might be able to do `virtual const vk::UniqueDevice& GetDevice() const = 0`. Which would give you the guarantees you want and make accidentally messing things up less likely.

Here's where I want to be: no one has an lvalue to a `vk::Device`.

jason-simmons: If other modules in this subsystem are interacting directly with Vulkan devices, then they will need to understand how Vulkan devices work and how their lifetime is managed.  For now I think the most idiomatic option is having `DeviceHolder::GetDevice` return a `const vk::Device&` and establishing a convention that anyone who may need the Vulkan device holds a `weak_ptr<DeviceHolder>` and should never hold a raw `vk::Device` instance.

The type system will not be able to prevent all forms of misuse as long as the other modules require direct access to the `vk::Device`.  If the convention above is insufficient, then we should hide all usage of the `vk::Device` behind some other interface.

impeller/renderer/backend/vulkan/pipeline_cache_vk.h
jason-simmons: A comment SGTM.  Later we can look into ways to make construction of `PipelineCacheVK` cleaner.

impeller/renderer/backend/vulkan/device_holder.h
gaaclarke: I think it's a loaded gun allowing `vk::Device&`.  I'd be down with making our own wrapper in the future.  For now I put in the references as requested.

gaaclarke:
@jason-simmons PTAL

jason-simmons (approved):

PR #41754, opened 2023-05-04
[Impeller] remove SDF code paths.

jonahwilliams:
From my investigations, we're not likely to be able to use SDF in the near term. To make planned refactors of the text rendering easier to land, I've moved the primary piece of code for generating the SDFs into a new TU for testing.

The shader itself has been deleted so that we don't ship + register it.


dnfield:
Why not just delete it entirely? It'll be in the git history if we really need it.

jonahwilliams:
I was keeping it around for youuuu 😂. Deleting sgtm

dnfield (approved):

PR #41761, opened 2023-05-05
[Impeller] Do not free a Vulkan command buffer if its command pool has already been destroyed

jason-simmons:
Fixes https://github.com/flutter/flutter/issues/125519

### DIFF
--- a/impeller/renderer/backend/vulkan/command_encoder_vk.cc
+++ b/impeller/renderer/backend/vulkan/command_encoder_vk.cc
@@ -37,6 +37,8 @@ class TrackedObjectsVK {
     if (!pool) {
       VALIDATION_LOG
           << "Command pool died before a command buffer could be recycled.";
+      // The buffer can not be freed if its command pool has been destroyed.
+      buffer_.release();
       return;
     }
     pool->CollectGraphicsCommandBuffer(std::move(buffer_));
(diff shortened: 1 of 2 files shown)


gaaclarke:
I was able to solve this by having the fence hold onto the Context in https://github.com/flutter/engine/pull/41748

That makes sure that the device lives long enough to kill the tracked objects.

gaaclarke:
> In this case the issue is the lifetime of the command pool, not the device - see [flutter/flutter#125519 (comment)](https://github.com/flutter/flutter/issues/125519#issuecomment-1535612572)
> 
> The command pool is destroyed by the `ClearAllPools` call made by the `ContextVK` destructor, which runs before the context's `FenceWaiterVK` is destructed. If the command pool has already been deleted, then it is no longer safe to free buffers associated with that pool.

Right, I get that.  The command pools before my change are actually cleaned up when the thread is destroyed since a thread_local variable has a strong reference.  That PR I linked actually removes the strong link when the context is destroyed.  Since the fence holds onto a strong reference to the tracked objects and the context the pool is guaranteed to be alive when the fence finishes.

jason-simmons:
`ClearAllPools` calls `Reset` on every pool associated with the `ContextVK`.  `Reset` deletes the `graphics_pool_`, which is the actual underlying Vulkan command pool object.

Once the `graphics_pool_` is gone, all of its command buffers are freed and are now unusable.  The `CommandPoolVK` should not make further Vulkan API calls using those buffers.

gaaclarke (approved):
Talked with jason offline this is a good change.  I've seen these command buffers crash when their deallocator is run late.  `release` is a misnomer and isn't releasing the ownership of the c++ wrapper.  This all looks good, thanks jason.

PR #41766, opened 2023-05-05
Optimize Vertices.raw() indices check, fix CkVertices.raw() checks, add tests - addresses some of #123085

timmaffett:
This PR is the most conservative thing we can do to address some of https://github.com/flutter/flutter/issues/123085
This reduces the number of ops needed to check the `indices[n]*2 <= positions.length` by avoiding the *2 by simply comparing `indices[n] < halfPositionsLength`.

Additionally, this PR also fixes the indices[] checking within the canvaskit web_ui version of Vertices.raw() (CkVertices.raw())
This routine was incorrectly testing `indices[n] < positions.length` instead of `indices[n]*2 < positions.length`.  It was also missing a check to verify that positions.length was even (because it holds pairs of positions for the raw()).
One note is that the checks within CkVertice() and CkVertice.raw() are using indices.any() instead of looping, and their error messages are less informative their Vertices() counterparts because of this.  They also check for indices<0 where Vertices() does not.  Should these be changed to be identical to Vertices() ?

I also added tests of the CkVertices() and CkVertices.raw() parameter checking - this was completely missing and is how the ckVertices.raw() was getting away with doing incorrect parameter checking.  (These are essentially identical to the test that are done against Vertices() within `testing\dart\painting_test.dart` with the exception of the less informative error messages mentioned above).

https://github.com/flutter/flutter/issues/123085 is further discussing placing all of the Vertices.raw() parameter checks behind an assert() .  Removing all checks for non-debug mode has additional considerations, while this change only optimizes operations with no change to the verification being performed.

I would be happy to add the additional assert() wrapper discussed in https://github.com/flutter/flutter/issues/123085 to the checks for a real additional performance gain :wink:

## Pre-launch Checklist

- [X] I signed the [CLA].

### DIFF
--- a/lib/ui/painting.dart
+++ b/lib/ui/painting.dart
@@ -4714,8 +4714,10 @@ class Vertices extends NativeFieldWrapperClass1 {
       throw ArgumentError('"positions" and "textureCoordinates" lengths must match.');
     }
     if (indices != null) {
+      final int halfPositions = positions.length>>1;  // len/2 - we already checked that positions.length is even
       for (int index = 0; index < indices.length; index += 1) {
-        if (indices[index] * 2 >= positions.length) {
+        // we need to check that `indices[index] * 2 >= positions.length`, so do it without multiplying
+        if (indices[index] >= halfPositions) {
           throw ArgumentError(
             '"indices" values must be valid indices in the positions list '
             '(i.e. numbers in the range 0..${positions.length ~/ 2 - 1}), '
--- a/lib/ui/painting.dart
+++ b/lib/ui/painting.dart
@@ -4714,7 +4714,7 @@ class Vertices extends NativeFieldWrapperClass1 {
       throw ArgumentError('"positions" and "textureCoordinates" lengths must match.');
     }
     if (indices != null) {
-      final int halfPositions = positions.length>>1;  // len/2 - we already checked that positions.length is even
+      final int halfPositions = positions.length >> 1;  // len/2 - we already checked that positions.length is even
       for (int index = 0; index < indices.length; index += 1) {
         // we need to check that `indices[index] * 2 >= positions.length`, so do it without multiplying
         if (indices[index] >= halfPositions) {
--- a/lib/web_ui/lib/src/engine/canvaskit/vertices.dart
+++ b/lib/web_ui/lib/src/engine/canvaskit/vertices.dart
@@ -47,6 +47,9 @@ class CkVertices implements ui.Vertices {
     Int32List? colors,
     Uint16List? indices,
   }) {
+    if (positions.length % 2 != 0) {
+      throw ArgumentError('"positions" must have an even number of entries (each coordinate is an x,y pair).');
+    }
     if (textureCoordinates != null &&
         textureCoordinates.length != positions.length) {
       throw ArgumentError(
--- a/lib/web_ui/lib/src/engine/canvaskit/vertices.dart
+++ b/lib/web_ui/lib/src/engine/canvaskit/vertices.dart
@@ -57,7 +57,7 @@ class CkVertices implements ui.Vertices {
     }
     // see ui.Vertices.raw() - all indices must be <(positions.length/2) because they refer to
     // pairs in the positions array.
-    final halfPositions = positions.length>>1;
+    final int halfPositions = positions.length >> 1;
     if (indices != null &&
         indices.any((int i) => i < 0 || i >= halfPositions)) {
       throw ArgumentError(
--- a/lib/web_ui/lib/src/engine/canvaskit/vertices.dart
+++ b/lib/web_ui/lib/src/engine/canvaskit/vertices.dart
@@ -55,8 +55,11 @@ class CkVertices implements ui.Vertices {
     if (colors != null && colors.length * 2 != positions.length) {
       throw ArgumentError('"positions" and "colors" lengths must match.');
     }
+    // see ui.Vertices.raw() - all indices must be <(positions.length/2) because they refer to
+    // pairs in the positions array.
+    final halfPositions = positions.length>>1;
     if (indices != null &&
-        indices.any((int i) => i < 0 || i >= positions.length)) {
+        indices.any((int i) => i < 0 || i >= halfPositions)) {
       throw ArgumentError(
           '"indices" values must be valid indices in the positions list.');
     }
--- a/lib/web_ui/test/canvaskit/vertices_test.dart
+++ b/lib/web_ui/test/canvaskit/vertices_test.dart
@@ -66,6 +66,60 @@ void testMain() {
         .draw(builder.build().layerTree);
     await matchGoldenFile('canvaskit_vertices_antialiased.png', region: region);
   }, skip: isSafari);
+
+  test('Vertices checks', () {
+    try {
+      ui.Vertices(
+        ui.VertexMode.triangles,
+        const <ui.Offset>[ui.Offset.zero, ui.Offset.zero, ui.Offset.zero],
+        indices: Uint16List.fromList(const <int>[0, 2, 5]),
+      );
+      throw 'Vertices did not throw the expected error.';
+    } on ArgumentError catch (e) {
+      expect('$e', 'Invalid argument(s): "indices" values must be valid indices in the positions list.');
+    }
+    ui.Vertices( // This one does not throw.
+      ui.VertexMode.triangles,
+      const <ui.Offset>[ui.Offset.zero],
+    ).dispose();
+    ui.Vertices( // This one should not throw.
+      ui.VertexMode.triangles,
+      const <ui.Offset>[ui.Offset.zero, ui.Offset.zero, ui.Offset.zero],
+      indices: Uint16List.fromList(const <int>[0, 2, 1, 2, 0, 1, 2, 0]), // Uint16List implements List<int> so this is ok.
+    ).dispose();
+  });
+
+  test('Vertices.raw checks', () {
+    try {
+      ui.Vertices.raw(
+        ui.VertexMode.triangles,
+        Float32List.fromList(const <double>[0.0]),
+      );
+      throw 'Vertices.raw did not throw the expected error.';
+    } on ArgumentError catch (e) {
+      expect('$e', 'Invalid argument(s): "positions" must have an even number of entries (each coordinate is an x,y pair).');
+    }
+    try {
+      ui.Vertices.raw(
+        ui.VertexMode.triangles,
+        Float32List.fromList(const <double>[0.0, 0.0, 0.0, 0.0, 0.0, 0.0]),
+        indices: Uint16List.fromList(const <int>[0, 2, 5]),
+      );
+      throw 'Vertices.raw did not throw the expected error.';
+    } on ArgumentError catch (e) {
+      expect('$e', '"indices" values must be valid indices in the positions list.');
+    }
+    ui.Vertices.raw( // This one does not throw.
+      ui.VertexMode.triangles,
+      Float32List.fromList(const <double>[0.0, 0.0]),
+    ).dispose();
+    ui.Vertices.raw( // This one should not throw.
+      ui.VertexMode.triangles,
+      Float32List.fromList(const <double>[0.0, 0.0, 0.0, 0.0, 0.0, 0.0]),
+      indices: Uint16List.fromList(const <int>[0, 2, 1, 2, 0, 1, 2, 0]),
+    ).dispose();
+  });
+
 }
 
 CkVertices _testVertices() {


timmaffett:
As it exists currently this PR is intended to optimize performance at zero change to the existing validity checks.

I agree that wrapping them in an assert like you illustrate in https://github.com/flutter/flutter/issues/123085 would be where the real performance gain would be.  It sounds like at least some of the team is on board - but it does take away the check entirely.
I checked within the engine and the indices array is assumed to be valid and never checked again.
This would be a little different then some assert() uses however, in that it even in release mode it would be conceivable that programmatically generated indices arrays could be buggy and contain invalid indices.  Therefor removing the assert does leave the underlying engine 'vulnerable' to possibly receiving invalid indices data.  In practice this probably does not matter, but is probably a decision best left to the coders of the internal engine routines to weigh in on.

skyne98:
Can we get their opinion on this somehow? I am not aware of the internal Flutter team structure, unfortunately, and for me the change sounds simple enough to follow through with just a "let's go" or two from the internal team.

lib/web_ui/lib/src/engine/canvaskit/vertices.dart
```diff
@@ -47,6 +47,9 @@ class CkVertices implements ui.Vertices {
     Int32List? colors,
     Uint16List? indices,
   }) {
+    if (positions.length % 2 != 0) {
```
kevmoo, line 50: there is an `isOdd` method on `int`, btw!

timmaffett: I changed that to `if (positions.length.isOdd) {`

The `%2 != 0` check matched the pre-existing checks within `painting.dart`. 
I did not go and change those pre-existing checks in `painting.dart`, they are still using `if (positions.length % 2 != 0) {`

Using `isOdd` is definitely more readable!

Do you have any idea what the difference is in the code that the dart compiler creates for `%2!=0` vs `isOdd` ?
I tried quickly take a look at https://godbolt.org/ but it did not have luck.

kevmoo: I'd hope it's all inlined. I have a vague memory that the code generated for JavaScript is better, though...

mdebbar:
The code itself looks good to me. I'm not familiar enough though to tell if this is a change we want to make or not.

cc @goderbauer @chinmaygarde @yjbanov @hterkelsen

timmaffett:
I have modified this PR to move the argument checking to asserts and add the assert argument checking to all of the web renderers.  (Previously the web renderers never checked the arguments.)

In the discussion of https://github.com/flutter/flutter/issues/123085 both @chinmaygarde and @goderbauer seemed agree that this would be a worthwhile change.  The performance of Vertices.raw() will improve for non web renderers and be identical for the web renderers. (and all web renderers now get argument checking parity with non web platforms).

mdebbar (commented):
@chinmaygarde @goderbauer would you mind reviewing this PR?

lib/ui/painting.dart
```diff
@@ -4744,24 +4744,9 @@ base class Vertices extends NativeFieldWrapperClass1 {
...
-            '"indices" values must be valid indices in the positions list '
-            '(i.e. numbers in the range 0..${positions.length - 1}), '
-            'but indices[$index] is ${indices[index]}, which is too big.',
-          );
-        }
-      }
-    }
+  }) : assert(textureCoordinates == null || textureCoordinates.length == positions.length,'"positions" and "textureCoordinates" lengths must match.'),
+    assert(colors == null || colors.length == positions.length,'"positions" and "colors" lengths must match.'),
+    assert(indices==null || indices.every((int i) => i >= 0 && i < positions.length),'"indices" values must be valid indices in the positions list.') {
```
goderbauer, line 4749: nit: add spaces around the `==`

goderbauer: (Here and everywhere else in this PR)

lib/ui/painting.dart
```diff
@@ -4828,27 +4813,10 @@ base class Vertices extends NativeFieldWrapperClass1 {
...
-            '(i.e. numbers in the range 0..${positions.length ~/ 2 - 1}), '
-            'but indices[$index] is ${indices[index]}, which is too big.',
-          );
-        }
-      }
-    }
+  }) : assert(positions.length.isEven,'"positions" must have an even number of entries (each coordinate is an x,y pair).'),
+    assert(textureCoordinates == null || textureCoordinates.length == positions.length,'"positions" and "textureCoordinates" lengths must match.'),
+    assert(colors == null || colors.length * 2 == positions.length,'"colors" length must be half the length of "positions".'),
+    assert(indices==null || indices.every((int i) => i >= 0 && i*2 < positions.length),'"indices" values must be valid indices in the positions list.') {
```
goderbauer, line 4819: nit: add spaces around the `*`

goderbauer: (here and elsewhere in the PR)

lib/web_ui/test/ui/vertices_test.dart
```diff
@@ -58,6 +65,120 @@ void testMain() {
     await drawPictureUsingCurrentRenderer(recorder.endRecording());
     await matchGoldenFile('ui_vertices_antialiased.png', region: region);
   }, skip: isHtml); // https://github.com/flutter/flutter/issues/127454
+
+  test('Vertices assert checks', () {
+    // We don't test textureCoordinate assert checks on html render because HTML renderer's SurfaceVertices() does not support textureCoordinates
+    if(renderer.rendererTag != 'html') {
```
goderbauer, line 71: nit: add space after `if`

goderbauer: also elsewhere in this PR

lib/web_ui/test/ui/vertices_test.dart
```diff
@@ -17,6 +18,12 @@ void main() {
 }
 
 void testMain() {
+  bool assertsEnabled = false;
```
goderbauer, line 21: I am not super familiar with how tests run in the engine, but in the framework they always run with asserts enabled. Is there evidence that that's not true here? Otherwise, we should simplify this remove this logic. If not, is it at least guaranteed that these tests in some environments run with asserts enabled so these tests actually test what we want them to test?

goderbauer (commented):
Overall this LGTM

lib/web_ui/test/ui/vertices_test.dart
timmaffett: In most cases asserts are enabled and this logic was not required, but there were at least one set of tests that were failing (without this logic) because asserts were not enabled (and it appeared to be compiling in release mode?) 
  I copied this assert enabled testing logic from other testing of asserts within the engine.

Hixie:
Is the engine resilient to bad data here, or will bad data cause a crash?

cc @chinmaygarde for a final decision on landing this. We should either decide to go for it or close the PR.

dnfield (changes requested):
Please update your tests so that if assertions are not enabled, you actually try to use the invalid objects.

I suspect flutter_tester will segfault in this case. If so, we have more to fix. If not we have coverage to show that it's safe to do this.

lib/web_ui/test/ui/vertices_test.dart
```diff
@@ -58,6 +65,120 @@ void testMain() {
...
+    if (renderer.rendererTag != 'html') {
+      try {
+        ui.Vertices(
+          ui.VertexMode.triangles,
+          const <ui.Offset>[ui.Offset.zero, ui.Offset.zero, ui.Offset.zero],
+          textureCoordinates: const <ui.Offset>[ui.Offset.zero],
+        );
+        if (assertsEnabled) {
+          throw AssertionError('Vertices did not throw the expected assert error.');
+        }
```
dnfield, line 80: for example, in an else block here call `Canvas.drawVertices(vertices, etc.)`

timmaffett: @dnfield I have added a function which uses and disposes of the created invalid `ui.Vertices` objects when assertions are not enabled.  Each of the various created (when assertions are disabled) invalid Vertices object's use will be tested this way.  In this way we will answer your question to determine "segfault vs. safe to use".

(additionally, within `testing\dart\painting_test.dart`,  I added calls to dispose() of the invalid `Vertices` objects that are created if assertions are not enabled (but here I did not attempt to use them).

```dart

// This function is used when asserts are NOT enabled and vertices() allows invalid 
// vertices objects to be created.  Here we attempt to use these invalid ui.Vertices
// objects to determine if they cause problems within drawVertices()
void useAndDisposeOfInvalidVerticesObject( ui.Vertices vertices ) {
  expect(vertices.debugDisposed, isFalse);

  final ui.PictureRecorder recorder = ui.PictureRecorder();
  final ui.Canvas canvas = ui.Canvas(
    recorder,
    const ui.Rect.fromLTRB(0, 0, 100, 100)
  );
  canvas.drawVertices(
    vertices,
    ui.BlendMode.srcOver,
    ui.Paint(),
  );
  vertices.dispose();
  expect(vertices.debugDisposed, isTrue);
}
```

Used like this throughout `vertices_test.dart` for all created/invalid `ui.Vertices` objects when assertions are not enabled:
```dart
try {
        final ui.Vertices invalidVertices = ui.Vertices(
          ui.VertexMode.triangles,
          const <ui.Offset>[ui.Offset.zero, ui.Offset.zero, ui.Offset.zero],
          textureCoordinates: const <ui.Offset>[ui.Offset.zero],
        );
        if (assertsEnabled) {
          throw AssertionError('Vertices did not throw the expected assert error.');
        } else {
          // Assertions are NOT ENABLED, so we will attempt to use this invalid ui.Vertices object
          // we just created to check if that causes exceptions/problems.
          useAndDisposeOfInvalidVerticesObject(invalidVertices);
        }
      } on AssertionError catch (e) {
        expect('$e', contains(r'\"positions\" and \"textureCoordinates\" lengths must match.'));
      }
```

kevmoo:
@timmaffett – we have a bit of distance to go to get this landed...I hope we do...but either way, this effort is AMAZING. Super thoughtful and thorough. "WE" (the Flutter ecosystem) are so lucky to have folks like you!

timmaffett:
> @timmaffett – we have a bit of distance to go to get this landed...I hope we do...but either way, this effort is AMAZING. Super thoughtful and thorough. "WE" (the Flutter ecosystem) are so lucky to have folks like you!

Thank you.
I thought @dnfield 's fields idea to actually attempt to use the invalid vertices objects (which would be created when asserts were not enabled) would be a interesting way to possibly answer @Hixie 's question concerning about what would happen if invalid vertices objects are actually used.

testing/dart/painting_test.dart
```diff
@@ -35,20 +67,24 @@ void main() {
       Vertices.raw(
         VertexMode.triangles,
         Float32List.fromList(const <double>[0.0]),
-      );
-      throw 'Vertices.raw did not throw the expected error.';
-    } on ArgumentError catch (e) {
-      expect('$e', 'Invalid argument(s): "positions" must have an even number of entries (each coordinate is an x,y pair).');
+      ).dispose();
```
dnfield, line 70: Can you also add something in here that tries to actually use these objects on a Canvas when assertions are not enabled?

We want to make sure that these invalid objects won't cause a crash.

zanderso:
From PR triage: The comments about tests from @dnfield are still relevant and need to be addressed before this PR can land. Adding myself as a reviewer to route to an engine team member when the tests are updated.

jtmcdole:
Still looking at one comment in the test file by Dan.

jonahwilliams (commented):
We shouldn't ship this. Replacing all of the bounds checks with asserts means mismatched positions/colors  could cause segfaults or other UB in profile/release builds.

lib/ui/painting.dart
```diff
@@ -4737,24 +4737,9 @@ base class Vertices extends NativeFieldWrapperClass1 {
...
-            '"indices" values must be valid indices in the positions list '
-            '(i.e. numbers in the range 0..${positions.length - 1}), '
-            'but indices[$index] is ${indices[index]}, which is too big.',
-          );
-        }
-      }
-    }
+  }) : assert(textureCoordinates == null || textureCoordinates.length == positions.length,'"positions" and "textureCoordinates" lengths must match.'),
+    assert(colors == null || colors.length == positions.length,'"positions" and "colors" lengths must match.'),
+    assert(indices == null || indices.every((int i) => i >= 0 && i < positions.length),'"indices" values must be valid indices in the positions list.') {
```
jonahwilliams, line 4742: We could definitely let the indices check be an assert. I think at most you'll get graphical corruption.

jonahwilliams: actually, on iOS/Metal we have to unroll triangle fans. If we end up using the indices to look up positions an invalid index could still lead to a crash.

timmaffett: @jonahwilliams Do you think I should go ahead and just close this ? - or would moving the checks to the c++ code closer where they are used make more sense ?  (and possibly be faster?)

jonahwilliams:
That depends on if it segfaults on real devices if the sizes or indices are mismatched.

jonahwilliams:
If the colors or texture coordinates is shorter than positions, then we'll read past the bounds of input data. This is a potential UB. We need to validate array lengths. 

https://github.com/flutter/engine/blob/main/lib/ui/painting/vertices.cc#L58-L64

If the lenght of colors or texture coordinates is too long then we just ignore the extras.

I'm not able to trigger any crashes in the index code though, I think we might be safe to make that assert only.

jonahwilliams:
We removed the expensive runtime checks in https://github.com/flutter/engine/pull/53558 . The bounds checks should remain as they are not expensive (compared to copying the data into the native heap) and are required for memory safety.

PR #41780, opened 2023-05-05
[Impeller] Use separate atlases and shaders for color and alpha

jonahwilliams:
Creates a separate atlas context and cache for color and alpha bitmap glyphs. Removes SDF shader and uses separate shader for full color glyphs.

Requires https://github.com/flutter/engine/pull/41754

Fixes https://github.com/flutter/flutter/issues/116818
Fixes https://github.com/flutter/flutter/issues/126101

This also fixes https://github.com/flutter/engine/pull/39383 but for light text on a dark background. This problem crops up when we switch to a full color atlas. In this context, the chosen glyph color is important. But with the alpha channel only atlas, its irrelevant.

See diff:

image

Example app:

```dart
// Copyright 2014 The Flutter Authors. All rights reserved.
// found in the LICENSE file.

import 'package:flutter/widgets.dart';

void main() =>
  runApp(
    Container(
      alignment: Alignment.center,
     child: Text(
      '''
 (Unicode Conference)، الذي سيعقد في 10-12 آذار 1997 بمدينة مَايِنْتْس، ألمانيا. و سيجمع المؤتمر بين خبراء من كافة قطاعات الصناعة على الشبكة العالمية انترنيت ويونيكود، حيث ستتم، على الصعيدين الدولي والمحلي على حد سواء مناقشة سبل استخدام يونكود في النظم القائمة وفيما يخص التطبيقات الحاسوبية، الخطوط، تصميم النصوص والحوسبة متعددة اللغات.

مَمِمّمَّمِّ😀 😃 😄 😁 😆''',
textDirection: TextDirection.rtl, style: TextStyle(fontSize: 24, color: Colors.white),
     ),
    ),
  );

```


jonahwilliams:
Still need to investigate if this fixes https://github.com/flutter/flutter/issues/126101

dnfield (approved):

bdero:
Skia seems to correct the alpha based on the paint color. I think the AA alpha dropoff becomes more aggressive as the color gets darker.

jonahwilliams:
Note that I did not actually change how glyphs are rendered, except in the case where there are emoji present - then previously we would promote to a full color atlas. Upon checking, it does appear that dark text gets darker, but the rendering is indentical in the case where there is no emoji.

So not exactly regressed, but not quite fixed yet. Let me double check the color that is being using in Skia. Skia only uses a single color in their atlas regardless of light/dark background so we should be able to also.

bdero:
Skia's internal atlas is probably not going to have the color-based alpha correction applied to it, lest they would have to create entirely new atlases for different colors. My guess would be that the alpha is linear in their internal atlas. Then, when they sample it, they just map it through curve (like en exponential that's tuned based on some computed property of the color -- like brightness).

jonahwilliams:
I don't think they're doing that. Here is the text fragment shader that gets generated:

```
#include <metal_stdlib>
#include <simd/simd.h>
using namespace metal;

struct sampler2D {
    texture2d<half> tex;
    sampler smp;
};
half4 sample(sampler2D i, float2 p, float b=-0.475) { return i.tex.sample(i.smp, p, bias(b)); }
half4 sample(sampler2D i, float3 p, float b=-0.475) { return i.tex.sample(i.smp, p.xy / p.z, bias(b)); }
half4 sampleLod(sampler2D i, float2 p, float lod) { return i.tex.sample(i.smp, p, level(lod)); }
half4 sampleLod(sampler2D i, float3 p, float lod) {
    return i.tex.sample(i.smp, p.xy / p.z, level(lod));
}
half4 sampleGrad(sampler2D i, float2 p, float2 dPdx, float2 dPdy) {
    return i.tex.sample(i.smp, p, gradient2d(dPdx, dPdy));
}

struct Inputs {
    float2 vTextureCoords_S0  [[user(locn0)]];
    float vTexIndex_S0  [[user(locn1)]];
    half4 vinColor_S0  [[user(locn2)]];
};
struct Outputs {
    half4 sk_FragColor [[color(0)]];
};
struct uniformBuffer {
    float4 sk_RTAdjust;
    float2 uAtlasSizeInv_S0;
};
struct Globals {
    sampler2D uTextureSampler_0_S0;
    constant uniformBuffer* _anonInterface0;
};
fragment Outputs fragmentMain(Inputs _in [[stage_in]], texture2d<half> uTextureSampler_0_S0_Tex [[texture(0)]], sampler uTextureSampler_0_S0_Smplr [[sampler(0)]], constant uniformBuffer& _anonInterface0 [[buffer(0)]], bool _frontFacing [[front_facing]], float4 _fragCoord [[position]]) {
    Globals _globals{{uTextureSampler_0_S0_Tex, uTextureSampler_0_S0_Smplr}, &_anonInterface0};
    (void)_globals;
    Outputs _out;
    (void)_out;
    half4 outputColor_S0 = _in.vinColor_S0;
    half4 texColor = sample(_globals.uTextureSampler_0_S0, _in.vTextureCoords_S0).xxxx;
    half4 outputCoverage_S0 = texColor;
    {
        _out.sk_FragColor = outputColor_S0 * outputCoverage_S0;
    }
    return _out;
}

```

They have an R8 glyph atlas with at scale rendered glyphs, and the sampler is a NN sampler

impeller/typographer/backends/skia/text_render_context_skia.cc
```diff
@@ -184,10 +183,8 @@ static void DrawGlyph(SkCanvas* canvas,
   sk_font.setHinting(SkFontHinting::kSlight);
   sk_font.setEmbolden(metrics.embolden);
 
-  auto glyph_color = has_color ? SK_ColorWHITE : SK_ColorBLACK;
-
   SkPaint glyph_paint;
-  glyph_paint.setColor(glyph_color);
+  glyph_paint.setColor(SK_ColorWHITE);
```
jonahwilliams, line 187: This needs to always be SK_ColorWhite, otherwise we end up darkening the dark text on light backgrounds. However this change does fix the issue that non-color glyphs were inconsistently rendered depending on whether or not we had promoted the atlas to full color.

bdero:
> I don't think they're doing that. 

In this case, it's probably getting baked in when the atlas is generated? I assume different subpixel AA variations of the same glyph are also getting baked into the atlas with this solution.

> Anyway, it turns out this does not fix the light text on dark background - and we're still bolder than Skia on the same font.

You mean text looks thinner than Skia for light text with a dark background, right? Black text on a white background should look nearly identical to Skia.

impeller/typographer/backends/skia/text_render_context_skia.cc
bdero: IIUC, this is going to make white text look correct, but black text to look too thick, regressing the cases solved for: https://github.com/flutter/flutter/issues/119805 and https://github.com/flutter/flutter/issues/119234.

jonahwilliams:
I can't tell if the goldens are better or worse 😆 . I'll take a closer look tomorrow.

jonahwilliams:
Sorry, I've confused everyone and myself. This does not fix the issue with light text on a dark background, but it does fix dark text on a light background.

On main, we render dark text differently depending on whether or not the single glyph atlas has been promoted to a full color atlas. Here are some screenshots.

## TOT / Current Main Branch Behavior



If you throw this in something like diff checker, you can see that the text rendered with the emoji - that is, the text rendering in RGBA atlas is significantly thicker than the text that gets rendered in the A8 atlas.

[cast.webm](https://user-images.githubusercontent.com/8975114/236983596-38844012-4ec6-4fbc-a96f-b77dce75d089.webm)

You can also see this with light text on a dark background, the text that is rendered into the full color atlas is thicker.

[cast (1).webm](https://user-images.githubusercontent.com/8975114/236983927-d574541c-6357-4698-80e1-ae4455463638.webm)

jonahwilliams:
So now I don't think it matters if we use SK_ColorWHITE or SK_ColorBLACK, becuase regular glyphs will always be drawn in the A8 atlas - meaning they shouldn't be as thick.

jonahwilliams:
(but of course I need to figure out if the goldens got thinner (expected) or thicker (unexpected))

bdero:
My worry is about the behavior of text drawn to the U8 atlas before and after this PR. I haven't tried running this PR yet, but my expectation is that all text drawn to the U8 atlas will become thicker, regressing https://github.com/flutter/flutter/issues/119805 and https://github.com/flutter/flutter/issues/119234.

jonahwilliams:
Put the color atlas check back, just in case we need it. I've also confirmed that with the color changed back we haven't adjusted how font was rendered _provided that we were already using the A8 atlas_ . Fonts should still be thinner in general if they were included in the full color atlas before. None of our goldens cover this specific case though.

[cast (2).webm](https://github.com/flutter/engine/assets/8975114/20925a68-f6a7-462f-a62c-801415a0d416)


[cast (3).webm](https://github.com/flutter/engine/assets/8975114/3a03d78f-cedd-43df-9afe-afa48e2f0a4c)

bdero (approved):
LGTM!

PR #41781, opened 2023-05-05
[Impeller][WIP] increase initial size of alpha bitmap atlas texture.

jonahwilliams:
Fixes https://github.com/flutter/flutter/issues/126103

Requires https://github.com/flutter/engine/pull/41780

Unlike Skia which starts with 2048x2048, I decreased the size of 1024x1024. Now that we'll never switch this to a color bitmap, the memory impact is dramatically lessened.


impeller/entity/contents/content_context.h
```diff
@@ -823,7 +823,8 @@ class ContentContext {
 
   bool is_valid_ = false;
   std::shared_ptr<Tessellator> tessellator_;
-  std::shared_ptr<GlyphAtlasContext> glyph_atlas_context_;
+  std::shared_ptr<GlyphAtlasContext> alpha_glyph_atlas_context_;
+  std::shared_ptr<GlyphAtlasContext> color_glyph_atlas_context_;
```
jonahwilliams, line 827: I could put these into a map or something else. Feels weird to make a second container but might as well. I don't think we'll end up with a third glyph atlas, at least not for a while

jonahwilliams:
Closing in favor of https://github.com/flutter/engine/pull/41880

PR #41784, opened 2023-05-05
[web] New platform view API to get view by ID

mdebbar:
In order to make `HtmlElementView` more useful, we need to expose an API for getting the created platform view element. ~~This would allow `HtmlElementView` to set attributes and styles on the element.~~

This would allow us to add a new callback to the `HtmlElementView` that would return the result of calling the registered platformViewFactory to the user, after it's been created and added to the DOM, so the user can do with it as they please.

Required by https://github.com/flutter/flutter/pull/126176


ditman:
(Modified the description a little bit, since we're going to use this API to implement the new `onPlatformViewCreated(Object contents)` or `onPlatformViewInjected(Object contents)` or whatever is called :P)

lib/web_ui/lib/src/engine/platform_views/content_manager.dart
```diff
@@ -60,6 +60,14 @@ class PlatformViewManager {
     return _contents.containsKey(viewId);
   }
 
+  /// Returns the HTML element created by a registered factory for [viewId].
+  ///
+  /// Throws an [AssertionError] if [viewId] hasn't been rendered before.
+  DomElement getViewById(int viewId) {
+    assert(knowsViewId(viewId), 'No view has been rendered for viewId: $viewId');
+    return _contents[viewId]!.firstElementChild!;
```
ditman, line 68: Please, document that we return the `firstElementChild` because we cache the created view wrapped with the `<flt-slot-content>` tag (right?)

lib/web_ui/lib/src/engine/platform_views/content_manager.dart
```diff
@@ -68,8 +76,13 @@ class PlatformViewManager {
   /// `factoryFunction` needs to be a [PlatformViewFactory].
   bool registerFactory(String viewType, Function factoryFunction,
       {bool isVisible = true}) {
-    assert(factoryFunction is PlatformViewFactory ||
-        factoryFunction is ParameterizedPlatformViewFactory);
+    assert(
```
ditman, line 79: I think you should revert these changes so this PR doesn't conflict with https://github.com/flutter/engine/pull/42255

ditman (approved):
LGTM, let's go!

lib/web_ui/lib/src/engine/platform_views/content_manager.dart
mdebbar: `_contents[viewId]` is the wrapper (i.e. the slot element). The first child of the slot is the element created by the factory.

mdebbar: https://github.com/flutter/engine/blob/f3f6a020595dd7f9f809bca2ebb604a42aae1cfc/lib/web_ui/lib/src/engine/platform_views/content_manager.dart#L133

mdebbar: Oh I think this is exactly what you are saying. Sorry I misread your comment.

Sounds good, I'll document why I'm returning `firstElementChild`.

lib/web_ui/lib/src/engine/platform_views/content_manager.dart
mdebbar: Good catch! I'll just move this change to the other PR.

PR #41803, opened 2023-05-07
[Impeller] Compute geometry for drawPoints in vertex shader, with fallback for optimized CPU routine.

jonahwilliams:
Example of using disabling rasterization and using a vertex shader for computing circle geometry. This is only wired up in Metal, but could be supported in GLES 3.0 / WebGL with  [Transform_Feedback](https://www.khronos.org/opengl/wiki/Transform_Feedback)

<google-sheets-html-origin><style type="text/css"></style>

IMpeller Before TEss Fix |   | 500ms
-- | -- | --
Impeller TOT |   | 52ms
Impeller CPU |   | 14ms
Impeller Vert |   | 2.4ms

Benchmark is pending here: https://github.com/flutter/flutter/pull/126728


impeller/entity/geometry.cc
```diff
@@ -883,4 +892,135 @@ std::optional<Rect> RRectGeometry::GetCoverage(const Matrix& transform) const {
...
+  if (scaled_radius < 4.0) {
+    return 8;
+  }
+  if (scaled_radius < 16) {
+    return 16;
+  }
+  if (scaled_radius < 32) {
+    return 32;
+  }
+  return 64;
```
jonahwilliams, line 1012: TODO

impeller/renderer/pipeline_builder.h
```diff
@@ -16,6 +16,15 @@
 
 namespace impeller {
 
+class NonRenderingFragment {
+ public:
+  static constexpr std::string_view kLabel = "NonRendering";
+  static constexpr std::string_view kEntrypointName = "non_rendering_main";
+  static constexpr ShaderStage kShaderStage = ShaderStage::kFragment;
+  static constexpr std::string_view kGeneratorName = "NonRendering";
```
jonahwilliams, line 24: TODO: right now the debug info comes from the fragment name but for non-rasterizing pipelines it should come from the vertex name.

impeller/renderer/pipeline_builder.h
```diff
@@ -52,35 +61,44 @@ struct PipelineBuilder {
...
-                       << "'.";
-        return false;
+
+      std::shared_ptr<const ShaderFunction> fragment_function;
+      if (!std::is_same<FragmentShader, NonRenderingFragment>::value) {
+        fragment_function = context.GetShaderLibrary()->GetFunction(
+            FragmentShader::kEntrypointName, ShaderStage::kFragment);
+      }
+
+      if (!std::is_same<FragmentShader, NonRenderingFragment>::value) {
```
jonahwilliams, line 87: TODO: make sure to validate vertex shader still

impeller/entity/geometry.h
```diff
@@ -290,4 +294,38 @@ class RRectGeometry : public Geometry {
   FML_DISALLOW_COPY_AND_ASSIGN(RRectGeometry);
 };
 
+class PointFieldGeometry : public Geometry {
```
jonahwilliams, line 297: TODO: figure out where to put the fallback to CPU computation. Could just add a fallback here?

impeller/entity/geometry.cc
dnfield: See `SkNextPow2` and `SkNextLog2`.

impeller/entity/geometry.h
jonahwilliams: Done

jonahwilliams:
This is working and has the correct fallbacks, last remaining task is to fix the subdivision curve.

impeller/entity/geometry.cc
dnfield: (Oh, I see, this is supposed to be modelling a curve not just always getting the next power of 2)

jonahwilliams:
Closing in favor of compute based approach

PR #41822, opened 2023-05-08
Get rid of "outrageous" default text styles for HTML renderer.

eyebrowsoffire:
The CanvasKit renderer doesn't do this, so we should just remove the red text global style.


yjbanov (approved):
lgtm

mdebbar (changes requested):
Thanks for removing this!

You probably want to remove this one too: https://github.com/flutter/engine/blob/edf0f90de1d2da51747c9185e11b7c49a680c152/lib/web_ui/lib/src/engine/text/canvas_paragraph.dart#L494

mdebbar (approved):
LGTM

PR #41824, opened 2023-05-08
Migrate mac host clang tidy to engine v2.

godofredoc:
This removes the additional gn command from the ci/lint tool and uses a single gn command.

## Pre-launch Checklist

- [X] I signed the [CLA].


ci/builders/standalone/mac_host_clang_tidy.json
```diff
@@ -0,0 +1,36 @@
...
+    "tests": [
+        {
+            "language": "python3",
+            "name": "test: lint host_debug",
+            "parameters": [
+                "--variant",
+                "host_debug",
+                "--lint-all",
+                "--shard-id=1",
+                "--shard-variants=host_debug"
```
zanderso, line 31: Sorry for remembering this a bit late, but if you look in the engine v1 recipe here:

https://flutter.googlesource.com/recipes/+/refs/heads/main/recipes/engine/engine_lint.py#101

Notice that the `shardVariant` is specified as `ios_debug_sim`. That's because the engine's linter script is looking at the compile commands for other variants, and only running a shard of the lint commands in the intersection, to avoid duplicating the work. That's why it was doing two `gn` commands.

As written, this will only lint half of the files. In order to duplicate the behavior of engine v1, we'd need to be able to run multiple `gn` commands. Another option would be to allow duplication between `host_debug` and `ios_debug_sim`, but to shard them with themselves. We could do that by running `list.sh` in parallel on two machines, one with `--shard-id=0` and the other with `--shard-id=1`.

Not sure which is easier.

godofredoc: I'll give it another try merging mac host and ios clang tidy into a single build to avoid duplication and move the scripts to run as global tests. I could add the double gn logic but that may add more complexity to all the other configurations to cover this single use case.

godofredoc: This is ready for review it is now making a 1:1 translation of the existing builds https://luci-milo.appspot.com/raw/build/logs.chromium.org/flutter/led/godofredoc_google.com/d3c09280c1196e2320dae4cee33f5ab5ea2044d27ef22554af37c71bfba6377c/+/build.proto

PR #41826, opened 2023-05-08
[tests] Remove unused fuchsia.sys protocol reference

drwez:
Remove reference to unused fuchsia.sys.Environment protocol in test Realm setup. Besides being unused,
the reference was not correctly matched by an #include, nor a GN build dependency.

Bug: fxbug.dev/81285


drwez:
@arbreng PTAL - this should be the final remaining change (indirectly) required to allow the legacy `CreateView()` to be removed from Fuchsia.

arbreng (approved):

PR #34574, opened 2022-07-10
[Windows] Fix GDI resource leaks in the software fallback path

sevenstars:
In [FlutterWindowWin32::OnBitmapSurfaceUpdated()](https://github.com/flutter/engine/blob/6c321db6885fdbc33a79ced39571341c06ab2383/shell/platform/windows/flutter_window_win32.cc#L247), the ReleaseDC function is not called to release the DC.
The ReleaseDC function must be called to release the DC.

This PR will fix issue: https://github.com/flutter/flutter/issues/107368

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [Flutter Style Guide] and the [C++, Objective-C, Java style guides].
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I added new tests to check the change I am making or feature I am adding, or Hixie said the PR is test-exempt. See [testing the engine] for instructions on
writing and running engine tests.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I signed the [CLA].
- [x] All existing and new tests are passing.

PR #41827, opened 2023-05-08
Fix leak when switching channels

robert-ancell:
As noticed in https://github.com/flutter/flutter/issues/126209

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/shell/platform/linux/fl_method_channel.cc
+++ b/shell/platform/linux/fl_method_channel.cc
@@ -88,7 +88,7 @@ static void channel_closed_cb(gpointer user_data) {
 static void fl_method_channel_dispose(GObject* object) {
   FlMethodChannel* self = FL_METHOD_CHANNEL(object);
 
-  // Note we don't have the clear the handler in messenger as it holds
+  // Note we don't have to clear the handler in messenger as it holds
   // a reference to this object so the following code is only run after
   // the messenger has closed the channel already.
 
--- a/shell/platform/linux/fl_method_channel.cc
+++ b/shell/platform/linux/fl_method_channel.cc
@@ -75,10 +75,6 @@ static void channel_closed_cb(gpointer user_data) {
   g_autoptr(FlMethodChannel) self = FL_METHOD_CHANNEL(user_data);
 
   self->channel_closed = TRUE;
-  // Clear the messenger so that disposing the channel will not clear the
-  // messenger's mapped channel, since `channel_closed_cb` means the messenger
-  // has abandoned this channel.
-  self->messenger = nullptr;
 
   // Disconnect handler.
   if (self->method_call_handler_destroy_notify != nullptr) {
@@ -92,10 +88,9 @@ static void channel_closed_cb(gpointer user_data) {
 static void fl_method_channel_dispose(GObject* object) {
   FlMethodChannel* self = FL_METHOD_CHANNEL(object);
 
-  if (self->messenger != nullptr) {
-    fl_binary_messenger_set_message_handler_on_channel(
-        self->messenger, self->name, nullptr, nullptr, nullptr);
-  }
+  // Note we don't have the clear the handler in messenger as it holds
+  // a reference to this object so the following code is only run after
+  // the messenger has closed the channel already.
 
   g_clear_object(&self->messenger);
   g_clear_pointer(&self->name, g_free);


shell/platform/linux/fl_method_channel.cc
```diff
@@ -92,7 +88,7 @@ static void channel_closed_cb(gpointer user_data) {
 static void fl_method_channel_dispose(GObject* object) {
   FlMethodChannel* self = FL_METHOD_CHANNEL(object);
 
-  if (self->messenger != nullptr) {
+  if (!self->channel_closed) {
     fl_binary_messenger_set_message_handler_on_channel(
```
stuartmorgan, line 92: I was thinking about this more yesterday and isn't this actually dead code that we could just remove? Setting the original handler provides an owning ref asking with the closed callback, so isn't it impossible to reach `dispose` without the handler already having been removed?

robert-ancell: Hmm, I think you've exposed another issue here - the channel should be dereferenced in `channel_closed_cb` as otherwise that reference is never dropped and the channel will never be disposed.

We do still need `channel_closed` though, as we can't otherwise tell if a given channel is the current handler for a binary messenger channel.

stuartmorgan: > Hmm, I think you've exposed another issue here - the channel should be dereferenced in `channel_closed_cb` as otherwise that reference is never dropped and the channel will never be disposed.

Isn't the owning reference passed to the messenger that's tracking the handlers to free? I assumed it was being freed at that lower layer when the handler was replaced.

If that's not the case, then how would the bug that led to this PR be happening? We'd only be setting a null handler if the object is being disposed, right?

> We do still need `channel_closed` though, as we can't otherwise tell if a given channel is the current handler for a binary messenger channel.

I just meant this line right here, where we clear the handler. I don't see any way to get into this method with `channel_closed` being false. (I'm actually not 100% clear on why we need the other use of this variable either now that I look at it, but that's not really in scope here.)

robert-ancell: The channel is being passed as an opaque piece of data, so the messenger doesn't free/unref it at all. It calls the destroy notify function which should do that. I don't think this PR is fixing any active bug, it was just noticed in the other bug that the messenger reference was being set to nullptr, which would leak if that code was executed.

If all the reference counting was working correctly, there would be two cases:
1. An active channel is being disposed because the Flutter app no longer requires it. In the dispose it would need to clear the message handler on the binary messenger so that would not longer contact a disposed object. (channel_closed=false).
2. An inactive channel is being disposed because the Flutter app no longer requires it. This channel is inactive because a second channel was created using the same name after this one. In the dispose it should not clear the message handler on the binary messenger as this would clear the handler of the second channel. (channel_closed=true).

stuartmorgan: > I don't think this PR is fixing any active bug, it was just noticed in the other bug that the messenger reference was being set to nullptr, which would leak if that code was executed.

Right, sorry; I got my wires crossed on this PR vs the original issue.

> 1. An active channel is being disposed because the Flutter app no longer requires it. In the dispose it would need to clear the message handler on the binary messenger so that would not longer contact a disposed object. (channel_closed=false).

I still don't understand how this case could happen though. Isn't the intent that [this ref of self](https://github.com/flutter/engine/blob/18cd6bb49d6000c39e808c3aaa3dd2bb3ea9ea29/shell/platform/linux/fl_method_channel.cc#L132) is lifetime-bound to the callback, and only removed in `channel_closed_cb`?

In which case in order to be disposed, the app/plugin would have to release any refs it has on the channel object, **and** the handler would have to be cleared or replaced so that `channel_closed_cb` would be called to remove that ref. Or to put it another way: if the ref I linked above is supposed to be balanced [here](https://github.com/flutter/engine/blob/18cd6bb49d6000c39e808c3aaa3dd2bb3ea9ea29/shell/platform/linux/fl_method_channel.cc#L74), then the last ref to `self` cannot be removed without [this line](https://github.com/flutter/engine/blob/18cd6bb49d6000c39e808c3aaa3dd2bb3ea9ea29/shell/platform/linux/fl_method_channel.cc#L77) also running.

Gustl22: Is there a test strategy for leaks like this? Haven't found a similar case for linux yet. Only found this: https://github.com/flutter/engine/pull/34574/files or `fml::MakeRefCounted` [here](https://github.com/flutter/engine/blob/main/shell/common/engine_unittests.cc)

Gustl22: @stuartmorgan @robert-ancell can you give me a hint where and how to add these leak tests? Best regards :)

robert-ancell: There's currently not any tests for the Linux code that could pick up leaks. What we probably should be doing is running some tests under valgrind and checking that reports no errors.

Gustl22: Okay thanks, I think these changes are out of my competences 😅

loic-sharma:
@robert-ancell @stuartmorgan Is this still on your radar? (Desktop triage)

stuartmorgan:
I was waiting on an answer to my last question about if/how part of this code could even be run.

shell/platform/linux/fl_method_channel.cc
robert-ancell: You're right @stuartmorgan, clearing the handler is never required due to the references. I'm a little concerned that if other code was to change this would no longer be true so I've tried to avoid this by adding a comment and noting this in the commit message.

stuartmorgan (approved):
LGTM

shell/platform/linux/fl_method_channel.cc
```diff
@@ -92,10 +88,9 @@ static void channel_closed_cb(gpointer user_data) {
 static void fl_method_channel_dispose(GObject* object) {
   FlMethodChannel* self = FL_METHOD_CHANNEL(object);
 
-  if (self->messenger != nullptr) {
-    fl_binary_messenger_set_message_handler_on_channel(
-        self->messenger, self->name, nullptr, nullptr, nullptr);
-  }
+  // Note we don't have the clear the handler in messenger as it holds
```
stuartmorgan, line 91: Typo: we don't have **to** clear the handler

PR #41829, opened 2023-05-08
[web] Cleanup assertionsEnabled

mdebbar:
For consistency with the rest of the flutter code, this PR explores the possibility of removing our proprietary `assertionsEnabled`.


lib/web_ui/lib/src/engine/canvaskit/native_memory.dart
```diff
@@ -130,9 +129,11 @@ class CountedRef<R extends StackTraceDebugger, T extends Object> {
...
   CountedRef(T nativeObject, R debugReferrer, String debugLabel) {
     _ref = UniqueRef<T>(this, nativeObject, debugLabel);
-    if (assertionsEnabled) {
+    assert(() {
       debugReferrers.add(debugReferrer);
-    }
+      return true;
+    }());
+    // TODO: @yjbanov what's the purpose of this? It's always true.
     assert(refCount == debugReferrers.length);
```
mdebbar, line 137: @yjbanov, a question for you :)

During construction, `refCount` is initialized to `1` and `debugReferrers` is initialized to an empty set then one item is added.

PR #41831, opened 2023-05-08
[web] generalize focusability in semantics

yjbanov:
Introduce 2 new classes for a11y focus management:

* `AccessibilityFocusManager`: a generic class that attaches "focus" and "blur" event handlers, and forwards the events to the framework as `SemanticsAction.didGainAccessibilityFocus` and `SemanticsAction.didLoseAccessibilityFocus` respectively. Provides the `changeFocus` method for the framework to move a11y focus to the target element.
* `Focusable`: a role manager that provides generic focus management functionality to `SemanticsObject`s that don't need anything special.

Rewrites focus management using the above two classes as follows:

* All focusable nodes except text fields and incrementables get the `Focusable` role (all custom focus stuff in `Tappable` was removed and delegated to `Focusable`).
* `Incrementable` uses a custom `<input>` internally and so it cannot use the `Focusable` role. Instead, it uses `AccessibilityFocusManager` to manage the focus on the `<input>` element.

Behavioral changes:

* Fixes https://github.com/flutter/flutter/issues/118737, but more generally fixes all nodes that use the `isFocusable` and `hasFocus` bits.
* `Tappable` only partially implemented focusability (e.g. it didn't generate the respective `SemanticsAction` events). Now by delegating to `Focusable`, it will inherit all the functionality.
* `Incrementable` and `Checkable` (checkboxes, radios, switches) get focus management features for the first time.
* Elements that are not inherently focusable (text, images) can now be focused if semantics requires them to be.
* `TextField` is left alone for now as focus and on-screen keyboard interact with each other in non-obvious ways.

### DIFF
--- a/lib/web_ui/lib/src/engine/semantics/focusable.dart
+++ b/lib/web_ui/lib/src/engine/semantics/focusable.dart
@@ -1,0 +1,193 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:ui/ui.dart' as ui;
+
+import '../dom.dart';
+import '../platform_dispatcher.dart';
+import '../util.dart';
+import 'semantics.dart';
+
+/// Supplies generic accessibility focus features to semantics nodes that have
+/// [ui.SemanticsFlag.isFocusable] set.
+///
+/// Assumes that the element being focused on is [SemanticsObject.element]. Role
+/// managers with special needs can implement custom focus management and
+/// exclude this role manager.
+///
+/// `"tab-index=0"` is used because `<flt-semantics>` is not intrinsically
+/// focusable. Examples of intrinsically focusable elements include:
+///
+///   * <button>
+///   * <input> (of any type)
+///   * <a>
+///   * <textarea>
+///
+/// See also:
+///
+///   * https://developer.mozilla.org/en-US/docs/Web/Accessibility/Keyboard-navigable_JavaScript_widgets
+class Focusable extends RoleManager {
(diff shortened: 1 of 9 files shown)


lib/web_ui/lib/src/engine/semantics/focusable.dart
```diff
@@ -0,0 +1,193 @@
...
+  DomEventListener domBlurListener,
+});
+
+/// Implements accessibility focus management for arbitrary elements.
+///
+/// Unlike [Focusable], which implements focus features on [SemanticsObject]s
+/// whose [SemanticsObject.element] is directly focusable, this class can help
+/// implementing focus features on custom elements. For example, [Incrementable]
+/// uses a custom `<input>` tag internally while its root-level element is not
+/// focusable. However, it can still use this class to managing the focus of the
```
mdebbar, line 71:
```suggestion
/// focusable. However, it can still use this class to manage the focus of the
```

lib/web_ui/lib/src/engine/semantics/focusable.dart
```diff
@@ -0,0 +1,193 @@
...
+        return true;
+      }());
+      return;
+    }
+
+    // Delay the focus request until the final DOM structure is established
+    // because the element may not yet be attached to the DOM, or it may be
+    // reparented and lose focus again.
+    _owner.addOneTimePostUpdateCallback(() {
+      if (_target != target) {
```
mdebbar, line 178: Good call!

mdebbar (approved):

lib/web_ui/lib/src/engine/semantics/focusable.dart
yjbanov: Done.

PR #41834, opened 2023-05-08
[Impeller] track path convexity and use it to simplify fill tessellation.

jonahwilliams:
We can do this on the GPU, but we'll need a fallback when we don't have polyline compute.

From experimentation, Skia has already computed this value when Impeller gets that path. Additionally, this will allow us to speed up tessellation of known convex shapes like ovals that we don't want to add special cases for in the API.

### DIFF
--- a/impeller/entity/entity_unittests.cc
+++ b/impeller/entity/entity_unittests.cc
@@ -2605,19 +2605,18 @@ TEST_P(EntityTest, TessellateConvex) {
     // Sanity check simple rectangle.
     auto [pts, indices] =
         TessellateConvex(PathBuilder{}
-                            .AddRect(Rect::MakeLTRB(0, 0, 10, 10))
-                            .TakePath()
-                            .CreatePolyline(1.0));
+                             .AddRect(Rect::MakeLTRB(0, 0, 10, 10))
+                             .TakePath()
+                             .CreatePolyline(1.0));
 
     std::vector<Point> expected = {
-      {0, 0},   {10, 0},  {4, 4}, //
-      {10, 0},  {10, 10}, {4, 4}, //
-      {10, 10}, {0, 10},  {4, 4}, //
-      {0, 10},  {0, 0},   {4, 4}, //
+        {0, 0},   {10, 0},  {4, 4},  //
+        {10, 0},  {10, 10}, {4, 4},  //
+        {10, 10}, {0, 10},  {4, 4},  //
+        {0, 10},  {0, 0},   {4, 4},  //
     };
-    std::vector<uint16_t> expected_indices = {
-      0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11
-    };
+    std::vector<uint16_t> expected_indices = {0, 1, 2, 3, 4,  5,
+                                              6, 7, 8, 9, 10, 11};
     ASSERT_EQ(pts, expected);
     ASSERT_EQ(indices, expected_indices);
   }
--- a/impeller/entity/entity_unittests.cc
+++ b/impeller/entity/entity_unittests.cc
@@ -2600,5 +2600,28 @@ TEST_P(EntityTest, TiledTextureContentsIsOpaque) {
   ASSERT_FALSE(contents.IsOpaque());
 }
 
+TEST_P(EntityTest, TessellateConvex) {
+  {
+    // Sanity check simple rectangle.
+    auto [pts, indices] =
+        TessellateConvex(PathBuilder{}
+                            .AddRect(Rect::MakeLTRB(0, 0, 10, 10))
+                            .TakePath()
+                            .CreatePolyline(1.0));
+
+    std::vector<Point> expected = {
+      {0, 0},   {10, 0},  {4, 4}, //
+      {10, 0},  {10, 10}, {4, 4}, //
+      {10, 10}, {0, 10},  {4, 4}, //
+      {0, 10},  {0, 0},   {4, 4}, //
+    };
+    std::vector<uint16_t> expected_indices = {
+      0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11
+    };
+    ASSERT_EQ(pts, expected);
+    ASSERT_EQ(indices, expected_indices);
+  }
+}
+
 }  // namespace testing
 }  // namespace impeller
--- a/impeller/entity/entity_unittests.cc
+++ b/impeller/entity/entity_unittests.cc
@@ -2610,13 +2610,27 @@ TEST_P(EntityTest, TessellateConvex) {
                              .CreatePolyline(1.0));
 
     std::vector<Point> expected = {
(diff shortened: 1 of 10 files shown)


jonahwilliams:
Note that once we add the GPU case, we could probably simplify things a bit and delete the specialized RRect geometry.

impeller/entity/geometry.cc
```diff
@@ -16,6 +16,46 @@
 
 namespace impeller {
 
+/// Given a convex polyline, create a triangle fan structure.
+std::pair<std::vector<Point>, std::vector<uint16_t>> TessellateConvex(
+    Path::Polyline polyline) {
+  std::vector<Point> output;
+  output.reserve(polyline.points.size() * 3);
+
+  for (auto j = 0u; j < polyline.contours.size(); j++) {
```
dnfield, line 25: I think this needs an assertion that there's only a single contour. Otherwise, please add some tests that show it works with multiple contours.

impeller/entity/geometry.cc
```diff
@@ -16,6 +16,46 @@
...
+    center_y /= end - start;
+
+    for (auto i = start + 1; i < end; i++) {
+      const auto& point_a = polyline.points[i - 1];
+      const auto& point_b = polyline.points[i];
+      output.emplace_back(point_a);
+      output.emplace_back(point_b);
+      output.emplace_back(Point(center_x, center_y));
+    }
+    // Some shapes don't seem to self close?
```
dnfield, line 44: If you fill a path that is missing an explicit close, you need to add one.

We should just document this behavior in the header doc.

impeller/entity/geometry.cc
```diff
@@ -16,6 +16,46 @@
...
+    }
+    center_x /= end - start;
+    center_y /= end - start;
+
+    for (auto i = start + 1; i < end; i++) {
+      const auto& point_a = polyline.points[i - 1];
+      const auto& point_b = polyline.points[i];
+      output.emplace_back(point_a);
+      output.emplace_back(point_b);
+      output.emplace_back(Point(center_x, center_y));
```
dnfield, line 42: Why not define this as a fan, and use `kTriangeFan` below?

impeller/entity/geometry.cc
```diff
@@ -16,6 +16,46 @@
...
+      output.emplace_back(polyline.points[end - 1]);
+      output.emplace_back(polyline.points[start]);
+      output.emplace_back(Point(center_x, center_y));
+    }
+  }
+
+  std::vector<uint16_t> index(output.size());
+  for (auto i = 0u; i < output.size(); i++) {
+    index[i] = i;
+  }
```
dnfield, line 55: We should drop this and just use `kTriangleFan`. We need to teach the rest of the pipeline to not require an index buffer in that case.

impeller/geometry/path.h
```diff
@@ -34,6 +34,13 @@ enum class FillType {
   kAbsGeqTwo,
 };
 
+enum class Convexity {
+  kUnknown,
+  kConvex,
+  kConcave,
+  kComplex,
```
dnfield, line 41: Document these - I'm not sure what you mean by complex (holes? multiple non-overlapping polygons with different convexity? self-intersection?)

impeller/geometry/path.h
```diff
@@ -94,6 +101,10 @@ class Path {
 
   FillType GetFillType() const;
 
+  void SetConvexity(Convexity value);
```
dnfield, line 104: I'd feel a little bit better about this if it lived on `PathBuilder` and only `PathBuilder` set it on `Path`. Right now the API is easy to misuse. I'd like to ideally get to a world where `impeller::Path` is immutable.

dnfield (changes requested):

dnfield (changes requested):
Mostly looks good

dnfield:
Although the golden change looks a little sad - 
image

impeller/entity/geometry.cc
jonahwilliams: I think that when we convert to the polyline we should add this close if it isn't present. Anyway, this is handled below. i'll update the comment

impeller/entity/geometry.cc
jonahwilliams: Triangle fan is not universally supported, and isn't included in the set of PrimitiveTypes

impeller/geometry/path.h
jonahwilliams: complex polygons have self intersection. I think for now since I'm not using these other cases, I'll drop it down to just unknown and convex

impeller/geometry/path.h
jonahwilliams: Sure, can do that

jonahwilliams:
I'll remove the center point and see if that improves the golden. Technically we could just use one of the vertices as the "center" point, maybe that gives a better result.

impeller/entity/geometry.cc
dnfield: (Talked offline, I guess Metal removed this, so ok - instead just copy the vertices to the vertex buffer and make the index buffer form them into a strip).

impeller/entity/geometry.cc
dnfield: If we do it in polyline conversion, we need to make sure to _not_ add a close if it's for a stroke (or if a close is already present).

impeller/entity/entity_unittests.cc
```diff
@@ -2610,13 +2610,27 @@ TEST_P(EntityTest, TessellateConvex) {
                              .CreatePolyline(1.0));
 
     std::vector<Point> expected = {
-        {0, 0},   {10, 0},  {4, 4},  //
-        {10, 0},  {10, 10}, {4, 4},  //
-        {10, 10}, {0, 10},  {4, 4},  //
-        {0, 10},  {0, 0},   {4, 4},  //
+        {0, 0}, {10, 0}, {10, 10}, {0, 10},  //
```
dnfield, line 2613: 🎉

impeller/geometry/path.h
```diff
@@ -101,9 +99,7 @@ class Path {
 
   FillType GetFillType() const;
 
-  void SetConvexity(Convexity value);
-
-  Convexity GetConvexity() const;
+  bool GetIsConvex() const;
```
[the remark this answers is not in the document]
jonahwilliams: Done

dnfield:
Looks like the golden change is still there but it's not quite as bad as I initially saw - there's some shift in how the line across the circle is stepped, but it's stepped a little unevenly in both cases.

dnfield (approved):

dnfield:
Actually I'm confused. The original diff looked like it had no large steps. The second diff I looked at both sides had steps. Now when I click the latest one I don't see any changes

jonahwilliams:
sorry I just clicked approve...

PR #41836, opened 2023-05-08
Makes android embedding to send full uri

chunhtai:
related https://github.com/flutter/flutter/issues/100624

## Pre-launch Checklist

- [ ] I signed the [CLA].

### DIFF
--- a/shell/platform/android/io/flutter/embedding/android/FlutterActivityAndFragmentDelegate.java
+++ b/shell/platform/android/io/flutter/embedding/android/FlutterActivityAndFragmentDelegate.java
@@ -525,16 +525,7 @@ import java.util.List;
     if (host.shouldHandleDeeplinking()) {
       Uri data = intent.getData();
       if (data != null) {
-        String fullRoute = data.getPath();
-        if (fullRoute != null && !fullRoute.isEmpty()) {
-          if (data.getQuery() != null && !data.getQuery().isEmpty()) {
-            fullRoute += "?" + data.getQuery();
-          }
-          if (data.getFragment() != null && !data.getFragment().isEmpty()) {
-            fullRoute += "#" + data.getFragment();
-          }
-          return fullRoute;
-        }
+        return data.toString();
       }
     }
     return null;
--- a/shell/platform/android/io/flutter/view/FlutterView.java
+++ b/shell/platform/android/io/flutter/view/FlutterView.java
@@ -367,10 +367,6 @@ public class FlutterView extends SurfaceView
     navigationChannel.setInitialRoute(route);
   }
 
-  public void pushRoute(String route) {
-    navigationChannel.pushRoute(route);
-  }
-
   public void popRoute() {
     navigationChannel.popRoute();
   }
--- a/shell/platform/android/io/flutter/view/FlutterView.java
+++ b/shell/platform/android/io/flutter/view/FlutterView.java
@@ -367,6 +367,10 @@ public class FlutterView extends SurfaceView
     navigationChannel.setInitialRoute(route);
   }
 
+  public void pushRoute(String route) {
+    navigationChannel.pushRoute(route);
+  }
+
   public void popRoute() {
     navigationChannel.popRoute();
   }
--- a/shell/platform/android/test/io/flutter/embedding/android/FlutterActivityAndFragmentDelegateTest.java
+++ b/shell/platform/android/test/io/flutter/embedding/android/FlutterActivityAndFragmentDelegateTest.java
@@ -309,7 +309,7 @@ public class FlutterActivityAndFragmentDelegateTest {
     ArgumentCaptor<FlutterEngineGroup.Options> optionsCaptor =
         ArgumentCaptor.forClass(FlutterEngineGroup.Options.class);
     verify(flutterEngineGroup, times(1)).createAndRunEngine(optionsCaptor.capture());
-    assertEquals("/initial_route", optionsCaptor.getValue().getInitialRoute());
+    assertEquals("foo://example.com/initial_route", optionsCaptor.getValue().getInitialRoute());
   }
 
   @Test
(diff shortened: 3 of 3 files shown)


shell/platform/android/test/io/flutter/embedding/android/FlutterActivityAndFragmentDelegateTest.java
```diff
@@ -838,7 +839,7 @@ public void itDoesNotSendPushRouteInformationMessageWhenOnNewIntentIsNonHierarch
     delegate.onNewIntent(mockIntent);
 
     // Verify that the navigation channel was not given a push route message.
-    verify(mockFlutterEngine.getNavigationChannel(), times(0))
+    verify(mockFlutterEngine.getNavigationChannel(), times(1))
```
chunhtai, line 842: This test was added https://github.com/flutter/engine/pull/29766

Which seems like a bug to me, it could have send the non hierarchical uri.

shell/platform/android/io/flutter/embedding/android/FlutterActivityAndFragmentDelegate.java
```diff
@@ -525,16 +525,7 @@ private String maybeGetInitialRouteFromIntent(Intent intent) {
...
-        if (fullRoute != null && !fullRoute.isEmpty()) {
-          if (data.getQuery() != null && !data.getQuery().isEmpty()) {
-            fullRoute += "?" + data.getQuery();
-          }
-          if (data.getFragment() != null && !data.getFragment().isEmpty()) {
-            fullRoute += "#" + data.getFragment();
-          }
-          return fullRoute;
-        }
+        return data.toString();
```
reidbaker, line 528: In general you should not rely on the behavior of tostring in production code. https://www.bekk.christmas/post/2019/4/never-use-tostring()-for-behaviour

shell/platform/android/io/flutter/view/FlutterView.java
```diff
@@ -360,10 +360,6 @@ public void setInitialRoute(String route) {
     navigationChannel.setInitialRoute(route);
   }
 
-  public void pushRoute(String route) {
```
reidbaker, line 363: Is this a breaking change?

reidbaker (commented):
Not sure this pr should land. I am not marking it request changes so that i am not blocking the pr. Deferring to @dnfield and @camsim99 for now.

shell/platform/android/io/flutter/view/FlutterView.java
chunhtai: I am not sure who is using this API, and if there is a way to access this API at all. I can restore it if this is needed though

shell/platform/android/io/flutter/embedding/android/FlutterActivityAndFragmentDelegate.java
chunhtai: I can't find a better way to convert uri to a serializable representation. all the suggestions i found is to use to string

shell/platform/android/io/flutter/view/FlutterView.java
dnfield: This is the old embedder v1 API. I think there were ongoing efforts to remove it but the person(s) working on that are no longer on the team.

shell/platform/android/test/io/flutter/embedding/android/FlutterActivityAndFragmentDelegateTest.java
dnfield: The linked patch was trying to avoid a crash. This assertion is probably unnecessary.

However, we should update the comment and have some kind of assertion about what arguments actually got sent through.

shell/platform/android/io/flutter/embedding/android/FlutterActivityAndFragmentDelegate.java
dnfield: I think in this case it's probably ok, since `Uri` provides both a `toString` and a `parse` method that knows how to deserialize from the string produced by toString...

shell/platform/android/test/io/flutter/embedding/android/FlutterActivityAndFragmentDelegateTest.java
```diff
@@ -308,7 +308,7 @@ public void flutterEngineGroupGetsInitialRouteFromIntent() {
     ArgumentCaptor<FlutterEngineGroup.Options> optionsCaptor =
         ArgumentCaptor.forClass(FlutterEngineGroup.Options.class);
     verify(flutterEngineGroup, times(1)).createAndRunEngine(optionsCaptor.capture());
-    assertEquals("/initial_route", optionsCaptor.getValue().getInitialRoute());
+    assertEquals("foo://example.com/initial_route", optionsCaptor.getValue().getInitialRoute());
```
dnfield, line 311: When these strings to go the framework, are they handled as dart `Uri` objects? If so this is probably fine, but if not won't this cause breakages in framework apps?

Even if so, won't this cause breakages to people using `toString` on the Uri?

chunhtai: yes the framework is treated it as uri object after https://github.com/flutter/flutter/pull/119968

chunhtai: > Even if so, won't this cause breakages to people using toString on the Uri?

I am not sure if i understand, can you add more details?

shell/platform/android/io/flutter/view/FlutterView.java
dnfield: If it's possible, you should put this back, Otherwise LGTM

dnfield (approved):
LGTM modulo avoiding the change to the v1 embedding.

PR #41857, opened 2023-05-09
[Impeller] started cleaning up pools when devices are deleted

gaaclarke:
fixes https://github.com/flutter/flutter/issues/125571

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/impeller/renderer/backend/vulkan/command_encoder_vk.cc
+++ b/impeller/renderer/backend/vulkan/command_encoder_vk.cc
@@ -157,10 +157,12 @@ bool CommandEncoderVK::Submit() {
     return false;
   }
 
-  return fence_waiter_->AddFence(
-      std::move(fence), [tracked_objects = std::move(tracked_objects_)] {
-        // Nothing to do, we just drop the tracked objects on the floor.
-      });
+  return fence_waiter_->AddFence(std::move(fence),
+                                 [tracked_objects = std::move(tracked_objects_),
+                                  strong_device = std::move(strong_device)] {
+                                   // Nothing to do, we just drop the tracked
+                                   // objects on the floor.
+                                 });
 }
 
 vk::CommandBuffer CommandEncoderVK::GetCommandBuffer() const {
--- a/impeller/renderer/backend/vulkan/command_encoder_vk.cc
+++ b/impeller/renderer/backend/vulkan/command_encoder_vk.cc
@@ -35,8 +35,6 @@ class TrackedObjectsVK {
     }
     auto pool = pool_.lock();
     if (!pool) {
-      VALIDATION_LOG
-          << "Command pool died before a command buffer could be recycled.";
       // The buffer can not be freed if its command pool has been destroyed.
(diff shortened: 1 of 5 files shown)


impeller/renderer/backend/vulkan/command_encoder_vk.cc
```diff
@@ -157,10 +157,12 @@ bool CommandEncoderVK::Submit() {
     return false;
   }
 
-  return fence_waiter_->AddFence(
-      std::move(fence), [tracked_objects = std::move(tracked_objects_)] {
-        // Nothing to do, we just drop the tracked objects on the floor.
-      });
+  return fence_waiter_->AddFence(std::move(fence),
+                                 [tracked_objects = std::move(tracked_objects_),
+                                  strong_device = std::move(strong_device)] {
```
jason-simmons, line 162: Why is this capturing a strong reference to the `ContextVK` in the lambda?

`TrackedObjectsVK` holds a weak reference to the `CommandPoolVK`.  That should allow cleanup of the tracked objects if the `CommandPoolVK` and the `ContextVK`/`vk::Device` are still alive (and do nothing if they are already gone)

impeller/renderer/backend/vulkan/command_pool_vk.cc
```diff
@@ -52,6 +48,9 @@ std::shared_ptr<CommandPoolVK> CommandPoolVK::GetThreadLocal(
 }
 
 void CommandPoolVK::ClearAllPools(const ContextVK* context) {
+  if (tls_command_pool.get()) {
+    tls_command_pool.get()->erase(context->GetHash());
```
jason-simmons, line 52: Move this to the end of the function so it runs after the block that calls `Reset` on every `CommandPoolVK` associated with this context.

In theory this should be unnecessary because nobody else should be holding on to a strong reference to this thread's `CommandPoolVK`, and the `erase` should destruct it and do all the necessary cleanup.  But for safety I think if would be prudent to ensure that the current thread's `CommandPoolVK` for this context is explicitly reset at the same time as the ones for other threads.

jason-simmons: Actually I just realized this is unnecessary.  If this thread's `CommandPoolVK` is still alive after the `tls_command_pool.erase`, then the corresponding `weak_ptr` in `g_all_pools` will still be promotable to a `shared_ptr` and will get the `Reset` call.

So the current implementation in the PR will work as intended.

impeller/renderer/backend/vulkan/command_encoder_vk.cc
gaaclarke: Our options are to keep a strong hold on the device in the fence, or to release the buffer without crashing.  I believe you prefer removing the strong link and calling `release()`.  I think I do too so I did that, although I'm unhappy about the C++ vulkan wrapper and how it looks like a leak.

jason-simmons (approved):

PR #41870, opened 2023-05-09
[web] Fix event offset for transformed widgets (and text input nodes).

htoor3:
Text inputs have moved outside of the shadowDOM and are now using the pointer event offset calculation algorithm that platform views use.  However, transforms (e.g. scaling) applied to the input element aren't currently accounted for, which leads to incorrect offsets and clicks being registered inaccurately.

This PR attempts to transform those offset coordinates using the transform matrix data that is included in the geometry information sent over to `text_editing.dart` from the framework.

## Issues

* Fixes https://github.com/flutter/flutter/issues/125948 (text editing)
* Fixes https://github.com/flutter/flutter/issues/126661 (platform view scaling)
* Fixes https://github.com/flutter/flutter/issues/126754

## Pre-launch Checklist

- [x] I signed the [CLA].


lib/web_ui/lib/src/engine/pointer_binding/event_position_helper.dart
```diff
@@ -25,6 +29,12 @@ ui.Offset computeEventOffsetToTarget(DomMouseEvent event, DomElement actualTarge
   }
 
   final bool isTargetOutsideOfShadowDOM = event.target != actualTarget;
+  final bool isInput = event.target.toString() == 'input';
+
+  if(isInput) {
+    return _computeOffsetForInputs(event);
+  }
```
mdebbar, line 36: We should also check that `geometry` is available.
```suggestion
  if (isInput) {
    final EditableTextGeomtery inputGeometry = textEditing.strategy.geometry;
    if (inputGeometry != null) {
      return _computeOffsetForInputs(event, inputGeometry);
    }
  }
```

lib/web_ui/lib/src/engine/pointer_binding/event_position_helper.dart
```diff
@@ -25,6 +29,12 @@ ui.Offset computeEventOffsetToTarget(DomMouseEvent event, DomElement actualTarge
   }
 
   final bool isTargetOutsideOfShadowDOM = event.target != actualTarget;
+  final bool isInput = event.target.toString() == 'input';
```
mdebbar, line 32: This won't work for `<textarea>`. Instead let's check if `event.target` is a descendant of the text editing host.

```suggestion
  final bool isInput =
      flutterViewEmbedder.textEditingHostNode.contains(event.target);
```

lib/web_ui/lib/src/engine/pointer_binding/event_position_helper.dart
```diff
@@ -33,6 +43,24 @@ ui.Offset computeEventOffsetToTarget(DomMouseEvent event, DomElement actualTarge
   return ui.Offset(event.offsetX, event.offsetY);
 }
 
+/// Computes the offsets for input nodes, which live outside of the shadowDOM.
+/// Since inputs can be transformed (scaled, translated, etc), we can't rely on
+/// `event.offset` to be accurate. `_computeOffsetRelativeToActualTarget` only
```
mdebbar, line 48: > Since inputs can be transformed (scaled, translated, etc), we can't rely on `event.offset` to be accurate.

You are rightfully stating the reason the previous implementation didn't work. But that's not the reason we can't rely on `event.offset`. The reason we can't use `event.offset` is because it is relative to the input itself and not to the flutter-view.

lib/web_ui/lib/src/engine/pointer_binding/event_position_helper.dart
```diff
@@ -33,6 +43,24 @@ ui.Offset computeEventOffsetToTarget(DomMouseEvent event, DomElement actualTarge
...
+/// handles the case where inputs are translated, but will have issues for scaled
+/// inputs (see: https://github.com/flutter/flutter/issues/125948).
+///
+/// We compute the offsets here by using the text input geometry data that is
+/// sent from the framework, which includes information on how to transform the
+/// underlying input element. We transform the `event.offset` points we receive
+/// using the values from the input's transform matrix.
+ui.Offset _computeOffsetForInputs(DomMouseEvent event) {
```
mdebbar, line 56: One more check we can do here is whether the event target is the same as the active input element i.e. `textEditing.strategy.domElement`.

lib/web_ui/lib/src/engine/pointer_binding/event_position_helper.dart
```diff
@@ -33,6 +43,24 @@ ui.Offset computeEventOffsetToTarget(DomMouseEvent event, DomElement actualTarge
...
+  final Float32List matrix = textEditing.strategy.geometry!.globalTransform;
+  final FastMatrix32 fastMatrix = FastMatrix32(matrix);
+  fastMatrix.transform(event.offsetX, event.offsetY);
```
mdebbar, line 59: @yjbanov is this the right API to use for transforming an offset?

mdebbar (commented):
Thanks for putting a fix together!

lib/web_ui/lib/src/engine/pointer_binding/event_position_helper.dart
htoor3: That's a good call.  Should we `assert` this or just put it in the outside conditional so the code has a chance to fall through to the fallback case?

lib/web_ui/lib/src/engine/pointer_binding/event_position_helper.dart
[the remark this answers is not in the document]
yjbanov: If this is performance-critical, I'd first check if the matrix is within the limits that `FastMatrix32` can deal with and only use it then, and fallback to full matrix otherwise. I think`transformKindOf` should give you the right info.

yjbanov: I think we want to support the general case. See my comment above for more details.

lib/web_ui/lib/src/engine/pointer_binding/event_position_helper.dart
mdebbar: If the event target is different from the active input, we are going to produce the wrong offsets regardless of which method we use. So I don't care much about which method we use in this case (`_computeOffsetForInputs` or fallback). What I do care about is being able to notice when this unexpected scenario occurs, so IMHO, we should assert.

lib/web_ui/lib/src/engine/pointer_binding/event_position_helper.dart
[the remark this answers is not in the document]
ditman: Why not https://developer.mozilla.org/en-US/docs/Web/API/DOMMatrixReadOnly / DOMMatrix?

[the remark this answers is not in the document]
htoor3: @ditman `DOMMatrixReadOnly` and its instance method `transformPoint` could certainly work, although it's missing documentation on MDN.  But the API seems well supported across different browsers.  We can also init it with `globalTransform` which should be a 16 length array. Do you think there would be significant advantages to using the browser API? Maybe their `transformPoint` is more optimized?

@yjbanov I don't see any docs outlining the limitations of the `FastMatrix32`, do you mean it only support matrices that perform simple 2D transforms and `Matrix4` should be used for 3D transforms?

lib/web_ui/lib/src/engine/pointer_binding/event_position_helper.dart
htoor3: Updated

lib/web_ui/lib/src/engine/pointer_binding/event_position_helper.dart
htoor3: This is updated btw

lib/web_ui/lib/src/engine/pointer_binding/event_position_helper.dart
```diff
@@ -33,6 +47,34 @@ ui.Offset computeEventOffsetToTarget(DomMouseEvent event, DomElement actualTarge
...
+/// sent from the framework, which includes information on how to transform the
+/// underlying input element. We transform the `event.offset` points we receive
+/// using the values from the input's transform matrix.
+ui.Offset _computeOffsetForInputs(DomMouseEvent event, EditableTextGeometry inputGeometry) {
+  final DomElement targetElement = event.target! as DomHTMLElement;
+  final DomHTMLElement domElement = textEditing.strategy.activeDomElement;
+  assert(targetElement == domElement, 'The targeted input element must be the active input element');
+  final Float32List transformValues = inputGeometry.globalTransform;
+  assert(transformValues.length == 16);
+  final Matrix4 transform = Matrix4(
```
htoor3, line 66: I replaced `FastMatrix32` with `Matrix4` @yjbanov 

cc @ditman

yjbanov: Since the transform is applied once per DOM event no extra optimization is necessary. The default transform should be fast enough.

yjbanov (approved):
LGTM if LGT @ditman

lib/web_ui/lib/src/engine/pointer_binding/event_position_helper.dart
```diff
@@ -33,6 +47,34 @@ ui.Offset computeEventOffsetToTarget(DomMouseEvent event, DomElement actualTarge
...
+    transformValues[0], transformValues[1], transformValues[2], transformValues[3],
+    transformValues[4], transformValues[5], transformValues[6], transformValues[7],
+    transformValues[8], transformValues[9], transformValues[10], transformValues[11],
+    transformValues[12], transformValues[13], transformValues[14], transformValues[15],
+  );
```
mdebbar, line 71: Maybe this is simpler?
```suggestion
  final Matrix4 transform = Matrix4.fromFloat32List(transformValues);
```

mdebbar (approved):

lib/web_ui/lib/src/engine/pointer_binding/event_position_helper.dart
htoor3: Definitely.  Good catch

ditman (approved):
Can you give this a quick test with this with this app:

* https://github.com/flutter/samples/tree/main/web_embedding/element_embedding_demo

And see if the input is still working as expected when the glassPane is being transformed too? (for example when the app is rotated a few degrees!)

Other than that, this is similar to what we discussed. That `globalTransform` matrix looks HANDY!  🙏

htoor3:
> Can you give this a quick test with this with this app:
> 
> * https://github.com/flutter/samples/tree/main/web_embedding/element_embedding_demo
> 
> And see if the input is still working as expected when the glassPane is being transformed too? (for example when the app is rotated a few degrees!)
> 
> Other than that, this is similar to what we discussed. That `globalTransform` matrix looks HANDY! 🙏

Just tested with your demo and added an input and it works well. I actually noticed an issue on the demo (without these changes) where if the embedded app is rotated at a steep angle, clicking anywhere in a focused input blurs it.  This also seems to be fixed with these changes!

lib/web_ui/lib/src/engine/pointer_binding/event_position_helper.dart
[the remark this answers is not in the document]
ditman: @yjbanov mentioned that DOMMatrix might not be fully supported everywhere, so I guess we're sticking to our own implementation!

ditman:
I've tested the fix against a bunch of test apps (most of them linked above), and it seems to improve things. Currently writing a test. (We should cherry-pick this into 3.10 stable @yjbanov @kevmoo)

yjbanov (approved):
lgtm

mdebbar (approved):
LGTM. Thanks @htoor3 and @ditman for fixing this!

PR #41875, opened 2023-05-09
[Android] Lifecycle defaults to focused instead of unfocused

gspencergoog:
## Description

In https://github.com/flutter/engine/pull/41702, the default state of "the focus bit is "false", assuming that Android will send an `onWindowFocusChanged(true)` when the window is first focused, but there appear to be some cases where that doesn't happen.

This change puts the initial state back to what it used to be: in the absence of focus change events, entering the "onResume" Android state will report the `resumed` state in Flutter. Before this PR, if no focus events were received, it would default to `inactive`.

## Tests
 - Updated tests to match.

### DIFF
--- a/shell/platform/android/io/flutter/embedding/engine/systemchannels/LifecycleChannel.java
+++ b/shell/platform/android/io/flutter/embedding/engine/systemchannels/LifecycleChannel.java
@@ -30,7 +30,7 @@ public class LifecycleChannel {
 
   private String lastAndroidState = "";
   private String lastFlutterState = "";
-  private boolean lastFocus = false;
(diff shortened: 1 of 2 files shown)


loic-sharma (approved):
Code changes look good! Is there a test verifying we'll default to focused even if we don't receive that first `onWindowFocusChanged(true)`?

gspencergoog:
> Code changes look good! Is there a test verifying we'll default to focused even if we don't receive that first onWindowFocusChanged(true)?

The changes in the unit test should do that.  It processes the resumed notification first, before any focus events are received.

PR #41877, opened 2023-05-09
[web] Move platformViewRegistry to ui_web

mdebbar:
In this PR, I'm leaving a getter for `platformViewRegistry` in `dart:ui` for backwards compatibility. At some point, we need to properly deprecate it (e.g. print a warning asking users to import it from `dart:ui_web` instead).

Next step is to migrate all call sites in flutter/flutter, flutter/packages and flutter/devtools (anything else?).

Issue: https://github.com/flutter/flutter/issues/126831

Fixes https://github.com/flutter/flutter/issues/41563


ditman:
> anything else?

We need to follow the [flutter deprecation policy](https://docs.flutter.dev/resources/compatibility#deprecation-policy), because this used to be a public API, and we know for a fact that people have used it on their own apps to define their custom platform views.

I don't think we own anything else that uses this. Maybe G3 has some internal usages too?

lib/web_ui/lib/ui.dart
```diff
@@ -14,6 +14,7 @@ import 'dart:math' as math;
 import 'dart:typed_data';
 
 import 'src/engine.dart' as engine;
+import 'ui_web/src/ui_web.dart' as ui_web;
```
ditman, line 17: How are users of `ui_web` going to import the package from outside of `flutter/engine`?

I think we should have a bunch of libraries ready for them to import, like:

* `import package:ui_web/platform_views.dart`
* `import package:ui_web/assets.dart`
* `import package:ui_web/initialization.dart`

(Importing from `src` from outside a package is [discouraged](https://dart.dev/guides/libraries/create-library-packages#organizing-a-library-package).)

ditman (commented):
This is fantastic! Small question about imports organization (this is a new package, so we can have whatever library structure we want within it.)

mdebbar:
> We need to follow the [flutter deprecation policy](https://docs.flutter.dev/resources/compatibility#deprecation-policy), because this used to be a public API, and we know for a fact that people have used it on their own apps to define their custom platform views.

As I mentioned in the PR description, I left a getter for `platformViewRegistry` in `dart:ui` so there's no breaking change in this PR. But we need to start planning the deprecation. I think printing a warning is a good first step while we continue migrating the projects that we own. What do you think?

lib/web_ui/lib/ui.dart
mdebbar: #### 1. Outside `flutter/engine`:

They simply import `dart:ui_web` like we did here: https://github.com/flutter/flutter/blob/df789c9e76098e82f80f2c5de8b5560f360afa40/packages/flutter_web_plugins/lib/src/navigation/url_strategy.dart#L8

#### 2. Inside `flutter/engine`:

Everything is within the same package (`flutter/lib/web_ui`). The new `ui_web/` is just another subfolder, nothing special about it. Same as `src/engine.dart` right above it.

ditman: Ahhh import from a **dart:** thing,, so we can't have a nicer file separation in there, shame!

ditman (approved):
LGTM!

> But we need to start planning the deprecation. I think printing a warning is a good first step while we continue migrating the projects that we own. What do you think?

SGTM!

Please consider tackling the webOnlyAssetManager next, that would allow to migrate all of flutter/packages out of this! :P

PR #41889, opened 2023-05-10
switch from MockCanvas to DisplayListBuilder in layer unit tests

flar:
Part of an ongoing set of efforts to address https://github.com/flutter/flutter/issues/106448

The layer unittests have been using a MockCanvas class to record the painting of trees of layers and then testing for the expected output.

A while back a similar mechanism was created to compare DisplayList output and to print out a human-friendly version of the differences found, but it was only used in a few tests written at the time it was created and a few since then.

This is the first in a series of PRs that will move all the rest of the unit tests onto the new DL comparison mechanism, starting with the layer types that just do basic drawing. Some of the remaining layers will require creating new hooks in, for instance, the Texture registry, the performance overlay TextBlob generation, etc.

### DIFF
--- a/display_list/display_list_unittests.cc
+++ b/display_list/display_list_unittests.cc
@@ -92,6 +92,22 @@ class DisplayListTestBase : public BaseT {
 };
 using DisplayListTest = DisplayListTestBase<::testing::Test>;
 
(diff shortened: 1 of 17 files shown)


jonahwilliams (approved):
LGTM!

PR #41907, opened 2023-05-10
Return fonts in a deterministic order.

eyebrowsoffire:
We need to make sure to return fonts in the order we passed them in, so that unit tests don't flake if they load in a different order.


yjbanov (approved):
lgtm

hterkelsen (approved):
LGTM

Hixie:
test-exempt: fixes tests

PR #41915, opened 2023-05-10
[macOS] Multi-view thread synchronizer

dkwingsmt:
This PR adapts `FlutterThreadSynchronizer` to multi-view.

### Problem
The `FlutterThreadSynchronizer` is a macOS engine class to ensure that, if the current window is resized, nothing will not be presented until the layer tree is drawn with the up-to-date size. 

This class is not compatible with multiview. A simple way to realize it is: while the class is owned by `FlutterView`, it blocks the _global_ platform thread and rasterizer thread - there is got to be some problems. Indeed, when I was testing with multiple windows (https://github.com/flutter/engine/pull/40399), the app freezes as soon as I resize a window.

The problem is because Flutter only have one rasterizer thread. When I'm resizing window A, A's synchronizer detects that the size mismatches, so it blocks the rasterizer thread to wait for an updated content with the updated size. However, window B is to be rendered next, and B's size matches and will try to rasterize, and will be blocked due to the blocked rasterizer thread, window A's updated content will never arrive, causing a deadlock.

### Changes
This PR removes the single-view assumption of `FlutterThreadSynchronizer` by making it created by `FlutterEngine` and shared by all `FlutterView`s, and that it prevents rasterization for all views if any view has a mismatched size.

The synchronizer is assigned to the view controller in the `attachToEngine:withId` method (now renamed to `setUpWithEngine:viewId:threadSynchronizer:`.

This PR also adds unit tests for `FlutterThreadSynchronizer`, which didn't have any unit tests at all.
* To achieve this, the `beginResizeForView:` method no longer checks whether  is called on the main thread, but whether it's called on the main queue. These are equivalent for the real main queue, since the documentation promises that the main queue always executes on the main thread. However, this change allows substituting the queue with a custom one for unit tests.

## Pre-launch Checklist

- [ ] I signed the [CLA].


dkwingsmt:
Actually let me investigate if I can only halt one view's rendering instead of all views'.

knopp:
I think sharing the thread synchronizer is the way to go, but it should probably be aware of multiview. 

So the API should look maybe something like:

```objc
@interface FlutterThreadSynchronizer : NSObject

- (void)view:(ViewId)viewId beginResize:(CGSize)size notify:(nonnull dispatch_block_t)notify;
- (void)view:(ViewId)viewId performCommit:(CGSize)size notify:(nonnull dispatch_block_t)notify;
- (void)viewShutDown:(ViewId)viewId
```

The waiting for content size would become a Map<ViewId, CGSize>. 

`beginResize` would block until all pending frames for ViewId->Size are satisfied.
`performCommit` would update the size for given viewId.

The `_shuttingDown` however would probably also needed to be tracked per view.

dkwingsmt:
@knopp Yeah, that's what I did in this PR. The API is a little different, but I can switch to yours. 

I don't think shutdown should be tracked per view, because `[FlutterView shutdown]` is only set when the engine is disposed. We can actually remove `[FlutterView shutdown]`.

And after thinking for a while, I don't think it's possible to allow other views to rendering while one view is waiting for its correctly sized content. The rasterization for the view being resized has to fully complete before the next view can be resized, since after all the rasterization is not made async. However, this does not harm the experience at all, since the wait time should be minimal. All views continue their animation while the view is being resized. 

The remaining question is, can we add a unit test? I think it'll need multi-threading (multiple task queues), but I'm not familiar with it. Can you help me?

shell/platform/darwin/macos/framework/Source/FlutterThreadSynchronizer.mm
```diff
@@ -102,6 +145,14 @@ - (void)performCommit:(CGSize)size notify:(nonnull dispatch_block_t)notify {
   event.Wait();
 }
 
+- (void)registerView:(int64_t)viewId {
+  _contentSizes[viewId] = CGSizeZero;
+}
+
+- (void)deregisterView:(int64_t)viewId {
+  _contentSizes.erase(viewId);
```
knopp, line 153: Is this meant to be called from platform thread? If not then maybe it should take the lock and notify `_condBlockResize` if it is actively waiting for content.

knopp:
My bad. The API seems fine, though I'm not sure what the cleanup process is. 

As for the tests, there are still just two threads, right? (platform and raster). I'd just use a serial background dispatch queue for simulating raster thread (`dispatch_queue_create` with `DISPATCH_QUEUE_SERIAL`).

shell/platform/darwin/macos/framework/Source/FlutterThreadSynchronizer.mm
dkwingsmt: I've asserted that this method is called from the platform thread.

dkwingsmt:
@knopp I've updated this PR and it should be ready for review. Changes include:
* Unit tests for the synchronizer, including its single-view capability and multiview capability.
* It no longer checks whether `beginResizeForView:` is called on the main _thread_, but whether it's called on the main _queue_. 
  * For the real main queue, these are equivalent, since the documentation said that the main queue always executes on the main thread. However, this change allows substituting the queue with a custom one for unit tests.
* Documentation updates.
* The thread synchronizer is now assigned to `FlutterViewController` via the `setupWithEngine:viewId:threadSynchronizer` method.

dkwingsmt:
Also cc @cbracken might want to take a look as well. 

cc @loic-sharma

shell/platform/darwin/macos/framework/Source/FlutterThreadSynchronizer.mm
```diff
@@ -41,18 +91,21 @@ - (void)blockUntilFrameAvailable {
...
 
   _beginResizeWaiting = NO;
 }
 
-- (void)beginResize:(CGSize)size notify:(nonnull dispatch_block_t)notify {
+- (void)beginResizeForView:(int64_t)viewId
+                      size:(CGSize)size
+                    notify:(nonnull dispatch_block_t)notify {
+  dispatch_assert_queue(_mainQueue);
   std::unique_lock<std::mutex> lock(_mutex);
```
knopp, line 106: I haven't tested this yet, but does the assert actually pass? I don't think the event handler actually runs from main dispatch queue (meaning while `CFRunLoop` is draining main dispatch queue), so technically this should fail. But then again, main dispatch queue is somewhat special so maybe the assert also checks for main thread.

dkwingsmt: Yes, it actually passes. I've merged this patch into our multiview-prototype branch and ran it with a multiwindow MVP app (see https://github.com/goderbauer/mvp) and the windows can correctly be resized without crashing.

I've also once incorrectly called `registerView:` in the test main instead of in `dispatchMainTask` and the test crashed as expected.

To explain, macOS associates the main queue with the main thread, and all other queues with a dynamic thread in a thread pool, as [the documentation](https://developer.apple.com/documentation/dispatch/dispatch_queue) says,

> Except for the dispatch queue representing your app's main thread, the system makes no guarantees about which thread it uses to execute a task.

shell/platform/darwin/macos/framework/Source/FlutterThreadSynchronizer.h
```diff
@@ -26,11 +34,65 @@
...
+ * A view must be registered before calling begineResizeForView: or
+ * performCommitForView:. It is typically done when the view controller is
+ * created.
+ */
+- (void)registerView:(int64_t)viewId;
+
+/**
+ * Requests the synchronizer to no longer track a view.
+ *
+ * It is typically done when the view controller is created.
```
knopp, line 53: The comment seems wrong. Also deregisterView does not seem to be called anywhere. Should it be called from `[FlutterView dealloc]`?

shell/platform/darwin/macos/framework/Source/FlutterView.mm
```diff
@@ -113,7 +115,6 @@ - (BOOL)layer:(CALayer*)layer
 }
 
 - (void)shutdown {
```
knopp, line 117: This is empty now, should it be removed?

shell/platform/darwin/macos/framework/Source/FlutterViewController.mm
```diff
@@ -372,6 +372,10 @@ @implementation FlutterViewController {
   std::shared_ptr<flutter::AccessibilityBridgeMac> _bridge;
 
   FlutterViewId _id;
+
+  // FlutterViewController does not actually uses the synchronizer, but only
+  // passes it to FlutterView.
+  __weak FlutterThreadSynchronizer* _weakViewThreadSynchronizer;
```
knopp, line 378: I don't think this needs to be weak.

knopp (approved):
Some nits, otherwise LGTM!

shell/platform/darwin/macos/framework/Source/FlutterView.mm
dkwingsmt: Removed.

shell/platform/darwin/macos/framework/Source/FlutterThreadSynchronizer.h
dkwingsmt: I fixed the comment, and removed register & deregister to `FVC` which handles lifecycle better.

shell/platform/darwin/macos/framework/Source/FlutterViewController.mm
dkwingsmt: Changed to strong, especially since `FVC` now directly uses it.

PR #41945, opened 2023-05-11
Convert public API NativeFieldWrapper classes to abstract interfaces

jason-simmons:
See https://github.com/flutter/flutter/issues/123756


lib/ui/compositing.dart
```diff
@@ -219,68 +230,8 @@ class ShaderMaskEngineLayer extends _EngineLayerWrapper {
...
-        currentLayer._debugChildren ??= <_EngineLayerWrapper>[];
-        currentLayer._debugChildren!.add(newLayer);
-      }
-      _layerStack.add(newLayer);
-      return true;
-    }());
-    return true;
-  }
+abstract class SceneBuilder {
+  factory SceneBuilder() => _NativeSceneBuilder();
```
dnfield, line 234: Here and elsewhere - should this be a redirecting factory instead, i.e.

```suggestion
  factory SceneBuilder() = _NativeSceneBuilder;
```

jason-simmons: done

dnfield (approved):
LGTM once the downstream https://github.com/flutter/flutter/pull/126570 lands - there's one other issue in CI about marking one of the `Shader` subclasses as base too

dnfield (approved):
LGTM once the downstream https://github.com/flutter/flutter/pull/126570 lands - there's one other issue in CI about marking one of the `Shader` subclasses as base too

PR #41951, opened 2023-05-11
Move checkerboard unit tests onto DisplayList mechanism

flar:
Part of an ongoing set of efforts to address https://github.com/flutter/flutter/issues/106448

Move the checkerboard layer unit tests onto the DisplayList version of the paint contexts and fix some bugs in the reusability of the DisplayListBuilder that this migration uncovered.

### DIFF
--- a/display_list/benchmarking/dl_complexity_helper.h
+++ b/display_list/benchmarking/dl_complexity_helper.h
@@ -120,7 +120,7 @@ class ComplexityCalculatorHelper
 
   void setAntiAlias(bool aa) override { current_paint_.setAntiAlias(aa); }
 
-  void setStyle(DlDrawStyle style) override {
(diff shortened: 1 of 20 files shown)


flar:
The number of changes seems larger than it should for a couple of reasons:

- Builder wasn't as "reusable" as I thought it was and this was uncovered by one of the tests. I've filed https://github.com/flutter/flutter/issues/126589 which will involve a minor yak shave to officially retire the feature
- One of the method names on DlPaint and DlOpReceiver had an inconsistent name. I made it more consistent with the current way the methods are named, but these names aren't necessarily fully consistent with the style guidelines. There is already an Issue filed to globally fix DL naming conventions, though: https://github.com/flutter/flutter/issues/114089

chinmaygarde (approved):

PR #41962, opened 2023-05-11
[Impeller] Make index buffer optional

dnfield:
Makes the index buffer optional. For calls that were just filling the index buffer with e.g. `std::iota`, it was wasteful. @jonahwilliams has also found cases where the index buffer will run out of room for drawPoints with lots of points.

Fixes https://github.com/flutter/flutter/issues/126572


dnfield:
Wow those goldens are not right at all. Marking as draft again.

jonahwilliams (approved):
LGTM

jonahwilliams:
Ship iiiiittttttttt

PR #41969, opened 2023-05-12
[Impeller] Dont synthesize fake index buffer if none was provided to VBB

jonahwilliams:
Followup from https://github.com/flutter/flutter/issues/126572


impeller/renderer/vertex_buffer_builder.h
```diff
@@ -121,29 +124,23 @@ class VertexBufferBuilder {
     return buffer->AsBufferView();
   }
 
-  std::vector<IndexType> CreateIndexBuffer() const {
-    if (indices_.size() > 0) {
-      return indices_;
```
dnfield, line 126: Ahh I misread the condition here when I was reading above.

dnfield (approved):

PR #41986, opened 2023-05-12
[web] Fix crash on old versions of Chrome (Intl.Segmenter missing)

mdebbar:
`Chrome < v87` have the `v8BreakIterator` API but don't have `Intl.Segmenter`. This caused crashes for some apps (see b/281377168).

This PR enhances our feature detection to take into account this API mismatch.


eyebrowsoffire (approved):
LGTM

PR #42002, opened 2023-05-12
Compile skwasm at -Oz.

eyebrowsoffire:
We want to start from a place of small size, and optimize from there with skwasm. Since there is no baseline expectation of performance at `-O3`, let's start with the smallest possible binary.


yjbanov (approved):
lgtm

PR #42019, opened 2023-05-13
Initial support for images in Skwasm

eyebrowsoffire:
This partially implements https://github.com/flutter/flutter/issues/126341

It does not implement image codecs, because they are going to get complicated with transferring video frames to the web worker and so on. I am going to deal with image codecs in a subsequent change.


lib/web_ui/lib/src/engine/skwasm/skwasm_impl/renderer.dart
```diff
@@ -285,9 +295,85 @@ class SkwasmRenderer implements Renderer {
...
+    if (targetWidth == width && targetHeight == height) {
+      // Not scaled
+      return null;
+    }
+    if (targetWidth == null) {
+      if (targetHeight == height) {
+        // Not scaled.
+        return null;
+      }
+      targetWidth = (width * targetHeight! / height).round();
```
yjbanov, line 316: Looks like the compiler failed to figure out that `targetHeight` is not null nere. Or am I missing something? Should we file an issue?

lib/web_ui/lib/src/engine/skwasm/skwasm_impl/renderer.dart
```diff
@@ -285,9 +295,85 @@ class SkwasmRenderer implements Renderer {
...
+  }) {
+    final ui.Size? scaledSize = _scaledSize(
+      width,
+      height,
+      targetWidth,
+      targetHeight
+    );
+    if (!allowUpscaling && scaledSize != null &&
+      (scaledSize.width > width || scaledSize.height > height)) {
+        domWindow.console.warn('Cannot apply targetWidth/targetHeight when allowUpscaling is false.');
```
yjbanov, line 347: I'm not sure we need to issue a warning. This seems to be the expected behavior. Sometimes you want to shrink the image to save memory, but upscaling can be done in the GPU so you don't want to waste CPU and memory on upscaling prior to rendering.

lib/web_ui/skwasm/shaders.cpp
```diff
@@ -136,3 +137,21 @@ SKWASM_EXPORT SkShader* shader_createRuntimeEffectShader(
                    childPointers.data(), childCount, nullptr)
       .release();
 }
+
+SKWASM_EXPORT SkShader* shader_createFromImage(SkImage* image,
+                                               SkTileMode tileModeX,
+                                               SkTileMode tileModeY,
+                                               FilterQuality quality,
+                                               SkScalar* matrix33) {
+  if (matrix33) {
```
yjbanov, line 146: Do we need to check? We seem to always pass a non-null matrix.

lib/web_ui/lib/src/engine/skwasm/skwasm_impl/shaders.dart
```diff
@@ -155,8 +155,43 @@ class SkwasmGradient extends SkwasmShader implements ui.Gradient {
...
+  SkwasmImageShader._(this.handle);
+
+  factory SkwasmImageShader.imageShader(
+    SkwasmImage image,
+    ui.TileMode tmx,
+    ui.TileMode tmy,
+    Float64List matrix4,
+    ui.FilterQuality? filterQuality,
+  ) => withStackScope((StackScope scope) {
+    final RawMatrix33 localMatrix = scope.convertMatrix4toSkMatrix(matrix4);
```
yjbanov, line 168: Since the matrix is never null, can we allocate on the heap to avoid copying the matrix twice, once here and a second time in `shader_createFromImage`?

lib/web_ui/lib/src/engine/skwasm/skwasm_impl/shaders.dart
```diff
@@ -171,15 +206,21 @@ class SkwasmFragmentProgram implements ui.FragmentProgram {
...
+      handle,
+      shaderData.floatCount,
+      shaderData.textureCount
+    );
   }
 
   RuntimeEffectHandle handle;
   String name;
+  int floatUniformCount;
+  int childShaderCount;
```
yjbanov, line 220: Do all of the fields need to be mutable?

lib/web_ui/lib/src/engine/skwasm/skwasm_impl/shaders.dart
```diff
@@ -240,7 +280,20 @@ class SkwasmFragmentShader extends SkwasmShader implements ui.FragmentShader {
 
   @override
   void setImageSampler(int index, ui.Image image) {
-    // TODO(jacksongardner): implement this when images are implemented
+    final SkwasmImageShader shader = SkwasmImageShader.imageShader(
+      image as SkwasmImage,
+      ui.TileMode.clamp,
+      ui.TileMode.clamp,
+      toMatrix64(Matrix4.identity().storage),
```
yjbanov, line 287: We should probably cache this.

lib/web_ui/lib/src/engine/skwasm/skwasm_impl/surface.dart
```diff
@@ -22,33 +24,55 @@ class SkwasmSurface {
...
+      _handle,
+      image.handle,
+      format.index,
+    );
+    final int context = await _registerCallback(callbackId);
+    final SkDataHandle dataHandle = SkDataHandle.fromAddress(context);
+    final int byteCount = skDataGetSize(dataHandle);
+    final Pointer<Uint8> dataPointer = skDataGetConstPointer(dataHandle).cast<Uint8>();
+    final Uint8List output = Uint8List(byteCount);
+    for (int i = 0; i < byteCount; i++) {
```
yjbanov, line 59: Is there any equivalent to `TypedArray.set` for copying one array into another in WasmGC? Then we could create a `Uint8List` view into the linear memory and copy it into the GC space as one bulk operation.

lib/web_ui/skwasm/surface.cpp
```diff
@@ -77,13 +94,23 @@ class Surface {
...
+    uint32_t callbackId = ++_currentCallbackId;
+    image->ref();
+
+    emscripten_dispatch_to_thread(_thread, EM_FUNC_SIG_VIIII,
+                                  reinterpret_cast<void*>(fRasterizeImage),
+                                  nullptr, this, image, format, callbackId);
+    return callbackId;
   }
 
   // Main thread only
```
yjbanov, line 111: Is there a way to validate this property at runtime?

lib/web_ui/skwasm/surface.cpp
```diff
@@ -176,20 +203,53 @@ class Surface {
     _surface->flush();
   }
 
+  void _rasterizeImage(SkImage* image,
+                       ImageByteFormat format,
+                       uint32_t callbackId) {
+    sk_sp<SkData> data;
+    if (format == ImageByteFormat::png) {
+      data = SkPngEncoder::Encode(_grContext.get(), image, {});
```
yjbanov, line 211: A-ha! I bet removing the dependency on `SkPngEncoder` will take us below 1MB :)

lib/web_ui/lib/src/engine/skwasm/skwasm_impl/image.dart
```diff
@@ -2,39 +2,67 @@
...
+    final SkDataHandle dataHandle = skDataCreate(pixels.length);
+    final Pointer<Uint8> dataPointer = skDataGetPointer(dataHandle).cast<Uint8>();
+    for (int i = 0; i < pixels.length; i++) {
+      dataPointer[i] = pixels[i];
+    }
+    final ImageHandle imageHandle = imageCreateFromPixels(
+      dataHandle,
+      width,
+      height,
+      format == ui.PixelFormat.bgra8888,
```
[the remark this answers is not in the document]
eyebrowsoffire: Well, we just have the two formats in web:
```
enum PixelFormat {
  rgba8888,
  bgra8888,
}
```
So I figured it was simpler to pass a bool for now. We can change it to an enum later if we have more than two.

lib/web_ui/lib/src/engine/skwasm/skwasm_impl/renderer.dart
eyebrowsoffire: I think expecting the compiler to figure out that `targetHeight` is non-null here is a little unrealistic. I might be able to restructure the code a little to make it more obvious to the compiler though... Hmmm...

lib/web_ui/lib/src/engine/skwasm/skwasm_impl/renderer.dart
eyebrowsoffire: I was doing this to be consistent with `skiaDecodeImageFromPixels`, which issues this warning under this condition.

lib/web_ui/lib/src/engine/skwasm/skwasm_impl/shaders.dart
eyebrowsoffire: Two stack allocations is likely cheaper than one heap allocation. Also, I'm not entirely sure it really wins anything to do things on the heap. `SkMatrix` is actually not a POD, so you have to actually call the constructor on it (we can't just try to fake construction of it through memory accesses). I wouldn't worry about it much, this is likely to be very fast.

lib/web_ui/lib/src/engine/skwasm/skwasm_impl/shaders.dart
eyebrowsoffire: Definitely not. I should change this.

lib/web_ui/lib/src/engine/skwasm/skwasm_impl/shaders.dart
eyebrowsoffire: Actually, I think I should just make this optional and not pass the data at all for the identity case.

lib/web_ui/lib/src/engine/skwasm/skwasm_impl/surface.dart
eyebrowsoffire: You want https://github.com/WebAssembly/gc/issues/367, which sadly won't make WasmGC MVP.

lib/web_ui/skwasm/shaders.cpp
eyebrowsoffire: I think I'm going to go the other direction and bubble the nullability up to the higher level APIs. There is at least one place I could avoid passing this data altogether.

lib/web_ui/skwasm/surface.cpp
eyebrowsoffire: Yes, I think I found one. I can make it an assert so that it only executes in debug mode.

lib/web_ui/skwasm/surface.cpp
eyebrowsoffire: I'm not really sure we can get rid of this though. It might be possible... maybe draw the image to a canvas and then do `toDataUrl()` on it? I don't know.

lib/web_ui/lib/src/engine/skwasm/skwasm_impl/image.dart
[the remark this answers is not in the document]
yjbanov: Oh, that looks like a bug. `PixelFormat` is part of `dart:ui`, which is expected to be identical across all platforms.

lib/web_ui/lib/src/engine/skwasm/skwasm_impl/renderer.dart
yjbanov: Heh, benefits of revisiting old code! :) Yeah, I don't think `skiaDecodeImageFromPixels` should warn either.

lib/web_ui/lib/src/engine/skwasm/skwasm_impl/shaders.dart
yjbanov: Gotcha. I thought the second one was on the heap.

lib/web_ui/skwasm/surface.cpp
yjbanov: Yep, exactly that. I'm not requesting this change in this PR. Just spotted an opportunity for later.

lib/web_ui/lib/src/engine/skwasm/skwasm_impl/surface.dart
yjbanov: Oh well...

lib/web_ui/lib/src/engine/skwasm/skwasm_impl/image.dart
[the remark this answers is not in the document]
eyebrowsoffire: I added support for it in Skwasm, but it's still not supported in all the renderers and the bitmap creation code path. I filed https://github.com/flutter/flutter/issues/126895 as a follow-on.

yjbanov (approved):
lgtm

PR #42026, opened 2023-05-14
Revert "Remove GN staging flag for save layer bounds"

CaseyHillers:
Reverts flutter/engine#41940

See b/282115120

This caused a golden failure on Google Testing where a rectangular avatar became circular.

### DIFF
--- a/tools/gn
+++ b/tools/gn
@@ -300,6 +300,8 @@ def to_gn_args(args):
   gn_args['skia_use_wuffs'] = True
   gn_args['skia_use_expat'] = args.target_os == 'android'
   gn_args['skia_use_fontconfig'] = args.enable_fontconfig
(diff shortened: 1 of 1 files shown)


zanderso (approved):

flar:
First, this change was expected to cause golden file failures. The new behavior is intentional, but not backwards compatible.

Second, it would seem to me that this kind of golden failure is exactly why this behavior has been bad. Those avatars are set inside a circular clip for a reason and this change fills that circular clip. The prior behavior failed to fill the circular clip because the ColorFilter did not fill past the implicit bounds of the children.

flar:
cc @lhkbob

CaseyHillers:
Reland available in https://github.com/flutter/engine/pull/42029. Thanks for investigating!

PR #42029, opened 2023-05-15
Reland "Remove GN staging flag for save layer bounds"

CaseyHillers:
Reverts flutter/engine#42026

After investigating, this is a bug fix for the scubas.


flar (approved):
LGTM

PR #42031, opened 2023-05-15
[Impeller] Add interactive DrawPaint blend test

bdero:
Made this to help debugging clear color problems and I've had it sitting around in my stage for a few days. Today, it ended up leading me to discover several other issues. For example, advanced blends are doing nothing in this case...

Screenshot 2023-05-14 at 7 14 43 PM


jonahwilliams (approved):
LGTM

PR #42033, opened 2023-05-15
[Impeller] Include AndroidSurfaceVulkanImpeller behind a flag

CaseyHillers:
Fixes b/282290672

Google Testing currently does not support vulkan, and constructors like this need to be behind a flag.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/shell/platform/android/platform_view_android.cc
+++ b/shell/platform/android/platform_view_android.cc
@@ -51,9 +51,16 @@ std::unique_ptr<AndroidSurface> AndroidSurfaceFactoryImpl::CreateSurface() {
       }
     case AndroidRenderingAPI::kVulkan:
       FML_DCHECK(enable_impeller_);
-      return std::make_unique<AndroidSurfaceVulkanImpeller>(
+      // TODO(kaushikiska@): Enable this after wiring a preference for Vulkan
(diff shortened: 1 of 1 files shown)


CaseyHillers:
\cc @iskakaushik as FYI. I duplicated your TODO from below on wiring up a flag for impeller here.

dnfield (approved):

CaseyHillers:
I'm going to submit as is to let this roll overnight to unblock Google testing. If there's a need for tests, I'll follow up in a separate PR.

Hixie:
FWIW, looks like in https://github.com/flutter/engine/pull/37486 we used `#if IMPELLER_ENABLE_VULKAN` rather than `#if false`.

test-exempt: removes code

That said, it would be good to add testing for both of these code paths.

Hixie:
why do we have IMPELLER_ENABLE_VULKAN in some places but not others?

dnfield:
> why do we have IMPELLER_ENABLE_VULKAN in some places but not others?

Vulkan is used on platforms/test harnesses besides Android.

Right now, we don't want anyone to turn Vulkan on on Android unless they have rebuilt the engine.

PR #42052, opened 2023-05-15
Fix bounds when painting style is stroke width for image shaders

dnfield:
Fixes https://github.com/flutter/flutter/issues/126739

When we're sampling from the texture, we need to sample from its origin regardless of how much offset the stroke width gives to the path.

### DIFF
--- a/impeller/aiks/aiks_unittests.cc
+++ b/impeller/aiks/aiks_unittests.cc
@@ -161,6 +161,13 @@ void CanRenderTiledTexture(AiksTest* aiks_test, Entity::TileMode tile_mode) {
       ColorSource::MakeImage(texture, tile_mode, tile_mode, {}, {});
   paint.color = Color(1, 1, 1, 1);
   canvas.DrawRect({0, 0, 600, 600}, paint);
+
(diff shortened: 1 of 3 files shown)


jonahwilliams (approved):
LGTM

dnfield:
Need to figure out what's changing in the goldens...

bdero (approved):
LGTM.

I'd recommend adding a golden that draws a stroke with `TiledTextureContents` as the color source. There are a bunch of combinations not covered by goldens yet.

dnfield:
> LGTM.
> 
> I'd recommend adding a golden that draws a stroke with `TiledTextureContents` as the color source. There are a bunch of combinations not covered by goldens yet.

I updated the goldens to draw both a stroke and a fill over the same area and make sure they match. I'm confused though about why some of them are changing. I'm going to be out today but will try to look tomorrow.

dnfield:
Ok, verified locally the golden updates are because I'm not placing the stroke exactly on top of the fill in some cases. Rather than try to figure that out I'm going to approve them. This would still catch regressions.

PR #42060, opened 2023-05-15
[Impeller] Fix 1d threadgroup dispatch size, Implement drawPoint geometry computation with compute, fallback to specialized CPU.

jonahwilliams:
Compute shader implementation of https://github.com/flutter/engine/pull/41803

I found this to be generally more straightforward, and we can revisit the vertex shader approach with GLES.

Benchmark is pending here: https://github.com/flutter/flutter/pull/126728

Note: rendering behavior is identical even though this is now a single draw call. This is becuase strokes have overdraw prevention enabled by default, while this new contents does not

image


impeller/base/validation.cc
```diff
@@ -11,7 +11,7 @@
 namespace impeller {
 
 static std::atomic_int32_t sValidationLogsDisabledCount = 0;
-static bool sValidationLogsAreFatal = false;
+static bool sValidationLogsAreFatal = true;
```
jonahwilliams, line 14: reminder: fix

impeller/entity/shaders/geometry/points.comp
```diff
@@ -0,0 +1,67 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include <impeller/texture.glsl>
+#include <impeller/types.glsl>
+
+// I dunno.
```
jonahwilliams, line 8: 🤷‍♂️

impeller/entity/geometry.cc
```diff
@@ -797,4 +807,210 @@ std::optional<Rect> RectGeometry::GetCoverage(const Matrix& transform) const {
...
+    Matrix effect_transform,
+    const ContentContext& renderer,
+    const Entity& entity,
+    RenderPass& pass) {
+  FML_UNREACHABLE();
+}
+
+/// @brief Compute the exact storage size needed to store the resulting
+/// buffer.
+/// @return
```
dnfield, line 963: Finish comment?

impeller/entity/geometry.cc
```diff
@@ -797,4 +807,210 @@ std::optional<Rect> RectGeometry::GetCoverage(const Matrix& transform) const {
...
+size_t PointFieldGeometry::ComputeCircleDivisions(Scalar scaled_radius,
+                                                  bool round) {
+  if (!round) {
+    return 4;
+  }
+
+  // Note: these values are approximated based on the values returned from
+  // the decomposition of 4 cubics performed by Path::CreatePolyline.
```
dnfield, line 971: Can you write a test that covers this?

impeller/entity/shaders/geometry/points.comp
dnfield: The way we use these shaders, these values are set by the `GridSize` API and get dropped.

I've tried to specify something meaningful if we were to use them, but you could just say `See PointFieldGeometry::GetPositionBuffer` or something.

impeller/entity/shaders/geometry/points.comp
```diff
@@ -0,0 +1,67 @@
...
+layout(local_size_x = 16) in;
+
+layout(std430) readonly buffer PointData {
+  vec2 points[];
```
dnfield, line 12: You should include a `uint count` here to guard against OOB access.

impeller/entity/shaders/geometry/points.comp
```diff
@@ -0,0 +1,67 @@
...
+}
+point_data;
+
+layout(std430) writeonly buffer GeometryData {
+  vec2 geometry[];
```
dnfield, line 17: Probably want a count here too?

impeller/entity/shaders/geometry/points.comp
dnfield: Oh, I see, you put it as a uniform.

impeller/entity/shaders/geometry/points.comp
```diff
@@ -0,0 +1,67 @@
...
+  float16_t radian_step;
+  uint points_per_circle;
+  int divisions_per_circle;
+}
+frame_info;
+
+void main() {
+  uint ident = gl_GlobalInvocationID.x;
+  // TODO(dnfield): https://github.com/flutter/flutter/issues/112683
+  // We should be able to use length here instead of an extra arrgument.
```
dnfield, line 34: I kind of doubt this will ever get fixed, we should probably just remove these.

dnfield (commented):
Neat! Mostly nits

impeller/entity/shaders/geometry/points.comp
jonahwilliams: See comment

impeller/entity/shaders/geometry/points.comp
jonahwilliams: Done

impeller/entity/geometry.cc
jonahwilliams: Done

impeller/entity/geometry.cc
jonahwilliams: Done

impeller/entity/geometry.cc
```diff
@@ -797,4 +807,210 @@ std::optional<Rect> RectGeometry::GetCoverage(const Matrix& transform) const {
...
+      renderer.GetContext()->GetResourceAllocator()->CreateBuffer(buffer_desc);
+
+  ComputeCommand cmd;
+  cmd.label = "Points Geometry";
+  cmd.pipeline = renderer.GetPointComputePipeline();
+
+  PS::FrameInfo frame_info;
+  frame_info.count = points_.size();
+  frame_info.radius = radius;
+  frame_info.radian_start = round_ ? 0.0f : 0.785398f;
```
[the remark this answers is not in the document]
jonahwilliams: Done

impeller/entity/shaders/geometry/points.comp
jonahwilliams: Done

dnfield (approved):

jonahwilliams:
I found a problem with how dispatch was being done. We were setting the threadGroup size as if it were the threadsize, which resulted in us doing way more work than necessary. I've now fixed this for one dimension, so that the thread group size is correctly computed as the units of work / maxThreadgroupSize. This is still broken for 2D. fyi @dnfield

impeller/renderer/backend/metal/compute_pass_mtl.mm
```diff
@@ -248,17 +248,29 @@ static bool Bind(ComputePassBindingsCache& pass,
     // https://github.com/flutter/flutter/issues/110619
 
     // For now, check that the sizes are uniform.
-    FML_DCHECK(grid_size == thread_group_size);
+    // FML_DCHECK(grid_size == thread_group_size);
```
jonahwilliams, line 251: I think we should rethink this API. What would make sense to me is 1. specifying the units of work in a grid size.

Maybe there are cases where we need to control threadgroup size? not sure

impeller/renderer/backend/metal/compute_pass_mtl.mm
```diff
@@ -246,19 +250,28 @@ static bool Bind(ComputePassBindingsCache& pass,
...
+              threadsPerThreadgroup:MTLSizeMake(maxTotalThreadsPerThreadgroup,
+                                                1, 1)];
+    } else {
+      while (width * height > maxTotalThreadsPerThreadgroup) {
+        width = std::max(1LL, width / 2);
+        height = std::max(1LL, height / 2);
+      }
+
+      auto size = MTLSizeMake(width, height, 1);
+      [encoder dispatchThreadgroups:size threadsPerThreadgroup:size];
```
dnfield, line 273: This is ok but we should really fix this in a separate PR

dnfield (approved):

PR #42061, opened 2023-05-15
[Windows] Add force redraw to the C++ client wrapper

loic-sharma:
This change adds a C++ client wrapper to Windows's "force redraw" C API. This API can be used to schedule a frame.

Part of https://github.com/flutter/flutter/issues/119415

## Background

The Windows runner has a race at startup:

1. **Platform thread**: creates a hidden window
2. **Platform thread**: launches the Flutter engine
3. **UI/Raster threads**: renders the first frame
4. **Platform thread**: Registers a callback to show the window once the next frame has been rendered.

Steps 3 and 4 happen in parallel and it is possible for step 3 to complete before step 4 starts. In this scenario, the next frame callback is never called and the window is never shown.

The Windows runner will be updated to call the "force redraw" API after it registers the next frame callback. If step 3 hasn't completed yet, the "force redraw" will no-op as a frame is already scheduled. If step 3 has already completed, the "force redraw" will schedule another frame, which will call the next frame callback and show the window.

See this discussion below on why we cannot avoid changing the Windows runner to fix this issue: https://github.com/flutter/engine/pull/42061#issuecomment-1550080722

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/shell/platform/windows/client_wrapper/flutter_view_controller.cc
+++ b/shell/platform/windows/client_wrapper/flutter_view_controller.cc
@@ -29,6 +29,10 @@ FlutterViewController::~FlutterViewController() {
   }
 }
 
+void FlutterViewController::ForceRedraw() {
+  FlutterDesktopViewControllerForceRedraw(controller_);
+}
+
 std::optional<LRESULT> FlutterViewController::HandleTopLevelWindowProc(
     HWND hwnd,
     UINT message,
--- a/shell/platform/windows/client_wrapper/flutter_view_controller_unittests.cc
+++ b/shell/platform/windows/client_wrapper/flutter_view_controller_unittests.cc
@@ -46,7 +46,9 @@ class TestWindowsApi : public testing::StubFlutterWindowsApi {
 
   bool engine_destroyed() { return engine_destroyed_; }
   bool view_controller_destroyed() { return view_controller_destroyed_; }
-  bool view_controller_force_redrawed() { return view_controller_force_redrawed_; }
+  bool view_controller_force_redrawed() {
+    return view_controller_force_redrawed_;
+  }
 
  private:
   bool engine_destroyed_ = false;
--- a/shell/platform/windows/client_wrapper/flutter_view_controller_unittests.cc
+++ b/shell/platform/windows/client_wrapper/flutter_view_controller_unittests.cc
@@ -28,6 +28,11 @@ class TestWindowsApi : public testing::StubFlutterWindowsApi {
   void ViewControllerDestroy() override { view_controller_destroyed_ = true; }
 
   // |flutter::testing::StubFlutterWindowsApi|
+  void ViewControllerForceRedraw() override {
+    view_controller_force_redrawed_ = true;
+  }
+
+  // |flutter::testing::StubFlutterWindowsApi|
   FlutterDesktopEngineRef EngineCreate(
       const FlutterDesktopEngineProperties& engine_properties) override {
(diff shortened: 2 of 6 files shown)


cbracken (commented):
The change itself seems technically correct and tested, but I wonder if there's a way to do this without adding public API -- i.e. trigger this from the internal window.cc when it goes from hidden to shown?

I don't think there's necessarily anything wrong with adding public API to support this -- we expose frame scheduling to the Dart side app too -- but it would be nice to just do the right thing automatically with no new code in the runner, if we can.

loic-sharma:
> but it would be nice to just do the right thing automatically with no new code in the runner, if we can.

I don't think we can avoid changing the runner. The "correct" solution would be to add the view and register the next frame callback atomically. This would require non-trivial embedder API changes. Currently both of these operations work by enqueueing work to the raster thread's task runner - nothing prevents these two tasks from being interleaved with raster operations submitted by the UI thread. In other words, we cannot avoid this race simply by moving/reordering when we add the view or the next frame callback.

Another potential solution would be to register the "show window" callback after we've created the view but before we've sent the window metrics:

https://github.com/flutter/engine/blob/5cf141f7c03c64f42a77d91335d13521f0402f4c/shell/platform/windows/flutter_windows.cc#L78-L87

This would require changes to the Windows embedder's API: the `FlutterViewController`'s constructor would need to accept a "show window" callback that is passed to the Windows embedder's `FlutterDesktopViewControllerCreate` (or some fork to avoid breaking changes). The runner would still need to be modified to something like:

```diff
bool FlutterWindow::OnCreate() {
  // ...
  flutter_controller_ = std::make_unique<flutter::FlutterViewController>(
-     frame.right - frame.left, frame.bottom - frame.top, project_);
+     frame.right - frame.left,
+     frame.bottom - frame.top,
+     project_,
+     [&]() {
+       this->Show();
+     });
  // Ensure that basic setup of the controller was successful.
  if (!flutter_controller_->engine() || !flutter_controller_->view()) {
    return false;
  }
  RegisterPlugins(flutter_controller_->engine());
  SetChildContent(flutter_controller_->view()->GetNativeWindow());

- flutter_controller_->engine()->SetNextFrameCallback([&]() {
-   this->Show();
- });

  return true;
}
```

The "force frame" approach requires the following runner modification:

```diff
```diff
bool FlutterWindow::OnCreate() {
  // ...
  flutter_controller_ = std::make_unique<flutter::FlutterViewController>(
     frame.right - frame.left, frame.bottom - frame.top, project_);
  // Ensure that basic setup of the controller was successful.
  if (!flutter_controller_->engine() || !flutter_controller_->view()) {
    return false;
  }
  RegisterPlugins(flutter_controller_->engine());
  SetChildContent(flutter_controller_->view()->GetNativeWindow());

  flutter_controller_->engine()->SetNextFrameCallback([&]() {
    this->Show();
  });

+ // It is possible Flutter rendered the first frame before the "show window"
+ // callback was registered. In case this happened, schedule another frame to
+ // ensure the window is shown. This no-ops if the first frame hasn't completed yet.
+ flutter_controller_->ForceFrame();

  return true;
}
```


This isn't possible without changing the embedder API. The "correct" solution would add the view and register a next frame callback atomically.

cbracken (approved):
Thanks for the detailed write-up; agreed the required embedder API changes to work around this aren't worth it, plus we already expose this to the Dart bits of the app and it's in the C API anyway.

lgtm

yaakovschectman (approved):
LGTM if we want this in the API

PR #42062, opened 2023-05-16
Implementing TextScaler for nonlinear text scaling

LongCatIsLooong:
`platformTextScaler` doesn't need to be per window, assuming the SP -> DiP mapping is always the same on the same device (unless the user changes the preference), and does not depend on the display device's pixel density (it's confirmed).

Design doc: https://docs.google.com/document/d/1-DlElw3zWRDlqoc9z4YfkU9PbBwRTnc7NhLM97ljGwk/edit#

## Pre-launch Checklist

- [ ] I signed the [CLA].


lib/ui/platform_dispatcher.dart
```diff
@@ -136,6 +136,9 @@ class PlatformDispatcher {
   static final PlatformDispatcher _instance = PlatformDispatcher._();
 
   _PlatformConfiguration _configuration = const _PlatformConfiguration();
+  // The expression is evaluated when `platformTextScaler` is accessed for the
+  // first time.
+  late final bool _platformTextScalerImplemented = _getScaledFontSize(14) != -1;
```
goderbauer, line 141: Why these magic numbers: "14" and "-1"?

lib/ui/platform_dispatcher.dart
```diff
@@ -968,21 +971,29 @@ class PlatformDispatcher {
...
+  @Deprecated(
+    'Use platformTextScaler instead. '
+    'This feature was deprecated after 3.11.0-5.0.pre.',
+  )
   double get textScaleFactor => _configuration.textScaleFactor;
 
-  /// A callback that is invoked whenever [textScaleFactor] changes value.
+  /// Deprecated. Will be removed in a future version of Flutter.
+  ///
+  /// This field is renamed to [onTextScalerChanged].
```
goderbauer, line 985: nit: it is not just a rename, right? The semantics also changed, presumably, it fires when textScaler changes and not when textscalefactor changes?

lib/ui/platform_dispatcher.dart
```diff
@@ -1269,6 +1280,40 @@ class PlatformDispatcher {
...
+  /// Given the font size the app developer specified, this [TextScaler]
+  /// converts it to the preferred font size that accounts for platform-wide
+  /// text scaling.
+  ///
+  /// The scaled value of the same font size input may change if the user changes
+  /// the text scaling preference (in system settings for example). The
+  /// [onTextScaleFactorChanged] callback can be used to monitor such changes.
+  ///
+  /// Instead of directly using this getter, applications should typically use
+  /// `MediaQuery.scaledFontSizeOf` instead to retrive the scaled font size in a
```
goderbauer, line 1297: put MediaQuery.scaledFontSizeOf in [] to link to it in docs?

lib/ui/platform_dispatcher.dart
```diff
@@ -1269,6 +1280,40 @@ class PlatformDispatcher {
...
+  /// widget tree, so text resizes properly when the text scaling preference
+  /// changes.
+  ///
+  /// See also:
+  ///
+  ///  * [WidgetsBindingObserver], for a mechanism at the widgets layer to
+  ///    observe when this callback is invoked.
+  TextScaler get platformTextScaler {
```
goderbauer, line 1305: Just to double check: This is on the PlatformDispatcher because it is a global thing for the entire app? This is not going to be a per-view setting?

lib/ui/platform_dispatcher.dart
```diff
@@ -1269,6 +1280,40 @@ class PlatformDispatcher {
...
+    final double textScaleFactor = _configuration.textScaleFactor;
+    if (textScaleFactor == 1) {
+      return TextScaler.noScaling;
+    }
+    return _platformTextScalerImplemented ? _SystemTextScaler(textScaleFactor) : TextScaler.linear(textScaleFactor);
+  }
+
+  // Returns a negative number when there's an error:
+  // -1 : GetScaledFontSize is not implemented on the current platform.
```
goderbauer, line 1314: Wondering if there's a way to make errors a little more structured to make this API a little more ergonomic and avoid magic numbers like -1.

lib/ui/platform_dispatcher.dart
```diff
@@ -2297,3 +2342,155 @@ enum DartPerformanceMode {
...
+    assert(fontSize >= 0);
+    assert(fontSize.isFinite);
+    switch (PlatformDispatcher._getScaledFontSize(fontSize)) {
+      case >= 0 && final double x when fontSize >= 0:
+        return x;
+      case == -1 when fontSize >= 0:
+        throw StateError('GetScaledFontSize is not implemented on the platform. This should not be reached.');
+      // Fallback to linear scaling when there's an error. Relevant error messages
+      // should already be printed to stderr.
+      case final double errorCode:
```
goderbauer, line 2480: Why does this case exist?

lib/web_ui/lib/platform_dispatcher.dart
```diff
@@ -108,15 +108,29 @@ abstract class PlatformDispatcher {
...
+  )
   double get textScaleFactor;
 
   bool get nativeSpellCheckServiceDefined => false;
 
   bool get brieflyShowPassword => true;
 
+  /// Deprecated. Will be removed in a future version of Flutter.
+  ///
+  /// This field is renamed to [onTextScalerChanged].
```
goderbauer, line 123: I believe, doc comments do not need to be duplicated in the web_ui

lib/web_ui/lib/platform_dispatcher.dart
```diff
@@ -492,3 +508,86 @@ enum DartPerformanceMode {
...
+  double scale(double fontSize);
+
+  TextScaler clamp(double minScaleFactor, double maxScaleFactor) {
+    return minScaleFactor == maxScaleFactor
+      ? TextScaler.linear(minScaleFactor)
+      : _ClampedTextScaler(this, minScaleFactor, maxScaleFactor);
+  }
+}
+
+final class _LinearTextScaler implements TextScaler {
```
goderbauer, line 528: Is there a way to share the implementation of this with regular dart:ui? This is just a straight-up copy, right? Would make this easier to maintain...

goderbauer: Same with _ClampedTextScaler.

goderbauer: (Makes me wonder if dart:ui should maybe just expose the `_getScaledFontSize` and the TestScaler stuff lives in the framework)...

lib/ui/platform_dispatcher.dart
```diff
@@ -2297,3 +2342,155 @@ enum DartPerformanceMode {
...
+        && maxScale == other.maxScale
+        && (minScale == maxScale || scaler == other.scaler);
+  }
+
+  @override
+  int get hashCode => minScale == maxScale ? minScale.hashCode : Object.hash(scaler, minScale, maxScale);
+}
+
+final class _SystemTextScaler extends TextScaler {
+  const _SystemTextScaler(this.textScaleFactor);
```
goderbauer, line 2465: Why is the systemTextScaler dependent on the textScaleFactor?

lib/ui/platform_dispatcher.dart
LongCatIsLooong: currently the android implementation returns -2 if the delegate is not set, for example.

lib/ui/platform_dispatcher.dart
LongCatIsLooong: It's been confirmed that this is per Android device not per monitor.  I was also told device manufacturers will not be able to override the text scaling curve: maybe we can avoid using FFI/JNI and just hard-code the formula in dart.  

For that we will need the Android build version and potentially some other information (it sounds like the curve is going to be different on TVs / watches than on phones). I think I'll ask for more information before proceeding since that sounds like a more performant and eaiser to implement solution than this.

LongCatIsLooong: If I add the `TextScaler` interface to the framework, what would be the easiest way to move it to dart:ui?

LongCatIsLooong:
I've updated the implementation. PTAL. 

There will be a new AndroidX API that also takes `double textScaleFactor` as a parameter, so the function will be 100% pure and we can remove the "configurationGeneration" parameter. @reidbaker is it ok to add additional AndroidX dependencies, or do we prefer using the platform API to reduce the binary size?

lib/ui/platform_dispatcher.dart
```diff
@@ -1366,6 +1469,16 @@ class _PlatformConfiguration {
...
+
+  /// A unique identifier for the platform configuration that this
+  /// [_PlatformConfiguration] represents.
+  ///
+  /// This identifier can be used to to prevent race conditions, when the
+  /// Flutter application synchronously calls a platform function, and the
+  /// platform's configuration has changed on the platform thread but a new
+  /// [_PlatformConfiguration] has not been created for the Flutter application.
+  /// See the [_getScaledFontSize] function for an example.
+  final int? configurationGeneration;
```
chunhtai, line 1481: generation seems like a weird name, maybe token or iteration?

lib/ui/window/platform_configuration.h
```diff
@@ -217,6 +217,26 @@ class PlatformConfigurationClient {
...
+  ///
+  /// @return     The scaled font size in logical pixels. Returns a negative
+  ///             number when an error is encountered:
+  ///              - Returns -1 if this method isn't implemented on the current
+  ///                platform.
+  ///              - Returns -2 if the given configuration_id does not match
+  ///                the platform's text scale factor value.
+  ///
+  virtual double GetScaledFontSize(double unscaled_font_size,
+                                   int configuration_id) const = 0;
```
chunhtai, line 238: configuration_id i assume this is configuration_generation? we should unify the name

lib/ui/platform_dispatcher.dart
```diff
@@ -1299,6 +1303,102 @@ class PlatformDispatcher {
...
+      return null;
+    }
+    final double? cachedValue = cache[unscaledFontSize];
+    if (cachedValue != null) {
+      assert(cachedValue >= 0);
+      return cachedValue;
+    }
+
+    final double unscaledFontSizeInt = unscaledFontSize.toDouble();
+    final double fontSize = PlatformDispatcher._getScaledFontSize(unscaledFontSizeInt, configurationGeneration);
```
chunhtai, line 1365: I am not familiar with FFI, is the performance bad enough that we need to cache the result?

lib/ui/platform_dispatcher.dart
```diff
@@ -1299,6 +1303,102 @@ class PlatformDispatcher {
...
+  /// changes.
+  double? scaleFontSize(double unscaledFontSize) {
+    assert(unscaledFontSize >= 0);
+    assert(unscaledFontSize.isFinite);
+
+    if (textScaleFactor == 1.0) {
+      return unscaledFontSize;
+    }
+
+    final int unscaledFloor = unscaledFontSize.floor();
```
chunhtai, line 1335: shouldn't the TypeValue.applyDimension be able to handle floating point fontsize? why do we convert to int and do linear intepolation?

lib/ui/platform_dispatcher.dart
LongCatIsLooong: renamed to `configurationId`

lib/ui/platform_dispatcher.dart
LongCatIsLooong: The Android team suggested caching the result as the `TypValue.applyDimension` call could be a bit slow.

lib/ui/platform_dispatcher.dart
LongCatIsLooong: Yeah it can handle floating point input. But with interpolation we can reduce the cache size, and the Android implementation does the same

goderbauer:
Can you give some context on how this is intended to be used? Are you going to implement a TextScaler on the framework side that calls the new `scaleFontSize` API? Curious: Wasn't the idea originally (and I think the PR description still describes this) to have dart:ui expose a TextScaler directly? What has changed?

shell/platform/android/io/flutter/embedding/engine/systemchannels/SettingsChannel.java
```diff
@@ -80,7 +105,31 @@ public void send() {
...
+              // Platform channels guarantees FIFO ordering, sentConfiguration
+              // should be either the first element or the second element in the
+              // linked list. Remove older configurations since Flutter ack'd that
+              // it received a more recent version.
+              while (pastConfigurations.size() > 0
+                  && !pastConfigurations.getFirst().equals(sentConfiguration)) {
+                pastConfigurations.remove();
+              }
+            }
+          });
```
goderbauer, line 132: This makes me wonder: why do we need to do the config ID dance in the first place if you know when flutter has acknowledged receipt. Couldn't `SettingsChannel.getDisplayMetrics` just always return the last DisplayMetrics for which Flutter has acknowledged receipt?

goderbauer: Oh, wait. Is the problem there that the acknowledgment is also send async so it would be possible that flutter already knows the latest config, but settings channel hasn't received that acknowledgement yet?

shell/platform/android/io/flutter/embedding/engine/systemchannels/SettingsChannel.java
```diff
@@ -17,13 +21,27 @@ public class SettingsChannel {
...
   public SettingsChannel(@NonNull DartExecutor dartExecutor) {
     this.channel = new BasicMessageChannel<>(dartExecutor, CHANNEL_NAME, JSONMessageCodec.INSTANCE);
   }
 
+  public static DisplayMetrics getPastDisplayMetrics(int configGeneration) {
+    final LinkedList<SentConfiguration> pastConfigurations = SentConfiguration.sentConfigurations;
+    while (pastConfigurations.size() > 0
+        && pastConfigurations.getFirst().generationNumber != configGeneration) {
+      pastConfigurations.remove();
+    }
```
goderbauer, line 37: Some of the higher level doc comments should mention that you cannot request font scaling for arbitrary generations, you have to use the latest generation you saw.

shell/platform/android/io/flutter/embedding/engine/systemchannels/SettingsChannel.java
```diff
@@ -17,13 +21,27 @@ public class SettingsChannel {
...
 
   public SettingsChannel(@NonNull DartExecutor dartExecutor) {
     this.channel = new BasicMessageChannel<>(dartExecutor, CHANNEL_NAME, JSONMessageCodec.INSTANCE);
   }
 
+  public static DisplayMetrics getPastDisplayMetrics(int configGeneration) {
+    final LinkedList<SentConfiguration> pastConfigurations = SentConfiguration.sentConfigurations;
+    while (pastConfigurations.size() > 0
+        && pastConfigurations.getFirst().generationNumber != configGeneration) {
+      pastConfigurations.remove();
```
goderbauer, line 36: My understanding is that `SentConfiguration.sentConfigurations` is modified from two separate threads: the code here is executed on the UI thread, and the code around line 129 is executed on the platform thread. Are these operations thread save? Looking at it naively, I would say they are not and if thread switching happens at a bad time we could end up with data loss?

Scenario: Let's say SentConfiguration.sentConfigurations looks like this: A -> B and the platform thread is still waiting for acknowledgement of B, which the UI thread has already received, but the ack has not been processed on the platform thread yet. For simplicity, let's assume that we only have one core to execute threads:

The UI thread is running and now asks for the display metrics of generation B. `getPastDisplayMetrics` above will now think it can delete A from the list. However, right before line 36 executes, the threads switch and the platform thread executes. It now processes the ack and proceeds to delete generation A. Some time after that, the threads switch again, the UI thread runs and it can finally execute line 36, which will now incorrectly delete generation B. 🔥 🔥 🔥 

Ideally, we would find a solution to not require complicated thread synchronization mechanisms... Would it be possible to somehow send the data required to do the non-linear scaling over to the UI thread with the rest of the platform metrics so we don't have to call back and deal with this race condition at all?

goderbauer: (Or if that's not an option and we need to do this complicated dance with generations we could probably just change it to have the UI thread only read this data, but never modify it)

lib/ui/platform_dispatcher.dart
```diff
@@ -1366,6 +1469,16 @@ class _PlatformConfiguration {
...
+
+  /// A unique identifier for the platform configuration that this
+  /// [_PlatformConfiguration] represents.
+  ///
+  /// This identifier can be used to to prevent race conditions, when the
+  /// Flutter application synchronously calls a platform function, and the
+  /// platform's configuration has changed on the platform thread but a new
+  /// [_PlatformConfiguration] has not been created for the Flutter application.
+  /// See the [_getScaledFontSize] function for an example.
+  final int? configurationId;
```
goderbauer, line 1481: document (or change): Why is this nullable? what does it mean if this is null?

goderbauer: Document: that this must be set if the platform wants to implement `scaleFontSize`. If it doesn't set this, `_scaleAndMemoize` will just silently pretend the platform doesn't support font scaling. (which is kinda a surprising side effect of this property....)

shell/platform/android/io/flutter/embedding/engine/systemchannels/SettingsChannel.java
goderbauer: FWIW, the naming `generation` to me seems a little more fitting, since this is not a classical ID.

shell/platform/android/io/flutter/embedding/engine/systemchannels/SettingsChannel.java
```diff
@@ -80,7 +105,31 @@ public void send() {
               + "\n"
               + "platformBrightness: "
               + message.get(PLATFORM_BRIGHTNESS));
-      channel.send(message);
+      final DisplayMetrics metrics = this.displayMetrics;
+      if (Build.VERSION.SDK_INT < 34 || metrics == null) {
```
goderbauer, line 109: If we guard this by SDK version, should we then also throw if `getPastDisplayMetrics` is accessed in an unsupported android version since it unexpectedly wouldn't return you any metrics.

lib/ui/platform_dispatcher.dart
```diff
@@ -1299,6 +1303,102 @@ class PlatformDispatcher {
...
+  // other platforms.
+  //
+  // The `configurationGeneration` parameter tells the embedder which platform
+  // configuration to use for computing the scaled font size. When the user
+  // changes the platform configuration, the configuration data will first be
+  // made available on the platform thread before being dispatched asynchronously
+  // to the Flutter UI thread. Since this call is synchronous, it could be using
+  // the newer configuration that has yet been sent to Flutter. This call should
+  // use the [PlatformDispatcher]'s latest configuration generation to avoid such
+  // a race.
```
goderbauer, line 1394: This should document that you can only use the last generation known by the PlatformDispatcher. You cannot use any arbitrary old generations. (which I think makes "generation" a better name for this then the generic "ID").

goderbauer:
The fact that we have to deal with these ugly threading issues is a smell in this implementation, IMHO. Is there any way we can avoid all this, e.g. by sending all the relevant data over to the PlatformDispatcher (or the framework) to do the text scaling right there instead of requiring these problematic calls back to the embedder?

lib/ui/platform_dispatcher.dart
```diff
@@ -1299,6 +1303,102 @@ class PlatformDispatcher {
...
+  /// Scales the `unscaledFontSize`, according to the user's platform
+  /// preferences.
+  ///
+  /// Many platforms allow users to scale text globally for better readability.
+  /// Given the font size the app developer specified in logical pixels, this
+  /// method converts it to the preferred font size (also in logical pixels) that
+  /// accounts for platform-wide text scaling.
+  ///
+  /// Returns null if the platform only supports linear text scaling, or does
+  /// not support platform-wide text scaling. The return value is always
```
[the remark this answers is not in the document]
LongCatIsLooong: I thought that gave a bit more information (so the framework knows whether the platform needs nonlinear scaling support) but it's probably not the most useful information and it makes the implementation more complicated. Changed the return type to `double`.

lib/ui/platform_dispatcher.dart
LongCatIsLooong: > Is it really that common to have font sizes in 0.1 increments in your app that we would actually benefit from hitting the cache multiple times?

In an animation that interpolates 2 `TextStyle`s with different font sizes? 

> Also, isn't the point of all of this to not do linear scaling like this does now for the x.0 to x.0 range?

Do you mean [x, x + 1] range where x is an integer? Yeah I've confirmed that the Android implementation does that so I guess we can make that assumption for now?

lib/ui/platform_dispatcher.dart
```diff
@@ -1299,6 +1303,102 @@ class PlatformDispatcher {
...
+    return switch ((_scaleAndMemoize(unscaledFloor), _scaleAndMemoize(unscaledCeil))) {
+      (null, _) => null,
+      (_, null) => null,
+      (final double lower, final double upper) => lower + (upper - lower) * (unscaledFontSize - unscaledFloor),
+    };
+  }
+
+  // The _cachedFontSizes variable is set to null if GetScaledFontSize is not
+  // implemented on the current platform.
+  Map<int, double>? _cachedFontSizes = <int, double>{};
```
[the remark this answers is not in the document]
LongCatIsLooong: The variable being null indicates that the platform does not support non-linear scaling (`_getScaledFontSize` returned -1) so the initial value can't be null. I didn't want to use a second flag for that information.

Yeah 1133's wrong thanks for catching that!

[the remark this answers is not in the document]
LongCatIsLooong: I just changed the return type of the public method to `double` so the framework doesn't know whether the scaling curve is linear (in which case there's no need to do caching) so this can only happen in the engine now. 

> Why do we cache in the first place? Are these calculations sooo expensive?

I can't see the implementation so I can't really answer that. I'll get someone that has the answer.

lib/ui/platform_dispatcher.dart
```diff
@@ -1299,6 +1303,102 @@ class PlatformDispatcher {
...
+      case -2:
+        assert(false, 'Flutter Error: incorrect configuration id.');
+        return null;
+      case final double errorCode:
+        assert(false, 'GetScaledFontSize failed with $errorCode');
+        return null;
+    }
+  }
+
+  // Scales the `unscaledFontSize` using the platform's text scaling curve.
```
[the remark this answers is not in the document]
LongCatIsLooong: > Looking at the configurationId dance I wonder if that approach wouldn't be simpler?

True. My understanding is the curve can change from version to version. Right now this seems to be the way to implement it that requires the least amount of maintenance in the future.  There's going to be an AndroidX API available in the future that we can use which doesn't require the configuration id thing.

shell/platform/android/io/flutter/embedding/engine/FlutterJNI.java
```diff
@@ -1282,6 +1285,24 @@ public String[] computePlatformResolvedLocale(@NonNull String[] strings) {
...
   // ----- End Localization Support ----
+  @Nullable
+  public float getScaledFontSize(float fontSize, int configGeneration) {
+    if (Build.VERSION.SDK_INT < 34) {
+      // Returns unimplemented on anything below Android U.
+      return -1f;
+    }
+    final DisplayMetrics metrics = SettingsChannel.getPastDisplayMetrics(configGeneration);
+    if (metrics == null) {
+      Log.e(
```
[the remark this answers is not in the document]
LongCatIsLooong: No it's not supposed to happen.

shell/platform/android/io/flutter/embedding/engine/systemchannels/SettingsChannel.java
LongCatIsLooong: The reply callback is called on the platform thread I think: https://github.com/flutter/engine/blob/9babd40c04b16659d44fa2caa0d1c2cc653c47ca/shell/platform/android/platform_message_response_android.cc#L27

LongCatIsLooong: There's going to be a new AndroidX API that takes the unscaled size and the scale factor, but I don't think it's available yet and I think it adds some additional AndroidX dependencies.

shell/platform/android/io/flutter/embedding/engine/systemchannels/SettingsChannel.java
LongCatIsLooong: well I'll go with generation here and everywhere else then.

shell/platform/android/io/flutter/embedding/engine/systemchannels/SettingsChannel.java
LongCatIsLooong: Let me see if there's an annotation for that.

shell/platform/android/io/flutter/embedding/engine/systemchannels/SettingsChannel.java
LongCatIsLooong: Yeah right now this channel's communication is async in both directions.

lib/ui/platform_dispatcher.dart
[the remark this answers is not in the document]
chunhtai: I am also wondering in why we need the configurationid, can the curve change within the lifetime of the app?

lib/ui/platform_dispatcher.dart
reidbaker: > > Is it really that common to have font sizes in 0.1 increments in your app that we would actually benefit from hitting the cache multiple times?
> 
> In an animation that interpolates 2 `TextStyle`s with different font sizes?
> 
> > Also, isn't the point of all of this to not do linear scaling like this does now for the x.0 to x.0 range?
> 
> Do you mean [x, x + 1] range where x is an integer? Yeah I've confirmed that the Android implementation does that so I guess we can make that assumption for now?

Consider linking to the android implementation in addition to your comment above for why it is safe.

shell/platform/android/io/flutter/embedding/engine/systemchannels/SettingsChannel.java
LongCatIsLooong: I'll update `getPastDisplayMetrics` so it doesn't modify the linked list and add a RW lock.

lib/ui/platform_dispatcher.dart
```diff
@@ -1299,6 +1303,102 @@ class PlatformDispatcher {
...
+    };
+  }
+
+  // The _cachedFontSizes variable is set to null if GetScaledFontSize is not
+  // implemented on the current platform.
+  Map<int, double>? _cachedFontSizes = <int, double>{};
+  double? _scaleAndMemoize(int unscaledFontSize) {
+    final Map<int, double>? cache = _cachedFontSizes;
+    final int? configurationGenerationId = _configuration.configurationId;
+    if (cache == null || configurationGenerationId == null) {
```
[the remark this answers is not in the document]
LongCatIsLooong: Is there a way to test these? Everything seems to be private.

shell/platform/android/test/io/flutter/embedding/engine/FlutterJNITest.java
```diff
@@ -255,4 +256,25 @@ public void setRefreshRateFPS_callsUpdateRefreshRate() {
     // --- Verify Results ---
     verify(flutterJNI, times(1)).updateRefreshRate();
   }
+
+  @Test
+  public void getScaledFontSizeWorks() {
+    final FlutterJNI flutterJNI = new FlutterJNI();
+
+    // Returns -1 below Android U.
+    if (Build.VERSION.SDK_INT < 34) {
```
reidbaker, line 265: Is this the normal pattern for testing on api versions? I would have expected an annotation declaring what apis this test should be run on. Basically I am surprised that this test can fail based on what api level it is runing against instead of defining what apis can run the test.

shell/platform/android/test/io/flutter/embedding/engine/FlutterJNITest.java
```diff
@@ -255,4 +256,25 @@ public void setRefreshRateFPS_callsUpdateRefreshRate() {
     // --- Verify Results ---
     verify(flutterJNI, times(1)).updateRefreshRate();
   }
+
+  @Test
+  public void getScaledFontSizeWorks() {
```
reidbaker, line 261: The name of this test does not really help describe what should happen also I think this test only covers error cases. Consider getScaledFontSizeErrorCases at a minimum if not breaking apart the test for the pre api 34 cases and the generation cases.

reidbaker (commented):
Great work on this!

lib/ui/platform_dispatcher.dart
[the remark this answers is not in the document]
goderbauer: We do have tests somewhere that simulate a metrics update. Might have to ask in the engine chat where those are. You should be able to repurpose one of those to check that text scaling still works after such an update?

reidbaker:
> I've updated the implementation. PTAL. Unfortunately I can't test the majority of the Android implementation since it's behind a `SDKInt > 34` check and robolectric can't seem to do `targetAPI = 34` yet.
> 
> There will be a new AndroidX API that also takes `double textScaleFactor` as a parameter, so the function will be 100% pure and we can replace the "configurationGeneration" parameter with `textScaleFactor` and stop storing `DisplayMetrics` in the Android `SettingsChannel` class. @reidbaker is it ok to add additional AndroidX dependencies, or do we prefer using the platform API to reduce the binary size?

Sorry in review I missed this comment. Feel free to add additional androidx dependencies if you think they would help the code.

goderbauer:
> There will be a new AndroidX API that also takes double textScaleFactor as a parameter, so the function will be 100% pure and we can replace the "configurationGeneration" parameter with textScaleFactor and stop storing DisplayMetrics in the Android SettingsChannel class.

This sounds like it would be the preferred way of doing this for us. Do we know what the timeline for this new API is? Will it be available before non-lienar text scaling becomes a thing in Android? I would prefer to just wait for that over the complexity introduced in this PR. (However, I defer to the Android experts if this is something we absolutely need to have before those APIs become available)

lib/ui/platform_dispatcher.dart
LongCatIsLooong: The implementation isn't in aosp yet unfortunately: https://cs.android.com/android/platform/superproject/main/+/main:frameworks/base/core/java/android/util/TypedValue.java;l=402-420?q=TypedValue.

lib/ui/platform_dispatcher.dart
LongCatIsLooong: Removed error code -2. Is there a way to share the error code definition between java and dart? The -1 literal is used once in each language so would comments that explain the return value suffice?

lib/ui/window/platform_configuration.h
```diff
@@ -217,6 +217,26 @@ class PlatformConfigurationClient {
...
   virtual void RequestDartDeferredLibrary(intptr_t loading_unit_id) = 0;
 
+  //--------------------------------------------------------------------------
+  /// @brief      Synchronously invokes platform-specific APIs to apply the
+  ///             system text scaling on the given unscaled font size.
+  ///
+  /// @param[in]  unscaled_font_size  The unscaled font size specified by the
+  ///                                 app developer. The value is guaranteed
+  ///                                 to be finite and non-negative.
+  /// @param[in]  configuration_id    The unique id of the configuration to use
```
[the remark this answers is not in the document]
LongCatIsLooong: sorry which field to link to?

shell/platform/android/io/flutter/embedding/engine/systemchannels/SettingsChannel.java
```diff
@@ -100,4 +134,120 @@ public enum PlatformBrightness {
...
+   * Flutter UI thread. Configurations exposed as functions that take parameters are typically not
+   * serializable. To allow the Flutter app to access these configurations, one possible solution is
+   * to create dart bindings that allow the Flutter framework to invoke these functions via JNI
+   * synchronously. To ensure the serialized configuration and these functions represent the same
+   * set of configurations at any given time, a "generation" identifier is used to specify the
+   * serialized configuration that the Flutter app most recently received and is currently using.
+   *
+   * <p>A unique generation identifier is generated by the {@link SettingsChannel} and associated
+   * with a configuration when it sends a serialized configuration to the Flutter framework. This
+   * queue keeps different generations of configurations that could be used by the Flutter
```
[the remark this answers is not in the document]
LongCatIsLooong: Changed the documentation a bit to mention that older configurations will be removed.

lib/ui/platform_dispatcher.dart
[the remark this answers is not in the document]
LongCatIsLooong: I tried adding a test to `flutter/lib/ui/fixtures/ui_test.dart`. But if I set `configurationId` to a non-null value there's no way to set it back to null and the default `_getScaledFontSize` implementation throws so there doesn't seem to be a way to test this without breaking other tests.

lib/ui/platform_dispatcher.dart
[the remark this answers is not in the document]
LongCatIsLooong: Yeah when the textScaleFactor changes the scaling curve (as a function of the input font size) changes.

LongCatIsLooong:
Not sure why the linter isn't happy even if I added `ChecksSdkIntAtLeast` so I had to suppress the lint.

> This sounds like it would be the preferred way of doing this for us. Do we know what the timeline for this new API is? Will it be available before non-lienar text scaling becomes a thing in Android? I would prefer to just wait for that over the complexity introduced in this PR. (However, I defer to the Android experts if this is something we absolutely need to have before those APIs become available)

I think it's mostly an implementation detail (although there are some differences, the newer API is an AndroidX API so I think that means nonlinear text scaling will be applied on earlier versions of Android too) that we can change easily?

shell/platform/android/io/flutter/embedding/engine/FlutterJNI.java
```diff
@@ -1282,6 +1285,20 @@ public String[] computePlatformResolvedLocale(@NonNull String[] strings) {
...
+          TAG,
+          "getScaledFontSize called with configurationId "
+              + String.valueOf(configurationId)
+              + ", which can't be found.");
+      return -1f;
+    }
+    return TypedValue.applyDimension(TypedValue.COMPLEX_UNIT_SP, fontSize, metrics)
+        / TypedValue.applyDimension(TypedValue.COMPLEX_UNIT_DIP, 1.0f, metrics);
```
odbol, line 1300: Alternatively, you could just use `metrics.density` instead of calling applyDimension() on 1dp: https://cs.android.com/android/platform/superproject/+/master:frameworks/base/core/java/android/util/TypedValue.java;l=409?q=typedvalue

odbol (commented):
LGTM from an Android text perspective. Thanks!

shell/platform/android/io/flutter/embedding/engine/systemchannels/SettingsChannel.java
```diff
@@ -17,13 +24,30 @@ public class SettingsChannel {
   private static final String BRIEFLY_SHOW_PASSWORD = "brieflyShowPassword";
   private static final String ALWAYS_USE_24_HOUR_FORMAT = "alwaysUse24HourFormat";
   private static final String PLATFORM_BRIGHTNESS = "platformBrightness";
+  private static final String CONFIGURATION_ID = "configurationId";
+
+  // When hasNonlinearTextScalingSupport() returns false, this will not be initialized.
```
goderbauer, line 29: Is this because Java will only initialize static fields on first access and when `hasNonlinearTextScalingSupport` is false this is never accessed?

shell/platform/android/io/flutter/embedding/engine/systemchannels/SettingsChannel.java
```diff
@@ -100,4 +141,122 @@ public enum PlatformBrightness {
...
+   * app most recently received and is currently using.
+   *
+   * <p>A unique generation identifier is generated by the {@link SettingsChannel} and associated
+   * with a configuration when it sends a serialized configuration to the Flutter framework. This
+   * queue keeps different generations of configurations that could be used by the Flutter
+   * framework, and cleans up old configurations that the Flutter framework no longer uses. When the
+   * Flutter framework invokes a function to access the configuration with a generation identifier,
+   * this queue finds the configuration with that identifier and also cleans up configurations that
+   * are no longer needed.
+   */
```
goderbauer, line 166: Somewhere (maybe here?) can we leave a breadcrumb to the future Android API that could replace all this and make this simpler? Just as a reminder for future readers of this code.

goderbauer (commented):
LGTM

Leaving it to the Android embedder people (i.e. @reidbaker) to give approval if they're ok with the changes there.

lib/ui/platform_dispatcher.dart
```diff
@@ -1299,6 +1303,97 @@ class PlatformDispatcher {
...
+        assert(false, 'Unknown error: GetScaledFontSize failed with $errorCode.');
+    }
+    return null;
+  }
+
+  // Calls the platform's text scaling implementation to scale the given
+  // `unscaledFontSize`.
+  //
+  // Currently this is only implemented on newer versions of Android (SDK level
+  // 34, using the `TypedValue#applyDimension` API). This function returns -1 on
```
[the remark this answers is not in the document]
LongCatIsLooong: I think I forgot to update the comments. Updated now.

shell/platform/android/io/flutter/embedding/engine/systemchannels/SettingsChannel.java
LongCatIsLooong: Yes.

LongCatIsLooong:
Hi @reidbaker could you take a look when you get a chance? I think I went through all comments but let me know if I missed any.

lib/ui/platform_dispatcher.dart
reidbaker: Do we have a policy against linking to code search results? If not then it would help a googler and after the change is merged into aosp someone investigation can probably use the path to find the right result.

shell/platform/android/test/io/flutter/embedding/engine/systemchannels/SettingsChannelTest.java
```diff
@@ -0,0 +1,70 @@
...
+    final SettingsChannel settingsChannel = new SettingsChannel(executor);
+
+    final ArgumentCaptor<ByteBuffer> messageCaptor = ArgumentCaptor.forClass(ByteBuffer.class);
+
+    settingsChannel.startMessage().setDisplayMetrics(mock(DisplayMetrics.class)).send();
+
+    verify(executor).send(eq("flutter/settings"), messageCaptor.capture(), isNull());
+  }
+
+  @Test
```
reidbaker, line 38: Should this test have a config = 34?

reidbaker (approved):

shell/platform/android/test/io/flutter/embedding/engine/systemchannels/SettingsChannelTest.java
LongCatIsLooong: I'll give that a try. But iirc Robolectric does not support that api level yet and the test will fail.

LongCatIsLooong: it gave me 
```
io.flutter.embedding.engine.systemchannels.SettingsChannelTest > configurationQueueWorks FAILED
    java.lang.IllegalArgumentException: API level 34 is not available
        at org.robolectric.plugins.UnknownSdk.verifySupportedSdk(UnknownSdk.java:45)
        at org.robolectric.RobolectricTestRunner.getSandbox(RobolectricTestRunner.java:298)
        at org.robolectric.RobolectricTestRunner.getSandbox(RobolectricTestRunner.java:66)
        at org.robolectric.internal.SandboxTestRunner$2.evaluate(SandboxTestRunner.java:244)
```

reidbaker: But you are calling an api 34 method. If roboletric does not support api 34 then update the dependency. Or at minimum use config minSdk to api 33 so that when api 34 is added this test starts running against api 34.

LongCatIsLooong: None of these methods directly or indirectly called is API 34 only I think.

reidbaker: If this code does not indirectly call 34 then shouldnt there be a test for the opposite of setDisplayMetricsDoesNothingOnAPILevel33. Something like setDisplayMetricsDoesXyzBehaviorOnApiLevel34?

reidbaker: I am sorry I am doing a poor job of communicating. Yes from your comments in this pr I see that the existing robolectric dependency does not support api 34. I believe you are adding code that splits and takes a different branch on api 34. If that is the case then you should add a test to cover this new feature otherwise it will be untested. If you need to update a dependency to test the new code that is what should happen. Otherwise this is new code with an untested branch which we want to avoid.

reidbaker: edit: I was mistaken about roboletric api 34 support. 

OK, can you file a follow up bug to add a test when we have a roboletric version that supports api 34?

PR #42087, opened 2023-05-17
[Impeller] Wire up compute tessellator.

jonahwilliams:
Supports compute based path tessellation using metal indirect draw arguments

### DIFF
--- a/impeller/entity/geometry.cc
+++ b/impeller/entity/geometry.cc
@@ -665,7 +665,7 @@ GeometryResult StrokePathGeometry::GetPositionBufferGPU(
   Scalar min_size = 1.0f / sqrt(std::abs(determinant));
   Scalar stroke_width = std::max(stroke_width_, min_size);
 
-  auto vertex_buffer = CreateHostVisibleDeviceBuffer<SS::VertexBuffer<2048>>(
+  auto vertex_buffer = CreateHostVisibleDeviceBuffer<SS::VertexBuffer<4096>>(
       renderer.GetContext(), "VertexBuffer");
   auto indirect_command_arguments =
       CreateHostVisibleDeviceBuffer<SS::IndirectCommandArguments>(
--- a/impeller/entity/geometry.cc
+++ b/impeller/entity/geometry.cc
@@ -4,6 +4,7 @@
 
 #include "impeller/entity/geometry.h"
 
+#include "flutter/fml/trace_event.h"
 #include "impeller/core/device_buffer.h"
 #include "impeller/entity/contents/content_context.h"
 #include "impeller/entity/entity.h"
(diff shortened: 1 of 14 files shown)


impeller/entity/geometry.cc
```diff
@@ -628,6 +638,67 @@ GeometryResult StrokePathGeometry::GetPositionBuffer(
   };
 }
 
+template <typename T>
+std::shared_ptr<DeviceBuffer> CreateHostVisibleDeviceBuffer(
```
jonahwilliams, line 642: TODO: we should find a way to use the transients buffer for these, or at least have a more efficient allocation strategy.

impeller/entity/geometry.cc
```diff
@@ -628,6 +638,67 @@ GeometryResult StrokePathGeometry::GetPositionBuffer(
...
+
+  return {
+      .type = PrimitiveType::kTriangleStrip,
+      .vertex_buffer =
+          {
+              .vertex_buffer = vertex_buffer->AsBufferView(),
+              .index_buffer = {},
+              .vertex_count = 1,
+              .index_type = IndexType::kNone,
+              .indirect_arguments = indirect_command_arguments->AsBufferView(),
```
jonahwilliams, line 694: magic pt 1

impeller/entity/geometry.cc
```diff
@@ -628,6 +638,67 @@ GeometryResult StrokePathGeometry::GetPositionBuffer(
...
+
+  auto vertex_buffer = CreateHostVisibleDeviceBuffer<SS::VertexBuffer<2048>>(
+      renderer.GetContext(), "VertexBuffer");
+  auto indirect_command_arguments =
+      CreateHostVisibleDeviceBuffer<SS::IndirectCommandArguments>(
+          renderer.GetContext(), "IndirectCommandArguments");
+
+  auto status = ComputeTessellator{}
+                    .SetStrokeWidth(stroke_width)
+                    // .SetStrokeJoin(stroke_join_)
```
jonahwilliams, line 676: Todo: actually set properties and check for supported functionality

impeller/renderer/stroke.comp
```diff
@@ -21,10 +21,18 @@ layout(binding = 1) buffer VertexBuffer {
...
 vertex_buffer;
 
-layout(binding = 2) buffer VertexBufferCount {
-  uint count;
+// layout(binding = 2) buffer VertexBufferCount {
+//   uint count;
+// }
+// vertex_buffer_count;
+
+layout(binding = 2) buffer IndirectCommandArguments {
```
jonahwilliams, line 29: This is the magic, a buffer with a specific struct layout (From https://developer.apple.com/documentation/metal/mtldrawprimitivesindirectarguments?language=objc) Can be supplied to a drawPrimitives call

PR #42088, opened 2023-05-17
Implement `ImageFilter`/`ColorFilter`/`MaskFilter` in Skwasm

eyebrowsoffire:
This implements https://github.com/flutter/flutter/issues/126342

This implements `ImageFilter`, `ColorFilter` and `MaskFilter` in Skwasm. This includes support on the `Paint` object, as well as the `SceneBuilder` layers that use these types.


lib/web_ui/lib/src/engine/scene_painting.dart
```diff
@@ -0,0 +1,20 @@
...
+
+// These are additional APIs that are not part of the `dart:ui` interface that
+// are needed internally to properly implement a `SceneBuilder` on top of the
+// generic Canvas/Picture api.
+abstract class SceneCanvas implements ui.Canvas {
+  void saveLayerWithFilter(ui.Rect? bounds, ui.Paint paint, ui.ImageFilter filter);
+}
+
+abstract class ScenePicture implements ui.Picture {
+  ui.Rect get cullRect;
```
yjbanov, line 15: Should this be called `bounds`? I'm not sure how a picture could be "culling" anything 🤔

lib/web_ui/lib/src/engine/scene_painting.dart
```diff
@@ -0,0 +1,20 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:ui/ui.dart' as ui;
+
+// These are additional APIs that are not part of the `dart:ui` interface that
+// are needed internally to properly implement a `SceneBuilder` on top of the
+// generic Canvas/Picture api.
+abstract class SceneCanvas implements ui.Canvas {
+  void saveLayerWithFilter(ui.Rect? bounds, ui.Paint paint, ui.ImageFilter filter);
```
yjbanov, line 11: Consider adding dartdocs to the additional methods.

lib/web_ui/lib/src/engine/skwasm/skwasm_impl/filters.dart
```diff
@@ -0,0 +1,132 @@
...
+  factory SkwasmMaskFilter.fromUiMaskFilter(ui.MaskFilter maskFilter) =>
+    SkwasmMaskFilter._(maskFilterCreateBlur(
+      maskFilter.webOnlyBlurStyle.index,
+      maskFilter.webOnlySigma
+    ));
+
+  final MaskFilterHandle handle;
+  bool _isDisposed = false;
+
+  void dispose() {
```
yjbanov, line 126: I believe the pattern we use is to `assert(!_isDisposed)` in all dispose methods because double-dispose likely points to some app-side bug.

lib/web_ui/lib/src/engine/skwasm/skwasm_impl/layers.dart
```diff
@@ -122,37 +125,54 @@ class ColorFilterLayer
...
-  void pre(ui.Canvas canvas, ui.Rect contentRect) {
-    // TODO(jacksongardner): Implement image filter
+  void pre(SceneCanvas canvas, ui.Rect contentRect) {
+    if (offset != ui.Offset.zero) {
+      canvas.save();
+      canvas.translate(offset.dx, offset.dy);
+    }
+    final ui.Rect adjustedContentRect =
+      (filter as SceneImageFilter).filterBounds(contentRect);
+    print('contentRect: $adjustedContentRect');
```
yjbanov, line 166: Debugging leftovers?

lib/web_ui/lib/src/engine/skwasm/skwasm_impl/paint.dart
```diff
@@ -20,6 +21,22 @@ class SkwasmPaint implements ui.Paint {
...
+  ui.MaskFilter? _maskFilter;
+
+  bool _invertColors = false;
+
+  static final SkwasmColorFilter _invertColorFilter = SkwasmColorFilter.fromEngineColorFilter(
+    const EngineColorFilter.matrix(<double>[
+      -1.0, 0, 0, 1.0, 0, // row
+      0, -1.0, 0, 1.0, 0, // row
+      0, 0, -1.0, 1.0, 0, // row
+      1.0, 1.0, 1.0, 1.0, 0
```
yjbanov, line 37: Add `// the boat` to the last line 😛

yjbanov (approved):

lib/web_ui/lib/src/engine/scene_painting.dart
eyebrowsoffire: This terminology comes from skia itself, see `SkPicture::cullRect`. I think the idea is that if that rectangle can be culled by another operation (like a clip operation or something), it can cull the picture and skip all of its drawing commands.

lib/web_ui/lib/src/engine/skwasm/skwasm_impl/filters.dart
eyebrowsoffire: Well, I don't actually have the finalization registry wired up yet, but the idea is for this to be safe to call more than once, since it might be called manually by the user and it might be called by the finalization registry. I think it makes sense for this to be idempotent, at least for how I envision this being used with the `UniqueRef` stuff.

lib/web_ui/lib/src/engine/skwasm/skwasm_impl/layers.dart
eyebrowsoffire: Whoops

lib/web_ui/lib/src/engine/skwasm/skwasm_impl/filters.dart
yjbanov: Yeah, as long as the object reference does not leak to the framework then it should be fine. The trouble starts when the framework references the object. Then you have a chicken and egg problem, the framework drops the reference and the object is GC'd, but then `FinalizationRegistry` doesn't have an object to call `dispose()` on. In CanvasKit each Skia object is represented using a pair of Dart objects, a `Ck*` object (e.g. `CkPaint`) and a `UniqueRef`. Then we do `FinalizationRegistry.register(ckObject, uniqueRef)` and hand the `ckObject` reference to the framework. When `ckObject` is GC'd, we call `uniqueRef.dispose()`. `UniqueRef` is the only place that could potentially be disposed more than once, once manually and once via `FinalizationRegistry`. However, `CkObject` should only be disposed once, which is where we put `assert(!isDisposed)`.

yjbanov: I'm guessing in Skwasm we don't need `UniqueRef` at all. I think `Pointer` can play the same role without an extra wrapper.

eyebrowsoffire: Yeah I see what you're saying. I know `FinalizationRegistry` basically needs an extra layer of indirection, but I haven't quite figured out how this is all going to work with the Skwasm objects. If you don't mind, I'm just going to kick the can on this issue and revisit it when I hook up the finalization registry stuff.

PR #42096, opened 2023-05-17
[web] Cleanup Vector3

mdebbar:
We have limited use of `Vector3` in the engine, so I converted it into a record and removed almost all of its methods.

Converting it to a record also means it's not mutable anymore, which is a positive in my book. That required a few minor changes in `Matrix4` but nothing major.


lib/web_ui/lib/src/engine/vector_math.dart
```diff
@@ -1091,315 +1062,21 @@ class Matrix4 {
...
-  /// optionally pass your own random number generator.
-  factory Vector3.random([math.Random? rng]) {
-    rng ??= math.Random();
-    return Vector3(rng.nextDouble(), rng.nextDouble(), rng.nextDouble());
-  }
-
-  final Float32List _v3storage;
+const Vector3 kUnitX = (x: 1.0, y: 0.0, z: 0.0);
+const Vector3 kUnitY = (x: 0.0, y: 1.0, z: 0.0);
+const Vector3 kUnitZ = (x: 0.0, y: 0.0, z: 1.0);
```
yjbanov, line 1067: Nice! This actually guarantees immutability, unlike the previous method of leaving comments with warnings.

yjbanov (approved):
lgtm

PR #42104, opened 2023-05-17
[Impeller] Don't rely on zeroing of threadgroup shared memory, don't early return in quad decomp.

jonahwilliams:
Fixes https://github.com/flutter/flutter/issues/126960 . This doesn't reproduce on macOS as that appears to always zero threadgroup shared memory. On iOS this wasn't getting zero'd out, and so we'd just get whatever garbage was in the memory. This would end up as a huge number that would go far oob and cause a memory fault or something

### DIFF
--- a/impeller/compiler/shader_lib/impeller/path.glsl
+++ b/impeller/compiler/shader_lib/impeller/path.glsl
@@ -121,7 +121,7 @@ QuadDecomposition DecomposeQuad(QuadData quad, float tolerance) {
   // This should never happen, but if it does happen be more defensive -
   // otherwise we'll get NaNs down the line.
   if (dd == vec2(0.)) {
-    return QuadDecomposition(0., 0., 0., 0., 0, 0.);
+    dd = vec2(1.0, 1.0);
   }
   float c = Cross(quad.p2 - quad.p1, dd);
   float x0 = dot(d01, dd) * 1. / c;
--- a/impeller/renderer/path_polyline.comp
+++ b/impeller/renderer/path_polyline.comp
@@ -72,10 +72,9 @@ void ProcessCubic(uint ident) {
   uint offset = 0;
   if (quad_count > 0) {
     scratch_sum[ident] = subgroupExclusiveAdd(scratch_count[ident]);
-
     offset = ComputePosition(ident) + quads.count;
   } else {
-    scratch_sum[ident] = 0;
+    scratch_sum[ident] = subgroupExclusiveAdd(scratch_count[ident]);
   }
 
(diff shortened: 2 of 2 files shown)


impeller/renderer/path_polyline.comp
```diff
@@ -72,9 +72,11 @@ void ProcessCubic(uint ident) {
   uint offset = 0;
   if (quad_count > 0) {
     scratch_sum[ident] = subgroupExclusiveAdd(scratch_count[ident]);
-
     offset = ComputePosition(ident) + quads.count;
+  } else {
+    scratch_sum[ident] = subgroupExclusiveAdd(scratch_count[ident]);
   }
```
dnfield, line 78: Consider instead:

```
scratch_sum[ident] = subgroupExclusiveAdd(scratch_count[ident]);
if (quad_count > 0) {
  offset = ComputePosition(ident) + quads.count;
}
```

And similar below

dnfield (approved):
LGTM with nit about restructuring if/elses

dnfield (approved):
LGTM with nit about restructuring if/elses

jonahwilliams:
I'm going to keep debugging this a bit. i think we're still relying on zero initialization somewhere

impeller/compiler/shader_lib/impeller/path.glsl
```diff
@@ -121,7 +121,7 @@ QuadDecomposition DecomposeQuad(QuadData quad, float tolerance) {
   // This should never happen, but if it does happen be more defensive -
   // otherwise we'll get NaNs down the line.
   if (dd == vec2(0.)) {
-    return QuadDecomposition(0., 0., 0., 0., 0, 0.);
+    dd = vec2(1.0, 1.0);
```
jonahwilliams, line 124: This is required to render correctly compared to the current tessellator

jonahwilliams:
Updated

jonahwilliams:
Still crashing on cubics

impeller/compiler/shader_lib/impeller/path.glsl
dnfield: This is the case where the quad is a straight line right?

In that case, the CPU based code just unconditionally puts p2 into the line.

I think here it would work the same if we did `return QuadDecomposition(0., 0., 0., 0., 1., 0.);` - to indicate that this generates a single line (right now it's indicating that it generates no lines).

jonahwilliams: Yeah, this works!

PR #42108, opened 2023-05-17
[web:a11y] support dialogs described by descendants

yjbanov:
Add a new `RouteName` semantic role for nodes with `namesRoute` set without a `scopesRoute`. Such nodes provide a description for the nearest ancestor dialog node. The web equivalent of this is when an element `role="dialog"` is described by pointing to one of its children using `aria-labelledby` and `aria-describedby`. Here's an example:

```
<div
  role="dialog"
  aria-labelledby="dialog1Title"
  aria-describedby="dialog1Desc">
  <h2 id="dialog1Title">Your personal details were successfully updated</h2>
  <p id="dialog1Desc">
    You can change your details at any time in the user account section.
  </p>
  <button>Close</button>
</div>
```

([Source](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/dialog_role))

Flutter currently does not distinguish between "labelled by" and "described by". In my testing, only `aria-describedby` resulted in an announcement of the dialog's description upon opening. `aria-labelledby` required that the dialog element itself be focusable, which is not the case. So I went with `aria-describedby`.

Fixes https://github.com/flutter/flutter/issues/126030

### DIFF
--- a/lib/web_ui/lib/src/engine/semantics/dialog.dart
+++ b/lib/web_ui/lib/src/engine/semantics/dialog.dart
@@ -25,7 +25,7 @@ class Dialog extends RoleManager {
             'namesRoute set, indicating a self-labelled dialog, but it is '
             'missing the label. A dialog should be labelled either by setting '
             'namesRoute on itself and providing a label, or by containing a '
-            'child node with namesRoute that can describe it with it content.'
+            'child node with namesRoute that can describe it with its content.'
           );
         }
         return true;
--- a/lib/web_ui/lib/src/engine/semantics/dialog.dart
+++ b/lib/web_ui/lib/src/engine/semantics/dialog.dart
@@ -77,8 +77,10 @@ class RouteName extends RoleManager {
         // Setting the label for the first time. Wait for the DOM tree to be
         // established, then find the nearest dialog and update its label.
         semanticsObject.owner.addOneTimePostUpdateCallback(() {
-          _lookUpNearestAncestorDialog();
-          _dialog?.describeBy(this);
+          if (!isDisposed) {
+            _lookUpNearestAncestorDialog();
+            _dialog?.describeBy(this);
+          }
         });
       }
     }
--- a/lib/web_ui/lib/src/engine/semantics/dialog.dart
+++ b/lib/web_ui/lib/src/engine/semantics/dialog.dart
@@ -13,26 +13,84 @@ class Dialog extends RoleManager {
   Dialog(SemanticsObject semanticsObject) : super(Role.dialog, semanticsObject);
 
   @override
-  void dispose() {
-    semanticsObject.element.removeAttribute('aria-label');
-    semanticsObject.clearAriaRole();
+  void update() {
+    // If semantic object corresponding to the dialog also provides the label
+    // for itself it is applied as `aria-label`. See also [describeBy].
+    if (semanticsObject.namesRoute) {
+      final String? label = semanticsObject.label;
+      assert(() {
+        if (label == null || label.trim().isEmpty) {
+          printWarning(
+            'Semantic node ${semanticsObject.id} had both scopesRoute and '
+            'namesRoute set, indicating a self-labelled dialog, but it is '
+            'missing the label. A dialog should be labelled either by setting '
+            'namesRoute on itself and providing a label, or by containing a '
+            'child node with namesRoute that can describe it with it content.'
+          );
+        }
+        return true;
+      }());
+      semanticsObject.element.setAttribute('aria-label', label ?? '');
+      semanticsObject.setAriaRole('dialog', true);
+    }
   }
+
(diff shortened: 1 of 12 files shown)


lib/web_ui/lib/src/engine/semantics/semantics.dart
```diff
@@ -827,6 +848,12 @@ class SemanticsObject {
   DomElement? _childContainerElement;
 
   /// The parent of this semantics object.
+  ///
+  /// This value is not final until the tree is finalized. It is not safe to
+  /// rely on this value in the middle of a semantics tree update. It is safe to
+  /// use this value in post-update callback (see
+  /// [EngineSemanticsOwner.addOneTimePostUpdateCallback]).
+  SemanticsObject? get parent => _parent;
```
mdebbar, line 856: Would be nice to have an assert that throws if we are in the middle of an update. Not sure what's the best way to do that though. Maybe a flag `EngineSemanticsOwner.isUpdating` and then:
```suggestion
  SemanticsObject? get parent {
    assert(!owner.isUpdating);
    return _parent;
  }
```

lib/web_ui/lib/src/engine/semantics/dialog.dart
```diff
@@ -13,26 +13,84 @@ class Dialog extends RoleManager {
...
+            'Semantic node ${semanticsObject.id} had both scopesRoute and '
+            'namesRoute set, indicating a self-labelled dialog, but it is '
+            'missing the label. A dialog should be labelled either by setting '
+            'namesRoute on itself and providing a label, or by containing a '
+            'child node with namesRoute that can describe it with its content.'
+          );
+        }
+        return true;
+      }());
+      semanticsObject.element.setAttribute('aria-label', label ?? '');
```
mdebbar, line 33: I don't see where the `aria-label` is set for the case where `namesRoute` is a descendant of `scopesRoute`.

Does it happen through the `LabelAndValue` role manager? If yes, then why are we doing it in the `Dialog` role manager too?

mdebbar (approved):
LGTM with a couple questions/comments.

lib/web_ui/lib/src/engine/semantics/dialog.dart
yjbanov: No, we don't set `aria-label` when a descendant provides  a description. We use `aria-describedby="ID_OF_DESCENDANT"` instead.

lib/web_ui/lib/src/engine/semantics/dialog.dart
```diff
@@ -13,26 +13,84 @@ class Dialog extends RoleManager {
...
+    if (semanticsObject.isLabelDirty) {
+      final Dialog? dialog = _dialog;
+      if (dialog != null) {
+        // Already attached to a dialog, just update the description.
+        dialog.describeBy(this);
+      } else {
+        // Setting the label for the first time. Wait for the DOM tree to be
+        // established, then find the nearest dialog and update its label.
+        semanticsObject.owner.addOneTimePostUpdateCallback(() {
+          _lookUpNearestAncestorDialog();
```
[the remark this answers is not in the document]
yjbanov: Good catch!

lib/web_ui/lib/src/engine/semantics/dialog.dart
mdebbar: And then that descendant is supposed to set the `aria-label`, right? I don't see `RouteName` doing it, so I'm assuming that `LabelAndValue` is doing it. But `LabelAndValue` should be doing it for `Dialog` as well. Maybe I'm missing something here.

mdebbar: Ah I see. I was confused by this test:
```dart
expectSemanticsTree('''
  <sem aria-describedby="flt-semantic-node-2" style="$rootSemanticStyle">
    <sem-c>
      <sem>
        <sem-c>
          <sem aria-label="$label"></sem>
        </sem-c>
      </sem>
    </sem-c>
  </sem>
''');
```

yjbanov: Oh, it's because the descendant can provide the description by applying `aria-label` to itself, but it can also be a group of elements that provide a description together.

lib/web_ui/lib/src/engine/semantics/semantics.dart
yjbanov: Done.

lib/web_ui/lib/src/engine/semantics/dialog.dart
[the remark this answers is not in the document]
yjbanov: Done.

mdebbar (approved):
Thanks for all the improvements! LGTM again!

PR #42113, opened 2023-05-17
[Impeller] Made other vulkan objects cleanup properly.

gaaclarke:
fixes https://github.com/flutter/flutter/issues/127058

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/impeller/renderer/backend/vulkan/context_vk.cc
+++ b/impeller/renderer/backend/vulkan/context_vk.cc
@@ -29,6 +29,24 @@
 
 VULKAN_HPP_DEFAULT_DISPATCH_LOADER_DYNAMIC_STORAGE
 
+namespace {
+template <typename T>
+class scoped_reset {
+ public:
+  scoped_reset(T* addr, T&& value) : addr_(addr) { *addr_ = std::move(value); }
+  ~scoped_reset() {
+    if (addr_) {
+      addr_->reset();
+    }
+  }
+  void release() { addr_ = nullptr; }
+  const typename T::element_type& get() const { return addr_->get(); }
+
+ private:
+  T* addr_;
+};
+}  // namespace
+
 namespace impeller {
 
 // TODO(csg): Fix this after caps are reworked.
@@ -278,7 +296,8 @@ void ContextVK::Setup(Settings settings) {
     VALIDATION_LOG << "Could not create logical device.";
     return;
   }
-  vk::UniqueDevice device = std::move(device_result.value);
+  scoped_reset<vk::UniqueDevice> device(&device_,
+                                        std::move(device_result.value));
 
   if (!caps->SetDevice(physical_device.value())) {
     VALIDATION_LOG << "Capabilities could not be updated.";
@@ -307,8 +326,7 @@ void ContextVK::Setup(Settings settings) {
   /// Setup the pipeline library.
   ///
   auto pipeline_library = std::shared_ptr<PipelineLibraryVK>(
-      new PipelineLibraryVK(weak_from_this(),                     //
-                            device.get(),                         //
+      new PipelineLibraryVK(shared_from_this(),                   //
                             caps,                                 //
                             std::move(settings.cache_directory),  //
                             settings.worker_task_runner           //
@@ -323,7 +341,7 @@ void ContextVK::Setup(Settings settings) {
       std::shared_ptr<SamplerLibraryVK>(new SamplerLibraryVK(device.get()));
 
   auto shader_library = std::shared_ptr<ShaderLibraryVK>(
-      new ShaderLibraryVK(device.get(),                    //
+      new ShaderLibraryVK(weak_from_this(),                //
                           settings.shader_libraries_data)  //
   );
 
@@ -365,7 +383,7 @@ void ContextVK::Setup(Settings settings) {
   instance_ = std::move(instance.value);
   debug_report_ = std::move(debug_report);
   physical_device_ = physical_device.value();
-  device_ = std::move(device);
+  device.release();  // Stop automatic reset of device_ ivar.
   allocator_ = std::move(allocator);
   shader_library_ = std::move(shader_library);
   sampler_library_ = std::move(sampler_library);
--- a/impeller/renderer/backend/vulkan/context_vk.cc
+++ b/impeller/renderer/backend/vulkan/context_vk.cc
@@ -281,8 +281,7 @@ void ContextVK::Setup(Settings settings) {
   device_ = std::move(device_result.value);
   // This makes sure that the device is deleted at the proper time if there is
   // an error.
-  std::unique_ptr<vk::UniqueDevice, std::function<void(vk::UniqueDevice*)>>
-      device_resetter(&device_, [](vk::UniqueDevice* ptr) { ptr->reset(); });
+  fml::ScopedCleanupClosure device_resetter([this]() { device_.reset(); });
 
   if (!caps->SetDevice(physical_device.value())) {
     VALIDATION_LOG << "Capabilities could not be updated.";
@@ -368,7 +367,7 @@ void ContextVK::Setup(Settings settings) {
   instance_ = std::move(instance.value);
   debug_report_ = std::move(debug_report);
   physical_device_ = physical_device.value();
-  device_resetter.release();
+  device_resetter.Release();
   allocator_ = std::move(allocator);
   shader_library_ = std::move(shader_library);
   sampler_library_ = std::move(sampler_library);
--- a/impeller/renderer/backend/vulkan/context_vk.cc
+++ b/impeller/renderer/backend/vulkan/context_vk.cc
@@ -279,6 +279,8 @@ void ContextVK::Setup(Settings settings) {
     return;
   }
   device_ = std::move(device_result.value);
+  // This makes sure that the device is deleted at the proper time if there is
+  // an error.
   std::unique_ptr<vk::UniqueDevice, std::function<void(vk::UniqueDevice*)>>
       device_resetter(&device_, [](vk::UniqueDevice* ptr) { ptr->reset(); });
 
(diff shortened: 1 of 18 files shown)


jason-simmons:
`ContextVK::Setup` is constructing all of the context objects locally and moving them into the instance's members only if setup is successful.

But I'm wondering whether that is necessary given that `Setup` is private and is only called by `ContextVK::Create`.  `Setup` could instead initialize each member directly and ensure that the members used by `DeviceHolder` are populated before a `weak_ptr<DeviceHolder>` is given to any other object.

If `Setup` fails, then `Create` will discard the incomplete `ContextVK` and any members that were initialized will be destructed.

gaaclarke:
> If `Setup` fails, then `Create` will discard the incomplete `ContextVK` and any members that were initialized will be destructed.

I was just trying to preserve the old logic but I think you are right.  I'll get rid of that smart pointer that resets the instance variable before review.

gaaclarke:
I'm still running into trouble getting the tests to run on CI.  They run fine locally but crash with no warning on CI.  I'm trying turning on debug build runs of the tests too to see if they give more information.  I'm noticing that the bots are x86, so maybe that is the x-factor that makes the tests crash.

impeller/renderer/backend/vulkan/context_vk.cc
```diff
@@ -278,7 +278,9 @@ void ContextVK::Setup(Settings settings) {
     VALIDATION_LOG << "Could not create logical device.";
     return;
   }
-  vk::UniqueDevice device = std::move(device_result.value);
+
+  // Make sure this ivar is set before handing out DeviceHolder references.
+  device_ = std::move(device_result.value);
```
jason-simmons, line 283: Set every `ContextVK` member as it is created in `Setup` instead of setting all of the members at the end of the method.

If `Setup` fails, then this will ensure that the `ContextVK` destructor deletes the objects in the proper order.  In particular, objects tied to the device should be deleted before the device, and the Vulkan instance should be deleted last.

impeller/renderer/backend/vulkan/pipeline_cache_vk.cc
```diff
@@ -105,7 +106,11 @@ PipelineCacheVK::PipelineCacheVK(std::shared_ptr<const Capabilities> caps,
   is_valid_ = !!cache_;
 }
 
-PipelineCacheVK::~PipelineCacheVK() = default;
+PipelineCacheVK::~PipelineCacheVK() {
+  if (!device_holder_.lock()) {
+    cache_.release();
```
jason-simmons, line 111: If `device_holder_.lock()` succeeds, then call `cache_.reset()` within that block.

This avoids a potential scenario where the Vulkan device is destroyed after `device_holder_.lock()` goes out of scope but before the members of `PipelineCacheVK` are destructed.

(This also applies to the `PipelineVK` and `ShaderFunctionVK` destructors)

impeller/renderer/backend/vulkan/shader_function_vk.h
```diff
@@ -25,8 +26,10 @@ class ShaderFunctionVK final
   friend class ShaderLibraryVK;
 
   vk::UniqueShaderModule module_;
+  std::weak_ptr<DeviceHolder> device_;
```
jason-simmons, line 29: Rename to `device_holder_` for consistency

impeller/renderer/backend/vulkan/pipeline_vk.cc
```diff
@@ -20,7 +20,14 @@ PipelineVK::PipelineVK(std::weak_ptr<PipelineLibrary> library,
   is_valid_ = pipeline_ && render_pass_ && layout_ && descriptor_set_layout_;
 }
 
-PipelineVK::~PipelineVK() = default;
+PipelineVK::~PipelineVK() {
+  if (!IsPipelineLive()) {
```
jason-simmons, line 24: Is there a reason to tie this to the lifetime of the `PipelineLibrary`?

AFAICT the constraint on calling these object deletion APIs is specific to Vulkan.  The APIs should be called if the Vulkan device is still valid and must not be called if it was destroyed.

Remove `IsPipelineLive` from the `Pipeline` superclass.  Add a `weak_ptr<DeviceHolder>` to `PipelineVK` and use that to check whether these objects can be deleted.

impeller/renderer/backend/vulkan/shader_function_vk.h
gaaclarke: done

impeller/renderer/backend/vulkan/pipeline_vk.cc
gaaclarke: It already existed and mostly represented the same thing.  It could have lead to problems though, done.

impeller/renderer/backend/vulkan/pipeline_cache_vk.cc
gaaclarke: Good call.  We have to do more than that though.  We have to hold onto the lock for the whole duration of the branch.  I made both changes.

impeller/renderer/backend/vulkan/context_vk.cc
gaaclarke: Refactor code that doesn't have to do with the change?  I'd rather do that as a followup PR if you want.

impeller/renderer/backend/vulkan/context_vk.cc
```diff
@@ -320,10 +321,10 @@ void ContextVK::Setup(Settings settings) {
   }
 
   auto sampler_library =
-      std::shared_ptr<SamplerLibraryVK>(new SamplerLibraryVK(device.get()));
+      std::shared_ptr<SamplerLibraryVK>(new SamplerLibraryVK(device_.get()));
```
gaaclarke, line 324: I didn't look into it but, I suspect this guy and the FenceWaiterVK may need device holders too.

impeller/renderer/backend/vulkan/context_vk.cc
jason-simmons: It's related to this change - in the current version of this PR if `Setup` exits early it will set `device_` in `ContextVK` but not any other members.

In particular, `instance_` is not set by `Setup` until the end.  So if a failure happens during `Setup`, the `instance` local to `Setup` will be destroyed but `device_` will not be destroyed until the `ContextVK` is destructed.

IIUC it's unsafe to call a Vulkan API like `vkDestroyDevice` if the instance used to create the device has already been destroyed.  So I think it would be safest to set all the `ContextVK` members as they are created and then rely on the `ContextVK` destructor to delete the context's Vulkan objects in the correct order.

impeller/renderer/pipeline.h
```diff
@@ -66,6 +66,8 @@ class Pipeline {
  protected:
   Pipeline(std::weak_ptr<PipelineLibrary> library, T desc);
 
+  bool IsPipelineLive() const { return static_cast<bool>(library_.lock()); }
```
jason-simmons, line 69: `IsPipelineLive` was added as part of this PR - remove it now that it's no longer being called

gaaclarke: done

impeller/renderer/backend/vulkan/context_vk.cc
gaaclarke: Ah, good observation.  That requires that the ivars of the context are in the correct order in the header which at least for the device and instance is the case.  It also means the that the order in which the ivars are set in `Setup` is meaningful.  If we space it out across this whole method, it would be easy to get wrong.  I'm back to liking the resetter on the stack that I had at first.  It makes it easy to verify the order and is the least invasive to what we already have.

I'm just seeing that you are out of office.  Let's go over our options when you get back.

gaaclarke: Thinking about it, if the order of the ivars is wrong, then the order on the stack would be wrong too.  So that doesn't solve the problem.  I don't think there is a way you can guarantee this code is safe to all failures.  Your proposal is better than what it is now though.  I'll make the change.

gaaclarke: Starting to implement it, I don't like it since on line X we initialize an instance variable and line X+100 we use it.  It isn't clear that that instance variable has been initialized at X+100.  By using values on the stack and initializing them when we create them we can't get them out of order.  Let's chat when you get back.

gaaclarke: I got rid of the special smart pointer and moved resetting the ivar to a `std::unique_ptr`.  I think you'll find that preferable.

> But I think the requirement that a constructor must initialize members before it uses them later is well understood.

It's not that it's not understood, it's impossible to verify when looking at the code without scrolling around because the method is so large.

> It will be more familiar to readers of the code than the scoped_reset class and the requirement to call release() on the scoped_reset at the right time.

The nice thing is that they don't have the `release`, the tests will fail.  The order of the `release` isn't super important as long as it remains in the same spot since the chances of errors in that block of code where all the ivars is getting set is practically zero (possibly zero, i didn't look up std::optional::value or std::string::string's signatures).

impeller/renderer/backend/vulkan/context_vk.cc
```diff
@@ -278,7 +278,11 @@ void ContextVK::Setup(Settings settings) {
     VALIDATION_LOG << "Could not create logical device.";
     return;
   }
-  vk::UniqueDevice device = std::move(device_result.value);
+  device_ = std::move(device_result.value);
+  // This makes sure that the device is deleted at the proper time if there is
+  // an error.
+  std::unique_ptr<vk::UniqueDevice, std::function<void(vk::UniqueDevice*)>>
```
jason-simmons, line 284: Use `fml::ScopedCleanupClosure` here

gaaclarke: done

jason-simmons (approved):

impeller/renderer/backend/vulkan/context_vk.cc
gaaclarke: Thanks Jason.  I have a followup draft PR that will expand the functionality of mock vulkan so that we can assert the behavior when there is an error at setup.

gaaclarke: I give up, I can't make Setup fail because in practice we `abort` if there is any vulkan error.  🤷

PR #42134, opened 2023-05-18
[web] Remove the JS API for url strategy

mdebbar:
Finally, we can remove this JS global function for customizing the url strategy.

Why I think we don't need to go through an official deprecation process:

1. It was initially made for internal use in Google3, and right now there are no references to it.
2. There's no public documentation of this JS function.
3. External users customize their url strategy through `flutter_web_plugins` which has been [migrated](https://github.com/flutter/flutter/pull/123443) already.


eyebrowsoffire (approved):
LGTM!

PR #42146, opened 2023-05-19
[Impeller] avoid creating multiple concurrent message loops for Andorid Vulkan

dnfield:
Fixes https://github.com/flutter/flutter/issues/127160

We should avoid creating multiple message loops that each have as many threads as processors on the machine.

### DIFF
--- a/impeller/renderer/backend/vulkan/context_vk.cc
+++ b/impeller/renderer/backend/vulkan/context_vk.cc
@@ -120,6 +120,12 @@ void ContextVK::Setup(Settings settings) {
     return;
   }
 
+  if (!settings.worker_task_runner) {
+    VALIDATION_LOG
+        << "Cannot set up a Vulkan context without a worker task runner.";
+  }
+  worker_task_runner_ = settings.worker_task_runner;
+
   auto& dispatcher = VULKAN_HPP_DEFAULT_DISPATCHER;
   dispatcher.init(settings.proc_address_callback);
 
--- a/impeller/renderer/backend/vulkan/context_vk.cc
+++ b/impeller/renderer/backend/vulkan/context_vk.cc
@@ -307,10 +307,11 @@ void ContextVK::Setup(Settings settings) {
   /// Setup the pipeline library.
   ///
   auto pipeline_library = std::shared_ptr<PipelineLibraryVK>(
-      new PipelineLibraryVK(weak_from_this(),                    //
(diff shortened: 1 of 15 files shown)


gaaclarke (commented):
Ah, this is the issue jonah was referring to in the other PR, eh?

This looks mostly good except I wouldn't muddy `DeviceHolder` and follow the pattern we had previously of the clients holding onto a shared pointer to the concurrent task runner.

impeller/renderer/backend/vulkan/device_holder.h
```diff
@@ -12,6 +13,8 @@ class DeviceHolder {
  public:
   virtual ~DeviceHolder() = default;
   virtual const vk::Device& GetDevice() const = 0;
+  virtual const std::shared_ptr<fml::ConcurrentTaskRunner>
+  GetConcurrentWorkerTaskRunner() const = 0;
```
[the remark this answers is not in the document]
dnfield: Fair enough, I'll remove this and still give a shared pointer ot the runner to the pipeline.

I wanted this to help with testing but I can still test it by exposing a getter on the context anyway.

impeller/renderer/backend/vulkan/device_holder.h
```diff
@@ -4,6 +4,7 @@
 
 #pragma once
 
+#include "flutter/fml/concurrent_message_loop.h"
```
gaaclarke, line 7: You can remove this now.

impeller/renderer/backend/vulkan/context_vk.cc
```diff
@@ -427,6 +434,11 @@ const vk::Device& ContextVK::GetDevice() const {
   return device_.get();
 }
 
+const std::shared_ptr<fml::ConcurrentTaskRunner>
+ContextVK::GetConcurrentWorkerTaskRunner() const {
+  return worker_task_runner_;
+}
+
```
gaaclarke, line 441: We don't need this anymore right?

dnfield: It's used in the test.

impeller/renderer/backend/vulkan/device_holder.h
dnfield: done

gaaclarke (approved):
lgtm!

PR #42160, opened 2023-05-19
[Impeller] remove final cmd buffer waitUntilScheduled on physical iOS

jonahwilliams:
From local testing, this doesn't seem to be necessary on physical iOS devices, with or without platform views.


jonahwilliams:
This came up while working on https://github.com/flutter/engine/pull/42028 . I observed that some of the saved time shifted into the waitUntilScheduled call here, reducing the impact of the concurrent encoding.

jonahwilliams:
Yeah no rush on this!

bdero (approved):
LGTM. Tried this out on the Wondrous map.

PR #42184, opened 2023-05-20
Implement Web Codecs for Skwasm

eyebrowsoffire:
This implements making images from web browser codecs.

This implements a JS support library as well, which allows both the main thread and the background thread to keep a registry of JS objects that can be referenced by a numeric ID and transferred between each other.


lib/web_ui/lib/src/engine/skwasm/skwasm_impl/codecs.dart
```diff
@@ -0,0 +1,33 @@
...
+
+  @override
+  ui.Image generateImageFromVideoFrame(VideoFrame frame) {
+    final int width = frame.codedWidth.toInt();
+    final int height = frame.codedHeight.toInt();
+    final SkwasmSurface surface = (renderer as SkwasmRenderer).surface;
+    final int videoFrameId = surface.acquireObjectId();
+    skwasmInstance.skwasmRegisterObject(videoFrameId.toJS, frame as JSAny);
+    skwasmInstance.skwasmTransferObjectToThread(videoFrameId.toJS, surface.threadId.toJS);
+    return SkwasmImage(imageCreateFromVideoFrame(
```
yjbanov, line 26: Is there a potential race between the transfer and SkImage initialization?

lib/web_ui/skwasm/library_skwasm_support.js
```diff
@@ -0,0 +1,81 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+mergeInto(LibraryManager.library, {
```
yjbanov, line 5: What sorcery is this? Is this Emscripten? Maybe leave a comment with some breadcrumbs?

lib/web_ui/skwasm/surface.cpp
```diff
@@ -2,330 +2,272 @@
...
+// Main thread only
+uint32_t Surface::renderPicture(SkPicture* picture) {
+  assert(emscripten_is_main_browser_thread());
+  uint32_t callbackId = ++_currentCallbackId;
+  picture->ref();
+  emscripten_dispatch_to_thread(_thread, EM_FUNC_SIG_VII,
+                                reinterpret_cast<void*>(fRenderPicture),
+                                nullptr, this, picture);
+
+  // After drawing to the surface, the browser implicitly flushed the drawing
```
yjbanov, line 51: nit: s/flushed/flushes/

lib/web_ui/lib/src/engine/skwasm/skwasm_impl/codecs.dart
eyebrowsoffire: No, there isn't, although I could see why you might think so. It is true that the `SkImage` is being created synchronously on the UI thread, but it's the image's *generator* that actually attempts to grab the video frame by its ID and convert it to a texture, and that generator is always run on the web worker right before the image is being either drawn to the actual graphics surface or being rasterized to pixels. Any call that would draw or rasterize the image is sent to the web worker via a postMessage, and the postMessage queue is serial, so those calls are guaranteed to happen *after* the postMessage that transfers the video frame.

lib/web_ui/skwasm/library_skwasm_support.js
eyebrowsoffire: Yes it's the emscripten JS library magic. This JS file is preprocessed by emscripten and folded into `skwasm.js`. I will leave some comments.

lib/web_ui/lib/src/engine/image_decoder.dart
```diff
@@ -0,0 +1,439 @@
...
+  }
+
+  /// The index of the frame that will be decoded on the next call of [getNextFrame];
+  int _nextFrameIndex = 0;
+
+  /// Creating a new decoder is expensive, so we cache the decoder for reuse.
+  ///
+  /// This decoder is closed and the field is nulled out after some time of
+  /// inactivity.
+  ImageDecoder? _cachedWebDecoder;
```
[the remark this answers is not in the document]
eyebrowsoffire: https://github.com/flutter/flutter/issues/127548

hterkelsen (approved):
LGTM

yjbanov (approved):
lgtm

PR #42192, opened 2023-05-20
[Impeller] Add UV compute shader.

jonahwilliams:
Adds a compute shader that computes the UV mapping for an arbitrary geometry. Use it with drawPoints since I didn't realize it was possible to use an image shader with this 😆 .

### DIFF
--- a/impeller/entity/geometry.cc
+++ b/impeller/entity/geometry.cc
@@ -10,6 +10,7 @@
 #include "impeller/entity/points.comp.h"
 #include "impeller/entity/position_color.vert.h"
 #include "impeller/entity/texture_fill.vert.h"
+#include "impeller/entity/uv.comp.h"
 #include "impeller/geometry/matrix.h"
 #include "impeller/geometry/path_builder.h"
 #include "impeller/renderer/command_buffer.h"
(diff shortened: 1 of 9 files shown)


impeller/entity/geometry.cc
```diff
@@ -936,25 +933,125 @@ GeometryResult PointFieldGeometry::GetPositionBufferCPU(
...
+
+  auto points_data =
+      host_buffer.Emplace(points_.data(), points_.size() * sizeof(Point),
+                          DefaultUniformAlignment());
+
+  DeviceBufferDescriptor buffer_desc;
+  buffer_desc.size = total * sizeof(Point);
+  buffer_desc.storage_mode = StorageMode::kDevicePrivate;
+
+  auto geometry_buffer = renderer.GetContext()
```
jonahwilliams, line 975: Small discovery, if I use buffers allocated from the compute passes transients buffer, then the second compute pass doesn't correctly synchronize on the completion on the first. I suspect metal is tracking buffer inputs/outputs to do automatic fences and the singular buffer defeats that detection.

impeller/entity/shaders/geometry/uv.comp
```diff
@@ -0,0 +1,60 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include <impeller/texture.glsl>
```
jonahwilliams, line 5: This isn't specific to draw points and can be used with any of our geometries.

dnfield (approved):
Flutter: The fastest way to draw points anywhere?

This makes me think I should start looking at getting compute going on Vulkan next.

LGTM

PR #42210, opened 2023-05-22
Skip and ignore boringssl/src/rust when looking for the licenses.

aam:
third_party/boringssl/src/rust/bssl-crypto/deny.toml coming in as part of https://dart-review.googlesource.com/c/sdk/+/304210 triggers 'potential license' license script error.

Rust sources of boringssl are not included into flutter, don't need to be scanned for licenses.

### DIFF
--- a/tools/licenses/lib/paths.dart
+++ b/tools/licenses/lib/paths.dart
@@ -58,6 +58,7 @@ final Set<String> skippedPaths = <String>{
   r'third_party/benchmark', // only used by tests
   r'third_party/boringssl/src/crypto/err/err_data_generate.go',
   r'third_party/boringssl/src/fuzz', // testing tools, not shipped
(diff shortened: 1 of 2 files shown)


aam:
This is to unblock dart->engine roller

tools/licenses/lib/paths.dart
```diff
@@ -370,6 +370,7 @@ final Set<String> skippedCommonFiles = <String>{
   r'codereview.settings',
   r'coderules.txt',
   r'configure-ac-style.md',
+  r'deny.toml',
```
zanderso, line 373: I think we should probably just skip everything under the rust bindings directory under https://boringssl.googlesource.com/boringssl/+/refs/heads/master/rust since we don't build/ship them. That exclusion would go above next to the other paths we exclude from boringssl.

PR #42211, opened 2023-05-22
[Impeller] Polyline decomp.

jonahwilliams:
Compute shader that turns a line segment + quadradic into a polyline, tracking countour and segment end for stroke generation.


impeller/entity/shaders/geometry/polyline.comp
```diff
@@ -0,0 +1,103 @@
...
+struct PolylineComponent {
+  vec2 pt;
+  // This is the end of a countour.
+  //
+  // that is path.close was called with this as the final
+  // point.
+  bool contour_end;
+
+  // This point is the end of a line segment.
+  bool segment_end;
```
jonahwilliams, line 39: We need to track both pieces of this to determine where to insert stroke joins and end caps.

I think that means we also need to track segment start too? I could re-arrange this to be a bit field instead if we need to pack more in.

PR #42161, opened 2023-05-19
[Impeller] use host image upload path on simulator

jonahwilliams:
This might help with https://github.com/flutter/flutter/issues/126768#issuecomment-1554469333

PR #42212, opened 2023-05-22
Revert "[Impeller] use host image upload path on simulator"

jonahwilliams:
Reverts flutter/engine#42161

As reported by uses, this did not help with the issue.


dnfield (approved):
LGTM.

We can leave this around for a little longer to see if anyone says it helps with https://github.com/flutter/flutter/issues/126768

PR #42215, opened 2023-05-22
[Impeller] Removed last ivars to raw vk::Device's.

gaaclarke:
Removes the last ivars that hold onto a `Device` directly.  This have recently been the cause of crashes, so the safer thing is to remove them all.

For the Allocator and the Sampler library, removing the device ivars is likely just a precaution since they should have only been accessed from the raster thread and killed before the Device.  The fence waiter however was potentially accessing the device from a separate thread, so that was much more risky and likely to cause crashes.

Testing: The existing `impeller_unittests` verify this is setup correctly.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/impeller/renderer/backend/vulkan/context_vk.h
+++ b/impeller/renderer/backend/vulkan/context_vk.h
@@ -33,7 +33,6 @@ class FenceWaiterVK;
 
 class ContextVK final : public Context,
                         public BackendCast<ContextVK, Context>,
-                        public DeviceHolder,
                         public std::enable_shared_from_this<ContextVK> {
  public:
   struct Settings {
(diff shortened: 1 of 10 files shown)


impeller/renderer/backend/vulkan/context_vk.h
```diff
@@ -131,10 +133,16 @@ class ContextVK final : public Context,
   std::shared_ptr<FenceWaiterVK> GetFenceWaiter() const;
 
  private:
-  vk::UniqueInstance instance_;
+  struct UniqueDeviceHolder : public DeviceHolder {
```
jason-simmons, line 136: Avoid the "Unique" name for something that does not provide an interface similar to `std::unique_ptr`.  Call this something like `DeviceHolderImpl`.

gaaclarke: done

jason-simmons (approved):

PR #42219, opened 2023-05-22
Make FML_LOG safe from static initialization

dnfield:
I ran into this while trying to get some printing going for places where we're creating thread local keys. 

Supposedly, just including `<iostream>` should statically initialize `std::cout/cerr`, but it gets really hard to reason about whether your statically initialized code is going to be initialized before or after that happens. I tried making sure that the TU for `fml/logging.cc` did that initialization statically, but that also failed in the verison of the test included here (it passed in some other iterations that modified run_all_unittests.cc). We _could_ make sure it happens each and every time we touch `std::cerr` but ... we could also just use `fprintf(stderr, ...)` and it works just fine.

/cc @flar who ran into problems around this a little while back and was asking about it.

### DIFF
--- a/fml/logging.cc
+++ b/fml/logging.cc
@@ -123,8 +123,9 @@ LogMessage::~LogMessage() {
   fx_logger_log_with_source(fx_log_get_logger(), fx_severity, nullptr, file_,
                             line_, stream_.str().c_str());
 #else
-  std::cerr << stream_.str();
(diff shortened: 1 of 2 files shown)


chinmaygarde (approved):

dnfield:
Fuchsia isn't happy with the `::signal` call, just skipping the test for that platform.

PR #42233, opened 2023-05-23
Build arm64 FlutterEmbedder.framework and lipo the result

zanderso:
This PR modifies the mac_host_engine build config to build the FlutterEmbedder.framework for arm64, and adds a generator that creates a fat framework containing both the intel and arm64 dylibs. The fat framework is uploaded to the gcs bucket instead of the thin intel-only one.

The script to create the lipo'd embedder framework is different enough from the desktop framework to have a separate script, but this still triggers the pylint duplicate-code check. Due to a bug in our version of pylint (https://github.com/pylint-dev/pylint/issues/214), it isn't possible to disable per-file, so this PR also disables the check globally.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -38,6 +38,7 @@ docstring-quote = double
 # multiple time (only on the command line, not in the configuration file where
 # it should appear only once).
 disable=
+    duplicate-code,
     exec-used,
     fixme,
     missing-class-docstring,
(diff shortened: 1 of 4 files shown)


zanderso:
cc @jiahaog after this lands could you try it out with https://github.com/jiahaog/flt ?

jiahaog:
Thanks Zach, I was able to use this in https://github.com/jiahaog/flt/commit/d8dd1e1690f0ed884afa54fdd8628c9bf3d4503c. I don't have an arm mac to verify though, can you give it a shot?

PR #41977, opened 2023-05-12
[macOS - TextInput] Insert new line only when TextInputAction.newline

bleroux:
## Description

This PR updates the macOS text input plugin to avoid adding a new line on a multiline text field when action is not set to `TextInputAction.newline`.

## Related Issue

macOS implementation for https://github.com/flutter/flutter/issues/125879.
(similar to the Linux implementation in https://github.com/flutter/engine/pull/41895).

## Tests

Adds 2 tests.

PR #42244, opened 2023-05-23
[Windows - TextInput] Insert new line only when TextInputAction.newline

bleroux:
## Description

This PR updates the Windows text input plugin to avoid adding a new line on a multiline text field when action is not set to `TextInputAction.newline`.

## Related Issue

Fixes https://github.com/flutter/flutter/issues/125879 as Linux and macOS implementations are merged.
Linux PR: https://github.com/flutter/engine/pull/41895
macOS PR: https://github.com/flutter/engine/pull/41977

## Tests

Adds 2 tests.

### DIFF
--- a/shell/platform/windows/text_input_plugin_unittest.cc
+++ b/shell/platform/windows/text_input_plugin_unittest.cc
@@ -208,14 +208,14 @@ TEST(TextInputPluginTest, VerifyComposingSendStateUpdate) {
 }
 
 TEST(TextInputPluginTest, VerifyInputActionNewlineInsertNewLine) {
-  std::vector<std::vector<uint8_t>> messages;
+  // Store messages as std::string for convenience.
+  std::vector<std::string> messages;
 
   TestBinaryMessenger messenger(
       [&messages](const std::string& channel, const uint8_t* message,
                   size_t message_size, BinaryReply reply) {
-        int length = static_cast<int>(message_size);
-        std::vector<uint8_t> last_message(length);
-        memcpy(&last_message[0], &message[0], length * sizeof(uint8_t));
+        std::string last_message(reinterpret_cast<const char*>(message),
+                                 message_size);
         messages.push_back(last_message);
       });
   BinaryReply reply_handler = [](const uint8_t* reply, size_t reply_size) {};
--- a/shell/platform/windows/text_input_plugin_unittest.cc
+++ b/shell/platform/windows/text_input_plugin_unittest.cc
@@ -27,6 +27,14 @@ static constexpr char kTextPlainFormat[] = "text/plain";
 static constexpr char kChannelName[] = "flutter/textinput";
 static constexpr char kEnableDeltaModel[] = "enableDeltaModel";
 static constexpr char kSetClientMethod[] = "TextInput.setClient";
+static constexpr char kAffinityDownstream[] = "TextAffinity.downstream";
+static constexpr char kTextKey[] = "text";
+static constexpr char kSelectionBaseKey[] = "selectionBase";
+static constexpr char kSelectionExtentKey[] = "selectionExtent";
+static constexpr char kSelectionAffinityKey[] = "selectionAffinity";
+static constexpr char kSelectionIsDirectionalKey[] = "selectionIsDirectional";
+static constexpr char kComposingBaseKey[] = "composingBase";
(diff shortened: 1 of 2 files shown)


shell/platform/windows/text_input_plugin_unittest.cc
```diff
@@ -36,6 +44,56 @@ static std::unique_ptr<std::vector<uint8_t>> CreateResponse(bool handled) {
   return JsonMessageCodec::GetInstance().EncodeMessage(*response_doc);
 }
 
+static std::unique_ptr<rapidjson::Document> EncodedClientConfig(
+    std::string type_name,
+    std::string input_action) {
+  auto arguments = std::make_unique<rapidjson::Document>(rapidjson::kArrayType);
+  auto& allocator = arguments->GetAllocator();
+  arguments->PushBack(42, allocator);
```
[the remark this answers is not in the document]
bleroux: This is the client ID. I duplicated this from existing code (see lines 173, 254 and 301).
I can add an inline comment such as `arguments->PushBack(42, allocator); // Client id.`or a constant such as `kDefaultClientId`, which one do you prefer?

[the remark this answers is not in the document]
yaakovschectman: I am inclined to prefer using a named constant, but either approach sounds good to me.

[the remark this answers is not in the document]
loic-sharma: +1 for named constant here

shell/platform/windows/text_input_plugin_unittest.cc
```diff
@@ -148,6 +206,109 @@ TEST(TextInputPluginTest, VerifyComposingSendStateUpdate) {
...
+TEST(TextInputPluginTest, VerifyInputActionNewlineInsertNewLine) {
+  std::vector<std::vector<uint8_t>> messages;
+
+  TestBinaryMessenger messenger(
+      [&messages](const std::string& channel, const uint8_t* message,
+                  size_t message_size, BinaryReply reply) {
+        int length = static_cast<int>(message_size);
+        std::vector<uint8_t> last_message(length);
+        memcpy(&last_message[0], &message[0], length * sizeof(uint8_t));
+        messages.push_back(last_message);
```
loic-sharma, line 218: What do you think of storing `std::string`s instead?

```suggestion
        std::string last_message(reinterpret_cast<const char*>(message), message_size);
        messages.push_back(last_message);
```

yaakovschectman: Would this still allow use of `std::equal` on https://github.com/flutter/engine/pull/42244/files/16ea94095b930f2c0a85e244b43abf31dbe690f0#diff-f913efa93fd3ac5c33b9fc924f41ff6e1a7f8300a5e548aa52fb0d30a3904b4cR247, as `update_state_message` will still be the same type?

bleroux: > What do you think of storing std::strings instead?

I choosed to rely on `std::vector<uint8_t>` because this is the type that is returned by `codec.EncodeMethodCall`, I think it is more understandable to stick to the type expectations of the existing APIs used in the test.

yaakovschectman (approved):
LGTM as long as @loic-sharma is good with the use of vectors as you've been discussing

shell/platform/windows/text_input_plugin_unittest.cc
loic-sharma: Storing `std::string`s would make debugging a bit easier if things break.

If you still prefer storing `uint8_t`s, could you replace the `memcpy`'s `&last_message[0]` argument with  `last_message.data()`?

loic-sharma: Personally for line 248 I would just have the expected JSON stored as a string. Here's an example where we do that:

https://github.com/flutter/engine/blob/7c0a639efd0a0f6c9e260da739a80089be75e17b/shell/platform/fuchsia/flutter/text_delegate_unittests.cc#L228-L231

loic-sharma (approved):
Consider using a `std::string` for the expected JSON. See this comment: https://github.com/flutter/engine/pull/42244#discussion_r1213450327

Otherwise, LGTM! Thank you for the patch! :)

PR #42252, opened 2023-05-23
[web] Hide JS types from dart:ui_web

mdebbar:
In order to use the exported `PlatformLocation` cross-platform, we need to remove any JS types from the interface. Namely, `DomEventListener`.

Required by https://github.com/flutter/flutter/pull/126851


lib/web_ui/lib/ui_web/src/ui_web/navigation/platform_location.dart
```diff
@@ -81,13 +84,13 @@ class BrowserPlatformLocation implements PlatformLocation {
...
-    domWindow.addEventListener('popstate', fn);
+  void addPopStateListener(EventListener fn) {
+    domWindow.addEventListener('popstate', createDomEventListener(fn));
   }
 
   @override
-  void removePopStateListener(DomEventListener fn) {
-    domWindow.removeEventListener('popstate', fn);
+  void removePopStateListener(EventListener fn) {
+    domWindow.removeEventListener('popstate', createDomEventListener(fn));
```
mdebbar, line 93: I just realized that this may not work because it creates a new function that's different from the one passed to `domWindow.addEventListener()`.

Does `Function.toJS` create a new JS function on each invocation? Or does it cache and return the same JS function? cc @joshualitt @eyebrowsoffire

joshualitt: I ran into this too when I tried to migrate the plugins. `Function.toJS` creates a new function.

My 'fix' was to create a cache locally in this file to do the deduping. Another option is to change this API to take a `JSFunction`.

mdebbar: This API used to take a `JSFunction` but I had to change it because we don't want to expose a JS type in the interface of the `PlatformLocation` class.

I think it's reasonable to have a cache inside the `BrowserPlatformLocation` class.

lib/web_ui/lib/ui_web/src/ui_web/navigation/platform_location.dart
```diff
@@ -80,14 +87,21 @@ class BrowserPlatformLocation implements PlatformLocation {
...
-    domWindow.addEventListener('popstate', fn);
+  void addPopStateListener(EventListener fn) {
+    domWindow.addEventListener('popstate', getOrCreateDomEventListener(fn));
   }
 
   @override
-  void removePopStateListener(DomEventListener fn) {
-    domWindow.removeEventListener('popstate', fn);
+  void removePopStateListener(EventListener fn) {
+    final DomEventListener jsFn = getOrCreateDomEventListener(fn);
```
joshualitt, line 102: Do we want to throw if a user tries to remove a listener that was never added?

joshualitt (approved):

PR #42258, opened 2023-05-23
[web] Update a11y announcements to append divs instead of setting content.

marcianx:
This also removes the appended divs after a short time so that screen readers don't navigate to it, especially when users are entering the DOM to enable accessiblity.

Fixes https://github.com/flutter/flutter/issues/127335.

## Pre-launch Checklist

- [X] I signed the [CLA].

### DIFF
--- a/lib/web_ui/lib/src/engine/semantics/accessibility.dart
+++ b/lib/web_ui/lib/src/engine/semantics/accessibility.dart
@@ -53,6 +53,8 @@ void initializeAccessibilityAnnouncements() {
 
 /// Duration for which a live message will be present in the DOM for the screen
 /// reader to announce it.
+///
+/// This was determined by trial and error with some extra buffer added.
 const Duration liveMessageDuration = Duration(milliseconds: 300);
 
 /// Makes accessibility announcements using `aria-live` DOM elements.
--- a/lib/web_ui/lib/src/engine/semantics/accessibility.dart
+++ b/lib/web_ui/lib/src/engine/semantics/accessibility.dart
@@ -55,7 +55,12 @@ void initializeAccessibilityAnnouncements() {
 /// reader to announce it.
 ///
 /// This was determined by trial and error with some extra buffer added.
-const Duration liveMessageDuration = Duration(milliseconds: 300);
+Duration liveMessageDuration = const Duration(milliseconds: 300);
+
+/// Sets [liveMessageDuration] to reduce the delay in tests.
+void setLiveMessageDurationForTest(Duration duration) {
+  liveMessageDuration = duration;
+}
 
 /// Makes accessibility announcements using `aria-live` DOM elements.
 class AccessibilityAnnouncements {
--- a/lib/web_ui/lib/src/engine/semantics/accessibility.dart
+++ b/lib/web_ui/lib/src/engine/semantics/accessibility.dart
@@ -2,6 +2,7 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+import 'dart:async';
 import 'dart:typed_data';
 
 import '../../engine.dart' show registerHotRestartListener;
@@ -23,7 +24,7 @@ enum Assertiveness {
 AccessibilityAnnouncements get accessibilityAnnouncements {
   assert(
     _accessibilityAnnouncements != null,
-    'AccessibilityAnnouncements not initialized. Call initializeAccessibilityAnnouncements() to innitialize it.',
+    'AccessibilityAnnouncements not initialized. Call initializeAccessibilityAnnouncements() to initialize it.',
   );
   return _accessibilityAnnouncements!;
 }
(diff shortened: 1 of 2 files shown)


lib/web_ui/lib/src/engine/semantics/accessibility.dart
```diff
@@ -50,6 +51,10 @@ void initializeAccessibilityAnnouncements() {
   });
 }
 
+/// Duration for which a live message will be present in the DOM for the screen
+/// reader to announce it.
+const Duration liveMessageDuration = Duration(milliseconds: 300);
```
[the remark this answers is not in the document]
marcianx: Done

yjbanov, line 58: Instead of making this a `const` can this be mutable so that tests can set a smaller duration than 300ms? I don't know how many tests we currently have, but if the union of all tests have to await on this 10 times, it lengthens the runtime of the test by 3 seconds. I think we could use 10 times smaller delays in the tests.

marcianx: Done!

yjbanov (approved):
lgtm

lib/web_ui/test/engine/semantics/accessibility_test.dart
```diff
@@ -43,76 +55,89 @@ void testMain() {
...
+      expectNoMessages();
     });
 
-    test('aria-live is polite when assertiveness is set to 0', () {
-      resetAccessibilityAnnouncements();
-      const Map<dynamic, dynamic> testInput = <dynamic, dynamic>{'data': <dynamic, dynamic>{'message': 'polite message', 'assertiveness': 0}};
-      accessibilityAnnouncements.handleMessage(codec, codec.encodeMessage(testInput));
-      expect(accessibilityAnnouncements.ariaLiveElementFor(Assertiveness.polite).text, 'polite message');
-      expect(accessibilityAnnouncements.ariaLiveElementFor(Assertiveness.assertive).text, '');
+    test('aria-live is polite when assertiveness is set to 0', () async {
```
htoor3, line 104: Do we know/care what happens if assertiveness is a negative int? Or is this not a scenario we'll run into

lib/web_ui/test/engine/semantics/accessibility_test.dart
```diff
@@ -43,76 +55,89 @@ void testMain() {
...
-      accessibilityAnnouncements.handleMessage(codec, codec.encodeMessage(testInput2));
-      expect(accessibilityAnnouncements.ariaLiveElementFor(Assertiveness.polite).text, 'Hello.');
-      expect(accessibilityAnnouncements.ariaLiveElementFor(Assertiveness.assertive).text, '');
-
-      // Now the "." is removed because the message without it will also look updated.
-      const Map<dynamic, dynamic> testInput3 = <dynamic, dynamic>{'data': <dynamic, dynamic>{'message': 'Hello'}};
-      accessibilityAnnouncements.handleMessage(codec, codec.encodeMessage(testInput3));
-      expect(accessibilityAnnouncements.ariaLiveElementFor(Assertiveness.polite).text, 'Hello');
-      expect(accessibilityAnnouncements.ariaLiveElementFor(Assertiveness.assertive).text, '');
+    test('Rapid-fire messages are each announced.', () async {
```
htoor3, line 112: Does this behavior change if mixed assertive/polite rapid-fire messages are sent, rather than just polite?

htoor3 (approved):
Just left some minor test comments, but lgtm otherwise

lib/web_ui/test/engine/semantics/accessibility_test.dart
marcianx: Nope. Nothing fundamentally changes. They go into different live divs, each of which is handled independently. These tests ensure that multiple messages of the same type don't clobber each other in any way.

mdebbar (approved):

lib/web_ui/test/engine/semantics/accessibility_test.dart
```diff
@@ -18,20 +21,29 @@ void main() {
...
+      present ? isNotNull : isNull,
+    );
+  }
+
+  tearDown(() async {
+    // Completely reset accessibility announcements for subsequent tests.
+    accessibilityAnnouncements.dispose();
+    await Future<void>.delayed(liveMessageDuration * 2);
+    initializeAccessibilityAnnouncements();
+    expectAnnouncementElements(present: true);
```
ditman, line 43: I find it slighly weird to "initializeX" inside of a `tearDown` method.

Would it make sense that `initializeAccessibilityAnnouncements` and (maybe) `expectAnnouncementElements` are the implementation of a `setUp` method?

I suspect this being out of order makes the tests in this file dependent on the execution order? (what happens if you don't run the "Initialization and disposal" test?)

ditman: Aggg too late :/

marcianx: Thanks for checking! Every test assumes that it's initialized at the start because that's what initializing the engine does on `setUpAll`. This is the same for the initialization and disposal test as well. So there should be no order dependence.

ditman: @marcianx yes! thanks for the response!

PR #42264, opened 2023-05-23
[Impeller] Set the coverage hint on everything

bdero:
Attaches a minimum coverage hint to all rendered entities. This makes ImageFilter blurs do less work when clipped. We can utilize this in a number of other filters as well.

Screenshot 2023-05-23 at 3 16 49 PM

The test I added also demonstrates a sampling issue in the gaussian blur that I'm trying to work out. Here it is without any clipping. This coverage hint change does not introduce the issue:

Screenshot 2023-05-23 at 3 08 15 PM

### DIFF
--- a/impeller/aiks/aiks_unittests.cc
+++ b/impeller/aiks/aiks_unittests.cc
@@ -2309,6 +2309,27 @@ TEST_P(AiksTest, CanRenderBackdropBlurHugeSigma) {
   ASSERT_TRUE(OpenPlaygroundHere(canvas.EndRecordingAsPicture()));
 }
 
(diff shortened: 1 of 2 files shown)


bdero:
In most Flutter apps, the 2-pass blur gets used for images and backdrops -- the vast majority of standalone blurs are SolidColor mask blurs on RRects, which we don't render using the 2-pass blur. I suspect that's why no one's reported this yet.

zanderso (approved):
@bdero please triage/accept the new golden images

chinmaygarde (approved):

PR #42265, opened 2023-05-23
[Impeller] Create an autorelease pool for Impeller tests running on macOS.

jason-simmons:
See https://github.com/flutter/flutter/issues/127358

### DIFF
--- a/fml/platform/darwin/scoped_nsautorelease_pool.cc
+++ b/fml/platform/darwin/scoped_nsautorelease_pool.cc
@@ -1,0 +1,26 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "flutter/fml/platform/darwin/scoped_nsautorelease_pool.h"
+
+#include <objc/message.h>
+#include <objc/runtime.h>
+
+namespace {
+typedef id (*msg_send)(void*, SEL);
+}  // anonymous namespace
+
+namespace fml {
+
+ScopedNSAutoreleasePool::ScopedNSAutoreleasePool() {
+  autorelease_pool_ = reinterpret_cast<msg_send>(objc_msgSend)(
+      objc_getClass("NSAutoreleasePool"), sel_getUid("new"));
+}
+
+ScopedNSAutoreleasePool::~ScopedNSAutoreleasePool() {
+  reinterpret_cast<msg_send>(objc_msgSend)(autorelease_pool_,
+                                           sel_getUid("drain"));
+}
+
+}  // namespace fml
--- a/fml/platform/darwin/scoped_nsautorelease_pool.h
+++ b/fml/platform/darwin/scoped_nsautorelease_pool.h
@@ -1,0 +1,26 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef FLUTTER_FML_PLATFORM_DARWIN_SCOPED_NSAUTORELEASEPOOL_H_
+#define FLUTTER_FML_PLATFORM_DARWIN_SCOPED_NSAUTORELEASEPOOL_H_
+
+#include "flutter/fml/macros.h"
+
+namespace fml {
+
+// Pushes an autorelease pool when constructed and pops it when destructed.
+class ScopedNSAutoreleasePool {
+ public:
+  ScopedNSAutoreleasePool();
+  ~ScopedNSAutoreleasePool();
+
+ private:
+  void* autorelease_pool_;
+
+  FML_DISALLOW_COPY_AND_ASSIGN(ScopedNSAutoreleasePool);
+};
+
+}  // namespace fml
+
+#endif  // FLUTTER_FML_PLATFORM_DARWIN_SCOPED_NSAUTORELEASEPOOL_H_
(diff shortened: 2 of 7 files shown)


fml/platform/darwin/scoped_nsautorelease_pool.h
```diff
@@ -0,0 +1,22 @@
...
+namespace fml {
+
+// Pushes an autorelease pool when constructed and pops it when destructed.
+class ScopedNSAutoreleasePool {
+ public:
+  ScopedNSAutoreleasePool();
+  ~ScopedNSAutoreleasePool();
+
+ private:
+  void* autorelease_pool_;
```
gaaclarke, line 17: add `FML_DISALLOW_COPY_AND_ASSIGN`

fml/platform/darwin/scoped_nsautorelease_pool.cc
```diff
@@ -0,0 +1,22 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "flutter/fml/platform/darwin/scoped_nsautorelease_pool.h"
+
+extern "C" {
+void* objc_autoreleasePoolPush(void);
```
gaaclarke, line 8: These are undocumented methods.  Can you just turn this to a .mm file and use the objc objects?

gaaclarke: Alternatively, just rename the playground test file to `golden_playground_test_mac.mm` and create and destroy the pool in the SetUp and TearDown (storing the pool in `GoldenPlaygroundTest::GoldenPlaygroundTestImpl`).

gaaclarke (changes requested):
It seems like we should have an autorelease pool drained after each test, instead of each test suite, no?

bdero:
I believe @dnfield already added an autorelease pool utility to the playground harness. Should we replace it with this new FML util? https://github.com/flutter/engine/pull/40748/files

dnfield (approved):
LGTM. Consider using this in https://github.com/flutter/engine/blob/main/impeller/playground/playground.cc#L265 as well. The feedback on that change was to move the class to FML.

LGTM

fml/platform/darwin/scoped_nsautorelease_pool.cc
dnfield: This needs to be importable into non-objc TUs See https://github.com/flutter/engine/pull/40748/files for example

gaaclarke: > This needs to be importable into non-objc TUs See https://github.com/flutter/engine/pull/40748/files for example

GoldenPlaygroundTest already handles this by using the pimpl pattern.  For `PlaygroundTest` it can just remain a `void*` which is no worse than what we have now.

jason-simmons:
> It seems like we should have an autorelease pool drained after each test, instead of each test suite, no?

The test fixture class containing the autorelease pool will be instantiated separately for each test case.

fml/platform/darwin/scoped_nsautorelease_pool.h
jason-simmons: done

jason-simmons:
> Consider using this in https://github.com/flutter/engine/blob/main/impeller/playground/playground.cc#L265 as well. The feedback on that change was to move the class to FML.

Replaced the autorelease wrapper in `Playground` with `fml::ScopedNSAutoreleasePool`

gaaclarke (commented):
> > It seems like we should have an autorelease pool drained after each test, instead of each test suite, no?
> 
> The test fixture class containing the autorelease pool will be instantiated separately for each test case.

You're right, I got mixed up with other testing libraries.  Everything LGTM modulo just using the documented API for autorelease pools.  Using undocumented functions is needlessly risky and complex imo.

fml/platform/darwin/scoped_nsautorelease_pool.cc
jason-simmons: This is based on Chromium's usage of these APIs in a similar class: see https://source.chromium.org/chromium/chromium/src/+/main:base/mac/scoped_nsautorelease_pool.cc

Relying on implementation details of the ObjC runtime is not ideal, but I don't think it will cause issues as long as it's hidden behind the `ScopedNSAutoreleasePool` abstraction.

gaaclarke: Okay, we keep the undocumented functions can you add this file to a `testonly` build target?  On iOS using undocumented functions can result in getting rejected from the App Store.  We want to make sure that it doesn't find it's way into the iOS code.

jason-simmons:
Changed to the implementation based on `objc_msgSend` that had been used in `Playground` - PTAL

gaaclarke (approved):
LGTM. I think this is needlessly convoluted in comparison to just using a `.mm` file but this will work.

gaaclarke:
> This is specifically needed for TUs where you can't just use a .mm file - if you could you would just do `@autoreleasepool {}` :)

You just compile the file you need depending on the platform.  You can't use `@autoreleasepool` because the code where the scope of pool isnt in objc++.  Is there something I'm missing?

## BUILD.gn
```python
executable("fml") {
  source = ["autorelease_pool.h"]
  if (is_mac || is_ios) {
    sources += ["autorelease_pool_darwin.mm"]
  } else {
    # Add a stub if you don't want to conditionally include the header.
    sources += ["autorelease_pool_stub.cc"]
  }
}
```

## autorelease_pool.h
```c++
class AutoreleasePool {
  private:
   void* pool_;
};
```

## autorelease_pool_darwin.mm
```c++
AutoreleasePool::AutoreleasePool() {
  pool_ = [[NSAutoreleasePool alloc] init];
}

AutoreleasePool::~AutoreleasePool() {
  [pool_ drain];
}
```

PR #42269, opened 2023-05-23
[Impeller] Limit rrect blur kernel to 1000x1000

bdero:
Impose the same large limit as the slower 2-pass blur.

Also renames "RRectShadow" to the much less ambiguous "SolidRRectBlur".

Before, a large enough blur sigma would result in nothing getting drawn. The Gaussian integral seems to be working fine against 16bit ops, so no adjustments were needed in the shader.

Screenshot 2023-05-23 at 4 08 41 PM


jonahwilliams (approved):
LGTM

chinmaygarde (approved):

PR #42275, opened 2023-05-24
[Impeller] Add golden repro for mask blur issues

bdero:
Part of https://github.com/flutter/flutter/issues/127013.

This is a repro to track the mask blur issues. If we accidentally trigger the wrong kind of blur, this golden will make it obvious because the gradient lines will become much more blurry.

Once this problem is fixed, the circle's mask should be fully blurred and the rounded rectangle should have no edge clamping (the diagonal lines should continue as the edges fade to black).

image


chinmaygarde (approved):
Trippy!

bdero:
With the sampling fix present, the playground looks like this:

Screenshot 2023-05-24 at 10 05 39 AM

PR #42277, opened 2023-05-24
[Impeller] Add 1 pixel border to snapshots of Contents

bdero:
Everything in Impeller is theoretically treated as if being drawn to layers of infinite size. Without this change, there's a coverage leak that occurs in some filters due to the customizable sampling mode. I should be able to get rid of this 1 pixel padding by refactoring the way we do sampling in the filter inputs (we already have most of this infrastructure in place), but this is a correct, safe, and not-that-expensive solution for the time being.

Also, this fixes a potential segfault due to an unsafe optional access... (would really love if there was a static analysis check for this -- it's by far the number one way I push crashers).

Before:

Screenshot 2023-05-23 at 10 51 43 PM

After:

Screenshot 2023-05-23 at 10 49 29 PM

### DIFF
--- a/impeller/entity/contents/contents.cc
+++ b/impeller/entity/contents/contents.cc
@@ -64,11 +64,18 @@ std::optional<Snapshot> Contents::RenderToSnapshot(
     bool msaa_enabled,
     const std::string& label) const {
   auto coverage = GetCoverage(entity);
-  if (coverage_limit.has_value()) {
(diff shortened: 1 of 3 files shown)


zanderso (approved):

zanderso:
It looks like https://clang.llvm.org/extra/clang-tidy/checks/bugprone/unchecked-optional-access.html is missing from our .clang-tidy file. I'll try adding it.

PR #42285, opened 2023-05-24
[Impeller] Allow batching compute workloads for geometry computations.

jonahwilliams:
Design:

The current design of the geometry system requires each geometry object to individually compute its own vertices to provide these synchronously to the rendering contents. We'd like to adjust this so that the geometry computations can be conditionally offloaded onto compute workloads, which should be significantly faster. 

A compute workload is going to be dispatched to a minmum number of threads, for example 1024 on macOS and iPhone 13. Some parallel algorithms used, like prefix sum, require barrier synchronization, meaning all threads must hit the same barriers and do some amount of work. Thus a compute workload operating over a 3 point polyline will perform about the same as a compute workload operating over a 300 point polyline. In addtion, there is some CPU side cost in creating and encoding compute commands. As a result of this, we don't want to map entity geometry objects 1:1 with compute commands. Instead we'd like to batch them up into as few compute commands as reasonable.

This PR introduces a system for batching up compute commands, which is called a GeometryPass. Each EntityPass will lazily contruct a GeometryPass object. This object will have methods that are specific to the type of geometry being computed, and will eagerly return IndirectCommandArguments which can be used by the rendering contents to encode commands whose vertex count we don't yet know. When the render pass is encoded, it will first enqueue and encode the commamd buffer, which insures all geometry data required by the render pass is ready before the vertex/fragment stages start.

In this PR, the indirect arguments are not yet used. Only the ConvexTessellation has been implemented, as this is the simplest algorithm for batching.

In performance testing locally, this doesn't perform noticably better or worse than the current CPU based convex tessellator. Large gains require polyline computation to also be moved to compute.  Note that batching commands requires a bit more state to be tracked. For convex tessellation, we need to know the start of the current piece of geometry. We'll likely need to encode more data for polyine computation, et cetera. This slightly increaases the costs of the shaders but without batching its not likely to be feasible to use compute except for extremely large geometry.

Other Adjustments:

* Grid size is moved to compute command instead of compute pass.


impeller/renderer/geometry_pass.cc
```diff
@@ -0,0 +1,94 @@
...
+
+GeometryPassResult GeometryPass::AddPolyline(Path::Polyline polyline,
+                                             const ContentContext& renderer) {
+  using CS = ConvexComputeShader;
+  auto result_size = polyline.points.size() * 3;
+  convex_pipeline_ = renderer.GetConvexComputePipeline();
+  auto& convex_command = GetOrCreateConvex(polyline.points.size());
+
+  // Emplace with no padding as we're going to treat this as a single buffer.
+  std::ignore = convex_command.input_buffer->Emplace(
```
jonahwilliams, line 46: This feels sort of hacky, but given that the HostBuffers already sort of act like temporary buffers I didn't want to create another local temporary buffer.

impeller/renderer/geometry_pass.cc
```diff
@@ -0,0 +1,94 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "impeller/renderer/geometry_pass.h"
+
+#include <iostream>
+#include <tuple>
+
+#include "impeller/core/host_buffer.h"
+#include "impeller/entity/contents/content_context.h"  // nogncheck
```
jonahwilliams, line 11: Sort of a hack, maybe move the geometry shaders to rendering?

impeller/renderer/backend/metal/compute_pass_mtl.mm
```diff
@@ -255,8 +255,10 @@ static bool Bind(ComputePassBindingsCache& pass,
 
     // Special case for linear processing.
     if (height == 1) {
-      int64_t threadGroups =
-          std::max(width / maxTotalThreadsPerThreadgroup, 1LL);
+      int64_t threadGroups = std::max(
+          static_cast<int64_t>(
```
jonahwilliams, line 259: Oops I still don't realize that C++ isn't Dart, subgroups were capped at 1 😆

jonahwilliams:
This has gotten a bit stale and I'm heads down on otherwork. Filled an issue to continue this here: https://github.com/flutter/flutter/issues/131304

PR #42293, opened 2023-05-24
[Impeller] fix UV and color blending.

jonahwilliams:
Fixes https://github.com/flutter/flutter/issues/127486

Combinations of UV mapping and color blending would cause us to double apply the UV mapping which rendered either nothing or garbage.


jonahwilliams:
Basically we'd end up applying the UV mapping from the VerticesGeometry twice, whereas the second time we really just want to sample the snapshotted rectangle.

I think we should update this implementation to avoid the snapshot for image sources, but that is a more involved change

bdero (approved):
LGTM

PR #42304, opened 2023-05-25
Apply the drawShadow bounds workaround to the Web HTML backend

jason-simmons:
Previously the Flutter framework had been inflating the rectangle computed for the bounds of a drawShadow operation in order to work around potential inaccuracies in the SkPicture's bounds calculation.

That workaround is now obsolete for most platforms and was removed from the framework (see https://github.com/flutter/flutter/pull/127052). But the Web HTML backend is using different code for computing shadow bounds.  This PR restores the workaround for Web HTML for consistency with the old behavior.

### DIFF
--- a/lib/web_ui/lib/src/engine/shadow.dart
+++ b/lib/web_ui/lib/src/engine/shadow.dart
@@ -73,12 +73,19 @@ ui.Rect computePenumbraBounds(ui.Rect shape, double elevation) {
   final double dx = elevation * tx;
   final double dy = elevation * ty;
   final ui.Offset offset = computeShadowOffset(elevation);
-  return ui.Rect.fromLTRB(
+  final ui.Rect bounds = ui.Rect.fromLTRB(
     shape.left - dx,
     shape.top - dy,
     shape.right + dx,
     shape.bottom + dy,
   ).shift(offset);
+
+  // Expand the bounds rectangle to compensate for inaccuracy in the shadow
+  // calculation.  This is similar to a workaround that had previously been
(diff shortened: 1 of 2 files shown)


eyebrowsoffire (approved):
LGTM!

lib/web_ui/lib/src/engine/shadow.dart
```diff
@@ -73,12 +73,18 @@ ui.Rect computePenumbraBounds(ui.Rect shape, double elevation) {
...
     shape.top - dy,
     shape.right + dx,
     shape.bottom + dy,
   ).shift(offset);
+
+  // Expand the bounds rectangle to compensate for inaccuracy in the shadow
+  // calculation.  This duplicates a workaround that had been previously used
+  // in the Flutter framework to adjust the bounds for shadows drawn by Skia.
+  // (See https://github.com/flutter/flutter/pull/127052)
+  return bounds.inflate(20);
```
flar, line 87: In the original workaround, the bounds computed was the join of the inflated shadow shape and the computed shadow bounds. It was basically enforcing a minimum padding of 20 pixels, but did not otherwise pad shadows larger than that.

Here you are returning an inflated value of the shadow bounds which pads all bounds even those that were originally larger than 20 pixels of shadow.

I'm not saying that is a bad thing, but it isn't the same thing...

jason-simmons: Changed the comment so it does not say that this workaround is identical to the old one

flar (approved):
LGTM

PR #42306, opened 2023-05-25
[Impeller] Simplify CreateContentsForGeometry overloads

bdero:
Came across this while working on generic fallbacks for mask blurs. For the unique_ptr case, we always capture ownership and convert to a shared_ptr anyhow. For the shared_ptr case, we always need to capture a copy, and so better to not take a const ref and allow the caller to move if possible.


jonahwilliams (approved):
LGTM

PR #42318, opened 2023-05-25
Implement Canvas.drawGlyphRun

xtyxtyx:
This PR implements the `drawGlyphRun` method for the `Canvas` class. This method offers a way to draw pre-laid out text, enhancing the text rendering capabilities and performance of Flutter:

Fixes: https://github.com/flutter/flutter/issues/96617 

Moreover, this enhancement serves as a prerequisite for resolving another issue: https://github.com/flutter/flutter/issues/58478.

```dart
final glyphRun = ui.GlyphRun(
  glyphs: Uint16List(...),
  positions: [Offset.zero,...],
  fontFamily: 'Menlo',
  fontSize: 14,
  fontWeight: FontWeight.w700,
  fontStyle: FontStyle.italic,
);

canvas.drawGlyphRun(glyphRun, Offset(5, 20), ui.Paint()..color = Colors.red);
```
CleanShot 2023-05-25 at 22 58 58@2x

> This PR is based on prior work from https://github.com/NanoMichael/engine/pull/1 .


knopp:
This looks great. Though I'm wondering if it could be approximated somehow with HTML renderer.

xtyxtyx:
> This looks great. Though I'm wondering if it could be approximated somehow with HTML renderer.

It appears that currently there is no API available for displaying glyphs in HTML, including SVG. Skia has proposed [an addition to Canvas2D](https://skia.org/docs/dev/design/text_c2d/), suggesting the use of 'context.fillGlyphs(glyphs, positions, Font)' to draw glyphs. However, it is not implemented in any browser yet..

Liloupar:
Can this be merged?

lib/ui/painting/canvas.cc
```diff
@@ -639,6 +640,77 @@ void Canvas::drawShadow(const CanvasPath* path,
...
+                               .GetFontCollection();
+
+    std::vector<SkString> sk_font_families;
+    sk_font_families.push_back(SkString(font_family_name.c_str()));
+
+    auto sk_typefaces =
+        font_collection->CreateSktFontCollection()->findTypefaces(
+            sk_font_families, font_style);
+
+    if (sk_typefaces.empty()) {
```
dnfield, line 687: @jason-simmons or @Rusino - will this be the case if the exact family/size/weight/etc. cannot be found, or will Skia potentially fall back here to some other font?

There also probably needs to be some kind of assertion that you've found only one font right? This is important so that the glyph IDs will actually match what you expected.

lib/ui/painting.dart
```diff
@@ -4936,6 +4936,59 @@ enum ClipOp {
...
+    required this.glyphs,
+    required this.positions,
+    required this.fontFamily,
+    required this.fontSize,
+    this.fontWeight = FontWeight.normal,
+    this.fontStyle = FontStyle.normal,
+  }) : assert(positions.length == glyphs.length * 2);
+
+  /// The id of the glyphs to draw. If a glyph id in this list does not exist
+  /// in the font given by [familyName], it takes no visual effect.
```
dnfield, line 4970: Here and below: style guide says that the first line of a doc comment should be a single-line sentence. 

We should document here how to get a glyph ID, what a glyph ID is, and importantly that it may be different for different fonts or even different versions of a font.

lib/ui/painting.dart
```diff
@@ -5756,6 +5809,14 @@ abstract class Canvas {
   ///
   /// The arguments must not be null.
   void drawShadow(Path path, Color color, double elevation, bool transparentOccluder);
+
+  /// Draw a list of glyphs given by [run]. All glyphs are baseline aligned.
+  /// [offset] specifies the origin point to draw the glyphs, so the position of
```
dnfield, line 5814: `The [offset] parameter` (start sentences with a capitcal letter).

lib/ui/painting/canvas.cc
jason-simmons: SkParagraph's `FontCollection::findTypefaces` may return some other font.  For example, if none of the available font managers can find a match for the specified font family, then `findTypefaces` may return the platform's default font.

I'd be reluctant to expose a public API that uses glyph IDs without a more complete set of low-level font management APIs similar to Skia's `SkFontMgr`/`SkTypeface`/`SkFont`.  These APIs would need to offer more control over font selection and a way to map Unicode code points to glyph IDs.

dnfield: I agree. If you're drawing with glyph IDs it will be very important to know you're getting the font you expect. 

For others reading this who aren't familiar, we get a number of bugs from users related to unexpected fallbacks/mismatched versions of fonts as it is. Sometimes a particular font has a bug and does not render as expected. It tends to be difficult today to track these down and understand them. That problem would be significantly worse with an API like this. If you're supplying glyph IDs based on the font on your local development environment, and your user has a different font installed, you suddenly will render either nothing or possibly _completely different text_ (because locally the glyph ID for `A` is now `B` on the installed font).

Hixie:
How does this interact with shaping, e.g. of Arabic text?

tgucio:
My understanding is this uses font GlyphIds i.e. what text rendering ends up with after font matching and shaping (e.g. GSUB and GPOS tables in OpenType fonts plus many script-specific tricks). For this to work we'd need to know the GlyphIds in the font we point to (`fontFamily: 'Menlo'`) in the example). The issues I see here are:
- there's no guarantee the font actually exists in the platform the app runs on (in which case a fallback font would be used and we'd get random output
- there's no guarantee the font version is the same the developer used: GlyphIds may change
- little usability for complex scripts (Indic, SE Asian, to less extent Arabic, Hebrew and many more)

I guess the way to properly implement this would be to:
- add an API in the engine to get glyph run info for paragraphs after layout (font name/ID)
- use this and alter given bits such as colour

Hixie:
That's consistent with my impression of this API. I understand the core use case, but it seems like this is a very brittle solution to that use case.

PR #42324, opened 2023-05-25
[Skwasm] Implement miscellaneous drawing APIs

eyebrowsoffire:
Implements `drawVertices`, `drawPoints`, and `drawAtlas`.

These are the last rendering APIs that are unimplemented in Skwasm! (Although we still need to add platform view support).


yjbanov (approved):
lgtm

PR #42328, opened 2023-05-25
Proper memory management in Skwasm

eyebrowsoffire:
This fixes https://github.com/flutter/flutter/issues/127243.

This ensures that native skwasm objects are cleaned up when their associated dart-side objects are garbage collected.

### DIFF
--- a/lib/web_ui/lib/src/engine/skwasm/skwasm_impl/canvas.dart
+++ b/lib/web_ui/lib/src/engine/skwasm/skwasm_impl/canvas.dart
@@ -18,9 +18,9 @@ class SkwasmCanvas implements SceneCanvas {
   SkwasmCanvas.fromHandle(this._handle);
   CanvasHandle _handle;
 
-  void delete() {
-    canvasDestroy(_handle);
-  }
+  // Note that we do not need to deal with the finalizer registry here, because
+  // the underlying native skia object is tied directly to the lifetime of the 
+  // associated SkPictureRecorder.
 
   @override
   void save() {
--- a/lib/web_ui/lib/src/engine/skwasm/skwasm_impl/canvas.dart
+++ b/lib/web_ui/lib/src/engine/skwasm/skwasm_impl/canvas.dart
@@ -19,7 +19,7 @@ class SkwasmCanvas implements SceneCanvas {
   CanvasHandle _handle;
 
   // Note that we do not need to deal with the finalizer registry here, because
-  // the underlying native skia object is tied directly to the lifetime of the 
+  // the underlying native skia object is tied directly to the lifetime of the
   // associated SkPictureRecorder.
 
   @override
--- a/lib/web_ui/lib/src/engine/skwasm/skwasm_impl/filters.dart
+++ b/lib/web_ui/lib/src/engine/skwasm/skwasm_impl/filters.dart
@@ -9,11 +9,11 @@ import 'package:ui/src/engine.dart';
 import 'package:ui/src/engine/skwasm/skwasm_impl.dart';
 import 'package:ui/ui.dart' as ui;
 
-class SkwasmImageFilter implements SceneImageFilter {
+class SkwasmImageFilter implements SceneImageFilter, SkwasmObjectWrapper<RawImageFilter> {
   SkwasmImageFilter._(this.handle) {
-    _registry.register(this, handle.address, this);
+    _registry.register(this);
   }
-  
+
   factory SkwasmImageFilter.blur({
     double sigmaX = 0.0,
     double sigmaY = 0.0,
@@ -62,9 +62,10 @@ class SkwasmImageFilter implements SceneImageFilter {
     return SkwasmImageFilter._(imageFilterCompose(nativeOuter.handle, nativeInner.handle));
   }
 
-  static final DomFinalizationRegistry _registry = 
-    DomFinalizationRegistry(createSkwasmFinalizer(imageFilterDispose));
+  static final SkwasmFinalizationRegistry<RawImageFilter> _registry =
+    SkwasmFinalizationRegistry<RawImageFilter>(imageFilterDispose);
 
+  @override
   final ImageFilterHandle handle;
   bool _isDisposed = false;
 
@@ -83,9 +84,9 @@ class SkwasmImageFilter implements SceneImageFilter {
   });
 }
 
-class SkwasmColorFilter {
+class SkwasmColorFilter implements SkwasmObjectWrapper<RawColorFilter> {
   SkwasmColorFilter._(this.handle) {
-    _registry.register(this, handle.address, this);
+    _registry.register(this);
   }
 
   factory SkwasmColorFilter.fromEngineColorFilter(EngineColorFilter colorFilter) =>
@@ -107,9 +108,10 @@ class SkwasmColorFilter {
     SkwasmColorFilter inner,
   ) => SkwasmColorFilter._(colorFilterCompose(outer.handle, inner.handle));
 
-  static final DomFinalizationRegistry _registry =
-    DomFinalizationRegistry(createSkwasmFinalizer(colorFilterDispose));
+  static final SkwasmFinalizationRegistry<RawColorFilter> _registry =
+    SkwasmFinalizationRegistry<RawColorFilter>(colorFilterDispose);
 
+  @override
   final ColorFilterHandle handle;
   bool _isDisposed = false;
 
@@ -121,9 +123,9 @@ class SkwasmColorFilter {
   }
 }
 
-class SkwasmMaskFilter {
+class SkwasmMaskFilter implements SkwasmObjectWrapper<RawMaskFilter> {
   SkwasmMaskFilter._(this.handle) {
-    _registry.register(this, handle.address, this);
+    _registry.register(this);
   }
 
   factory SkwasmMaskFilter.fromUiMaskFilter(ui.MaskFilter maskFilter) =>
@@ -132,9 +134,10 @@ class SkwasmMaskFilter {
       maskFilter.webOnlySigma
     ));
 
-  static final DomFinalizationRegistry _registry =
-    DomFinalizationRegistry(createSkwasmFinalizer(maskFilterDispose));
+  static final SkwasmFinalizationRegistry<RawMaskFilter> _registry =
+    SkwasmFinalizationRegistry<RawMaskFilter>(maskFilterDispose);
 
+  @override
   final MaskFilterHandle handle;
   bool _isDisposed = false;
 
--- a/lib/web_ui/lib/src/engine/skwasm/skwasm_impl/filters.dart
+++ b/lib/web_ui/lib/src/engine/skwasm/skwasm_impl/filters.dart
@@ -10,8 +10,10 @@ import 'package:ui/src/engine/skwasm/skwasm_impl.dart';
 import 'package:ui/ui.dart' as ui;
 
 class SkwasmImageFilter implements SceneImageFilter {
-  SkwasmImageFilter._(this.handle);
-
+  SkwasmImageFilter._(this.handle) {
+    _registry.register(this, handle.address, this);
+  }
+  
   factory SkwasmImageFilter.blur({
     double sigmaX = 0.0,
     double sigmaY = 0.0,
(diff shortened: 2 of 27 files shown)


lib/web_ui/lib/src/engine/skwasm/skwasm_impl/canvas.dart
```diff
@@ -18,9 +18,9 @@ class SkwasmCanvas implements SceneCanvas {
   SkwasmCanvas.fromHandle(this._handle);
   CanvasHandle _handle;
 
-  void delete() {
-    canvasDestroy(_handle);
-  }
+  // Note that we do not need to deal with the finalizer registry here, because
+  // the underlying native skia object is tied directly to the lifetime of the
+  // associated SkPictureRecorder.
```
yjbanov, line 23: Good point. I wonder if in debug mode we should watch for recorders dropped on the floor and issue an error.

lib/web_ui/lib/src/engine/skwasm/skwasm_impl/filters.dart
```diff
@@ -60,16 +62,20 @@ class SkwasmImageFilter implements SceneImageFilter {
...
-    }
-  }
+  static final SkwasmFinalizationRegistry<RawImageFilter> _registry =
+    SkwasmFinalizationRegistry<RawImageFilter>(imageFilterDispose);
 
+  @override
   final ImageFilterHandle handle;
   bool _isDisposed = false;
 
+  void dispose() {
```
kevmoo, line 72: Drive-by-comment: this code is duplicated a LOT – worth making it a `mixin` or something?

lib/web_ui/lib/src/engine/skwasm/skwasm_impl/filters.dart
```diff
@@ -100,33 +108,43 @@ class SkwasmColorFilter {
...
   ) => SkwasmColorFilter._(colorFilterCompose(outer.handle, inner.handle));
 
-  void dispose() {
-    if (!_isDisposed) {
-      colorFilterDispose(handle);
-      _isDisposed = true;
-    }
-  }
+  static final SkwasmFinalizationRegistry<RawColorFilter> _registry =
+    SkwasmFinalizationRegistry<RawColorFilter>(colorFilterDispose);
```
yjbanov, line 112: Per-type finalization registry is an interesting choice. Why can't we use one registry for all types? I'm not sure what the trade-offs are actually.

lib/web_ui/lib/src/engine/skwasm/skwasm_impl/font_collection.dart
```diff
@@ -19,28 +19,38 @@ import 'package:ui/src/engine/skwasm/skwasm_impl.dart';
 const String _robotoUrl =
     'https://fonts.gstatic.com/s/roboto/v20/KFOmCnqEu92Fr1Me5WZLCzYlKw.ttf';
 
-class SkwasmTypeface {
-  SkwasmTypeface(SkDataHandle data) : handle = typefaceCreate(data);
+class SkwasmTypeface implements SkwasmObjectWrapper<RawTypeface> {
```
yjbanov, line 22: Do we even hold onto typefaces? Could we accidentally delete a typeface that's still in use?

lib/web_ui/lib/src/engine/skwasm/skwasm_impl/image.dart
```diff
@@ -9,8 +9,10 @@ import 'package:ui/src/engine.dart';
 import 'package:ui/src/engine/skwasm/skwasm_impl.dart';
 import 'package:ui/ui.dart' as ui;
 
-class SkwasmImage implements ui.Image {
-  SkwasmImage(this.handle);
+class SkwasmImage implements SkwasmObjectWrapper<RawImage>, ui.Image {
+  SkwasmImage(this.handle) {
```
yjbanov, line 13: According to the `clone()` method below, it is the responsibility of the caller of the constructor to "pre-bump" the ref count of the handle. Does that hold for all invocations of this constructor?

lib/web_ui/lib/src/engine/skwasm/skwasm_impl/filters.dart
eyebrowsoffire: I took a couple different passes at it trying to DRY this out, this is actually more concise than it was in my previous passes. I did try actually factoring this out into a mixin, but there were two things that made it difficult:

(1) Each object has a different cleanup function for its handle, and since a mixin can't have a generative constructor it's hard for the adopting class to pass it in.
(2) The finalization registry is a static on the derived class, so the derived class can access it. I can't put the registry on the mixin because you'd only get one finalization registry for the mixin and not one per derived class.

There are probably clever ways to avoid these issues, but at a certain point I stopped refactoring and figured this is good enough. If you feel strongly, I can go back and try to figure out a way to fold some of this logic into `SkwasmObjectWrapper` though... maybe...

kevmoo: Looking at the code, it's clear we'd need some funky new language features to generalize this. Nevermind. 😢

lib/web_ui/lib/src/engine/skwasm/skwasm_impl/memory.dart
```diff
@@ -0,0 +1,31 @@
...
+
+import 'package:ui/src/engine.dart';
+
+abstract class SkwasmObjectWrapper<T extends NativeType> {
+  Pointer<T> get handle;
+}
+
+typedef DisposeFunction<T extends NativeType> = void Function(Pointer<T>);
+
+class SkwasmFinalizationRegistry<T extends NativeType> {
```
yjbanov, line 16: Do we actually need the `<T>` type parameter? If not, then I think we can use a shared `DomFinalizationRegistry` instance for all objects. But, again, I'm not sure what kind of tradeoff exists in using one vs many registries 🤷  Just pointing out a possibility.

lib/web_ui/lib/src/engine/skwasm/skwasm_impl/paint.dart
```diff
@@ -8,15 +8,26 @@ import 'package:ui/src/engine.dart';
...
+class SkwasmPaint implements SkwasmObjectWrapper<RawPaint>, ui.Paint {
+  factory SkwasmPaint() => SkwasmPaint._fromHandle(paintCreate());
+
+  SkwasmPaint._fromHandle(this.handle) {
+    _registry.register(this);
+  }
+
+  void dispose() {
+    assert(!_isDisposed);
+    _registry.unregister(this);
```
yjbanov, line 20: This pattern is used in all types. I wonder it can be rolled into `SkwasmFinalizationRegistry`. We could have a nicely symmetric API:

```dart
registry.register(this);
registry.dispose(this); // also unregisters since it doesn't need to be GC'd any more
```

lib/web_ui/lib/src/engine/skwasm/skwasm_impl/memory.dart
```diff
@@ -0,0 +1,31 @@
...
+  SkwasmFinalizationRegistry(DisposeFunction<T> dispose)
+    : registry = DomFinalizationRegistry(((JSNumber address) =>
+      dispose(Pointer<T>.fromAddress(address.toDart.toInt()))
+    ).toJS);
+
+  final DomFinalizationRegistry registry;
+
+  void register(SkwasmObjectWrapper<T> wrapper) {
```
yjbanov, line 24: One nice feature of `UniqueRef`/`CountedRef` is that they integrate with `Instrumentation` which helps you find leaks and double-frees. But maybe this can be added in a follow-up PR, if the design allows it.

lib/web_ui/lib/src/engine/skwasm/skwasm_impl/filters.dart
kevmoo: Because each type has its own dispose function?

lib/web_ui/lib/src/engine/skwasm/skwasm_impl/filters.dart
yjbanov: `SkwasmFinalizationRegistry` already takes a reference to the type-specific clean-up function.  Maybe this is doable via `SkwasmFinalizationRegistry`. (see https://github.com/flutter/engine/pull/42328#discussion_r1206056640)

lib/web_ui/lib/src/engine/skwasm/skwasm_impl/filters.dart
kevmoo: Back to the mixin discussion, you could have a `Map<Type, DisposeFunction>`. Then you have one registry, but you need to do a map access on every dispose. 🤷

eyebrowsoffire: Each handle has a different native disposal function. The alternative would be to have a single finalization registry, but add an extra layer of indirection that wraps the handle and is polymorphic on the dispose method (this is basically what you're doing on the CK side with your Ref objects). For the skwasm case, I found it a little cleaner to just do the per-object finalization registry and not end up with an extra layer of indirection.

lib/web_ui/lib/src/engine/skwasm/skwasm_impl/font_collection.dart
eyebrowsoffire: The typeface is a shared pointer under the hood, and the font collection hangs onto it. The `SkwasmFontCollection` does hold onto typeface objects explicitly. We used to need it in order to scan incoming text to find missing code points, but Skia does that for it for us now. However, when we reset the font collection (which is just a debug only situation for now), we re-register the base Roboto font with the new collection, so keeping the typeface around is useful in that scenario (rather than having to recreate it from data again). So I don't know if there is a super important reason to hang onto them in prod right now, but I don't think it hurts anything and makes the debug reset part easier.

eyebrowsoffire: Rereading my comment, I'm not sure my first point was clear, but what I mean is that we don't have to worry about the typefaces dangling if we don't hang onto them, because the FontCollection has its own shared reference to each typeface. Our hanging onto it or not hanging onto it won't affect the operation of the FontCollection itself.

lib/web_ui/lib/src/engine/skwasm/skwasm_impl/image.dart
eyebrowsoffire: Basically yes. This constructor (and all skwasm object wrappers constructors that just take a raw handle) expect a handle that has +1 ref count on it, and the object basically takes ownership of that ref count. All the native `create` functions return a handle with +1 ref count on it, and when cloning we bump the ref count for this same reason.

lib/web_ui/lib/src/engine/skwasm/skwasm_impl/memory.dart
eyebrowsoffire: We do need this type parameter, because each finalization registry has its own disposal function that it invokes and that disposal function takes a specific pointer type. I mentioned this in the other comment that it is possible to do this with another wrapper which basically "type-erases" this problem away, but I'd rather not add another layer of indirection.

lib/web_ui/lib/src/engine/skwasm/skwasm_impl/memory.dart
```diff
@@ -7,25 +7,40 @@ import 'dart:js_interop';
...
 
 class SkwasmFinalizationRegistry<T extends NativeType> {
-  SkwasmFinalizationRegistry(DisposeFunction<T> dispose)
+  SkwasmFinalizationRegistry(this.dispose)
     : registry = DomFinalizationRegistry(((JSNumber address) =>
       dispose(Pointer<T>.fromAddress(address.toDart.toInt()))
     ).toJS);
 
   final DomFinalizationRegistry registry;
+  final DisposeFunction<T> dispose;
```
kevmoo, line 36: nit: could these fields be private?

lib/web_ui/lib/src/engine/skwasm/skwasm_impl/memory.dart
```diff
@@ -7,25 +7,40 @@ import 'dart:js_interop';
...
-  Pointer<T> get handle;
+class SkwasmObjectWrapper<T extends NativeType> {
+  SkwasmObjectWrapper(this.handle, this.registry) {
+    registry.register(this);
+  }
+  final SkwasmFinalizationRegistry<T> registry;
+  final Pointer<T> handle;
+  bool _isDisposed = false;
+
+  void dispose() {
```
kevmoo, line 18: LOVE!

yjbanov (approved):
lgtm

lib/web_ui/lib/src/engine/skwasm/skwasm_impl/memory.dart
yjbanov: One unusual aspect of our codebase is that everything under `src/engine/` is automatically private, because we rewrite the code into `dart:_engine`. So it's OK to leave things public if it helps with testing and such.

lib/web_ui/lib/src/engine/skwasm/skwasm_impl/memory.dart
yjbanov: Aye!

lib/web_ui/lib/src/engine/skwasm/skwasm_impl/memory.dart
kevmoo: defaulting to private makes it easy to understand what's used. It also helps find when things are no longer used.

PR #42329, opened 2023-05-25
[licenses] remove obsolete license

XilaiZhang:
context: b/267810411

According to the source at https://github.com/openssl/openssl/tree/master/include/openssl, it seems like this license is no longer there. Would be great if we could remove this license from github, so that we don't need to add it to g3.


Hixie:
This file is autogenerated, so you can't just remove it like that. You have to rerun the license script to generate these files.

Also this particular file is just the debug output, not the actual license file. See the README in the license script directory for details.

The file being edited in this PR tells you that the file that contains the license in question is 
```
FILE: ../../../third_party/boringssl/src/include/openssl/x509.h
```

That's where I would start if you want to remove this license.

XilaiZhang:
Thanks for explaining! Ahh yeah there is a readme at third_party/tools/licenses, sure will do it the proper way

XilaiZhang:
Looks like the license in this PR exists in a very old version of boring ssl. updated in b/267810411 and will close this pr.

PR #42331, opened 2023-05-25
[Impeller] Disable the wide gamut settings flag on iOS simulators

jason-simmons:
Wide gamut is not supported on the iOS simulator, and the iOS platform's FlutterView.isWideGamutSupported checked for this.  However, the ImageDecoder decides whether to enable wide gamut based on the engine's settings, so the flag in the settings needs to be modified for the simulator.


gaaclarke (approved):
lgtm

PR #42332, opened 2023-05-26
Use ShellTest::Config struct to create Shell

dkwingsmt:
This PR refactors `ShellTest::CreateShell` so that instead of taking a list of optional parameters, it takes a `Config` struct, almost all fields of which have default values. This makes calling this method cleaner, and makes it easier to add more parameters. A short version is also kept that takes `settings` and `task_runners`, since this version is used so often.

Currently `ShellTest::CreateShell`'s full version has 2 mandatory parameters and 5 optional parameters. Many calls to this method have to list a long list of default parameters before going to the single one that needs overwriting, and adding new parameters to this call is even more tedious.

## Pre-launch Checklist

- [ ] I signed the [CLA].

### DIFF
--- a/shell/common/shell_test.h
+++ b/shell/common/shell_test.h
@@ -31,22 +31,29 @@ namespace testing {
 
 class ShellTest : public FixtureTest {
  public:
+  struct Config {
+    // Required.
+    const Settings& settings;
+    // Defaults to GetTaskRunnersForFixture().
+    std::optional<TaskRunners> task_runners = {};
+    bool simulate_vsync = false;
+    std::shared_ptr<ShellTestExternalViewEmbedder>
+        shell_test_external_view_embedder = nullptr;
+    bool is_gpu_disabled = false;
+    ShellTestPlatformView::BackendType rendering_backend =
+        ShellTestPlatformView::BackendType::kDefaultBackend;
+    // Defaults to calling ShellTestPlatformView::Create with the provided
+    // arguments.
+    Shell::CreateCallback<PlatformView> platform_view_create_callback;
+  };
+
   ShellTest();
 
   Settings CreateSettingsForFixture() override;
-  std::unique_ptr<Shell> CreateShell(const Settings& settings,
-                                     bool simulate_vsync = false);
   std::unique_ptr<Shell> CreateShell(
       const Settings& settings,
-      TaskRunners task_runners,
-      bool simulate_vsync = false,
-      const std::shared_ptr<ShellTestExternalViewEmbedder>&
-          shell_test_external_view_embedder = nullptr,
-      bool is_gpu_disabled = false,
-      ShellTestPlatformView::BackendType rendering_backend =
-          ShellTestPlatformView::BackendType::kDefaultBackend,
-      Shell::CreateCallback<PlatformView> platform_view_create_callback =
-          nullptr);
+      std::optional<TaskRunners> task_runners = {});
+  std::unique_ptr<Shell> CreateShell(const Config& config);
   void DestroyShell(std::unique_ptr<Shell> shell);
   void DestroyShell(std::unique_ptr<Shell> shell,
                     const TaskRunners& task_runners);
--- a/shell/common/shell_test.h
+++ b/shell/common/shell_test.h
@@ -36,12 +36,7 @@ class ShellTest : public FixtureTest {
     const Settings& settings;
     // Defaults to GetTaskRunnersForFixture().
     std::optional<TaskRunners> task_runners = {};
-    bool simulate_vsync = false;
-    std::shared_ptr<ShellTestExternalViewEmbedder>
-        shell_test_external_view_embedder = nullptr;
     bool is_gpu_disabled = false;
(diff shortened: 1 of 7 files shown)


shell/common/shell_test.h
```diff
@@ -31,22 +31,31 @@ namespace testing {
 
 class ShellTest : public FixtureTest {
  public:
+  struct Config {
+    // Required.
+    const Settings& settings;
+    // Defaults to &GetTaskRunnersForFixture().
+    const TaskRunners* task_runners;
```
dnfield, line 38: Please don't use a raw pointer here.

Just keep this as a value - avoids one more semantic change in the way these tests work, and avoids ambiguity around whether this can be nullptr and what that means.

dnfield: Also in general here: instead of specifying defaults in a comment, initialize the values here with a default.

dnfield (changes requested):
I like this idea, the config object could be cleaned up a bit.

If assigning defaults needs more than simple logic (like perhaps the default for task_runners), consider using a builder pattern.

shell/common/shell_test.h
```diff
@@ -31,22 +31,29 @@ namespace testing {
 
 class ShellTest : public FixtureTest {
  public:
+  struct Config {
+    // Required.
+    const Settings& settings;
+    // Defaults to GetTaskRunnersForFixture().
+    std::optional<TaskRunners> task_runners = {};
```
dnfield, line 38: This is where I think the builder could be helpful - you don't want the public interface here to actually be optional.

But if you don't want to use a builder I think we should make the `std::optional` be on the constructor for this struct and make the member non-optional - because otherwise callers will always have to check (and get linted against not checking now) whether the optional has a value when it really always does.

dkwingsmt: I can create a `TaskRunnersBuilder`, which has two ctors:
```
TaskRunnersBuilder(TaskRunner); // implicit conversion
TaskRunnersBuilder(); // Defaults to GetTaskRunnersForFixture()
```

But I think it's really just reinventing the wheel for `std::optional` because it's not semantically obvious what the default behavior is (and it's now hidden even deeper).

And what's worse, because this field is a value, not a pointer of a function, I can't really specify an "empty" value if not for the default ctor.

> because otherwise callers will always have to check (and get linted against not checking now) whether the optional has a value when it really always does

I don't think this is a problem, because this struct really is only used by `ShellTest::CreateShell`.

dkwingsmt:
Also I created a builder for `PlatformView`. Let me know what you think!

shell/common/shell_test.h
dkwingsmt: Another option is to mark `task_runners` as mandatory so that you always explicit write `GetTaskRunnersForFixture()` for every call.

shell/common/shell_test_platform_view.cc
```diff
@@ -57,5 +59,33 @@ std::unique_ptr<ShellTestPlatformView> ShellTestPlatformView::Create(
   }
 }
 
+ShellTestPlatformViewBuilder::ShellTestPlatformViewBuilder(Config config)
+    : config_(config) {}
+
+std::unique_ptr<PlatformView> ShellTestPlatformViewBuilder::operator()(
```
dnfield, line 65: nit: avoid defining a custom operator for this, just make a `Build` method.

dnfield (approved):
LGTM with nit

shell/common/shell_test_platform_view.cc
dkwingsmt: The custom operator is to make this class assignable to the current `std::function` parameter. We can turn this into a `Build` method instead, but then the several cases that currently define a custom closure will have to define a new class that accepts a closure. Should I do it?

dkwingsmt:
```cpp
  Shell::CreateCallback<PlatformView> platform_view_create_callback =
      [task_runners, main_context](flutter::Shell& shell) {
        auto result = std::make_unique<TestPlatformView>(shell, task_runners);
        ON_CALL(*result, CreateRenderingSurface())
            .WillByDefault(::testing::Invoke([main_context] {
              auto surface = std::make_unique<MockSurface>();
              ON_CALL(*surface, GetContext())
                  .WillByDefault(Return(main_context.get()));
              ON_CALL(*surface, IsValid()).WillByDefault(Return(true));
              ON_CALL(*surface, MakeRenderContextCurrent())
                  .WillByDefault(::testing::Invoke([] {
                    return std::make_unique<GLContextDefaultResult>(true);
                  }));
              return surface;
            }));
        return result;
      };

  auto shell = CreateShell({
      .settings = settings,
      .task_runners = task_runners,
      .platform_view_create_callback = platform_view_create_callback,
  });
```

https://github.com/flutter/engine/blob/6699bbb43b9d90d49d374ca215c52b383a77b7c1/shell/common/shell_unittests.cc#L1608

dnfield: That's fine then.

PR #42333, opened 2023-05-26
[Impeller] Switch back to shared buffer UI image uploading on iOS

bdero:
Resolves https://github.com/flutter/flutter/issues/126878.

This is just a soft revert of the new blit upload path that was added in https://github.com/flutter/engine/pull/40410.

Switches back to uploading via `SetContents`. The blit pass that generates the mipmaps still fails if the app is backgrounded, but not writing all the mip levels is safe for Metal and Vulkan -- GLES is the problem child.

The downside of course is that we lose lossy compression for the time being (supported on Family 8/A15/iPhone 13 and above). Perhaps we can come back around to this and add, say, a command buffer queue that defers execution until apps are foregrounded.

### DIFF
--- a/lib/ui/painting/image_decoder_impeller.cc
+++ b/lib/ui/painting/image_decoder_impeller.cc
@@ -452,13 +452,8 @@ void ImageDecoderImpeller::Decode(fml::RefPtr<ImageDescriptor> descriptor,
           // texture is implemented on other platforms.
           sk_sp<DlImage> image;
           std::string decode_error;
-#ifdef FML_OS_IOS
-          std::tie(image, decode_error) = UploadTextureToPrivate(
-              context, bitmap_result.device_buffer, bitmap_result.image_info);
-#else
           std::tie(image, decode_error) =
               UploadTextureToShared(context, bitmap_result.sk_bitmap);
-#endif
           result(image, decode_error);
         };
         // TODO(jonahwilliams): https://github.com/flutter/flutter/issues/123058


lib/ui/painting/image_decoder_impeller.cc
```diff
@@ -452,13 +452,8 @@ void ImageDecoderImpeller::Decode(fml::RefPtr<ImageDescriptor> descriptor,
           // texture is implemented on other platforms.
           sk_sp<DlImage> image;
           std::string decode_error;
-#ifdef FML_OS_IOS
-          std::tie(image, decode_error) = UploadTextureToPrivate(
-              context, bitmap_result.device_buffer, bitmap_result.image_info);
-#else
```
dnfield, line 458: Instead of this, you can use the GPU sync switch to conditionally use this path on iOS.

dnfield (changes requested):
We can use the sync switch for this.

bdero:
I believe @dnfield is preparing a different patch for this with a test. So I'm gonna hold off on touching up this PR for the moment.

lib/ui/painting/image_decoder_impeller.cc
jonahwilliams: You need to disable mipmaps as well

bdero: I mentioned mipmaps in the description.

> The blit pass that generates the mipmaps still fails if the app is backgrounded, but not writing all the mip levels is safe for Metal and Vulkan -- GLES is the problem child.

dnfield: On some versions of iOS we've seen crashes when accessing command encoders in the background

bdero:
Closing in favor of https://github.com/flutter/engine/pull/42349

PR #42337, opened 2023-05-26
Started executing vulkan unit tests with validation on macos

gaaclarke:
fixes https://github.com/flutter/flutter/issues/127634

This PR depends on a buildroot change at https://github.com/flutter/buildroot/pull/737

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/testing/run_tests.py
+++ b/testing/run_tests.py
@@ -481,7 +481,10 @@ def run_cc_tests(build_dir, executable_filter, coverage, capture_core_dump):
             '1',  # Validates accesses to threadgroup memory.
         'MTL_SHADER_VALIDATION_TEXTURE_USAGE':
             '1',  # Validates that texture references are not nil.
+        # Note: built from //third_party/swiftshader
         'VK_ICD_FILENAMES': os.path.join(build_dir, 'vk_swiftshader_icd.json'),
+        # Note: built from //third_party/vulkan_validation_layers:vulkan_gen_json_files
+        # and //third_party/vulkan_validation_layers.
         'VK_LAYER_PATH': os.path.join(build_dir, 'vulkan-data'),
         'VK_INSTANCE_LAYERS': 'VK_LAYER_KHRONOS_validation',
     }
--- a/testing/run_tests.py
+++ b/testing/run_tests.py
@@ -75,19 +75,23 @@ def run_cmd(
   print(f'Running command "{command_string}"')
 
   start_time = time.time()
-  collect_output = forbidden_output or allowed_failure_output
-  stdout_pipe = sys.stdout if not collect_output else subprocess.PIPE
-  stderr_pipe = sys.stderr if not collect_output else subprocess.PIPE
 
   process = subprocess.Popen(
       cmd,
-      stdout=stdout_pipe,
-      stderr=stderr_pipe,
+      stdout=subprocess.PIPE,
+      stderr=subprocess.STDOUT,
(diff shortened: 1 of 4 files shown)


testing/run_tests.py
```diff
@@ -75,19 +75,23 @@ def run_cmd(
...
   start_time = time.time()
-  collect_output = forbidden_output or allowed_failure_output
-  stdout_pipe = sys.stdout if not collect_output else subprocess.PIPE
-  stderr_pipe = sys.stderr if not collect_output else subprocess.PIPE
 
   process = subprocess.Popen(
       cmd,
-      stdout=stdout_pipe,
-      stderr=stderr_pipe,
+      stdout=subprocess.PIPE,
```
gaaclarke, line 81: This change allows us to stream the output to stdout but also capture it for inspection.  The tests take a long enough time, it's a pain not seeing the streaming results.

testing/run_tests.py
```diff
@@ -75,19 +75,23 @@ def run_cmd(
...
   )
-  stdout, stderr = process.communicate()
+  output = ''
+
+  for line in iter(process.stdout.readline, ''):
+    output += line
+    sys.stdout.write(line)
+
+  sys.stdout.flush()
+  process.wait()
```
zanderso, line 94: Does `process.stdout.readline` drain the buffer to avoid the deadlock concern with `wait()` called out in the docs? https://docs.python.org/3/library/subprocess.html#subprocess.Popen.wait

testing/run_tests.py
```diff
@@ -75,19 +75,23 @@ def run_cmd(
...
       env=env,
       universal_newlines=True,
       **kwargs
   )
-  stdout, stderr = process.communicate()
+  output = ''
+
+  for line in iter(process.stdout.readline, ''):
+    output += line
+    sys.stdout.write(line)
```
zanderso, line 91: Previously the test script would only generate output when a test fails. This is nice when diagnosing failures on CI since the succeeding tests add only minimal output to the logs. Am I misreading what this does? Is there some way to make this configurable?

testing/run_tests.py
gaaclarke: That's my understanding.  It's reading the pipe until EOF so there should be no more waiting on the pipe by the time we call `wait`.

testing/run_tests.py
gaaclarke: I believe you are misremembering how it works.  If `collect_output == true` the output would be printed on line 119.  If `collect_output == false` it would get printed out by virtue of using `stdout=sys.stdout`.

testing/run_tests.py
```diff
@@ -490,6 +482,8 @@ def make_test(name, flags=None, extra_env=None):
         'MTL_SHADER_VALIDATION_TEXTURE_USAGE':
             '1',  # Validates that texture references are not nil.
         'VK_ICD_FILENAMES': os.path.join(build_dir, 'vk_swiftshader_icd.json'),
+        'VK_LAYER_PATH': os.path.join(build_dir, 'vulkan-data'),
```
zanderso, line 485: A comment pointing to the GN file with the string that has to be the same as this one would be good?

zanderso (approved):
lgtm w/ one comment.

testing/run_tests.py
gaaclarke: done

PR #42349, opened 2023-05-26
[Impeller] Avoid encoding metal commands while the GPU is unavailable when decoding images.

dnfield:
Fixes https://github.com/flutter/flutter/issues/126878

This disables device private upload on iOS when backgrounded, and disables mipmap generation when backgrounded.

We don't have a good way to test the core problem in this repo because it only reproduces on physical iOS hardware - simulators don't really care if you do this stuff in the background.

I'll write a devicelab test in the framework to capture this. In the mean time it can be reviewed.

We could consider making the IOManager a shared_ptr instead of having an fml::WeakPtr and that'd clean up some of the extra arguments to engine construction - or we could consider vending the sync switch from impeller::Context unconditionally, but it's pretty iOS specific...

### DIFF
--- a/lib/ui/painting/image_decoder.h
+++ b/lib/ui/painting/image_decoder.h
@@ -27,7 +27,8 @@ class ImageDecoder {
       const Settings& settings,
       const TaskRunners& runners,
       std::shared_ptr<fml::ConcurrentTaskRunner> concurrent_task_runner,
-      fml::WeakPtr<IOManager> io_manager);
+      fml::WeakPtr<IOManager> io_manager,
+      const std::shared_ptr<fml::SyncSwitch>& gpu_disabled_switch = nullptr);
 
   virtual ~ImageDecoder();
 
--- a/lib/ui/painting/image_decoder.h
+++ b/lib/ui/painting/image_decoder.h
@@ -28,7 +28,7 @@ class ImageDecoder {
       const TaskRunners& runners,
       std::shared_ptr<fml::ConcurrentTaskRunner> concurrent_task_runner,
       fml::WeakPtr<IOManager> io_manager,
-      const std::shared_ptr<fml::SyncSwitch>& gpu_disabled_switch = nullptr);
+      const std::shared_ptr<fml::SyncSwitch>& gpu_disabled_switch);
 
   virtual ~ImageDecoder();
 
--- a/lib/ui/painting/image_decoder_impeller.cc
+++ b/lib/ui/painting/image_decoder_impeller.cc
@@ -248,18 +248,11 @@ DecompressResult ImageDecoderImpeller::DecompressTexture(
                           .image_info = scaled_bitmap->info()};
 }
 
-std::pair<sk_sp<DlImage>, std::string>
-ImageDecoderImpeller::UploadTextureToPrivate(
+/// Only call this method if the GPU is available.
+static std::pair<sk_sp<DlImage>, std::string> UnsafeUploadTextureToPrivate(
     const std::shared_ptr<impeller::Context>& context,
     const std::shared_ptr<impeller::DeviceBuffer>& buffer,
     const SkImageInfo& image_info) {
-  TRACE_EVENT0("impeller", __FUNCTION__);
-  if (!context) {
-    return std::make_pair(nullptr, "No Impeller context is available");
-  }
-  if (!buffer) {
-    return std::make_pair(nullptr, "No Impeller device buffer is available");
-  }
   const auto pixel_format =
       impeller::skia_conversions::ToPixelFormat(image_info.colorType());
   if (!pixel_format) {
@@ -321,11 +314,40 @@ ImageDecoderImpeller::UploadTextureToPrivate(
 }
 
 std::pair<sk_sp<DlImage>, std::string>
+ImageDecoderImpeller::UploadTextureToPrivate(
+    const std::shared_ptr<impeller::Context>& context,
+    const std::shared_ptr<impeller::DeviceBuffer>& buffer,
+    const SkImageInfo& image_info,
+    std::shared_ptr<SkBitmap> bitmap,
+    const std::shared_ptr<fml::SyncSwitch>& gpu_disabled_switch) {
+  TRACE_EVENT0("impeller", __FUNCTION__);
+  if (!context) {
+    return std::make_pair(nullptr, "No Impeller context is available");
+  }
+  if (!buffer) {
+    return std::make_pair(nullptr, "No Impeller device buffer is available");
+  }
+
+  std::pair<sk_sp<DlImage>, std::string> result;
+  gpu_disabled_switch->Execute(
+      fml::SyncSwitch::Handlers()
+          .SetIfFalse([&result, context, buffer, image_info] {
+            result = UnsafeUploadTextureToPrivate(context, buffer, image_info);
+          })
+          .SetIfTrue([&result, context, bitmap, gpu_disabled_switch] {
+            // create_mips is false because we already know the GPU is disabled.
+            result = UploadTextureToShared(context, bitmap, gpu_disabled_switch,
+                                           false);
+          }));
+  return result;
+}
+
+std::pair<sk_sp<DlImage>, std::string>
 ImageDecoderImpeller::UploadTextureToShared(
     const std::shared_ptr<impeller::Context>& context,
     std::shared_ptr<SkBitmap> bitmap,
-    bool create_mips,
-    const std::shared_ptr<fml::SyncSwitch>& gpu_disabled_switch) {
+    const std::shared_ptr<fml::SyncSwitch>& gpu_disabled_switch,
+    bool create_mips) {
   TRACE_EVENT0("impeller", __FUNCTION__);
   if (!context) {
     return std::make_pair(nullptr, "No Impeller context is available");
@@ -374,12 +396,12 @@ ImageDecoderImpeller::UploadTextureToShared(
 
   if (texture_descriptor.mip_count > 1u && create_mips) {
     std::optional<std::string> decode_error;
-#if FML_OS_IOS
+
+    // The only platform that needs mipmapping unconditionally is GL.
+    // GL based platforms never disable GPU access.
+    // This is only really needed for iOS.
     gpu_disabled_switch->Execute(fml::SyncSwitch::Handlers().SetIfFalse(
         [context, &texture, &decode_error] {
-#else
-    [context, &texture, &decode_error] {
-#endif  // FML_OS_IOS
           auto command_buffer = context->CreateCommandBuffer();
           if (!command_buffer) {
             decode_error =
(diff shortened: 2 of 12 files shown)


lib/ui/painting/image_decoder_impeller.cc
```diff
@@ -370,32 +373,44 @@ ImageDecoderImpeller::UploadTextureToShared(
...
-    blit_pass->GenerateMipmap(texture);
-
-    blit_pass->EncodeCommands(context->GetResourceAllocator());
-    if (!command_buffer->SubmitCommands()) {
-      std::string decode_error("Failed to submit blit pass command buffer.");
-      FML_DLOG(ERROR) << decode_error;
-      return std::make_pair(nullptr, decode_error);
+    std::optional<std::string> decode_error;
+#if FML_OS_IOS
+    gpu_disabled_switch->Execute(fml::SyncSwitch::Handlers().SetIfFalse(
```
gaaclarke, line 378: Should the true branch be setting `decode_error`?

lib/ui/painting/image_decoder.h
```diff
@@ -27,7 +27,8 @@ class ImageDecoder {
       const Settings& settings,
       const TaskRunners& runners,
       std::shared_ptr<fml::ConcurrentTaskRunner> concurrent_task_runner,
-      fml::WeakPtr<IOManager> io_manager);
+      fml::WeakPtr<IOManager> io_manager,
+      const std::shared_ptr<fml::SyncSwitch>& gpu_disabled_switch = nullptr);
```
gaaclarke, line 31: Everyone should have a syncswitch so there's no need to support nullptr.

lib/ui/painting/image_decoder_impeller.cc
```diff
@@ -446,24 +461,39 @@ void ImageDecoderImpeller::Decode(fml::RefPtr<ImageDescriptor> descriptor,
...
-          std::tie(image, decode_error) =
-              UploadTextureToShared(context, bitmap_result.sk_bitmap);
-#endif
+          FML_DCHECK(gpu_disabled_switch);
+          if (gpu_disabled_switch) {
+            gpu_disabled_switch->Execute(fml::SyncSwitch::Handlers().SetIfFalse(
+                [&needs_upload, &image, &decode_error, &context,
+                 &bitmap_result] {
+                  needs_upload = false;
+                  FML_LOG(ERROR) << "Yay";
```
gaaclarke, line 479: Stray log.

lib/ui/painting/image_decoder_impeller.cc
```diff
@@ -446,24 +461,39 @@ void ImageDecoderImpeller::Decode(fml::RefPtr<ImageDescriptor> descriptor,
...
 #ifdef FML_OS_IOS
-          std::tie(image, decode_error) = UploadTextureToPrivate(
-              context, bitmap_result.device_buffer, bitmap_result.image_info);
-#else
-          std::tie(image, decode_error) =
-              UploadTextureToShared(context, bitmap_result.sk_bitmap);
-#endif
+          FML_DCHECK(gpu_disabled_switch);
+          if (gpu_disabled_switch) {
+            gpu_disabled_switch->Execute(fml::SyncSwitch::Handlers().SetIfFalse(
```
gaaclarke, line 475: It's worth adding a comment about the omission of a true branch.

gaaclarke (changes requested):
Looks good for the most part.  The only thing that should get changed is removing the conditional compilation.  I think we should treat sync_switch as a nonnull parameter instead.  Also, I want clarity in for the true branches of the syncswitches.  We should have comments that at least explain their omission.

lib/ui/painting/image_decoder_impeller.cc
```diff
@@ -370,32 +373,44 @@ ImageDecoderImpeller::UploadTextureToShared(
...
-    blit_pass->SetLabel("Mipmap Blit Pass");
-    blit_pass->GenerateMipmap(texture);
-
-    blit_pass->EncodeCommands(context->GetResourceAllocator());
-    if (!command_buffer->SubmitCommands()) {
-      std::string decode_error("Failed to submit blit pass command buffer.");
-      FML_DLOG(ERROR) << decode_error;
-      return std::make_pair(nullptr, decode_error);
+    std::optional<std::string> decode_error;
+#if FML_OS_IOS
```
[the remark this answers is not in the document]
dnfield: Sounds good, that will actually make this more testable too.

lib/ui/painting/image_decoder_impeller.cc
dnfield: Done

dnfield: (removed the one below too)

lib/ui/painting/image_decoder_impeller.cc
dnfield: This got moved and there's no longer an omission.

dnfield:
Added a test.

Ideally this would get a screenshot based test on a real device, but that will probably take some more time to write. For now the test I added validates that if the sync switch is disabled we do not create any command buffers.

lib/ui/painting/image_decoder_impeller.cc
```diff
@@ -318,10 +313,40 @@ ImageDecoderImpeller::UploadTextureToPrivate(
...
+  std::pair<sk_sp<DlImage>, std::string> result;
+  gpu_disabled_switch->Execute(
+      fml::SyncSwitch::Handlers()
+          .SetIfFalse([&result, context, buffer, image_info] {
+            result = UnsafeUploadTextureToPrivate(context, buffer, image_info);
+          })
+          .SetIfTrue([&result, context, bitmap, gpu_disabled_switch] {
+            // create_mips is false because we already know the GPU is disabled.
+            result = UploadTextureToShared(context, bitmap, gpu_disabled_switch,
+                                           false);
```
gaaclarke, line 340:
```suggestion
                                           /*create_mips=*/false);
```

lib/ui/painting/image_decoder_impeller.cc
```diff
@@ -318,10 +313,40 @@ ImageDecoderImpeller::UploadTextureToPrivate(
...
+  }
+
+  std::pair<sk_sp<DlImage>, std::string> result;
+  gpu_disabled_switch->Execute(
+      fml::SyncSwitch::Handlers()
+          .SetIfFalse([&result, context, buffer, image_info] {
+            result = UnsafeUploadTextureToPrivate(context, buffer, image_info);
+          })
+          .SetIfTrue([&result, context, bitmap, gpu_disabled_switch] {
+            // create_mips is false because we already know the GPU is disabled.
```
gaaclarke, line 338: Can't this result in suboptimal rendering based on when the gpu is disabled?  Are we going to get bug reports about blurry images that we can't source back to coming from backgrounded apps?

lib/ui/painting/image_decoder_impeller.cc
gaaclarke: We ripped out opengl for iOS right?

lib/ui/painting/image_decoder_impeller.cc
```diff
@@ -446,24 +479,27 @@ void ImageDecoderImpeller::Decode(fml::RefPtr<ImageDescriptor> descriptor,
...
           std::tie(image, decode_error) = UploadTextureToPrivate(
-              context, bitmap_result.device_buffer, bitmap_result.image_info);
+              context, bitmap_result.device_buffer, bitmap_result.image_info,
+              bitmap_result.sk_bitmap, gpu_disabled_switch);
 #else
-          std::tie(image, decode_error) =
-              UploadTextureToShared(context, bitmap_result.sk_bitmap);
-#endif
+          std::tie(image, decode_error) = UploadTextureToShared(
+              context, bitmap_result.sk_bitmap, gpu_disabled_switch, true);
```
gaaclarke, line 495:
```suggestion
              context, bitmap_result.sk_bitmap, gpu_disabled_switch, /*create_mips=*/true);
```

shell/common/engine_unittests.cc
```diff
@@ -162,7 +162,8 @@ TEST_F(EngineTest, Create) {
         /*animator=*/std::move(animator_),
         /*io_manager=*/io_manager_,
         /*font_collection=*/std::make_shared<FontCollection>(),
-        /*runtime_controller=*/std::move(runtime_controller_));
+        /*runtime_controller=*/std::move(runtime_controller_),
+        /*gpu_disabled_switch=*/nullptr);
```
gaaclarke, line 166: nit: make syncswitches here since technically it isn't supposed to be null.  We could also add a dcheck that it isn't null.

gaaclarke (approved):

lib/ui/painting/image_decoder_impeller.cc
```diff
@@ -370,32 +395,40 @@ ImageDecoderImpeller::UploadTextureToShared(
...
+          blit_pass->SetLabel("Mipmap Blit Pass");
+          blit_pass->GenerateMipmap(texture);
+
+          blit_pass->EncodeCommands(context->GetResourceAllocator());
+          if (!command_buffer->SubmitCommands()) {
+            decode_error = "Failed to submit blit pass command buffer.";
+            return;
+          }
+          command_buffer->WaitUntilScheduled();
+        }));
```
bdero, line 427: This isn't something we should solve now, but we'll eventually need a solution to defer generating mipmaps. Not generating mipmaps will cause min filtered images to continue appearing more blurry than Skia once we start properly setting the mip bias.

lib/ui/painting/image_decoder_impeller.cc
```diff
@@ -370,32 +395,40 @@ ImageDecoderImpeller::UploadTextureToShared(
...
+          }
+          command_buffer->SetLabel("Mipmap Command Buffer");
+
+          auto blit_pass = command_buffer->CreateBlitPass();
+          if (!blit_pass) {
+            decode_error = "Could not create blit pass for mipmap generation.";
+            return;
+          }
+          blit_pass->SetLabel("Mipmap Blit Pass");
+          blit_pass->GenerateMipmap(texture);
```
bdero, line 419: The `texture.NeedsMipmapGeneration()` check also needs to be removed from `render_pass_mtl.mm`.

bdero (changes requested):

lib/ui/painting/image_decoder_impeller.cc
dnfield: It can result in suboptimal rendering. We could choose to build mips when the GPU is available again, but that's going to be a more invasive/risky change. I'd like this to be cherry pickable.

lib/ui/painting/image_decoder_impeller.cc
dnfield: Correct

lib/ui/painting/image_decoder_impeller.cc
dnfield: Should we change that to a DLOG on iOS?

dnfield: I ifdef'd it out for IOS.

bdero (approved):
LGTM

lib/ui/painting/image_decoder_impeller.cc
bdero: I think we need to rethink these safety checks in general and have the HAL be less opinionated about it. While it's unsafe for GLES and probably worth tracking in some way, mipmaps can be uploaded directly (albeit, Impeller doesn't support this yet). One example is storing radiance maps for image based lighting.

PR #42352, opened 2023-05-26
[web] Use the new js_util.callMethod(Object, Object)

mdebbar:
Now that `js_util.callMethod` can take an `Object` as a method name, we can use it directly for methods that are accessed through a `Symbol`.


joshualitt (approved):

Hixie:

PR #42368, opened 2023-05-26
Remove legacy xcode properties/dependencies

keyonghan:
Now only $flutter/osx_sdk property is being used, and it's safe to remove deprecated entries from ci.yaml.

Part of https://github.com/flutter/flutter/issues/127534

### DIFF
--- a/.ci.yaml
+++ b/.ci.yaml
@@ -54,7 +54,6 @@ platform_properties:
       device_type: none
       cpu: x86
       os: Mac-12
-      xcode: 14e222b
       $flutter/osx_sdk : >-
         {
           "sdk_version": "14e222b"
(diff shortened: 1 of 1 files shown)


.ci.yaml
```diff
@@ -54,7 +54,6 @@ platform_properties:
       device_type: none
       cpu: x86
       os: Mac-12
-      xcode: 14e222b
       $flutter/osx_sdk : >-
         {
           "sdk_version": "14e222b"
```
yusuf-goog, line 59: Do you need the runtime versions listed anywhere?

yusuf-goog (approved):

.ci.yaml
keyonghan: No on the platform level. We have only one target needing runtimes. See `Mac Unopt` below.

PR #42386, opened 2023-05-28
[Impeller] Bias towards sampling from lower mip levels

bdero:
Resolves https://github.com/flutter/flutter/issues/127174.

Sets the same mip bias as Skia uses. Biases towards sampling from lower (larger) mip levels -- the tradeoff is a slightly higher cache miss rate in exchange for sharper/less blurry bilinear min-filtered images.

This makes scaled down images render identically between Impeller and Skia.

### Original image
image

### Impeller before
Screenshot 2023-05-25 at 11 08 02 PM

### Impeller after
Screenshot 2023-05-27 at 9 34 33 PM

### Skia
Screenshot 2023-05-25 at 11 11 10 PM


bdero:
I think this will be a pretty important quality improvement to track/retain, but it's a bit hard to capture with a sensible golden.

Maybe a good way to get a human interpretable golden for this would be to render a scaled down rasterized VG image to a second pass, and then scale it way up with nearest sampling. There's not really a way to do that through the Aiks surface so I'll need to backdoor something.

PR #42411, opened 2023-05-30
[Impeller] Add Impeller support in the embedder API

bdero:


dkwingsmt:
@loic-sharma Thanks! They look good to me (in terms of multi-view macOS).

flow/embedded_views.cc
```diff
@@ -42,8 +42,10 @@ bool DisplayListEmbedderViewSlice::recording_ended() {
   return builder_ == nullptr;
 }
 
-void ExternalViewEmbedder::SubmitFrame(GrDirectContext* context,
-                                       std::unique_ptr<SurfaceFrame> frame) {
+void ExternalViewEmbedder::SubmitFrame(
+    GrDirectContext* context,
```
chinmaygarde, line 46: Ooh, that's a bit unfortunate. But we can clean this up later.

shell/platform/embedder/embedder_surface_metal_impeller.mm
```diff
@@ -0,0 +1,100 @@
...
+}
+
+EmbedderSurfaceMetalImpeller::~EmbedderSurfaceMetalImpeller() = default;
+
+bool EmbedderSurfaceMetalImpeller::IsValid() const {
+  return valid_;
+}
+
+std::unique_ptr<Surface> EmbedderSurfaceMetalImpeller::CreateGPUSurface() API_AVAILABLE(ios(13.0)) {
+  if (@available(iOS 13.0, *)) {
```
chinmaygarde, line 63: I don't see the point of this availability guard. There is no call here that the compiler will complain about. And we support targets below 13.0.

shell/platform/embedder/embedder_external_view.cc
```diff
@@ -83,15 +85,28 @@ bool EmbedderExternalView::Render(const EmbedderRenderTarget& render_target) {
...
-  if (!surface) {
+  auto impeller_target = render_target.GetImpellerRenderTarget();
+  if (impeller_target.has_value()) {
+    auto aiks_context = render_target.GetAiksContext();
+
+    DisplayListBuilder dl_builder;
+    dl_builder.SetTransform(&surface_transformation_);
+    slice_->render_into(&dl_builder);
+
+    auto dispatcher = impeller::DlDispatcher();
```
chinmaygarde, line 96: Aha. I've been trying to making this the default in the engine but it seems so counter to convention. My reasoning is that `auto foo = FooType{};` makes it impossible to have foo uninitialized.

chinmaygarde: But the dl_builder above uses the other style? Just use the same style everywhere?

shell/platform/embedder/embedder_render_target.cc
```diff
@@ -4,24 +4,38 @@
...
   // TODO(38468): The optimization to elide backing store updates between frames
   // has not been implemented yet.
   backing_store_.did_update = true;
-  FML_DCHECK(render_surface_);
+
+  if (impeller_target_) {
+    FML_DCHECK(aiks_context_);
+    FML_DCHECK(!render_surface_);
+  } else {
+    FML_CHECK(render_surface_);
```
chinmaygarde, line 37: `DCHECK` turned to `CHECK`. Perhaps keep them the same?

shell/platform/embedder/BUILD.gn
```diff
@@ -122,11 +126,15 @@ template("embedder_source_set") {
         "embedder_external_texture_metal.mm",
         "embedder_surface_metal.h",
         "embedder_surface_metal.mm",
+        "embedder_surface_metal_impeller.h",
```
chinmaygarde, line 129: For testing, (in a later patch is fine) I suggest converting the `impeller::EmbedderTest` fixture to be a parameterized fixture with the impeller parameterization setting up a context object with the right flag set. Then in the setup, you can skip the test on non-supported backends. The updates should be minimal. Likely just patching the TEST_F calls and instantiating the suite. But you'd double your test coverage.

Just a thought. And we can attempt this in a follow up patch.

chinmaygarde (commented):
So cool! I left a few comments with nits and suggestions. But lgtm otherwise.

chinmaygarde (approved):

shell/platform/embedder/embedder_surface_metal_impeller.mm
bdero: Whoops, this is because I just duplicated the Skia variation and modified relevant parts. Removed.

shell/platform/embedder/embedder_external_view.cc
bdero: Done. I also prefer `auto foo = ...`, but tend to heavily bias towards whatever's nearby.

shell/platform/embedder/embedder_render_target.cc
bdero: Fixed.

shell/platform/embedder/embedder_surface_metal_impeller.mm
bdero: It turns out we need the check because CAMetalLayer is iOS 13+

chinmaygarde: The docs say 8.0. Perhaps its the iOS simulator?
Screenshot 2023-05-31 at 10 38 34 AM

bdero:
We're still not building Impeller with Fuchsia. I'm pulling on that thread now that the `AiksContext` is being lugged around everywhere.

shell/platform/embedder/embedder_surface_metal_impeller.mm
chinmaygarde: We should copy this sort of macro into formats_mtl and use it for the availability check. https://github.com/google/skia/blob/a8c3132457176d453c063247eaa6e2cab1cb700e/include/gpu/mtl/GrMtlTypes.h#L28

chinmaygarde: So:

```
#if TARGET_OS_SIMULATOR
#define IMPELLER_METAL_AVAILABLE @available(macos(10.11), ios(13.0))
#else  // TARGET_OS_SIMULATOR
#define IMPELLER_METAL_AVAILABLE @available(macos(10.11), ios(8.0))
#endif  // TARGET_OS_SIMULATOR
```

Have macros for `IMPELLER_MACOS_METAL_AVAILABLE`, `IMPELLER_IOS_METAL_AVAILABLE` set to the appropriate versions.

zanderso:
There's a big regression in backdrop filter perf on this PR: https://flutter-flutter-perf.skia.org/e/?queries=sub_result%3D90th_percentile_frame_rasterizer_time_millis%26sub_result%3D99th_percentile_frame_rasterizer_time_millis%26sub_result%3Daverage_frame_rasterizer_time_millis%26sub_result%3Dworst_frame_rasterizer_time_millis%26test%3Dpost_backdrop_filter_perf_ios__timeline_summary&queries=sub_result%3D90th_percentile_frame_rasterizer_time_millis%26sub_result%3D99th_percentile_frame_rasterizer_time_millis%26sub_result%3Daverage_frame_rasterizer_time_millis%26sub_result%3Dworst_frame_rasterizer_time_millis%26test%3Dbackdrop_filter_perf_ios__timeline_summary%26test%3Dpost_backdrop_filter_perf_ios__timeline_summary&selected=commit%3D35187%26name%3D%252Carch%253Dintel%252Cbranch%253Dmaster%252Cconfig%253Ddefault%252Cdevice_type%253DiPhone_11%252Cdevice_version%253Dnone%252Chost_type%253Dmac%252Csub_result%253Daverage_frame_rasterizer_time_millis%252Ctest%253Dbackdrop_filter_perf_ios__timeline_summary%252C

PR #42417, opened 2023-05-30
[Impeller] offload all text computation into vertex shader

jonahwilliams:
TextContents::Render occassionally shows up in the highest CPU functions. We can actually unload most of this computation into the vertex shader.


bdero (approved):

bdero:
LGTM. I still have a bit of PTSD from https://github.com/flutter/engine/pull/40912 TBH, but this makes sense. ;)

PR #42418, opened 2023-05-30
Reland: "Adding app lifecycle notification for macOS and Linux, add `hidden` state." (#40542)

gspencergoog:
## Description

This reverts commit 57e491a to re-land #40542 to fix the lint warning which caused the original PR to be reverted (the lint warnings were turned on after the original was landed, which is why it wasn't caught earlier).

### DIFF
--- a/shell/platform/darwin/macos/framework/Source/FlutterEngineTest.mm
+++ b/shell/platform/darwin/macos/framework/Source/FlutterEngineTest.mm
@@ -805,55 +805,75 @@ TEST_F(FlutterEngineTest, HandleAccessibilityEvent) {
   EXPECT_TRUE(announced);
 }
 
-TEST_F(FlutterEngineTest, RunWithEntrypointUpdatesDisplayConfig) {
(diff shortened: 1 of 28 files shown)


cbracken (approved):

shell/platform/darwin/macos/framework/Source/FlutterEngineTest.mm
```diff
@@ -805,55 +805,75 @@ - (nonnull NSView*)createWithViewIdentifier:(int64_t)viewId arguments:(nullable
   EXPECT_TRUE(announced);
 }
 
-TEST_F(FlutterEngineTest, RunWithEntrypointUpdatesDisplayConfig) {
```
goderbauer, line 808: Presumably, it wasn't the intention of this PR to remove this test and the next one? Probably just a bad merge?

gspencergoog: Huh.  No, definitely not intended.

PR #42421, opened 2023-05-30
[Impeller] Support tessellation of non-convex shapes via X/Y-Monotone polygons.

jonahwilliams:
Implements but does not use a tessellation routine that works for all x/y monotone polygon.

A polygon is l-monotone if a line L only intersects the points of the polygon at two locations, (or across one line segment). Many of the non-convex shapes in the gallery are X or Y monotone:

(This one is probably monotone, though I haven't exactly verified that).

image

These shapes have a much simpler tessellation algorithm, all we need to do is decompose them into two pieces (by identifying the largest/smallest vertices in the X/Y direction) and then performing a fairly constrained ear clip. This can be done from left to right in a single pass and overall is O(N).


jonahwilliams:
I don't think I'm going to get a chance to work on this. We should pursue more general solutions like StC

PR #42429, opened 2023-05-30
[Impeller] Fixed TypographerTest.MaybeHasOverlapping

gaaclarke:
fixes https://github.com/flutter/flutter/issues/127714

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/impeller/typographer/typographer_unittests.cc
+++ b/impeller/typographer/typographer_unittests.cc
@@ -9,6 +9,7 @@
 #include "impeller/typographer/lazy_glyph_atlas.h"
 #include "third_party/skia/include/core/SkBitmap.h"
 #include "third_party/skia/include/core/SkData.h"
+#include "third_party/skia/include/core/SkFontMgr.h"
 #include "third_party/skia/include/core/SkTextBlob.h"
 
 // TODO(zanderso): https://github.com/flutter/flutter/issues/127701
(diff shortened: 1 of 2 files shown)


jonahwilliams (approved):
RSLGTM

Why do we get away with this on iOS? relying on some internal state somewhere?

jonahwilliams (approved):
RSLGTM

Why do we get away with this on iOS? relying on some internal state somewhere?

gaaclarke:
> Why do we get away with this on iOS? relying on some internal state somewhere?

Well, at first I thought it was because the default font was different per operating system.  But I just looked and this still passes on macos with arial 12, where on linux I needed arial 0.5 to get them to overlap.

Here are the first 2 glyph rects:
```
(x:0.000000 y:-10.000000 w:6.000000 h:11.000000)
(x:5.673828 y:-10.000000 w:9.000000 h:11.000000)
```

I don't know why they overlap.  I could guess the ttf files are different or there is some ifdef code or skia is using different libraries for typesetting?  If you have an optimization that requires that the glyphs aren't overlapping, I don't think it's getting hit as much as you had hoped.

jonahwilliams:
Yeah I think its making some assumptions based on what the font will do. Perhaps instead we should hardcode the font?

Its entirely reasonable that they would overlap, and this test might just be poorly written to assume they wont

jonahwilliams:
but at any rate, this fix is still fine if using a different font stabilizes it

gaaclarke:
@jonahwilliams I tried switching the test to use a specific font:
```c++
TEST_P(TypographerTest, MaybeHasOverlapping) {
  static constexpr const char* kTestFontFixture = "Roboto-Regular.ttf";
  auto mapping = OpenFixtureAsSkData(kTestFontFixture);
  FML_CHECK(mapping);
  SkFont sk_font(SkTypeface::MakeFromData(mapping), 12.0);

  auto frame = TextFrameFromTextBlob(SkTextBlob::MakeFromString("1", sk_font));
  // Single character has no overlapping
  ASSERT_FALSE(frame.MaybeHasOverlapping());

  auto frame_2 =
      TextFrameFromTextBlob(SkTextBlob::MakeFromString("123456789", sk_font));
  // Characters probably have overlap due to low fidelity text metrics, but this
  // could be fixed.
  ASSERT_TRUE(frame_2.MaybeHasOverlapping());
}
```

This still passes on macos but fails on linux.

PR #42432, opened 2023-05-30
[web] Move announcement live elements to the end of the DOM and make them `div`s instead of `label`s.

marcianx:
- Moving them to the end prevents the screen reader from landing on them before the relevant content.
- Making them `div`s instead of `label`s prevents some screen readers (ChromeVox in particular) from landing on the live elements when the live elements are empty.

Fixes https://github.com/flutter/flutter/issues/127862.

## Pre-launch Checklist

- [X] I signed the [CLA].


lib/web_ui/lib/src/engine/embedder.dart
```diff
@@ -216,8 +218,12 @@ class FlutterViewEmbedder {
         .instance.semanticsHelper
         .prepareAccessibilityPlaceholder();
 
+    final DomElement announcementsElement = createDomElement('div');
```
yjbanov, line 221: Does it literally need to be `<div>`? For consistency with other top-level elements it would be good for it to have a custom name, such as `<flt-announcement-host>`.

yjbanov (approved):

lib/web_ui/lib/src/engine/embedder.dart
marcianx: Sure, updated per your suggestion and also updated the two live elements and confirmed that things still work fine on ChromeVox on our app.

yjbanov (approved):
lgtm

lib/web_ui/lib/src/engine/embedder.dart
```diff
@@ -216,8 +218,12 @@ class FlutterViewEmbedder {
         .instance.semanticsHelper
         .prepareAccessibilityPlaceholder();
 
+    final DomElement announcementsElement = createDomElement('ftl-announcement-host');
+    _accessibilityAnnouncements = AccessibilityAnnouncements(hostElement: announcementsElement);
+
     shadowRoot.append(accessibilityPlaceholder);
     shadowRoot.append(_sceneHostElement!);
+    shadowRoot.append(announcementsElement);
```
ditman, line 226: Do we need announcements on every possible view, or can we have a single global instance?

marcianx: 1. Where would we put that instance if it's not `flutterViewEmbedder`? Would that be a separate singleton object? In other words, would you like this to be re-singleton-ified? If so, then the question in the multi-view work is, what can be used to ensure that the live elements end up at/near the end of the DOM? (That being said placing it at the end, this might be slightly less critical after the two other fixes in this PR and the last.)
2. AFAIAA, there's also not a strict issue with having multiple live elements. Each live element is already implemented to support multiple announcements at once.

lib/web_ui/lib/src/engine/embedder.dart
```diff
@@ -216,8 +218,12 @@ class FlutterViewEmbedder {
         .instance.semanticsHelper
         .prepareAccessibilityPlaceholder();
 
+    final DomElement announcementsElement = createDomElement('ftl-announcement-host');
```
mdebbar, line 221:
```suggestion
    final DomElement announcementsElement = createDomElement('flt-announcement-host');
```

lib/web_ui/lib/src/engine/embedder.dart
ditman: If we can have multiple live elements with no problem, let's go with this, that's why it's LGTM'd :)

(Without wanting to derail this PR: since you said that this needs to be near the end of the DOM, we could put it as the "last child" of the "body" tag. And this wouldn't need to be a re-singletonized: instances would need to either create the DOM object in the DOM if it wasn't there already, or reuse the one created by other instance previously, since all they do is adding some elements to the DOM when asked to announce something.)

ditman:
(Restarted failing checks, they seemed to be an infra problem)

lib/web_ui/lib/src/engine/embedder.dart
marcianx: The challenge is that inserting the live element div at the end of the DOM only works if subsequently initialized DOM elements don't also append themselves to `document.body`, thereby ending up after the live element. Effectively, you'd still need some view-level abstraction that can be used to append more elements to a view without adding them after the accessibility DOM for reliability.

marcianx:
I've re-run the failing tests twice and they have consistently failed:

* Linux Framework Smoke Tests: "Infra Failure: Step('Framework test.test: Framework test') (canceled) (retcode: 0)"
* Linux Web Framework tests: "Empty summaryMarkdown"

Why is this PR showing these errors when some of the other PRs I sampled are not? Most of the others don't trigger "Linux Web Framework tests", but some do trigger "Linux Framework Smoke Tests" which seem to be passing for them.

mdebbar:
@marcianx you might be stuck on a bad commit or something. I rebased the PR, let's see if things go green now 🤞

mdebbar (approved):

PR #42438, opened 2023-05-31
Wait for GL command completion in the ExternalTextureGLRefreshedTooOften test

jason-simmons:
This test sometimes caused assertion failures when running on Windows with ANGLE.  The process may be unable to safely clean up global objects if GL commands are pending when the test exits.


loic-sharma (approved):
LGTM. Thank you for investigating this issue!

PR #42486, opened 2023-06-01
Reland "[web] Remove the JS API for url strategy (#42134)"

mdebbar:
Initially landed in https://github.com/flutter/engine/pull/42134
Then reverted in https://github.com/flutter/engine/pull/42468

It failed because we changed `UrlStrategy` to an `interface` which prevents "extending".

The only change in the reland is the removal of the `interface` keyword.


jonahwilliams (approved):
RSLGTM

PR #42493, opened 2023-06-01
MultiView changes for dart:ui

goderbauer:
Fixes https://github.com/flutter/flutter/issues/124991.

Framework tests are failing because I need to do a pre-migration of `flutter_test` for the Object -> int change of viewId. That's in https://github.com/flutter/flutter/pull/128092.

### DIFF
--- a/lib/ui/platform_dispatcher.dart
+++ b/lib/ui/platform_dispatcher.dart
@@ -2414,7 +2414,7 @@ class SemanticsActionEvent {
   /// Optional arguments for the action.
   final Object? arguments;
 
-  static const _noArgumentPlaceholder = Object();
+  static const Object _noArgumentPlaceholder = Object();
 
   /// Create a clone of the [SemanticsActionEvent] but with provided parameters
   /// replaced.
--- a/lib/ui/platform_dispatcher.dart
+++ b/lib/ui/platform_dispatcher.dart
@@ -33,7 +33,11 @@ typedef PointerDataPacketCallback = void Function(PointerDataPacket packet);
 typedef KeyDataCallback = bool Function(KeyData data);
 
 /// Signature for [PlatformDispatcher.onSemanticsAction].
+// TODO(goderbauer): Deprecate/remove this when the framework has migrated to SemanticsActionEventCallback.
 typedef SemanticsActionCallback = void Function(int nodeId, SemanticsAction action, ByteData? args);
+
+/// Signature for [PlatformDispatcher.onSemanticsActionEvent].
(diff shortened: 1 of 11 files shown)


lib/ui/pointer.dart
```diff
@@ -178,11 +179,15 @@ class PointerData {
     this.rotation = 0.0,
   });
 
-  /// Unique identifier that ties the [PointerEvent] to embedder event created it.
+  /// The ID of the [FlutterView] this [PointerEvent] originated from.
+  final int viewId;
+
+  /// Unique identifier that ties the [PointerEvent] to embedder event created
```
loic-sharma, line 185: Nit:
```suggestion
  /// Unique identifier that ties the [PointerEvent] to the embedder
  /// event that created it.
```

(My formatting may be off)

lib/ui/platform_dispatcher.dart
```diff
@@ -1158,6 +1168,22 @@ class PlatformDispatcher {
...
+  ///
+  /// This callback is used when the user expresses the action they wish to
+  /// perform based on the semantics node supplied by updateSemantics.
+  ///
+  /// The framework invokes this callback in the same zone in which the
+  /// callback was set.
+  SemanticsActionEventCallback? get onSemanticsActionEvent => _onSemanticsActionEvent;
+  SemanticsActionEventCallback? _onSemanticsActionEvent;
+  Zone _onSemanticsActionEventZone = Zone.root;
+  set onSemanticsActionEvent(SemanticsActionEventCallback? callback) {
```
loic-sharma, line 1182: Should we make it an error to set both `onSemanticsActionEvent` and `onSemanticsAction`?

loic-sharma, line 1182: From my understanding, the engine will not invoke the `SemanticsActionEventCallback` currently. Should we make that work as part of this pull request before making this API available to users?

goderbauer: I opted for both are getting called if you set both. Do you feel strongly about only allowing one at a time?

loic-sharma: Ah sorry, I can't read. I thought that'd be in `hooks.dart`, but your approach is good too. Marking as resolved.

loic-sharma: Gotcha, calling both sounds good! Marking as resolved

loic-sharma (approved):
LGTM! Thanks for doing this 🎉

goderbauer:
Another framework pre-migration for this change is landing in https://github.com/flutter/flutter/pull/128152.

PR #42497, opened 2023-06-01
Handle deprecation of Dart_TimelineEvent Embedder API

derekxu16:
This PR changes usages of `Dart_TimelineEvent` to `Dart_RecordTimelineEvent` as `Dart_TimelineEvent` was deprecated in https://dart-review.googlesource.com/c/sdk/+/308721.

### DIFF
--- a/shell/common/pipeline.h
+++ b/shell/common/pipeline.h
@@ -112,12 +112,14 @@ class Pipeline {
         std::function<PipelineProduceResult(ResourcePtr, size_t)>;
 
     Continuation continuation_;
-    size_t trace_id_;
+    uint64_t trace_id_;
 
-    ProducerContinuation(const Continuation& continuation, size_t trace_id)
+    ProducerContinuation(const Continuation& continuation, uint64_t trace_id)
         : continuation_(continuation), trace_id_(trace_id) {
+      TRACE_EVENT_ASYNC_BEGIN0_WITH_FLOW_IDS("flutter", "PipelineItem",
+                                             trace_id_, /*flow_id_count=*/1,
+                                             /*flow_ids=*/&trace_id);
       TRACE_FLOW_BEGIN("flutter", "PipelineItem", trace_id_);
-      TRACE_EVENT_ASYNC_BEGIN0("flutter", "PipelineItem", trace_id_);
       TRACE_EVENT_ASYNC_BEGIN0("flutter", "PipelineProduce", trace_id_);
     }
 
--- a/flow/frame_timings.h
+++ b/flow/frame_timings.h
@@ -14,9 +14,11 @@
 #include "flutter/fml/time/time_delta.h"
 #include "flutter/fml/time/time_point.h"
 
-#define TRACE_EVENT_WITH_FRAME_NUMBER(recorder, category_group, name) \
-  TRACE_EVENT1(category_group, name, "frame_number",                  \
-               recorder->GetFrameNumberTraceArg())
+#define TRACE_EVENT_WITH_FRAME_NUMBER(recorder, category_group, name,       \
+                                      flow_id_count, flow_ids)              \
+  TRACE_EVENT1_WITH_FLOW_IDS(category_group, name, flow_id_count, flow_ids, \
+                             "frame_number",                                \
+                             recorder->GetFrameNumberTraceArg())
 
 namespace flutter {
 
--- a/fml/trace_event.cc
+++ b/fml/trace_event.cc
@@ -30,6 +30,8 @@ std::atomic<TimelineMicrosSource> gTimelineMicrosSource = DefaultMicrosSource;
 inline void FlutterTimelineEvent(const char* label,
                                  int64_t timestamp0,
                                  int64_t timestamp1_or_async_id,
+                                 intptr_t flow_id_count,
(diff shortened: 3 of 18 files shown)


shell/common/pipeline.h
```diff
@@ -116,8 +116,10 @@ class Pipeline {
 
     ProducerContinuation(const Continuation& continuation, size_t trace_id)
         : continuation_(continuation), trace_id_(trace_id) {
+      TRACE_EVENT_ASYNC_BEGIN0_WITH_FLOW_IDS("flutter", "PipelineItem",
+                                             trace_id_, /*flow_id_count=*/1,
+                                             /*flow_ids=*/&trace_id_);
       TRACE_FLOW_BEGIN("flutter", "PipelineItem", trace_id_);
```
derekxu16, line 122: This doesn't do anything because flow events don't bind to async events in Chrome's trace format.

dnfield (changes requested):
This needs a test, and it looks like CI isn't happy.

What's a case where we'd ever want more than 1 flow ID? It looks like it's always either 1 or 0. Can we just make the API take a single parameter and assume the count is always 1 if it's not nullptr?

derekxu16:
> This needs a test, and it looks like CI isn't happy.

I haven't landed the Dart CL yet, because it will break Dart SDK into Flutter Engine rolls. I'd like to get this PR approved first, and then I'll be able to include these changes with the rolling of the embedder API changes. CI will only pass when these changes are included with the rolling of the embedder API changes. I'm not sure how to write a test for this, do you have an example I can follow?

> What's a case where we'd ever want more than 1 flow ID? It looks like it's always either 1 or 0. Can we just make the API take a single parameter and assume the count is always 1 if it's not nullptr?

`Animator` has events that are linked to multiple flow events
https://github.com/flutter/engine/blob/adca97c136528ca6697497b6a4f4da4a7a09530a/shell/common/animator.cc#L71

https://github.com/flutter/engine/blob/adca97c136528ca6697497b6a4f4da4a7a09530a/shell/common/animator.cc#L264

dnfield:
> I haven't landed the Dart CL yet, because it will break Dart SDK into Flutter Engine rolls

If we can in any way avoid the need for a manual roll we should. It makes things like reverts/relands much harder, and ends up making the release process more complicated. Can we avoid that here?

dnfield:
(Thanks for the pointer about multiple flow ids)

derekxu16:
> If we can in any way avoid the need for a manual roll we should. It makes things like reverts/relands much harder, and ends up making the release process more complicated. Can we avoid that here?

I think a manual roll has to be done here, because on the Dart side we are adding new required parameters to the `Dart_TimelineEvent` function, which will make all existing calls to `Dart_TimelineEvent` have the wrong number of arguments. We can't get around it by adding them as optional parameters, because `dart_tools_api.h` is only supposed to use C features (https://dart-review.googlesource.com/c/sdk/+/302963/comment/a70bff54_1920bed7/).

We can make the number of Engine changes included in the manual roll smaller though. There aren't many direct calls to `Dart_TimelineEvent`, so we can just pass `flow_id_count=0` and `flow_ids=nullptr` to those when doing the roll. Then we can update the `TRACE_EVENT_*` macros in a separate PR. Would that be any better?

dnfield:
Why can't we deprecate the old method and create a new one?

derekxu16:
I just discussed this with @bkonyi. We would prefer not to go through the extra deprecation / cleanup process since this function is only used in a couple of places in the Engine. Let us know if you still strongly desire to avoid the manual roll.

zanderso:
@derekxu16 @bkonyi It looks like there's a straightforward way to do a soft transition for this API, so that's what we should do. Manual rolls are for emergencies or cases where a soft transition is impossible.

derekxu16:
I'll be able to test this in https://github.com/flutter/engine/blob/main/testing/dart/observatory/tracing_test.dart once https://dart-review.googlesource.com/c/sdk/+/303022 lands. I need that CL to land because it will allow me to decode Perfetto-format trace protos and check that flow IDs are present in them. That CL is currently blocked for the reasons described here: https://dart-review.googlesource.com/c/sdk/+/303022/comment/d77751a4_f9e8fea8. I'll revist this PR once that CL lands.

derekxu16:
The Linux linux_unopt check is failing because it's detecting a leak. I think it's because `flutter_tester` is not cleaning up the VM properly. I added print statements to `third_party/dart/runtime/vm/dart.cc`, built the engine with `--no-prebuilt-dart-sdk`, ran the `--type=dart` tests, and never saw the string that I printed.

We have a test in the VM that passes flow ID arguments to `Dart_RecordTimelineEvent`, and ASAN doesn't detect leaks when running that test. https://github.com/dart-lang/sdk/blob/b64fc318e715081c3ce99ae3242e598886cb3ce2/runtime/vm/dart_api_impl_test.cc#L10416

dnfield:
(From offline discussion - we leak the VM by default, the tester should not need to leak the VM, setting it to not leak the VM makes ASAN happy)

dnfield (approved):
LGTM as long as bots are happy - I rekicked one that seemed like an infra failure.

PR #42504, opened 2023-06-02
[Impeller] Add an assert macro that can be used to satisfy the bugprone-unchecked-optional-access check

jason-simmons:
In clang-tidy the bugprone-unchecked-optional-access linter will recognize Chromium's version of CHECK but does not know about Flutter's FML_CHECK macros.  This macro can be used to tell the linter that a std::optional is known to contain a value in a particular code path.

### DIFF
--- a/fml/logging.h
+++ b/fml/logging.h
@@ -95,4 +95,10 @@ bool ShouldCreateLogMessage(LogSeverity severity);
     ::fml::KillProcess();                          \
   }
 
+// Asserts that a std::optional contains a value.  This can be used to satisfy
+// the clang-tidy bugprone-unchecked-optional-access check.
+#define ASSERT_OPTIONAL_HAS_VALUE(opt) \
+  if (!opt.has_value())                \
+    FML_UNREACHABLE();
+
 #endif  // FLUTTER_FML_LOGGING_H_
--- a/impeller/renderer/backend/metal/surface_mtl.mm
+++ b/impeller/renderer/backend/metal/surface_mtl.mm
@@ -50,6 +50,7 @@ static std::optional<RenderTarget> WrapTextureWithRenderTarget(
   // rendering but also creates smaller intermediate passes.
   ISize root_size;
   if (requires_blit) {
+    ASSERT_OPTIONAL_HAS_VALUE(clip_rect);
     root_size = ISize(clip_rect->size.width, clip_rect->size.height);
   } else {
     root_size = {static_cast<ISize::Type>(texture.width),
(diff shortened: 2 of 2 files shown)


fml/logging.h
```diff
@@ -95,4 +95,10 @@ bool ShouldCreateLogMessage(LogSeverity severity);
     ::fml::KillProcess();                          \
   }
 
+// Asserts that a std::optional contains a value.  This can be used to satisfy
+// the clang-tidy bugprone-unchecked-optional-access check.
+#define ASSERT_OPTIONAL_HAS_VALUE(opt) \
```
zanderso, line 100: Cool. This looks useful.

impeller/renderer/backend/metal/surface_mtl.mm
```diff
@@ -50,6 +50,7 @@
   // rendering but also creates smaller intermediate passes.
   ISize root_size;
   if (requires_blit) {
+    ASSERT_OPTIONAL_HAS_VALUE(clip_rect);
```
zanderso, line 53: This is fine if there's unit test coverage that will cause the assert to trigger if someone violates the assumptions that make the assert correct today. Otherwise we're no better off than we were when we didn't have the clang-tidy lint running.

If the test that covers the assumptions is at a different level, then should this instead be this?
```C++
if (!clipt_rect.has_value()) {
  VALIDATION_LOG << "...";
  return std::nullopt;
}
```

That way this code can be defensive against someone calling `WrapTextureWithRenderTarget` incorrectly by accident.

impeller/renderer/backend/metal/surface_mtl.mm
```diff
@@ -243,6 +244,7 @@ static bool ShouldWaitForCommandBuffer() {
       return false;
     }
     auto blit_pass = blit_command_buffer->CreateBlitPass();
+    ASSERT_OPTIONAL_HAS_VALUE(clip_rect_);
```
zanderso, line 247: Same question. Should this be like:
```C++
if (!clip_rect_.has_value()) {
  return false;
}
```

impeller/renderer/backend/metal/surface_mtl.mm
jason-simmons: If `requires_blit` is set, then `clip_rect` must always be set and there is no fallback path.

The assert macro using `FML_UNREACHABLE` is intended to represent that the `!has_value()` case should never happen here.

bdero: I'm somewhat biased towards returning an empty result in these particular cases.

On the one hand, we've heavily biased towards returning null resources/error flags for clear "this should never happen" cases across render routines, which normally results in a frame getting skipped. That's actually saved us from needlessly crashing in some cases where a user's app might just get a janky animation due to some skipped frames that are triggering an unexpected case.
On the other hand, it's often quite annoying to find these sources of error, and an explicit crash would be quicker to pinpoint. However, that issue is mostly solved through validation logging, which is largely missing here in `SurfaceMTL` at the moment.

PR #42520, opened 2023-06-02
Generate a unique temporary directory name in the iOS scenario test script

jason-simmons:
This ensures that a retry of this script on CI will not collide with the output directory of a previous run.


godofredoc (approved):

zanderso (commented):
lgtm

Is there any analogous change to make in https://github.com/flutter/engine/blob/main/testing/scenario_app/run_android_tests.sh ?

jason-simmons:
AFAICT there isn't any use of temporary paths in the Android scenario test script

PR #42523, opened 2023-06-02
[macOS] Top-left origin for PlatformView container

cbracken:
For consistency with Flutter (and all other platforms), Flutter views in the macOS embedder set `isFlipped` to ensure a co-ordinate system with the origin in the top-left, with y co-ordinates increasing towards the bottom edge of the view.

Previously, we were using a stock NSView as the container, which uses a bottom-left origin by default. Instead we extract the PlatformView container view as its own class with `isFlipped` true.

This doesn't correct the issue of the view anchorpoint/position but does correct rotation direction.

This also applies the transform back to origin prior to other transforms when adjusting the platformview position rather than after.

Issue: https://github.com/flutter/flutter/issues/124490

## Pre-launch Checklist

- [X] I signed the [CLA].

### DIFF
--- a/shell/platform/darwin/macos/framework/Source/FlutterMutatorView.mm
+++ b/shell/platform/darwin/macos/framework/Source/FlutterMutatorView.mm
@@ -34,7 +34,8 @@
 
 - (BOOL)isFlipped {
   // Flutter transforms assume a coordinate system with an upper-left corner origin, with y
-  // coordinate values increasing downwards.
+  // coordinate values increasing downwards. This affects the view, view transforms, and
+  // sublayerTransforms.
   return YES;
 }
 
@@ -58,7 +59,8 @@
 
 - (BOOL)isFlipped {
   // Flutter transforms assume a coordinate system with an upper-left corner origin, with y
-  // coordinate values increasing downwards.
+  // coordinate values increasing downwards. This affects the view, view transforms, and
+  // sublayerTransforms.
   return YES;
 }
 
--- a/shell/platform/darwin/macos/framework/Source/FlutterMutatorView.mm
+++ b/shell/platform/darwin/macos/framework/Source/FlutterMutatorView.mm
@@ -12,8 +12,8 @@
 #include "flutter/shell/platform/embedder/embedder.h"
 
 @interface FlutterMutatorView () {
-  /// Each of these views clips to a CGPathRef. These views, if present,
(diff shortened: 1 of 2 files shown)


shell/platform/darwin/macos/framework/Source/FlutterMutatorView.mm
```diff
@@ -26,6 +26,20 @@ @interface FlutterMutatorView () {
 
 @end
 
+/// Superview container for platform views, to which sublayer transforms are applied.
+@interface FlutterPlatformViewContainer : NSView
+@end
+
+@implementation FlutterPlatformViewContainer
+
+- (BOOL)isFlipped {
```
gspencergoog, line 35: Maybe `isYFlipped` or `isVerticallyFlipped`? Someday there might be a system where the origin is in the lower right...

gspencergoog (approved):

shell/platform/darwin/macos/framework/Source/FlutterMutatorView.mm
```diff
@@ -26,6 +26,20 @@ @interface FlutterMutatorView () {
...
+  // Flutter transforms assume a coordinate system with an upper-left corner origin, with y
+  // coordinate values increasing downwards.
```
knopp, line 37: Maybe mention that the value also affects the layer.sublayerTransform rotation direction. Even though it kinda makes sense, I haven't found it in any AppKit/QuartzCore documentation.

shell/platform/darwin/macos/framework/Source/FlutterMutatorView.mm
knopp: This is Cocoa method that needs to be overriden to switch into a sane coordinate system...

knopp (approved):

shell/platform/darwin/macos/framework/Source/FlutterMutatorView.mm
cbracken: Done. Good call. Mentioned it affects view coords, transforms, and sublayerTransforms.

shell/platform/darwin/macos/framework/Source/FlutterMutatorView.mm
knopp: > Someday there might be a system where the origin is in the lower right...

Please don't give them any suggestions...

PR #42525, opened 2023-06-02
Revert "Reland "add non-rendering operation culling to DisplayListBuilder" (#41463)"

flar:
Reverts flutter/engine#42330

Some golden failures down the line were discovered in internal testing. See b/285539451


flar:
Thanks to some help from @jason-simmons the root cause was found. The test code was rendering a "backwards rectangle"  (i.e. Left > Right or Top > Bottom) which has been an undocumented feature of the ui.Canvas methods for some time. The DisplayList-driven engine code would output the right thing because both Skia and Impeller will render such rectangles sorted, but the internal bounds code in DisplayList would accumulate them incorrectly as it assumed that such rectangles were "empty". See https://github.com/flutter/flutter/issues/128159

PR #42552, opened 2023-06-04
Raster cache should preserve RTree for overlay layers

knopp:
Fixes https://github.com/flutter/flutter/issues/116069

When raster cache paints to a layer over platform view, it will draw the individual rectangles from the content RTree, to ensure that the destination layer RTree is not clobbered. This is necessary for raster cache to not break unobstructed platform views and/or hit testing.

When raster cache paints in the root (bottom-most) layer the behavior it just copies the entire image as it did before. This is to minimise risk of performance regressions. 

## Pre-launch Checklist

- [X] I signed the [CLA].

### DIFF
--- a/flow/layers/display_list_layer_unittests.cc
+++ b/flow/layers/display_list_layer_unittests.cc
@@ -318,8 +318,8 @@ TEST_F(DisplayListLayerTest, RasterCachePreservesRTree) {
                               &paint_context(), false);
 
   DisplayListBuilder expected_root_canvas(true);
-  context->raster_cache->Draw(display_list_layer->caching_key_id(),
-                              expected_root_canvas, nullptr, true);
+  ASSERT_TRUE(context->raster_cache->Draw(display_list_layer->caching_key_id(),
+                                          expected_root_canvas, nullptr, true));
   auto root_canvas_dl = expected_root_canvas.Build();
   const auto root_canvas_rects =
       root_canvas_dl->rtree()->searchAndConsolidateRects(kGiantRect, true);
@@ -329,8 +329,9 @@ TEST_F(DisplayListLayerTest, RasterCachePreservesRTree) {
   EXPECT_EQ(root_canvas_rects_expected, root_canvas_rects);
 
   DisplayListBuilder expected_overlay_canvas(true);
-  context->raster_cache->Draw(display_list_layer->caching_key_id(),
-                              expected_overlay_canvas, nullptr, false);
+  ASSERT_TRUE(context->raster_cache->Draw(display_list_layer->caching_key_id(),
+                                          expected_overlay_canvas, nullptr,
+                                          false));
   auto overlay_canvas_dl = expected_overlay_canvas.Build();
   const auto overlay_canvas_rects =
       overlay_canvas_dl->rtree()->searchAndConsolidateRects(kGiantRect, true);
--- a/flow/layers/display_list_layer_unittests.cc
+++ b/flow/layers/display_list_layer_unittests.cc
@@ -319,7 +319,8 @@ TEST_F(DisplayListLayerTest, RasterCachePreservesRTree) {
 
   DisplayListBuilder expected_root_canvas(true);
   ASSERT_TRUE(context->raster_cache->Draw(display_list_layer->caching_key_id(),
-                                          expected_root_canvas, nullptr, true));
+                                          expected_root_canvas, nullptr,
+                                          false));
   auto root_canvas_dl = expected_root_canvas.Build();
   const auto root_canvas_rects =
       root_canvas_dl->rtree()->searchAndConsolidateRects(kGiantRect, true);
@@ -331,7 +332,7 @@ TEST_F(DisplayListLayerTest, RasterCachePreservesRTree) {
   DisplayListBuilder expected_overlay_canvas(true);
   ASSERT_TRUE(context->raster_cache->Draw(display_list_layer->caching_key_id(),
                                           expected_overlay_canvas, nullptr,
-                                          false));
+                                          true));
   auto overlay_canvas_dl = expected_overlay_canvas.Build();
   const auto overlay_canvas_rects =
       overlay_canvas_dl->rtree()->searchAndConsolidateRects(kGiantRect, true);
--- a/flow/layers/display_list_layer_unittests.cc
+++ b/flow/layers/display_list_layer_unittests.cc
@@ -307,17 +307,22 @@ TEST_F(DisplayListLayerTest, RasterCachePreservesRTree) {
   use_skia_raster_cache();
 
   auto context = preroll_context();
-  display_list_layer->Preroll(preroll_context());
-  EXPECT_EQ(context->renderable_state_flags, 0);
+  {
+    auto mutator = context->state_stack.save();
+    mutator.transform(SkMatrix::Scale(2.0, 2.0));
+    display_list_layer->Preroll(preroll_context());
+    EXPECT_EQ(context->renderable_state_flags, 0);
 
-  // Pump the DisplayListLayer until it is ready to cache its DL
-  display_list_layer->Preroll(preroll_context());
-  display_list_layer->Preroll(preroll_context());
-  display_list_layer->Preroll(preroll_context());
-  LayerTree::TryToRasterCache(*preroll_context()->raster_cached_entries,
-                              &paint_context(), false);
+    // Pump the DisplayListLayer until it is ready to cache its DL
+    display_list_layer->Preroll(preroll_context());
+    display_list_layer->Preroll(preroll_context());
+    display_list_layer->Preroll(preroll_context());
+    LayerTree::TryToRasterCache(*preroll_context()->raster_cached_entries,
+                                &paint_context(), false);
+  }
 
   DisplayListBuilder expected_root_canvas(true);
+  expected_root_canvas.Scale(2.0, 2.0);
   ASSERT_TRUE(context->raster_cache->Draw(display_list_layer->caching_key_id(),
                                           expected_root_canvas, nullptr,
                                           false));
@@ -325,11 +330,12 @@ TEST_F(DisplayListLayerTest, RasterCachePreservesRTree) {
   const auto root_canvas_rects =
       root_canvas_dl->rtree()->searchAndConsolidateRects(kGiantRect, true);
   std::list<SkRect> root_canvas_rects_expected = {
-      SkRect::MakeLTRB(13, 13, 28, 28),
+      SkRect::MakeLTRB(26, 26, 56, 56),
   };
   EXPECT_EQ(root_canvas_rects_expected, root_canvas_rects);
 
   DisplayListBuilder expected_overlay_canvas(true);
+  expected_overlay_canvas.Scale(2.0, 2.0);
   ASSERT_TRUE(context->raster_cache->Draw(display_list_layer->caching_key_id(),
                                           expected_overlay_canvas, nullptr,
                                           true));
@@ -339,9 +345,9 @@ TEST_F(DisplayListLayerTest, RasterCachePreservesRTree) {
 
   // Same bounds as root canvas, but preserves individual rects.
   std::list<SkRect> overlay_canvas_rects_expected = {
-      SkRect::MakeLTRB(13, 13, 23, 18),
-      SkRect::MakeLTRB(13, 18, 28, 23),
-      SkRect::MakeLTRB(18, 23, 28, 28),
+      SkRect::MakeLTRB(26, 26, 46, 36),
+      SkRect::MakeLTRB(26, 36, 56, 46),
+      SkRect::MakeLTRB(36, 46, 56, 56),
   };
   EXPECT_EQ(overlay_canvas_rects_expected, overlay_canvas_rects);
 };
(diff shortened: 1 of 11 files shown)


knopp:
cc @cyanglaz as this potentially affects unobstructed platform views on iOS.

flar:
Layer trees should never be rendered from cache with a PlatformView present (see LayerRasterCacheItem::PrerollFinalize). DisplayList layers will already have an rtree. Does that simplify the implementation here?

flow/raster_cache.cc
```diff
@@ -73,8 +93,12 @@ std::unique_ptr<RasterCacheResult> RasterCache::Rasterize(
...
   }
 
-  DlSkCanvasAdapter canvas(surface->getCanvas());
-  canvas.Clear(DlColor::kTransparent());
+  // Clear the surface canvas directly to avoid RTree storing a giant clear
+  // rect.
+  auto surfaceCanvas = surface->getCanvas();
+  surfaceCanvas->clear(SK_ColorTRANSPARENT);
+
+  DisplayListBuilder canvas(true);
```
flar, line 101: Could the RTree just be passed in from the caller? It should only be needed with DL cache items and they should already have an RTree in their DL.

flow/raster_cache.cc
```diff
@@ -39,8 +46,21 @@ void RasterCacheResult::draw(DlCanvas& canvas, const DlPaint* paint) const {
...
-  canvas.DrawImage(image_, {bounds.fLeft, bounds.fTop},
-                   DlImageSampling::kNearestNeighbor, paint);
+  if (is_root_canvas || !rtree_) {
+    canvas.DrawImage(image_, {bounds.fLeft, bounds.fTop},
+                     DlImageSampling::kNearestNeighbor, paint);
+  } else {
+    // On some platforms RTree from overlay layers is used for unobstructed
+    // platform views and hit testing. To preserve the RTree raster cache must
+    // paint individual rects instead of the whole image.
+    auto rects = rtree_->searchAndConsolidateRects(kGiantRect);
```
flar, line 56: It would be nice if we could quickly and easily estimate the density of the RTree and then just go through the single DrawImage above. Some (not sure of percentage) cache items will have a bg fill at the start.

flow/raster_cache.cc
knopp: Good point. It looks like the `DisplayList` of `DisplayListLayer` is always created with an `RTree`. We shouldn't duplicate this process unnecessarily.

flow/raster_cache.cc
knopp: Maybe compare overall bounds (all RTree rectangles joined together) with area of largest rect? That would be pretty good indicator of a fill. It would require one pass through all rectangles (or gather this when constructing the RTree). But it should be able to detect fill.

flar: But you'd have to account for overdraw? The Region of the RTree might make a better source for that data.

Which reminds me, I was going to check if the RTree cached its Region or not.

knopp: I'm not entirely sure what we're trying to save here. If the region is rectangular it will only yield one rectangle covering the entire area anyway.

knopp: `RTree` doesn't cache the `Region`. There might be different region built depending on the query rect though so if we want to cache it we might need to do it per query rect.

flar: We could make a Region for the whole RTree before intersecting with the rect, though. We'd then need a Region::IntersectRect method, but that could run faster than the existing searchAndConsolidate. Could that amortize the Region work?

Food for thought...

flow/layers/display_list_layer_unittests.cc
```diff
@@ -294,6 +294,56 @@ TEST_F(DisplayListLayerTest, CachedIncompatibleDisplayListOpacityInheritance) {
...
+
+  // Pump the DisplayListLayer until it is ready to cache its DL
+  display_list_layer->Preroll(preroll_context());
+  display_list_layer->Preroll(preroll_context());
+  display_list_layer->Preroll(preroll_context());
+  LayerTree::TryToRasterCache(*preroll_context()->raster_cached_entries,
+                              &paint_context(), false);
+
+  DisplayListBuilder expected_root_canvas(true);
+  context->raster_cache->Draw(display_list_layer->caching_key_id(),
```
flar, line 321: Doesn't this return a boolean that it succeeded? You should ASSERT_EQ on that, otherwise the rest of the test is invalid.

flar: Same below.

flow/raster_cache.cc
```diff
@@ -39,8 +46,29 @@ void RasterCacheResult::draw(DlCanvas& canvas, const DlPaint* paint) const {
...
+
+    // Bounds of original display list.
+    SkRect rtree_bounds = SkRect::MakeEmpty();
```
flar, line 59: rtree should probably have an overall bounds which would make this unnecessary. It can be retrieved for free from the top-level node which bounds the entire tree.

flow/raster_cache.cc
```diff
@@ -39,8 +46,29 @@ void RasterCacheResult::draw(DlCanvas& canvas, const DlPaint* paint) const {
...
+    for (auto rect : rects) {
+      rtree_bounds.join(rect);
+    }
+    for (auto rect : rects) {
+      SkRect src(rect);
+      src.offset(-rtree_bounds.fLeft, -rtree_bounds.fTop);
```
flar, line 65: Could this just be a canvas.translate outside of the loop?

flow/raster_cache.cc
```diff
@@ -39,8 +46,29 @@ void RasterCacheResult::draw(DlCanvas& canvas, const DlPaint* paint) const {
...
+      SkRect dst(src);
+      dst.offset(bounds.fLeft, bounds.fTop);
```
flar, line 67: Or `dst = src.makeOffset()`

flow/raster_cache.cc
flar: If not, then same comment as below - `src = rect.makeOffset()`

flow/raster_cache.cc
```diff
@@ -84,19 +114,21 @@ std::unique_ptr<RasterCacheResult> RasterCache::Rasterize(
   }
 
   auto image = DlImage::Make(surface->makeImageSnapshot());
-  return std::make_unique<RasterCacheResult>(image, context.logical_rect,
-                                             context.flow_type);
+  return std::make_unique<RasterCacheResult>(
```
flar, line 117: I just noticed that the checkerboard will not contribute to the rtree. I'm not sure if we care, but if we do then we can just `rtree = nullptr` when there is a checkerboard.

flow/raster_cache.h
```diff
@@ -143,9 +148,12 @@ class RasterCache {
   // if the item was disabled due to conditions discovered during |Preroll|
   // or if the attempt to populate the entry failed due to bounds overflow
   // conditions.
+  // If |is_root_canvas| is false, the raster cache will preserve the original
+  // RTree of cached content.
   bool Draw(const RasterCacheKeyID& id,
             DlCanvas& canvas,
-            const DlPaint* paint) const;
+            const DlPaint* paint,
+            bool is_root_canvas = true) const;
```
flar, line 156: Perhaps at this level the flag should be named something more specific to its duties. Whether or not there is a root canvas in the system is outside of the context of these classes. So, while `is_root_canvas` makes sense at the layer painting level, here it should be something like `use_rtree` or `preserve_rtree` (which would invert its value) or `ignore_rtree` (which would have the same value as "is root", but feels more awkward of a name). These classes don't care if there is anything about a root canvas, they just want to know if they should use the supplied rtree or not...

The other flag used in the layers could be named something like `rendering_to_platform_view` to make it compatible with `preserve_rtree` here without inverting it?

flar (commented):
Nothing really wrong in this review, just a couple of options for efficiency and a naming issue.

I have no idea if we should care about checkerboards in this stuff, so I'll leave the answer to that one to you. I will point out that checkerboards are a debugging tool, so performance isn't an issue, but would a developer want to have them on, but still have more precise interaction with their performance overlays?

flow/raster_cache.cc
knopp: Getting a subregion by intersecting region with a rect is likely quite a lot cheaper than constructing region from rectangles. I created a new issue here since `DlRegion::MakeIntersection` is out of scope of this PR, though a version of it is implemented here: https://github.com/flutter/engine/pull/42620

For this though it would be best to have an intersection of region and rectangle, instead of two regions.

flow/raster_cache.cc
flar: Perhaps having the checkerboard be sliced up by the rtree would provide an important debugging aid so developers can see exactly which parts of the overlay are being imposed upon? It might look odd, but it would provide information that isn't normally visible.

On the other hand, perhaps we might consider adding an overlay checkerboard feature specific to how the platform view slices are divided and recombined in the stack of platform views? Note that the rendering of the cached DL over the platform view might not represent the actual real estate consumed because the embedders often combine the rtree results into fewer larger rectangles, so slicing it up here doesn't give an accurate picture of how much real estate is being coopted...

flow/raster_cache.cc
knopp: I added RTree::bounds() implementation: 
```cc
const SkRect& DlRTree::bounds() const {
  if (!nodes_.empty()) {
    return nodes_.back().bounds;
  } else {
    return empty_;
  }
}
``` 

This should be correct, right? Unless I'm misunderstanding how the R-Tree works.

flow/raster_cache.cc
knopp: I'm leaning towards leaving this as it is. I have bad feeling about changing the behavior (disabling the R-Tree) upon a debug flag. It this turns out problematic we can reconsider.

flow/raster_cache.cc
flar: Absolutely!

flow/raster_cache.cc
flar: Note that we flush the cache when the flag changes, so this would only be modified while they have the mechanism enabled. But I agree that nothing needs to be done here and we can reevaluate our debugging tools later.

flow/raster_cache.h
knopp: I like `preserve_rtree`. I think the flag in `PaintContext` could be `is_overlay_layer` or `rendering_above_platform_view`.

knopp: Went with `rendering_above_platform_view`. To me it communicates the z-index clearer, but it's probably subjective.

flow/raster_cache.cc
```diff
@@ -39,8 +46,24 @@ void RasterCacheResult::draw(DlCanvas& canvas, const DlPaint* paint) const {
...
+    // On some platforms RTree from overlay layers is used for unobstructed
+    // platform views and hit testing. To preserve the RTree raster cache must
+    // paint individual rects instead of the whole image.
+    auto rects = rtree_->searchAndConsolidateRects(kGiantRect);
+
+    canvas.Translate(bounds.fLeft, bounds.fTop);
+
+    SkRect rtree_bounds = rtree_->bounds();
+    for (auto rect : rects) {
+      rect.offset(-rtree_bounds.fLeft, -rtree_bounds.fTop);
```
flar, line 62: This could also be factored into the translate above.

knopp: This affects the `src` position (where we copy from the image), so I don't quite see how.

flow/raster_cache.h
knopp: My bad. Pre-push check failed and I didn't notice that the push didn't get through.

knopp:
I rebased my other branch on top of this and it seems to have regressed. Setting to WIP until I figure out why.

knopp:
Fixed by properly transforming local r-tree coordinates to screen space. I also updated the test to use a non-identity transform to catch this.

flow/raster_cache.h
```diff
@@ -148,12 +148,15 @@ class RasterCache {
   // if the item was disabled due to conditions discovered during |Preroll|
   // or if the attempt to populate the entry failed due to bounds overflow
   // conditions.
-  // If |is_root_canvas| is false, the raster cache will preserve the original
-  // RTree of cached content.
+  // If |preserve_rtree| is true, the raster cache will preserve the original
+  // RTree of cached content by blitting individual rectangles from the cached
+  // image of the canvas according to the original layer R-Tree (if present).
```
flar, line 153: "of the canvas" doesn't seem necessary and I'm not sure it really describes the DL cache items very well...?

flow/raster_cache.h
```diff
@@ -148,12 +148,15 @@ class RasterCache {
...
+  // This is to ensure that the target surface R-Tree with be cloberred with
```
flar, line 154: "will be clobbered" - missing a "not"? Also bad spelling on "clobbered"

flar (approved):
Looks good to go except for some typos in one of the comments.

flow/raster_cache.h
knopp: That should have been *from the cached image **to** the canvas*. Canvas referring to the destination canvas.

PR #42567, opened 2023-06-05
[Impeller] Make conical gradient work as expected

ColdPaleLight:
fix https://github.com/flutter/flutter/issues/128012

In many instances (https://github.com/flutter/flutter/issues/128012#issuecomment-1577029031), the current implementation of conical gradient in impeller produces incorrect results. This pull request proposes to migrate the conical gradient algorithm from skia to impeller.

Please see https://github.com/google/skia/blob/ddf987d2ab3314ee0e80ac1ae7dbffb44a87d394/src/sksl/sksl_graphite_frag.sksl#L541-L666

Test result
FGcZZRCQZJ

## Pre-launch Checklist

- [ ] I signed the [CLA].

### DIFF
--- a/impeller/compiler/shader_lib/impeller/gradient.glsl
+++ b/impeller/compiler/shader_lib/impeller/gradient.glsl
@@ -7,27 +7,124 @@
 
 #include <impeller/texture.glsl>
 
+mat3 IPMapToUnitX(vec2 p0, vec2 p1) {
+  // Returns a matrix that maps [p0, p1] to [(0, 0), (1, 0)]. Results are
+  // undefined if p0 = p1.
+  return mat3(0.0, -1.0, 0.0, 1.0, 0.0, 0.0, 0.0, 0.0, 1.0) *
+         inverse(mat3(p1.y - p0.y, p0.x - p1.x, 0.0, p1.x - p0.x, p1.y - p0.y,
+                      0.0, p0.x, p0.y, 1.0));
+}
+
 /// Compute the t value for a conical gradient at point `p` between the 2
-/// circles defined by (c0, r0) and (c1, r1).
-///
-/// This assumes that c0 != c1.
-float IPComputeConicalT(vec2 c0, float r0, vec2 c1, float r1, vec2 p) {
-  float w = 1.0;
-  float result = 0.0;
-  vec2 ab = c1 - c0;
-  float dr = r1 - r0;
-  // Set sample rate to a minimum for the case where c0 and c1 are close.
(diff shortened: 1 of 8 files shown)


impeller/compiler/shader_lib/impeller/gradient.glsl
```diff
@@ -7,27 +7,124 @@
...
-  while (w >= 0.0) {
-    vec2 cw = w * ab + c0;
-    float rw = w * dr + r0;
-    if (length(p - cw) <= rw) {
-      result = w;
-      break;
+/// circles defined by (c0, r0) and (c1, r1). The returned vec2 encapsulates 't'
+/// as its x component and validity status as its y component, with positive y
+/// indicating a valid result.
+vec2 IPComputeConicalT(vec2 c0, float r0, vec2 c1, float r1, vec2 pos) {
```
gaaclarke, line 22: You should add a comment that points to the graphite implementation.  If there are upstream bugfixes we'll want to know where to grab them.

impeller/entity/contents/conical_gradient_contents.cc
```diff
@@ -51,18 +51,6 @@ void ConicalGradientContents::SetFocus(std::optional<Point> focus,
   focus_radius_ = radius;
 }
 
-bool ConicalGradientContents::IsOpaque() const {
```
gaaclarke, line 54: Was this removed for the cases where the gradient doesn't occupy the full square?  That wasn't the crux of the reported issue, right?  I'm not sure if it's worth detecting those scenarios for IsOpaque.

gaaclarke (changes requested):
Nice find!  The graphite code is so much easier to migrate.  I spent so much time reading the math papers (which I swear are wrong) and reading through the skia code.

The code looks good.  Just needs a comment linking the source of the code also I need confirmation that it's fine to opt out of IsOpaque since we could implement it if we want.

impeller/entity/contents/conical_gradient_contents.cc
gaaclarke: @bdero are you fine with us opting out of IsOpaque?  I think we could calculate this by expanding on what we previously had if it's worth it.

bdero: Yeah this is just used to optimize the blend mode for HSR and prevent unnecessary overdraw, so it's safe to just remove for now if it's wrong.

gaaclarke (approved):
LGTM modulo comment linking skia code.

PR #42568, opened 2023-06-05
Revert Dart SDK to 3.1.0-166.0.dev

zanderso:
Rolls to the Framework are failing repeatedly in tool integration tests since the attempt at https://github.com/flutter/flutter/pull/128168


bkonyi (approved):

aam:
More than likely the roll needs https://github.com/flutter/flutter/pull/128084, which have matching changes to the list of expression evaluation request arguments

aam:
> > More than likely the roll needs [flutter/flutter#128084](https://github.com/flutter/flutter/pull/128084), which have matching changes to the list of expression evaluation request arguments
> 
> I don't have any concrete suggestions about how to go about it, but it generally causes a lot less disruption to do soft-breaking changes when possible. Just a wild guess, but it looks like in this case, the service protocol could maybe have been made robust against the new parameters being missing?

My intent was to include quoted PR into the roll.
The change is to the text-based interface between frontend server and flutter tools - there is no soft-breaking I can see that can be done (other than to introduce alternative `compile-expression` command, which seems an overkill, given that we do have ability to patch manual rolls)

zanderso:
> My intent was to include quoted PR into the roll. The change is to the text-based interface between frontend server and flutter tools - there is no soft-breaking I can see that can be done (other than to introduce alternative `compile-expression` command, which seems an overkill, given that we do have ability to patch manual rolls)

We do have the ability to patch rolls, but my view is that that is more for emergencies rather than our regular way of doing things. When the rollers get blocked up, it has a bunch of bad effects downstream, so we should be trying really hard to avoid that.

aam:
https://github.com/flutter/flutter/pull/128255 with engine roll and patch to expression evaluation invocation(https://github.com/flutter/flutter/pull/128084)  passes the tests.

PR #42576, opened 2023-06-05
Ensure Dart roll script picks up udpates to dart-sdk/tools

zanderso:
The build of the Dart SDK depends on this package, so it needs to be kept up to date by https://github.com/flutter/buildroot/blob/master/tools/dart/create_updated_flutter_deps.py, which is run by the autoroller.


zanderso:
The absence of autorolling this package is one of the causes of https://github.com/flutter/flutter/issues/127606.

PR #42581, opened 2023-06-05
Convert pixel formats for animated PNG frames using the source blend mode

jason-simmons:
The destination buffer format may not exactly match the source format. In particular, the alpha type of the source may need to be converted to the alpha type of the destination.

Fixes https://github.com/flutter/flutter/issues/128102


bdero (approved):

bdero:
LGTM!

PR #42545, opened 2023-06-04
[Impeller] Reland: Add Impeller Metal support in the embedder API (#42411)

bdero:
This reverts commit https://github.com/flutter/engine/commit/05f958c7f3b1c66d543e9170d54ecc1b5a253df4.

Original PR: https://github.com/flutter/engine/pull/42411

Part of https://github.com/flutter/flutter/issues/112230.

PR #42593, opened 2023-06-06
Revert "[Impeller] Reland: Add Impeller Metal support in the embedder API (#42411)"

zanderso:
Reverts flutter/engine#42545

Not sure if this is needed, but getting a revert started for the regression noted here: https://github.com/flutter/engine/pull/42545#issuecomment-1579034704.


jonahwilliams:
https://github.com/flutter/engine/pull/42594#issuecomment-1579114014 might be easier to land

jonahwilliams:
or cherry pick that is, if we need to

zanderso:
The reland (https://github.com/flutter/engine/pull/42545) missed the beta cut, so no need to CP anything.

PR #42594, opened 2023-06-06
[Impeller] fix partial repaint texture sizing.

jonahwilliams:
Locally this fixes the embedder support regression caused by partial repaint textures being too large.


zanderso:
A bit confused. Wasn't this the fix for the previous regression? https://github.com/flutter/engine/pull/42545/commits/3aebe098249540ece2dc57bb54fd79c4cff4c8ff.

jonahwilliams:
that's wrong then ...

We should revert for now until we figure this out I guess

jonahwilliams:
There may be some confusion as to whether we're wrapping the surface texture or creating a new texture. Whenever we perform a partial repaint, the surface texture isn't used and instead a new texture is created that is a smaller size.

bdero:
This is indeed wrong, but there's another bug in here where I'm re-wrapping the texture with an incorrect descriptor.

PR #42595, opened 2023-06-06
Use proper sampling options in Skwasm's drawAtlas.

eyebrowsoffire:
Make Skwasm consistent with the CanvasKit backend's sampling options in `drawAtlas`. Also, fix the atlas rendering tests so the sprites don't bleed into each other.


yjbanov (approved):
lgtm

PR #42597, opened 2023-06-06
[Impeller] Reland 2: Add Impeller Metal support in the embedder API (#42411)

bdero:
This reverts commit https://github.com/flutter/engine/commit/350a12b3dde49a0eb159d146384924681d31a0b1 + fixes bugs.

Reworked the surface wrapping to use a valid descriptor when wrapping the drawable.


jonahwilliams (approved):
LGTM

bdero:
OK, so looking at the benchmarks, I think this next attempt is good to go. None of the benchmarks are printing validation errors anymore, and the "simple animation" benchmark also didn't regress during the last attempt.
Another thing I fixed with my last attempt at landing this was an extra RT copy into the optional, which was surprisingly expensive enough to show up in the CPU profile because it wasn't doing a move assignment.

PR #42598, opened 2023-06-06
[web] Remove /#/ from home page URL

mdebbar:
To make the `HashUrlStrategy` a bit cleaner (at least for the home page), let's get rid of the notorious `/#/` when the app is in the home page.

Non-home pages continue to have a hash in them. After this PR, here are some example URL changes:

- `http://domain.com/#/` => `http://domain.com/`
- `http://domain.com/#/page1` => (remains the same)

This change is backwards compatible, i.e. if you write `http://domain.com/#/` in the address bar, the app will load and open the home page correctly; and flutter will automatically change the URL to `http://domain.com/`.

Fixes https://github.com/flutter/flutter/issues/127608


kevmoo (approved):
LGTM

PR #42615, opened 2023-06-07
[iOS] Fix TextInputAction.continueAction sends wrong action to framework

bleroux:
## Description

This PR fixes an issue related to text input plugin, on IOS, sending a wrong action string to the engine for `FlutterTextInputActionContinue`.

## Related Issue

Fixes https://github.com/flutter/flutter/issues/126922

## Tests

Adds 1 test.


justinmc (approved):
LGTM :+1: 

I wonder why we named it `continueAction` instead of just `continue` in the first place?

PR #42616, opened 2023-06-07
Revert "[Android] Return keyboard pressed state"

dkwingsmt:
Reverts flutter/engine#41695 due to internal bug report b/284945818.

cc @bleroux let's investigate some day.


jiahaog:
We should try to create a repro from the internal bug so that it's possible for us to debug the issue and reland this :)

PR #42625, opened 2023-06-07
[Impeller] Makes validation layers flag work for android

gaaclarke:
This builds and links in the validation layers for android.  They then can be turned on or off with a manifest field.

fixes https://github.com/flutter/flutter/issues/123788
depends on https://github.com/flutter/buildroot/pull/741

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/shell/config.gni
+++ b/shell/config.gni
@@ -10,6 +10,7 @@ declare_args() {
   shell_enable_vulkan = false
 
   shell_enable_software = true
+  enable_vulkan_validation_layers = false
 }
 
 declare_args() {
--- a/shell/config.gni
+++ b/shell/config.gni
@@ -10,7 +10,6 @@ declare_args() {
   shell_enable_vulkan = false
 
   shell_enable_software = true
-  enable_vulkan_validation_layers = false
 }
 
 declare_args() {
--- a/testing/scenario_app/android/gradle.properties
+++ b/testing/scenario_app/android/gradle.properties
@@ -1,4 +1,4 @@
-org.gradle.jvmargs=-Xmx1536M
+org.gradle.jvmargs=-Xmx2048M
 android.useAndroidX=true
 android.enableJetifier=true
 android.builder.sdkDownload=false
--- a/tools/gn
+++ b/tools/gn
@@ -80,7 +80,7 @@ def to_command_line(gn_args):
   def merge(key, value):
     if isinstance(value, bool):
       return '%s=%s' % (key, 'true' if value else 'false')
-    elif isinstance(value, int):
+    if isinstance(value, int):
       return '%s=%d' % (key, value)
     return '%s="%s"' % (key, value)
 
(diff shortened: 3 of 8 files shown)


shell/config.gni
```diff
@@ -10,6 +10,7 @@ declare_args() {
   shell_enable_vulkan = false
 
   shell_enable_software = true
+  enable_vulkan_validation_layers = false
```
zanderso, line 13: It looks like there's a convention here to prefix the argument names with `shell_`.

tools/gn
```diff
@@ -548,11 +550,8 @@ def to_gn_args(args):
     gn_args['use_fstack_protector'] = True
 
   if args.enable_vulkan_validation_layers:
-    if args.target_os != 'fuchsia':
-      print(
-          'Vulkan validation layers are currently only supported on Fuchsia targets.'
-      )
-      sys.exit(1)
+    if args.target_os == 'android':
+      gn_args['android_api_level'] = int(26)
```
zanderso, line 554: Is the `int()` needed?

vulkan/config.gni
```diff
@@ -3,10 +3,4 @@
 # found in the LICENSE file.
 
 declare_args() {
-  # Whether to include vulkan validation layers, if available.
```
zanderso, line 6: Probably also useful to retain this comment when copying the arg to the new location.

Can this file be deleted?

Alternately, why not just leave the arg here and include this `config.gni` where it's needed?

shell/config.gni
gaaclarke: Moved it back to vulkan/config.gni

tools/gn
gaaclarke: nope, just residual, removed

vulkan/config.gni
gaaclarke: Moved it back, that area of the code is fuchsia specific so I thought `shell` might make more sense.  By name alone though it should be relevant.

testing/scenario_app/android/gradle.properties
```diff
@@ -1,4 +1,4 @@
-org.gradle.jvmargs=-Xmx1536M
+org.gradle.jvmargs=-Xmx2048M
```
zanderso, line 1: What's this for?

zanderso (approved):
cc @jonahwilliams @dnfield

testing/scenario_app/android/gradle.properties
gaaclarke: I was just running into it.  I'm not 100% sure it caused by this change.   My theory is that having those extra .so files in the flutter.jar file somehow requires more memory when the build is happening.

dnfield: (this increases the JVM heap that gradle uses. It's probably fine to increase this)

PR #42628, opened 2023-06-07
[Impeller] Make interleaved layout (more) explicit in generated headers.

jonahwilliams:
Working on https://github.com/flutter/engine/pull/42415 , I found that it was difficult to customize the buffer layout as the interleaved layout is implicitly confiured in the backends. Rather than bake another built in layout, I've pulled (most) information about buffer layout into the generated headers so it is explicitly confiured, which should allow easier customization as the backend has fewer choices to make.

TBD whether or not we need to do something weird for GLES since stride has a different meaning there...

Work towards https://github.com/flutter/flutter/issues/116168


impeller/renderer/vertex_descriptor.cc
```diff
@@ -36,18 +40,26 @@ size_t VertexDescriptor::GetHash() const {
...
+  for (const auto& layout : layouts_) {
+    fml::HashCombineSeed(seed, layout.GetHash());
+  }
   return seed;
 }
 
 // |Comparable<VertexDescriptor>|
 bool VertexDescriptor::IsEqual(const VertexDescriptor& other) const {
-  return inputs_ == other.inputs_;
+  return inputs_ == other.inputs_ && layouts_ == other.layouts_;
```
jonahwilliams, line 51: This doesn't actually seem like the right comparison but unsure if its actually used.

impeller/compiler/code_gen_template.h
```diff
@@ -86,7 +86,8 @@ struct {{camel_case(shader_name)}}{{camel_case(shader_stage)}}Shader {
     {{stage_input.type.type_name}},     // type
     {{stage_input.type.bit_width}}u,    // bit width of type
     {{stage_input.type.vec_size}}u,     // vec size
-    {{stage_input.type.columns}}u       // number of columns
+    {{stage_input.type.columns}}u,      // number of columns
+    {{stage_input.offset}}u,            // offset for interleaved layout
```
jonahwilliams, line 90: Essentially the addtions here are:

1. ShaderStageIO must have an offset
2. A secondary array that has binding+stride info

We may need more info but this is good enough for now. TBD weirdness about stride in GLES

jonahwilliams:
actually just text thankfully

jonahwilliams:
@dnfield @bdero PTAL, this should be working now. Doesn't fully allow customizing vertex layout, but completely moves offset computation offline and removes the need to sort. Should be a better base to build https://github.com/flutter/engine/pull/42415 on

impeller/renderer/vertex_descriptor.cc
```diff
@@ -10,32 +10,28 @@ VertexDescriptor::VertexDescriptor() = default;
...
-    size_t count) {
+    size_t count,
+    const ShaderStageBufferLayout* const stage_layout[],
+    size_t layout_count) {
   inputs_.reserve(inputs_.size() + count);
+  layouts_.reserve(layouts_.size() + layout_count);
   for (size_t i = 0; i < count; i++) {
     inputs_.emplace_back(*stage_inputs[i]);
   }
-  // TODO(jonahwilliams): vulkan shader stage needed sorting too. Remove once
```
jonahwilliams, line 20: Remove, offsets are already computed so no sorting needs to happen.

impeller/compiler/reflector.cc
```diff
@@ -402,8 +403,36 @@ std::shared_ptr<fml::Mapping> Reflector::InflateTemplate(
...
+    const spirv_cross::SmallVector<spirv_cross::Resource>& resources) const {
+  std::vector<size_t> offsets(resources.size(), 0);
+  if (resources.size() == 0) {
+    return offsets;
+  }
+  for (const auto& resource : resources) {
+    const auto type = compiler_->get_type(resource.type_id);
+    auto location = compiler_->get_decoration(
+        resource.id, spv::Decoration::DecorationLocation);
+    // Malformed shader, will be caught later on.
```
dnfield, line 416: Why not have some kind of VALIDATION_LOG here?

jonahwilliams: So this happens with invalid inputs, there is already a validation log that fires and exit that happens, so we'd either need to perform all of that work twice (since this code only runs on vertex shaders), or we do what I did here and let the later stage catch it.

dnfield (approved):

PR #42642, opened 2023-06-07
[web] Move webOnlyAssetManager to `dart:ui_web`

mdebbar:
Moving `webOnlyAssetManager` to `dart:ui_web` and renaming it to simply `assetManager`. This will make it easier for plugins and apps to get a hold of the asset manager without doing `dart:ui` shims or other shenanigans.

Part of https://github.com/flutter/flutter/issues/126831


ditman (approved):
This LGTM! (I thought I had already reviewed this, but I was confusing it with the flutter/packages migration, which is already using the changes here!)

PR #42648, opened 2023-06-08
[Linux] composing delta fixes

Renzo-Olivares:
Fixes a few issues with the Linux text editing delta implementation that was causing some issues when composing CJK text.

`im_preedit_changed_cb` dispatches a delta with the delta range set as the current composing region causing a crash when composing because the new composing region range might not yet exist in the `oldText` that the delta is applied to. We should instead send the composing region before the change to the text as that will be the correct range that is being modified in the `oldText`.

`im_preedit_end_cb` dispatches a delta with an empty `oldText`, a not empty `deltaText`, and a delta range of `(-1,-1)` causing a crash when composing ends because the delta range will never be within the range of the text. We should instead send the current text value as the `oldText` and not send a `deltaRange` or `deltaText` since a `(-1,-1)` range essentially means nothing in the text changed.

`im_commit_cb` does not account for the case when we were previously composing when the text was committed. This causes the dispatched delta to have a delta range that always inserts at the current the selection, when it should instead replace the previous composing range. We should account for the case when we are previously composing before committing the text, and dispatch a delta with the previous composing region.

Fixes https://github.com/flutter/flutter/issues/113909

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/shell/platform/linux/fl_text_input_plugin.cc
+++ b/shell/platform/linux/fl_text_input_plugin.cc
@@ -276,6 +276,7 @@ static void im_preedit_changed_cb(FlTextInputPlugin* self) {
   }
   priv->text_model->UpdateComposingText(buf);
   priv->text_model->SetSelection(flutter::TextRange(cursor_offset));
+
   if (priv->enable_delta_model) {
     std::string text(buf);
     flutter::TextEditingDelta delta = flutter::TextEditingDelta(
--- a/shell/platform/linux/fl_text_input_plugin.cc
+++ b/shell/platform/linux/fl_text_input_plugin.cc
@@ -276,7 +276,7 @@ static void im_preedit_changed_cb(FlTextInputPlugin* self) {
   }
   priv->text_model->UpdateComposingText(buf);
   priv->text_model->SetSelection(flutter::TextRange(cursor_offset));
-      g_warning("preedit changed");
+  g_warning("preedit changed");
   if (priv->enable_delta_model) {
     std::string text(buf);
     flutter::TextEditingDelta delta = flutter::TextEditingDelta(
@@ -292,7 +292,7 @@ static void im_commit_cb(FlTextInputPlugin* self, const gchar* text) {
   FlTextInputPluginPrivate* priv = static_cast<FlTextInputPluginPrivate*>(
       fl_text_input_plugin_get_instance_private(self));
   std::string text_before_change = priv->text_model->GetText();
-    flutter::TextRange composing_before_change = priv->text_model->composing_range();
+  flutter::TextRange composing_before_change = priv->text_model->composing_range();
   flutter::TextRange selection_before_change = priv->text_model->selection();
   gboolean wasComposing = priv->text_model->composing();
 
@@ -301,9 +301,9 @@ static void im_commit_cb(FlTextInputPlugin* self, const gchar* text) {
     priv->text_model->CommitComposing();
   }
 
-      g_warning("im commit");
+  g_warning("im commit");
   if (priv->enable_delta_model) {
-     flutter::TextEditingDelta* delta;
+    flutter::TextEditingDelta* delta;
     if (wasComposing) {
         delta = new flutter::TextEditingDelta(text_before_change, composing_before_change, text);  
     } else {
@@ -324,7 +324,7 @@ static void im_preedit_end_cb(FlTextInputPlugin* self) {
       fl_text_input_plugin_get_instance_private(self));
   std::string text_before_change = priv->text_model->GetText();
   priv->text_model->EndComposing();
-        g_warning("preedit end");
+  g_warning("preedit end");
   if (priv->enable_delta_model) {
     flutter::TextEditingDelta delta = flutter::TextEditingDelta(
         text_before_change, flutter::TextRange(-1, -1), priv->text_model->GetText());
--- a/shell/platform/linux/fl_text_input_plugin.cc
+++ b/shell/platform/linux/fl_text_input_plugin.cc
@@ -307,9 +307,11 @@ static void im_commit_cb(FlTextInputPlugin* self, const gchar* text) {
   if (priv->enable_delta_model) {
     flutter::TextEditingDelta* delta;
     if (wasComposing) {
-      delta = new flutter::TextEditingDelta(text_before_change, composing_before_change, text);
+      delta = new flutter::TextEditingDelta(text_before_change,
+      					    composing_before_change, text);
     } else {
-      delta = new flutter::TextEditingDelta(text_before_change, selection_before_change, text);
+      delta = new flutter::TextEditingDelta(text_before_change,
+         				    selection_before_change, text);
     }
     // flutter::TextEditingDelta delta = flutter::TextEditingDelta(
     //     text_before_change, selection_before_change, text);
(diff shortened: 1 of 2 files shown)


shell/platform/linux/fl_text_input_plugin.cc
```diff
@@ -290,17 +292,27 @@ static void im_commit_cb(FlTextInputPlugin* self, const gchar* text) {
...
-        text_before_change, selection_before_change, text);
-    update_editing_state_with_delta(self, &delta);
+    flutter::TextEditingDelta* delta;
+    if (wasComposing) {
+      delta = new flutter::TextEditingDelta(text_before_change,
+                                            composing_before_change, text);
+    } else {
+      delta = new flutter::TextEditingDelta(text_before_change,
+                                            selection_before_change, text);
+    }
```
cbracken, line 313: Possibly more succinct:
```cc
flutter::TextRange replace_range = wasComposing ? composing_before_change : selection_before_change;
flutter::TextEditingDelta* delta =
    std::make_unique<flutter::TextEditingDelta>(text_before_change, replacement_range, text);

File: shell/platform/linux/fl_text_input_plugin.cc [hunk 2]
```diff
@@ -290,17 +292,27 @@ static void im_commit_cb(FlTextInputPlugin* self, const gchar* text) {
...
   }
 
   if (priv->enable_delta_model) {
-    flutter::TextEditingDelta delta = flutter::TextEditingDelta(
-        text_before_change, selection_before_change, text);
-    update_editing_state_with_delta(self, &delta);
+    flutter::TextEditingDelta* delta;
+    if (wasComposing) {
+      delta = new flutter::TextEditingDelta(text_before_change,
+                                            composing_before_change, text);
```
cbracken, line 309: Avoid new/delete in C++ code. Use `std::unique_ptr<Foo>`, which you can create with `std::make_unique<Foo>(constructor, args, go, here)`;

shell/platform/linux/fl_text_input_plugin.cc
```diff
@@ -290,17 +292,27 @@ static void im_commit_cb(FlTextInputPlugin* self, const gchar* text) {
   FlTextInputPluginPrivate* priv = static_cast<FlTextInputPluginPrivate*>(
       fl_text_input_plugin_get_instance_private(self));
   std::string text_before_change = priv->text_model->GetText();
+  flutter::TextRange composing_before_change =
+      priv->text_model->composing_range();
   flutter::TextRange selection_before_change = priv->text_model->selection();
+  gboolean wasComposing = priv->text_model->composing();
```
cbracken, line 298:
```suggestion
  gboolean was_composing = priv->text_model->composing();
```

shell/platform/linux/fl_text_input_plugin.cc
```diff
@@ -310,10 +322,12 @@ static void im_commit_cb(FlTextInputPlugin* self, const gchar* text) {
 static void im_preedit_end_cb(FlTextInputPlugin* self) {
   FlTextInputPluginPrivate* priv = static_cast<FlTextInputPluginPrivate*>(
       fl_text_input_plugin_get_instance_private(self));
+  std::string text_before_change = priv->text_model->GetText();
   priv->text_model->EndComposing();
   if (priv->enable_delta_model) {
     flutter::TextEditingDelta delta = flutter::TextEditingDelta(
-        "", flutter::TextRange(-1, -1), priv->text_model->GetText());
+        text_before_change, flutter::TextRange(-1, -1),
```
cbracken, line 329: I'm surprised this works. The type of the base/extent positions are `size_t`, which is unsigned.

justinmc (approved):
LGTM, but I'm interested if @cbracken has a better way to represent the (-1,-1) TextRange.

Thanks for fixing these, I probably introduced these bugs when I wrote this :grin: .

shell/platform/linux/fl_text_input_plugin.cc
cbracken: Here's a prototype of something we could use to do this safely:
https://github.com/flutter/engine/compare/main...cbracken:flutter_engine:add-invalid-range?expand=1

If this seems useful, I could commit it, but when I added TextRange originally my thought was to see how far along we could make it without modelling things using invalid positions like `-1` since some platforms (macOS, iOS) don't model any positions with negative numbers, so we're going to need to do translation between models somewhere regardless and TextRange was cleaner without the concept of negative values.

cbracken (changes requested):
Thanks for patching! Added a few comments/suggestions.

shell/platform/linux/fl_text_input_plugin.cc
cbracken: Generally always use RAII-based smart pointers (primarily unique_ptr) unless it's impossible, they're guaranteed to clean themselves up at end of scope whether it's a return or throw, and they're effectively zero overhead.

In this case there's only one way out of the scope, but there's a function call in there that could do something crazy like throw (but almost certainly doesn't 😉). The more ways out of a scope the more `delete`s you'll need to make sure not to forget. Smart pointers guarantee cleanup so why even risk it?

In modern c++ new/delete are generally considered bad practice unless absolutely necessary. In our codebase there are maybe two spots I've found where we couldn't use either unique_ptr or shared_ptr, and we still have a new/delete, but we should avoid adding any more.

More reading in [the C++ FAQ](https://isocpp.org/wiki/faq/freestore-mgmt#memory-leaks) and [C++ Core Guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#S-resource).

cbracken (approved):

shell/platform/linux/fl_text_input_plugin.cc
```diff
@@ -290,17 +292,23 @@ static void im_commit_cb(FlTextInputPlugin* self, const gchar* text) {
...
   if (priv->text_model->composing()) {
     priv->text_model->CommitComposing();
   }
 
   if (priv->enable_delta_model) {
-    flutter::TextEditingDelta delta = flutter::TextEditingDelta(
-        text_before_change, selection_before_change, text);
-    update_editing_state_with_delta(self, &delta);
+    flutter::TextRange replace_range =
+        was_composing ? composing_before_change : selection_before_range;
```
cbracken, line 307:
```suggestion
        was_composing ? composing_before_change : selection_before_change;
```

cbracken: Oops - sorry about that!

shell/platform/linux/fl_text_input_plugin.cc
Renzo-Olivares: I think in this case it may be correct to use composing_before_change. So this may not be needed. Thanks for checking that out!

shell/platform/linux/fl_text_input_plugin.cc
Renzo-Olivares: Ooops not sure how I missed that! Thanks for patching!

shell/platform/linux/fl_text_input_plugin.cc
Renzo-Olivares: correction, a delta range of `(-1,-1)`, is needed but we should just use `TextEditingDelta(const std::u16string & text)` for this case.

PR #42585, opened 2023-06-06
[Impeller] Added a switch to turn on vulkan

gaaclarke:
fixes https://github.com/flutter/flutter/issues/128286

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [Flutter Style Guide] and the [C++, Objective-C, Java style guides].
- [x] I listed at least one issue that this PR fixes in the description above.
- [ ] I added new tests to check the change I am making or feature I am adding, or Hixie said the PR is test-exempt. See [testing the engine] for instructions on writing and running engine tests.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I signed the [CLA].
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/flutter/wiki/Tree-hygiene#overview
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[C++, Objective-C, Java style guides]: https://github.com/flutter/engine/blob/main/CONTRIBUTING.md#style
[testing the engine]: https://github.com/flutter/flutter/wiki/Testing-the-engine
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/wiki/Chat

PR #42660, opened 2023-06-08
Revert "[Impeller] Added a switch to turn on vulkan"

zanderso:
Reverts flutter/engine#42585

Reverting to investigate https://github.com/flutter/flutter/issues/128484

### DIFF
--- a/common/settings.h
+++ b/common/settings.h
@@ -217,9 +217,6 @@ struct Settings {
   bool enable_impeller = false;
 #endif
 
-  // Requests a particular backend to be used (ex "opengles" or "vulkan")
-  std::optional<std::string> impeller_backend;
-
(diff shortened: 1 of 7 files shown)


gaaclarke:
This was discussed in another channel, but sounds like the runtime flag was adding too much size to the engine.

gaaclarke:
Yea, I guess I'm a bit concerned because if our plan is to have a runtime check for Vulkan vs OpenGLES based on the capabilities of the device this is going to be a cost we will have to incur.  If that's not something we are willing to accept we'll have to come up with a different strategy for Android.

The PR only would have affected the stripping of code related to `AndroidSurfaceVulkanImpeller` and its dependencies.  The switch only changed conditional compilation, it didn't change how things are built.

PR #42664, opened 2023-06-08
[Impeller] sort all vertex inputs by location.

jonahwilliams:
Fixes https://github.com/flutter/flutter/issues/128283

Why doesn't this happen elsewhere? might depend on how spirv compiles shaders or the order they run in. Long term fix is https://github.com/flutter/engine/pull/42628 which computes all offset data in the compiler

flutter_04


impeller/renderer/vertex_descriptor.cc
```diff
@@ -17,6 +17,13 @@ bool VertexDescriptor::SetStageInputs(
   for (size_t i = 0; i < count; i++) {
     inputs_.emplace_back(*stage_inputs[i]);
   }
+  // TODO(jonahwilliams): vulkan shader stage needed sorting too. Remove once
+  // autogenerated headers include offset information.
```
dnfield, line 21: Please link an issue tracking this work.

dnfield (approved):

impeller/renderer/vertex_descriptor.cc
jonahwilliams: Done

jonahwilliams:
Will be actually fixed via https://github.com/flutter/engine/pull/42628 . I'm not worried about it regressing

PR #42667, opened 2023-06-08
[fuchsia] Bump the target API level to 12, and pass it to fidlc

mk12:
This makes the GN build pass the target Fuchsia API level to fidlc. Currently it relies on the default of HEAD, meaning all FIDL changes are immediately visible when the SDK rolls. This was never the intention: flutter should be targeting a specific API level.

This unblocks https://fxrev.dev/864297 which makes the fidlc --available flag required, and which uncovered this problem.

I also copied the changes to gen_response_file.py from https://fxrev.dev/865020 to make it forward --available to fidlc.


tools/fuchsia/fidl/fidl_library.gni
```diff
@@ -22,6 +22,9 @@ template("fidl_library") {
 
   assert(defined(invoker.sources), "A FIDL library requires some sources.")
 
+  assert(fuchsia_target_api_level != -1,
```
tarobins, line 25: are there existing usages of this template?  is fuchsia_target_api_level set for all on them?

arbreng:
Generally LGTM, but before merging I pinged mitchell offline with some questions I have

tools/fuchsia/fidl/fidl_library.gni
mk12: Yes it's used to make FIDL bindings for Fuchsia SDK libraries (see tools/fuchsia/sdk/sdk_targets.gni). I copied the != -1 assertion from tools/fuchsia/fuchsia_archive.gni. `fuchsia_target_api_level` is a global GN arg set here https://github.com/flutter/engine/blob/main/tools/gn#L570. The build should fail if there's a case where it's not set.

mk12:
Amended to also bump the API level to 12 as discussed with @arbreng

tarobins:
LGTM

mk12:
@tarobins It says you "self-requested a review" - did you forget to approve, or are you waiting for @arbreng?

tarobins (approved):

tarobins:
No, I just don't know what I'm doing.

mk12:
👍 Thanks. Looks like I still can't merge even with approval, could you do that as well?

arbreng (approved):

PR #41562, opened 2023-04-27
Add Support for Z-Index on CanvasKit PlatformView Layer, Don't Require Additional WebGL Contexts for Layers Drawn Behind the Canvas

jezell:
Implements https://github.com/flutter/flutter/issues/123277

*If you had to change anything in the [flutter/tests] repo, include a link to the migration guide as per the [breaking change policy].*

## Pre-launch Checklist

- [ ] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [Flutter Style Guide] and the [C++, Objective-C, Java style guides].
- [x] I listed at least one issue that this PR fixes in the description above.
- [ ] I added new tests to check the change I am making or feature I am adding, or Hixie said the PR is test-exempt. See [testing the engine] for instructions on writing and running engine tests.
- [ ] I updated/added relevant documentation (doc comments with `///`).
- [x] I signed the [CLA].
- [ ] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/flutter/wiki/Tree-hygiene#overview
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[C++, Objective-C, Java style guides]: https://github.com/flutter/engine/blob/main/CONTRIBUTING.md#style
[testing the engine]: https://github.com/flutter/flutter/wiki/Testing-the-engine
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/wiki/Chat

PR #42672, opened 2023-06-08
Use a single OffscreenCanvas for rendering in CanvasKit

harryterkelsen:
This changes CanvasKit's rendering model. Previously, each overlay canvas induced a new WebGL context and SkSurface. Now, there is only ever one SkSurface (backed by an OffscreenCanvas), and the overlays are HTML canvases with a bitmaprenderer context. The content is rendered using the SkSurface and passed to the overlays using `transferToImageBitmap`. This way, there is only one WebGL context in the lifetime of the app.

## Pre-launch Checklist

- [x] I signed the [CLA].


lib/web_ui/lib/src/engine/canvaskit/render_canvas.dart
```diff
@@ -0,0 +1,165 @@
...
+  int _pixelHeight = -1;
+
+  DomCanvasRenderingContextBitmapRenderer? renderContext;
+
+  ui.Size? _currentCanvasPhysicalSize;
+  ui.Size? _currentRenderSize;
+  double _currentDevicePixelRatio = -1;
+
+  bool _addedToScene = false;
+  void addToScene() {
```
yjbanov, line 55: It looks suspicious to me that the canvas knows where to add itself among other canvases and platform views in the scene host. It's like `int` knowing how to add itself to a `List<int>`.

Suggestion: can we move this logic to the place that makes decisions about how render canvases and platform views are composited?

lib/web_ui/lib/src/engine/canvaskit/render_canvas.dart
```diff
@@ -0,0 +1,165 @@
...
+      _pixelWidth = size.width.ceil();
+      _pixelHeight = size.height.ceil();
+      _updateLogicalHtmlCanvasSize();
+    }
+
+    // This is the first frame we have rendered with this canvas.
+    if (_currentCanvasPhysicalSize == null) {
+      _addedToScene = false;
+
+      _createNewCanvas(size);
```
yjbanov, line 152: Instead of delaying the creation of the canvas, can we create it right away and make it `final` non-null, and then simply update its size? There will be less state to track. We can just update `_currentCanvasPhysicalSize` and move on. Additionally, with the suggestion above, if `RenderCanvas` doesn't need to be responsible for adding itself to the scene host, then we also don't need `_addedToScene`.

lib/web_ui/lib/src/engine/canvaskit/render_canvas_factory.dart
```diff
@@ -0,0 +1,142 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+import 'package:meta/meta.dart';
+
+import '../../engine.dart';
+
+/// Caches canvases used to overlay platform views.
+class RenderCanvasFactory {
```
yjbanov, line 9: Do we need the cache any more? "bitmaprenderer" canvases seem to be "free" in that they don't have their own bitmap memory. Instead, they just take ownership of the `ImageBitmap` memory that's given to them. Maybe what we can do is remember the last "composition" of render canvases and platform views (i.e. the sequence `[R1, V1, R2, V2, ..., Rk, Vk]`) and then optimistically render into them until we find that they need to be recomposited, which happens rarely (when platform views appear/disappear, or when the order changes). If/when we see that a recomposition is needed, we can then compute the minimum necessary "edit" to the sequence (we do that in the HTML renderer and in semantics), fix it, and then render into the new one.

yjbanov (commented):
Reviewed the first 30% of the code so far.

lib/web_ui/lib/src/engine/canvaskit/render_canvas.dart
```diff
@@ -0,0 +1,165 @@
...
+    // If `physicalSize` is not precise, use a slightly bigger canvas. This way
+    // we ensure that the rendred picture covers the entire browser window.
+    _pixelWidth = physicalSize.width.ceil();
+    _pixelHeight = physicalSize.height.ceil();
+    final DomCanvasElement htmlCanvas = createDomCanvasElement(
+      width: _pixelWidth,
+      height: _pixelHeight,
+    );
+    canvasElement = htmlCanvas;
+    renderContext = htmlCanvas.contextBitmapRenderer;
```
yjbanov, line 97: Looking at our implementation of `get contextBitmapRenderer` we simply call `getContext('bitmaprenderer')`. According to [the spec](https://html.spec.whatwg.org/multipage/canvas.html#the-imagebitmaprenderingcontext-interface) the default value of `alpha` is `false`, which means "the canvas is forced to always be opaque". But we do want canvases to be transparent.

yjbanov: Or maybe I'm wrong. It says `false` in the `getContext` signature:

```
context = canvas.getContext('bitmaprenderer' [, { [ alpha: false ] } ])
```

But the dictionary says it's `true`:

```
dictionary ImageBitmapRenderingContextSettings {
  boolean alpha = true;
};
```

lib/web_ui/lib/src/engine/canvaskit/embedded_views.dart
```diff
@@ -409,26 +367,27 @@ class HtmlViewEmbedder {
...
-    _updateOverlays(diffResult);
+    final List<OverlayGroup>? overlayGroups = _updateOverlays(diffResult);
+    if (overlayGroups != null) {
+      _activeOverlayGroups = overlayGroups;
+    }
     assert(
-      _context.pictureRecorders.length == _overlays.length,
-      'There should be the same number of picture recorders '
+      _context.pictureRecorders.length >= _overlays.length,
+      'There should at least as many picture recorders '
```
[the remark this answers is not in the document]
harryterkelsen: Done

lib/web_ui/lib/src/engine/canvaskit/render_canvas.dart
harryterkelsen: > If the alpha setting is provided and set to false, then the canvas is forced to always be opaque.

It seems that `alpha: false` must be provided explicitly for the forced opaque behavior.

lib/web_ui/lib/src/engine/canvaskit/render_canvas.dart
harryterkelsen: Done

lib/web_ui/lib/src/engine/canvaskit/render_canvas.dart
```diff
@@ -0,0 +1,165 @@
...
+    // accessible.
+    htmlCanvas.setAttribute('aria-hidden', 'true');
+
+    htmlCanvas.style.position = 'absolute';
+    _updateLogicalHtmlCanvasSize();
+    htmlElement.append(htmlCanvas);
+  }
+
+  /// Ensures that this canvas can draw a frame of the given [size].
+  void ensureSize(ui.Size size) {
```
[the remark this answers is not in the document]
harryterkelsen: Done

lib/web_ui/lib/src/engine/canvaskit/render_canvas.dart
harryterkelsen: Done.

lib/web_ui/lib/src/engine/canvaskit/render_canvas_factory.dart
harryterkelsen: I agree, right now this factory basically just holds the base canvas and the `pictureToImageSurface`, the caching doesn't help us much. I'll remove this class in my forthcoming PR which reduces the number of overlays required.

jezell:
So this is an interesting approach, it seems to use a single canvas to render each layer, and then use canvas to canvas drawing operations to output it to a non webgl canvas. The goals of this change, I believe, are to avoid the multiple webGL context limit enforced by the browser. Some observations:

This still doesn't actually solve the root problems related to too many layers when HTML elements get introduced into the DOM and we need overlay canvases. Instead of the browsers built-in error limit, now we have a new software defined overlay canvas limit (maximumOverlays), which is 7 vs the browsers 16 (and flutter's current 8?). 

Am I missing something, or will every case that would fail before still fail under this proposed approach, with the caveat that in theory you can make that number whatever you want (if it's exposed) and avoid the hard cap of the browser, but in return we need to do tons of bitmap copy operations which might blow up perf (Firefox and Safari I'm looking at you). 

I guess the proof is in the perf as to whether this is an improvement or not, but how does it solve the problems behind #37890 and the other related issues if it doesn't reduce the number of overlays required and has an even lower limit of overlays due to the perf required for the copies? 

The proposal in #41562 solves the problem that leads to the additional overlays being needed for things like video elements and image elements. Do we need to resurrect that even if this change gets merged?

Instead of going through all this work, why not just allow the overlays to render themselves into the canvas? The open issues related to Video and Image elements causing canvases to be created would require a lot less copying if they could just paint their media element directly into the canvas instead of copying around everything besides them. Skia supports it already so no reason it can't be supported with a special image constructor in web_ui that accepts an HTMLMediaElement.

harryterkelsen:
@jezell I have a follow-up change which reduces the overlays induced by HTML elements which uses the algorithm here: https://docs.google.com/document/d/1aCIOEEYe31HqCaCGUKsAmUu9T0U6o2zUMGpqdPEdJoQ/edit#heading=h.kj5jgtvn1u89

eyebrowsoffire:
@jezell Also, if all you're looking for is a way to create a `ui.Image` from a browser resource, I recently added a web-specific API for making a `ui.Image` from an `ImageBitmap`: https://github.com/flutter/engine/pull/45256

You can use `createImageBitmap` to make an image bitmap from most browser sources you might care about, such as a video or image element.

jezell:
@harryterkelsen @eyebrowsoffire awesome news!

jezell:
@harryterkelsen you can in many cases make a further optimization and reduce the required platform views to one as I did in #41562. As long as the platform views are rectangular, opaque, and ignore pointer events, you can place them behind the canvas and render their region with a clear operation. While some platform views might not meet things conditions, many commonly problematic ones such as camera / video feeds or jpegs do. To the viewer, it appears that they are sitting in the correct location because the things that would have been rendered on top by placing them behind the canvas get erased during the drawing operation instead of creating a new canvas. I believe his approach is superior in these cases as it completely eliminates the need for the overlays when these constraints hold true.

eyebrowsoffire:
@jezell We have actually discussed the "cutout" approach in the past, but as you've already described it only works on completely opaque rectangular views, so we cannot use it as a general purpose solution for platform views.

To help reassure a bit, we have done benchmarking of this change and haven't found any major issues with the performance of this strategy. This is also roughly equivalent to the strategy we are using in the experimental Skwasm renderer, which has proven to be quite performant so far. The amount of copying happening under the hood is actually quite a bit less than you might imagine, as `transferFromImageBitmap` does not actually do any copying, but just replaces the backing store of the canvas itself with the contents of the `ImageBitmap`.

The main thing here is that practically speaking, there are hard limits and very real perf ramifications to creating multiple GL contexts, but those same limitations do not apply to making `ImageBitmap` objects. I don't know of any upper limit to the number of `ImageBitmap` objects one can create, but if there is one it's nowhere near the scale of the limit of the number of WebGL contexts that can be created. We have spent some time discussing this with browser folks too. There is a lot of intelligent resource pooling and reuse that happens under the hood in the browser when creating `ImageBitmap` objects, and they have said this is a reasonable use pattern.

The main thing we achieve from having a single GL context is that we can safely use the same GPU-side resources across multiple overlays. Right now, having multiple GL contexts is unfortunately pretty error prone and has performance cliffs when we have to move resources from one context to another.

Anyway, we really appreciate your engagement (and your patience) on these issues. I'm hoping the new `ui_web` API might be the best way forward for your use-case to avoid making platform views at all, but I'd be curious to hear about the outcomes of any experimentations with it, so please keep us informed. Thank you!

jezell:
@eyebrowsoffire I agree that the ImageBitmap approach is the best general purpose solution, looking forward to giving it a shot. Should solve a lot of the cases for us that get us into trouble. Is the video_player web package being updated to use it? I'll work with the flutter_webrtc guys and see if we can get them to move over to it as well. I believe it should be good enough to solve our current blockers.

lib/web_ui/lib/src/engine/canvaskit/surface.dart
```diff
@@ -205,70 +206,38 @@ class Surface {
...
       _grContext?.releaseResourcesAndAbandonContext();
       _grContext?.delete();
       _grContext = null;
 
       _createNewCanvas(size);
       _currentCanvasPhysicalSize = size;
-    } else if (window.devicePixelRatio != _currentDevicePixelRatio) {
-      _updateLogicalHtmlCanvasSize();
-    }
+    } else if (window.devicePixelRatio != _currentDevicePixelRatio) {}
```
eyebrowsoffire, line 232: I guess this entire else block should be removed? (It's just an empty block now)

eyebrowsoffire (approved):
LGTM! One small comment.

wanjm:
> @eyebrowsoffire I agree that the ImageBitmap approach is the best general purpose solution, looking forward to giving it a shot. Should solve a lot of the cases for us that get us into trouble. Is the video_player web package being updated to use it? I'll work with the flutter_webrtc guys and see if we can get them to move over to it as well. I believe it should be good enough to solve our current blockers.

@jezell  if you have any examples on how to use it in video_player or flutter_webrtc, could you please share to me?
currently , I just guess, we don't need to use platform view in web, just set up html element but not render it and then setup a timer to create ui.Image from the element and then show it?

lib/web_ui/lib/src/engine/canvaskit/surface.dart
harryterkelsen: Done.

PR #42674, opened 2023-06-08
[Impeller] [Vulkan] Make sure to set transfer flags

dnfield:
Even if device transient was requested, we might have decided it can't be device transient and may still put it through a blit pass.


jonahwilliams (approved):
LGTM

dnfield:
https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkImageUsageFlags.html indicates that one of the flags MUST be set, but nothing about any flags being incomptible.

dnfield:
This causes a validation failure in its current form and it's not clear to me how much it's helping locally vs just flakiness. Going to close until I better understand what's going on.

PR #42676, opened 2023-06-08
[Impeller] Specify blend mode on blend filter commands

bdero:
Side quest of https://github.com/flutter/flutter/issues/127232.

This is particularly useful for debugging pipeline blend variations of the blend filter, since we otherwise have to look at the raw pipeline blend ops and piece together what the blend mode must've been.


jonahwilliams (approved):
LGTM

PR #42678, opened 2023-06-08
[Impeller] Reorder blend filter checks to avoid unnecessary destination snapshot

bdero:
Side quest of https://github.com/flutter/flutter/issues/127232.

Tiny reordering to avoid unnecessary input resolution. :) Happened to notice we were hitting this case while going through some frame captures of reported fidelity issues.

Before:
Screenshot 2023-06-08 at 1 45 48 PM

After:
Screenshot 2023-06-08 at 1 48 49 PM


jonahwilliams (approved):
LGTM!

PR #42685, opened 2023-06-09
[Impeller] add explicit VMA flush to device memory writes.

jonahwilliams:
According to my read of the documentation for VMA/Vulkan allocation docs, setting VK_MEMORY_PROPERTY_HOST_COHERENT_BIT means we shouldn't need to flush the memory. nevertheless, this fixes the corruption of host buffers observed in https://github.com/flutter/flutter/issues/124040

Fixes https://github.com/flutter/flutter/issues/124040

> VK_MEMORY_PROPERTY_HOST_COHERENT_BIT bit specifies that the host cache management commands [vkFlushMappedMemoryRanges](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkFlushMappedMemoryRanges.html) and [vkInvalidateMappedMemoryRanges](https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkInvalidateMappedMemoryRanges.html) are not needed to flush host writes to the device or make device writes visible to the host, respectively.

🤷‍♂️

### DIFF
--- a/impeller/renderer/backend/vulkan/device_buffer_vk.cc
+++ b/impeller/renderer/backend/vulkan/device_buffer_vk.cc
@@ -46,6 +46,8 @@ bool DeviceBufferVK::OnCopyHostBuffer(const uint8_t* source,
   if (source) {
     ::memmove(dest + offset, source + source_range.offset, source_range.length);
   }
+  // See https://github.com/flutter/flutter/issues/128556 . Some devices do not
+  // have support for coherent host memory and require an explicit flush.
   ::vmaFlushAllocation(allocator_, allocation_, offset, source_range.length);
 
   return true;
--- a/impeller/renderer/backend/vulkan/device_buffer_vk.cc
+++ b/impeller/renderer/backend/vulkan/device_buffer_vk.cc
@@ -46,6 +46,7 @@ bool DeviceBufferVK::OnCopyHostBuffer(const uint8_t* source,
   if (source) {
     ::memmove(dest + offset, source + source_range.offset, source_range.length);
   }
+  ::vmaFlushAllocation(allocator_, allocation_, offset, source_range.length);
 
   return true;
 }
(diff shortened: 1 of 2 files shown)


impeller/renderer/backend/vulkan/device_buffer_vk.cc
```diff
@@ -46,6 +46,7 @@ bool DeviceBufferVK::OnCopyHostBuffer(const uint8_t* source,
   if (source) {
     ::memmove(dest + offset, source + source_range.offset, source_range.length);
   }
+  ::vmaFlushAllocation(allocator_, allocation_, offset, source_range.length);
```
dnfield, line 49: `#ifdef FML_OS_ANDROID`?

jonahwilliams: are you sure? we can't actually test all the devices this will run on, and this will always be safe to call...

jonahwilliams: I think we should probably just not request coherent memory if we don't need it and add the explict flushes

dnfield (approved):

impeller/renderer/backend/vulkan/device_buffer_vk.cc
dnfield: Ok.

It looks like we should be querying `vmaGetMemoryProperties` to see if the https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkPhysicalDeviceMemoryProperties.html has a VkMemoryType that supports host coherent.

jonahwilliams:
I removed the coherent bit. If we add a cap check then we risk unreproducible failures if the flush gets deleted.

dnfield (approved):
This LGTM, consider adding a comment on why we're not asking for HOST_COHERENT and that if we want to ask for it we need to have some kind of caps check.

dnfield:
I think we'd need some kind of golden test for this on Android - we can get to that once vulkan is available in something the devicelab can see, or maybe try to add one to scenario_app?

PR #42686, opened 2023-06-09
[Impeller] disable buffer to texture blit for Vulkan.

jonahwilliams:
This fixes some f'd up ness in the images. Will need to debug this separately on Android. For now, we can just turn it off in the caps.

### DIFF
--- a/impeller/renderer/backend/vulkan/capabilities_vk.cc
+++ b/impeller/renderer/backend/vulkan/capabilities_vk.cc
@@ -328,7 +328,7 @@ bool CapabilitiesVK::SupportsSSBO() const {
 
 // |Capabilities|
 bool CapabilitiesVK::SupportsBufferToTextureBlits() const {
-  return true;
+  return false;
 }
 
 // |Capabilities|


impeller/renderer/backend/vulkan/capabilities_vk.cc
```diff
@@ -328,7 +328,7 @@ bool CapabilitiesVK::SupportsSSBO() const {
 
 // |Capabilities|
 bool CapabilitiesVK::SupportsBufferToTextureBlits() const {
-  return true;
+  return false;
```
dnfield, line 331: Consider an ifdef with `!FML_OS_ANDROID` and a comment explaining this is implemented and working on non-Android platforms.

jonahwilliams: I'm not confident in that given that I can run the gallery on vulkan outside of android

jonahwilliams: err can't

dnfield (approved):

PR #42692, opened 2023-06-09
[Impeller] Add CPU implementations for all color filters

bdero:
Part of https://github.com/flutter/flutter/issues/127232.

Adds a CPU implementation for ColorMatrix and the color space conversions. Also changes the blend signature for consistency.

These will be necessary to make the mask blur fast path continue working in the presence of color filters. In general, we can use these to eliminate a needlessly expensive image-based color filter for the non-image color sources.


jonahwilliams (approved):
LGTM

impeller/entity/entity_unittests.cc
```diff
@@ -1931,238 +1931,6 @@ TEST_P(EntityTest, SrgbToLinearFilter) {
   ASSERT_TRUE(OpenPlaygroundHere(callback));
 }
 
-TEST_P(EntityTest, TTTBlendColor) {
```
dnfield, line 1934: What's replacing these tests?

See also https://github.com/flutter/flutter/issues/128606. It's possible some of these tests are encoding bad behavior, but it seems veyr likely to me that if they are not encoding bad behavior they're missing cases where there's bad behavior.

dnfield: Sorry, I see below the new tests.

bdero: Yeah I just moved this existing test into `GeometryTest.ColorBlend`. And yes, some of them are wrong -- we used to use these for absorbing the DrawPaints, which introduced bugs. Some blends are also missing from the test at the moment.

PR #42700, opened 2023-06-09
Capture additional final inset states in ImeSyncDeferringInsetsCallback

gmackall:
Fixes https://github.com/flutter/flutter/issues/118761.

For some context: ImeSyncDeferringInsetsCallback extends [WindowInsetsAnimation.Callback](https://developer.android.com/reference/android/view/WindowInsetsAnimation.Callback) specifically to handle insets related to keyboard animations.

When a keyboard animation happens, we get a call to onPrepare with the animation, and then to onApplyWindowInsets with [WindowInsets](https://developer.android.com/reference/android/view/WindowInsets) that represent the post animation state. For example, for hiding the keyboard, we would get WindowInsets with height of 0 for insets of type WindowInsets.Type.ime() (input method editor). We save these post-animation insets. 

We then get calls to onProgress for each frame of the animation, and use the WindowInsets passed to onProgress to build new WindowInsets based on the saved post-animation insets, but with the keyboard insets overridden by the WindowInsets passed to onProgress.

Finally, we get a call to onEnd, where we apply the saved insets.

However, there can be multiple animation running at the same time. For example, dismissing the keyboard can result in an animation associated solely with the keyboard happening, as well as an animation associated only with the navigation bar happening. And they can start at slightly different times. This can result in the situation described in this comment: https://github.com/flutter/flutter/issues/118761#issuecomment-1574284505

Re-setting needsSave to false ensures we don't ignore any updates to the final state, ensuring correct insets in onProgress and onEnd.

## Pre-launch Checklist

- [x] I signed the [CLA].


gmackall:
@jason-simmons I requested your review as you had made the most changes to this callback, but let me know if you're not the right person to review! I've also reached out on #hackers-engine to see if anyone there knows who best to review

jason-simmons (approved):

PR #42708, opened 2023-06-09
[Impeller] Added helper messaging around validation usage

gaaclarke:
Adds some helper messages around using validation layers.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/impeller/renderer/backend/vulkan/context_vk.cc
+++ b/impeller/renderer/backend/vulkan/context_vk.cc
@@ -190,9 +190,9 @@ void ContextVK::Setup(Settings settings) {
       ss << vk::to_string(validation) << " ";
     }
     ss << "]";
-    FML_DLOG(ERROR) << ss.str();
+    FML_LOG(ERROR) << ss.str();
 #if !defined(IMPELLER_ENABLE_VULKAN_VALIDATION_LAYERS) && FML_OS_ANDROID
-    VALIDATION_LOG << "Vulkan validation layers found but the gn argument "
+    FML_LOG(ERROR) << "Vulkan validation layers turned on but the gn argument "
                       "`--enable-vulkan-validation-layers` is missing.";
 #endif
     instance_info.pNext = &validation;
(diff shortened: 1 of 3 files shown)


impeller/renderer/backend/vulkan/context_vk.cc
```diff
@@ -184,6 +184,17 @@ void ContextVK::Setup(Settings settings) {
...
   if (caps->AreValidationsEnabled()) {
+    std::stringstream ss;
+    ss << "Enabling validation layers, features: [";
+    for (const auto& validation : enabled_validations) {
+      ss << vk::to_string(validation) << " ";
+    }
+    ss << "]";
+    FML_DLOG(ERROR) << ss.str();
+#if !IMPELLER_ENABLE_VULKAN_VALIDATION_LAYERS && FML_OS_ANDROID
+    VALIDATION_LOG << "Attempting to use validation layers without "
```
dnfield, line 195: I don't think VALIDATION_LOG is right here - unless I'm misunderstanding and it's an error to end up in this state (although it seems like there are multiple ways now to turn on validation and I'm not clear on which path(s) are allowed or not).

gaaclarke: Yep, it's an error.  We want people to be using the validation layers that we are building.  There is no reason to turn on validation layers without building and packaging up validation layers.  Jonah ran into this the other day where he thought just turning on the environment variable was sufficient then had to go through the rigamarole to package prebuilt layers.

dnfield (approved):
LGTM with nit about message wording

impeller/renderer/backend/vulkan/context_vk.cc
gaaclarke: I updated it.  I made it a bit stronger and also made sure to mention `gn` (as opposed to runtime arguments).

gaaclarke: (sorry, uploaded now, i accidentally committed the local hack i need to make scenario_app work on my machine instead)

jonahwilliams: VALIDATION_LOG  isn't fatal by default and won't show up in logcat unless fatal validations are enabled

gaaclarke: So, this will crash in a way that is hard to know what is going on if the runtime flag is on, but it wasn't built to support it.  Sounds like we want an `FML_LOG(ERROR)` instead.

gaaclarke: Switched both things to `FML_LOG(ERROR)`. There is no need to use `DLOG` since using validation layers should never happen in the wild anyways.  `ERROR` since we know it will probably crash.

PR #42715, opened 2023-06-09
[Impeller] Fix vulkan gradient nan issue.

jonahwilliams:
Fixes https://github.com/flutter/flutter/issues/128603

We were hitting a nan issue. GLSL has different relaxed precision semantics than vulkan does, and didn't hit it for some reason.


ColdPaleLight (approved):
LGTM

PR #42753, opened 2023-06-12
[Impeller] Correct attachment description for offscreen MSAA resolve.

jonahwilliams:
Two issues:

1. The resolve texture is created from the same impeller attachment as the msaa attachment, so the store mode was getting set to dont care instead of store.

2. the image layout for the resolve attachment should be ePresentSrcKHR, at least from following the guide at https://vulkan-tutorial.com/Multisampling

Fixing both of these locally fixes all of the weird offscreen cursed rendering.

Fixes https://github.com/flutter/flutter/issues/128600

### DIFF
--- a/impeller/renderer/backend/vulkan/formats_vk.h
+++ b/impeller/renderer/backend/vulkan/formats_vk.h
@@ -427,8 +427,7 @@ constexpr vk::AttachmentDescription CreateAttachmentDescription(
     SampleCount sample_count,
     LoadAction load_action,
     StoreAction store_action,
-    vk::ImageLayout current_layout,
-    bool resolve_texture = false) {
+    vk::ImageLayout current_layout) {
   vk::AttachmentDescription vk_attachment;
 
   vk_attachment.format = ToVKImageFormat(format);
@@ -467,9 +466,7 @@ constexpr vk::AttachmentDescription CreateAttachmentDescription(
   switch (kind) {
     case AttachmentKind::kColor:
       vk_attachment.initialLayout = current_layout;
-      vk_attachment.finalLayout =
-          resolve_texture ? vk::ImageLayout::eGeneral
-                          : vk::ImageLayout::eColorAttachmentOptimal;
+      vk_attachment.finalLayout = vk::ImageLayout::eGeneral;
       break;
     case AttachmentKind::kDepth:
     case AttachmentKind::kStencil:
--- a/impeller/renderer/backend/vulkan/formats_vk.h
+++ b/impeller/renderer/backend/vulkan/formats_vk.h
@@ -468,7 +468,7 @@ constexpr vk::AttachmentDescription CreateAttachmentDescription(
     case AttachmentKind::kColor:
       vk_attachment.initialLayout = current_layout;
       vk_attachment.finalLayout =
-          resolve_texture ? vk::ImageLayout::ePresentSrcKHR
+          resolve_texture ? vk::ImageLayout::eGeneral
                           : vk::ImageLayout::eColorAttachmentOptimal;
       break;
     case AttachmentKind::kDepth:
--- a/impeller/renderer/backend/vulkan/formats_vk.h
+++ b/impeller/renderer/backend/vulkan/formats_vk.h
@@ -427,7 +427,8 @@ constexpr vk::AttachmentDescription CreateAttachmentDescription(
     SampleCount sample_count,
     LoadAction load_action,
     StoreAction store_action,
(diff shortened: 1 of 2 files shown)


jonahwilliams:
lol now we get a validation error. I think I just need to correct the layout info.

jonahwilliams:
hilariously I misread the tutorial, but visually things are fixed for me 😮‍💨

jonahwilliams:
So I suspect that rather than fixing this, the error has introduced some pessimization which is causing the synchronization issue to resolve itself.

impeller/renderer/backend/vulkan/formats_vk.h
```diff
@@ -466,7 +467,9 @@ constexpr vk::AttachmentDescription CreateAttachmentDescription(
   switch (kind) {
     case AttachmentKind::kColor:
       vk_attachment.initialLayout = current_layout;
-      vk_attachment.finalLayout = vk::ImageLayout::eColorAttachmentOptimal;
+      vk_attachment.finalLayout =
+          resolve_texture ? vk::ImageLayout::ePresentSrcKHR
+                          : vk::ImageLayout::eColorAttachmentOptimal;
```
dnfield, line 472: Is this the right switch value?

Spec says that eColorAttachmentOptimal "must only" be used as a color ro resolve in a `VkFramebuffer`. 

Perhaps the argument should instead be `bool is_framebuffer` or something? Or perhaps an enum where the enumerated values are something like `kFramebuffer` and `kNonFramebuffer`.

dnfield: I'm not sure I really understand what `ePresentSrcKHR` is really for. Did you already try `eGeneral`?

impeller/renderer/backend/vulkan/render_pass_vk.cc
```diff
@@ -95,7 +98,7 @@ SharedHandleVK<vk::RenderPass> RenderPassVK::CreateVKRenderPass(
     if (color.resolve_texture) {
       resolve_refs[bind_point] =
           vk::AttachmentReference{static_cast<uint32_t>(attachments.size()),
-                                  vk::ImageLayout::eColorAttachmentOptimal};
+                                  vk::ImageLayout::ePresentSrcKHR};
```
dnfield, line 101: (Ditto here: maybe eGeneral instead?)

impeller/renderer/backend/vulkan/render_pass_vk.cc
```diff
@@ -49,14 +49,17 @@ static vk::AttachmentDescription CreateAttachmentDescription(
 
   if (desc.storage_mode == StorageMode::kDeviceTransient) {
```
dnfield, line 50: Is the real problem that we're setting the storage mode to device transient?

I'm not quite clear on when we should set dontcare or not though.

impeller/renderer/backend/vulkan/formats_vk.h
jonahwilliams: I don't really understand either, but it was the value used in https://vulkan-tutorial.com/Multisampling so I just went with it. I'll try general

impeller/renderer/backend/vulkan/render_pass_vk.cc
jonahwilliams: no that doesn't seem to be it, at least things are already working for me

impeller/renderer/backend/vulkan/formats_vk.h
jonahwilliams: eGeneral works! lets see if it validates

impeller/renderer/backend/vulkan/formats_vk.h
```diff
@@ -466,7 +467,9 @@ constexpr vk::AttachmentDescription CreateAttachmentDescription(
   switch (kind) {
     case AttachmentKind::kColor:
       vk_attachment.initialLayout = current_layout;
-      vk_attachment.finalLayout = vk::ImageLayout::eColorAttachmentOptimal;
+      vk_attachment.finalLayout =
+          resolve_texture ? vk::ImageLayout::eGeneral
```
dnfield, line 471: The nitpick here is using "resolve_texture" as the reason to do this.

The not nitpick here is that I think we should go with eGeneral unless we know for sure that this is for the framebuffer. eGeneral should always be safe if I'm understanding the spec, whereas the eColorAttachmentOptimal is only safe if we know we're using this for framebuffer.

dnfield (approved):

PR #42789, opened 2023-06-12
Predictive back breakage fix

justinmc:
This PR attempts to fix breakage(s) caused by https://github.com/flutter/engine/pull/39208.

b/286627757

## Main problem

There were breakages in Google's code where classes that implemented PlatformMessageHandler did not provide an implementation of setFrameworkHandlesBack.

```
... is not abstract and does not override abstract method setFrameworkHandlesBack(boolean) in PlatformMessageHandler
```

I've tried to fix this by adding a default empty implementation to PlatformMessageHandler.  Is that ok to do there, even though other nearby methods don't have default implementations there?

Also, I noticed that I could probably do the same thing to avoid my empty implementation in FlutterFragment, so I added one to PlatformPluginDelegate as well.  That did allow me to remove the implementation in FlutterFragment.  Is that ok as well?  Do I need to provide both empty implementations?

## Other cleanup

I noticed that when I renamed frameworkHandlesBacks to frameworkHandlesBack, I forgot to rename some cases, which I've cleaned up here.

### DIFF
--- a/shell/platform/android/io/flutter/embedding/engine/systemchannels/PlatformChannel.java
+++ b/shell/platform/android/io/flutter/embedding/engine/systemchannels/PlatformChannel.java
@@ -517,7 +517,7 @@ public class PlatformChannel {
     void setSystemUiOverlayStyle(@NonNull SystemChromeStyle systemUiOverlayStyle);
 
     /** The Flutter application would or would not like to handle navigation pop events itself. */
-    void setFrameworkHandlesBack(boolean frameworkHandlesBack);
+    default void setFrameworkHandlesBack(boolean frameworkHandlesBack) {}
 
     /**
      * The Flutter application would like to pop the top item off of the Android app's navigation
--- a/shell/platform/android/io/flutter/embedding/engine/systemchannels/PlatformChannel.java
+++ b/shell/platform/android/io/flutter/embedding/engine/systemchannels/PlatformChannel.java
@@ -516,7 +516,13 @@ public class PlatformChannel {
      */
     void setSystemUiOverlayStyle(@NonNull SystemChromeStyle systemUiOverlayStyle);
 
-    /** The Flutter application would or would not like to handle navigation pop events itself. */
+    /**
+     * The Flutter application would or would not like to handle navigation pop events itself.
+     *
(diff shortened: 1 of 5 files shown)


shell/platform/android/io/flutter/plugin/platform/PlatformPlugin.java
```diff
@@ -57,7 +57,13 @@ public interface PlatformPluginDelegate {
...
     boolean popSystemNavigator();
 
-    void setFrameworkHandlesBack(boolean frameworkHandlesBacks);
+    /**
+     * The Flutter application would or would not like to handle navigation pop events itself.
+     *
+     * <p>Relevant for registering and unregistering the app's OnBackInvokedCallback for the
+     * Predictive Back feature.
+     */
+    default void setFrameworkHandlesBack(boolean frameworkHandlesBack) {}
```
camsim99, line 66: I think this works as a good alternative to the empty implementation in `FlutterFragment`. Maybe the doc can expand on where this is currently relevant (`FlutterActivity` alone now I suppose)?

shell/platform/android/io/flutter/embedding/engine/systemchannels/PlatformChannel.java
```diff
@@ -517,7 +517,7 @@ public interface PlatformMessageHandler {
     void setSystemUiOverlayStyle(@NonNull SystemChromeStyle systemUiOverlayStyle);
 
     /** The Flutter application would or would not like to handle navigation pop events itself. */
-    void setFrameworkHandlesBack(boolean frameworkHandlesBack);
+    default void setFrameworkHandlesBack(boolean frameworkHandlesBack) {}
```
camsim99, line 520: This makes sense to me since it is only used in a particular place. I'd suggest expanding the doc to mention where exactly this is relevant currently (`FlutterActivity`), just so it's clear we don't expect it to work or be used elsewhere.

camsim99 (approved):
This approach looks good to me given that `setFrameworkHandlesBack` is only relevant in particular places. Just left suggestions on expanding the documentation to make that fact clearer to the reader!

PR #42792, opened 2023-06-12
[Impeller] Fix text jitter on Vulkan.

jonahwilliams:
Fixes https://github.com/flutter/flutter/issues/128624


gaaclarke (approved):
Hopefully this shows up in the swiftshader tests.  I'm not sure of another easy way to test that we don't break this.

jonahwilliams:
It probably wont. Vulkan Relaxed precision can be ignored by backends, and I'd expect anything CPU backed to ignore relaxed precision since they won't actually have 16bit registers.

PR #25064, opened 2021-03-18
Decode empty message to nil in standard codec

cbracken:
This updates Flutter.*Codec `decode:` implementations to treat empty
messages as equivalent to nil or NSNull, which avoids a crash when we
subsequently otherwise try to read the type/value from the message. This
handles the case where the sender were to send `null` over the channel.
e.g.,

    final channel = BasicMessageChannel<Object?>('somechannel', StandardMessageCodec());
    channel.send(null);

It also updates the macOS embedder to send nil to the decoder when a
zero-length message is received in order to be consistent with the iOS
embedding.

Previously, the macOS embedder always encoded platform messages as
NSData regardless of length:
https://github.com/flutter/engine/blob/ba93431e6885f94614d91c291ae0336bac6b70b0/shell/platform/darwin/macos/framework/Source/FlutterEngine.mm#L498-L500

The iOS embedder did not have this issue since it special-cased
zero-length messages as nil:
https://github.com/flutter/engine/blob/ba93431e6885f94614d91c291ae0336bac6b70b0/shell/platform/darwin/ios/framework/Source/platform_message_router.mm#L23-L26

Bug: flutter/flutter#78003

## Pre-launch Checklist

- [X] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [X] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [X] I read and followed the [Flutter Style Guide] and the [C++, Objective-C, Java style guides].
- [X] I listed at least one issue that this PR fixes in the description above.
- [X] I added new tests to check the change I am making or feature I am adding, or Hixie said the PR is test-exempt. See [testing the engine] for instructions on
writing and running engine tests.
- [X] I updated/added relevant documentation (doc comments with `///`).
- [X] I signed the [CLA].
- [X] All existing and new tests are passing.
- [ ] The reviewer has submitted any presubmit flakes in this PR using the [engine presubmit flakes form] before re-triggering the failure.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/flutter/wiki/Tree-hygiene#overview
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[C++, Objective-C, Java style guides]: https://github.com/flutter/engine/blob/master/CONTRIBUTING.md#style
[...]

PR #42808, opened 2023-06-13
[Linux] Allow BasicMessageChannel sending and responding to null message

bleroux:
## Description

This PR changes how a null message is handled by the Linux engine for a basic message channel.

**Before**:

- when receiving a null message a warning was emitted.
- `fl_binary_messenger_send_response` was called but failed and the application exited.

```
** (bug:9866): WARNING **: 23:13:42.109: Failed to decode message: Unexpected end of data

** (bug:9866): CRITICAL **: 23:13:42.109: gboolean send_response(FlBinaryMessenger *, FlBinaryMessengerResponseHandle *, GBytes *, GError **): assertion 'response_handle->response_handle != nullptr' failed
```

**After**:

- Receiving a null message is handled as expected from the framework side documentation:

https://github.com/flutter/flutter/blob/9287e81d52d69560222c14170a2fd7def613fe61/packages/flutter/lib/src/services/platform_channel.dart#L149-L150

## Related Issue

Fixes https://github.com/flutter/flutter/issues/128704

## Tests

Adds 2 tests.

### DIFF
--- a/shell/platform/linux/fl_basic_message_channel.cc
+++ b/shell/platform/linux/fl_basic_message_channel.cc
@@ -232,7 +232,6 @@ G_MODULE_EXPORT void fl_basic_message_channel_send(FlBasicMessageChannel* self,
                                                    GAsyncReadyCallback callback,
                                                    gpointer user_data) {
   g_return_if_fail(FL_IS_BASIC_MESSAGE_CHANNEL(self));
-  g_return_if_fail(message != nullptr);
 
   g_autoptr(GTask) task =
       callback != nullptr ? g_task_new(self, cancellable, callback, user_data)
--- a/shell/platform/linux/fl_standard_message_codec.cc
+++ b/shell/platform/linux/fl_standard_message_codec.cc
@@ -420,6 +420,10 @@ static GBytes* fl_standard_message_codec_encode_message(FlMessageCodec* codec,
 static FlValue* fl_standard_message_codec_decode_message(FlMessageCodec* codec,
                                                          GBytes* message,
                                                          GError** error) {
+  if (g_bytes_get_size(message) == 0) {
+    return fl_value_new_null();
+  }
+
   FlStandardMessageCodec* self =
       reinterpret_cast<FlStandardMessageCodec*>(codec);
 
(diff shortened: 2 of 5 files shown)


shell/platform/linux/fl_basic_message_channel.cc
```diff
@@ -232,7 +232,6 @@ G_MODULE_EXPORT void fl_basic_message_channel_send(FlBasicMessageChannel* self,
                                                    GAsyncReadyCallback callback,
                                                    gpointer user_data) {
   g_return_if_fail(FL_IS_BASIC_MESSAGE_CHANNEL(self));
-  g_return_if_fail(message != nullptr);
```
stuartmorgan, line 235: This will mean calling `fl_message_codec_encode_message` with null below; does that work? Is there unit testing of it? Does it return a non-null data pointer, such that line 246 won't return an error?

shell/platform/linux/fl_standard_message_codec.cc
```diff
@@ -420,6 +420,10 @@ static GBytes* fl_standard_message_codec_encode_message(FlMessageCodec* codec,
 static FlValue* fl_standard_message_codec_decode_message(FlMessageCodec* codec,
                                                          GBytes* message,
                                                          GError** error) {
+  if (g_bytes_get_size(message) == 0) {
+    return fl_value_new_null();
```
stuartmorgan, line 424: @cbracken Null messages strike again :P

How do you think we should handle null messages here: null data, or a value whose value is null? IIRC you found that this was inconsistent in https://github.com/flutter/engine/pull/25064 but I forget whether you cleaned it up later.

shell/platform/linux/fl_basic_message_channel.cc
robert-ancell: Please add '(allow-none)' to the documentation string in `public/flutter_linux/fl_basic_message_channel.h`

shell/platform/linux/fl_standard_message_codec.cc
robert-ancell: You have to return the FlValue object here, because returning NULL implies there was an error (GObject convention).

robert-ancell (approved):
With the docstring change this looks good, on the assumption that this is the correct behaviour for receiving an empty message.

shell/platform/linux/fl_basic_message_channel.cc
bleroux: > Please add '(allow-none)' to the documentation string in public/flutter_linux/fl_basic_message_channel.h

Done, thanks for noticing this :+1:

stuartmorgan (approved):
LGTM, thanks!

PR #42818, opened 2023-06-13
[Impeller] Use explicit float types in Vulkan shaders

bdero:


impeller/compiler/shader_lib/impeller/types.glsl
```diff
@@ -9,7 +9,12 @@
 #extension GL_AMD_gpu_shader_half_float_fetch : enable
 #extension GL_EXT_shader_explicit_arithmetic_types_float16 : enable
 
-#ifndef IMPELLER_TARGET_METAL_IOS
+// OpenGLES 2 targets GLSL ES 1.0, which doesn't support explicit arithmetic
+// types on individual declarations. So for GLES, we set the float types to
+// `mediump` by default and provide macros for 16 bit types to keep writing
+// cross API shaders easy.
+
+#ifdef IMPELLER_TARGET_OPENGLES
```
jonahwilliams, line 17: I don't think this will work in general, the 16bit extensions aren't available everywhere and we can't actually support a backend that uses GLES and has 16bit uniform values, as the bindings will be incorrect.

jonahwilliams: TIL.

we use the same generated headers on GLES and Vulkan though, so we'd need to solve that issue too - probably generate a different set of bindings assuming 32bit and 16bit float and select at runtime.

bdero: > we use the same generated headers on GLES and Vulkan though

Just for Android, right? On MacOS, I think we end up using the Metal headers for all of the backends, which maybe is another problem that needs to be solved. :/

I'm still not really sure what the best design approach would be for reconciling the different backend headers. Maybe we can automate the common case and create a generic lowest common denominator meta-header with special UBO/VBO builders to dispatch the specific packing behavior of the backend in use. But importantly, using this is optional, and we can still  use the per-backend struct directly for special backend-specific optimizations whenever there's a need.

bdero:
I have give up on solving the 16 bit buffer access problems for the time being, so closing this out.

Continuation issues:
- Vulkan: https://github.com/flutter/flutter/issues/128836
- GLES: https://github.com/flutter/flutter/issues/128881

PR #42820, opened 2023-06-13
[Impeller] Reland attempt Vulkan setup and fallback to GLES.

jonahwilliams:


shell/platform/android/context/android_context.h
```diff
@@ -16,6 +16,9 @@ enum class AndroidRenderingAPI {
   kSoftware,
   kOpenGLES,
   kVulkan,
+  /// @brief Attempt to create a Vulkan surface, if that fails then fall back
+  /// to GLES.
+  kAutoselect,
```
jonahwilliams, line 21: I don't know if adding this will cause issues elsewhere, I''m going to run it on CI and see.

Alternatively could just use a separate flag below.

shell/platform/android/platform_view_android.cc
```diff
@@ -74,8 +74,8 @@ static std::shared_ptr<flutter::AndroidContext> CreateAndroidContext(
   if (enable_impeller) {
     // TODO(gaaclarke): We need to devise a more complete heuristic about what
```
gaaclarke, line 75: You can get rid of this TODO now.

gaaclarke (approved):

chinmaygarde (approved):

shell/platform/android/platform_view_android.cc
jonahwilliams: Done

jason-simmons:
Removed the autosubmit label - this should wait until https://github.com/flutter/engine/pull/42821 lands.

Without https://github.com/flutter/engine/pull/42821 the process may crash while tearing down the `ContextVK` if Vulkan setup did not succeed.

jonahwilliams:
Thanks @jason-simmons !

dnfield:
Other patch landed, re-adding autosubmit.

PR #42825, opened 2023-06-13
Make validation logs ERROR level

dnfield:
INFO logs do not show up in normal logging unless an extra flag is passed by the tool (and it is a different flag from the "-v" flag).

This means bug reports that include logs about impeller are missing the validation errors, which makes them harder to triage and dedupe.

Fixes https://github.com/flutter/flutter/issues/128808

### DIFF
--- a/impeller/base/validation.cc
+++ b/impeller/base/validation.cc
@@ -46,7 +46,7 @@ void ImpellerValidationBreak(const char* message) {
   if (sValidationLogsAreFatal) {
     FML_LOG(FATAL) << stream.str();
   } else {
-    FML_LOG(INFO) << stream.str();
+    FML_LOG(ERROR) << stream.str();
   }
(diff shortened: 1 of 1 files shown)


chinmaygarde (approved):

gaaclarke (approved):
Isn't this going to potentially print out anytime we use opengl as the fallback?  Should this only be ERROR in debug builds?

dnfield:
> Isn't this going to potentially print out anytime we use opengl as the fallback? Should this only be ERROR in debug builds?

It's only going to print out if `IMPELLER_ENABLE_VALIDATION` is true, which is set if `IMPELLER_DEBUG` is true, which is only set in debug and profile mode for Flutter right now.

dnfield:
(and knowing why we failed to choose vulkan may be valuable information anyway)

PR #42830, opened 2023-06-14
[web] Fix for Safari autofill flickering bug 

htoor3:
Safari web autofill has an issue where the autofill dialog flickers and doesn't completely render and allow users to fill any forms. 

This is caused by a collision between default browser behavior on the `pointerdown` event and the programatic focusing of our inputs. 

### Problem

When we click into an input, the element is created, placed, and (explicitly) focused.  However, all of this is done before the `pointerdown` event can finish.  Since all `pointerdown` elements target the `flutter-view` (formerly `flt-glass-pane`), default browser behavior is to trigger a `blur` event since the target doesn't match what's currently focused, which is the input element.  This doesn't manifest into most text editing issues because we listen for `blur` events on the `input` and call `focus()`.  However, in Safari, this near-instant focus/blur results in the disappearance of the autofill popup. 

The current chain of events looks like: 

`pointerdown` event starts -> input is created + focused -> `pointerdown` event ends, and triggers a `blur` -> input refocuses on `blur`

### Solution

This change ensures that we don't focus the input until after the `pointerdown` event concludes, thus preventing any rapid-fire `blur` event from being emitted.  We do this via wrapping the focus logic within a zero-duration Timer.  

The new chain of events looks like:

`pointerdown` event starts -> `pointerdown` event ends ->  input is created + focused

### Alternative approach

Note: Another option was to call `preventDefault()` on the `pointerdown` event to prevent the `blur` from occurring that way.  There may be unintended side effects from that approach, especially as it relates to platform views.  The surface area of the chosen approach is much more contained and should result in no side effects outside of Safari Desktop's text editing strategy. 

Screenshot 2023-06-15 at 7 34 02 PM

Fixes https://github.com/flutter/flutter/issues/127960

## Pre-launch Checklist

- [x] I signed the [CLA].


mdebbar:
Change looks good to me!

Let's add some tests and this should be good to go.

htoor3:
> Change looks good to me!
> 
> Let's add some tests and this should be good to go.

All done! Requesting re-review

lib/web_ui/test/engine/text_editing_test.dart
```diff
@@ -532,6 +532,12 @@ Future<void> testMain() async {
               Matrix4.translationValues(10.0, 20.0, 30.0).storage.toList());
       sendFrameworkMessage(codec.encodeMethodCall(setSizeAndTransform));
 
+      // This timer exists because Desktop Safari element is focused after
+      // zero-duration timer to prevent autofill dialog
+      if(isSafari){
+        await Future<void>.delayed(Duration.zero);
+      }
```
mdebbar, line 539: This block (comment + code) is being repeated a lot. Should we maybe extract this into a function?
```dart
/// On Desktop Safari, the editing element is focused after zero-duration timer
/// to prevent autofill dialog ...
Future<void> waitForDesktopSafariFocus() async {
  if (editingStrategy is SafariDesktopTextEditingStrategy) {
    await Future<void>.delayed(Duration.zero);
  }
}
```

```suggestion
      await waitForDesktopSafariFocus();
```

mdebbar (approved):

PR #42902, opened 2023-06-16
[Impeller] correct default PSO pixel format and sample count.

jonahwilliams:
Despite compiling an initial PSO variant for every shader, in practice these were completely useless as they had a sample count of 1 (instead of 4 for MSAA) and an invalid pixel format.

Correct the defaults so they may be usful.

Not enough to fix https://github.com/flutter/flutter/issues/128963 , because we still have far too much state for Vulkan to handle


chinmaygarde (approved):

jonahwilliams:
Holding off on landing this, I'd like to make sure that this and @gaaclarke 's patch roll in separately so we can have an independent accounting of performance impact (i'd expect the change from this to be minor overall though)

PR #42928, opened 2023-06-16
Print a warning when a message channel is used on the wrong thread.

dnfield:
Fixes https://github.com/flutter/flutter/issues/128746

Prints a warning the first time a platform channel sends a message from the wrong thread with instructions/link to the site about how to fix this.


dnfield:
Most of this change is related to making logging/thread checking testable.

shell/common/shell.cc
```diff
@@ -992,11 +992,27 @@ void Shell::OnPlatformViewSetViewportMetrics(const ViewportMetrics& metrics) {
...
 }
 
+std::set<std::string> Shell::misbehaving_message_channels_ = {};
+
 // |PlatformView::Delegate|
 void Shell::OnPlatformViewDispatchPlatformMessage(
     std::unique_ptr<PlatformMessage> message) {
   FML_DCHECK(is_setup_);
-  FML_DCHECK(task_runners_.GetPlatformTaskRunner()->RunsTasksOnCurrentThread());
+  if (!task_runners_.GetPlatformTaskRunner()->RunsTasksOnCurrentThread()) {
```
dnfield, line 1001: The heart of the change.

Previously this was only a DCHECK, which helps engine developers but not plugin developers/users.

fml/memory/thread_checker.h
```diff
@@ -67,7 +78,10 @@ class ThreadChecker final {
     return is_creation_thread_current;
   }
 
+  static void DisableNextThreadCheckFailure() { disable_next_ = true; }
```
gaaclarke, line 81: This isn't safe since we can't deterministically know what thread check is next, so we can't reasonably disable it.

Can you make `disable_next_` thread local?

fml/logging.cc
```diff
@@ -72,61 +72,75 @@ LogMessage::LogMessage(LogSeverity severity,
   }
 }
 
+// static
+std::shared_ptr<std::ostringstream> LogMessage::test_stream_ = nullptr;
```
gaaclarke, line 76:
```suggestion
std::shared_ptr<std::ostringstream> LogMessage::capture_next_log_stream_ = nullptr;
```

It's probably better that the method name and the variable have the same verbiage.

gaaclarke, line 76: Sounds like you want this to be thread local too, no?

fml/logging.h
```diff
@@ -27,7 +27,10 @@ class LogMessage {
 
   std::ostream& stream() { return stream_; }
 
+  static void CaptureNextLog(const std::shared_ptr<std::ostringstream>& stream);
```
gaaclarke, line 30: This should probably be a free function in `fml` right?  The relationship between FML_LOG and LogMessage isn't clear.

shell/common/shell.cc
```diff
@@ -992,11 +992,27 @@ void Shell::OnPlatformViewSetViewportMetrics(const ViewportMetrics& metrics) {
   }
 }
 
+std::set<std::string> Shell::misbehaving_message_channels_ = {};
```
gaaclarke, line 995: This isn't threadsafe either.  In the error case you are talking to this on multiple threads.

gaaclarke (changes requested):
This is on the right track.  I just want us to be careful and do the threading right.  I don't want use to crash in weird ways just so we can get better messaging.

fml/memory/thread_checker.h
dnfield: This is only meant to be used in tests so that we can intentionally violate a thread check.

Making it thread local sounds good.

shell/common/shell.cc
dnfield: I could introduce a mutex and then make this only for non-release.

dnfield: The problem with making it thread local is that it could still potentially print out very frequently in a plugin that's getting callbacks on new threads each time.

dnfield: I'm just doing this per shell. In 99% of cases there will only be one shell anyway, and if someone gets duplicate messages due to more than one shell that's probably fine.

fml/logging.cc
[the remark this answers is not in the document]
gaaclarke: This still is doesn't conform to the c++ style guide.  Probably the easiest thing to do would be `std::optional<std::ostringstream&>`.  You could technically get a dangling pointer, but it's up do you if you think it's worth having a raii wrapper.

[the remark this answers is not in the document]
dnfield: It's thread_local now right?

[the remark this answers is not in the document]
gaaclarke: The current state of your code?  yea.

gaaclarke (approved):
LGTM, thanks!

PR #27342, opened 2021-07-12
Add the ability to change the CanvasKit url at runtime.

hterkelsen:
Allows setting the CanvasKit URL at runtime. Useful in tests to allow using a local version of CanvasKit and reducing flakiness.

This will allow us to cache the CanvasKit files on LUCI: https://github.com/flutter/flutter/issues/86125

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [Flutter Style Guide] and the [C++, Objective-C, Java style guides].
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I added new tests to check the change I am making or feature I am adding, or Hixie said the PR is test-exempt. See [testing the engine] for instructions on
writing and running engine tests.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I signed the [CLA].
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/flutter/wiki/Tree-hygiene#overview
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[C++, Objective-C, Java style guides]: https://github.com/flutter/engine/blob/master/CONTRIBUTING.md#style
[testing the engine]: https://github.com/flutter/flutter/wiki/Testing-the-engine
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/wiki/Chat

PR #42941, opened 2023-06-16
[web] Removes patchCanvasKitModule.

ditman:
**This must land _after_ https://github.com/flutter/flutter/pull/129032**

Flutter web uses requireJS in `debug` mode to assemble a DDC-compiled app from a bunch of small files ("modules").

This caused that `canvaskit.js` (then, but probably all other modules that used a browserify-like loading header) didn't work because it attempted to use the `define` function provided by Flutter's instance of `requireJS` (which kept the defined modules private, rather than as globals on the page, as the users of the JS expected).

A [fix](https://github.com/flutter/engine/pull/27342) was added to `flutter/engine` to trick loaders into *not* using the `requireJS` module loader, but a recent change in the fix's js-interop layer *subtly* changed its JS output on the page (objects went from `undefined` to `null`), causing this:

* https://github.com/flutter/flutter/issues/126131 (and others)

After flutter/flutter#129032, the engine fix shouldn't be required anymore, so this PR removes it.

## Issues

* Fixes https://github.com/flutter/flutter/issues/126131 (and possibly others)

## Testing

* Manually tested with some test apps, and miscellanous JS scripts as reported by users.

## Pre-launch Checklist

- [x] I signed the [CLA].


hterkelsen (approved):
LGTM

mdebbar (approved):

ditman:
I think I'll test this by initializing an engine (ensuring that it's downloaded canvaskit) and then checking that `module` and `module.exports` don't exist.

ditman:
Added a test for the change.

(Also rename other initialization tests inside the initialization directory to not start with initialization, because there's too many initializations in those paths!)

PR #42960, opened 2023-06-17
[macOS] Implement unobstructed platform views

knopp:
Fixes https://github.com/flutter/flutter/issues/129073

## Changes to Embedder API

Introduced `FlutterRegion` to represent arbitrary region:
```cpp
typedef struct {
  /// The size of this struct. Must be sizeof(FlutterRegion).
  size_t struct_size;
  /// Number of rectangles in the region.
  size_t rects_count;
  /// The rectangles that make up the region.
  FlutterRect* rects;
} FlutterRegion;
```

Note that this is identical to struct `FlutterDamage` with more generic naming. Maybe down the line we could deprecate `FlutterDamage` and use `FlutterRegion` instead.

Introduced `FlutterBackingStorePresentInfo`:
```cpp

typedef struct {
  size_t struct_size;

  /// The area of the backing store that contains Flutter contents. Pixels
  /// outside of this area are transparent and the embedder may choose not
  /// to render them. Coordinates are in physical pixels.
  FlutterRegion* paint_region;
} FlutterBackingStorePresentInfo;
```
In future this struct may also contain more precise hit test region (when framework supports it) and/or information relevant to partial repaint (buffer damage, frame damage).

Added a `backing_store_present_info` field to `FlutterLayer`:

```cpp
typedef struct {
  ...
  /// Extra information for the backing store that the embedder may
  /// use during presentation.
  FlutterBackingStorePresentInfo* backing_store_present_info;
} FlutterLayer;
```

## Changes to the macOS embedder

This PR adds support for `FLTEnableSurfaceDebugInfo` flag in main bundle `Info.plist` that enables visual indicators of overlay layers.

## Example of unobstructed platform views

https://github.com/flutter/flutter/assets/96958/09a75eee-316b-4d53-a8b4-d6bb4e1e52f7

## Pre-launch Checklist

- [X] I signed the [CLA].

### DIFF
--- a/shell/platform/darwin/macos/framework/Source/FlutterCompositor.mm
+++ b/shell/platform/darwin/macos/framework/Source/FlutterCompositor.mm
@@ -60,6 +60,7 @@ bool FlutterCompositor::Present(FlutterViewId view_id,
         FlutterBackingStorePresentInfo* present_info = layer->backing_store_present_info;
         if (present_info != nullptr && present_info->paint_region != nullptr) {
           auto paint_region = present_info->paint_region;
+          // Safe because the size of FlutterRect is not expected to change.
           info.paintRegion = std::vector<FlutterRect>(
               paint_region->rects, paint_region->rects + paint_region->rects_count);
         }
--- a/shell/platform/darwin/macos/framework/Source/FlutterCompositor.mm
+++ b/shell/platform/darwin/macos/framework/Source/FlutterCompositor.mm
@@ -57,6 +57,12 @@ bool FlutterCompositor::Present(FlutterViewId view_id,
         info.surface = surface;
         info.offset = CGPointMake(layer->offset.x, layer->offset.y);
         info.zIndex = i;
+        FlutterBackingStorePresentInfo* present_info = layer->backing_store_present_info;
+        if (present_info != nullptr && present_info->paint_region != nullptr) {
+          auto paint_region = present_info->paint_region;
+          info.paintRegion = std::vector<FlutterRect>(
+              paint_region->rects, paint_region->rects + paint_region->rects_count);
+        }
         [surfaces addObject:info];
       }
     }
--- a/shell/platform/darwin/macos/framework/Source/FlutterSurfaceManager.mm
+++ b/shell/platform/darwin/macos/framework/Source/FlutterSurfaceManager.mm
@@ -29,6 +29,9 @@
 
   // Currently visible layers.
   NSMutableArray<CALayer*>* _layers;
+
+  NSNumber* _enableSurfaceDebugInfo;
+  CATextLayer* _infoLayer;
 }
 
 /**
@@ -37,6 +40,56 @@
 - (void)commit:(NSArray<FlutterSurfacePresentInfo*>*)surfaces;
 
 @end
+
+static NSColor* GetBorderColorForLayer(int layer) {
+  NSArray* colors = @[
+    [NSColor yellowColor],
+    [NSColor cyanColor],
+    [NSColor magentaColor],
+    [NSColor greenColor],
+    [NSColor purpleColor],
+    [NSColor orangeColor],
+    [NSColor blueColor],
+  ];
+  return colors[layer % colors.count];
+}
+
+static void UpdateContentSubLayers(CALayer* layer,
+                                   IOSurfaceRef surface,
+                                   CGFloat scale,
+                                   CGSize surfaceSize,
+                                   NSColor* borderColor,
+                                   const std::vector<FlutterRect>& covered_area) {
+  while (layer.sublayers.count > covered_area.size()) {
+    [layer.sublayers.lastObject removeFromSuperlayer];
+  }
+
+  while (layer.sublayers.count < covered_area.size()) {
+    CALayer* newLayer = [CALayer layer];
+    [layer addSublayer:newLayer];
+  }
+
+  for (size_t i = 0; i < covered_area.size(); i++) {
+    CALayer* subLayer = [layer.sublayers objectAtIndex:i];
+    const auto& rect = covered_area[i];
+    subLayer.frame = CGRectMake(rect.left / scale, rect.top / scale,
+                                (rect.right - rect.left) / scale, (rect.bottom - rect.top) / scale);
+
+    double w = surfaceSize.width;
+    double h = surfaceSize.height;
+
+    subLayer.contentsRect = CGRectMake(rect.left / w, rect.top / h, (rect.right - rect.left) / w,
+                                       (rect.bottom - rect.top) / h);
+
+    if (borderColor != nil) {
+      // Visualize sublayer
+      subLayer.borderColor = borderColor.CGColor;
+      subLayer.borderWidth = 1.0;
+    }
+
+    subLayer.contents = (__bridge id)surface;
+  }
+}
 
 @implementation FlutterSurfaceManager
 
@@ -77,6 +130,17 @@
   return surface;
 }
 
+- (BOOL)enableSurfaceDebugInfo {
+  if (_enableSurfaceDebugInfo == nil) {
+    _enableSurfaceDebugInfo =
+        [[NSBundle mainBundle] objectForInfoDictionaryKey:@"FLTEnableSurfaceDebugInfo"];
+    if (_enableSurfaceDebugInfo == nil) {
+      _enableSurfaceDebugInfo = @NO;
+    }
+  }
+  return [_enableSurfaceDebugInfo boolValue];
+}
+
 - (void)commit:(NSArray<FlutterSurfacePresentInfo*>*)surfaces {
   FML_DCHECK([NSThread isMainThread]);
 
@@ -100,15 +164,37 @@
     [_layers addObject:layer];
   }
 
+  bool enableSurfaceDebugInfo = self.enableSurfaceDebugInfo;
+
   // Update contents of surfaces.
   for (size_t i = 0; i < surfaces.count; ++i) {
     FlutterSurfacePresentInfo* info = surfaces[i];
     CALayer* layer = _layers[i];
     CGFloat scale = _containingLayer.contentsScale;
-    layer.frame = CGRectMake(info.offset.x / scale, info.offset.y / scale,
-                             info.surface.size.width / scale, info.surface.size.height / scale);
-    layer.contents = (__bridge id)info.surface.ioSurface;
+    if (i == 0) {
+      layer.frame = CGRectMake(info.offset.x / scale, info.offset.y / scale,
+                               info.surface.size.width / scale, info.surface.size.height / scale);
+      layer.contents = (__bridge id)info.surface.ioSurface;
+    } else {
+      layer.frame = CGRectZero;
+      NSColor* borderColor = enableSurfaceDebugInfo ? GetBorderColorForLayer(i - 1) : nil;
+      UpdateContentSubLayers(layer, info.surface.ioSurface, scale, info.surface.size, borderColor,
+                             info.paintRegion);
+    }
     layer.zPosition = info.zIndex;
+  }
+
+  if (enableSurfaceDebugInfo) {
+    if (_infoLayer == nil) {
+      _infoLayer = [[CATextLayer alloc] init];
+      [_containingLayer addSublayer:_infoLayer];
+      _infoLayer.fontSize = 15;
+      _infoLayer.foregroundColor = [NSColor yellowColor].CGColor;
+      _infoLayer.frame = CGRectMake(15, 15, 300, 100);
+      _infoLayer.contentsScale = _containingLayer.contentsScale;
+      _infoLayer.zPosition = 100000;
+    }
+    _infoLayer.string = [NSString stringWithFormat:@"Surface count: %li", _layers.count];
   }
 }
 
--- a/shell/platform/darwin/macos/framework/Source/FlutterSurfaceManager.mm
+++ b/shell/platform/darwin/macos/framework/Source/FlutterSurfaceManager.mm
@@ -30,6 +30,8 @@
   // Currently visible layers.
   NSMutableArray<CALayer*>* _layers;
 
+  // Whether to highlight borders of overlay surfaces. Determined by
+  // FLTEnableSurfaceDebugInfo value in main bundle Info.plist.
   NSNumber* _enableSurfaceDebugInfo;
   CATextLayer* _infoLayer;
 }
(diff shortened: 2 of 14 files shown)


loic-sharma:
> Note that this is identical to struct `FlutterDamage` with more generic naming...

Are we allowed to rename structs in the embedder API? The embedder API allows [renaming members](https://github.com/flutter/engine/blob/08aaa88bf67fcd5a18d4309d89c1e940a544e49a/shell/platform/embedder/embedder.h#L39-L41), I **_think_** renaming a struct wouldn't break ABI.

chinmaygarde:
The ABI remains stable but the API breaks. I'd rather not break. You could
use typedefs.

shell/platform/embedder/embedder_layers.h
```diff
@@ -42,6 +43,10 @@ class EmbedderLayers {
       mutations_referenced_;
   std::vector<std::unique_ptr<std::vector<const FlutterPlatformViewMutation*>>>
       mutations_arrays_referenced_;
+  std::vector<std::unique_ptr<FlutterBackingStorePresentInfo>>
+      present_info_refereced_;
```
loic-sharma, line 47:
```suggestion
      present_info_referenced_;
```

shell/platform/embedder/embedder.h
```diff
@@ -1568,6 +1568,24 @@ typedef enum {
   kFlutterLayerContentTypePlatformView,
 } FlutterLayerContentType;
 
+typedef struct {
+  /// The size of this struct. Must be sizeof(FlutterRegion).
+  size_t struct_size;
+  /// Number of rectangles in the region.
+  size_t num_rects;
```
loic-sharma, line 1575: We've been inconsistent, but I believe the preferred naming pattern is `rects_count`

loic-sharma:
@chinmaygarde I deleted my comment as I figured it'd better to keep `FlutterDamage` and `FlutterRegion` separate structs in case we ever need to add new members to `FlutterDamage` that don't make sense in `FlutterRegion`.

knopp:
> The ABI remains stable but the API breaks. I'd rather not break. You could use typedefs.
> […](#)

`FlutterDamage` has field called `damage` while `FlutterRegion` has field called `rects`. Not sure how the typedef would help unless we union the fields, at which point maybe it's better to just keep the structs separate.

shell/platform/embedder/embedder.h
```diff
@@ -1587,6 +1605,10 @@ typedef struct {
   FlutterPoint offset;
   /// The size of the layer (in physical pixels).
   FlutterSize size;
+
+  /// Extra information for the backing store that the embedder may
+  /// use during presentation.
+  FlutterBackingStorePresentInfo* backing_store_present_info;
```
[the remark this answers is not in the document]
knopp: Eventually, we'll also need to pass preexisting damage (from embedder client to embedder) and `buffer_damage` + `frame_damage` (from embedder to client) to support partial repaint. I'm not entirely sure where these belong. Technically we don't even need them per buffer, they are only once for all layers. But I haven't yet figured out partial repaint with multiple layers, the tricky part there is we may need to make sure that we preserve z-order.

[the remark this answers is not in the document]
loic-sharma: I'm not very familiar with this area so I'll let others provide guidance :)

> I'm not sure why this would break the ABI though?

Apologies, I misspoke. My concern is that `FlutterBackingStore`'s contains a [union of structs](https://github.com/flutter/engine/blob/08aaa88bf67fcd5a18d4309d89c1e940a544e49a/shell/platform/embedder/embedder.h#L1542-L1553). We would break the ABI if we add a member after this union and then add members to structs within that union.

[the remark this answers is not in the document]
knopp: The client sets struct size though. So if we add a new union field and client sets `backing_store.size = size_of(FlutterBackingStoreWithoutNewMember)` we could detect it and convert the struct after the callback.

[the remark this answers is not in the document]
knopp: I'm guessing though a complication here is that we pass the struct later in present callback, but not the original struct - a copy that embedder makes. And it gets hairy if the embedder wants to update fields after the union. Still doable, but we may internally need to keep union {  FlutterBackingStore, FlutterBackingStoreLegacy } around and access them based on client provided size.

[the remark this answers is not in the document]
knopp: Example of adding fields to backing store the preserve abi compatibility: 
https://gist.github.com/knopp/9dbabd64a0501113ad73464463a51031. Instead of using the backing store directly the embedder would use a proxy class that is aware of different backing store versions.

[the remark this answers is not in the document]
knopp: I added commit moving the new field into `FlutterBackingStore` in a way that preverves the ABI even if we need to add another field to the union.

loic-sharma:
@knopp FYI, Chris Bracken is out on vacation until July 10th and will likely not be able to review until then. Apologies for the delay!

knopp:
> @knopp FYI, Chris Bracken is out on vacation until July 10th and will likely not be able to review until then. Apologies for the delay!

Thanks for the heads-up!

cbracken:
I'm back and taking a look at this and the others! Thanks so much for sending them!

shell/platform/embedder/embedder.h
```diff
@@ -1527,6 +1527,24 @@ typedef enum {
   kFlutterBackingStoreTypeSoftware2,
 } FlutterBackingStoreType;
 
+typedef struct {
```
cbracken, line 1530: Add a doc comment for the struct.

shell/platform/embedder/embedder.h
```diff
@@ -1527,6 +1527,24 @@ typedef enum {
...
+  /// The size of this struct. Must be sizeof(FlutterRegion).
+  size_t struct_size;
+  /// Number of rectangles in the region.
+  size_t rects_count;
+  /// The rectangles that make up the region.
+  FlutterRect* rects;
+} FlutterRegion;
+
+typedef struct {
```
cbracken, line 1539: Similar comment here. This is really just for IDE hinting for folks who work against the embedder API; you've already documented the important stuff on the field.

shell/platform/embedder/embedder_external_view_embedder.cc
```diff
@@ -238,8 +238,16 @@ void EmbedderExternalViewEmbedder::SubmitFrame(
       // platform view.
       if (external_view->HasEngineRenderedContents()) {
         const auto& exteral_render_target = matched_render_targets.at(view_id);
+        const auto& external_view = pending_views_.at(view_id);
+        auto rect_list =
+            external_view->GetEngineRenderedContentsRegion(SkRect::MakeIWH(
+                pending_frame_size_.width(), pending_frame_size_.height()));
+        std::vector<SkIRect> rects;
```
cbracken, line 245:
```suggestion
        std::vector<SkIRect> rects;
        rects.reserve(rect_list.size());
```

shell/platform/embedder/embedder_layers.cc
```diff
@@ -35,6 +38,33 @@ void EmbedderLayers::PushBackingStoreLayer(const FlutterBackingStore* store) {
   layer.size.width = transformed_layer_bounds.width();
   layer.size.height = transformed_layer_bounds.height();
 
+  auto paint_region_rects = std::make_unique<std::vector<FlutterRect>>();
+  paint_region_rects->reserve(paint_region_vec.size());
+
+  for (const auto& rect : paint_region_vec) {
+    auto transformed_rect =
+        root_surface_transformation_.mapRect(SkRect::Make(rect));
+    paint_region_rects->push_back(FlutterRect{
```
cbracken, line 47: I _think_ this could be `emplace_back(transformed_rect.x(), ...)`?

cbracken (commented):
Overall looks great - I'm still looking at the embedder API side changes. Sent the rest of the comments.

shell/platform/darwin/macos/framework/Source/FlutterSurfaceManager.mm
```diff
@@ -77,6 +130,17 @@ - (FlutterSurface*)surfaceForSize:(CGSize)size {
   return surface;
 }
 
+- (BOOL)enableSurfaceDebugInfo {
```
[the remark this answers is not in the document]
cbracken: (FWIW I can definitely see the argument for keeping it around until https://github.com/flutter/engine/pull/43301 is complete)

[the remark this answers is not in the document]
knopp: I think the usefulness of this right now overweights the amount of code needed to get this working (which isn't really a whole lot).

shell/platform/embedder/embedder_layers.cc
knopp: Flutter rect doesn't have a 4 argument constructor so the compiler is not too happy about it.

shell/platform/embedder/embedder.h
```diff
@@ -1632,6 +1653,10 @@ typedef struct {
     // The description of the Vulkan backing store.
     FlutterVulkanBackingStore vulkan;
   };
+
+  /// Extra information set by the engine for the backing store that the
+  /// embedder may use during presentation.
+  FlutterBackingStorePresentInfo* present_info;
```
cbracken, line 1659: Adding this at the end isn't safe since the offset of this field isn't guaranteed given that both (a) things might get added to this union that make it consume more space and (b) things might get added to structs/unions in the structs in this union (transitively down) that might affect the size of this union.

cbracken (changes requested):
Just a quick note on the embedder.h changes that I'll send out now while I continue to review.

shell/platform/darwin/macos/framework/Source/FlutterSurfaceManager.mm
```diff
@@ -38,6 +43,59 @@ - (void)commit:(NSArray<FlutterSurfacePresentInfo*>*)surfaces;
...
+
+/// Creates sublayers for given layer, each one displaying a portion of the
+/// of the surface determined by a rectangle in the provided paint region.
+static void UpdateContentSubLayers(CALayer* layer,
+                                   IOSurfaceRef surface,
+                                   CGFloat scale,
+                                   CGSize surfaceSize,
+                                   NSColor* borderColor,
+                                   const std::vector<FlutterRect>& paintRegion) {
+  while (layer.sublayers.count > paintRegion.size()) {
```
cbracken, line 67:
```suggestion
  // Adjust sublayer count to paintRegion size.
  while (layer.sublayers.count > paintRegion.size()) {
```

shell/platform/embedder/embedder.h
knopp: The idea is that adding things to union would produce another version. The logic in embedder_backing_store would ensure that we're not exposing fields outside of current version.

This would get tricky if the backing stores themselves or inner fields change. I assume that was not likely, but I can see in the PR that I was wrong.

We have backed ourself into a corner with the union not having pointer members :-/

I can put the present info back to the `FlutterLayer`. I think ti would be less disruptive than introducing `FlutterBackingStore2`.

shell/platform/embedder/embedder_backing_store.cc
```diff
@@ -0,0 +1,17 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "flutter/shell/platform/embedder/embedder_backing_store.h"
+
+namespace flutter {
+
+bool EmbedderBackingStore::SetPresetInfo(FlutterBackingStorePresentInfo* info) {
```
cbracken, line 9:
```suggestion
bool EmbedderBackingStore::SetPresentInfo(FlutterBackingStorePresentInfo* info) {
```

shell/platform/embedder/embedder_backing_store.h
```diff
@@ -0,0 +1,94 @@
...
+    // By default assume the oldest version of backing store. Embedder can
+    // override this by setting the struct_size field.
+    backing_store_ = {
+        .struct_size = sizeof(FlutterBackingStoreVersion1),
+    };
+  }
+
+  bool IsValid() const {
+    return backing_store_.struct_size == sizeof(FlutterBackingStoreVersion1) ||
+           backing_store_.struct_size == sizeof(FlutterBackingStoreVersion2);
```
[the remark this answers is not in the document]
cbracken: I think even the new backing store case is tricky since once there are fields after it, it's possible for two versions to have the same number of bytes. We'd really need an explicit version field to be safe.

I think the cumulative level of danger of accidentally breaking things is a bit scary here -- people touching FlutterMetalBackingStore likely aren't thinking about the effects their change are likely to have in other places in the embedder API and ideally people reviewing shouldn't need to keep a pile of edge-cases paged into their brains. We're looking at adding tests that would watch for such breakages in the meantime. The original API here (and a few other places with nested structs/unions) is pretty unfortunate.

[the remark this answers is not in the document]
knopp: No argument here. 
Conceptually this is also problematic because the backing store is populated by client and then modified by engine, which just seems wrong.

shell/platform/embedder/embedder_backing_store.h
```diff
@@ -0,0 +1,94 @@
...
+    return version1_.software2;
+  }
+
+  const FlutterMetalBackingStore& metal() const { return version1_.metal; }
+
+  const FlutterVulkanBackingStore& vulkan() const { return version1_.vulkan; }
+
+  void SetDidUpdate(bool did_update) { version1_.did_update = true; }
+
+  bool SetPresetInfo(FlutterBackingStorePresentInfo* info);
```
cbracken, line 75:
```suggestion
  bool SetPresentInfo(FlutterBackingStorePresentInfo* info);
```

knopp:
I have removed the commit [Move FlutterBackingStorePresentInfo into FlutterBackingStore]. This moves `backing_store_present_info` back to `FlutterLayer`. Question is, whether this should be another union - maybe down the line we'll have additional present information for other layers as well?

shell/platform/embedder/embedder_backing_store.cc
```diff
@@ -0,0 +1,17 @@
...
+  if (backing_store_.struct_size == sizeof(FlutterBackingStoreVersion2)) {
+    version2_.present_info = info;
+    return true;
+  } else {
+    return false;
+  }
```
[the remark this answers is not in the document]
knopp: EmbedderBackingStore is gone.

shell/platform/embedder/embedder_backing_store.cc
knopp: EmbedderBackingStore has been removed.

shell/platform/embedder/embedder_backing_store.h
knopp: EmbedderBackingStore has been removed.

shell/platform/embedder/embedder.h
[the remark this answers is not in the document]
knopp: And now I have removed the commit. It was a bad idea. I don't think there is any reasonable way to extend `FlutterBackingStore` right now.

shell/platform/darwin/macos/framework/Source/FlutterCompositor.mm
```diff
@@ -57,6 +57,12 @@
         info.surface = surface;
         info.offset = CGPointMake(layer->offset.x, layer->offset.y);
         info.zIndex = i;
+        FlutterBackingStorePresentInfo* present_info = layer->backing_store_present_info;
+        if (present_info != nullptr && present_info->paint_region != nullptr) {
+          auto paint_region = present_info->paint_region;
+          info.paintRegion = std::vector<FlutterRect>(
+              paint_region->rects, paint_region->rects + paint_region->rects_count);
```
cbracken, line 64: Not sure if it's worth commenting here (or somewhere) that this is safe because FlutterRect's size is never expected to change in the future. (It has no struct_size member as well, for that reason).

shell/platform/embedder/embedder.h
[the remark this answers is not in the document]
cbracken: Agreed. I think there's no way to safely modify that struct at this point and it should be locked down.

cbracken (approved):
lgtm - thanks for your patience on this one, in particular on the embedder API changes.

shell/platform/darwin/macos/framework/Source/FlutterSurfaceManager.mm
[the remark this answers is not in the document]
knopp: I'd leave it for now. Would possibly revisit if/when iOS moves towards embedder API.

PR #42962, opened 2023-06-17
Revert "[Impeller] correct default PSO pixel format and sample count."

jonahwilliams:
Reverts flutter/engine#42902

This is breaking iOS builds:

https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8778085761769326737/+/u/run_platform_views_scroll_perf_non_intersecting_impeller_ios__timeline_summary/test_stdout


fluttergithubbot (approved):

chinmaygarde (approved):

jonahwilliams:
The reason this is breaking is that the default pixel format can change in the presence of wide gamut + platform views. a potential second surface will have a different pixel value (?) and so the lookup fails. The "fix" is to cache the default options object and reuse that, though if we ever do wide gamut for Vulkan we're going to have a heck of time with the PSO variations.

PR #42991, opened 2023-06-19
Basic view management for engine classes

dkwingsmt:
_This PR is part of the multiview engine project. For a complete roadmap, see [this doc](https://docs.google.com/document/d/10APhzRDR7XqjWdbYWpFfKur7DPiz_HvSKNcLvcyA9vg/edit?resourcekey=0-DfGcg4-XWRMMZF__C1nmcA)._

------

This PR adds view management to all engine classes that need it. View management here basically means `AddView` and `RemoveView` methods, and most importantly, how to handle the implicit view.

The implicit view is a special view that's handled differently than all the other "regular views", since it keeps the behavior of the current single view of Flutter. Detailed introduction can be found in `Settings.implicit_view_enabled`.

The following two graphs show the difference between initializing with/without the implicit view and creating regular views.

image

image

image

## Pre-launch Checklist

- [ ] I signed the [CLA].


common/settings.h
```diff
@@ -224,6 +224,10 @@ struct Settings {
   // must be available to the application.
   bool enable_vulkan_validation = false;
 
+  // Enable the implicit view, so that the shell should create a view with
+  // kFlutterImplicitViewId without the platform calling FlutterEngineAddView.
+  bool enable_implicit_view = true;
```
loic-sharma, line 229: Can we give a brief explanation on what the implicit view is (has view ID 0, allows app to make single view assumptions, used for mobile apps) and mention that `PlatformDispatcher.implicitView` will be non-null if this is true?

lib/ui/hooks.dart
```diff
@@ -4,6 +4,20 @@
 
 part of dart.ui;
 
+@pragma('vm:entry-point')
+void _addView(
+  int id,
+) {
```
loic-sharma, line 10: Does the arguments need to be on multiple lines? It looks like this fits under 80 characters:

```suggestion
void _addView(int id) {
```

lib/ui/hooks.dart
```diff
@@ -4,6 +4,20 @@
...
+  PlatformDispatcher.instance._addView(id);
+}
+
+@pragma('vm:entry-point')
+void _removeView(
+  int id,
+) {
```
loic-sharma, line 17:
```suggestion
void _removeView(int id) {
```

lib/ui/natives.dart
```diff
@@ -135,3 +135,7 @@ _ScheduleImmediateClosure _getScheduleMicrotaskClosure() => _scheduleMicrotask;
 // rendering.
 @pragma('vm:entry-point')
 bool _impellerEnabled = false;
+
+// Used internally to indicate whether the platform uses the implicit view.
```
loic-sharma, line 139: Consider:

```suggestion
// Used internally to indicate whether the embedder enables the implicit view.
```

lib/ui/platform_dispatcher.dart
```diff
@@ -244,6 +248,28 @@ class PlatformDispatcher {
     _onMetricsChangedZone = Zone.current;
   }
 
+  FlutterView _createView(int id) {
+    return FlutterView._(id, this);
+  }
+
+  void _addView(int id) {
+    assert(id != _kImplicitViewId, 'The implicit view #$id can not be added.');
```
loic-sharma, line 256: This means that even if the implicit view is disabled, the embedder cannot create a view with ID 0. Is that a necessary limitation?

loic-sharma: Also see this comment: https://github.com/flutter/engine/blob/e6b8292705a832a56ab31b13fd819030070962c0/shell/platform/darwin/macos/framework/Source/FlutterView.h#L21

dkwingsmt: This is my intended design. The idea is that, since the implicit view and the regular views have different models, some code might distinguish them by testing the view ID, since they might not have the signal on whether the id-0 view is an implicit one. It's much safer (and actually easier to understand) to just preserve ID 0 for the implicit view.

common/settings.h
dkwingsmt: I wrote some doc. Let me know what you think :)

lib/ui/platform_dispatcher.dart
```diff
@@ -283,9 +309,7 @@ class PlatformDispatcher {
   ) {
     final _ViewConfiguration previousConfiguration =
         _viewConfigurations[id] ?? const _ViewConfiguration();
-    if (!_views.containsKey(id)) {
-      _views[id] = FlutterView._(id, this);
-    }
+    _views.putIfAbsent(id, () => _createView(id));
```
goderbauer, line 312: We should probably remove this? You shouldn't be able to update metrics for non-existing views.

common/settings.h
```diff
@@ -224,6 +224,32 @@ struct Settings {
...
+  // The implicit view is a compatibility mode to help the transition from
+  // single-view APIs to multi-view APIs, which use different models for views.
+  // Although single-view APIs will eventually be replaced by their multi-view
+  // variants, during the deprecation period, the single-view APIs will coexist
+  // with and work with the multi-view APIs as if the other views don't exist.
+  // To achieve this, the platform can enable "implicit view", which is a
+  // special view that keeps all behaviors of "the view" in a legacy
+  // single-view Flutter app, allowing single-view APIs to continue working,
+  // while new-style "regular" views created by new ways must be operated by
+  // multi-view APIs.
```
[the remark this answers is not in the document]
goderbauer: Maybe document that the default value is true for backwards compatibility, but multi-window capable platforms likely want to set this to false.

lib/ui/platform_dispatcher.dart
goderbauer: I don't think we should give these IDs any semantic meaning like that publicly. These are just identifiers identifying a FlutterView, but the exact value doesn't have a meaning. If we come across a use case where it is important to distinguish between implicit view and regular views, we should add an explicit API for that (we already have this in the dart:ui world with PlatformDispatcher.implicitView).

lib/ui/platform_dispatcher.dart
```diff
@@ -325,16 +352,7 @@ class PlatformDispatcher {
...
 
-    final FlutterView? view = _views[id];
-    if (id == _kImplicitViewId && view == null) {
-      // TODO(goderbauer): Remove the implicit creation of the implicit view
-      //   when we have an addView API and the implicit view is added via that.
-      _views[id] = FlutterView._(id, this, viewConfiguration);
-    } else {
-      assert(view != null);
-      view!._viewConfiguration = viewConfiguration;
-    }
```
goderbauer, line 336: Good riddance!

lib/ui/window/platform_configuration.cc
```diff
@@ -35,7 +35,15 @@ PlatformConfigurationClient::~PlatformConfigurationClient() {}
...
 PlatformConfiguration::PlatformConfiguration(
     PlatformConfigurationClient* client)
-    : client_(client) {}
+    : client_(client) {
+  if (client_->ImplicitViewEnabled()) {
+    // Add PlatformConfiguration's window now, but don't add dart:ui's View
+    // here. The dart:ui needs another way to add the implicit view
+    // synchronously so that the view is available before the main function.
+    // See _implicitViewEnabled in natives.dart.
+    AddWindowRecord(kFlutterImplicitViewId);
```
goderbauer, line 44: (unrelated to this PR: We should come up with a plan to clean up the Window/View naming in the engine, it's gonna get really confusing once we introduce actual windows. I'll add an item for this to the work tracker.)

shell/common/rasterizer_unittests.cc
```diff
@@ -29,6 +29,8 @@ using testing::ReturnRef;
 
 namespace flutter {
 namespace {
+
+constexpr int64_t kDefaultViewId = 0;
```
goderbauer, line 33: Should we stick with the implicitview naming scheme here?

shell/common/shell.h
```diff
@@ -298,6 +298,28 @@ class Shell final : public PlatformView::Delegate,
   ///
   bool IsSetup() const;
 
+  /// @brief  Add a non-implicit view.
```
goderbauer, line 301: I am actually not sure what the style is, but the docs of the next two methods all use this grammatical form:

```suggestion
  /// @brief  Adds a non-implicit view.
```

shell/common/shell.h
```diff
@@ -298,6 +298,28 @@ class Shell final : public PlatformView::Delegate,
...
+  ///         because the rasterizer thread can be blocked by the platform
+  ///         thread to render platform views.
+  ///
+  ///         The implicit view is added internally on Shell initialization
+  ///         depending on `Settings.enable_implicit_view`. Trying to add
+  ///         `kFlutterImplicitViewId` is a no-op with warning.
+  /// @param view_id
+  void AddView(int64_t view_id);
+
+  /// @brief  Remove a non-implicit view.
```
goderbauer, line 314:
```suggestion
  /// @brief  Removes a non-implicit view.
```

common/settings.h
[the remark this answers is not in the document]
dkwingsmt: > Can we rephrase this to not imply that single-view (or implicit view mode) is going away? implicit view mode is just a different mode of operation and we will likely have to support that "forever" 

Good point. I've rewritten this doc with this idea in mind. Can you take a look?

common/settings.h
```diff
@@ -224,6 +226,38 @@ struct Settings {
...
 
+  // Enable the implicit view.
+  //
+  // The implicit view mode is a compatibility mode to help the transition from
+  // the older single-view APIs to the newer multi-view APIs. The two sets of
+  // APIs use different models for view management. By enabling the implicit
+  // view, single-view APIs can operate a special view as if other views don't
+  // exist.
+  //
+  // In the regular multi-view model where implicit view is disabled, all views
```
loic-sharma, line 237:
```suggestion
  // In the regular multi-view model where the implicit view is disabled, all views
```

common/settings.h
```diff
@@ -224,6 +226,38 @@ struct Settings {
...
+  // this view.
+  //
+  // The single-view APIs, which are APIs that do not specify view IDs, operate
+  // the implicit view. The multi-view APIs can operate all views, including the
+  // implicit view with the correct ID (kFlutterImplicitViewId), unless
+  // specified otherwise,
+  //
+  // The enable_implicit_view defaults to true. Multi-view apps will probably
+  // want to disable the implicit view eventually for the simplicity of a
+  // unified model.
```
loic-sharma, line 258: Long-term our plan is to keep the implicit view enabled on mobile platforms. I would remove the recommendation that the implicit view should be disabled:

```suggestion
  // The enable_implicit_view defaults to true.
```

common/settings.h
loic-sharma: Excellent comment. Marking this as resolved :)

common/settings.h
```diff
@@ -224,6 +226,38 @@ struct Settings {
...
+  // view, single-view APIs can operate a special view as if other views don't
+  // exist.
+  //
+  // In the regular multi-view model where implicit view is disabled, all views
+  // should be created by Shell::AddView before being used, and removed by
+  // Shell::RemoveView to signify that they are gone. If a view is added or
+  // removed, the framework (PlatformDispatcher) will be notified. New view IDs
+  // are always unique, never reused. Operating a non-existing view is an error.
+  //
+  // If implicit view is enabled, in addition to the "regular views" as above,
```
loic-sharma, line 243:
```suggestion
  // If the implicit view is enabled, in addition to the "regular views" as above,
```

lib/ui/platform_dispatcher.dart
```diff
@@ -249,6 +248,28 @@ class PlatformDispatcher {
     _onMetricsChangedZone = Zone.current;
   }
 
+  FlutterView _createView(int id) {
```
loic-sharma, line 251: It looks like only `_doAddView` calls `_createView`. Can we inline this logic in `_doAddView`?

lib/ui/platform_dispatcher.dart
```diff
@@ -249,6 +248,28 @@ class PlatformDispatcher {
...
+    return FlutterView._(id, this, const _ViewConfiguration());
+  }
+
+  void _addView(int id) {
+    assert(id != _implicitViewId, 'The implicit view #$id can not be added.');
+    _doAddView(id);
+    _invoke(onMetricsChanged, _onMetricsChangedZone);
+  }
+
+  void _doAddView(int id) {
```
loic-sharma, line 261: I think this would be a little easier to read without `_doAddView`:

```dart
  PlatformDispatcher._() {
    _setNeedsReportTimings = _nativeSetNeedsReportTimings;
    if (_implicitViewId != null) {
      _views[_implicitViewId] = FlutterView._(id, this, const _ViewConfiguration());
    }
  }

...

  void _addView(int id) {
    assert(id != _implicitViewId, 'The implicit view #$id can not be added.');
    assert(!_views.containsKey(id), 'View ID $id already exists.');
    FlutterView._(id, this, const _ViewConfiguration());
    _invoke(onMetricsChanged, _onMetricsChangedZone);
  }
```

I don't feel strongly about this, feel free to skip

lib/ui/window.dart
```diff
@@ -407,7 +407,7 @@ class SingletonFlutterWindow extends FlutterView {
     'This feature was deprecated after v3.7.0-32.0.pre.'
   )
   SingletonFlutterWindow._() : super._(
-    _kImplicitViewId,
+    _implicitViewId!,
```
loic-sharma, line 410: Will this result in an error if the implicit view is disabled, even if the user never uses the `window` global? We should be able to disable the implicit view without causing errors, even in phase 1. What are our other options?

lib/ui/window/platform_configuration.cc
```diff
@@ -35,7 +32,15 @@ PlatformConfigurationClient::~PlatformConfigurationClient() {}
 
 PlatformConfiguration::PlatformConfiguration(
     PlatformConfigurationClient* client)
-    : client_(client) {}
+    : client_(client) {
+  if (client_->ImplicitViewEnabled()) {
+    // Add PlatformConfiguration's window now, but don't add dart:ui's View
+    // here. The dart:ui needs another way to add the implicit view
+    // synchronously so that the view is available before the main function.
+    // See _implicitViewEnabled in natives.dart.
```
loic-sharma, line 40:
```suggestion
    // See _implicitViewId in natives.dart.
```

lib/ui/window/platform_configuration.cc
```diff
@@ -35,7 +32,15 @@ PlatformConfigurationClient::~PlatformConfigurationClient() {}
...
+    AddWindowRecord(kFlutterImplicitViewId);
```
loic-sharma, line 41: `AddWindowRecord` uses `library_` which isn't set until `DidCreateIsolate`. 

Can we move this logic to `DidCreateIsolate`?

lib/ui/window/platform_configuration.cc
```diff
@@ -77,12 +86,43 @@ void PlatformConfiguration::DidCreateIsolate() {
...
+
+void PlatformConfiguration::AddWindowRecord(int64_t view_id) {
+  windows_.emplace(
+      view_id, std::make_unique<Window>(library_, view_id,
+                                        ViewportMetrics{1.0, 0.0, 0.0, -1, 0}));
+}
+
+void PlatformConfiguration::AddView(int64_t view_id) {
+  FML_DCHECK(view_id != kFlutterImplicitViewId);
+  AddWindowRecord(view_id);
```
loic-sharma, line 101: Should we check that `dart_state` is valid before calling `AddWindowRecord`? Same comment for `RemoveView`'s logic

lib/ui/window/platform_configuration.h
```diff
@@ -268,6 +268,10 @@ class PlatformConfiguration final {
   ///
   void DidCreateIsolate();
 
+  void AddView(int64_t view_id);
```
loic-sharma, line 271: Please add docs to these new members

lib/ui/window/platform_configuration.cc
```diff
@@ -77,12 +86,43 @@ void PlatformConfiguration::DidCreateIsolate() {
...
                       Dart_GetField(library, tonic::ToDart("_reportTimings")));
 
-  // TODO(loicsharma): This should only be created if the embedder enables the
-  // implicit view.
-  // See: https://github.com/flutter/flutter/issues/120306
-  windows_.emplace(kImplicitViewId,
-                   std::make_unique<Window>(
-                       kImplicitViewId, ViewportMetrics{1.0, 0.0, 0.0, -1, 0}));
+  library_.Set(tonic::DartState::Current(),
+               Dart_LookupLibrary(tonic::ToDart("dart:ui")));
```
loic-sharma, line 90: Can we reuse the `library` created on [line 48](https://github.com/flutter/engine/pull/42991/files#diff-d9b11435542f935fb2c739d1abdf3c8f92b5f3bab797e614ea74324e52c91b98R48)?

```suggestion
               library);
```

runtime/runtime_controller.cc
```diff
@@ -115,11 +113,7 @@ std::unique_ptr<RuntimeController> RuntimeController::Clone() const {
 }
 
 bool RuntimeController::FlushRuntimeStateToIsolate() {
-  // TODO(dkwingsmt): Needs a view ID here (or platform_data should probably
-  // have multiple view metrics).
-  return SetViewportMetrics(kFlutterImplicitViewId,
-                            platform_data_.viewport_metrics) &&
-         SetLocales(platform_data_.locale_data) &&
+  return SetLocales(platform_data_.locale_data) &&
```
loic-sharma, line 116: It looks like this is only used by `RuntimeController::LaunchRootIsolate`. Should we send the implicit view's viewport metrics here to maintain backwards compatibility?

runtime/runtime_controller.h
```diff
@@ -163,6 +163,10 @@ class RuntimeController : public PlatformConfigurationClient {
   ///
   std::unique_ptr<RuntimeController> Clone() const;
 
+  bool AddView(int64_t view_id);
+
+  bool RemoveView(int64_t view_id);
```
loic-sharma, line 168: Please document new members

shell/common/engine.h
```diff
@@ -675,6 +675,10 @@ class Engine final : public RuntimeDelegate, PointerDataDispatcher::Delegate {
   ///
   std::optional<uint32_t> GetUIIsolateReturnCode();
 
+  void AddView(int64_t view_id);
+
+  void RemoveView(int64_t view_id);
```
loic-sharma, line 680: Please document new members

shell/common/fixtures/shell_test.dart
```diff
@@ -482,3 +482,42 @@ Future<void> testThatAssetLoadingHappensOnWorkerThread() async {
...
   notifyNative();
 }
+
+@pragma('vm:external-name', 'NativeReportViewIdsCallback')
+external void nativeReportViewIdsCallback(bool hasImplicitView, List<int> viewIds);
+
+List<int> getCurrentViewIds() {
+  return PlatformDispatcher.instance.views
+      .map((FlutterView view) => view.viewId)
+      .toSet()
```
loic-sharma, line 492: Why the `.toSet()`? It would be unexpected if there's repeated view ID values.

shell/common/rasterizer.h
```diff
@@ -204,6 +204,14 @@ class Rasterizer final : public SnapshotDelegate,
 
   fml::TaskRunnerAffineWeakPtr<SnapshotDelegate> GetSnapshotDelegate() const;
 
+  //----------------------------------------------------------------------------
+  /// @brief      Add a view, implicit or not.
+  void AddView(int64_t view_id);
+
+  //----------------------------------------------------------------------------
+  /// @brief      Remove a view, implicit or not.
+  void RemoveSurface(int64_t view_id);
```
loic-sharma, line 213: Let's be consistent between the add/remove methods (`AddView`/`RemoveView` or `AddSurface`/`RemoveSurface`)

```suggestion
  void RemoveView(int64_t view_id);
```

shell/common/rasterizer.h
```diff
@@ -204,6 +204,14 @@ class Rasterizer final : public SnapshotDelegate,
 
   fml::TaskRunnerAffineWeakPtr<SnapshotDelegate> GetSnapshotDelegate() const;
 
+  //----------------------------------------------------------------------------
+  /// @brief      Add a view, implicit or not.
+  void AddView(int64_t view_id);
+
+  //----------------------------------------------------------------------------
+  /// @brief      Remove a view, implicit or not.
```
loic-sharma, line 212: It looks like this file's style is to end with a trailing `///` line:

```suggestion
  /// @brief      Add a view, implicit or not.
  ///
  void AddView(int64_t view_id);

  //----------------------------------------------------------------------------
  /// @brief      Remove a view, implicit or not.
  ///
```

shell/common/rasterizer.h
```diff
@@ -204,6 +204,14 @@ class Rasterizer final : public SnapshotDelegate,
 
   fml::TaskRunnerAffineWeakPtr<SnapshotDelegate> GetSnapshotDelegate() const;
 
+  //----------------------------------------------------------------------------
+  /// @brief      Add a view, implicit or not.
+  void AddView(int64_t view_id);
```
loic-sharma, line 209: It looks like this file's style is to end with a trailing `///` line:

```suggestion
  //----------------------------------------------------------------------------
  /// @brief      Add a view, implicit or not.
  ///
  void AddView(int64_t view_id);
```

Same comment for `RemoveSurface` below

shell/common/rasterizer.h
```diff
@@ -204,6 +204,14 @@ class Rasterizer final : public SnapshotDelegate,
 
   fml::TaskRunnerAffineWeakPtr<SnapshotDelegate> GetSnapshotDelegate() const;
 
+  //----------------------------------------------------------------------------
+  /// @brief      Add a view, implicit or not.
```
loic-sharma, line 208: It looks like this file's style is to end with a trailing `///` line:

```suggestion
  /// @brief      Add a view, implicit or not.
  ///
```

Same comment for `RemoveSurface` below

shell/common/shell.cc
```diff
@@ -2023,6 +2037,71 @@ bool Shell::OnServiceProtocolReloadAssetFonts(
   return true;
 }
 
+void Shell::AddView(int64_t view_id) {
+  TRACE_EVENT0("flutter", "Shell::AddView");
+  FML_DCHECK(is_set_up_);
+  FML_DCHECK(task_runners_.GetPlatformTaskRunner()->RunsTasksOnCurrentThread());
+  if (view_id == kFlutterImplicitViewId) {
+    FML_DLOG(WARNING)
```
loic-sharma, line 2045: This seems like a logical bug that shouldn't happen? If so, prefer `ERROR` so that it shows up in logs.

```suggestion
    FML_DLOG(ERROR)
```

shell/common/shell.cc
```diff
@@ -2023,6 +2037,71 @@ bool Shell::OnServiceProtocolReloadAssetFonts(
...
+      [&latch,                                  //
+       rasterizer = rasterizer_->GetWeakPtr(),  //
+       view_id                                  //
+  ]() {
+        if (rasterizer) {
+          rasterizer->RemoveSurface(view_id);
+        }
+        latch.Signal();
+      });
+  latch.Wait();
```
loic-sharma, line 2102: Do we need a second latch so that we can wait until the UI thread has removed its view?

Also, should the raster thread wait until the UI thread has removed its view? If not, could you add a quick comment explaining why this is safe?

shell/common/shell.cc
```diff
@@ -2023,6 +2037,71 @@ bool Shell::OnServiceProtocolReloadAssetFonts(
...
+        << kFlutterImplicitViewId
+        << ". This view should never be added. This call is no-op.";
+    return;
+  }
+
+  task_runners_.GetUITaskRunner()->PostTask([engine = engine_->GetWeakPtr(),  //
+                                             view_id                          //
+  ] {
+    if (engine) {
+      engine->AddView(view_id);
```
loic-sharma, line 2056: Why is the view added to the UI thread before the raster thread? Wouldn't that mean the UI thread might attempt to render to a view before the raster thread has added that view? Should we swap this order such that the view is added to the rasterizer before the engine?

lib/ui/platform_dispatcher.dart
dkwingsmt: Thanks for finding this. It used to be reused for `ViewConfiguration`.

lib/ui/platform_dispatcher.dart
dkwingsmt: I think this method contains sufficient parts that I don't want to repeat it. I've renamed it to `_insertNewView` and moved it to before `_addView`. Hopefully it's a bit better.

lib/ui/window.dart
dkwingsmt: I had the same question, but I have the unit test `ShellWithImplicitViewDisabledStartsWithoutImplicitView` that disables the implicit view, and it passes.

Originally I had the global `window` variable marked as `late`, but Dart marked the `late` unnecessary. I guess global object variables are late inherently.

lib/ui/window/platform_configuration.cc
dkwingsmt: Done. I've also searched through the repo to make sure there are no other cases.

lib/ui/window/platform_configuration.cc
dkwingsmt: That's right! I wonder why it didn't cause any errors.

dkwingsmt: Oh, because I pass in the reference to `library_` and the value that it contains is changed. I don't think the movement is necessary, and putting it at the constructor might be clearer.

dkwingsmt: Oh but if we put it here we don't need a reference any more...

lib/ui/window/platform_configuration.cc
dkwingsmt: Don't we check `dart_state` with the if clause? Do you mean if the dart state is invalid we should not silently return but log an error?

shell/common/fixtures/shell_test.dart
dkwingsmt: I've changed it to an assertion that there are no duplicate view ID.

lib/ui/window/platform_configuration.cc
loic-sharma: My concern is that `PlatformConfiguration`'s state is updated (by `AddWindowRecord`) even if `dart_state` is invalid. That could mean that `PlatformConfiguration`'s state no longer matches `dart_state`. I *think* this is ok, but perhaps it'd be better to do the check first anyways?

runtime/runtime_controller.cc
dkwingsmt: `PlatformData` now contains a full map from view ID to view metrics, and this function flushes all of them to Dart by adding a view with this view metrics.

lib/ui/hooks.dart
```diff
@@ -30,9 +84,98 @@ void _updateDisplays(
   PlatformDispatcher.instance._updateDisplays(displays);
 }
 
+List<DisplayFeature> _decodeDisplayFeatures({
```
dkwingsmt, line 87: Both `_decodeDisplayFeatures` and `_buildViewConfiguration` were moved from `PlatformDispatcher`.

lib/ui/window/platform_configuration.cc
dkwingsmt: I'm keeping my implementation since it's safe enough.

lib/ui/window/platform_configuration.cc
dkwingsmt: To reconfirm, the question is why we allow modifying `windows_` when `dart_state` is invalid. The platform configuration states and the platform dispatcher states should be kept in sync. An alternative solution is to `FML_DCHECK` the validity of `dart_state`.

I've thought about it too, but I don't think it's that concerning because:
* Existing methods have been behaving in the same way and running fine.
* If I understand correctly, `PlatformConfiguration` is supposed to exist only when the Dart state exists. The Dart state might become invalid when then app starts shutting down, and in that case we'd prefer robustness over data correctness (since the app's ending).

dkwingsmt:
Until recently, I had been initializing the implicit view in `Shell::OnPlatformViewCreated`, both for the rasterizer and for the engine. But I discovered that the platform view might be destroyed and re-created during an app's lifetime, which is useful when the app goes background on mobile. I've therefore adjusted how different classes handle views:

The `Rasterizer` no longer requires callers to register a view - they are created automatically on rasterization. The states are removed when the during `CollectView` (for a specific view, during view removal) and `TearDown` (for all views, during platform view destruction). This is ok because the rasterizer's view states are "last-frame" records, which are only useful for performance during `DrawLastLayerTree` and don't affect functionality.

The `Engine` (including `PlatformConfiguration` and the Dart VM) still requires callers to properly register views and deregister views. These states are not affected during platform view creation and destruction. The implicit view is added during `Shell::Setup`, which is earlier than `Shell::OnPlatformViewCreated`. This is ok because these states do not rely on `Surface` and don't need `PlatformView` to be ready anyway.

lib/ui/window.dart
dkwingsmt: We discussed last Thursday and this is our thought process and conclusion (to my understanding).
* The engine will be allowed to disable implicit views at the end of phase 2, which is earlier than the official removal of `window`.
* Using `window` in whatever ways should not trigger errors when the implicit view is disabled.
* Instead of mapping `window` to an arbitrary existing view, which should provide expected behaviors: the `window` corresponds to the implicit view, such as `FlutterEngine.viewController`.
* Therefore, using `window` when the implicit view is disabled/closed should be no-ops.

shell/common/shell.cc
dkwingsmt: I've revised the blocking logic in `Shell::{Add,Remove}View` with some docs as well. Let me know what you think!

shell/common/shell.cc
dkwingsmt: Duplicate issue as https://github.com/flutter/engine/pull/42991#discussion_r1280159459

dkwingsmt:
I've addressed almost all issues except for https://github.com/flutter/engine/pull/42991#discussion_r1279981169 (how to make `window` compatible), which I've left a TODO with issue https://github.com/flutter/flutter/issues/131651.

This PR should be ready for another review. @loic-sharma can you take a look?

common/settings.h
```diff
@@ -224,6 +229,36 @@ struct Settings {
...
+  // kFlutterImplicitViewId. This view, called the implicit view, is available
+  // throughout the lifetime of the shell. Shell::AddView or RemoveView must not
+  // be called for this view. Even when the window that shows the view is
+  // closed, the framework is unaware and might continue render into or operate
+  // this view.
+  //
+  // The single-view APIs, which are APIs that do not specify view IDs, operate
+  // the implicit view. The multi-view APIs can operate all views, including the
+  // implicit view with the correct ID (kFlutterImplicitViewId), unless
+  // specified otherwise,
```
loic-sharma, line 257:
```suggestion
  // specified otherwise.
```

shell/common/rasterizer.h
```diff
@@ -204,6 +204,19 @@ class Rasterizer final : public SnapshotDelegate,
 
   fml::TaskRunnerAffineWeakPtr<SnapshotDelegate> GetSnapshotDelegate() const;
 
+  //----------------------------------------------------------------------------
+  /// @brief      Deallocate the resources for displaying a view.
+  ///
+  ///             This method should be called a view is removed.
```
loic-sharma, line 210:
```suggestion
  ///             This method should be called when a view is removed.
```

shell/common/rasterizer.cc
```diff
@@ -156,6 +156,12 @@ void Rasterizer::NotifyLowMemoryWarning() const {
   context->performDeferredCleanup(std::chrono::milliseconds(0));
 }
 
+void Rasterizer::CollectView(int64_t view_id) {
+  if (view_id == kFlutterImplicitViewId) {
```
loic-sharma, line 160: Since this logic will need to be updated for multi-view, please add a TODO comment and assert that `view_id` is `kFlutterImplicitViewId`

shell/common/shell.cc
```diff
@@ -986,8 +990,11 @@ void Shell::OnPlatformViewSetViewportMetrics(int64_t view_id,
       });
 
   {
+    // TODO(dkwingsmt): Use the correct view ID when this method supports
+    // multi-view.
+    int64_t view_id = kFlutterImplicitViewId;
```
loic-sharma, line 995: This method already accepts a `view_id` argument. Should we remove this local variable?

shell/common/shell.cc
```diff
@@ -1967,7 +1975,9 @@ bool Shell::OnServiceProtocolRenderFrameWithRasterStats(
 
     response->AddMember("snapshots", snapshots, allocator);
 
-    const auto& frame_size = expected_frame_size_;
+    // TODO(dkwingsmt): Not sure what these fields are supposed to be after
+    // multi-view. For now it sends the implicit view's dimension as before.
```
loic-sharma, line 1979: Should we create a tracking issue for this? We will likely forget to fix this otherwise until something breaks

lib/ui/hooks.dart
loic-sharma: Is this a strictly necessary change? If this is a refactoring, I'd recommend creating a separate follow-up pull request as this pull request is a bit large already

lib/ui/window/platform_configuration.cc
```diff
@@ -77,12 +76,31 @@ void PlatformConfiguration::DidCreateIsolate() {
...
+                                    const ViewportMetrics& view_metrics) {
+  auto [window_iterator, insertion_happened] = windows_.emplace(
+      view_id, std::make_unique<Window>(library_, view_id, view_metrics));
+  FML_DCHECK(insertion_happened);
+  // Make the new window send an AddView message to Dart.
+  window_iterator->second->AddView();
+}
+
+void PlatformConfiguration::RemoveView(int64_t view_id) {
+  FML_DCHECK(view_id != kFlutterImplicitViewId);
```
loic-sharma, line 92: Should we assert that the `view_id` is in `windows_`?

lib/ui/window/platform_configuration.h
```diff
@@ -268,6 +258,27 @@ class PlatformConfiguration final {
...
+  ///
+  /// @param[in]  view_id           The ID of the new view.
+  /// @param[in]  viewport_metrics  The initial viewport metrics for the view.
+  ///
+  void AddView(int64_t view_id, const ViewportMetrics& view_metrics);
+
+  //----------------------------------------------------------------------------
+  /// @brief      Notify the framework that a view is no longer available.
+  ///
+  ///             Removing a view that has not been added triggers assertion.
```
loic-sharma, line 276: Should we mention that the implicit view cannot be removed?

lib/ui/window/platform_configuration.cc
```diff
@@ -77,12 +76,31 @@ void PlatformConfiguration::DidCreateIsolate() {
...
+  library_.Set(tonic::DartState::Current(), library);
+}
+
+void PlatformConfiguration::AddView(int64_t view_id,
+                                    const ViewportMetrics& view_metrics) {
+  auto [window_iterator, insertion_happened] = windows_.emplace(
+      view_id, std::make_unique<Window>(library_, view_id, view_metrics));
+  FML_DCHECK(insertion_happened);
+  // Make the new window send an AddView message to Dart.
+  window_iterator->second->AddView();
```
loic-sharma, line 88: Currently this calls `PlatformDispatcher._addView` for all views, including the implicit view. The `PlatformDispatcher._addView` handles the implicit view specially by sending a window metric update as the implicit view was already created.

This seems to have two drawbacks:
1. It would be unexpected if `PlatformDispatcher._addView` was called twice for the implicit view. Currently, it is difficult to assert this
2. This sends an additional window metric update at startup. This might cause the framework to do unnecessary rebuilds.

An alternative approach might be to make `PlatformConfiguration::AddView` not call `PlatformDispatcher._addView` for the implicit view by not calling `Window::AddView`. We could then assert that `PlatformDispatcher._addView` is never called on the implicit view. What are your thoughts?

shell/common/shell.cc
```diff
@@ -2023,6 +2033,66 @@ bool Shell::OnServiceProtocolReloadAssetFonts(
...
 }
 
+void Shell::AddView(int64_t view_id, const ViewportMetrics& viewport_metrics) {
+  TRACE_EVENT0("flutter", "Shell::AddView");
+  FML_DCHECK(is_set_up_);
+  FML_DCHECK(task_runners_.GetPlatformTaskRunner()->RunsTasksOnCurrentThread());
+  if (view_id == kFlutterImplicitViewId) {
+    FML_DLOG(ERROR) << "Unexpected request to add the implicit view #"
+                    << kFlutterImplicitViewId
+                    << ". This view should never be added. This call is no-op.";
```
loic-sharma, line 2043: Should this be an assertion that exits? FYI, `FML_DCHECK` can log a message. For example, see:

https://github.com/flutter/engine/blob/9c4db1788f915b85d9f92fa5860a58b40884ae65/fml/shared_thread_merger.cc#L79

shell/common/shell.cc
```diff
@@ -2023,6 +2033,66 @@ bool Shell::OnServiceProtocolReloadAssetFonts(
...
+        // that adds back the view record.
+        task_runners.GetRasterTaskRunner()->PostTask([rasterizer, view_id]() {
+          if (rasterizer) {
+            rasterizer->CollectView(view_id);
+          }
+        });
+        latch.Signal();
+      });
+
+  latch.Wait();
```
loic-sharma, line 2093: Instead of blocking the platform thread, should this method accept a callback that is executed once the rasterizer has been updated?

loic-sharma:
@dkwingsmt Apologies for the delay, I missed your comment. I left some comments, but your new approach is looking good to me! :)

shell/common/shell.cc
dkwingsmt: I was planning to make AddView and RemoveView more tolerant to implicit views by returning with a log. But making them assert works too.

lib/ui/hooks.dart
dkwingsmt: Yes, this is necessary, because this part of logic is now shared between `_updateWindowMetrics` and `_addView` to parse the window metrics packet. I'd be happy to submit a preliminary PR for this.

shell/common/shell.cc
dkwingsmt: Done. Issue at https://github.com/flutter/flutter/issues/131892.

lib/ui/window/platform_configuration.cc
dkwingsmt: > This sends an additional window metric update at startup

At startup, this function is called by `FlushRuntimeStateToIsolate` when the isolate is ready/ The accumulated states with one call for each view, either `AddView` or `UpdateWindowMetrics`.

lib/ui/hooks.dart
loic-sharma: Ah gotcha makes sense. Yup a preliminary PR would be great 👍

lib/ui/platform_dispatcher.dart
```diff
@@ -218,10 +210,29 @@ class PlatformDispatcher {
...
+      } else {
+        debugLastImplicitView = result;
+        debugHasRecordedLastImplicitView = true;
+      }
+      return true;
+    }());
+    return result;
+  }
+  FlutterView? debugLastImplicitView;
+  bool debugHasRecordedLastImplicitView = false;
```
loic-sharma, line 235: Should these be private?

lib/ui/window/platform_configuration.cc
```diff
@@ -77,12 +76,34 @@ void PlatformConfiguration::DidCreateIsolate() {
...
+}
+
+void PlatformConfiguration::RemoveView(int64_t view_id) {
+  FML_DCHECK(view_id != kFlutterImplicitViewId)
+      << "The implicit view #" << view_id << " should never be removed.";
+  size_t erased_elements = windows_.erase(view_id);
+  FML_DCHECK(erased_elements != 0) << "View #" << view_id << " doesn't exist.";
+  (void)erased_elements;
```
loic-sharma, line 96: What's the `(void)` cast for?

lib/ui/window/platform_configuration.cc
loic-sharma: Marking this as resolved now that `PlatformDispatcher`'s constructor does not create the implicit view

lib/ui/window/platform_configuration.h
```diff
@@ -268,6 +258,30 @@ class PlatformConfiguration final {
   ///
   void DidCreateIsolate();
 
+  //----------------------------------------------------------------------------
+  /// @brief      Notify the framework that a new view is available.
+  ///
+  ///             A view must be added before other methods can refer to it,
+  ///             including the implicit view. Adding a view that has been
+  ///             added triggers assertion.
```
loic-sharma, line 266:
```suggestion
  ///             exists already triggers an assertion.
```

lib/ui/window/platform_configuration.h
```diff
@@ -268,6 +258,30 @@ class PlatformConfiguration final {
...
+  ///
+  void AddView(int64_t view_id, const ViewportMetrics& view_metrics);
+
+  //----------------------------------------------------------------------------
+  /// @brief      Notify the framework that a view is no longer available.
+  ///
+  ///             Removing a view that has not been added triggers assertion.
+  ///
+  ///             The implicit view (kFlutterImplicitViewId) should never be
+  ///             removed. Doing so triggers assertion.
```
loic-sharma, line 279:
```suggestion
  ///             removed. Doing so triggers an assertion.
```

lib/ui/window/window.cc
```diff
@@ -13,14 +13,48 @@
 
 namespace flutter {
 
-Window::Window(int64_t window_id, ViewportMetrics metrics)
-    : window_id_(window_id), viewport_metrics_(std::move(metrics)) {
-  library_.Set(tonic::DartState::Current(),
-               Dart_LookupLibrary(tonic::ToDart("dart:ui")));
-}
+Window::Window(tonic::DartPersistentValue& library,
```
loic-sharma, line 16: Is there any chance we can rename this to `View`? If yes, should this be part of this change?

lib/ui/window/window.h
```diff
@@ -20,18 +20,24 @@
...
+         int64_t window_id,
+         ViewportMetrics metrics);
 
   ~Window();
 
   int window_id() const { return window_id_; }
 
   const ViewportMetrics& viewport_metrics() const { return viewport_metrics_; }
 
+  // Send an AddView message to Dart with the current view metrics.
```
loic-sharma, line 33: Consider mentioning that it is an error to add a view with the same `id` multiple times.

```suggestion
  // Send an AddView message to Dart with the current view metrics.
  // This must not be called more than once per `window_id`.
```

lib/ui/window/platform_configuration.cc
dkwingsmt: This is to avoid a lint error that says `erased_elements` is unused in prod. It is also used in places such as `flow/frame_timings.cc`.

lib/ui/window/window.cc
dkwingsmt: We should definitely rename it to `View`. I don't want to do this since this PR is already pretty big. But mostly, we're probably removing `Window`. The only place that uses it is `DrawShadow`. (`RuntimeController` also uses it now but can be migrated.)

runtime/runtime_controller.cc
```diff
@@ -115,11 +113,13 @@ std::unique_ptr<RuntimeController> RuntimeController::Clone() const {
...
 
 bool RuntimeController::FlushRuntimeStateToIsolate() {
-  // TODO(dkwingsmt): Needs a view ID here (or platform_data should probably
-  // have multiple view metrics).
-  return SetViewportMetrics(kFlutterImplicitViewId,
-                            platform_data_.viewport_metrics) &&
-         SetLocales(platform_data_.locale_data) &&
+  for (auto const& [view_id, viewport_metrics] :
+       platform_data_.viewport_metrics_for_views) {
+    if (!AddView(view_id, viewport_metrics)) {
```
loic-sharma, line 118: Shouldn't this set the viewport metrics instead of adding the view? Wouldn't this cause errors if `RuntimeController::FlushRuntimeStateToIsolate` is called multiple times?

dkwingsmt: Currently I don't see why `RuntimeController::FlushRuntimeStateToIsolate` will be called multiple times because it is only called when the root isolate becomes ready. I can add a boolean for it.

shell/common/shell.cc
```diff
@@ -2023,6 +2031,61 @@ bool Shell::OnServiceProtocolReloadAssetFonts(
...
 }
 
+void Shell::AddView(int64_t view_id, const ViewportMetrics& viewport_metrics) {
+  TRACE_EVENT0("flutter", "Shell::AddView");
+  FML_DCHECK(is_set_up_);
+  FML_DCHECK(task_runners_.GetPlatformTaskRunner()->RunsTasksOnCurrentThread());
+  FML_DCHECK(view_id != kFlutterImplicitViewId)
+      << "Unexpected request to add the implicit view #"
+      << kFlutterImplicitViewId
+      << ". This view should never be added. This call is no-op.";
```
loic-sharma, line 2041: The assertion failure should exit the program. Consider:

```suggestion
      << ". This view should never be added.";
```

runtime/runtime_controller.cc
dkwingsmt: Also this needs to add the view because for `PlatformDispatcher` no views have been added at all; it has just been constructed.

shell/common/shell.cc
loic-sharma: Same comment for `RemoveView` assertion

shell/common/fixtures/shell_test.dart
```diff
@@ -482,3 +482,69 @@ Future<void> testThatAssetLoadingHappensOnWorkerThread() async {
...
+  }
+  return result;
+}
+
+@pragma('vm:external-name', 'NativeReportViewWidthsCallback')
+external void nativeReportViewWidthsCallback(List<int> viewWidthPacket);
+
+// This entrypoint reports whether there's an implicit view and the list of view
+// IDs using nativeReportViewIdsCallback on initialization and every
+// onMetricsChanged.
```
loic-sharma, line 543: This comment seems incorrect

lib/ui/window/platform_configuration.cc
loic-sharma: Ah gotcha. I'd consider using `FML_DCHECK(windows_.find(view_id) != windows_.end())` instead, but I don't feel strongly. Feel free to keep as is

lib/ui/window/window.cc
loic-sharma: Sounds good to me, marking as resolved

runtime/runtime_controller.cc
loic-sharma: Ahh interesting, I didn't realize `AddView` was ignored before the isolate is ready. Could we assert that `RuntimeController::FlushRuntimeStateToIsolate` is called no more than once?

runtime/runtime_controller.h
```diff
@@ -621,9 +645,6 @@ class RuntimeController : public PlatformConfigurationClient {
 
   bool FlushRuntimeStateToIsolate();
```
loic-sharma, line 646: Can we add a comment that this should be called only once when the Dart isolate is initialized?

shell/common/shell.cc
```diff
@@ -713,6 +713,9 @@ bool Shell::Setup(std::unique_ptr<PlatformView> platform_view,
   weak_rasterizer_ = rasterizer_->GetWeakPtr();
   weak_platform_view_ = platform_view_->GetWeakPtr();
 
+  if (settings_.enable_implicit_view) {
+    engine_->AddView(kFlutterImplicitViewId, ViewportMetrics{});
```
loic-sharma, line 717: From my understanding, at this point the Dart isolate isn't ready yet to receive the view? If yes, could you add a quick comment that this view won't actually be added until `RuntimeController::FlushRuntimeStateToIsolate`.

shell/common/fixtures/shell_test.dart
dkwingsmt: Thanks! Changed to 
```dart
// This entrypoint reports the list of views and their widths using
// nativeReportViewWidthsCallback on initialization and every onMetricsChanged.
```

shell/common/shell.cc
dkwingsmt: The Dart isolate isn't ready to receive this view for sure, as well as all other views that are added immediately after `FlutterEngine` is constructed. These view operations are stored in `RuntimeController` and not sent to Dart until `FlushRuntimeStateToIsolate`. I can note this in `FlushRuntimeStateToIsolate`.

shell/common/shell.cc
dkwingsmt: (Ah there's another comment I failed to address)
Yeah we might not need to wait for the UI task to finish. Let me remove the block first, and we'll see if we actually want a callback (we'll need a decision when we add the embedder API.)

dkwingsmt: Actually I'm not sure if we need to block until this task to finish at all. Theoretically, if an embedder requests to remove a view, even if the view ID is referred in the future, the embedder can just silently ignore them. I think I'll remove the block for now, and we'll see if we need a callback in the future (we'll need to make a decision when we add the embedder API.)

dkwingsmt: As a result I added the following doc:
```dart
  ///         This method returns immediately and does not wait for the task on
  ///         the UI thread to finish. This means that the Dart VM might still
  ///         send messages regarding this view ID for a short while, even
  ///         though this view ID is already invalid.
```

runtime/runtime_controller.h
dkwingsmt: I've added a boolean to guarantee that this method is only called once.

dkwingsmt:
For reviewers' convenience, I've written a short doc that explains my roadmap for landing engine changes, which starts with this PR.
https://docs.google.com/document/d/10APhzRDR7XqjWdbYWpFfKur7DPiz_HvSKNcLvcyA9vg/edit?resourcekey=0-DfGcg4-XWRMMZF__C1nmcA

dnfield (changes requested):
Putting a request changes on this while we discuss.

At a high level, I think a better API would be for embedders to say whether or not they are providing an implicit view and what ID it's using, rather than asking the engine to manage an implicit view or not for the embedder. We should also avoid adding an API right now that would always be an error to use.

goderbauer:
> At a high level, I think a better API would be for embedders to say whether or not they are providing an implicit view and what ID it's using, rather than asking the engine to manage an implicit view or not for the embedder.

This needs to be backwards compatible, though: Existing embedders assume that there is an implicit view and this needs to stay true without requiring them to update anything in their code. New embedders can make the choice that the engine should not have this implicit view it creates for legacy embedders.

dnfield:
> This needs to be backwards compatible, though: Existing embedders assume that there is an implicit view and this needs to stay true without requiring them to update anything in their code. New embedders can make the choice that the engine should not have this implicit view it creates for legacy embedders.

I'd rather defer the API surface for that until we have an implementation - right now the settings parameter as written in this patch would just make rendering fail if set to false, and I'm not entirely convinced that it's the most ergonomic spot for dealing with things without seeing more of a fleshed out use case.

lib/ui/window/platform_configuration.cc
```diff
@@ -77,12 +76,34 @@ void PlatformConfiguration::DidCreateIsolate() {
...
+  FML_DCHECK(insertion_happened);
+  // Make the new window send an AddView message to Dart.
+  window_iterator->second->AddView();
+}
+
+void PlatformConfiguration::RemoveView(int64_t view_id) {
+  FML_DCHECK(view_id != kFlutterImplicitViewId)
+      << "The implicit view #" << view_id << " should never be removed.";
+  size_t erased_elements = windows_.erase(view_id);
+  FML_DCHECK(erased_elements != 0) << "View #" << view_id << " doesn't exist.";
```
dnfield, line 95: Should these be DCHECK or CHECKs?

DCHECK will only run in a local unopt engine build. Flutter developers will never see these.

It seems like this should always be an error log at the least. Alternatively, it could be a DCHECK and an error log with a fizzle, e.g.

```cpp
if (view_id == kFlutterImplicitViewId) {
  FML_LOG(ERROR) << "The implicit view #" << view_id << " cannot be removed.";
  FML_DCHECK(false);
  return;
}
```

dnfield: Similar comment for the other DCHECK on 95

lib/ui/window/window.cc
```diff
@@ -13,14 +13,48 @@
...
 Window::~Window() {}
 
+void Window::AddView() {
+  std::shared_ptr<tonic::DartState> dart_state = library_.dart_state().lock();
+  if (!dart_state) {
+    return;
+  }
+  tonic::DartState::Scope scope(dart_state);
+  tonic::CheckAndHandleError(tonic::DartInvokeField(
+      library_.value(), "_addView",
```
dnfield, line 32: Please cache the field lookup somewhere, see how that's done in `PlatformConfiguration::DidCreateIsolate`.

lib/ui/window/window.cc
```diff
@@ -33,26 +67,26 @@ void Window::UpdateWindowMetrics(const ViewportMetrics& metrics) {
...
-          tonic::ToDart(metrics.physical_system_gesture_inset_top),
-          tonic::ToDart(metrics.physical_system_gesture_inset_right),
-          tonic::ToDart(metrics.physical_system_gesture_inset_bottom),
-          tonic::ToDart(metrics.physical_system_gesture_inset_left),
-          tonic::ToDart(metrics.physical_touch_slop),
-          tonic::ToDart(metrics.physical_display_features_bounds),
-          tonic::ToDart(metrics.physical_display_features_type),
-          tonic::ToDart(metrics.physical_display_features_state),
-          tonic::ToDart(metrics.display_id),
+          tonic::ToDart(viewport_metrics_.device_pixel_ratio),
```
dnfield, line 70: This should probably get cached as well (The lookup of `_updateWindowMetrics`).

dnfield (commented):
Couple more issues but after that I think should be good to go

lib/ui/window/platform_configuration.cc
dkwingsmt: I've changed one of them to the "if" pattern you suggested. I think both should stay DCHECK since they should never happen if `Shell` is implemented correctly, and serve as debug-only assertions. I changed the first check to an `if` to prevent the implicit view being deleted, but didn't for the second check so to make sure the message is always sent to the Dart VM

dkwingsmt:
I've also removed `Window`, but merging it to `PlatformConfiguration`. We've always wanted to do this, not only because `Window` has only one actual property (`view_configuration`) but also that the name is obsolete. I found any ways to cache the two methods would be more verbose than just throwing the class away right now.

common/settings.h
```diff
@@ -22,6 +22,32 @@
...
+// view ID of `kFlutterImplicitViewId` and is available throughout the lifetime
+// of the shell. `Shell::AddView` or `RemoveView` must not be called for this
+// view. Even when the window that shows the view is closed, the framework is
+// unaware and might continue rendering into or operating this view.
+//
+// The single-view APIs, which are APIs that do not specify view IDs, operate
+// the implicit view. The multi-view APIs can operate all views, including the
+// implicit view if the target ID is `kFlutterImplicitViewId`, unless specified
+// otherwise.
+constexpr int64_t kFlutterImplicitViewId = 0;
```
dnfield, line 49: nit: Consider moving this to window.h or platform_configuration.h.

dnfield (approved):

PR #43011, opened 2023-06-20
Skwasm platform views

eyebrowsoffire:
This implements platform views in Skwasm. There are a number of substantial changes rolled up in this change, including:
* Reworked the rendering system to use multiple canvases with a single WebGL context, via ImageBitmap rendering.
* Reworked our object management and bindings to use the `__externref_t` construct in C code.
* Upgraded emscripten to a much newer version
* Generified skwasm's scene builder to be able to work with any renderer that can produce `ImageBitmap` objects from Pictures, and whose `Canvas`, `Picture` and `ImageFilter` objects conform to `SceneCanvas`/`ScenePicture`/`SceneImageFilter`

For platform views themselves, most stuff is implemented except for clipping. I plan on doing that in a subsequent change.


lib/web_ui/lib/src/engine/layers.dart
```diff
@@ -0,0 +1,750 @@
...
+  @override
+  void post(SceneCanvas canvas, ui.Rect contentRect) {
+    if (clip == ui.Clip.antiAliasWithSaveLayer) {
+      canvas.restore();
+    }
+    canvas.restore();
+  }
+
+  @override
+  PlatformViewStyling createPlatformViewStyling() {
```
harryterkelsen, line 76: imo it would be cleaner to separate out the clip styling into a separate class. I don't think the layers should have this logic in them

lib/web_ui/lib/src/engine/scene_view.dart
```diff
@@ -0,0 +1,235 @@
...
+
+  Future<void> renderScene(EngineScene scene) async {
+    if (queuedRenders >= kMaxQueuedRenders) {
+      return;
+    }
+    queuedRenders += 1;
+
+    scene.beginRender();
+    final List<LayerSlice> slices = scene.rootLayer.slices;
+    final Iterable<Future<DomImageBitmap?>> renderFutures = slices.map(
```
harryterkelsen, line 44: nit: might be cleaner to use whereType to filter out the PlatformViewSlices rather than make a list of nullable futures

harryterkelsen (approved):
Awesome! LGTM with nits

lib/web_ui/lib/src/engine/scene_view.dart
eyebrowsoffire: The reason I did it this way is so that I can extract them by index when I'm iterating the slices themselves. If we use `whereType` there is no simple way to correlate the image bitmap with the slice it was rendered from.

PR #43014, opened 2023-06-20
[Impeller] Force initialize more possible color source variants on Vulkan, ensure stencil is always present to reduce variants further.

jonahwilliams:
Work towards https://github.com/flutter/flutter/issues/129050

Initializes all possible variants for solid color and gradient pipelines.
Adds check that forces stencils to always be attached for Android to avoid generating extra PSO variants.
Turns default sample count back to 1, but initializes all PSOs with a samplecount of 4


impeller/entity/contents/content_context.cc
```diff
@@ -352,11 +231,21 @@ std::shared_ptr<Texture> ContentContext::MakeSubpass(
   if (context->GetCapabilities()->SupportsOffscreenMSAA() && msaa_enabled) {
     subpass_target = RenderTarget::CreateOffscreenMSAA(
         *context, texture_size, SPrintF("%s Offscreen", label.c_str()),
-        RenderTarget::kDefaultColorAttachmentConfigMSAA, std::nullopt);
+        RenderTarget::kDefaultColorAttachmentConfigMSAA  //
+#ifndef FML_OS_ANDROID
```
jonahwilliams, line 235: Optimizing out the stencil forces a new variant on Vulkan. Perhaps there is a faster way to attach a stencil and not use it, but this seems reasonable?

impeller/entity/contents/pipeline_macros.h
```diff
@@ -0,0 +1,128 @@
...
+    PrimitiveType::kTriangle, PrimitiveType::kTriangleStrip};
+
+template <class T>
+using Variants = std::unordered_map<ContentContextOptions,
+                                    std::unique_ptr<T>,
+                                    ContentContextOptions::Hash,
+                                    ContentContextOptions::Equal>;
+
+template <typename PipelineT>
+static std::unique_ptr<PipelineT> CreatePipeline(
```
jonahwilliams, line 93: started life as a macro but became a template 🤷

jonahwilliams:
Based on some initial studies of the current PSO construction time, I don't think we can actually afford to initialize any more PSOs at startup. I'm going to close this and open a new PR for the stencil and sample count change.

PR #43026, opened 2023-06-20
[Impeller] Add Rect::Project

bdero:
- Utility to project a rectangle onto another rectangle.
- Fix constness for Rect::Expand.
- Add named methods for scaling.

PR #43028, opened 2023-06-21
[Impeller] Correctly compute UVs in texture fill

bdero:
Resolves https://github.com/flutter/flutter/issues/128353.

Depends on https://github.com/flutter/engine/pull/43026.

* Maps the texel coordinate space correctly relative to the geometry pixel coordinates.
* Use decal sampling for subpass textures.
* Simplify the computation by using new rect utilities such as `Rect::Project`.

Texture in subpass (the issue was that when copying to the parent pass, this image gets screwed up):
image

Before Impeller:
image

After Impeller:
image

### DIFF
--- a/impeller/entity/contents/texture_contents.cc
+++ b/impeller/entity/contents/texture_contents.cc
@@ -115,12 +115,14 @@ bool TextureContents::Render(const ContentContext& renderer,
     return true;  // Nothing to render.
   }
 
-  auto half_texel_size =
-      Size(0.5 / texture_->GetSize().width, 0.5 / texture_->GetSize().height);
   auto texture_coords =
       Rect::MakeSize(texture_->GetSize()).Project(source_rect_);
+
+  // Expand the texture coordinates
+  auto half_texel_size =
+      Size(0.5 / texture_->GetSize().width, 0.5 / texture_->GetSize().height);
   texture_coords =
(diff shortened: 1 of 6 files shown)


bdero:
The subpass image is getting copied to the parent pass perfectly, and the remaining visible imperfections are due to having only 4 subpixel weights with MSAA.

jonahwilliams (approved):
LGTM!

bdero:
I fixed an issue where I was not mapping the source rect in terms of the texture size, and ended up writing a solution that's much nicer/easier to understand.

impeller/entity/contents/texture_contents.cc
```diff
@@ -104,37 +104,31 @@ std::optional<Snapshot> TextureContents::RenderToSnapshot(
...
       texture_ == nullptr || texture_->GetSize().IsEmpty()) {
-    return true;
+    return true;  // Nothing to render.
   }
 
+  // Expand the source rect by half a texel, which aligns sampled texels to the
+  // pixel grid if both the source rect and destination rect are given in screen
+  // space.
+  auto texture_coords =
+      Rect::MakeSize(texture_->GetSize()).Project(source_rect_.Expand(0.5));
```
bdero, line 122: This line is all that's needed to correctly compute the texture coordinates. :)

PR #43040, opened 2023-06-21
[Impeller] Perform one queue submission for all render passes, post submitKHR to worker pool.

jonahwilliams:
Locally this improves performance quite a bit.

1. There is a substantial overhead for each queue submit. But we can still can get the same ordering guarantees by submitting at once.
2. The submit khr synchronization is already handled by the fences/semaphores, and since it blocks we can post to a worker.

Long term, we should be able to run encoding of render passes on worker threads, but we'll need to join and submit still. This architecture is sort of a half way point to there.


impeller/renderer/backend/vulkan/swapchain_impl_vk.cc
```diff
@@ -377,8 +378,33 @@ bool SwapchainImplVK::Present(const std::shared_ptr<SwapchainImageVK>& image,
...
+    for (const auto& encoder : encoders) {
+      buffers.push_back(encoder->GetCommandBuffer());
+    }
+    submit_info.setCommandBuffers(buffers);
+    if (context.GetGraphicsQueue()->Submit(submit_info, *fence) !=
+        vk::Result::eSuccess) {
+      return false;
+    }
+
+    if (!context.GetFenceWaiter()->AddFence(
```
jonahwilliams, line 403: I think there must be a way to also join this fence/submit with the one below, but I didnt figure it out.

jonahwilliams:
More or less inspired by some suggestions in https://zeux.io/2020/02/27/writing-an-efficient-vulkan-renderer/ aswell as the discussion @bdero and I had yesterday about concurrent encode. I wanted to find an intermediate step where encoding works more like macOS but without needing to immediately implement the joined submit, so I took a bit of a half measure that should be easy to eventually update into fully concurrent encode.

impeller/renderer/command_buffer.cc
```diff
@@ -51,6 +51,25 @@ bool CommandBuffer::SubmitCommandsAsync(
   return SubmitCommands(nullptr);
 }
 
+bool CommandBuffer::SubmitCommandsAsync(
```
bdero, line 54: I haven't combed over this completely yet, but it seems a bit odd that we need to add a new submit method for this. Is there ever a case where we shouldn't emulate Metal's good behavior on the Vulkan backend?

jonahwilliams: We're using SubmitCommandsAsync on macOS too. Fundamentally the old design was always 1. Encoder 2. Submit, but pushing that entire thing to a worker thread made the ownership a bit more complicated, having the command buffer take ownership of the render pass or blit pass makes things a bit clearer.

bdero: Why do we have both `SubmitCommands` and `SubmitCommandsAsync`? From reading the docstrings, I can't distinguish why one should ever need to use `SubmitCommands`.

When is using `SubmitCommandsAsync` unsafe or undesirable?

jonahwilliams: Oh at this point there is no reason to use SubmitCommands. Though in theory both should work, I can pare this down and just make SubmitCommandsAsync the only option (and rename it to SubmitCommands)

jonahwilliams: actually there are two reasons we use this today:

1. Testing, sometimes we latch on the callback for the compute unittests.
2. The blit pass workloads that happen outside of the frame. Technically the way I've implemented submit commands async on vulkan assumes that there is some future step that will submit N command buffers, but there is no equivalent end of frame workload on image uploads.

So I don't know that we could entirely remove it, but we could re-arrange things to make it more obviously not the right thing to use for frame workload operations.

impeller/renderer/backend/vulkan/command_buffer_vk.cc
```diff
@@ -47,6 +48,48 @@ const std::shared_ptr<CommandEncoderVK>& CommandBufferVK::GetEncoder() const {
...
+    return false;
+  }
+  context_vk.GetCommandBufferQueue()->Enqueue(std::move(encoder_));
+  return true;
+}
+
+bool CommandBufferVK::SubmitCommandsAsync(
+    std::shared_ptr<RenderPass> render_pass) {
+  TRACE_EVENT0("impeller", "CommandBufferVK::SubmitCommandsAsync");
+  if (!render_pass->IsValid() || !IsValid()) {
```
gaaclarke, line 75: nit: these can be the same function instead of duplicating the logic

```c++
namespace {
  template<typename Pass, typename Context, typename Encoder, typename EncodeFunc>
  {
    bool SubmitCommandsAsync(Pass pass, Context, Encoder encoder, Encoder encode_func) {
    // ...
   }
}
```

jonahwilliams:
I've confirmed that I am able to get more than one swapchain image at a time with this change:

```
E/flutter ( 6628): [ERROR:flutter/impeller/renderer/backend/vulkan/swapchain_impl_vk.cc(481)] Done Present: 1
E/flutter ( 6628): [ERROR:flutter/impeller/renderer/backend/vulkan/swapchain_impl_vk.cc(341)] GOT IMAGE: 2
E/flutter ( 6628): [ERROR:flutter/impeller/renderer/backend/vulkan/swapchain_impl_vk.cc(459)] Present: 2
E/flutter ( 6628): [ERROR:flutter/impeller/renderer/backend/vulkan/swapchain_impl_vk.cc(341)] GOT IMAGE: 3
E/flutter ( 6628): [ERROR:flutter/impeller/renderer/backend/vulkan/swapchain_impl_vk.cc(481)] Done Present: 2
```

However this generally only happens if we miss the frame budget, since if we render fast enough it finishes quickly

impeller/renderer/backend/vulkan/context_vk.cc
```diff
@@ -437,6 +439,15 @@ std::shared_ptr<PipelineLibrary> ContextVK::GetPipelineLibrary() const {
   return pipeline_library_;
 }
 
+std::shared_ptr<CommandBufferQueue> ContextVK::GetCommandBufferQueue() const {
```
gaaclarke, line 442: Do we want a command buffer queue to be associated with a frame?  If we have concurrent frames being encoded couldn't this get jumbled up?

The way this is written couldn't one frame be written, then another one gets started written to and he's using the command buffer queue which gets flushed from the first frame?

impeller/renderer/backend/vulkan/context_vk.h
```diff
@@ -31,6 +31,20 @@ class CommandEncoderVK;
 class DebugReportVK;
 class FenceWaiterVK;
 
+class CommandBufferQueue {
+ public:
+  void Enqueue(std::shared_ptr<CommandEncoderVK> encoder) {
```
gaaclarke, line 36: tidy should complain about this, `const std::shared_ptr<CommandEncoderVK>&`

impeller/renderer/backend/vulkan/context_vk.h
```diff
@@ -160,6 +179,8 @@ class ContextVK final : public Context,
   std::shared_ptr<FenceWaiterVK> fence_waiter_;
   std::string device_name_;
   std::shared_ptr<fml::ConcurrentTaskRunner> worker_task_runner_;
+  std::shared_ptr<fml::ConcurrentMessageLoop> submission_task_runner_;
```
gaaclarke, line 182: Can't we just use the worker task runner?

impeller/renderer/backend/vulkan/swapchain_impl_vk.cc
```diff
@@ -429,32 +454,36 @@ bool SwapchainImplVK::Present(const std::shared_ptr<SwapchainImageVK>& image,
...
-      [[fallthrough]];
-    case vk::Result::eSuccess:
-      return true;
-    default:
-      VALIDATION_LOG << "Could not present queue: " << vk::to_string(result);
-      return false;
-  }
-  FML_UNREACHABLE();
-  return false;
+  context.GetSubmissionTaskRunner()->PostTask([&, current_frame, index]() {
```
gaaclarke, line 457: From what I'm understanding, this is the most important part of your change.  Do we want to split this up with the change that is enqueuing the command buffers?

gaaclarke (commented):
Sounds super promising.  Here are my concerns in summary:
1) Should we split this up between presenting on a worker thread and batching the submits?
1) Did you run the gallery test locally to measure the performance difference?
1) Having one command buffer queue per context could jumble up submission between frames, couldn't that lead to problems?

impeller/renderer/backend/vulkan/swapchain_impl_vk.cc
```diff
@@ -377,8 +378,32 @@ bool SwapchainImplVK::Present(const std::shared_ptr<SwapchainImageVK>& image,
...
 
-  const auto& sync = synchronizers_[current_frame_];
+  // Submit all command buffers.
+  {
+    auto [fence_result, fence] = context.GetDevice().createFenceUnique({});
+    if (fence_result != vk::Result::eSuccess) {
+      return false;
+    }
+
+    const auto& encoders = context.GetCommandBufferQueue()->Take();
```
gaaclarke, line 391: Related to my question above, i guess if this is guaranteed to be called before encoding another image, it should be fine for the context to have one command buffer queue.  I don't know if that's the case off the top of my head.

jonahwilliams:
Yes @gaaclarke I think you're absolutely right that for sanity's sake we should split this up into multiple changes. Still, presented here for coherency

impeller/renderer/backend/vulkan/context_vk.h
jonahwilliams: I thought I had read there might be some threading restriction on which threads we can call submit from, so it needed to be the same one or something. I can't find that now, let me try this out and see

impeller/renderer/backend/vulkan/context_vk.cc
jonahwilliams: Yeah that part of this is definitely a bit fiddly. I don't think its currently possible for us to concurrent encode multiple frames, in that case as long as we reset the queue in submit it should be perfectly safe...

impeller/renderer/backend/vulkan/context_vk.h
jonahwilliams: this seems to work with just using the concurrent worker

jonahwilliams:
Testing this on gallery on a pixel 6, I don't see much difference - but then most of the app has only one or two render passes anyway. I'm seeing about ~1ms on average for certain wonderous screens.

I'm not super thrilled about either measurement - I think there is something to the fence waiter causing stalls too.

I'm going to work on a reduce benchmark that guarantees we create a large number of render passes, so I can check if performance improves as I'd expect.

jonahwilliams:
So far we've identified the issue with Tracked Object deallocation blocking and we've re-introduced a higher priority queue for graphics workloads. This might become feasible again.

The current issue is that the command buffers are all created on the raster thread with the raster thread's command pool, but then destructed on a worker thread. We need to delay command buffer construction so that we can get the appropriate thread local command pool, as using a command pool from multiple threads is a big no-no (at least according to the validation layers).

PR #43053, opened 2023-06-21
[web] Don't get break type from v8BreakIterator

mdebbar:
In some languages, when the text contains a new line "\n", the `v8BreakIterator` starts returning different values from `breakType()`. This caused our code to think that those line breaks were hard line breaks when in fact they were soft line breaks. Still unclear if this is a `v8BreakIterator` bug or a wrong expectation on our side regarding how `breakType()` is supposed to work.

Instead of using `v8BreakIterator.breakType()`, let's do the `soft` vs `hard` detection ourselves (we already have all the necessary pieces).

Fixes https://github.com/flutter/flutter/issues/127379

### DIFF
--- a/lib/web_ui/lib/src/engine/text/line_breaker.dart
+++ b/lib/web_ui/lib/src/engine/text/line_breaker.dart
@@ -84,8 +84,6 @@ List<LineBreakFragment> breakLinesUsingV8BreakIterator(String text, JSString jsT
   iterator.adoptText(jsText);
   iterator.first();
   while (iterator.next() != -1) {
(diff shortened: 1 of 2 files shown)


mdebbar:
According to [this code](https://chromium.googlesource.com/v8/v8/+/b3014e0ef772415cf0827f330b3abc59b5d1ef28/src/objects/js-break-iterator.cc#221) in V8, it looks like `breakType()` is only meant for `words`.

yjbanov (approved):
lgtm

PR #43058, opened 2023-06-21
[web] - Fix for Safari partially autofilling form 

htoor3:
We have an issue where autofill in Safari Desktop is only partially autofilling the form.  The only input that gets filled is the currently focused one. 

This is happening for a couple of reasons:
- We hide non-focused inputs that are part of the autofill form by sizing them to 0px height and width, among other styles.
- This doesn't cause issues in most browsers.  However, Safari does not respect offscreen or 0-sized inputs, and this leads to the partial form autofill issue. 

The solution is to make sure we hide the inputs in the rest of the form from the users, without hiding it from Safari. 

Fixes https://github.com/flutter/flutter/issues/71275

## Pre-launch Checklist

- [x] I signed the [CLA].


lib/web_ui/test/engine/text_editing_test.dart
```diff
@@ -2290,6 +2290,63 @@ Future<void> testMain() async {
...
+          autofillForm.formElement.childNodes.toList()
+              as List<DomHTMLInputElement>;
+      final DomHTMLInputElement username = formChildNodes[0];
+      final DomHTMLInputElement password = formChildNodes[1];
+      expect(username.name, 'username');
+      expect(password.name, 'current-password');
+      expect(username.style.width, isNot(equals('0px')));
+      expect(username.style.height, isNot(equals('0px')));
+      expect(password.style.width, isNot(equals('0px')));
+      expect(password.style.height, isNot(equals('0px')));
```
mdebbar, line 2347: The `equals` is not necessary.
```suggestion
      expect(username.style.width, isNot('0px'));
      expect(username.style.height, isNot('0px'));
      expect(password.style.width, isNot('0px'));
      expect(password.style.height, isNot('0px'));
```

mdebbar, line 2347: Let's also check for `pointer-events: none`?

lib/web_ui/lib/src/engine/text_editing/text_editing.dart
```diff
@@ -126,6 +126,16 @@ void _hideAutofillElements(DomHTMLElement domElement,
...
   }
 
+  if (shouldHideElement) {
+    elementStyle
+      ..width = '0'
+      ..height = '0';
+  }
+
+  if (shouldDisablePointerEvents) {
+    elementStyle.pointerEvents = 'none';
```
mdebbar, line 136: This seems like a good thing to do in all browsers, what do you think?

lib/web_ui/lib/src/engine/text_editing/text_editing.dart
```diff
@@ -201,7 +212,10 @@ class EngineAutofillForm {
       e.preventDefault();
     }));
 
-    _hideAutofillElements(formElement);
+    // We need to explicitly disable pointer events on the form in Safari Desktop,
+    // so that we don't have pointer event collisions if users hover over or click
+    // into the invisible autofill elements within the form.
+    _hideAutofillElements(formElement, shouldDisablePointerEvents: isSafariDesktopStrategy);
```
mdebbar, line 218: It's not clear to me why we are not passing `shouldHideElement: !isSafariDesktopStrategy` here.

lib/web_ui/lib/src/engine/text_editing/text_editing.dart
htoor3: We probably could but I noticed one side effect if we do this, where the dev tools inspector can never directly inspect any of the inputs.  It's a minor annoyance, but since setting `pointer-events` to `none` doesn't seem to have an effect on browsers outside of Safari, I decided to keep the change exclusive to Safari.

lib/web_ui/lib/src/engine/text_editing/text_editing.dart
htoor3: For the autofill to work, we only need the actual `<input>` elements visible for Safari and don't have to worry about the `<form>`.

As for why we need to disable pointer events here on an "invisible" form, Safari has this weird behavior where the `<form>` element still registers as the target, even if the input child has disabled pointer events.  This doesn't occur on Chrome or Firefox from my testing.

mdebbar: Oh nvm. I didn't notice that this is for `<form>` while the other one below is for `<input>` elements.

mdebbar (approved):
LGTM

PR #43061, opened 2023-06-21
Revert "[Impeller] dont use concurrent runner to decode images on Android."

jonahwilliams:
Reverts flutter/engine#42944

This didn't improve any of the benchmarks, which I think at least disproves my theory on overloading. Lets go back to the prior strategy and look for improvements elsewhere.


jonahwilliams:
Friendly ping @dnfield @chinmaygarde

PR #43085, opened 2023-06-22
[Impeller] remove Vulkan pipeline cache mutex.

jonahwilliams:
At least from what I've found online, this API is already safe to call from multiple threads. From testing, at startup the presence of the mutex is adding ~100ms to shader creation time, as all compilations across N threads are fighting for the same lock.

https://github.com/flutter/flutter/issues/129050

### DIFF
--- a/impeller/renderer/backend/vulkan/pipeline_cache_vk.cc
+++ b/impeller/renderer/backend/vulkan/pipeline_cache_vk.cc
@@ -68,13 +68,6 @@ PipelineCacheVK::PipelineCacheVK(std::shared_ptr<const Capabilities> caps,
       OpenCacheFile(cache_directory_, kPipelineCacheFileName, vk_caps);
 
   vk::PipelineCacheCreateInfo cache_info;
-
-  // TODO(csg): VK_PIPELINE_CACHE_CREATE_EXTERNALLY_SYNCHRONIZED_BIT is behind
-  // an extension. Check it and set it. If not, the implementation is doing
-  // unnecessary synchronization.
-  // cache_info.flags =
-  // vk::PipelineCacheCreateFlagBits::eExternallySynchronized;
-
   if (existing_cache_data) {
     cache_info.initialDataSize = existing_cache_data->GetSize();
     cache_info.pInitialData = existing_cache_data->GetMapping();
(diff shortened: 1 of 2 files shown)


jonahwilliams:
If we end up creating a bunch of variants on the fly this would also explain why it sometimes takes longer.

impeller/renderer/backend/vulkan/pipeline_cache_vk.cc
```diff
@@ -68,13 +68,6 @@ PipelineCacheVK::PipelineCacheVK(std::shared_ptr<const Capabilities> caps,
       OpenCacheFile(cache_directory_, kPipelineCacheFileName, vk_caps);
 
   vk::PipelineCacheCreateInfo cache_info;
-
-  // TODO(csg): VK_PIPELINE_CACHE_CREATE_EXTERNALLY_SYNCHRONIZED_BIT is behind
-  // an extension. Check it and set it. If not, the implementation is doing
-  // unnecessary synchronization.
-  // cache_info.flags =
-  // vk::PipelineCacheCreateFlagBits::eExternallySynchronized;
```
dnfield, line 76: Based on what I'm reading in https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkPipelineCacheCreateFlagBits.html it sounds like the idea would be to use this extension so we could use the mutex ourselves.

But since we don't have the extension it ends up being both us and the implementation blocking.

So we might as well just let the implementation do it for us.

dnfield (approved):
LGTM.

Another quote 

> VUID-vkCreateGraphicsPipelines-pipelineCache-02876
If pipelineCache was created with VK_PIPELINE_CACHE_CREATE_EXTERNALLY_SYNCHRONIZED_BIT, host access to pipelineCache must be [externally synchronized](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#fundamentals-threadingbehavior)

dnfield:
Actually per https://registry.khronos.org/vulkan/specs/1.3-extensions/html/vkspec.html#pipelines-cache-destroy it looks like we still have to synchronize the destruction of the cache.

dnfield:
...but we haven't been doing that anyway.

jonahwilliams:
I think we can get away with that since most/all of our Vk objects should be destructed on the raster thread, but I don't know.

dnfield:
The pipeline cache can get destroyed either on the raster thread or on the worker thread that persists the cache to disk.

PR #43089, opened 2023-06-22
[Impeller] default sample count back to 1 (but configure to 4 in defaults).

jonahwilliams:
Also forces the creation of a stencil attachment always on Android to reduce the number of PSO variants.

https://github.com/flutter/flutter/issues/129050

### DIFF
--- a/impeller/entity/contents/content_context.cc
+++ b/impeller/entity/contents/content_context.cc
@@ -355,7 +355,7 @@ std::shared_ptr<Texture> ContentContext::MakeSubpass(
     subpass_target = RenderTarget::CreateOffscreenMSAA(
         *context, texture_size, SPrintF("%s Offscreen", label.c_str()),
         RenderTarget::kDefaultColorAttachmentConfigMSAA  //
-#ifndef FML_OS_ANDROID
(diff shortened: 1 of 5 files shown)


impeller/entity/contents/content_context.cc
```diff
@@ -352,11 +354,21 @@ std::shared_ptr<Texture> ContentContext::MakeSubpass(
   if (context->GetCapabilities()->SupportsOffscreenMSAA() && msaa_enabled) {
     subpass_target = RenderTarget::CreateOffscreenMSAA(
         *context, texture_size, SPrintF("%s Offscreen", label.c_str()),
-        RenderTarget::kDefaultColorAttachmentConfigMSAA, std::nullopt);
+        RenderTarget::kDefaultColorAttachmentConfigMSAA  //
+#ifndef FML_OS_ANDROID
```
bdero, line 358: Mind adding another comment nearby about how keeping the stencil attachment is just to keep PSO variants to a minimum on Android (similar to the one below)?

bdero (approved):
LGTM!

impeller/entity/contents/content_context.cc
jonahwilliams: Done

PR #43091, opened 2023-06-22
Workaround a release blocker after libc++ change

chingjun:
The code breaks in C++20 mode after libc++ removes comparisons for `std::vector` and replaces them with 'operator <=>'.

See cl/542541552 for context.

### DIFF
--- a/shell/platform/common/client_wrapper/include/flutter/encodable_value.h
+++ b/shell/platform/common/client_wrapper/include/flutter/encodable_value.h
@@ -215,6 +215,11 @@ class EncodableValue : public internal::EncodableValueVariant {
     }
     return std::get<int64_t>(*this);
   }
+
+  // Avoid operator<=> problems with std::variant in C++20 mode.
+  friend bool operator<(const EncodableValue& lhs, const EncodableValue& rhs) {
+    return static_cast<const super&>(lhs) < static_cast<const super&>(rhs);
+  }
 };
 
 }  // namespace flutter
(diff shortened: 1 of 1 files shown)


shell/platform/common/client_wrapper/include/flutter/encodable_value.h
```diff
@@ -215,6 +215,11 @@ class EncodableValue : public internal::EncodableValueVariant {
     }
     return std::get<int64_t>(*this);
   }
+
+  // Avoid operator<=> problems with std::variant in C++20 mode.
```
stuartmorgan, line 219: What are the problems? Even reading the internal change I'm not clear on what problem is being solved.

And is this permanent, or a temporary workaround? If the latter, what are the conditions for removing it?

shell/platform/common/client_wrapper/include/flutter/encodable_value.h
```diff
@@ -215,6 +215,11 @@ class EncodableValue : public internal::EncodableValueVariant {
...
+  friend bool operator<(const EncodableValue& lhs, const EncodableValue& rhs) {
+    return static_cast<const super&>(lhs) < static_cast<const super&>(rhs);
```
stuartmorgan, line 221: I'm not following this code. What does an explicit cast to a `super` reference do? This needs an explanatory comment.

stuartmorgan (changes requested):
What's the failure mode here? Ideally we should be testing this.

stuartmorgan:
I've added a suggested comment for the method internally; if everyone is fine with that text I think we're good.

stuartmorgan (approved):
LGTM

PR #43094, opened 2023-06-22
[Impeller] Add validation forbidding SamplerAddressMode::kDecal on the OpenGLES backend

bdero:
Fixes uncontrolled crash when kDecal is used (as seen on https://github.com/flutter/engine/pull/43087).

### DIFF
--- a/impeller/renderer/backend/gles/sampler_library_gles.cc
+++ b/impeller/renderer/backend/gles/sampler_library_gles.cc
@@ -19,6 +19,9 @@ SamplerLibraryGLES::~SamplerLibraryGLES() = default;
 // |SamplerLibrary|
 std::shared_ptr<const Sampler> SamplerLibraryGLES::GetSampler(
     SamplerDescriptor descriptor) {
+  // TODO(bdero): Change this validation once optional support for kDecal is
+  //              added to the OpenGLES backend:
+  //              https://github.com/flutter/flutter/issues/129358
   if (descriptor.width_address_mode == SamplerAddressMode::kDecal ||
       descriptor.height_address_mode == SamplerAddressMode::kDecal ||
       descriptor.depth_address_mode == SamplerAddressMode::kDecal) {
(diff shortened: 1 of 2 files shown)


jonahwilliams (commented):
What happens if we attempt to bind a null Sampler?

impeller/renderer/backend/gles/sampler_library_gles.cc
```diff
@@ -17,6 +19,14 @@ SamplerLibraryGLES::~SamplerLibraryGLES() = default;
 // |SamplerLibrary|
 std::shared_ptr<const Sampler> SamplerLibraryGLES::GetSampler(
     SamplerDescriptor descriptor) {
+  if (descriptor.width_address_mode == SamplerAddressMode::kDecal ||
+      descriptor.height_address_mode == SamplerAddressMode::kDecal ||
+      descriptor.depth_address_mode == SamplerAddressMode::kDecal) {
+    VALIDATION_LOG << "SamplerAddressMode::kDecal is not supported by the "
```
zanderso, line 25: A TODO linking to a GItHub issue would probably be helpful here.

zanderso (approved):

impeller/renderer/backend/gles/sampler_library_gles.cc
bdero: Done & filed this issue: https://github.com/flutter/flutter/issues/129358

bdero:
> What happens if we attempt to bind a null Sampler?

The `command.BindResource` call will early return and not add the sampler to the command, and then the GLES backend will fail to find the sampler slot. This triggers a chain of early returns that bubble up to an [FML_CHECK](https://github.com/flutter/engine/blob/1225ac5dd6e986b1fe1fecce91408d3eb6db2ed7/impeller/renderer/backend/gles/render_pass_gles.cc#L535) that will fail.

PR #43097, opened 2023-06-22
[Impeller] Fix and reland drawPaint collapsing optimization.

gaaclarke:
Collapses DrawPaint calls into clear colors when possible.

issue https://github.com/flutter/flutter/issues/129292
relands https://github.com/flutter/engine/pull/41711

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/impeller/aiks/aiks_unittests.cc
+++ b/impeller/aiks/aiks_unittests.cc
@@ -1992,6 +1992,18 @@ TEST_P(AiksTest, OpacityPeepHoleApplicationTest) {
   ASSERT_TRUE(delegate->CanCollapseIntoParentPass(entity_pass.get()));
 }
 
+TEST_P(AiksTest, DrawPaintAbsorbsClears) {
+  Canvas canvas;
+  canvas.DrawPaint({.color = Color::Red(), .blend_mode = BlendMode::kSource});
+  canvas.DrawPaint(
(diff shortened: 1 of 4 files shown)


chinmaygarde (approved):

gaaclarke:
I checked the golden image results:
1) CanRenderOffscreenCheckerboard - This doesn't reproduce locally, it looks like the save layer has a color that isn't transparent.  I think we need to conditionally set the clear color instead of always setting it.
2) CanRenderDestructiveSaveLayer - This is drawing wrong too, the correct drawing does not have a red background.  Well have to figure out why that isn't blending correctly.

gaaclarke:
For the `CanRenderOffscreenCheckerboard` problem it was because when we call `EntityPass::SetClearColor`, that clear color was used as the clear color for every subpass in an entity pass.  That test uses a multiply blend mode so the cream color would get multiplied, creating the error.  I've updated it so that when you call EntityPass::SetClearColor, the clear color only applies to the `offscreen_target`.  After typing this out though, this might fix Vulkan and break Metal since it isn't going down the same path.  We'll see.

gaaclarke:
CanRenderDestructiveSaveLayer should be fixed by the same fix.  It's difficult to know for sure, I think Skia Gold draws transparent black as white and locally we draw it as black.

PR #43111, opened 2023-06-22
[web] Move web-only initialization APIs to `dart:ui_web`

mdebbar:
| Old API in `dart:ui` | New API in `dart:ui_web` |
|-|-|
| ~`webOnlyInitializePlatform`~ | ~`ui_web.initializePlatform`~ |
| `webOnlyWarmupEngine` | `ui_web.bootstrapEngine` |
| `debugEmulateFlutterTesterEnvironment` | `ui_web.debugEmulateFlutterTesterEnvironment` |
| `webOnlySetPluginHandler` | `ui_web.setPluginHandler` |

Part of https://github.com/flutter/flutter/issues/126831


lib/web_ui/lib/ui_web/src/ui_web/initialization.dart
```diff
@@ -0,0 +1,66 @@
...
+/// the [FlutterLoaderExtension.didCreateEngineInitializer] JS callback, to
+/// delegate bootstrapping the app to the programmer.
+///
+/// If said callback is not defined, this assumes that the Flutter Web app is
+/// initializing "automatically", as was normal before this feature was
+/// introduced. This will immediately run the `initializeEngine` and `runApp`
+/// methods (via [AppBootstrap.autoStart]).
+///
+/// This method should NOT trigger the download of any additional resources
+/// (except when the app is in "autoStart" mode).
```
yjbanov, line 36: Maybe we should move most of this under a `//` doc and write a fresh one that doesn't rely on internal details?

lib/web_ui/lib/ui_web/src/ui_web/initialization.dart
```diff
@@ -0,0 +1,66 @@
...
+Future<void> warmupEngine({
+  Function? registerPlugins,
+  Function? runApp,
```
yjbanov, line 39: Can we use more specific function signatures? Looks like they are just `void Function()`?

lib/web_ui/lib/initialization.dart
```diff
@@ -23,94 +23,66 @@
...
-    engine.domWindow.console.debug('Flutter Web Bootstrap: Auto.');
-    await bootstrap.autoStart();
-  } else {
-    // Yield control of the bootstrap procedure to the user.
-    engine.domWindow.console.debug('Flutter Web Bootstrap: Programmatic.');
-    loader.didCreateEngineInitializer(bootstrap.prepareEngineInitializer());
-  }
+// TODO(mdebbar): Deprecate this and remove it.
+// https://github.com/flutter/flutter/issues/127395
+bool get debugEmulateFlutterTesterEnvironment {
```
ditman, line 60: Is this something that is needed for end users, or is this only for our testing? If it's only for our engine testing, it probably should live within `engine` or similar, rather than be exposed to `web_ui`?

ditman: (Never mind, this is used all over the bootstrap and the framework!)

yjbanov: Alternatively, this could be hidden behind JS-interop, if we don't want it to show up in the public APIs.

mdebbar: `debugEmulateFlutterTesterEnvironment` has effect on several things including the usage of Ahem font. If users want their tests to produce real text (e.g. for screenshots), they have to disable `debugEmulateFlutterTesterEnvironment`.

lib/web_ui/lib/ui_web/src/ui_web/initialization.dart
```diff
@@ -0,0 +1,66 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:ui/src/engine.dart';
+
+/// Performs a full initialization of the web environment that supports the
+/// Flutter framework.
+///
+/// Unlike [warmupEngine], this method initializes UI and non-UI services.
+Future<void> initializePlatform() async {
```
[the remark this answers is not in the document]
mdebbar: No reason. I just removed the `webOnly` prefix from the name. I think `initializeEngine` makes more sense, and is more consistent with `warmupEngine`.

lib/web_ui/lib/ui_web/src/ui_web/initialization.dart
```diff
@@ -0,0 +1,66 @@
...
+  await initializeEngine();
+}
+
+/// Initializes essential bits of the engine before it fully initializes.
+///
+/// When [FlutterLoaderExtension.didCreateEngineInitializer] is set, it delegates
+/// engine initialization and app startup to the programmer. Else, it immediately
```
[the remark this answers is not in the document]
mdebbar: @ditman any ideas here?

Most people won't see or need this. It's done by `flutter.js` for them. The people who care about this are the ones who want to provide their own replacement for `flutter.js`.

mdebbar:
1. Not moving `webOnlyInitializePlatform` to `dart:ui_web` anymore. We might be able to switch all call sites to use `bootstrapEngine` instead.
2. Renamed `warmupEngine` to `bootstrapEngine`.
3. Updated the docs of `bootstrapEngine`.

lib/web_ui/lib/ui_web/src/ui_web/initialization.dart
```diff
@@ -0,0 +1,57 @@
...
+/// Bootstraps the Flutter Web engine and app.
+///
+/// If the app uses plugins, then the [registerPlugins] callback can be provided
+/// to register those plugins. This is done typically by calling
+/// `registerPlugins` from the auto-generated `web_plugin_registrant.dart` file.
+///
+/// The [runApp] callback is invoked to run the app after the engine is fully
+/// initialized.
+///
+/// For more information, see what the `flutter_tools` doesin the entrypoint
```
yjbanov, line 17: nit: s/doesin/does in/

lib/web_ui/lib/ui_web/src/ui_web/plugins.dart
```diff
@@ -0,0 +1,22 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'dart:typed_data';
+
+import 'package:ui/src/engine.dart';
+import 'package:ui/ui.dart' as ui;
+
+typedef PluginMessageHandler = Future<void> Function(
```
yjbanov, line 10: Does this need a dartdoc?

yjbanov (approved):

lib/web_ui/lib/ui_web/src/ui_web/plugins.dart
mdebbar: I removed this. We already have the same typedef in `dart:ui` so I used that instead.

PR #43115, opened 2023-06-22
Add osx_sdk context to mac_clang_tidy

keyonghan:
With context, recipes logic should be simplified without checking corresponding platform/property conditions.

With the `osx_sdk` context, it will install xcode accordingly.

### DIFF
--- a/ci/builders/mac_clang_tidy.json
+++ b/ci/builders/mac_clang_tidy.json
@@ -60,6 +60,9 @@
                 "host_debug",
                 "ios_debug_sim"
             ],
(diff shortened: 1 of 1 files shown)


godofredoc:
Does this depend on the recipes change?

keyonghan:
> Does this depend on the recipes change?

This is a no-op if landing before the recipes cl changes, but will take effect only after the recipes cl lands.

ricardoamador (approved):
LGTM

godofredoc (approved):

PR #43118, opened 2023-06-22
Add a flag to `ParagraphBuilder` for rounding hack migration

LongCatIsLooong:
The goal is to remove the rounding applied in skparagraph and in the framework: https://github.com/flutter/flutter/issues/31707

The plumbing is done via a new static variable `ParagraphBuilder.shouldDisableRoundingHack` that toggles the rounding behavior in skparagraph and the flag is read by framework code. Application code and test code can either use `ParagraphBuilder.setDisableRoundingHack` or `--dart-define="SKPARAGRAPH_REMOVE_ROUNDING_HACK=1"` to opt-in. 
Once the internal migration is finished the default value of the flag will be set to true.

## Pre-launch Checklist

- [ ] I signed the [CLA].


eyebrowsoffire (commented):
`ParagraphStyle` is part of `dart:ui` which is exposed directly to users, and so I am hesitant to add something temporary like this if we don't have to. I am assuming the plan is to remove this hack completely or shift it to the framework side, and this flag is needed in order to make changes in lock step with the framework. If so, it feels like maybe some sort of global flag that determines this behavior would be easier to remove than piping it through the paragraph style itself, which feels more disruptive to the API overall.

Also, for what it's worth, you probably can just not change any of the skwasm stuff at all. We don't run any framework tests that produce goldens or scubas in Google3 with the skwasm renderer yet, so there are less concerns about the migration there.

LongCatIsLooong:
I removed skwasm-related changes and moved the global flag to `dart:ui`.

jason-simmons (approved):

lib/ui/text.dart
```diff
@@ -3014,6 +3014,28 @@ abstract class ParagraphBuilder {
   /// [Paragraph].
   factory ParagraphBuilder(ParagraphStyle style) = _NativeParagraphBuilder;
 
+  /// Whether the rounding hack enabled by default in SkParagraph and TextPainter
+  /// is disabled.
+  ///
+  /// Do not rely on this getter as it exists for migration purposes only and
+  /// will soon be removed.
+  static bool get shouldDisableRoundingHack {
```
goderbauer, line 3022: Can we mark this one as `@deprecated` to give a stronger signal that it is not to be used? That's how we usually mark temporary things for migration purposes in the framework.

PR #43121, opened 2023-06-23
Add docs for magic variables.

godofredoc:
Adds documentation for env variable substitution in test execution.

Bug: https://github.com/flutter/flutter/issues/129280

## Pre-launch Checklist

- [X] I signed the [CLA].

### DIFF
--- a/ci/builders/README.md
+++ b/ci/builders/README.md
@@ -297,7 +297,12 @@ default is empty which means no interpreter will be used to run the script
 and it is assumed the script is already an executable with the right
 permissions to run in the target platform.
 * **name** - the name of the step running the script.
-* **parameters** - flags or parameters passed to the script.
+* **parameters** - flags or parameters passed to the script. Parameters
+accept magic environment variables(placeholders replaced before executing
+the test). Magic environment variables have the following limitations:
+only `${FLUTTER_LOGS_DIR}` is currently supported and it needs to be used
+alone within the parameter string(e.g. `["${FLUTTER_LOGS_DIR}"]` is OK
+but `["path=${FLUTTER_LOGS_DIR}"]` is not).
 * **Script** - the path to the script to execute relative to the checkout
 directory.
 * **contexts** - a list of available contexts to add to the text execution step.


ci/builders/README.md
```diff
@@ -297,7 +297,12 @@ default is empty which means no interpreter will be used to run the script
 and it is assumed the script is already an executable with the right
 permissions to run in the target platform.
 * **name** - the name of the step running the script.
-* **parameters** - flags or parameters passed to the script.
+* **parameters** - flags or parameters passed to the script. Parameters
+accept magic environment variables(placeholders replaced before executing
```
ricardoamador, line 301: I am not sure I understand here... So this insert the string literal ${FLUTTER_LOG_DIR} for later interpolation? How do you know then what this variable will be? It has to be set somewhere, why do this instead of letting these be set as passed in?

godofredoc: FLUTTER_LOG_DIR is an environment variable set by engine v2 recipes to save logs that will be automatically uploaded to GCS.

Assuming recipes set FLUTTER_LOG_DIR=[tmp]/a/ and we set the parameters to: ['param1', 'param2', '${FLUTTER_LOG_DIR}'] the list of parameters will be translated to ['param1', 'param2', '[tmp]/a'] before the test execution.

This is required because we don't know the sandbox path before the build execution.

ricardoamador: So this is set internally in the recipe? Is it mandatory?

godofredoc: So this is set internally in the recipe?
Yes the env variable is set internally in the recipe. 

Is it mandatory?
No, scripts controlled by flutter should read the env variable directly to know where to put logs. Support for this was added to pass the environment variable value as a script parameter in scripts that we don't own. This is an example: https://github.com/flutter/engine/blob/main/ci/builders/linux_android_aot_engine.json#L114

ci/builders/README.md
```diff
@@ -297,7 +297,12 @@ default is empty which means no interpreter will be used to run the script
...
+the test). Magic environment variables have the following limitations:
+only `${FLUTTER_LOGS_DIR}` is currently supported and it needs to be used
+alone within the parameter string(e.g. `["${FLUTTER_LOGS_DIR}"]` is OK
+but `["path=${FLUTTER_LOGS_DIR}"]` is not).
```
zanderso, line 305: Curious about the reason for this restriction. Could you point me at the implementation?

godofredoc: The restriction is not technical, it was just to simplify the initial implementation. https://cs.opensource.google/flutter/recipes/+/main:recipes/engine_v2/builder.py;l=101

ricardoamador (approved):
LGTM

ci/builders/README.md
zanderso: kk. Is there an issue filed to follow-up on that? Looping over the keys and doing a find/replace seems like a small, easy improvement.

godofredoc: Created a tracking bug: https://github.com/flutter/flutter/issues/129469

PR #5632, opened 2018-06-28
Path iterator

dnfield:
This patch exposes most of `SkPath::Iter` in the engine.  This allows more inspection of `Path` objects - e.g. if you get a path, you could get all the necessary information to serialize it.  

Compared to `PathMetrics` (aka PathMeasure), it gives the actual points/verbs/weights used by Skia for the segments of a path (instead of their lengths, subpaths, and tangents).

The basic use case is "I want to be able to serialize a Path" - which could then be persisted and deserialized, or converted to some other format (e.g. a potential Flutter vector format, or SVG Path data).

Although there is no code currently in the engine that uses `SkPath::Iter` directly, it is used indirectly by other calls already.  At the bottom of it all, this code is really just exposing the verb and point arrays, and the weights for conic segments.

PR #43124, opened 2023-06-23
Allow `Path` and its siblings to be serialized

fzyzcjy:
As usual, I will add tests if you think this looks roughly OK :)

This is a part of https://github.com/flutter/flutter/issues/129098

### Why

In order to record the current frame, the `Path` has to be dumped and restored.

### Will it work for Impeller?

Yes. Currently it uses the `SkPath`'s built-in serialization. However, in the future, if we use `impeller::geometry::Path`, we can still write a simple serialize/deserialize function for it, since the impeller `Path` is nothing but a struct with the following fields:

```c++
class Path {
  FillType fill_ = FillType::kNonZero;
  Convexity convexity_ = Convexity::kUnknown;
  std::vector<ComponentIndexPair> components_;
  std::vector<LinearPathComponent> linears_;
  std::vector<QuadraticPathComponent> quads_;
  std::vector<CubicPathComponent> cubics_;
  std::vector<ContourComponent> contours_;
...
}
```

And all those fields IMHO looks well serializable.

I can make a PR when `SkPath` is removed and impeller Path is used instead, just ping me then :)

## Pre-launch Checklist

- [x] I signed the [CLA].


jonahwilliams (changes requested):
We must not expose a specific serialization structure for path, especially an entirely undocumented one. 

See https://api.skia.org/classSkPath.html#ab34720f80c02d5854cc6b09675b5d907

> the format used for [SkPath](https://api.skia.org/classSkPath.html) in memory is not guaranteed.

fzyzcjy:
@jonahwilliams Thanks for the reply. However, IMHO it is reasonable, because it should be treated as a blackbox. In other words, users never understand what are those bytes, and the only usage of the bytes is to be passed to `Path.deserialize()`. Just like the skia doc says:

> Use only be used in concert with readFromMemory()

Thinking in another view: If the new API (Path.serialize/Path.deserialize) is unreasonable, why does Skia expose writeToMemory/readFromMemory?

dnfield:
See prior discussion at https://github.com/flutter/engine/pull/5632

dnfield:
High level summary: exposing internals of these intentionally opaque objects will make it more difficult to optimize them. Clients will break when we find better ways e.g. to decompose arcs to cubics or cubics to quads or lines and when we do that. We don't want that.

fzyzcjy:
Thank you!

fzyzcjy:
Ah I see even you (hey you guys are definitely expert Flutter engineers in Google!) got rejected for this (similar) proposal as me...

PR #43129, opened 2023-06-23
[Impeller] re-enable buffer to texture blit Vulkan.

jonahwilliams:
I was missing the VMA flush.

Because skia is writing directly to the exposed ptr, we don't go through the OnSetContents which makes a flush call. Add a flush API and implement it for Vk DeviceBuffers.

### DIFF
--- a/lib/ui/painting/image_decoder_impeller.cc
+++ b/lib/ui/painting/image_decoder_impeller.cc
@@ -510,6 +510,9 @@ ImpellerAllocator::ImpellerAllocator(
 
 std::optional<std::shared_ptr<impeller::DeviceBuffer>>
 ImpellerAllocator::GetDeviceBuffer() const {
+  if (buffer_.has_value()) {
+    buffer_.value()->Flush();
+  }
   return buffer_;
 }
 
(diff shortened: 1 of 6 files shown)


lib/ui/painting/image_decoder_impeller.cc
```diff
@@ -510,6 +510,9 @@ ImpellerAllocator::ImpellerAllocator(
 
 std::optional<std::shared_ptr<impeller::DeviceBuffer>>
 ImpellerAllocator::GetDeviceBuffer() const {
+  if (buffer_.has_value()) {
+    buffer_.value()->Flush();
+  }
```
dnfield, line 515: Is this the right time to flush?

Why not flush it explicitly in the blit pass where it's getting used?

jonahwilliams: we only need to flush here because we've written directly to the ptr due to the skia interop. Adding the flush anywhere else would cause extra work, as if the buffer was populated with setcontents that would be redundant

dnfield: I misread which class this was in. LGTM.

dnfield (approved):

PR #43139, opened 2023-06-23
[impeller] Explicitly cast enum class to int before passing it to a formatted string

CaseyHillers:
Next version of the crosstool will start warning about this (see: https://github.com/llvm/llvm-project/issues/38717)

## Pre-launch Checklist

- [X] I signed the [CLA].

### DIFF
--- a/impeller/renderer/blit_pass.cc
+++ b/impeller/renderer/blit_pass.cc
@@ -48,8 +48,8 @@ bool BlitPass::AddCopy(std::shared_ptr<Texture> source,
     VALIDATION_LOG << SPrintF(
         "The source sample count (%d) must match the destination sample count "
         "(%d) for blits.",
-        source->GetTextureDescriptor().sample_count,
-        destination->GetTextureDescriptor().sample_count);
+        static_cast<int>(source->GetTextureDescriptor().sample_count),
+        static_cast<int>(destination->GetTextureDescriptor().sample_count));
     return false;
   }
 


dnfield (approved):
LGTM

impeller/renderer/blit_pass.cc
```diff
@@ -48,8 +48,8 @@ bool BlitPass::AddCopy(std::shared_ptr<Texture> source,
     VALIDATION_LOG << SPrintF(
         "The source sample count (%d) must match the destination sample count "
         "(%d) for blits.",
-        source->GetTextureDescriptor().sample_count,
-        destination->GetTextureDescriptor().sample_count);
+        static_cast<int>(source->GetTextureDescriptor().sample_count),
+        static_cast<int>(destination->GetTextureDescriptor().sample_count));
```
bdero, line 52: Nit: static_cast<std::underlying_type<SampleCount>>?

bdero (approved):

impeller/renderer/blit_pass.cc
bdero: I guess it makes sense to always interpret as int here. The cast should fail if it were even possible for something exotic to happen (won't happen).

alexfh: This should be `int`, since `%d` placeholders require an `int` argument.

Hixie:
test-exempt: fixes lint in newer compiler

PR #43146, opened 2023-06-23
Explicitly cast enum class to int before passing it with a format string

chingjun:
Next version of the crosstool will start warning about this (see: https://github.com/llvm/llvm-project/issues/38717)

For context, see cl/542803761

Reviewers, please leave comment in the internal change, thanks.

### DIFF
--- a/impeller/renderer/blit_pass.cc
+++ b/impeller/renderer/blit_pass.cc
@@ -48,8 +48,8 @@ bool BlitPass::AddCopy(std::shared_ptr<Texture> source,
     VALIDATION_LOG << SPrintF(
         "The source sample count (%d) must match the destination sample count "
         "(%d) for blits.",
-        source->GetTextureDescriptor().sample_count,
-        destination->GetTextureDescriptor().sample_count);
+        static_cast<int>(source->GetTextureDescriptor().sample_count),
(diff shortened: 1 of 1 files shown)


dnfield (approved):
LGTM

impeller/renderer/blit_pass.cc
```diff
@@ -48,8 +48,8 @@ bool BlitPass::AddCopy(std::shared_ptr<Texture> source,
     VALIDATION_LOG << SPrintF(
         "The source sample count (%d) must match the destination sample count "
         "(%d) for blits.",
-        source->GetTextureDescriptor().sample_count,
-        destination->GetTextureDescriptor().sample_count);
+        static_cast<int>(source->GetTextureDescriptor().sample_count),
+        static_cast<int>(destination->GetTextureDescriptor().sample_count));
```
bdero, line 52: Nit: `static_cast<std::underlying_type<SampleCount>>`?

bdero (approved):

impeller/renderer/blit_pass.cc
dnfield: `%d` is for ints.

PR #43149, opened 2023-06-23
Refactor JSNumber.toDart and Object.toJS

srujzs:
JSNumber.toDart will now be two functions: toDartDouble and toDartInt. Note that some code that looks like `toDart.toInt()` was kept as `toDartDouble.toInt()` instead of `toDartInt`, since `toDartInt` throws if the value is not an integer.

Object.toJS is now Object.toJSBox. An actual box is introduced on all backends now, and is unwrapped in JSBoxedDartObject.toDart. There are many usages of toJSAnyShallow that we should modify, but that's for a later CL.

This is to help land this CL: https://dart-review.googlesource.com/c/sdk/+/309082

https://dart-review.googlesource.com/c/sdk/+/309081 is the CL that added the new methods.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/lib/web_ui/lib/src/engine/profiler.dart
+++ b/lib/web_ui/lib/src/engine/profiler.dart
@@ -12,8 +12,7 @@ import 'platform_dispatcher.dart';
 import 'safe_browser_api.dart';
 
 @JS('window._flutter_internal_on_benchmark')
-external JSAny? get _onBenchmark;
-Object? get onBenchmark => _onBenchmark?.toObjectShallow;
+external JSBoxedDartObject? get onBenchmark;
 
 /// A function that computes a value of type [R].
 ///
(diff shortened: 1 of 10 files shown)


lib/web_ui/lib/src/engine/profiler.dart
```diff
@@ -12,8 +12,7 @@ import 'platform_dispatcher.dart';
 import 'safe_browser_api.dart';
 
 @JS('window._flutter_internal_on_benchmark')
-external JSAny? get _onBenchmark;
```
srujzs, line 15: Do note that this means there is a proper box in JS now for this code. It doesn't seem like there will be performance implications but I'm not entirely sure. Here is the framework PR that needs to be landed after this one: https://github.com/flutter/flutter/pull/129436. If we're worried about performance, I'm happy to use the helpers like `toObjectShallow` there, but the framework will need to know whether the platform is Wasm or not to reuse those helpers.

srujzs:
@eyebrowsoffire PTAL, thanks!

eyebrowsoffire (approved):
LGTM! Thanks!

PR #43150, opened 2023-06-23
[iOS] Avoid crash when backdrop filter is null for PlatformViews

cyanglaz:
During certain partial repaint logic, the backdrop filter layer is present but the filter itself is null. We need to account for it when deciding whether to add such filter to the mutator stack.

Fixes https://github.com/flutter/flutter/issues/127095

## Pre-launch Checklist

- [x] I signed the [CLA].


testing/scenario_app/lib/src/platform_view.dart
```diff
@@ -1443,8 +1443,6 @@ class PlatformViewWithOtherBackDropFilter extends PlatformViewScenario {
 /// A simple platform view for testing backDropFilter with a platform view in the scene.
 ///
 /// The stack would look like: picture 1 -> pv1 -> picture 2 -> filter -> pv2 - > picture 3.
-/// Because backdrop filter on platform views has not been implemented(see: https://github.com/flutter/flutter/issues/43902),
```
cyanglaz, line 1446: Removing this out dated comment, the backdrop filter on platform views have been implemented.

cyanglaz:
@hellohuanlin Friendly ping :)

testing/scenario_app/lib/src/platform_view.dart
```diff
@@ -1533,6 +1531,68 @@ class TwoPlatformViewsWithOtherBackDropFilter extends Scenario
   }
 }
 
+/// A simple platform view for testing backDropFilter with a platform view in the scene.
+///
+/// The backdrop filter sigma value is negative, which tries to reproduce a crash, see:
```
hellohuanlin, line 1536: curious what does a negative sigma mean? is it not supposed to happen? maybe make sense to print a warning if so.

hellohuanlin (approved):

testing/scenario_app/lib/src/platform_view.dart
cyanglaz: I'm not entirely sure why the negative sigma is used in one of the example code to reproduce the bug, however, the fix is more about accounting the possibility where the filter object is null. 

As a follow up, I think we also need to disable partial repaint when there is a platform view.

sang-makebetter:
@hellohuanlin Hello, can I ask on which Flutter release that will apply these changes. Thank you!

ChurikiTenna:
Is this fixed yet?
My flutter version is Flutter 3.10.6 • channel stable, and still seeing the crash.

PR #43152, opened 2023-06-23
[Impeller] Supply surface cull rect for Vulkan & GLES dispatchers

bdero:
Similar to Metal, have the DL cull entities early that lie entirely outside the surface texture.

### DIFF
--- a/shell/gpu/gpu_surface_gl_impeller.cc
+++ b/shell/gpu/gpu_surface_gl_impeller.cc
@@ -103,10 +103,13 @@ std::unique_ptr<SurfaceFrame> GPUSurfaceGLImpeller::AcquireFrame(
           return false;
         }
 
+        auto cull_rect =
+            surface->GetTargetRenderPassDescriptor().GetRenderTargetSize();
+
         impeller::DlDispatcher impeller_dispatcher;
         display_list->Dispatch(
             impeller_dispatcher,
-            surface->GetTargetRenderPassDescriptor().GetRenderTargetSize());
+            SkIRect::MakeWH(cull_rect.width, cull_rect.height));
         auto picture = impeller_dispatcher.EndRecordingAsPicture();
 
         return renderer->Render(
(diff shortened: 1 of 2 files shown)


gaaclarke (commented):
LGTM, what are your plans for testing?

jonahwilliams (approved):
LGTM

I don't really think there are reasonable ways to test this sort of thing, besides rendering correctness and performance.

gaaclarke:
> I don't really think there are reasonable ways to test this sort of thing, besides rendering correctness and performance.

Neither of those things would show up if this was broken here.  I plan on writing a test that uses it, we can consider that coverage at least.

jonahwilliams:
You could stand up one of the platform view performance integration tests on Impeller/Android. I think its Skia only right now.

jonahwilliams:
Yes that's why we have a devicelab

shell/gpu/gpu_surface_gl_impeller.cc
```diff
@@ -104,7 +104,9 @@ std::unique_ptr<SurfaceFrame> GPUSurfaceGLImpeller::AcquireFrame(
         }
 
         impeller::DlDispatcher impeller_dispatcher;
-        display_list->Dispatch(impeller_dispatcher);
+        display_list->Dispatch(
```
jonahwilliams, line 107: Note: need to do the SkConversions like in the metal surface.

flar:
Yeah, I don't see a good way to test this without instrumenting the code being tested. Does the Heisenberg Uncertainty Principle apply here?

flar:
You can't test that the culling is happening, but you can test if the culling causes an error maybe. Draw a scene that has a mix of in-bounds, out-of-bounds, and on-the-bounds rendering ops and then make it a golden test. It would have no idea if the out-of-bounds ops got culled from that, but it would test for unintended consequences.

Could we somehow create a franken-render-target that was large but reported a small bounds and then see if the primitives were rendered outside the bounds?

jonahwilliams:
We have benchmarks that are not currently configured to run Impeller, but could. These would show a performance improvement on this change. Benchmark improvements count as tests

bdero:
Looking into `android_view_scroll_perf` for this at the moment.

PR #43162, opened 2023-06-23
[Impeller] Disable blending for Source blend mode

bdero:
All of the backends already respect this property for the first color attachment in the pipeline. ~~Our Vulkan backend disables blending for a whole pipeline if the first color attachment has blending turned off.~~ This has been under our nose but I haven't checked it until now.

HSR might not be working on some Vulkan & OpenGLES drivers because of this.

### DIFF
--- a/impeller/entity/contents/content_context.cc
+++ b/impeller/entity/contents/content_context.cc
@@ -42,6 +42,7 @@ void ContentContextOptions::ApplyToPipelineDescriptor(
       color0.src_color_blend_factor = BlendFactor::kZero;
       break;
     case BlendMode::kSource:
+      color0.blending_enabled = false;
       color0.dst_alpha_blend_factor = BlendFactor::kZero;
       color0.dst_color_blend_factor = BlendFactor::kZero;
       color0.src_alpha_blend_factor = BlendFactor::kOne;


bdero:
Relevant discussion that prompted this: https://docs.google.com/document/d/1Lqf1BRn4uCcUfv9dZlDyAgdSMeQ3FTnPgPjKF9yQ3MI/edit?disco=AAAAz-_REKM

jonahwilliams:
I'm not sure if I understand the description here. 

> Our Vulkan backend disables blending for a whole pipeline if the first color attachment has blending turned off

Is this what you are adding? or a non-deseriable behavior. Sorry, unlcear.

I do remember the discussion about disabling blending and IIRC it made no different for metal, but not surprised if it does for Vulkan.

bdero:
> Is this what you are adding? or a non-deseriable behavior. Sorry, unlcear.

Sorry, disregard -- I explained it wrong in the desc and the behavior of the Vulkan backend is correct. Blend configs (including enabling/disabling) is per-attachment. This patch just flips the (already working) flag off when it makes sense. :)

bdero:
> I do remember the discussion about disabling blending and IIRC it made no different for metal, but not surprised if it does for Vulkan.

Yeah, Apple seems to be smart about it. HSR Just Works when the blend ops match `Source` blending.

jonahwilliams (approved):
LGTM

PR #43163, opened 2023-06-23
[Impeller] Fix float issue rrect blur.

jonahwilliams:
Fixes https://github.com/flutter/flutter/issues/128284

Rearrange the gaussian imports so that we can get away with a single `precision highp float` in the rrect blur shader.

### DIFF
--- a/impeller/compiler/shader_lib/impeller/gaussian_highp.glsl
+++ b/impeller/compiler/shader_lib/impeller/gaussian_highp.glsl
@@ -1,0 +1,23 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#ifndef GAUSSIAN_HIGHP_GLSL_
+#define GAUSSIAN_HIGHP_GLSL_
+
+/// Gaussian distribution function.
+float IPGaussian(float x, float sigma) {
+  // sqrt(2 * pi)
+  const float kSqrtTwoPi = 2.50662827463;
+  float variance = sigma * sigma;
+  return exp(-0.5f * x * x / variance) / (kSqrtTwoPi * sigma);
+}
+
+/// Simpler (but less accurate) approximation of the Gaussian integral.
+vec2 IPVec2FastGaussianIntegral(vec2 x, float sigma) {
+  // sqrt(3)
+  const float kSqrtThree = 1.73205080757;
+  return 1.0 / (1.0 + exp(-kSqrtThree / sigma * x));
+}
+
+#endif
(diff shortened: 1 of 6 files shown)


impeller/compiler/shader_lib/impeller/gaussian_highp.glsl
```diff
@@ -0,0 +1,23 @@
...
+#ifndef GAUSSIAN_HIGHP_GLSL_
+#define GAUSSIAN_HIGHP_GLSL_
+
+/// Gaussian distribution function.
+float IPGaussian(float x, float sigma) {
+  // sqrt(2 * pi)
+  const float kSqrtTwoPi = 2.50662827463;
+  float variance = sigma * sigma;
+  return exp(-0.5f * x * x / variance) / (kSqrtTwoPi * sigma);
+}
```
bdero, line 14: Can we just keep the existing pattern with unique in the same file? We don't mix e.g. `IPHalfGaussian` and `IPVec2FastGaussianIntegral`, but someday we might, and this pattern may get unruly if we wind up expanding it to the rest of the variations.

jonahwilliams: I'm using this is as a trick to not import types.glsl which is where there is a precision mediump float which conflicts with the declaration in rrect. Maybe there is a way to jumble around the includes so that isn't an issue?

bdero: Oh! Would there be any issue with removing the precision declarations from `types.glsl` and just allowing the (pretty good) defaults of `highp` for vertex shaders and `mediump` for fragment shaders, and then allow individual shaders to override the defaults as necessary? I think this will be needed for fixing some shaders on OpenGLES anyhow, since I believe the GLSL ES 1 semantics we need to target for GLES2 only allow configuring precision on a per-shader basis and not on individual vars.

bdero: Maybe we should land this and fix it up in a future PR since it's a very different refactor.

bdero (approved):

impeller/compiler/shader_lib/impeller/gaussian_highp.glsl
jonahwilliams: I think that change would probably be fine. its a bit late to land this anyway (and there are some failing tests anyway), so I'll give that a shot next week and see what malioc says.

jonahwilliams: I tried this here: https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8777362086417627201/+/u/test:_malioc_diff/stdout

Everything appears to be opted into highp float if there is no precision specifier in the file. Could be an issue with shaderc or spirvcross configuration.

As an alternative, I could also remove the precision specifier in types.glsl and then add it to each fragment shader.

bdero: > I could also remove the precision specifier in types.glsl and then add it to each fragment shader.

Explicitly specifying it in every `frag` file seems reasonable to me.

And for future us: Fragment shaders that are using any 32bit values for Metal/highp values for Vulkan will eventually need highp precision for GLES to resolve whatever precision problems we've previously encountered on with GLES/Metal. Of course, we'll probably then need to pepper mediump across the vars that we'd rather keep lower precision for Vulkan in those cases.

jonahwilliams:
Fixed by adjusting specifiers

PR #43168, opened 2023-06-23
[Impeller] Collapse DrawRects into clear colors optimization

gaaclarke:
design doc: https://docs.google.com/document/d/1Lqf1BRn4uCcUfv9dZlDyAgdSMeQ3FTnPgPjKF9yQ3MI/edit

fixes: https://github.com/flutter/flutter/issues/129292

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/impeller/aiks/canvas.cc
+++ b/impeller/aiks/canvas.cc
@@ -246,8 +246,7 @@ void Canvas::DrawRect(Rect rect, const Paint& paint) {
   entity.SetContents(paint.WithFilters(
       paint.CreateContentsForGeometry(Geometry::MakeRect(rect))));
 
-  if (xformation_stack_.size() == 1 &&  // If we're recording the root pass,
-      GetCurrentPass().GetElementCount() == 0 &&  // and this is the first item,
+  if (GetCurrentPass().GetElementCount() == 0 &&
       (paint.blend_mode == BlendMode::kSourceOver ||
        paint.blend_mode == BlendMode::kSource) &&
       paint.color.alpha >= 1.0f) {
--- a/impeller/aiks/canvas.cc
+++ b/impeller/aiks/canvas.cc
@@ -178,8 +178,7 @@ void Canvas::DrawPaint(const Paint& paint) {
   entity.SetBlendMode(paint.blend_mode);
   entity.SetContents(paint.CreateContentsForEntity({}, true));
 
-  if (xformation_stack_.size() == 1 &&  // If we're recording the root pass,
-      GetCurrentPass().GetElementCount() == 0 &&  // and this is the first item,
+  if (GetCurrentPass().GetElementCount() == 0 &&  // and this is the first item,
       (paint.blend_mode == BlendMode::kSourceOver ||
        paint.blend_mode == BlendMode::kSource) &&
       paint.color.alpha >= 1.0f) {
--- a/impeller/aiks/canvas.cc
+++ b/impeller/aiks/canvas.cc
@@ -178,18 +178,7 @@ void Canvas::DrawPaint(const Paint& paint) {
   entity.SetBlendMode(paint.blend_mode);
   entity.SetContents(paint.CreateContentsForEntity({}, true));
 
-  if (GetCurrentPass().GetElementCount() == 0 &&  // and this is the first item,
-      (paint.blend_mode == BlendMode::kSourceOver ||
-       paint.blend_mode == BlendMode::kSource) &&
-      paint.color.alpha >= 1.0f) {
-    // Then we can absorb this drawPaint as the clear color of the pass.
-    GetCurrentPass().AddConditionalClear((EntityPass::ConditionalClear){
-        .color = paint.color,
-        .entity = entity,
-    });
-  } else {
-    GetCurrentPass().AddEntity(entity);
-  }
+  GetCurrentPass().AddEntity(entity);
 }
 
 bool Canvas::AttemptDrawBlurredRRect(const Rect& rect,
@@ -245,15 +234,7 @@ void Canvas::DrawRect(Rect rect, const Paint& paint) {
   entity.SetContents(paint.WithFilters(
       paint.CreateContentsForGeometry(Geometry::MakeRect(rect))));
 
-  if (GetCurrentPass().GetElementCount() == 0 &&
-      (paint.blend_mode == BlendMode::kSourceOver ||
-       paint.blend_mode == BlendMode::kSource) &&
-      paint.color.alpha >= 1.0f) {
-    GetCurrentPass().AddConditionalClear(
-        (EntityPass::ConditionalClear){.color = paint.color, .entity = entity});
-  } else {
-    GetCurrentPass().AddEntity(entity);
-  }
+  GetCurrentPass().AddEntity(entity);
 }
 
 void Canvas::DrawRRect(Rect rect, Scalar corner_radius, const Paint& paint) {
--- a/impeller/aiks/canvas.cc
+++ b/impeller/aiks/canvas.cc
@@ -172,23 +172,25 @@ void Canvas::DrawPath(const Path& path, const Paint& paint) {
 }
 
 void Canvas::DrawPaint(const Paint& paint) {
-  if (xformation_stack_.size() == 1 &&  // If we're recording the root pass,
-      GetCurrentPass().GetElementCount() == 0 &&  // and this is the first item,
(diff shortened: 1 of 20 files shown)


impeller/aiks/canvas.cc
```diff
@@ -172,23 +172,24 @@ void Canvas::DrawPath(const Path& path, const Paint& paint) {
 }
 
 void Canvas::DrawPaint(const Paint& paint) {
-  if (xformation_stack_.size() == 1 &&  // If we're recording the root pass,
```
gaaclarke, line 175: @bdero note that I got rid of this criteria.  I don't think think it's important that it's the root pass, just that the render target for the render pass is covered.  Let me know if there is something I'm missing.

gaaclarke:
This is broken, i'm going to drop it back into draft and look into the failures tomorrow.

bdero (commented):
Okay, so I had a big thonk about this... It occurs to me that Aiks won't be able to pack the `conditional_clears_` list properly in the future when we start reversing the draw order.

What if instead of relying on Aiks to append items to a list, we push this entire concern into the Entities layer by adding the necessary APIs to perform this conversion internally? I.e.:
1. Add `virtual bool Geometry::CoversArea(const Entity& entity, Rect area) const` for determining if the given rect is completely covered by the geometry.
    * Default implementation just returns `false`.
    * Add an override for `CoverGeometry` which returns `true` to handle the `DrawPaint` case that we're already handling.
    * Add an override for `RectGeometry` to handle the new case we're trying to tackle.
2. Add `virtual std::optional<Color> Contents::AsBackgroundColor(const Entity& entity, ISize target_size) const`, which returns a `Color` if the Contents can safely be replaced with a background color.
    * Default implementation just returns `std::nullopt`.
    * Add an override for `SolidColorContents` that returns the color if `geometry_.CoversArea(entity, Rect::MakeSize(Size(target_size)))` returns `true`.
3. Add `std::optional<Color> Entity::AsBackgroundColor(ISize target_size) const` as a helper to call `Contents::AsBackgroundColor` (common pattern).
4. Then, before the render target for a pass is formed in `EntityPass`, preprocess the elements up until either a subpass is found or `Entity::AsBackgroundColor()` returns `std::nullopt`. Set the resulting clear color on the render target.
5. Pass a `start_element` index to `OnRender()` to offset the start of the element iteration.

A design like this would allow us to:
* Make the optimization independent from the `Element` draw order.
* Avoid copying the entire `Element` list.
* Expand this optimization in the future by just adding more `Geometry::CoversArea()` and `Contents::AsBackgroundColor()` overrides.

gaaclarke:
@bdero I'm thinking about it.  It may be cleaner in terms of bookkeeping but complicates the API.

> Avoid copying the entire Element list.

This isn't copying the Element list, it's copying references to the elements.  It isn't ideal but it isn't as bad as copying the elements.

gaaclarke:
I think we're going to have to do like brandon suggested and avoid holding the conditional clears separately.  There is order information getting lost by having that parallel data structure.

gaaclarke:
Transitioning this back from a draft to allow golden image testing since the tests are passing.

gaaclarke:
@bdero yay, goldens are passing.  I'm just going to push one more cleanup then re-request a review.

impeller/entity/contents/solid_color_contents.cc
```diff
@@ -101,4 +101,23 @@ std::unique_ptr<SolidColorContents> SolidColorContents::Make(const Path& path,
...
+  }
+
+  std::optional<Rect> coverage = GetCoverage(entity);
+  if (!coverage.has_value()) {
+    return {};
+  }
+
+  Rect target_rect = Rect::MakeSize(target_size);
+  return coverage.value().Contains(target_rect) ? GetColor()
+                                                : std::optional<Color>();
```
bdero, line 120: Checking if the `Entity` coverage covers the whole pass isn't enough to determine that the drawn geometry will cover all fragments in the subpass. Additional restrictions are needed.

For example, if the geometry is a stroke path that draws a line across the diagonal of the screen, then this check will incorrectly pass.

I'd suggest adding something like `virtual bool Geometry::CoversArea(const Entity& entity, Rect area) const` for determining if the given rect is completely covered by the geometry.
  * Default implementation just returns `false`.
  * Add an override for `CoverGeometry` which returns `true` to handle the `DrawPaint` case that we're already handling.
  * Add an override for `RectGeometry` to handle the new case we're trying to tackle.

For the `RectGeometry` case, checking against the geometry's coverage will work if the entity's transform is known to be translation+scale only (we've already got a helper that can be used to check for that).

impeller/entity/entity_pass.cc
```diff
@@ -896,8 +923,17 @@ void EntityPass::SetClearColor(Color clear_color) {
   clear_color_ = clear_color;
 }
 
-Color EntityPass::GetClearColor() const {
-  return clear_color_;
+Color EntityPass::GetClearColor(ISize target_size) const {
+  Color result = clear_color_;
```
bdero, line 927: Any reason we should keep the `clear_color_`? Can we just remove it along with `SetClearColor` now that we're handling everything below Aiks?

impeller/entity/entity_pass.cc
```diff
@@ -704,7 +719,19 @@ bool EntityPass::OnRender(
...
+    // Skip elements that are incorporated into the clear color.
+    if (is_collapsing_clear_colors) {
+      std::optional<Color> entity_color =
+          AsBackgroundColor(element, root_pass_size);
+      if (entity_color.has_value()) {
+        continue;
+      } else {
+        is_collapsing_clear_colors = false;
+      }
+    }
```
bdero, line 733: This skip behavior LGTM and will be applied to every `EntityPass`, but right now the clear color is only being applied to the root `EntityPass` in this PR (which has it's render target created in `Render()`).

We also need to set the clear color when constructing subpasses, which happens in [`GetEntityForElement()`](https://github.com/flutter/engine/blob/241ca5c1d6be881b4a5cb5a30611ff15de8189a8/impeller/entity/entity_pass.cc#L477-L481). There, you could just use `subpass->GetClearColor()` to grab it and feed it to the RT creation routine.

impeller/entity/contents/solid_color_contents.cc
gaaclarke: Oh thanks.  I wondered about this but somehow convinced myself we were only hitting this with rect geometry.  I'll create a new test for this case.

impeller/entity/entity_pass.cc
gaaclarke: I'll look into it.  I considered it, but since it didn't show up as a requirement to get the tests passing didn't pursue it.

gaaclarke: Done

impeller/entity/contents/solid_color_contents.cc
gaaclarke: Done

impeller/entity/entity_pass.cc
gaaclarke: I'll do this is a follow up PR since it isn't strictly necessary to get the optimization I want in Gallery.  In the gallery the first operation is adjusting the clip actually, although I think it is a noop.  It would be nice to get that to collapse too if it's the case.

bdero: If you defer this fix to a future PR, I think we'll need to add a mechanism to ensure this skipping behavior to only occurs on the root pass, lest some apps will render incorrectly due to the unhandled dropping of entities on subpasses.

gaaclarke: I thought this was just an optimization.  This is surprising that it isn't covered in existing tests.  How could i make sure to hit this in a test?  Would savelayer with a drawrect be enough?

bdero: We just have to be careful to avoid the opacity peephole subpass collapse optimization. Drawing a fullscreen rect in a SaveLayer, but giving that SaveLayer an interesting blend mode like `Modulate` should work.

gaaclarke: Okay done.  I used multiply since it was easier to verify, let me know if you think that may be a problem in the future.  I think that should be good.

bdero (approved):
LGTM!

impeller/entity/entity_pass.cc
bdero: `Multiply` is even more "interesting" than `Modulate`, so we should be good. 😄

PR #43214, opened 2023-06-26
[Impeller] Treat SubOptimalKHR as rotated.

jonahwilliams:
If we don't follow the guidelines at https://developer.android.com/games/optimize/vulkan-prerotation , then any presentations while the screen is rotated will result in suboptimalKHR. Use a change in state across presentations from sucess -> suboptimal or suboptimal -> sucess to detect whether the swapchain needs to be recreated.

Fixes https://github.com/flutter/flutter/issues/129459 (mostly)


chinmaygarde (approved):
I filed https://github.com/flutter/flutter/issues/129580 to followup. But LGTM otherwise.

PR #43215, opened 2023-06-26
[Impeller] Add Vulkan allocator traces.

jonahwilliams:
I suspect this is where we're blocking on the texture uploads right now.


jonahwilliams:
maybe separate allocators will actually work once we've moved PSO construction onto a higher priority thread.

chinmaygarde (approved):

PR #43274, opened 2023-06-27
Initialize skwasm codecs before handing them back to the user.

eyebrowsoffire:
Benchmarks were failing because the code was reading the `frameCount` and `repetitionCount` before reading any frames out of the codec. The codec gets implicitly initialized when you read a frame, but we should return it to the user initialized so that `frameCount` and `repetitionCount` work even if you haven't read a frame yet. This is consistent with how CanvasKit's codec works.

Also, modified our unit tests so that they exercise the codecs in this way.

### DIFF
--- a/lib/web_ui/lib/src/engine/skwasm/skwasm_impl/renderer.dart
+++ b/lib/web_ui/lib/src/engine/skwasm/skwasm_impl/renderer.dart
@@ -369,11 +369,12 @@ class SkwasmRenderer implements Renderer {
     if (contentType == null) {
       throw Exception('Could not determine content type of image from data');
     }
-    final ui.Codec baseDecoder = SkwasmImageDecoder(
+    final SkwasmImageDecoder baseDecoder = SkwasmImageDecoder(
       contentType: contentType,
       dataSource: list.toJS,
       debugSource: 'encoded image bytes',
     );
+    await baseDecoder.initialize();
     if (targetWidth == null && targetHeight == null) {
       return baseDecoder;
     }
@@ -395,11 +396,13 @@ class SkwasmRenderer implements Renderer {
     if (contentType == null) {
       throw Exception('Could not determine content type of image at url $uri');
     }
-    return SkwasmImageDecoder(
+    final SkwasmImageDecoder decoder = SkwasmImageDecoder(
       contentType: contentType,
       dataSource: response.body as JSAny,
       debugSource: uri.toString(),
     );
+    await decoder.initialize();
+    return decoder;
   }
 
   @override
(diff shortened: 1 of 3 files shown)


yjbanov (approved):
lgtm

lib/web_ui/lib/src/engine/skwasm/skwasm_impl/renderer.dart
```diff
@@ -395,11 +396,13 @@ class SkwasmRenderer implements Renderer {
...
       throw Exception('Could not determine content type of image at url $uri');
     }
-    return SkwasmImageDecoder(
+    final SkwasmImageDecoder decoder = SkwasmImageDecoder(
       contentType: contentType,
       dataSource: response.body as JSAny,
       debugSource: uri.toString(),
     );
+    await decoder.initialize();
+    return decoder;
```
ditman, line 405: If `initialize` didn't swallow the returned future type here:

* https://github.com/flutter/engine/blob/6f9cc079605cc71c112a2c8c151fa923e8979e40/lib/web_ui/lib/src/engine/image_decoder.dart#L92-L94

(And returned the initialized `this` or something.

You could just `return decoder.initialize();`

lib/web_ui/lib/src/engine/skwasm/skwasm_impl/renderer.dart
```diff
@@ -369,11 +369,12 @@ class SkwasmRenderer implements Renderer {
     if (contentType == null) {
       throw Exception('Could not determine content type of image from data');
     }
-    final ui.Codec baseDecoder = SkwasmImageDecoder(
+    final SkwasmImageDecoder baseDecoder = SkwasmImageDecoder(
       contentType: contentType,
       dataSource: list.toJS,
       debugSource: 'encoded image bytes',
     );
+    await baseDecoder.initialize();
```
ditman, line 377: Not sure how the error manifested on the failing tests, but it may make sense to spit out a warning on the `SkwasmImageDecoder` if users attempt to call method instances before they call `initialize()`? It doesn't look like an optional method to call, right?

ditman (approved):
LGTM! 🔥

lib/web_ui/lib/src/engine/skwasm/skwasm_impl/renderer.dart
eyebrowsoffire: Yeah, the whole point of this change is that we should never hand the codec back to the user in an uninitialized state. Particularly because the `ui.Codec` API doesn't have the `initialize` method on it. So I don't think there is any reason to have warnings, vending an uninitialized codec is an error of the engine.

PR #43283, opened 2023-06-27
[Impeller] Correct advanced blending behavior.

bdero:
Investigated as part of https://github.com/flutter/flutter/issues/127232.

This took some time to work out, but:
* The amount of blending applied to the source color needs to be weighted by the destination alpha, which is pretty sensible behavior. This is in addition to applying source-over behavior with the resulting blended color.
* All of the blend functions assume that the color is already premultiplied, so remove the unpremultiply/premultiply surrounding the blend call.

All of the blend modes now visually match up with the [Flutter docs](https://api.flutter.dev/flutter/dart-ui/BlendMode.html), except for ColorBurn and Saturation, which appear to have a slight miscalculation going on with the red channel. Going to investigate and fix in a follow-up.

### DIFF
--- a/impeller/entity/shaders/blending/advanced_blend.glsl
+++ b/impeller/entity/shaders/blending/advanced_blend.glsl
@@ -38,16 +38,14 @@ void main() {
                               ) *
                        blend_info.dst_input_alpha;
 
-  f16vec4 dst = IPHalfUnpremultiply(dst_sample);
+  f16vec4 dst = dst_sample;
   f16vec4 src = blend_info.color_factor > 0.0hf
                     ? blend_info.color
-                    : IPHalfUnpremultiply(
-                          Sample(texture_sampler_src,  // sampler
-                                 v_src_texture_coords  // texture coordinates
-                                 ) *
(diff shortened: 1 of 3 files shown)


jonahwilliams (approved):
RSLGTM

bdero:
The malioc diff is looking kinda happy about this. I guess removing the premultiplying ceremony was more impactful than the additional dst.a lerp.

bdero:
This fix to the blend modes does exacerbate the problem that I'm ultimately working on fixing visible in [this issue](https://github.com/flutter/flutter/issues/127232). Prior to this change, we're multiplying the destination alpha into the result for the advanced blend modes, which is wrong for computing the final _color_.
However, that just so happens to emulate limiting the blend output to the geometry mask in a limited number of cases.

bdero:
We have a couple of goldens (`CanRenderForegroundAdvancedBlendWithMaskBlur` and `ForegroundBlendSubpassCollapseOptimization`) that reproduce this issue and will get repaired once we fix the underlying problem and handle masks properly.

PR #43293, opened 2023-06-28
[Impeller] Don't decompress into device buffer on Vulkan.

jonahwilliams:
https://github.com/flutter/flutter/issues/129392

It seems that both allocation of the device buffers and the device private textures can compete with raster tasks. I've tried a few different approaches that haven't panned out.

### DIFF
--- a/lib/ui/painting/image_decoder_impeller.cc
+++ b/lib/ui/painting/image_decoder_impeller.cc
@@ -34,6 +34,7 @@ namespace flutter {
 
 namespace {
 
+#ifdef FML_OS_ANDROID
 class MallocDeviceBuffer : public impeller::DeviceBuffer {
  public:
   MallocDeviceBuffer(impeller::DeviceBufferDescriptor desc, uint8_t* data)
@@ -73,6 +74,7 @@ std::shared_ptr<MallocDeviceBuffer> CreateMallocDeviceBuffer(
   return std::make_shared<MallocDeviceBuffer>(
       desc, static_cast<uint8_t*>(malloc(desc.size)));
 }
+#endif  // FML_OS_ANDROID
 
 /**
  *  Loads the gamut as a set of three points (triangle).
--- a/lib/ui/painting/image_decoder_impeller.cc
+++ b/lib/ui/painting/image_decoder_impeller.cc
@@ -33,6 +33,47 @@
 namespace flutter {
 
 namespace {
+
+class MallocDeviceBuffer : public impeller::DeviceBuffer {
+ public:
+  MallocDeviceBuffer(impeller::DeviceBufferDescriptor desc, uint8_t* data)
+      : impeller::DeviceBuffer(desc) {
+    data_ = data;
+  }
+
+  ~MallocDeviceBuffer() override { free(data_); }
+
+  bool SetLabel(const std::string& label) override { return true; }
+
+  bool SetLabel(const std::string& label, impeller::Range range) override {
+    return true;
+  }
+
+  void Flush() override {}
+
+  uint8_t* OnGetContents() const override { return data_; }
+
+  bool OnCopyHostBuffer(const uint8_t* source,
+                        impeller::Range source_range,
+                        size_t offset) override {
+    for (auto i = 0u; i < source_range.length; i++) {
+      data_[offset + i] = source[source_range.offset + i];
+    }
+    return true;
+  }
+
+ private:
+  uint8_t* data_;
+
+  FML_DISALLOW_COPY_AND_ASSIGN(MallocDeviceBuffer);
+};
+
+std::shared_ptr<MallocDeviceBuffer> CreateMallocDeviceBuffer(
+    impeller::DeviceBufferDescriptor desc) {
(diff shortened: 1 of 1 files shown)


jonahwilliams:
This does improve performance noticably for me locally. But I don't think this is the right long term approach.

lib/ui/painting/image_decoder_impeller.cc
```diff
@@ -33,6 +33,49 @@
...
+  void Flush() override {}
+
+  uint8_t* OnGetContents() const override { return data_; }
+
+  bool OnCopyHostBuffer(const uint8_t* source,
+                        impeller::Range source_range,
+                        size_t offset) override {
+    for (auto i = 0u; i < source_range.length; i++) {
+      data_[offset + i] = source[source_range.offset + i];
+    }
```
gaaclarke, line 62:
```suggestion
    memcpy(data_ + offset), source + source_range.offset, source_range.length);
```
`memcpy` will always be as fast or faster than your own loop (usually faster).

lib/ui/painting/image_decoder_impeller.cc
```diff
@@ -33,6 +33,49 @@
 namespace flutter {
 
 namespace {
+
+#ifdef FML_OS_ANDROID
```
gaaclarke, line 37: This doesn't actually need to be conditionally compiled, there's nothing here that requires android.  The usage is where the branching should happen.  The less conditional compilation, the better.

lib/ui/painting/image_decoder_impeller.cc
```diff
@@ -528,7 +575,14 @@ bool ImpellerAllocator::allocPixelRef(SkBitmap* bitmap) {
   descriptor.size = ((bitmap->height() - 1) * bitmap->rowBytes()) +
                     (bitmap->width() * bitmap->bytesPerPixel());
 
+#ifdef FML_OS_ANDROID
```
gaaclarke, line 578: Instead of having 2 conditional compilations you an use conditional compilation to set 1 const so that these 2 branches can be linked semantically

```c++
#ifdef FML_OS_ANDROID
static constexpr bool kShouldUseMallocDeviceBuffer = true;
#else
static constexpr bool kShouldUseMallocDeviceBuffer = false;
#endif
```

lib/ui/painting/image_decoder_impeller.cc
```diff
@@ -33,6 +33,49 @@
...
+ private:
+  uint8_t* data_;
+
+  FML_DISALLOW_COPY_AND_ASSIGN(MallocDeviceBuffer);
+};
+
+std::shared_ptr<MallocDeviceBuffer> CreateMallocDeviceBuffer(
+    impeller::DeviceBufferDescriptor desc) {
+  return std::make_shared<MallocDeviceBuffer>(
+      desc, static_cast<uint8_t*>(malloc(desc.size)));
```
gaaclarke, line 75: nit: since it's an error to `free` memory that wasn't malloc'd, it's prudent to put the malloc in the constructor.

lib/ui/painting/image_decoder_impeller.cc
```diff
@@ -483,7 +526,11 @@ void ImageDecoderImpeller::Decode(fml::RefPtr<ImageDescriptor> descriptor,
                                                  gpu_disabled_switch]() {
           sk_sp<DlImage> image;
           std::string decode_error;
+#ifdef FML_OS_ANDROID
+          if (false) {
+#else
           if (context->GetCapabilities()->SupportsBufferToTextureBlits()) {
```
gaaclarke, line 532: Should this just be a different question for the capabilities?  "ShouldUseBufferToTextureBlits"?

gaaclarke (commented):
So, the theory is that allocating through the vulkan driver is interfering with other work the driver should be doing?  If that's the case, shouldn't we just make a new `impeller::Allocator` instead of putting in a branch in the image decoder logic?

gaaclarke:
> So, the theory is that allocating through the vulkan driver is interfering with other work the driver should be doing?

It just occurred to me that you are probably more suspecting that our `vmaCreateBuffer` calls are using suboptimal settings (and not interested in the other feedback haha)?  `malloc` is just sidestepping those settings.

I checked out VkCreateBufferInfo settings and nothing stuck out to me.  My first thought that is if we are going through some allocation path that would require synchronization with the GPU, like what would happen with a discrete GPU but isn't required in our case.  So, maybe allocating host visible memory would be faster for example since it would know it doesn't need to synchronize.

gaaclarke:
It may be worth considering if a vma pool could help us.  My understanding is they can cut some of the overhead from vmaCreateBuffer.

There is some conversation about the performance of vmaCrateBuffer here: https://github.com/GPUOpen-LibrariesAndSDKs/VulkanMemoryAllocator/issues/178#issuecomment-860520590

> A good workaround is to create custom pool for linear resources, another for OPTIMAL images to keep them separate and use VMA_POOL_CREATE_IGNORE_BUFFER_IMAGE_GRANULARITY_BIT on both.

PR #43294, opened 2023-06-28
[Impeller] Fix advanced CPU blend modes

bdero:
Resolves https://github.com/flutter/flutter/issues/128606.

All CPU blends now match the behavior of Impeller's GPU blends, which (after https://github.com/flutter/engine/pull/43283) closely matches Skia's blends! Now we can finally use these.

https://github.com/flutter/engine/assets/919017/f9dc7323-fd14-4cdc-ba8b-930622be4206

### DIFF
--- a/impeller/aiks/aiks_unittests.cc
+++ b/impeller/aiks/aiks_unittests.cc
@@ -2098,7 +2098,7 @@ static Picture BlendModeTest(BlendMode blend_mode,
   ///
 
   canvas.Save();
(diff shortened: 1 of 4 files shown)


jonahwilliams (approved):
LGTM!

bdero:
I refactored this to replace the internal blend applicator lambdas with inlined functions because they were permanently capturing values between invocations of `Color::Blend` on a subset of targets. Really wacky stuff.

PR #43309, opened 2023-06-28
[fuchsia] Remove strict equality check for SkMatrix comparison for Fl…

uysalere:
Flatland version of https://github.com/flutter/engine/commit/a4775c7a7bc3dfa158ce83b62f8d89394ec65c15

Bug: fxb/126648

### DIFF
--- a/shell/platform/fuchsia/flutter/flatland_external_view_embedder.cc
+++ b/shell/platform/fuchsia/flutter/flatland_external_view_embedder.cc
@@ -235,8 +235,26 @@ void FlatlandExternalViewEmbedder::SubmitFrame(
         const ViewMutators view_mutators =
             ParseMutatorStack(view_params.mutatorsStack());
         const SkSize view_size = view_params.sizePoints();
-        FML_CHECK(view_mutators.total_transform ==
-                  view_params.transformMatrix());
+
+        // Verify that we're unpacking the mutators' transform matrix correctly.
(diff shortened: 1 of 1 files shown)


shell/platform/fuchsia/flutter/flatland_external_view_embedder.cc
```diff
@@ -235,8 +235,29 @@ void FlatlandExternalViewEmbedder::SubmitFrame(
...
             ParseMutatorStack(view_params.mutatorsStack());
         const SkSize view_size = view_params.sizePoints();
-        FML_CHECK(view_mutators.total_transform ==
-                  view_params.transformMatrix());
+
+        // Verify that we're unpacking the mutators' transform matrix correctly
+        // on debug builds. Use built-in get method for SkMatrix to get values
+        // See:
+        // https://source.corp.google.com/piper///depot/google3/third_party/skia/HEAD/include/core/SkMatrix.h;l=391
+#ifdef NDEBUG
```
arbreng, line 243: Doing this on only debug builds is useless because we never run C++ debug builds, even during developement.

Please remove the #ifdef entirely and change the FML_LOG below from FATAL to an ERROR.

arbreng (approved):

shell/platform/fuchsia/flutter/flatland_external_view_embedder.cc
uysalere: Done

PR #43315, opened 2023-06-28
Handle null images in Metal Screenshot mechanism

flar:
Recent resource leaks discovered in the way Metal generates its screenshots resulted in null pointers being returned from various methods and then objects would try to either release those null pointers or dereference them. While this PR doesn't address any test code that might get a null screenshot from the mechanism (which shouldn't happen in practice unless a test fails), at least it fixes internal cases where we might have null pointers and try to access or release them within the mechanism itself.

### DIFF
--- a/impeller/golden_tests/metal_screenshoter.mm
+++ b/impeller/golden_tests/metal_screenshoter.mm
@@ -51,6 +51,9 @@ std::unique_ptr<MetalScreenshot> MetalScreenshoter::MakeScreenshot(
 
   CGImageRef cgImage = [cicontext createCGImage:flipped
                                        fromRect:[ciImage extent]];
+  if (!cgImage) {
+    return {};
+  }
 
   return std::unique_ptr<MetalScreenshot>(new MetalScreenshot(cgImage));
 }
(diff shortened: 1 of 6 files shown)


impeller/golden_tests/metal_screenshoter.mm
```diff
@@ -51,6 +51,9 @@
 
   CGImageRef cgImage = [cicontext createCGImage:flipped
                                        fromRect:[ciImage extent]];
+  if (!cgImage) {
+    return {};
```
gaaclarke, line 55: Instead of making MetalScreenshot exist in a world where it can be a screenshot or not I'd prefer to make it mean a screenshot, then MakeScreenshot could return a `std::optional<std::unique_ptr<MetalScreenshot>>`.

gaaclarke (changes requested):

impeller/golden_tests/metal_screenshoter.mm
gaaclarke: Or just have it return nullptr.

flar: Why is it a unique ptr instead of shared?

flar: Also optional would seem to be more appropriate when the answer could validly be "no", as in there is a documentable reason not to produce a screen shot.

Failing to produce a screen shot here isn't really "the answer is no", it's more "Oops".

gaaclarke: > Why is it a unique ptr instead of shared?

Because there is just one owner.

> Failing to produce a screen shot here isn't really "the answer is no", it's more "Oops".

You could use fml::status_or then.  Either is fine with me.

flar: Why is there just one owner? Why can't a save a screen shot aside for later reference and it can be deallocated when nobody needs it any more. Note that even if the CG data needs to have one owner, MetalScreenshot could be shared.

In any case I'll close the PR. It's really only valuable while we have a resource leak that causes lots of screen shots to crash which Jason is already looking into. If it is valuable for other reasons, I'll let you manage that.

PR #43341, opened 2023-06-29
Use Windows Display Language

yaakovschectman:
Get the Windows Display Language for locale selection instead of the preferred languages.

https://github.com/flutter/flutter/issues/129786

## Pre-launch Checklist

- [ ] I signed the [CLA].


shell/platform/windows/system_utils_unittests.cc
```diff
@@ -12,38 +12,6 @@
 namespace flutter {
 namespace testing {
 
-class MockWindowsRegistry : public WindowsRegistry {
```
yaakovschectman, line 15: The `WindowsRegistry` class is now unused. I am unsure if we should remove it altogether or not.

shell/platform/windows/system_utils.cc
```diff
@@ -54,33 +54,9 @@ std::wstring GetPreferredLanguagesFromMUI() {
 std::vector<std::wstring> GetPreferredLanguages(
```
yaakovschectman, line 54: We now have no more use for `GetPreferredLanguagesFromRegistry`. Should we remove the function altogether?

loic-sharma: Yes please remove this. If needed, we can always use git's history to resurrect the code.

shell/platform/windows/system_utils_unittests.cc
loic-sharma: Sadly, yes please remove this. We can use git's history to resurrect this code when we need the registry again.

yaakovschectman:
Hmmm... it looks like the devicelab machine may actually return an empty list for `GetThreadPreferredUILanguages`, in which case, unit testing is going to be tricky...

loic-sharma:
> it looks like the devicelab machine may actually return an empty list for GetThreadPreferredUILanguages, in which case, unit testing is going to be tricky...

I'd consider using the [`WindowsProcTable`](https://github.com/flutter/engine/blob/f091085377774c31f872dc01de2c5c25e4932fbd/shell/platform/windows/windows_proc_table.h#L17) here. Currently the `WindowsProcTable` is used for dynamic loading of APIs that are unavailable on older Windows versions, however, we can expand it to also be used to mock out Win32 APIs.

The `WindowsProcTable` would expose a virtual method that mirrors the `GetThreadPreferredUILanguages` Win32 API. The implementation would call the `GetThreadPreferredUILanguages` API directly (there's no need to dynamically load this API), and tests would override this implementation using mocks.

shell/platform/windows/system_utils.cc
```diff
@@ -8,79 +8,48 @@
...
-    return std::wstring();
-  }
-  return buffer;
-}
-
 std::wstring GetPreferredLanguagesFromMUI() {
   ULONG buffer_size;
   ULONG count = 0;
   DWORD flags = MUI_LANGUAGE_NAME | MUI_UI_FALLBACK;
   if (!GetThreadPreferredUILanguages(flags, &count, nullptr, &buffer_size)) {
```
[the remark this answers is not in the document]
yaakovschectman: It looks like they are using the same method as here, there. And re: your comment on `WindowsProcTable`, this shouldn't be necessary. Looks like I missed a nuance of the parameters to the function that didn't make a difference locally but did on devicelab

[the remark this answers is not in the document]
loic-sharma: > It looks like they are using the same method as here, there

Ah you're right, I got confused by their helper's name.

> And re: your comment on WindowsProcTable, this shouldn't be necessary.

The current tests depend on the machines environment. For example, the tests would fail if somehow a machine was configured such that `GetThreadPreferredUILanguages` returns no languages.

In my mind, the ideal test would be an integration test that mocks `GetThreadPreferredUILanguages` and then checks that the Windows embedder sends the expected `FlutterLocale` to the embedder API at startup. Through our mocks we can guarantee that the machine's language configuration can't cause unexpected test flakes.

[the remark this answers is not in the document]
loic-sharma: > Are we sure that:
> This is a concern, e.g. that the environment can return no preferred languages?

This is a hypothetical scenario, yes. Another hypothetical scenario would be that `GetThreadPreferredUILanguages` fails and returns `false`. These hypothetical scenarios would cause flakes. To reduce this likelihood, our tests should be deterministic and we should control their inputs.

> Doing so will not obscure the purpose of this test?

My only concern is using the real `GetThreadPreferredUILanguages` implementation in the test. I'm flexible on the test itself, I'd be happy with your current test too if `GetThreadPreferredUILanguages` was mocked out :)

shell/platform/windows/windows_proc_table.h
```diff
@@ -26,6 +26,15 @@ class WindowsProcTable {
   virtual BOOL GetPointerType(UINT32 pointer_id,
                               POINTER_INPUT_TYPE* pointer_type);
 
+  // Get the preferred languages for the thread, and optionally the process,
+  // and system, in that order, depending on the flags.
+  // See
+  // https://learn.microsoft.com/en-us/windows/win32/api/winnls/nf-winnls-getthreadpreferreduilanguages
```
loic-sharma, line 32:
```suggestion
  // https://learn.microsoft.com/windows/win32/api/winnls/nf-winnls-getthreadpreferreduilanguages
```

shell/platform/windows/windows_proc_table.h
```diff
@@ -26,6 +26,15 @@ class WindowsProcTable {
   virtual BOOL GetPointerType(UINT32 pointer_id,
                               POINTER_INPUT_TYPE* pointer_type);
 
+  // Get the preferred languages for the thread, and optionally the process,
+  // and system, in that order, depending on the flags.
+  // See
+  // https://learn.microsoft.com/windows/win32/api/winnls/nf-winnls-getthreadpreferreduilanguages
+  virtual LRESULT GetThreadPreferredUILanguages(DWORD flags,
```
loic-sharma, line 33: Could you update the comment starting on line 15 for `WindowsProcTable` to mention this can also be used to mock out Windows APIs?

shell/platform/windows/system_utils_unittests.cc
```diff
@@ -7,71 +7,46 @@
...
 
 TEST(SystemUtils, GetPreferredLanguages) {
-  WindowsRegistry registry;
-  std::vector<std::wstring> languages = GetPreferredLanguages(registry);
+  MockWindowsProcTable proc_table;
+  ON_CALL(proc_table, GetThreadPreferredUILanguages)
+      .WillByDefault(
+          [](DWORD flags, PULONG count, PZZWSTR languages, PULONG size) {
+            static const wchar_t lang[] = L"en-US\0\0";
+            static const size_t lang_len = sizeof(lang) / sizeof(wchar_t);
```
loic-sharma, line 32: Is `sizeof(lang)` right here? I'd expect this to be `wcslen`

yaakovschectman: `lang` is a `wchar_t[...]`, so `sizeof(lang) / sizeof(wchar_t)` returns the number of characters in `lang` including the terminating null. I've confirmed that, when the extra null I erroneously added is removed (which I will push soon), the expression evaluates to 7, which is correct.

loic-sharma: Huh, TIL, I would've expected that to return the size of a pointer. Thanks for the details!

loic-sharma (approved):
Thanks for fixing this, excellent work!

PR #37677, opened 2022-11-16
[Impeller] Reland: Refactor color source resolution to use explicit factory types

bdero:
This reverts commit d721647766ddf697d26f05f4ac56d4142ef77742.

Relands https://github.com/flutter/engine/pull/37656, but with with a missing null check (reverted in https://github.com/flutter/engine/pull/37673)

PR #43342, opened 2023-06-29
[Impeller] Add explicit color filter types to Aiks.

bdero:
Part of https://github.com/flutter/flutter/issues/127232.

Adds an explicit ColorFilter concept to Aiks which wraps the GPU and CPU implementations. We're not using the CPU implementations yet, but I'll be getting it wired up in a follow-up patch.


jonahwilliams (approved):
IIRC the last time we tried to land something like this all of the rendering broke in the gallery. Did we ever root cause that? 

The change itself LGTM

bdero:
Hmm, I don't recall landing something like this. Might you be referring to one of my previously failed ColorSource refactors that I had to roll back due to CPU benchmark regressions? I wound up taking a different approach for that in the end.

In terms of avoiding fidelity regressions I think we've been doing pretty well now that we have a lot of stuff covered by goldens.

jonahwilliams:
I'm thinking of this one: https://github.com/flutter/engine/pull/37677

jonahwilliams:
I guess that was color sources and not filters.

PR #43348, opened 2023-06-29
[Impeller] Remove blend mode restrictions on clear color optimization

bdero:
Now that all of the CPU blend modes work, we can allow back-to-back fullscreen solid color draws to get blended together regardless of their blend mode/color opacity.

### DIFF
--- a/impeller/entity/contents/contents.h
+++ b/impeller/entity/contents/contents.h
@@ -117,10 +117,14 @@ class Contents {
   ///        Use of this method is invalid if CanAcceptOpacity returns false.
   virtual void SetInheritedOpacity(Scalar opacity);
 
+  /// @brief Returns a color if this Contents will flood the given `target_size`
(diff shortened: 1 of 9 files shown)


impeller/entity/contents/contents.h
```diff
@@ -117,10 +117,14 @@ class Contents {
   ///        Use of this method is invalid if CanAcceptOpacity returns false.
   virtual void SetInheritedOpacity(Scalar opacity);
 
+  /// @brief Returns a color if this Contents will flood the given `target_size`
+  ///        with a color. This output color is the "Source" color that will be
+  ///        used for the Entity's blend operation.
+  ///
+  ///        This is useful for absorbing full screen solid color draws into
+  ///        subpass clear colors.
```
gaaclarke, line 125: 👍

gaaclarke (approved):
LGTM, nice

PR #43353, opened 2023-06-30
Apply the transform of an image filter layer to paint bounds in the CanvasKit backend

jason-simmons:
Fixes https://github.com/flutter/flutter/issues/128788


hterkelsen (approved):
LGTM

hterkelsen:
Note that once https://skia-review.googlesource.com/c/skia/+/721299 lands this code should be unnecessary since we can just use Skia to do this calculation

PR #43361, opened 2023-06-30
[canvaskit] Enable CanvasKit to compute tight SkPicture bounds

hterkelsen:
Passes the `computeBounds` bool argument to `PictureRecorder.beginRecording`. This will allow us to get tight bounds of the draw commands in the Picture, which enables us to optimize away Pictures which don't paint anything, or optimize away overlays that don't overlap with Platform views.

## Pre-launch Checklist

- [x] I signed the [CLA].


lib/web_ui/test/canvaskit/picture_test.dart
```diff
@@ -68,6 +68,61 @@ void testMain() {
       expect(data!.lengthInBytes, 10 * 15 * 4);
       expect(data.buffer.asUint32List().first, color.value);
     });
-  // TODO(hterkelsen): https://github.com/flutter/flutter/issues/60040
+
+    test('cullRect bounds are tight', () async {
```
yjbanov, line 72: Should we also test cases where the estimated size blows up to infinity? E.g. when the app does `canvas.drawColor` or `canvas.drawPaint`? What's the expected value of `cullRect`, and how should it be interpreted by the engine code?

hterkelsen: In this case it should default to the cullRect passed in when the PictureRecorder begins recording (we use ui.Rect.largest). I added a test to confirm this.

yjbanov (approved):
LGTM but an extra LGTM from @eyebrowsoffire would be nice, especially from the WasmGC point of view.

lib/web_ui/lib/src/engine/canvaskit/picture.dart
```diff
@@ -15,17 +15,18 @@ import 'surface_factory.dart';
 
 /// Implements [ui.Picture] on top of [SkPicture].
 class CkPicture implements ui.Picture {
```
eyebrowsoffire, line 17: I actually made this abstract class that you might want to conform to here: https://github.com/flutter/engine/blame/893ab3bf7bb9cf104846ec50b79baf4864089253/lib/web_ui/lib/src/engine/scene_painting.dart#L15C3-L15C3

I actually wrote Skwasm's scene builder in an agnostic way where it could theoretically be reused for any renderer that implements the normal `dart:ui` interfaces plus a few odds and ends that aren't exposed (such as this `cullRect` function). It might be interesting to see if CanvasKit could be migrated to that scene builder and do some perf measurements at some point, I think it does a little less work in some cases than the CanvasKit one (although it doesn't support platform views yet, so we wouldn't want to switch over to that until the platform views are implemented).

I also think a getter for `cullRect` makes more sense than a function, but up to you.

eyebrowsoffire (changes requested):
One other bounds calculation gotcha I had to deal with when I was implementing Skwasm was the `ImageFilterEngineLayer`. The filter itself might actually change the bounds of the content it is filtering. Skia has an API called `filterBounds` on their image filters that takes an input cull rect and gives the output cull rect. They do this on the native side as well. See https://github.com/flutter/engine/blame/40a8732a5de0e63f2903bf0e565383f9671ce93c/lib/web_ui/lib/src/engine/skwasm/skwasm_impl/layers.dart#L164 for how I did this in Skwasm, I think you'll need a similar adjustment on the CanvasKit side (or else it cuts off some of the content of blurred images, for example)

lib/web_ui/lib/src/engine/canvaskit/picture.dart
hterkelsen: Thanks! I'm getting started on the `filterBounds` CanvasKit change

hterkelsen: https://skia-review.googlesource.com/c/skia/+/721299 is out for review

hterkelsen:
Filed https://github.com/flutter/flutter/issues/130453 to track the work of adjusting the image filter bounds using `filterBounds`

lib/web_ui/lib/src/engine/canvaskit/picture.dart
hterkelsen: Done.

eyebrowsoffire (approved):
LGTM!

hterkelsen:
Fixes https://github.com/flutter/flutter/issues/130453

PR #43364, opened 2023-06-30
Rename default views to implicit views

dkwingsmt:
This PR renames a number of internal variables from "default view ID" to "implicit view ID", as we have pretty much settled on naming this mechanism. Some docs are also fixed accordingly.

## Pre-launch Checklist

- [ ] I signed the [CLA].


shell/platform/darwin/macos/framework/Headers/FlutterPluginRegistrarMacOS.h
```diff
@@ -36,11 +36,9 @@ FLUTTER_DARWIN_EXPORT
 @property(nonnull, readonly) id<FlutterTextureRegistry> textures;
 
 /**
- * The default view displaying Flutter content.
+ * The view displaying Flutter content.
```
loic-sharma, line 39: Given this is a user-facing header file, I'd add more details. We should make it clear this makes single view assumptions. Maybe consider something like:

> The implicit view that displays Flutter content.
>
> This property is provided for backwards compatibility for apps
> that assume a single view.
>
> This method may return |nil|, for instance in a headless environment
> or for apps that do not assume a single view.

loic-sharma: The comment on `FlutterView.h` is excellent and may be worth copying here as well:

```
 * Some single-view APIs will eventually be replaced by their multi-view
 * variant. During the deprecation period, the single-view APIs will coexist with
 * and work with the multi-view APIs as if the other views don't exist.  For
 * backward compatibility, single-view APIs will always operate on the view with
```

dkwingsmt: Done. I was thinking to remove all references to multi-view since from the view of the developers, multi-view isn't supported at all. But I think announcing the plan beforehand also works and makes developers prepared too.

shell/platform/darwin/macos/framework/Headers/FlutterPluginRegistrarMacOS.h
```diff
@@ -36,11 +36,16 @@ FLUTTER_DARWIN_EXPORT
...
  *
- * This method may return |nil|, for instance in a headless environment.
+ * Some single-view APIs will eventually be replaced by their multi-view
+ * variant. During the deprecation period, the single-view APIs will coexist
+ * with and work with the multi-view APIs as if the other views don't exist.
+ * For backward compatibility, single-view APIs will always operate on this
+ * view, which is the first view assigned to the engine.
```
loic-sharma, line 45: > which is the first view assigned to the engine.

This property will be nil if you don't link a view controller to the implicit view.

loic-sharma: > For backward compatibility, single-view APIs will always operate on this view

This conflicts with the "will eventually be replaced" comment. Which single-view APIs do you have in mind here? 1P plugins will move off this API.

shell/platform/darwin/macos/framework/Headers/FlutterPluginRegistrarMacOS.h
```diff
@@ -36,11 +36,16 @@ FLUTTER_DARWIN_EXPORT
...
  *
- * The default view is a special view operated by single-view APIs.
+ * This method may return |nil|, for instance in a headless environment, or when
+ * multi-view is supported in the future and the compatible mode is disabled.
```
loic-sharma, line 48: Consider:

```suggestion
 * multi-view is supported in the future and single-view APIs are deprecated.
```

dkwingsmt: I don't think this is correct. After phase 1, single-view APIs might or might not be deprecated immediately. And the compatible mode (i.e. implicit view enabled) can be enabled or disabled. They are not related. This field will be null when the compatible mode is disabled (or enabled but the implicit view is not assigned.)

dkwingsmt:
I rewrote the doc entirely. Let me know what you think :)

shell/platform/darwin/macos/framework/Headers/FlutterPluginRegistrarMacOS.h
```diff
@@ -36,11 +36,30 @@ FLUTTER_DARWIN_EXPORT
...
+ * all behaviors of "the single view" (which is called "the implicit view") are
+ * preserved, allowing legacy single-view APIs to continue working, while
+ * new-style views created by new ways must be operated by the upcoming
+ * multi-view APIs.
+ *
+ * Plugins written for a single view can keep operating on this view and expect
+ * unchanged behavior for the implicit view. This includes that:
+ *
+ *  - The first view controller attached to the engine will be linked to the
+ *    implicit view.
```
loic-sharma, line 56: This is only true if the implicit view is enabled. I'd consider removing this line, I'm not sure users need to know about this behavior.

dkwingsmt: If the implicit view is not enabled, then this field is null. This field is the implicit view, therefore to explain this field the users need to understand the implicit view.

dkwingsmt: (As far as I understand)

dkwingsmt:
Also cc @goderbauer to take a look at the docs :)

shell/platform/darwin/macos/framework/Headers/FlutterPluginRegistrarMacOS.h
loic-sharma: Agreed, but the line is incorrect if the implicit view is disabled:

> The first view controller attached to the engine will be linked to the
> implicit view.

If the implicit view is disabled, the first view controller will create a new view which won't be the implicit view.

shell/platform/darwin/macos/framework/Source/FlutterEngine_Internal.h
```diff
@@ -127,11 +127,11 @@ typedef NS_ENUM(NSInteger, FlutterAppExitResponse) {
...
  * Dissociate the given view controller from this engine.
  *
- * Practically, since FlutterEngine can only be attached with one controller,
- * the given controller must be the default view controller.
- *
  * If the view controller is not associated with this engine, this call throws an
  * assertion.
+ *
+ * Practically, since FlutterEngine can only be attached with one controller for
+ * now, the given controller must be the current view controller.
```
goderbauer, line 134: I would remove the "for now" here. There isn't enough context here to understand when/how this would change and I don't think it is important for understanding the functionality of the method.

shell/platform/darwin/macos/framework/Headers/FlutterPluginRegistrarMacOS.h
```diff
@@ -36,11 +36,30 @@ FLUTTER_DARWIN_EXPORT
...
+ *  - Single-view Flutter APIs will operate the implicit view.
+ *
+ * This method may return |nil| if the view is not assigned. In single-view
+ * apps, this means that the app is running headlessly. In multi-view apps,
+ * this means that the compatible mode is disabled, or that the compatible mode
```
goderbauer, line 61: We don't define what "compatible mode" is anywhere and if we can avoid it I wouldn't introduce it as a new concept. Can we instead explain this in terms of existing concepts like the implicit view? E.g. maybe we can just say that this being null means the app has been configured to run without an implicit view. Or something like that.

shell/platform/darwin/macos/framework/Headers/FlutterPluginRegistrarMacOS.h
```diff
@@ -36,11 +36,30 @@ FLUTTER_DARWIN_EXPORT
...
 
 /**
- * The default view displaying Flutter content.
+ * The view displaying Flutter content.
  *
- * This method may return |nil|, for instance in a headless environment.
+ * Currently Flutter only supports one view, and this is the view.
  *
- * The default view is a special view operated by single-view APIs.
+ * Flutter plans to support multiple views in the future. Although single-view
```
goderbauer, line 43: I feel a bit uncomfortable about making sooo many concrete promises about how this will work in the future without having an actual implementation to back this up. I would keep this a little lighter and stick to what this does today, maybe with just a one-sentence hint that at some point flutter will support multiple views and at some point after that people will eventually have to migrate away from this API after the proper deprecation period. As we implement more stuff, the doc can evolve then.

Spending 90% of the doc on describing what this may do at some unknown point in the future seems odd.

dkwingsmt: Yeah, that's why my initial PR included nothing but the current status of this property. 
```
/**
 * The view displaying Flutter content.
 *
 * This method may return |nil|, for instance in a headless environment.
 */
```
What do you think of this one?

goderbauer: I like that more than making promises about the future 😝

loic-sharma: I still think we should nudge users away from this property though. If we want to give minimal explanation, perhaps we could reconsider this suggestion: https://github.com/flutter/engine/pull/43364#discussion_r1248322959

```
The implicit view that displays Flutter content.

This property is provided for backwards compatibility for apps
that assume a single view. This will eventually be replaced by
a multi-view API variant.

This method may return |nil|, for instance in a headless environment
or for apps that do not assume a single view.
```

dkwingsmt: @loic-sharma Currently there's nowhere no nudge users to. It's not about being minimal, but that everything it talks about is not there yet. Imagine if you're a user, you see this comment that announces an upcoming deprecation, what can you do? You can not migrate, you just keep worrying until the official support is here?

dkwingsmt: And to be honest the doc here is not the point of this PR at all. I only wanted to rename the variables because my other two PRs will add more similar constants (as you have seen) and I don't want them to use the legacy name. Then I found this doc is outdated so I fixed it too.

goderbauer: Loïc's suggestion in https://github.com/flutter/engine/pull/43364#discussion_r1256172796 also seems fine. It doesn't make any promises about HOW this may work in the future.

dkwingsmt: Ok, I'll use that one.

dkwingsmt:
@loic-sharma @goderbauer Is this PR good to go?

loic-sharma (approved):
Looks good! 🎉

goderbauer (approved):
LGTM

PR #43388, opened 2023-07-03
[Impeller] Add support to embedder for Impeller on GL (via Angle on Windows).

jonahwilliams:
Actually works now, though there is some issue with the default fbo stencil so I've filled https://github.com/flutter/flutter/issues/130048

Other issues:
* ~~Rendering looks wrong~~
* ~~Resizing window hangs~~
* ~~Reactor isn't set up correctly and all blit passes are currently failing.~~
* ~~Needs to handle falling back to sample count of 1 like we do on Android~~.


shell/platform/windows/angle_surface_manager.cc
```diff
@@ -67,13 +67,10 @@ bool AngleSurfaceManager::InitializeEGL(
...
 
 bool AngleSurfaceManager::Initialize() {
-  // TODO(dnfield): Enable MSAA here, see similar code in android_context_gl.cc
-  // Will need to plumb in argument from project bundle for sampling rate.
-  // https://github.com/flutter/flutter/issues/100392
   const EGLint config_attributes[] = {EGL_RED_SIZE,   8, EGL_GREEN_SIZE,   8,
                                       EGL_BLUE_SIZE,  8, EGL_ALPHA_SIZE,   8,
                                       EGL_DEPTH_SIZE, 8, EGL_STENCIL_SIZE, 8,
-                                      EGL_NONE};
+                                      EGL_SAMPLES,    4};
```
jonahwilliams, line 73: We could configure this per backend but turning on MSAA for Skia might not be a bad thing to do? w/e TOOD

impeller/renderer/backend/gles/render_pass_gles.cc
```diff
@@ -450,25 +450,25 @@ struct RenderPassData {
...
-    if (pass_data.discard_stencil_attachment) {
-      attachments.push_back(is_default_fbo ? GL_STENCIL_EXT
-                                           : GL_STENCIL_ATTACHMENT);
-    }
-    gl.DiscardFramebufferEXT(GL_FRAMEBUFFER,      // target
-                             attachments.size(),  // attachments to discard
-                             attachments.data()   // size
-    );
-  }
+  // if (gl.DiscardFramebufferEXT.IsAvailable()) {
```
jonahwilliams, line 453: This was causing the problem, but haven't actually debugged it yet.

shell/platform/embedder/embedder_surface_gl_impeller.cc
```diff
@@ -0,0 +1,198 @@
...
+// |EmbedderSurface|
+std::shared_ptr<impeller::Context>
+EmbedderSurfaceGLImpeller::CreateImpellerContext() const {
+  return impeller_context_;
+}
+
+// |EmbedderSurface|
+sk_sp<GrDirectContext> EmbedderSurfaceGLImpeller::CreateResourceContext()
+    const {
+  sk_sp<GrDirectContext> resource_context;
```
jonahwilliams, line 183: This seems wrong but I'm not sure how else to make it work. 🤷‍♂️

shell/gpu/gpu_surface_gl_impeller.cc
```diff
@@ -80,10 +80,14 @@ std::unique_ptr<SurfaceFrame> GPUSurfaceGLImpeller::AcquireFrame(
     return nullptr;
   }
 
+  GLFrameInfo frame_info = {static_cast<uint32_t>(size.width()),
+                            static_cast<uint32_t>(size.height())};
+  const GLFBOInfo fbo_info = delegate_->GLContextFBO(frame_info);
```
jonahwilliams, line 85: Adding this is enough to get window resizing working.

shell/platform/windows/flutter_windows_engine.cc
```diff
@@ -369,9 +375,16 @@ bool FlutterWindowsEngine::Run(std::string_view entrypoint) {
     args.aot_data = aot_data_.get();
   }
 
-  FlutterRendererConfig renderer_config = surface_manager_
-                                              ? GetOpenGLRendererConfig()
-                                              : GetSoftwareRendererConfig();
+  FlutterRendererConfig renderer_config;
+
+  if (enable_impeller_) {
+    // Impeller does not support a Software backend. Avoid failling back and
```
loic-sharma, line 381:
```suggestion
    // Impeller does not support a Software backend. Avoid falling back and
```

shell/platform/windows/flutter_windows_engine.cc
```diff
@@ -369,9 +375,16 @@ bool FlutterWindowsEngine::Run(std::string_view entrypoint) {
...
+    // confusing the engine on which renderer is selected.
+    renderer_config = GetOpenGLRendererConfig();
```
loic-sharma, line 383: Should we `return false` here if surface creation failed and the `surface_manager_` is null?

loic-sharma: If yes, I'd also test this behavior here using something similar to this: https://github.com/flutter/engine/blob/1de141c6a6e744b30a1de179370aa0805e4e7b94/shell/platform/windows/flutter_windows_engine_unittests.cc#L173

shell/platform/embedder/embedder_surface_gl_impeller.cc
```diff
@@ -0,0 +1,195 @@
...
+    return;
+  }
+
+  auto worker_id = impeller_context_->AddReactorWorker(worker_);
+  if (!worker_id.has_value()) {
+    FML_LOG(ERROR) << "Could not add reactor worker.";
+    return;
+  }
+
+  FML_LOG(ERROR) << "Using the Impeller rendering backend (OpenGL).";
```
loic-sharma, line 89: Is `ERROR` correct here? I'd expect this to be an `INFO` or `WARNING`

jonahwilliams: ERROR is the only message that shows up across the board and is what we're using for other platforms to disambiguate whether or not impeller is running. We may remove this at some point in the future.

loic-sharma: Gotcha, marking as resolved

loic-sharma:
For other folks reviewing, you can test this locally by:

1. Update the tool to support Impeller:

```patch
diff --git a/packages/flutter_tools/lib/src/windows/windows_device.dart b/packages/flutter_tools/lib/src/windows/windows_device.dart
index d94b8bbaef..5fc49d423a 100644
--- a/packages/flutter_tools/lib/src/windows/windows_device.dart
+++ b/packages/flutter_tools/lib/src/windows/windows_device.dart
@@ -40,6 +40,9 @@ class WindowsDevice extends DesktopDevice {
   @override
   String get name => 'Windows';

+  @override
+  bool get supportsImpeller => true;
+
   @override
   Future<TargetPlatform> get targetPlatform async => TargetPlatform.windows_x64;
```

2. Enable impeller in run: `flutter run -d windows --local-engine host_debug_unopt --enable-impeller`

yaakovschectman:
Are we testing this in any way?

jonahwilliams:
we can turn on a devicelab test that runs on impeller, however we've generally kept them in staging (for Android) as we don't want regressions to experimental software to close the tree.

That said, we'll need a devicelab machine that can acquire a hardware surface.

jonahwilliams:
I need to make sure impeller is getting enabled in the test as I expect, but that is waiting for me to get home and get access to a windows machine.

jonahwilliams:
Technically this also allows any other embedder requesting the GL backend to use Impeller too.

shell/platform/windows/flutter_windows_engine.cc
jonahwilliams: Done

shell/platform/windows/flutter_windows_engine.cc
jonahwilliams: Done

shell/platform/windows/flutter_windows_engine.cc
```diff
@@ -369,9 +375,21 @@ bool FlutterWindowsEngine::Run(std::string_view entrypoint) {
...
+  FlutterRendererConfig renderer_config;
+
+  if (enable_impeller_) {
+    if (!surface_manager_) {
+      FML_LOG(ERROR) << "Could not create surface manager. Impeller backend "
+                        "does not support software rendering.";
+      return false;
+    }
+    // Impeller does not support a Software backend. Avoid falling back and
+    // confusing the engine on which renderer is selected.
```
loic-sharma, line 387: This coment is a bit late as at this point we've already checked that we won't use the software backend. Please move this comment up a bit

shell/platform/windows/flutter_windows_engine_unittests.cc
```diff
@@ -223,6 +243,44 @@ TEST_F(FlutterWindowsEngineTest, RunWithoutANGLEUsesSoftware) {
   modifier.embedder_api().Shutdown = [](auto engine) { return kSuccess; };
 }
 
+TEST_F(FlutterWindowsEngineTest, RunWithoutANGLEOnImpellerFailsToStart) {
+  SetEnvironmentVariable("FLUTTER_ENGINE_SWITCHES", "1");
+  SetEnvironmentVariable("FLUTTER_ENGINE_SWITCH_1", "enable-impeller=true");
```
loic-sharma, line 248: Hm, I believe this test runs in parallel with the `FlutterProjectBundle` tests. For example:

https://github.com/flutter/engine/blob/1b44d7ee1a609cb9743940431704f5f1a50653e5/shell/platform/windows/flutter_project_bundle_unittests.cc#L69-L78

If so, we may run into concurrency issues. It looks like `FlutterProjectBundle` supported overriding the switches using `SetSwitches`, but that's currently broken:

https://github.com/flutter/engine/blob/3f50b7d51678a8899ee390ea22c0ba7a03bcdbef/shell/platform/windows/flutter_project_bundle.cc#L85-L92

Would you mind fixing that `SetSwitches` method and then moving this test to `SetSwitches`? Apologies for the mess. Feel free to skip this if the concurrency concern is invalid. Or also feel free to slap a TODO and I can fix it up in a follow-up change.

loic-sharma:
> I need to make sure impeller is getting enabled in the test as I expect, but that is waiting for me to get home and get access to a windows machine.

Impeller is indeed enabled 😄

shell/platform/windows/flutter_windows_engine_unittests.cc
jonahwilliams: Yeah I noticed that but I wasn't sure if this was the correct way to do it or not. If that is supposed to work then happy to fix it, seems like a better way to configure test state..

loic-sharma (approved):
Windows embedder changes LGTM, but consider getting another approval from someone that's more familiar with the engine surface changes

jonahwilliams:
PTAL @chinmaygarde @dnfield

impeller/renderer/backend/gles/render_pass_gles.cc
```diff
@@ -460,7 +461,15 @@ struct RenderPassData {
       attachments.push_back(is_default_fbo ? GL_DEPTH_EXT
                                            : GL_DEPTH_ATTACHMENT);
     }
+
+// TODO(jonahwilliams): discarding the stencil on the default fbo when running
+// on Windows. Causes Angle to discard the entire render target. Until we know
```
dnfield, line 466:
```suggestion
// on Windows causes Angle to discard the entire render target. Until we know
```

shell/platform/embedder/embedder.cc
```diff
@@ -66,7 +66,16 @@ extern const intptr_t kPlatformStrongDillSize;
...
 #include "flutter/shell/platform/embedder/embedder_external_texture_gl.h"
-#endif
+#ifdef IMPELLER_SUPPORTS_RENDERING
+#include "flutter/shell/platform/embedder/embedder_render_target_impeller.h"  // nogncheck
+#include "flutter/shell/platform/embedder/embedder_surface_gl_impeller.h"  // nogncheck
+#include "impeller/core/texture.h"                        // nogncheck
+#include "impeller/renderer/backend/gles/context_gles.h"  // nogncheck
+#include "impeller/renderer/backend/gles/texture_gles.h"  // nogncheck
+#include "impeller/renderer/context.h"                    // nogncheck
+#include "impeller/renderer/render_target.h"              // nogncheck
```
dnfield, line 76: Why are these nognchecked?

shell/platform/embedder/embedder_surface_gl_impeller.cc
```diff
@@ -0,0 +1,195 @@
...
+
+EmbedderSurfaceGLImpeller::EmbedderSurfaceGLImpeller(
+    EmbedderSurfaceGL::GLDispatchTable gl_dispatch_table,
+    bool fbo_reset_after_present,
+    std::shared_ptr<EmbedderExternalViewEmbedder> external_view_embedder)
+    : gl_dispatch_table_(std::move(gl_dispatch_table)),
+      fbo_reset_after_present_(fbo_reset_after_present),
+      external_view_embedder_(std::move(external_view_embedder)),
+      worker_(std::make_shared<ReactorWorker>()) {
+  worker_->SetReactionsAllowedOnCurrentThread(true);
```
dnfield, line 52: nit: should this be further down after the early returns?

shell/platform/embedder/embedder_surface_gl_impeller.cc
```diff
@@ -0,0 +1,195 @@
...
+
+  impeller_context_ =
+      impeller::ContextGLES::Create(std::move(gl), shader_mappings);
+
+  if (!impeller_context_) {
+    FML_LOG(ERROR) << "Could not create Impeller context.";
+    return;
+  }
+
+  auto worker_id = impeller_context_->AddReactorWorker(worker_);
```
dnfield, line 83: e.g. right before this line?

shell/platform/embedder/embedder_surface_gl_impeller.cc
dnfield: I don't think you need this.

You should just see if you can make the resource context current, and if you can mark the worker as allowed to react on this thread. And then always return nullptr.

Returning a valid context from here should not be required and we should not be creating a GrContext when using Impeller.

dnfield (changes requested):
Mostly LG, some nits.

We should do something about the resource context bit before landing this.

shell/platform/embedder/embedder_surface_gl_impeller.cc
jonahwilliams: Done

shell/platform/embedder/embedder_surface_gl_impeller.cc
jonahwilliams: Done

shell/platform/embedder/embedder_surface_gl_impeller.cc
jonahwilliams: Ahh OK, that works and makes much more sense. I briefly read through the code and thought this still needed to be constructed

impeller/renderer/backend/gles/render_pass_gles.cc
jonahwilliams: Done

shell/platform/embedder/embedder.cc
jonahwilliams: Pushing up a change to remove these to see what complains

dnfield (approved):
LGTM once CI is happy

shell/platform/embedder/embedder_surface_gl_impeller.cc
dnfield: The resource context is a performance improving optional thing. Impeller handles things differently by just using worker threads and the reactor mechanism - Skia OTOH expects you to manage a context for the IO thread.

Returning nullptr here should be tolerated anywhere. We may eventually want some better way to signal what this method is actually doing but right now it's a little hard to think about the best way to clean up all the duplicated branches where we've baked Skia concepts into things (like methods that return nullptr for GrContexts in various spots, or nullptr for impeller::Contexts in others).

jonahwilliams:
Ahh the problem was the `#ifdef` instead of `#if`

jonahwilliams:
ahh I guess not, linux fuchsia seems to be trying to enable impeller :(

chinmaygarde (approved):

jonahwilliams:
@dnfield I put the nognchecks back. I'm not sure exactly how to untangle this, but it seems like its mostly a problem with however we're configuring fuchsia builds. Impeller ends up in a partially enabled state and that is confusing the build rules.

PR #43394, opened 2023-07-04
Stop fl_method_call_get_type() not being exported.

robert-ancell:
Causes the build in https://github.com/flutter/flutter/pull/123731 to fail


robert-ancell:
Found better method in https://github.com/flutter/engine/pull/43395

PR #43425, opened 2023-07-05
Check FlutterAppDelegate selector support before calling

gspencergoog:
## Description

This adds checks for the app delegate to make sure that it supports the Flutter-specific selectors before calling them, so that a non `FlutterAppDelegate` can be used for the `NSApplicationDelegate` on `NSApp`.

## Related Issues
 - https://github.com/flutter/flutter/issues/124829
 - https://github.com/flutter/flutter/issues/127476

## Tests
 - No tests yet (and so still draft).


gspencergoog:
@cbracken How would I set the delegate to a non-`FlutterAppDelegate` in a test?

cbracken:
You should be able to just create a class that extends `NSObject <NSApplicationDelegate>` and set the `delegate` property of the current application at runtime. IIRC you'll want to hold on to the previous one then set it back at the end of the test to avoid test ordering bugs.

shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
```diff
@@ -178,9 +178,11 @@ - (instancetype)initWithEngine:(FlutterEngine*)engine
     // allow tests to override it so that an actual exit doesn't occur.
     [[NSApplication sharedApplication] terminate:sender];
   };
-  FlutterAppDelegate* appDelegate =
-      (FlutterAppDelegate*)[[NSApplication sharedApplication] delegate];
-  appDelegate.terminationHandler = self;
+  id<NSApplicationDelegate> appDelegate = [[NSApplication sharedApplication] delegate];
+  if ([[appDelegate class] instancesRespondToSelector:@selector(setTerminationHandler:)]) {
```
cbracken, line 182: Here and below, you should be able to cut out the middle man and just do:
```objc
if ([appDelegate respondsTo:@selector(setTerminationHandler:)]) {
```

cbracken (commented):
Overall lgtm - just the test to do.

gspencergoog (commented):
Okay, this is ready for a "real" review. I added a test.

PR #43429, opened 2023-07-05
Release log capture at end of test

dnfield:
Fixes https://github.com/flutter/flutter/issues/130036

Also fixes a potential issue where the shell wasn't getting destroyed at the end of the test.

Since no log is actually printed ont he second go, the destructor of `LogMessage` never clears the static pointer, and the next test that tries to print a log tries to print to that pointer which is now garbage.

### DIFF
--- a/fml/logging.cc
+++ b/fml/logging.cc
@@ -75,9 +75,21 @@ LogMessage::LogMessage(LogSeverity severity,
 // static
 thread_local std::ostringstream* LogMessage::capture_next_log_stream_ = nullptr;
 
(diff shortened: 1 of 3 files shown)


dkwingsmt:
Requiring a manual cleaning seems volatile. Can we make a guarding object out of it?

dnfield:
> Requiring a manual cleaning seems volatile. Can we make a guarding object out of it?

Refactored this to use a RAII object.

dkwingsmt (approved):
Looks great!

PR #43434, opened 2023-07-06
fix: use 4x4 matrix on TransformLayer from natives

Kingtous:
Hi from [Dora](https://www.dora.run/) team, which is powered by Flutter Web to boost devs to build their 3D websites in just a few seconds. We've encountered the 3d transform problem when running in HTML mode and we found the same issue on the Native engine.

This pr fixes part of https://github.com/flutter/flutter/issues/128480, which fixes the transform running on natives but not Flutter web(html renderer), Because the html renderer needs more context to be told to have 3d transform like `preverse-3d`, etc.

image

I've found that the transform layer doesn't accept a 4x4 transform, which contains Z-axis. Without the information on Z-axis, the 3d nested transform cannot be applied correctly and it will remain 2d transform when combining 2 or more transform layers.

Note that the Flutter web(canvaskit, HTML) is already 4x4 transform but the transform layer from native is 3x3, which has confused the developers and it exactly performs incorrectly. As for more, the reason why flutter web which is rendered by HTML renderer cannot perform well is lacking `preserve-3d` properties on some DOM nodes, which requires further investigating, because after filling the `preserve-3d` on the transform nodes and parents, many golden test cannot be passed due to this property.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/flow/layers/transform_layer.cc
+++ b/flow/layers/transform_layer.cc
@@ -8,9 +8,8 @@
 
 namespace flutter {
 
-TransformLayer::TransformLayer(const SkMatrix& transform)
-    : transform_(transform) {
-  // Checks (in some degree) that SkMatrix transform_ is valid and initialized.
+TransformLayer::TransformLayer(const SkM44& transform) : transform_(transform) {
+  // Checks (in some degree) that SkM44 transform_ is valid and initialized.
   //
   // If transform_ is uninitialized, this assert may look flaky as it doesn't
   // fail all the time, and some rerun may make it pass. But don't ignore it and
@@ -47,7 +46,7 @@ void TransformLayer::Preroll(PrerollContext* context) {
   SkRect child_paint_bounds = SkRect::MakeEmpty();
   PrerollChildren(context, &child_paint_bounds);
 
-  transform_.mapRect(&child_paint_bounds);
+  transform_.asM33().mapRect(&child_paint_bounds);
   set_paint_bounds(child_paint_bounds);
 }
 
--- a/flow/layers/transform_layer.cc
+++ b/flow/layers/transform_layer.cc
@@ -3,7 +3,6 @@
 // found in the LICENSE file.
 
 #include "flutter/flow/layers/transform_layer.h"
-#include "third_party/skia/src/core/SkMatrixPriv.h"
 
 #include <optional>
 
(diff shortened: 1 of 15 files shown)


flow/layers/transform_layer.cc
```diff
@@ -42,12 +41,12 @@ void TransformLayer::Diff(DiffContext* context, const Layer* old_layer) {
...
 void TransformLayer::Preroll(PrerollContext* context) {
   auto mutator = context->state_stack.save();
-  mutator.transform(transform_);
+  mutator.transform(transform_.asM33());
 
   SkRect child_paint_bounds = SkRect::MakeEmpty();
   PrerollChildren(context, &child_paint_bounds);
 
-  transform_.mapRect(&child_paint_bounds);
+  transform_.asM33().mapRect(&child_paint_bounds);
```
dnfield, line 49: We should be using a 4x4 mapping function here. Skia has one in SkMatrixPriv because it makes assumptions it doesn't want to advertise as public, but we can live with those assumptions. Just grab it out from there.

flow/layers/transform_layer.cc
```diff
@@ -42,12 +41,12 @@ void TransformLayer::Diff(DiffContext* context, const Layer* old_layer) {
 
 void TransformLayer::Preroll(PrerollContext* context) {
   auto mutator = context->state_stack.save();
-  mutator.transform(transform_);
+  mutator.transform(transform_.asM33());
```
dnfield, line 44: We should probably file a bug for this, since it'll affect platform views.

dnfield: Specifically a but tracking whether we can make the mutator stack work with full 4x4 matrices.

flow/raster_cache_unittests.cc
```diff
@@ -673,7 +673,7 @@ TEST(RasterCache, PrepareLayerTransform) {
       std::make_shared<DlBlurImageFilter>(5, 5, DlTileMode::kClamp);
   auto blur_layer = std::make_shared<ImageFilterLayer>(blur_filter);
   SkMatrix matrix = SkMatrix::Scale(2, 2);
-  auto transform_layer = std::make_shared<TransformLayer>(matrix);
+  auto transform_layer = std::make_shared<TransformLayer>(SkM44(matrix));
```
dnfield, line 676: (nit) Here and elsewhere: just make these all SkM44 to start with?

lib/ui/painting/matrix.cc
```diff
@@ -32,6 +32,23 @@ SkMatrix ToSkMatrix(const tonic::Float64List& matrix4) {
   return sk_matrix;
 }
 
+SkM44 ToSkM44(const tonic::Float64List& matrix4) {
+  FML_DCHECK(matrix4.data());
+  SkM44 sk_m44;
+  for (int i = 0; i < 16; ++i) {
+    sk_m44.setRC(i % 4, i / 4, SafeNarrow(matrix4[i]));
```
dnfield, line 39: Do we want to assume the compiler is optimizing this nicely for us? These are potentially a lot of instructions to save a few lines of code.

lib/ui/painting/matrix.cc
```diff
@@ -32,6 +32,23 @@ SkMatrix ToSkMatrix(const tonic::Float64List& matrix4) {
...
+  }
+  return sk_m44;
+}
+
+tonic::Float64List ToMatrix4FromSkM44(const SkM44& m44) {
+  tonic::Float64List matrix4(Dart_NewTypedData(Dart_TypedData_kFloat64, 16));
+  for (int i = 0; i < 16; ++i) {
+    matrix4[i] = m44.row(i % 4)[i / 4];
```
dnfield, line 47: Ditto.

dnfield (changes requested):
Mostly good, but let's not drop back to SkMatrix to do the rect mapping.

flow/layers/transform_layer.cc
Kingtous: > 

Yes, I agree. I've found the platform views will affect if it comes to a 4x4 transform which is only an identity.

flow/raster_cache_unittests.cc
Kingtous: I've tried to make all matrix to be SkM44 to start with, but there comes the problem that many classes in the test cases only accept SkMatrix, maybe it will change a lot of codes, which I think maybe it's not quite a good solution?

lib/ui/painting/matrix.cc
Kingtous: This function has no reference and it's unnecessary. I'll delete it :)

flow/layers/transform_layer.cc
Kingtous: May be we need to use `extern` to use this symbol because:

```console
The build continued as if that argument was unspecified.

ERROR at //flutter/flow/layers/transform_layer.cc:6:11: Including a private header.
#include "third_party/skia/src/core/SkMatrixPriv.h"
          ^---------------------------------------
This file is private to the target //third_party/skia:skia
```

dnfield: We can't directly use SkMatrixPriv.h because it's private to Skia. But we can write a method that does the same thing (perhaps copying a lot of Skia code to do it).

dnfield:
@flar ended up fixing this separately.

flar:
> @flar ended up fixing this separately.

Except that I didn't bring over MapRect(SkM44), which I'll now have to integrate...

PR #43442, opened 2023-07-06
Handle nested display list clips in Impeller dispatcher

dnfield:
Fixes https://github.com/flutter/flutter/issues/130084

If a display list is drawn into another display list and the child display list establishes a small clip, subsequent drawing operations are discarded when really they should not be.

The test is expected to render both a blue and a red square; before the fix it renders only the blue square since the red square is incorrectly clipped out.

See also https://github.com/dnfield/flutter_svg/issues/938

### DIFF
--- a/impeller/display_list/dl_dispatcher.cc
+++ b/impeller/display_list/dl_dispatcher.cc
@@ -1052,6 +1052,10 @@ void DlDispatcher::drawDisplayList(
   Matrix saved_initial_matrix = initial_matrix_;
   int restore_count = canvas_.GetSaveCount();
 
+  // The display list may alter the clip, which must be restored to the current
+  // clip at the end of playback.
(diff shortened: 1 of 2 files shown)


impeller/display_list/dl_dispatcher.cc
```diff
@@ -1052,6 +1052,10 @@ void DlDispatcher::drawDisplayList(
   Matrix saved_initial_matrix = initial_matrix_;
   int restore_count = canvas_.GetSaveCount();
 
+  // The display list may alter the clip, which must be restored to the current
+  // clip at the end of playback.
+  canvas_.Save();
```
dnfield, line 1057: (we restore back to restore_count below)

chinmaygarde (approved):

bdero (approved):

PR #43468, opened 2023-07-07
[Impeller] Determine font scaling at render time rather than at canvas draw time

dnfield:
Fixes https://github.com/flutter/flutter/issues/130142

Before this patch, the new test would render blurry text. After it, it renders at the correct scale.

### DIFF
--- a/impeller/aiks/aiks_unittests.cc
+++ b/impeller/aiks/aiks_unittests.cc
@@ -2851,5 +2851,20 @@ TEST_P(AiksTest, CanCanvasDrawPicture) {
   ASSERT_TRUE(OpenPlaygroundHere(canvas.EndRecordingAsPicture()));
 }
 
+TEST_P(AiksTest, DrawPictureWithText) {
+  Canvas subcanvas;
+  ASSERT_TRUE(RenderTextInCanvas(
+      GetContext(), subcanvas,
+      "the quick brown fox jumped over the lazy dog!.?", "Roboto-Regular.ttf"));
+  auto picture = subcanvas.EndRecordingAsPicture();
+
+  Canvas canvas;
+  canvas.Translate({200, 200});
+  canvas.Scale(Vector2(3.5, 3.5));  // The text must not be blurry after this.
+  canvas.DrawPicture(picture);
+
+  ASSERT_TRUE(OpenPlaygroundHere(canvas.EndRecordingAsPicture()));
+}
+
 }  // namespace testing
 }  // namespace impeller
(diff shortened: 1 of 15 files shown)


impeller/typographer/lazy_glyph_atlas.cc
```diff
@@ -50,8 +51,8 @@ std::shared_ptr<GlyphAtlas> LazyGlyphAtlas::CreateOrGetGlyphAtlas(
     i++;
     return &result;
   };
-  auto atlas =
-      text_context->CreateGlyphAtlas(type, std::move(atlas_context), iterator);
+  auto atlas = text_context->CreateGlyphAtlas(type, std::move(atlas_context),
+                                              iterator, scale);
```
bdero, line 55: The scale needs to be updated every time the transformation matrix changes (there's an earlier cache return in this method), otherwise text will render at the wrong size.

impeller/aiks/aiks_unittests.cc
```diff
@@ -2851,5 +2851,28 @@ TEST_P(AiksTest, CanCanvasDrawPicture) {
   ASSERT_TRUE(OpenPlaygroundHere(canvas.EndRecordingAsPicture()));
 }
 
+TEST_P(AiksTest, DrawPictureWithText) {
+  Canvas subcanvas;
+  ASSERT_TRUE(RenderTextInCanvas(
+      GetContext(), subcanvas,
+      "the quick brown fox jumped over the lazy dog!.?", "Roboto-Regular.ttf"));
+  auto picture = subcanvas.EndRecordingAsPicture();
```
bdero, line 2859: Please extend this test to render two text frames at different canvas scales in subcanvas. I don't think this patch renders this scenario correctly:

```
RenderTextInCanvas(...);
canvas.Translate(0, 100);
canvas.Scale(3, 3);
RenderTextInCanvas(...);
```

bdero (commented):
Rather than snaking the scale through the atlas, can we just have `TextContents` update the `TextFrame` with an updated scale metric before handing it off to the atlas?

jonahwilliams (commented):
I'm not sure if this will work as written. The reason we recorded the scales early is that the first time we ask for the atlas, we need to generate it based off all of the glyphs and their metrics.

But now, we don't record the scale. So the first time we ask for the atlas, we generate but we only have the correct scale information for the first text contents. Instead, I think we'd need to visit each text contents to determine its scale.

impeller/typographer/lazy_glyph_atlas.cc
bdero: Er, like @jonahwilliams said, this would actually need to be done for all of the frames here. But I don't think there's enough context to connect those frames to their respective `TextContents` here.

impeller/aiks/aiks_unittests.cc
dnfield: So this works today but ends up with blurry text, because the _first_ call with a scale is winning.

dnfield:
I'm going to close this - discussed what should be a better approach offline.

PR #43474, opened 2023-07-07
[Impeller] Merge lazy glyph atlases in DrawPicture

dnfield:
This is a re-do of https://github.com/flutter/engine/pull/43468

The test has been slightly altered and now the glyph atlas has all the right text sizes in it.

Fixes https://github.com/flutter/flutter/issues/130142

### DIFF
--- a/impeller/entity/contents/text_contents.cc
+++ b/impeller/entity/contents/text_contents.cc
@@ -224,4 +224,17 @@ bool TextContents::Render(const ContentContext& renderer,
                       cmd);
 }
 
+// |Contents|
+void TextContents::AdoptLazyGlyphAtlas(
+    std::shared_ptr<LazyGlyphAtlas> lazy_glyph_atlas,
+    Scalar scale) {
+  FML_DCHECK(lazy_atlas_);
+  if (!lazy_atlas_) {
+    VALIDATION_LOG << "TextContents asked to adopt a lazy glyph atlas, but "
+                      "does not have one to begin with.";
+    return;
+  }
+  lazy_atlas_->Absorb(lazy_glyph_atlas, scale);
+}
+
 }  // namespace impeller
--- a/impeller/typographer/font.h
+++ b/impeller/typographer/font.h
@@ -42,6 +42,8 @@ class Font : public Comparable<Font> {
     Scalar skewX = 0.0f;
     Scalar scaleX = 1.0f;
 
+    Metrics Scaled(Scalar new_scale) const;
+
     constexpr bool operator==(const Metrics& o) const {
       return scale == o.scale && point_size == o.point_size &&
              embolden == o.embolden && skewX == o.skewX && scaleX == o.scaleX;
@@ -63,6 +65,8 @@ class Font : public Comparable<Font> {
 
   const Metrics& GetMetrics() const;
 
+  Font Scaled(Scalar scale) const;
+
   // |Comparable<Font>|
   std::size_t GetHash() const override;
 
@@ -80,6 +84,7 @@ class Font : public Comparable<Font> {
 template <>
 struct std::hash<impeller::Font::Metrics> {
   constexpr std::size_t operator()(const impeller::Font::Metrics& m) const {
-    return fml::HashCombine(m.scale, m.point_size);
+    return fml::HashCombine(m.scale, m.point_size, m.embolden, m.skewX,
+                            m.scaleX);
   }
 };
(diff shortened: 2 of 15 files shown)


impeller/typographer/font.h
```diff
@@ -42,6 +42,8 @@ class Font : public Comparable<Font> {
     Scalar skewX = 0.0f;
     Scalar scaleX = 1.0f;
 
+    Metrics Scaled(Scalar new_scale) const;
```
bdero, line 45: Nit: `appended_scale`?

impeller/entity/contents/text_contents.cc
```diff
@@ -224,4 +224,17 @@ bool TextContents::Render(const ContentContext& renderer,
...
+void TextContents::AdoptLazyGlyphAtlas(
+    std::shared_ptr<LazyGlyphAtlas> lazy_glyph_atlas,
+    Scalar scale) {
+  FML_DCHECK(lazy_atlas_);
+  if (!lazy_atlas_) {
+    VALIDATION_LOG << "TextContents asked to adopt a lazy glyph atlas, but "
+                      "does not have one to begin with.";
+    return;
+  }
+  lazy_atlas_->Absorb(lazy_glyph_atlas, scale);
```
bdero, line 237: I think `Absorb` looks good, but if we do this here, won't we end up pushing copies of all text frames each time we draw text in the picture getting appended?

Could we just performing this merge once during `Canvas::DrawPaint`? Perhaps we can just add a handle to the lazy glyph atlas to the root EntityPass to facilitate this, and then we can just `this->lazy_glyph_atlas_.Absorb(pass->GetGlyphAtlas())` in `Canvas::DrawPaint`, for example.

bdero: Alternatively, we could add a `LazyGlyphAtlas` field to `Picture` and populate it when we end recording. Kind of like that idea better...

impeller/entity/contents/text_contents.cc
```diff
@@ -224,4 +224,17 @@ bool TextContents::Render(const ContentContext& renderer,
                       cmd);
 }
 
+// |Contents|
+void TextContents::AdoptLazyGlyphAtlas(
```
bdero, line 228: It doesn't look like anything here is actually reassigning the `lazy_atlas_`, and so when the picture text renders, it's still going to be using the atlas from the picture canvas.

Since we shouldn't perform the merge here, we could get away with just making the existing glyph atlas setter virtual instead.

impeller/entity/contents/text_contents.cc
bdero: Oh, it looks like the opposite is happening here, actually. We're copying all of the parent Canvas's text frames into the Picture Canvas's lazy atlas.

If we do the opposite and:
1. Copy all of the text frames from the Picture's lazy atlas into the parent Canvas's lazy atlas, scaling them according to the CTM of the parent canvas, and
2. Reassign the `lazy_atlas_` field in all of the Picture's `TextContents`,

then we can avoid the need for tracking the atlases separately.

Mutating contents like this is kind of screwed up in general... it can only work out if the graph of DrawPictures is a DAG with one leaf (since everything will collapse into the same lazy atlas in the end). But things will get screwed up if there are multiple leaves.

bdero: I think we need to support deep copying entities/contents for `EntityPass::Clone()` or come up with an alternative design that involves `EntityPass` supplying contents the lazy atlas at render time.

```cpp
Canvas root, childA, childB;
canvas*.DrawTextFrame(...); // Draw some different text in each of the 3 canvases.

root.DrawPicture(childA.EndRecordingAsPicture());
childB.DrawPicture(childA.EndRecordingAsPicture()); // All of the childA atlas refs are switched to the childB atlas!

// When the childA contents are rendered as part of root, it'll end up rendering to the childB atlas...
// Things would work out here if we forced pictures to only ever be drawn once, though.
root.EndRecordingAsPicture().Render(...);
```

dnfield: Yes, Id di this backwards and am working on fixing it.

Canvas::DrawPicture has no way of getting a "root canvas". There's no such concept.

dnfield: I was assuming that Clone was making a deep copy, since we're updating all the transforms?

dnfield: I see, the entity gets a new copy but the contents are shared...

dnfield:
Talked offline again, I'm going to do yet another try on this.

PR #43479, opened 2023-07-07
[Impeller] join all queue submissions into one.

jonahwilliams:
While we knew that this would reduce encoding time, this may also help scenarios where we accidentally or intentionally destrory images during the raster task thread.


jonahwilliams:
This didn't help as much but it represents another step in concurrent encoding.

jonahwilliams:
So it looks like I was almost right about acquireNextDrawable applying backpressure. There is currently an edge case where once we've exhausted the swapchain, blocking on the next image fence will result in touching a vk fence from multiple threads (since we acquire on a different thread than we present on).I think this is easily fixable by using a frames in flight semaphore in addition to the synchronizer fence, but just an FYI in case you see a synchronization validation error.

PR #43485, opened 2023-07-08
[Impeller] Check for a null command buffer in InlinePassContext::EndPass

jason-simmons:
This could happen if the pass is immediately ended due to clear color optimization.

Fixes https://github.com/flutter/flutter/issues/130167


bdero (approved):
LGTM

PR #43493, opened 2023-07-09
[Impeller] Don't decompress into device buffer for Vulkan/GLES.

jonahwilliams:
My observations on the Pixel 6 device are that performing device allocations from multiple threads can dramatically slow down the raster task workload. As a stopgap solution, we can adjust image upload to only touch the device allocator on the IO thread which reduces the parallel access.

This doesn't have any impact on the S10, but locally on the Pixel 6 it is a night and day difference. I am testing using jonahwilliams/forked_gallery and navigating to the Reply demo. This demo has a large number of images, several of which are quite large.

Work towards https://github.com/flutter/flutter/issues/129392

### Before

Page transition is ~4 frames.

image

### After

Page transition is ~20 frames.

image


lib/ui/painting/image_decoder_impeller.cc
```diff
@@ -483,14 +522,16 @@ void ImageDecoderImpeller::Decode(fml::RefPtr<ImageDescriptor> descriptor,
...
+              context->GetCapabilities()->SupportsBufferToTextureBlits()) {
             std::tie(image, decode_error) = UploadTextureToPrivate(
                 context, bitmap_result.device_buffer, bitmap_result.image_info,
                 bitmap_result.sk_bitmap, gpu_disabled_switch);
             result(image, decode_error);
           } else {
-            std::tie(image, decode_error) = UploadTextureToShared(
+            std::tie(image, decode_error) = UploadTextureToStorage(
                 context, bitmap_result.sk_bitmap, gpu_disabled_switch,
+                impeller::StorageMode::kDevicePrivate,
```
jonahwilliams, line 534: This works because both GLES and Vulkan don't really have a host visible abstraction and always do an upload.

dnfield (approved):

PR #43519, opened 2023-07-10
[Impeller] Apply color filters on the CPU for solid colors & gradients.

bdero:
Apply color filters on the CPU for solid colors and gradients. This fixes https://github.com/flutter/flutter/issues/127232 for the purely color-based color sources.

We still need to make the other special color blending paths respect the geometry mask in order to fully resolve the underlying issue:
* SaveLayers (if they have an interesting matrix filter -- without a matrix filter, it'll always work today)
* TiledTextureContents
* TextContents
* VerticesContents
* AtlasContents

### DIFF
--- a/impeller/aiks/aiks_unittests.cc
+++ b/impeller/aiks/aiks_unittests.cc
@@ -395,6 +395,30 @@ TEST_P(AiksTest, CanRenderLinearGradientDecal) {
   CanRenderLinearGradient(this, Entity::TileMode::kDecal);
 }
 
+TEST_P(AiksTest, CanRenderLinearGradientDecalWithColorFilter) {
+  Canvas canvas;
+  canvas.Scale(GetContentScale());
+  Paint paint;
+  canvas.Translate({100.0f, 0, 0});
+
+  std::vector<Color> colors = {Color{0.9568, 0.2627, 0.2118, 1.0},
+                               Color{0.1294, 0.5882, 0.9529, 0.0}};
+  std::vector<Scalar> stops = {0.0, 1.0};
+
+  paint.color_source = ColorSource::MakeLinearGradient(
+      {0, 0}, {200, 200}, std::move(colors), std::move(stops),
+      Entity::TileMode::kDecal, {});
+  // Overlay the gradient with 50% green. This should appear as the entire
+  // rectangle being drawn with 50% green, including the border area outside the
+  // decal gradient.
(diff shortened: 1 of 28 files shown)


bdero:
Let's see those goldens.

bdero:
Those goldens are looking mighty fine.

jonahwilliams (commented):
How does this work with decal gradients + color blending that changes transparent black?

jonahwilliams:
I'm not really sure what the correct result should be in this case, but I think If you apply a color filter to a decal texture you can observe the transparent black pixels getting changed, where a CPU blend afterwards might not do that?

bdero:
> I'm not really sure what the correct result should be in this case, but I think If you apply a color filter to a decal texture you can observe the transparent black pixels getting changed, where a CPU blend afterwards might not do that?

Not sure if I'm fully answering your question, but:

Today, when applying a color blend to the color source, pixels that lie outside the geometry can be affected, which is incorrect behavior. This patch fixes that behavior for solid colors and gradients.

Fixing it for the other drawing ops will require more expensive tricks. For example, tracking the geometry mask as a separate texture and compositing post-blend, or drawing the geometry outline to a stencil and then using it inside the GPU blend filter.

jonahwilliams:
Here is an example of what I'm talking about: https://dartpad.dev/?id=80d0af639ec8952a3e31bac084fb202b

With decal and certain gradient parameters you can inset the color source in the geometry, and in that case you'd need to handle blending being done with transparent black pixels

bdero:
Ah yeah, that needs to be fixed. I'll fix it by throwing the blended border color in the shader uniforms. It makes sense to allow the border color to be set in our emulated decal anyhow. [Metal allows this](https://developer.apple.com/documentation/metal/mtlsamplerdescriptor/2092299-bordercolor), for example.

bdero:
Great catch, btw. :)

bdero:
Alrighty, added `CanRenderLinearGradientDecalWithColorFilter` to demonstrate that this is handled now.

Before:

Screenshot 2023-07-10 at 6 38 35 PM

After:

Screenshot 2023-07-10 at 6 35 14 PM

bdero:
Goldens are good.

jonahwilliams (approved):
LGTM

At some point we'll need to reckon with how complex we've made gradients...

PR #43535, opened 2023-07-10
[Impeller] Fixes for asymmetric stencil descriptors

johnmccutchan:
Fixes for asymmetric stencil descriptors

- Fix GLES implementation to properly set back and front face options in the case that they aren't identical.
- Add write_mask to hash function.

https://github.com/flutter/flutter/issues/129855

### DIFF
--- a/impeller/renderer/renderer_unittests.cc
+++ b/impeller/renderer/renderer_unittests.cc
@@ -1046,6 +1046,142 @@ TEST_P(RendererTest, VertexBufferBuilder) {
   ASSERT_EQ(vertex_builder.GetVertexCount(), 4u);
 }
 
+TEST_P(RendererTest, StencilMask) {
+  using VS = BoxFadeVertexShader;
+  using FS = BoxFadeFragmentShader;
+  auto context = GetContext();
+  ASSERT_TRUE(context);
+  using BoxFadePipelineBuilder = PipelineBuilder<VS, FS>;
+  auto desc = BoxFadePipelineBuilder::MakeDefaultPipelineDescriptor(*context);
+  ASSERT_TRUE(desc.has_value());
+
+  // Vertex buffer.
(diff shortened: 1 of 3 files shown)


bdero:
Interface LGTM. 👍

impeller/renderer/renderer_unittests.cc
```diff
@@ -1046,6 +1046,137 @@ TEST_P(RendererTest, VertexBufferBuilder) {
...
+      auto pass = buffer->CreateRenderPass(render_target);
+      if (!pass) {
+        return false;
+      }
+      pass->SetLabel("Stencil Buffer");
+      ImGui::Begin("Controls", nullptr, ImGuiWindowFlags_AlwaysAutoResize);
+      ImGui::SliderInt("Stencil Write Value", &stencil_reference_write, 0,
+                       0xFF);
+      ImGui::SliderInt("Stencil Compare Value", &stencil_reference_read, 0,
+                       0xFF);
```
bdero, line 1134: Nit: It would be nice if the front face stencil behavior could be checked separately from this tool. One way would be to:
1. set the `stencil_compare` to `kLessEqual`, and then
2. add an `x_mirror` bool to the controls that, when multiplies the mvp by `Matrix::MakeScale(Vector2(-1, -1))`

bdero (approved):
LGTM, just one suggestion

impeller/renderer/renderer_unittests.cc
johnmccutchan: Done.

bdero: Whoops, means to say invert just x with `Matrix::MakeScale(Vector2(-1, 1))` and then the image would flip + the checkerboard pattern should invert.

bdero: And keep the front face as `kGreater`, but just change the back face to `kLessEqual`.

PR #37394, opened 2022-11-07
Reland layer state stack 2

flar:
This reland fixes the cached rendering issues (see https://github.com/flutter/flutter/issues/114359) resulting from doubling up the transform when preparing a state stack for updating a cache entry.

Many of the performance regressions seem to be better now with these recent fixes, but I've been trying to track down a 10-20% regression in some of the peephole benchmarks. Most notably the "grid of opacity" benchmark (see the 2 recent bumps in [this Skia Perf graph](https://flutter-flutter-perf.skia.org/e/?begin=1663036524&end=1667853776&queries=sub_result%3Daverage_frame_rasterizer_time_millis%26test%3Dopacity_peephole_grid_of_opacity_perf__e2e_summary&requestType=0&selected=commit%3D31839%26name%3D%252Carch%253Dintel%252Cbranch%253Dmaster%252Cconfig%253Ddefault%252Cdevice_type%253DMoto_G__4_%252Cdevice_version%253Dnone%252Chost_type%253Dlinux%252Csub_result%253Daverage_frame_rasterizer_time_millis%252Ctest%253Dopacity_peephole_grid_of_opacity_perf__e2e_summary%252C)). I've tracked this down to the most probable cause being the copies and manipulations of SkM44 matrices that can be reduced or eliminated with more work.

I'm creating this PR now so that some developers can have access to the bug fixes before I fix the last of the performance regressions.

PR #43536, opened 2023-07-10
Use full 4x4 matrix transforms in TransformLayer

flar:
Fixes: https://github.com/flutter/flutter/issues/82961
Fixes: https://github.com/flutter/flutter/issues/113346

The fix was a simple fallout from the previous work to add support for SkM44 throughout the DL and Diff mechanisms (see https://github.com/flutter/flutter/issues/82955, https://github.com/flutter/flutter/issues/116198, https://github.com/flutter/engine/pull/37394)

Tested with its own existing and new unit tests as well as the test case from https://github.com/flutter/flutter/issues/113346

### DIFF
--- a/flow/layers/transform_layer.cc
+++ b/flow/layers/transform_layer.cc
@@ -8,9 +8,8 @@
 
 namespace flutter {
 
-TransformLayer::TransformLayer(const SkMatrix& transform)
-    : transform_(transform) {
-  // Checks (in some degree) that SkMatrix transform_ is valid and initialized.
+TransformLayer::TransformLayer(const SkM44& transform) : transform_(transform) {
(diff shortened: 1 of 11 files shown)


chinmaygarde (approved):

flow/layers/transform_layer.cc
```diff
@@ -47,7 +46,7 @@ void TransformLayer::Preroll(PrerollContext* context) {
   SkRect child_paint_bounds = SkRect::MakeEmpty();
   PrerollChildren(context, &child_paint_bounds);
 
-  transform_.mapRect(&child_paint_bounds);
+  transform_.asM33().mapRect(&child_paint_bounds);
```
dnfield, line 49: We should port over SkMatrixPriv::mapRect for this.

PR #43542, opened 2023-07-11
remove include of SkPicture files from non-Fuchsia sources

flar:
Addresses most of https://github.com/flutter/flutter/issues/128060

Most of the uses of SkPicture and Recorder are removed from the engine sources. The few that remain are:

- DisplayList <-> Skia consistency testing code
- Legacy code only used from Fuchsia
- Dart CanvasKit uses which aren't actually using the local Skia sources or libraries

These are all comment and include file changes and so the testing is in the building.


jonahwilliams (approved):
LGTM

PR #43558, opened 2023-07-11
Listen to window notifications to update application lifecycle

yaakovschectman:
Use `WM_SIZE, WM_SHOWWINDOW, WM_SETFOCUS, and WM_KILLFOCUS` to update the application lifecycle state on Windows.

https://github.com/flutter/flutter/issues/103637

## Pre-launch Checklist

- [ ] I signed the [CLA].


shell/platform/windows/flutter_window_unittests.cc
```diff
@@ -229,6 +230,8 @@ TEST(FlutterWindowTest, OnPointerStarSendsDeviceType) {
                           kDefaultPointerDeviceId, WM_LBUTTONDOWN);
   win32window.OnPointerLeave(10.0, 10.0, kFlutterPointerDeviceKindStylus,
                              kDefaultPointerDeviceId);
+
+  win32window.SetView(nullptr);
```
yaakovschectman, line 234: Added because `FlutterWindow` will try to send a `HIDE` update to the binding delegate upon its deconstruction, and this test sets the binding delegate to the address of a local stack variable, which means it is no longer valid once execution leaves the scope of this test function. Alternatively, we could replace `delegate` with a `unique_ptr` and get its raw pointer before moving it to pass to `EXPECT_CALL`

shell/platform/windows/flutter_window.cc
```diff
@@ -328,4 +336,25 @@ bool FlutterWindow::NeedsVSync() {
   return true;
 }
 
+void FlutterWindow::OnWindowStateEvent(WindowStateEvent event) {
+  switch (event) {
+    case SHOW:
+      restored_ = true;
+      break;
+    case HIDE:
+      restored_ = focused_ = false;
```
cbracken, line 345: Nit: I _thought_ there was something in the google C++ style guide about avoiding chained assignments but I can't for the life of me actually find it. If we do this elsewhere in the code then it's fine to keep it, but if not, we should separate these onto separate lines.

shell/platform/windows/flutter_window_unittests.cc
```diff
@@ -61,6 +61,7 @@ class MockFlutterWindow : public FlutterWindow {
   MOCK_METHOD1(Win32MapVkToChar, uint32_t(uint32_t));
   MOCK_METHOD0(GetPlatformWindow, HWND());
   MOCK_METHOD0(GetAxFragmentRootDelegate, ui::AXFragmentRootDelegateWin*());
+  // MOCK_METHOD1(OnWindowStateEvent, void(WindowStateEvent));
```
cbracken, line 64: Delete this if unneeded.

shell/platform/windows/flutter_window_unittests.cc
```diff
@@ -324,5 +327,30 @@ TEST(FlutterWindowTest, AlertNode) {
...
+      });
+
+  win32window.InjectWindowMessage(WM_SIZE, 0, 0);
+  EXPECT_EQ(last_event, HIDE);
+
+  win32window.InjectWindowMessage(WM_SIZE, 0, MAKEWORD(1, 1));
+  EXPECT_EQ(last_event, SHOW);
+
+  win32window.InjectWindowMessage(WM_SETFOCUS, 0, 0);
+  ;
```
cbracken, line 348: Nit: Delete this line.

shell/platform/windows/windows_lifecycle_manager.cc
```diff
@@ -75,6 +75,24 @@ bool WindowsLifecycleManager::WindowProc(HWND hwnd,
...
+
+    case WM_SHOWWINDOW:
+      if (!wpar) {
+        OnWindowStateEvent(hwnd, HIDE);
+      }
+      break;
+
+    case WM_DESTROY:
+      OnWindowStateEvent(hwnd, HIDE);
+      break;
```
yaakovschectman, line 95: Note in case anyone is curious: The top level runner window receives a SETFOCUS message when it first comes up, but then does not receive further SETFOCUS or KILLFOCUS messages when the Flutter application gains or loses focus, so we are not listening for it here.

shell/platform/windows/flutter_window_unittests.cc
cbracken: From a lifetime perspective, if the delegate is always strictly owned by the window and its lifetime is constrained to (at most) the lifetime of the owning window, then using a unique_ptr seems reasonable. We should do a quick audit of the code to be sure. If not, then I'd add a one-line comment explaining the rationale here as a breadcrumb for future archaeologists.

shell/platform/windows/flutter_window.cc
```diff
@@ -69,16 +69,24 @@ static HCURSOR GetCursorByName(const std::string& cursor_name) {
 }  // namespace
 
 FlutterWindow::FlutterWindow(int width, int height)
-    : binding_handler_delegate_(nullptr) {
+    : binding_handler_delegate_(nullptr), restored_(false), focused_(false) {
```
loic-sharma, line 72: Consider setting these default values on the fields themselves to shorten this line. I don't feel strongly about this

shell/platform/windows/windows_lifecycle_manager.h
```diff
@@ -9,12 +9,23 @@
...
 #include <optional>
+#include <set>
+
+#include "flutter/shell/platform/common/app_lifecycle_state.h"
 
 namespace flutter {
 
 class FlutterWindowsEngine;
 
+enum WindowStateEvent {
```
loic-sharma, line 22: Flutter's style seems to be `enum class` (see [this](https://stackoverflow.com/questions/18335861/why-is-enum-class-preferred-over-plain-enum)) and names like `kShow`, `kHide`, etc...

shell/platform/windows/flutter_window_unittests.cc
yaakovschectman: Actually, it looks to be the other way around. The delegate i.e. `FlutterWindowsView` owns a pointer to the handler i.e. `FlutterWindow`, but where we're using a `MockWindowBindingHandlerDelegate` in these unit tests, the window is just scoped to the function it's created in. For now I will put a comment here, though moving ownership of `win32window` in these tests to inside the mock delegate may be feasible.

shell/platform/windows/flutter_window.cc
gspencergoog: Even if it's not explicitly in the style guide, I would consider this less readable (it's easy to miss one of the assignments if you don't read carefully).

shell/platform/windows/flutter_window.h
```diff
@@ -174,6 +177,12 @@ class FlutterWindow : public Window, public WindowBindingHandler {
   RECT cursor_rect_;
 
   FML_DISALLOW_COPY_AND_ASSIGN(FlutterWindow);
+
+  // The window receives resize and focus messages before its view is set, so
+  // these values cache the state of the window in the meantime so that the
+  // proper application lifecycle state can be updated once the view is set.
+  bool restored_;
+  bool focused_;
```
loic-sharma, line 185: Please move these above the `FML_DISALLOW_COPY_AND_ASSIGN`

shell/platform/windows/windows_lifecycle_manager.h
cbracken, line 22: Even though it's pretty self-evident, consider adding a doc comment here.

shell/platform/windows/windows_lifecycle_manager.h
```diff
@@ -34,12 +45,23 @@ class WindowsLifecycleManager {
...
+  // Intercept top level window WM_CLOSE message and listen to events that may
+  // update the application lifecycle.
   bool WindowProc(HWND hwnd, UINT msg, WPARAM w, LPARAM l, LRESULT* result);
 
   // Signal to start consuming WM_CLOSE messages.
   void BeginProcessingClose();
 
+  virtual void SetLifecycleState(AppLifecycleState state);
+
+  virtual void OnWindowStateEvent(HWND hwnd, WindowStateEvent event);
```
cbracken, line 61: Similarly add doc comments here.

shell/platform/windows/windows_lifecycle_manager_unittests.cc
```diff
@@ -0,0 +1,61 @@
...
+
+namespace flutter {
+namespace testing {
+
+class WindowsLifecycleManagerTest : public WindowsTest {};
+
+TEST_F(WindowsLifecycleManagerTest, StateTransitions) {
+  WindowsLifecycleManager manager(nullptr);
+  HWND win1 = (HWND)1;
+  HWND win2 = (HWND)2;
```
cbracken, line 18: Use `reinterpret_cast<HWND>(1)` here.

shell/platform/windows/flutter_window.cc
cbracken: Agreed. In fact, bool class fields initialise to false by default, but from a readability point of view, agreed it's better to be explicit about it to show the intent.

shell/platform/windows/windows_lifecycle_manager.cc
```diff
@@ -75,6 +75,24 @@ bool WindowsLifecycleManager::WindowProc(HWND hwnd,
     case WM_DWMCOMPOSITIONCHANGED:
       engine_->OnDwmCompositionChanged();
       break;
+
+    case WM_SIZE:
```
loic-sharma, line 79: The app can have native (non-Flutter) windows. Alternatively, the app can have two separate engines each with their own window.

Should this logic be inspecting these other windows when generating app lifecycle events?

yaakovschectman: I am not aware of any way to respond to window notifications in non-flutter windows other than by expecting the developer to make a call to our API from their window's WndProc. In what case does the app have two engines?

cbracken (commented):
Overall looks great - just a few comments.

shell/platform/windows/windows_lifecycle_manager.cc
loic-sharma: > In what case does the app have two engines?

Currently that's the only way to do a multi-window app using Flutter.

> I am not aware of any way to respond to window notifications in non-flutter windows other than by expecting the developer to make a call to our API from their window's WndProc

For the application exist lifecycle event, the app looks up all other windows and examines their state. Can we do similar tricks for hidden/shown state?

Taking a step back, what are the uses experiences we're hoping to enable with this change? It's not clear to me that sending potentially incorrect application lifecycle events is better than not sending application lifecycle events.

gspencergoog: The use case I'm thinking of are an application that wants to pause an operation when no windows are visible.  The framework also pauses frame generation when no windows are visible (see [here](https://github.com/gspencergoog/flutter/blob/link_examples/packages/flutter/lib/src/scheduler/binding.dart#L404).

gspencergoog: So, I can see an app wanting to sort of hibernate when no windows (including non-Flutter) windows are visible.

Mostly, I want the concept to be simple to understand, with as few caveats as possible: e.g. "The app will be in the 'hidden' state when zero of your application windows are visible to the user.", rather than having to qualify/complicate that concept with "just Flutter views (Additional platform windows don't count)".

yaakovschectman: So are you suggesting that we should not merge this PR until we have a way to respond to changes in non-Flutter windows' states?

gspencergoog: Yes, like we talked about in the meeting: Loïc pointed out that we don't want to be sending incorrect events, and he's right. @loic-sharma you had ideas for how to listen to the app-wide events, right?

If it's not possible to listen to the events on other windows, then the backup plan would be to provide an embedder API that allows an app to tell us when all the non-Flutter windows lose or gain focus or visibility. Probably we would want to just let them indicate how many are visible and if any are focused.

yaakovschectman: Well I'm trying to use Spy++, but when opening a log for a Window it doesn't actually log any messages, so that's not very helpful.

Anyway, from experience I've observed that the top level window receives SETFOCUS when it first appears and never receives KILLFOCUS, and that while the top level window receives SHOWWINDOW events, the Flutter window does not. Note that even the MS docs state for some paired messages that a window may receive one and not the other.

This means we cannot just process all messages for the process in the same way. I think providing an API function that a developer can call from their WndProc may be preferable anyway, because that will allow a developer to exclude certain windows from the lifecycle logic if they so choose, and to implement custom logic for focus or visibility if necessary

loic-sharma: > @loic-sharma you had ideas for how to listen to the app-wide events, right?

The idea I suggested was to listen to the [Windows message loop](https://github.com/flutter/flutter/blob/ef43463c86bf52fbb5906a4526188b04570f73a0/packages/flutter_tools/templates/app_shared/windows.tmpl/runner/main.cpp.tmpl#L35-L39). When we receive a message that indicates a window was minimized/shown, we [enumerate the app's windows](https://github.com/flutter/engine/blob/4b274b5c2c3a3cd6e9e04041fea730952605ac57/shell/platform/windows/windows_lifecycle_manager.cc#L126-L135) and check their visibility. If a window was minimized and all windows are minimized, we send the `AppLifecycleState.hidden` state. If a window was shown and it is the only visible window, we send the `AppLifecycleState.resumed` state.

> I think providing an API function that a developer can call from their WndProc may be preferable anyway, because that will allow a developer to exclude certain windows from the lifecycle logic if they so choose, and to implement custom logic for focus or visibility if necessary

This is a viable alternative too.

### Comparison

Option 1 - introduce APIs that the developer can call from their window procedure:
* ✅ Works on an app where all windows are created by Flutter
* ⚠️ Requires modifications to the runner app template
* ⚠️ Plugins that create windows - if there are any - will need to be updated
* ⚠️ Flutter will send incorrect lifecycle events if:
    * A developer uses win32 APIs to create a window but forgets to notify Flutter
    * A developer uses code they do not control to create a window. For example, the app uses a win32 API which creates a new window whose code - including the window procedure - is entirely owned by Windows
 
Option 2 - listen to the windows event loop
* ✅ Works on an app where all windows are created by Flutter
* ⚠️ Requires modifications to the runner app template
* ✅ Plugins that create windows - if there are any - will work
* ⚠️ This might be trickier to implement properly

yaakovschectman: What modifications to the template are you envisioning us needing to make for option 1? As I see it, I do not see why we would need to do so.

shell/platform/windows/windows_lifecycle_manager.cc
```diff
@@ -163,4 +181,65 @@ void WindowsLifecycleManager::BeginProcessingClose() {
...
+  }
+
+  std::lock_guard guard(state_update_lock_);
+  switch (event) {
+    case WindowStateEvent::kShow: {
+      bool first_shown_window = visible_windows_.empty();
+      auto pair = visible_windows_.insert(hwnd);
+      if (first_shown_window && pair.second &&
+          state_ == AppLifecycleState::kHidden) {
+        SetLifecycleState(AppLifecycleState::kInactive);
```
loic-sharma, line 212: Is this the right event?

yaakovschectman: I think so. In other words, if a window has just been shown, and there were previously no visible windows, and we are currently in the hidden state, enter the inactive state. Does this sound incorrect or not represented by the above code?

shell/platform/windows/client_wrapper/include/flutter/flutter_view_controller.h
```diff
@@ -57,6 +57,14 @@ class FlutterViewController {
                                                   WPARAM wparam,
                                                   LPARAM lparam);
 
+  // Non-Flutter windows added manually to an application shall pass window
+  // messages from their WndProc functions to this method call in order to
+  // implement the application lifecycle for itself.
+  void HandleExternalWindowMessage(HWND hwnd,
+                                   UINT message,
+                                   WPARAM wparam,
+                                   LPARAM lparam);
```
loic-sharma, line 66: What's the reasoning for putting this on the `FlutterViewController`? Keep in mind that Flutter can be in a "headless" mode where there's no Flutter views, however, they may be external windows.

shell/platform/windows/windows_lifecycle_manager.cc
yaakovschectman: I suspect option 2 could also introduce some other issues if/when a window handles visibility/focus differently than default, like our top level runner window. Also, there's an issue of whether we listen to messages on all windows or only top level windows. In the former case, win32 controls like text edits and buttons will count as visible windows to keep track of, and in the latter, the FlutterView itself is excluded.

shell/platform/windows/client_wrapper/include/flutter/flutter_view_controller.h
yaakovschectman: It seemed reasonable to handle this in a similar manner to `HandleTopLevelWindowProc`. I am unsure of where else we would be able to expose this API call so it can be called from an app with an external window and in turn be able to access a/the `FlutterWindowsEngine` that holds the `WindowsLifecycleManager`. Are you aware of a better sounding alternative?

loic-sharma: I'd lean towards moving this to `FlutterEngine`.

Here's what a multi-view app will look like on Windows: https://github.com/goderbauer/mvp/pull/9/files#diff-f3c022dee233416d90c619b80526ec0cb3377657c88a45b2be0f34a7cf77a061R32

You'll be able to create a `FlutterEngine` and then pass it to the `FlutterViewController`'s constructor. For now, you can access the `FlutterEngine` using `FlutterViewController::engine`.

yaakovschectman: That's reasonable. I'll make that change momentarily, thanks.

shell/platform/windows/windows_lifecycle_manager.cc
loic-sharma: Agreed. If we go with option 1 we should:

1. Document how to notify Flutter of external windows
2. Determine which Windows plugins, if any, create windows. We should work with their maintainers to ensure it does not break Flutter's app lifecycle events

Also, we should make sure the API supports this scenario where the developer cannot modify a window's procedure:
> A developer uses code they do not control to create a window. For example, the app uses a win32 API which creates a new window whose code - including the window procedure - is entirely owned by Windows

yaakovschectman: Those first two should be doable. Modifying an existing window's WndProc is probably possible (e.g. see https://stackoverflow.com/questions/63143237/change-wndproc-of-the-window). Only thing is we would need a place to store a pointer to the original WndProc to call from the replacement one. We could add a `map<HWND, WNDPROC>` somewhere, perhaps in `FlutterWindowsEngine`. Any thoughts?

shell/platform/windows/windows_lifecycle_manager.cc
loic-sharma: Is the app focused if I click on the app from the Windows taskbar? In what scenarios would an app be shown but not focused?

Since it's not readily apparent why the app receives the `Inactive` state when a window is shown, I would add a quick comment that the resumed state is sent when the view's `HWND` is focused.

yaakovschectman: Under most circumstances there may not be a scenario when a window goes from hidden/minimized to shown/restored without gaining focus, but it is always possible for a window to be explicitly shown through `ShowWindow` or restored programmatically, which does not necessarily affect focus. Also, I believe @gspencergoog mentioned at one point it is preferable to send the framework a notification of entering the inactive state in between transitioning from hidden to resumed.

gspencergoog: The framework will automatically generate a transition to inactive in between if you go from hidden to resumed, but I'd prefer that the embedding generate the transition itself in response to real events.

The idea is that all platforms have the same state machine, so application writers can depend on it.

shell/platform/windows/client_wrapper/include/flutter/flutter_view_controller.h
loic-sharma: FYI `FlutterWindowsEngine` isn't part of the Flutter Windows's public API. You'll need to expose this API using `//shell/platform/windows/client_wrapper/include/flutter/flutter_engine.h` and `//shell/platform/windows/public/flutter_windows.h`.

gspencergoog (approved):

shell/platform/windows/client_wrapper/include/flutter/flutter_view_controller.h
loic-sharma: @yaakovschectman It looks like this is still pending. Please let me know if I'm missing something!

shell/platform/windows/flutter_windows_engine_unittests.cc
```diff
@@ -895,5 +896,52 @@ TEST_F(FlutterWindowsEngineTest, EnableApplicationLifecycle) {
...
+  engine->window_proc_delegate_manager()->OnTopLevelWindowProc(
+      (HWND)1, WM_SIZE, SIZE_MINIMIZED, 0);
+  EXPECT_EQ(engine->GetLifecycleManager()->GetLifecycleState(),
+            AppLifecycleState::kHidden);
+
+  engine->window_proc_delegate_manager()->OnTopLevelWindowProc(
+      (HWND)1, WM_SIZE, SIZE_RESTORED, 0);
+  EXPECT_EQ(engine->GetLifecycleManager()->GetLifecycleState(),
+            AppLifecycleState::kInactive);
+}
```
loic-sharma, line 944: Do we have tests for external windows' messages? Also, do we have tests for a `FlutterWindow` that receives messages before its view is attached?

shell/platform/windows/client_wrapper/include/flutter/flutter_view_controller.h
yaakovschectman: Please see if the new changes are in line with what you had in mind? Thanks.

shell/platform/windows/flutter_windows_engine_unittests.cc
yaakovschectman: We do now, please check if they fulfil what you think is necessary.

shell/platform/windows/client_wrapper/flutter_engine_unittests.cc
```diff
@@ -74,6 +83,8 @@ class TestFlutterWindowsApi : public testing::StubFlutterWindowsApi {
     next_frame_callback_ = nullptr;
   }
 
+  UINT last_message() { return last_message_; }
```
loic-sharma, line 86: Consider calling this `last_external_message`. I don't feel strongly, feel free to skip

```suggestion
  UINT last_external_message() { return last_external_message_; }
```

shell/platform/windows/windows_lifecycle_manager.h
```diff
@@ -34,12 +47,33 @@ class WindowsLifecycleManager {
                     std::optional<LPARAM> lparam,
                     UINT exit_code);
 
-  // Intercept top level window messages, only paying attention to WM_CLOSE.
+  // TODO(schectman): Provide an API function for non-Flutter windows' WndProc
+  // function to call in order to contribute to the application lifecycle.
+  // https://github.com/flutter/flutter/issues/103637
```
loic-sharma, line 52: Should this be removed now?

shell/platform/windows/flutter_windows_engine.cc
```diff
@@ -796,4 +795,39 @@ void FlutterWindowsEngine::OnApplicationLifecycleEnabled() {
...
+        event = wparam == SIZE_MINIMIZED ? flutter::WindowStateEvent::kHide
+                                         : flutter::WindowStateEvent::kShow;
+        break;
+      case WM_SETFOCUS:
+        event = flutter::WindowStateEvent::kFocus;
+        break;
+      case WM_KILLFOCUS:
+        event = flutter::WindowStateEvent::kUnfocus;
+        break;
+    }
```
loic-sharma, line 825: This message translation logic is very similar to the logic in `WindowsLifecycleManager`. Can we push this translation logic down to `WindowsLifecycleManager`?

Also, why is this logic to translate external window messages to `WindowStateEvent`s different from the translation logic for Flutter's  windows messages? Shouldn't they be the same?

shell/platform/windows/flutter_windows_engine_unittests.cc
loic-sharma: Is there a test that uses `FlutterWindowsEngine::ProcessExternalWindowMessage` and then check's the lifecycle manager's resulting state? (I see `FlutterEngineTest.ProcessExternalWindowMessage`, but that doesn't seem to reach the lifecycle manager)

shell/platform/windows/flutter_windows_engine.cc
yaakovschectman: They cannot be the same, because the flutter window never expects to receive `WM_SHOWWINDOW`, and when it receives `WM_SIZE`, it does not indicate in `wparam` the change in window size as is usually the case.

shell/platform/windows/client_wrapper/include/flutter/flutter_view_controller.h
loic-sharma: Yup, looks good!

shell/platform/windows/flutter_windows_engine.cc
loic-sharma: > the flutter window never expects to receive `WM_SHOWWINDOW`, and when it receives `WM_SIZE`, it does not indicate in `wparam` the change in window size as is usually the ca

Are you referring to the messages received by `Window::HandleMessage`? I was referring to the logic for the top-level windows that are now in `WindowsLifecycleManager::WindowProc` and `WindowsLifecycleManager::ExternalWindowMessage`, I would expect these to mirror each other however there's differences. For example, on `WM_SHOWWINDOW` with a `TRUE` `wparam`:

* `WindowsLifecycleManager::WindowProc` doesn't do anything
* `WindowsLifecycleManager::ExternalWindowMessage` creates a `WindowStateEvent::kShow` event

shell/platform/windows/flutter_windows_engine_unittests.cc
loic-sharma: Looks like these were added, marking this as resolved

shell/platform/windows/flutter_windows_engine.cc
loic-sharma: Sorry, I don't fully understand the differences. Here's my understanding of the current logic:

Scenario | [Flutter's top-level window](https://github.com/yaakovschectman/flutter-engine/blob/bad630200785a8e02c9a05b7cb4e6ec04ee94d68/shell/platform/windows/windows_lifecycle_manager.cc#L41-L98) | [External top-level window](https://github.com/yaakovschectman/flutter-engine/blob/bad630200785a8e02c9a05b7cb4e6ec04ee94d68/shell/platform/windows/windows_lifecycle_manager.cc#L245-L271)
-- | -- | --
`WM_CLOSE` | If last window, requests application quit | Ignored
`WM_SIZE` | If `wparam` indicates window was maximized or restored, generates "show" window state event<br><br>If `wparam` indicates window was minimized, generates "hide" window state event | If `wparam` indicates window was minimized, generates "hide" window state event. Otherwise, generates "show" window state event
`WM_SHOWWINDOW` | If `wparam` is `FALSE`, generates "hide" window state event | If `wparam` is `FALSE`, generates "hide" window state event.<br><br>If `wparam` is `TRUE`, generates "show" window event
`WM_DESTROY` | Generates "hide" window state event | Ignored
`WM_SETFOCUS` | Ignored | Generates "focus" window state event
`WM_KILLFOCUS` | Ignored | Generates "unfocus" window state event

The differences in `WM_SETFOCUS` and `WM_KILLFOCUS` makes sense, we handle this message using the Flutter's "inner" window.

However, I'm still curious about these differences:

1. If the external top-level window receives a `WM_CLOSE` message, why don't we request application exist if it is the last window?
2. If the external top-level window receives a `WM_CLOSE` message with `wparam` of `SIZE_MAXHIDE` or `SIZE_MAXSHOW`, why does that generate a "show" window state event?
3. If the Flutter top-level window receives a `WM_SHOWWINDOW` with a `wparam` of `TRUE`, why don't we generate a "show" window state event? We do this if the external top-level window receives such a message, would it be better to be consistent?
4. Why don't we generate a "hide" window message if an external top-level receives a `WM_DESTROY` message?

If these differences are intentional, please add comments to explain.

yaakovschectman: 1.) The logic for `WM_CLOSE` was not part of this PR and was in place beforehand. I believe it is the case that it does not send the request to the framework to check if it should cancel quitting; it will just process the message as usual and close if appropriate. The app quit request worked not by explicitly quitting when allowed to do so, but by forwarding the close message to the rest of the WndProc logic, which would usually end up at the default WndProc and end up closing the program.
2.) Do you mean `WM_SIZE`? If so, the messages may be redundant, but to my understanding, they should at least not end up inaccurate.
3.) That is probably fine to generate the show event as well.
4.) Same as 3.

yaakovschectman: Does the recent commit address these concerns any?

loic-sharma: > 1.) The logic for `WM_CLOSE` was not part of this PR and was in place beforehand. I believe it is the case that it does not send the request to the framework to check if it should cancel quitting; it will just process the message as usual and close if appropriate. The app quit request worked not by explicitly quitting when allowed to do so, but by forwarding the close message to the rest of the WndProc logic, which would usually end up at the default WndProc and end up closing the program.

I talked to @gspencergoog about this scenario, and we're thinking that the Dart app should be notified of app exit if the last window is an "external" non-Flutter window. As a result, the Windows embedder will need to be able to consume an external window's `WM_CLOSE`. I'm fine with punting this work until a follow-up PR, but, we should update `FlutterEngine::ProcessExternalWindowMessage` and `FlutterDesktopEngineProcessExternalWindowMessage` to allow consuming messages (ideally they should mirror `FlutterViewController::HandleTopLevelWindowProc` and `FlutterDesktopViewControllerHandleTopLevelWindowProc`).

> 2.) Do you mean `WM_SIZE`? If so, the messages may be redundant, but to my understanding, they should at least not end up inaccurate.

Ah yes, I meant `WM_SIZE` 👍 

> Does the recent commit address these concerns any?

The new implementation looks good modulo the new external window `WM_CLOSE` comment above. Thanks for updating!

yaakovschectman: This API call now returns a `bool` indicating whether or not the handler should continue processing the message. For now, it always returns `false`, i.e. the message was not consumed, continue to process it.

loic-sharma: Should we mirror the top-level Flutter window message handler's API? The [C++ API returns a `std::optional<LRESULT>`](https://github.com/flutter/engine/blob/182e1189873bbec33d96b59f20d3b365e7aaf9ba/shell/platform/windows/client_wrapper/include/flutter/flutter_view_controller.h#L50-L58), the [C API has an `LRESULT* result` out parameter](https://github.com/flutter/engine/blob/182e1189873bbec33d96b59f20d3b365e7aaf9ba/shell/platform/windows/public/flutter_windows.h#L119-L125).

shell/platform/windows/windows_lifecycle_manager.cc
```diff
@@ -163,4 +183,105 @@ void WindowsLifecycleManager::BeginProcessingClose() {
...
+  }
+}
+
+bool WindowsLifecycleManager::ExternalWindowMessage(HWND hwnd,
+                                                    UINT message,
+                                                    WPARAM wparam,
+                                                    LPARAM lparam) {
+  std::optional<flutter::WindowStateEvent> event = std::nullopt;
+
+  switch (message) {
```
loic-sharma, line 253: Please add a TODO for the `WM_CLOSE` message here with a link to a tracking issue.

shell/platform/windows/flutter_windows_engine.cc
yaakovschectman: I think it still should return a `bool` to indicate whether the message still needs to be processed, but I can an add an `LRESULT*`.

shell/platform/windows/flutter_window.cc
```diff
@@ -74,11 +74,19 @@ FlutterWindow::FlutterWindow(int width, int height)
   current_cursor_ = ::LoadCursor(nullptr, IDC_ARROW);
 }
 
-FlutterWindow::~FlutterWindow() {}
+FlutterWindow::~FlutterWindow() {
+  OnWindowStateEvent(WindowStateEvent::kHide);
```
loic-sharma, line 78: Imagine I'm an add-to-app scenario where most of my app's content is built using win32 APIs, but one page is built using Flutter:

1. The user launches the app. A top-level window is created using win32 APIs
2. The user navigates to the "Home" page, which is built using win32 APIs
3. The user navigates to the "Settings" page, which is built using Flutter. The app creates a `FlutterViewController`, which creates a `FlutterWindowsView` and `FlutterWindow`.
4. The user navigates back to the "Home" win32 page. The app no longer needs the `FlutterViewController`, so it destroys it which also destroys the `FlutterWindowsView` and `FlutterWindow`.

The app's window is visible and focused at the end of this scenario. However, from my understanding this line will update the app's state to hidden.

Some questions:
1. Can we add a test that covers this scenario? It'd be easy to regress this behavior
2. Should we remove this line? We send the "hide" state event when the top-level Flutter window receives a `WM_DESTROY` - is that enough?

shell/platform/windows/flutter_windows_engine.cc
loic-sharma: For the C++ API, could we use the std::optional<LRESULT> approach there too? If it has a `LRESULT` value, it's the equivalent of returning `true` with the out parameter set. If it doesn't have a value, it's the equivalent of returning `false`. The benefits of the `std::optional` approach:

1. It aligns with the `FlutterViewController` API
2. It's more idiomatic for C++

shell/platform/windows/flutter_window.cc
loic-sharma: Another similar concern: an add-to-app scenario where the Flutter view's `HWND` has width & height `0` but the top-level window has non-zero size. This should not result in a hidden lifecycle event.

yaakovschectman: I do not follow your logic in the first concern. It will eventually call `WindowsLifecycleManager::OnWindowStateEvent`, which will only update the state to hidden when there are no other currently visible windows (`visible_windows_.empty()`).

Regarding the second, do you propose we should not respond to the flutter window's `WM_SIZE` messages? That could work, but the developer would need to make sure the API function is called if they want top-level window focus to effect the lifecycle state, because, again, the top-level window under normal circumstances receives `GETFOCUS` messages upon initialization with no paired `KILLFOCUS` when the flutter app loses focus, so we cannot listen for them in the top-level message handler.

loic-sharma: Sorry, I forgot that both the top-level and inner `HWND`s have separate window states.

> Regarding the second, do you propose we should not respond to the flutter window's `WM_SIZE` messages?

I meant to say that the app should not receive a `AppLifecycleState.hidden`. Sending a "hidden" window state event here seems reasonable. It seems your code covers this concern: the app won't receive a "hidden" state as the top-level window remains visible.

Would you mind adding some tests to cover these scenarios though? Perhaps:

1. There's an engine and view. Both the top-level and inner `HWND`s are visible. The app remains in the "resumed" state if the view is destroyed.
2. There's an engine and view. Both the top-level and inner `HWND`s are visible. The app remains in the "resumed" state if the view is resized to width & height `0`.

Feel free to do whichever variants you think would be most valuable :)

shell/platform/windows/client_wrapper/testing/stub_flutter_windows_api.h
```diff
@@ -6,6 +6,7 @@
 #define FLUTTER_SHELL_PLATFORM_WINDOWS_WRAPPER_TESTING_STUB_FLUTTER_WINDOWS_API_H_
 
 #include <memory>
+#include <optional>
```
loic-sharma, line 9: Can this be removed?

```suggestion
```

shell/platform/windows/flutter_windows_engine.cc
```diff
@@ -796,4 +795,27 @@ void FlutterWindowsEngine::OnApplicationLifecycleEnabled() {
...
+  lifecycle_manager_->OnWindowStateEvent(hwnd, event);
+}
+
+std::optional<LRESULT> FlutterWindowsEngine::ProcessExternalWindowMessage(
+    HWND hwnd,
+    UINT message,
+    WPARAM wparam,
+    LPARAM lparam) {
+  if (lifecycle_manager_) {
+    switch (message) {
```
loic-sharma, line 809: Would it be better to remove this `switch` and just let the lifecycle manager decide which messages it responds to? I don't feel strongly about this

shell/platform/windows/flutter_windows_engine.h
```diff
@@ -262,6 +262,22 @@ class FlutterWindowsEngine {
   // Registers the top level handler for the WM_CLOSE window message.
   void OnApplicationLifecycleEnabled();
 
+  // Called when a Window receives an event that may alter the application
+  // lifecycle state.
+  void OnWindowStateEvent(HWND hwnd, WindowStateEvent event);
+
+  WindowsLifecycleManager* GetLifecycleManager() {
```
loic-sharma, line 269: Nit: Our style is to name getter functions like variables

```suggestion
  WindowsLifecycleManager* lifecycle_manager() {
```

shell/platform/windows/windows_lifecycle_manager.cc
loic-sharma: I think everything in this thread is addressed as we decided not to modify external window's procedures. Feel free to mark this as resolved :)

shell/platform/windows/flutter_window.cc
yaakovschectman: When you say "the view is destroyed", does this mean the `FlutterWindowsView` is destructed? Because, if so, doing so stops the engine, so platform messages to the framework will cease. Also, in either case, should the app stay in resumed, or inactive, if the inner HWND was the only focused one before being hidden? I would think inactive, rather than resumed.

shell/platform/windows/windows_lifecycle_manager.h
```diff
@@ -34,12 +47,43 @@ class WindowsLifecycleManager {
                     std::optional<LPARAM> lparam,
                     UINT exit_code);
 
-  // Intercept top level window messages, only paying attention to WM_CLOSE.
+  // Intercept top level window WM_CLOSE message and listen to events that may
+  // update the application lifecycle.
   bool WindowProc(HWND hwnd, UINT msg, WPARAM w, LPARAM l, LRESULT* result);
```
loic-sharma, line 52: Perhaps consider adding a little more context in comments for outsiders unfamiliar with this codebase:

1. `WindowProc`: for top-level Flutter `HWND`s
2. `ExternalWindowMessage`: for top-level "external" `HWND`s
3. `OnWindowStateEvent`: for "inner" `HWND`s that back Flutter views

Perhaps a different name could help clarify too? Perhaps `OnTopLevelFlutterWindowMessage`, `OnTopLevelExternalWindowMessage`, and `OnWindowStateEvent`? I don't feel strongly about any of this, feel free to use different names or keep the current ones :)

shell/platform/windows/windows_lifecycle_manager.h
```diff
@@ -34,12 +47,43 @@ class WindowsLifecycleManager {
...
+  // Respond to a change in window state. Transitions as follows:
+  // When the only visible window is hidden, transition from resumed or
+  // inactive to hidden.
+  // When the only focused window is unfocused, transition from resumed to
+  // inactive.
+  // When a window is focused, transition from inactive to resumed.
+  // When a window is shown, transition from hidden to inactive.
+  virtual void OnWindowStateEvent(HWND hwnd, WindowStateEvent event);
+
+  AppLifecycleState GetLifecycleState() { return state_; }
```
loic-sharma, line 71: Nit:

```suggestion
  AppLifecycleState lifecycle_state() { return state_; }
```

Also, consider moving this to be the last public member. I don't feel strongly, feel free to skip.

shell/platform/windows/windows_lifecycle_manager.cc
loic-sharma: This seems reasonable, marking as resolved

loic-sharma (approved):
LGTM once the last few nits are addressed. Nice work! 🎉

shell/platform/windows/windows_lifecycle_manager.h
loic-sharma, line 52: Consider making this also return `std::optional<LRESULT>` for consistency. I don't feel strongly as it's not a public API, feel free to skip this if you'd like

shell/platform/windows/flutter_windows_engine.cc
yaakovschectman: That is probably a better idea if we want to also respond to `WM_CLOSE` in the future, or make similar additions.

shell/platform/windows/windows_lifecycle_manager.h
yaakovschectman: This method gets used, albeit indirectly, by a call to `RegisterTopLevelWindowProcDelegate`, with an argument of a function with a certain signature and must return `bool`. So to make this change would mean adding some extra logic to a lambda in `FlutterWindowsEngine`, so I think it may be better to leave it for now.

PR #43574, opened 2023-07-11
Release the mutex in the CanCreateConcurrentMessageLoop test before waking the latch

jason-simmons:
If the wake happens while thread_ids_mutex is still held, then the test function may exit and destroy thread_ids_mutex before the pool thread task releases it.

Fixes https://github.com/flutter/flutter/issues/130344


zanderso (approved):
Thanks!

PR #43584, opened 2023-07-12
Make GOMA state automatic by default

Hixie:
We used to default to force-enabled, which would fail on non-GOMA setups.

### DIFF
--- a/tools/gn
+++ b/tools/gn
@@ -214,6 +214,13 @@ def buildtools_dir():
 def setup_goma(args):
   goma_gn_args = {}
 
+  # args.goma has three states, True (--goma), False (--no-goma), and
+  # None (default). In True mode, we force GOMA to be used (and fail
+  # if we cannot enable it) unless the selected target definitely does
+  # not support it (in which case we print a warning). In False mode,
+  # we disable GOMA regardless. In None mode, we enable it if we can
+  # autodetect a configuration, and otherwise disable it.
+
   # When running in CI, the recipes use their own goma install, and take
   # care of starting and stopping the compiler proxy.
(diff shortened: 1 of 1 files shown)


chinmaygarde (approved):
LGTM but some CI actions may depend on the default being true and may now no longer use goma. cc @zanderso or @godofredoc may know.

zanderso:
Yeah the presub failures are timeouts. The engine v2 recipes rely on the default being `True`:

https://flutter.googlesource.com/recipes/+/refs/heads/main/recipe_modules/build_util/api.py#18

The old recipes passed it explicitly:

https://flutter.googlesource.com/recipes/+/refs/heads/main/recipes/engine/engine.py#280

Unfortunately, we can't just pass `--goma` explicitly in the v2 build configs (like here https://github.com/flutter/engine/blob/main/ci/builders/linux_android_aot_engine.json#L22) since goma has to be disabled on release builds. So switching the default to `False` will require an update to the v2 recipe in the spot I linked above, e.g. to remove `--goma` from the gn args on release builds, but not sure how @godofredoc would want to handle it.

Hixie:
I really have no idea what I'm doing editing these scripts, but FWIW my goal was to make it default to neither true nor false and have it pick the right option based on whether GOMA is configured or not. I'm guessing from what you're saying that what I actually did is make it default to False?

zanderso:
> I'm guessing from what you're saying that what I actually did is make it default to False?

Yes. I get surprised by these things sometimes, but I'm guessing that with your change it's defaulting to `None`, which gets treated as `False` when branched on.

Hixie:
Ok, let's try this more explicit solution then. :-)

Hixie:
Looks like that did it.

PR #43589, opened 2023-07-12
header file cleanup focusing on removing unnecessary SkPicture includes

flar:
Most of the #include directives for SkPicture are removed except where they are still functional. Many comments rewritten to no longer be SkPicture-centric.

- DL unit tests still use it for consistency testing
- rasterizer/engine still use it for screen shot support
- Fuchsia still uses it extensively

### DIFF
--- a/display_list/testing/dl_test_snippets.h
+++ b/display_list/testing/dl_test_snippets.h
@@ -244,18 +244,6 @@ struct DisplayListInvocation {
   DlInvoker invoker;
   bool supports_group_opacity_ = false;
 
-  bool sk_version_matches() {
-    return (static_cast<int>(op_count_) == sk_op_count_ &&
-            byte_count_ == sk_byte_count_);
-  }
-
-  // A negative sk_op_count means "do not test this op".
-  // Used mainly for these cases:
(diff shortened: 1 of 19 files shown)


flar:
Two unit tests were discovered to be completely identical to their DisplayList counterparts having been edited in a previous de-skia-fication pass. Both are deleted as redundant.

display_list/testing/dl_test_snippets.h
```diff
@@ -267,11 +255,6 @@ struct DisplayListInvocation {
   // to |DisplayList.byte_count().
   size_t byte_count() { return sizeof(DisplayList) + byte_count_; }
 
-  int sk_op_count() { return sk_op_count_; }
-  // byte count for the ops with DisplayList overhead as translated
-  // through an SkCanvas interface, comparable to |DisplayList.byte_count().
-  size_t sk_byte_count() { return sizeof(DisplayList) + sk_byte_count_; }
```
flar, line 273: These 2 fields can be deleted now, but it would cascade into a huge edit of the long list of test cases that are expressed using this structure, so I'm leaving that for a future pass.

chinmaygarde (approved):

PR #43593, opened 2023-07-12
Document (and assert) that channel names can't contains nulls

Hixie:
Fixes https://github.com/flutter/flutter/issues/116652


chinmaygarde (approved):

Hixie:
@chinmaygarde FYI i adjusted the test to be skipped when asserts are enabled, and converted some of the other tests to use `expect` instead of `assert`.

PR #43605, opened 2023-07-12
[Impeller] defer construction of VkCommandBuffer until encoding begins.

jonahwilliams:
VkCommandBuffers are allocated out of a thread local VkCommandPool. THis command buffer is only safe to use on the thread which allocates it.

We'd like to attempt concurrent encoding on Vulkan like we've done for iOS, which means the worker threads will be accessing the VkCommandBuffer. In order for this to be safe, the VkCommandBuffer must be allocated out of a VkCommandPool from that worker thread. As a simple fix, we make the Impeller CommandBuffer defer construction of the VkCommandBuffer object, allowing us to record impeller commands on the raster thread, move the command buffer to a worker thread, create the VkCommandBuffer, and then begin encoding.


impeller/renderer/backend/vulkan/command_encoder_vk.cc
```diff
@@ -97,25 +97,9 @@ class TrackedObjectsVK {
...
-  begin_info.flags = vk::CommandBufferUsageFlagBits::eOneTimeSubmit;
-  if (tracked_objects_->GetCommandBuffer().begin(begin_info) !=
-      vk::Result::eSuccess) {
-    VALIDATION_LOG << "Could not begin command buffer.";
-    return;
-  }
-  device_holder_ = device_holder;
-  queue_ = queue;
+    : fence_waiter_(std::move(fence_waiter)), context_(context) {
   is_valid_ = true;
```
jason-simmons, line 103: This flag is now always true and can be removed.

impeller/renderer/backend/vulkan/command_encoder_vk.h
```diff
@@ -68,11 +67,14 @@ class CommandEncoderVK {
...
   friend class ContextVK;
 
-  std::weak_ptr<const DeviceHolder> device_holder_;
-  std::shared_ptr<QueueVK> queue_;
+  bool Prepare() const;
+
   std::shared_ptr<FenceWaiterVK> fence_waiter_;
-  std::shared_ptr<TrackedObjectsVK> tracked_objects_;
+  std::weak_ptr<const ContextVK> context_;
+  mutable std::shared_ptr<QueueVK> queue_;
```
jason-simmons, line 74: Remove the `const` qualifier from the affected methods instead of making these members mutable.

jason-simmons (commented):
How will the `CommandEncoderVK` be passed to the worker threads when using concurrent encoding?

If there is a place where the worker receives the encoder before using it, then it might make sense to instead pass a factory object.  The worker can then have the factory construct an encoder that is initialized for use on that thread.  This will make a clearer separation between the factory (which does not have thread affinity) and the encoder (which does).

jonahwilliams:
My idea, which hopefully works, is to do a rough analog of what the mtl command buffer is doing:

https://github.com/flutter/engine/blob/main/impeller/renderer/backend/metal/command_buffer_mtl.mm#L179

That is we'll move the impeller::CommandBuffer and impeller:RenderPass to a worker thread, then invoke impeller::RenderPass::Encode which will in turn access CommandEncoderVK to perform the actual encoding.

impeller/renderer/backend/vulkan/blit_pass_vk.cc
```diff
@@ -35,7 +36,11 @@ bool BlitPassVK::EncodeCommands(
     return false;
   }
 
-  auto encoder = encoder_.lock();
+  auto command_buffer = command_buffer_.lock();
+  if (!command_buffer) {
+    return false;
+  }
+  auto encoder = command_buffer->GetEncoder();
```
jonahwilliams, line 43: I've updated RenderPass/BlitPass/And ComputePass to be non-const so that the first time command_buffer->GetEncoder() is called we construct the VkCommandEncoder.

jason-simmons (approved):

PR #43608, opened 2023-07-12
Add comment to use of 3x3 mapRect in TransformLayer

flar:
Addresses comment in https://github.com/flutter/flutter/issues/130303#issuecomment-1632895439

No tests as this is just a comment.


dnfield (approved):
Thanks for researching this!

LGTM

PR #43615, opened 2023-07-12
move rtree and canvas_spy sources to Fuchsia sub-directory

flar:
These sources have long been obsolete for most of the engine, except for the Fuchsia sources. They do not need to be built for any other platform so this PR moves them into a Fuchsia platform-specific build directory so that they are only present in the Fuchsia binaries.


cyanglaz (approved):
LGTM

PR #43620, opened 2023-07-12
[web] fix clicks on merged semantic nodes

yjbanov:
This should significantly improve the situation described in https://github.com/flutter/flutter/issues/130162.

When the framework merges semantics trees of multiple widgets into a single node, it can expand the tap area of one of the descendants to the size of the combined node as follows:

Screenshot 2023-07-07 at 4 11 30 PM

When a screen reader taps on the combined node, the pointer events and the click all land in the center. This produces a stalemate resulting in the user action never producing a tap/click gesture in the framework:

* The web engine, seeing `pointerdown`/`pointerup`, switches to the gesture mode ignores the `click`, believing that the framework will interpret the pointer events as one.
* The framework, seeing pointer events landing outside the checkbox, never reacts to the pointer events.

This PR mostly solves the issue by delaying event sequences starting with pointerdown for up to 200ms. If within those 200ms a click is observed, `SemanticsAction.tap` is sent to the framework. Otherwise, the pointer events are sent to the framework and the DOM `click` event is dropped. The tradeoff is that even when the drag gesture detector is the only one present, there's a 200ms delay before dragging can start. However, it seems to be a better trade-off than missing clicks entirely.

### DIFF
--- a/lib/web_ui/lib/src/engine/pointer_binding.dart
+++ b/lib/web_ui/lib/src/engine/pointer_binding.dart
@@ -349,7 +349,10 @@ class ClickDebouncer {
     // It's only interesting to debounce clicks when both `pointerdown` and
     // `pointerup` land on the same element.
     if (event.type == 'pointerup') {
-      final bool targetChanged = event.target != state.target;
+      // TODO(yjbanov): this is a bit mouthful, but see https://github.com/dart-lang/sdk/issues/53070
+      final DomEventTarget? eventTarget = event.target;
+      final DomElement stateTarget = state.target;
+      final bool targetChanged = eventTarget != stateTarget;
       if (targetChanged) {
         _flush();
       }
--- a/lib/web_ui/lib/src/engine/pointer_binding.dart
+++ b/lib/web_ui/lib/src/engine/pointer_binding.dart
@@ -279,7 +279,7 @@ class ClickDebouncer {
     }
 
     if (isListening) {
-      // There's a pending queue of pointer events. Prefer sendind the tap action
+      // There's a pending queue of pointer events. Prefer sending the tap action
       // instead of pointer events, because the pointer events may not land on the
       // combined semantic node and miss the click/tap.
       final DebounceState state = _state!;
@@ -415,15 +415,9 @@ class ClickDebouncer {
   ///
   /// This object can be used as if it was just initialized.
   void reset() {
-    final DebounceState? state = _state;
+    _state?.timer.cancel();
     _state = null;
     _lastFlushedPointerUpTimeStamp = null;
-
-    if (state == null) {
-      return;
-    }
-
-    state.timer.cancel();
   }
 }
 
--- a/lib/web_ui/lib/src/engine/pointer_binding.dart
+++ b/lib/web_ui/lib/src/engine/pointer_binding.dart
@@ -111,6 +111,7 @@ class PointerBinding {
   void dispose() {
     _adapter.clearListeners();
     _pointerDataConverter.clearPointerState();
+    clickDebouncer.reset();
   }
 
   final DomElement flutterViewElement;
@@ -173,9 +174,11 @@ class PointerBinding {
   }
 }
 
+@visibleForTesting
 typedef QueuedEvent = ({ DomEvent event, Duration timeStamp, List<ui.PointerData> data });
 
-typedef _DebounceState = ({
+@visibleForTesting
+typedef DebounceState = ({
   DomElement target,
   Timer timer,
   List<QueuedEvent> queue,
@@ -207,10 +210,39 @@ typedef _DebounceState = ({
 ///
 /// This mechanism is in place to deal with https://github.com/flutter/flutter/issues/130162.
 class ClickDebouncer {
-  _DebounceState? _state;
+  DebounceState? _state;
 
+  @visibleForTesting
+  DebounceState? get debugState => _state;
+
+  // The timestamp of the last "pointerup" DOM event that was flushed.
+  //
+  // Not to be confused with the time when it was flushed. The two may be far
+  // apart because the flushing can happen after a delay due to timer, or events
+  // that happen after the said "pointerup".
+  Duration? _lastFlushedPointerUpTimeStamp;
+
+  /// Returns true if the debouncer has a non-empty queue of pointer events that
+  /// were withheld from the framework.
+  ///
+  /// This value is normally false, and it flips to true when the first
+  /// pointerdown is observed that lands on a tappable semantics node, denoted
+  /// by the presence of the `flt-tappable` attribute.
   bool get isDebouncing => _state != null;
 
+  /// Processes a pointer event.
+  ///
+  /// If semantics are off, simply forwards the event to the framework.
+  ///
+  /// If currently debouncing events (see [isDebouncing]), adds the event to
+  /// the debounce queue, unless the target of the event is different from the
+  /// target that initiated the debouncing process, in which case stops
+  /// debouncing and flushes pointer events to the framework.
+  ///
+  /// If the event is a `pointerdown` and the target is `flt-tappable`, begins
+  /// debouncing events.
+  ///
+  /// In all other situations forwards the event to the framework.
   void onPointerData(DomEvent event, List<ui.PointerData> data) {
     if (!EnginePlatformDispatcher.instance.semanticsEnabled) {
       _sendToFramework(event, data);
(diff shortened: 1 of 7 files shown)


lib/web_ui/lib/src/engine/pointer_binding.dart
```diff
@@ -156,20 +159,195 @@ class PointerBinding {
...
+      _startDebouncing(event, data);
+    } else {
+      _sendToFramework(event, data);
+    }
+  }
+
+  /// Notifies the debouncer about a browser-recognized click event.
+  ///
+  /// Returns true if the click event should be deduplicated. Returns false if
+  /// the click event should not be deduplicated.
```
mdebbar, line 232: At least for me, I find the "send to framework" terminology clearer than "deduplicate", i.e. `Returns true if the click event should be sent to the framework`. Same goes for method names.

This is a personal opinion & preference. Not a strong change request.

lib/web_ui/lib/src/engine/semantics/tappable.dart
```diff
@@ -33,42 +30,44 @@ class Button extends PrimaryRoleManager {
...
 class Tappable extends RoleManager {
-  Tappable(SemanticsObject semanticsObject)
-      : super(Role.tappable, semanticsObject);
+  Tappable(SemanticsObject semanticsObject) : super(Role.tappable, semanticsObject) {
+    _clickListener = createDomEventListener((DomEvent click) {
+      final bool shouldDeduplicateClickEvent = PointerBinding.instance!.clickDebouncer.onClick(click);
+
+      if (!_isListening || shouldDeduplicateClickEvent) {
+        return;
+      }
```
mdebbar, line 39: When `!_isListening`, do we need to call `clickDebouncer.onClick(click)`?

i.e. something like this:
```suggestion
      if (!_isListening) {
        return;
      }
      final bool shouldDeduplicateClickEvent = PointerBinding.instance!.clickDebouncer.onClick(click);

      if (shouldDeduplicateClickEvent) {
        return;
      }
```

lib/web_ui/lib/src/engine/pointer_binding.dart
```diff
@@ -156,20 +159,195 @@ class PointerBinding {
...
+/// * A `click` event arrives. If the event queue has not been flushed to the
+///   framework, the event is forwarded to the framework as a
+///   `SemanticsAction.tap`, and all the pointer events are dropped. If, by the
+///   time the click event arrives, the queue was flushed (but no more than 50ms
+///   ago), then the click event is dropped instead under the assumption that
+///   the flushed pointer events are interpreted by the framework as the desired
+///   gesture.
+///
+/// This mechanism is in place to deal with https://github.com/flutter/flutter/issues/130162.
+class ClickDebouncer {
```
mdebbar, line 209: It's slightly unintuitive that we are feeding both pointer events AND clicks to the `ClickDebouncer`, but the `ClickDebouncer` only flushes pointer events while leaving the click events up to the caller to flush.

Is it possible to move `invokeOnSemanticsAction(...);` to the inside of `ClickDebouncer`?

mdebbar (commented):
The overall idea and logic look good to me. I have some suggestions/questions.

lib/web_ui/lib/src/engine/pointer_binding.dart
yjbanov: That makes sense. I'll change it.

lib/web_ui/lib/src/engine/semantics/tappable.dart
yjbanov: In either case (`isListeening` being true or false) the node should tell the debouncer to stop debouncing because the click gesture has ended. That's why `onClick` is called unconditionally. Otherwise, the debouncer will have to wait for the timer to expire before going back to normal state.

However, barring the [issue with scrollables](https://github.com/flutter/flutter/issues/130162) it should be a near-impossible situation that `isListening` is false _and_ the debouncer is debouncing events. This is because the debouncer only start debouncing if `target.hasAttribute('flt-tappable')` is true. If the node is not listening, its DOM element won't have the `flt-tappable` attribute (see changes in `Tappable.update` method).

## But!

Maybe there should be a second callback for the "not listening" case. Instead of `onClick`, maybe we should have `onNotListening`. The difference is that the debouncer would flush the events to the framework instead of dropping them.

🤔

lib/web_ui/lib/src/engine/pointer_binding.dart
yjbanov: That might be cleaner! I'll explore that. In particular, the `onClick` method won't need to return any value, and all the logic concerning sending events to the framework can be consolidated in one place.

yjbanov:
I did the clean up. `onClick` is now `void` and encapsulates all the logic related to figuring out what to forward to the framework and does the actual forwarding. PTAL.

lib/web_ui/lib/src/engine/pointer_binding.dart
```diff
@@ -198,65 +445,50 @@ class _Listener {
...
   final bool useCapture;
-  final bool isNative;
 
   void unregister() {
-    if (isNative) {
-      target.removeEventListener(event, handler, useCapture);
-    } else {
-      target.removeEventListener(event, handler, useCapture);
-    }
+    target.removeEventListener(event, handler, useCapture);
```
ditman, line 491: There's a cool property coming: `signal` when registering the event listener that would allow us to simplify this Listener class quite a bit. Instead of having to keep a reference to the passed-in handler and useCapture values, we can just keep an instance of an "AbortSignal" object, and call its `.abort()` method when it's time to stop remove the listener. (See [this example](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#add_an_abortable_listener))

(It's Safari 15 only so maybe in a couple of years... :) )

lib/web_ui/test/engine/pointer_binding_test.dart
```diff
@@ -3087,6 +3087,385 @@ void testMain() {
       packets.clear();
     },
   );
+
+  group('$ClickDebouncer', () {
```
ditman, line 3091:
```suggestion
  group('ClickDebouncer', () {
```

We don't do the `$` anymore (at least on the ecosystem :P)

lib/web_ui/test/engine/pointer_binding_test.dart
```diff
@@ -3087,6 +3087,385 @@ void testMain() {
...
+    expect(binding.clickDebouncer.isDebouncing, false);
+    expect(semanticsActions, isEmpty);
+  });
+
+  testWithSemantics('Forwards to framework when not debouncing', () async {
+    expect(EnginePlatformDispatcher.instance.semanticsEnabled, true);
+    expect(binding.clickDebouncer.isDebouncing, false);
+
+    // This test DOM element is missing the `flt-tappable` attribute on purpose
+    // so that the debouncer does not debounce events and simply let's
```
ditman, line 3162:
```suggestion
    // so that the debouncer does not debounce events and simply lets
```

lib/web_ui/test/engine/pointer_binding_test.dart
```diff
@@ -3087,6 +3087,385 @@ void testMain() {
...
+  });
+
+
+  testWithSemantics('Forwards click if enough time passed after the last flushed pointerup', () async {
+    expect(EnginePlatformDispatcher.instance.semanticsEnabled, true);
+    expect(binding.clickDebouncer.isDebouncing, false);
+
+    final DomElement testElement = createDomElement('flt-semantics');
+    testElement.setAttribute('flt-tappable', '');
+    flutterViewEmbedder.semanticsHostElement!.appendChild(testElement);
```
ditman, line 3418: From what I've seen, `flt-tappable` always goes in a `flt-semantics`? Can it be the case that we have something like:

```
flt-semantics
 ┗ some-elements...
    ┗ [flt-tappable]
```

(So the flt-tappable is not directly on the element of tag flt-semantics?)

ditman (approved):
Nothing major, this LGTM. Ship it!

lib/web_ui/lib/src/engine/pointer_binding.dart
```diff
@@ -178,6 +409,22 @@ class PointerBinding {
     }
     EnginePlatformDispatcher.instance.invokeOnPointerDataPacket(packet);
```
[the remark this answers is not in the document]
yjbanov: Yeah, one of many things for us to fix once we work on multi-view.

lib/web_ui/lib/src/engine/pointer_binding.dart
```diff
@@ -178,6 +409,22 @@ class PointerBinding {
...
+  void reset() {
+    final DebounceState? state = _state;
+    _state = null;
+    _lastFlushedPointerUpTimeStamp = null;
+
+    if (state == null) {
+      return;
+    }
+
+    state.timer.cancel();
```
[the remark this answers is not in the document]
yjbanov: Nope, `state` has been null-checked above.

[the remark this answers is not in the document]
yjbanov: Oh wait, you're suggesting just a simplification.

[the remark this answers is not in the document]
yjbanov: Simplified even further (this method used to have much more logic in previous iterations).

lib/web_ui/lib/src/engine/pointer_binding.dart
yjbanov: Interesting. Yeah, that's a much more reasonable pattern.

lib/web_ui/test/engine/pointer_binding_test.dart
yjbanov: Done.

lib/web_ui/test/engine/pointer_binding_test.dart
yjbanov: Done.

lib/web_ui/test/engine/pointer_binding_test.dart
yjbanov: No, it is guaranteed by the `Tappable` role manager that the `flt-tappable` attribute is set on the `flt-semantics` element itself. However, I'm realizing that I don't have a test for that. Adding a test.

ditman (approved):
I still think this PR makes sense after the latest changes :)

PR #43662, opened 2023-07-14
[Impeller] Fixes text scaling issues again, this time with perspective

dnfield:
Difference from last time is in f19025a.

Fixes https://github.com/flutter/flutter/issues/130476

The failure case is when we have a subpass that has a perspective matrix. Translating a perspective matrix may result in a scale change, since you're potentially translating it across the Z plane to account for perspective. This was missed before because I looked at the code and thought "ah it's just a translate so it's ok."

That commit adds a test that previously would fail.

The meat of this change is to wait to populte the glyph atlas until after we have constructed the entity list that we will actually render. To make that work, I'm capturing the entities, the pass context, and the render pass for that entity, iterating that list once to populate the glyph atlas (since these are the entities with the correct matrix), and then a second time to call render (and potentially end the inline pass context).

### DIFF
--- a/impeller/entity/contents/text_contents.cc
+++ b/impeller/entity/contents/text_contents.cc
@@ -29,18 +29,15 @@ void TextContents::SetTextFrame(const TextFrame& frame) {
   frame_ = frame;
 }
 
-void TextContents::SetGlyphAtlas(std::shared_ptr<LazyGlyphAtlas> atlas) {
-  lazy_atlas_ = std::move(atlas);
-}
-
 std::shared_ptr<GlyphAtlas> TextContents::ResolveAtlas(
     GlyphAtlas::Type type,
+    const std::shared_ptr<LazyGlyphAtlas>& lazy_atlas,
     std::shared_ptr<GlyphAtlasContext> atlas_context,
     std::shared_ptr<Context> context) const {
-  FML_DCHECK(lazy_atlas_);
-  if (lazy_atlas_) {
-    return lazy_atlas_->CreateOrGetGlyphAtlas(type, std::move(atlas_context),
-                                              std::move(context));
+  FML_DCHECK(lazy_atlas);
+  if (lazy_atlas) {
+    return lazy_atlas->CreateOrGetGlyphAtlas(type, std::move(atlas_context),
+                                             std::move(context));
   }
 
   return nullptr;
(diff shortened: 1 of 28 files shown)


dnfield:
(Added one more commit cause I realized I forgot the branch where we can't render to the onscreen pass)

bdero:
Something isn't adding up here:
If the flattened geometry is visibly stretching as the subpass offset is applied, then we're definitely applying the subpass offset incorrectly.
 If it doesn't visibly stretch, then max basis length is probably not the right heuristic to use for atlas glyph scaling.

dnfield:
> Something isn't adding up here:
> 
> 1. If the flattened geometry is visibly stretching as the subpass offset is applied, then we're definitely applying the subpass offset incorrectly and rendering wrong to begin with.
> 2. If it doesn't visibly stretch even though the basis vectors are changing, then max basis length is not the right heuristic to use for atlas glyph scaling.

I'm not sure if we're rendering wrong - the list wheel viewport stuff (cupertino date picker) looks ok to me. But when you translate something with perspective you may be affecting the scale.

We're definitely broken today but just getting lucky because the scales are close enough where it's not making a huge difference - but we're drawing text into the glyph atlas in the perspective case at one scale and then rendering it at another.

impeller/entity/contents/text_contents.cc
```diff
@@ -78,14 +75,43 @@ std::optional<Rect> TextContents::GetCoverage(const Entity& entity) const {
...
+bool TextContents::Render(const ContentContext& renderer,
+                          const Entity& entity,
+                          RenderPass& pass) const {
+  auto color = GetColor();
+  if (color.IsTransparent()) {
+    return true;
+  }
+
+  auto type = frame_.GetAtlasType();
+  auto scale = entity.DeriveTextScale();
```
bdero, line 93: Rather than using the render-time entity to derive the correct scale of the text, why not record the derived scale (or even the full matrix) to `TextContents` when it's being built by Aiks? We know the CTM in that moment. Then, we just use that value for both the rendering lookup + `PopulateGlyphAtlas` without having to worry about fuzzy lookup problems.

bdero: `TextContents` could technically get shared between entities given it's a `shared_ptr`, but in practice we don't share contents between different rendering operations today.

And if we did, we'd likely only do this for color sources, not special rendering operations that cross color source concerns with geometry concerns like with text.

dnfield:
Closing this in favor of https://github.com/flutter/engine/pull/43695

PR #43663, opened 2023-07-14
[web] always add secondary role managers

yjbanov:
Always add secondary role managers irrespective of the initial state of the semantic node, and have role manager decide whether it applies to the node or not.

Fixes https://github.com/flutter/flutter/issues/130546

### DIFF
--- a/lib/web_ui/lib/src/engine/semantics/focusable.dart
+++ b/lib/web_ui/lib/src/engine/semantics/focusable.dart
@@ -38,6 +38,20 @@ class Focusable extends RoleManager {
 
   @override
   void update() {
+    if (!_focusManager.isManaging && !semanticsObject.isFocusable) {
+      // Nothing to do about this node. It's neither focusable, nor being managed.
+      return;
+    }
+
+    if (!_focusManager.isManaging) {
+      // This line is only reachable iff the node is focusable but not being
+      // managed. So the focus manager is told to start managing it.
+      _focusManager.manage(semanticsObject.id, semanticsObject.element);
+    }
+    _updateFocus();
+  }
+
+  void _updateFocus() {
     _focusManager.changeFocus(semanticsObject.hasFocus && (!semanticsObject.hasEnabledState || semanticsObject.isEnabled));
   }
 
@@ -78,6 +92,9 @@ class AccessibilityFocusManager {
   final EngineSemanticsOwner _owner;
 
   _FocusTarget? _target;
+
+  /// Whether this focus manager is managing a focusable target.
+  bool get isManaging => _target != null;
 
   /// Starts managing the focus of the given [element].
   ///
--- a/lib/web_ui/lib/src/engine/semantics/focusable.dart
+++ b/lib/web_ui/lib/src/engine/semantics/focusable.dart
@@ -30,9 +30,7 @@ import 'semantics.dart';
 class Focusable extends RoleManager {
   Focusable(SemanticsObject semanticsObject)
       : _focusManager = AccessibilityFocusManager(semanticsObject.owner),
-        super(Role.focusable, semanticsObject) {
-    _focusManager.manage(semanticsObject.id, semanticsObject.element);
-  }
+        super(Role.focusable, semanticsObject);
 
   final AccessibilityFocusManager _focusManager;
 
@@ -137,6 +135,7 @@ class AccessibilityFocusManager {
   /// Stops managing the focus of the current element, if any.
   void stopManaging() {
     final _FocusTarget? target = _target;
+    _target = null;
 
     if (target == null) {
       /// Nothing is being managed. Just return.
@@ -145,7 +144,11 @@ class AccessibilityFocusManager {
 
     target.element.removeEventListener('focus', target.domFocusListener);
     target.element.removeEventListener('blur', target.domBlurListener);
-    _target = null;
+
+    // Blur the element after removing listeners. If this method is being called
+    // it indicates that the framework already knows that this node should not
+    // have focus, and there's no need to notify it.
+    target.element.blur();
   }
 
   void _setFocusFromDom(bool acquireFocus) {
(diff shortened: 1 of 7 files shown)


lib/web_ui/lib/src/engine/semantics/semantics.dart
```diff
@@ -509,6 +538,23 @@ abstract class PrimaryRoleManager {
...
+          _secondaryRoleManagers ??= <RoleManager>[];
+          _secondaryRoleManagers!.add(secondaryRoleManager);
+        }
+      });
+
+      // A role manager should only be added once. So remove any factories that
+      // contributed a role manager.
+      _secondaryRoleManagers?.forEach((RoleManager roleManager) {
+        secondaryRoleFactories.remove(roleManager.role);
+      });
```
ditman, line 555: Is this because this update can be called multiple times? Does every update restore the roleFactories for this role manager? Is the role manager disposable?

lib/web_ui/lib/src/engine/semantics/semantics.dart
```diff
@@ -509,6 +538,23 @@ abstract class PrimaryRoleManager {
   /// the object.
   @mustCallSuper
   void update() {
+    final Map<Role, RoleManagerFactory>? secondaryRoleFactories = _secondaryRoleFactories;
+    if (secondaryRoleFactories != null && secondaryRoleFactories.isNotEmpty) {
+      secondaryRoleFactories.forEach((Role role, RoleManagerFactory roleFactory) {
+        final RoleManager? secondaryRoleManager = roleFactory(semanticsObject);
+        if (secondaryRoleManager != null) {
```
ditman, line 545: This for loop looks a little bit weird to me. Wouldn't it be better that this knows what `availableSecondaryRoles` can be used, and the `semanticsObject` knows what `configuredSecondaryRoles` it has?

That way instead of this loop checking on every roleFactory if it can support the semanticsObject, an intersection would narrow down the actually needed Roles (which could have the roleFactory associated to the Role an immutable `const Map<Role, RoleManagerFactory>`, or whatever?)

ditman (commented):
I don't have strong opinions on this, but the determination of what secondary role factories are needed is a little bit itchy :/

lib/web_ui/lib/src/engine/semantics/semantics.dart
```diff
@@ -451,51 +462,69 @@ abstract class PrimaryRoleManager {
...
   void addFocusManagement() {
-    if (semanticsObject.isFocusable) {
-      addSecondaryRole(Focusable(semanticsObject));
-    }
+    addSecondaryRole(Role.focusable, (SemanticsObject semanticsObject) {
+      if (semanticsObject.isFocusable) {
+        return Focusable(semanticsObject);
+      }
+      return null;
+    });
```
mdebbar, line 470: Looks like these factories are always the same and have no dependency on the `PrimaryRoleManager` instance. Can we put them in a global map:
```dart
final Map<Role, RoleManagerFactory> secondaryRoleFactories = <Role, RoleManagerFactory> {
  Role.focusable: (SemanticsObject obj) => obj.isFocusable ? Focusable(obj) : null,
  ...
};
```
or combine them in a switch expression somewhere (I prefer this to make sure we exhaust all cases for `Role`):
```dart
extension RoleFactoryExtension on Role {
  RoleManager? createManager() {
    return switch (this) {
      Role.focusable => this.isFocusable ? Focusable(this) : null,
      ...
    }
  }
}
```

And the primary role manager only maintains a `List<Role>` and uses it during `update()` to decide what secondary role managers to create.

lib/web_ui/lib/src/engine/semantics/semantics.dart
mdebbar, line 555: To clarify my suggestion above, here's how I'm imagining this method to look like:
```dart
    final List<Role> secondaryRoles = _secondaryRoles;
    for (Role role in secondaryRoles) {
      final RoleManager? secondaryRoleManager = role.createManager();
      if (secondaryRoleManager != null) {
        _secondaryRoleManagers ??= <RoleManager>[];
        _secondaryRoleManagers!.add(secondaryRoleManager);
        _secondaryRoles.remove(role); // Is this allowed inside the for loop?
      }
    }
```

yjbanov:
Updated the design based on comments and discussion. We now always add secondary role managers, and have them figure out in their `update()` methods whether their functionality is needed or not.

yjbanov:
PTAL.

ditman (approved):
I think I like this better now, with the relevant logic in the subclasses!

lib/web_ui/lib/src/engine/semantics/focusable.dart
```diff
@@ -38,7 +38,14 @@ class Focusable extends RoleManager {
...
   @override
   void update() {
-    _focusManager.changeFocus(semanticsObject.hasFocus && (!semanticsObject.hasEnabledState || semanticsObject.isEnabled));
+    if (semanticsObject.isFocusable) {
+      if (!_focusManager.isManaging) {
+        _focusManager.manage(semanticsObject.id, semanticsObject.element);
+      }
+      _focusManager.changeFocus(semanticsObject.hasFocus && (!semanticsObject.hasEnabledState || semanticsObject.isEnabled));
+    } else {
+      _focusManager.stopManaging();
```
mdebbar, line 47: Do we need to `changeFocus(false)` when the node becomes unfocusable?

mdebbar (approved):
LGTM!

lib/web_ui/lib/src/engine/semantics/focusable.dart
yjbanov: Good point. I also noticed that I unconditionally start managing the node in the constructor. Fixing.

yjbanov: Done.

PR #43675, opened 2023-07-14
Use utf8.encode() instead of longer const Utf8Encoder.convert()

mkustermann:
The change in [0] has propagated now everywhere, so we can use 
`utf8.encode()` instead of the longer `const Utf8Encoder.convert()`.

Also it cleans up code like

```
  Uint8List bytes;
  bytes.buffer.asByteData();
```

as that is not guaranteed to be correct, the correct version would be

```
  Uint8List bytes;
  bytes.buffer.asByteData(bytes.offsetInBytes, bytes.length);
```

a shorter hand for that is:

```
  Uint8List bytes;
  ByteData.sublistView(bytes);
```

[0] https://github.com/dart-lang/sdk/issues/52801

### DIFF
--- a/shell/platform/fuchsia/flutter/tests/integration/mouse-input/mouse-input-view/lib/mouse-input-view.dart
+++ b/shell/platform/fuchsia/flutter/tests/integration/mouse-input/mouse-input-view/lib/mouse-input-view.dart
@@ -108,7 +108,7 @@ class MyApp {
       double wheelXPhysicalPixel,
       double wheelYPhysicalPixel}) {
     print('mouse-input-view reporting mouse input to MouseInputListener');
-    final message = utf8.encode(json.encode({
+    final message = ByteData.sublistView(utf8.encode(json.encode({
         'method': 'MouseInputListener.ReportMouseInput',
         'local_x': localX,
         'local_y': localY,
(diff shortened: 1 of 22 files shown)


shell/platform/fuchsia/flutter/tests/integration/mouse-input/mouse-input-view/lib/mouse-input-view.dart
```diff
@@ -108,20 +108,19 @@ class MyApp {
...
+        'local_y': localY,
+        'time_received': timeReceived,
+        'component_name': 'touch-input-view',
+        'buttons': buttons,
+        'phase': 'asdf',
+        'wheel_x_physical_pixel': wheelXPhysicalPixel,
+        'wheel_y_physical_pixel': wheelYPhysicalPixel,
+      }))
+      .buffer
+      .asByteData();
```
sstrickl, line 123: Make this `final message = ByteData.sublistView(utf8.encode(json.encode({...})));` like the change in `parent_view.dart`?

sstrickl (changes requested):

shell/platform/fuchsia/flutter/tests/integration/mouse-input/mouse-input-view/lib/mouse-input-view.dart
mkustermann: Thanks, missed that one.

mkustermann (commented):
PTAL

sstrickl (approved):

PR #43686, opened 2023-07-14
[Impeller] remove requests for dedicated allocations.

jonahwilliams:
This change never improved performance (in fact I think it regressed it slightly), so lets revert it and see if things change.


chinmaygarde (approved):
I'd leave limits_vk arounds just the same. We'll add to it eventually. You call though.

jonahwilliams:
Will do

PR #43688, opened 2023-07-14
More fixes for the new clang-tidy roll on iOS targets

jason-simmons:
See https://github.com/flutter/engine/pull/43661


zanderso (approved):
Thanks!

PR #43690, opened 2023-07-14
Replace deprecated [UIScreen mainScreen] in FlutterViewController.mm and FlutterViewControllerTest.mm

mossmana:
Issue: https://github.com/flutter/flutter/issues/128260

## Pre-launch Checklist

- [x] I signed the [CLA].


shell/platform/darwin/ios/framework/Source/FlutterViewController.mm
```diff
@@ -1871,7 +1876,6 @@ - (void)performOrientationUpdate:(UIInterfaceOrientationMask)new_preferences {
         if (!windowScene) {
           // When the view is not loaded, it does not make sense to access the interface
           // orientation, bail.
-          FML_LOG(WARNING) << "Trying to access the window scene before the view is loaded.";
```
hellohuanlin, line 1874: is this intended change?

hellohuanlin (approved):
looks clean

shell/platform/darwin/ios/framework/Source/FlutterViewController.mm
mossmana: The log is a bit redundant, since the windowSceneIfViewLoaded function that is called in the line above already logs the message.

mossmana: Although, I guess there could be a case where the windowScene could be null for other reasons. Maybe rather than delete the line, I should log out the comment instead?

hellohuanlin: yeah even if both are logged, one is for window scene, and the other is for screen, so not exactly duplicate.

shell/platform/darwin/ios/framework/Source/FlutterViewControllerTest.mm
```diff
@@ -174,10 +174,22 @@ - (void)tearDown {
   self.messageSent = nil;
 }
 
-- (id)setupMockMainScreenAndView:(FlutterViewController*)viewControllerMock
-                       viewFrame:(CGRect)viewFrame
-                  convertedFrame:(CGRect)convertedFrame {
-  OCMStub([viewControllerMock mainScreenIfViewLoaded]).andReturn(UIScreen.mainScreen);
+- (id)setupMockScreen {
```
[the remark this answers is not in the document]
mossmana: Ah yes, I wasn't even considering the appropriate verb vs. the noun usage when I modeled against existing functions.  I'll go ahead and make the change for this test.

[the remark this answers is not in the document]
hellohuanlin: Or `createMockScreen` may be more accurate about what the function is doing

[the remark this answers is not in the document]
mossmana: Historically, in the context of a test (not just for iOS), I've usually seen `setUp` more than `create`. They can be considered synonyms so, I think I'll go with `setUp`.

hellohuanlin:
I went through the places you early return if view is not loaded - I think they are all safe to return, since they are all called after the view is loaded, e.g. scroll callbacks. 

CC @cyanglaz to double check, especially for add-to-app scenario that I'm not familiar with

hellohuanlin:
Circling back from the triage meeting with chris and stuart - the add-to-app scenario should be fine. i think we are good to land this one.

PR #43695, opened 2023-07-14
[Impeller] Fix text scaling issues again, this time with perspective when a save layer is involved

dnfield:
Alternative to https://github.com/flutter/engine/pull/43662

Records the basis vector of the current CTM into text contents and reuses that rather than trying to get it again at render time. That method breaks if perspective is involved in the CTM and a subpass gets translated, which can modify the scale (and rotation) of the matrix.

We're definitely not doing things quite right with perspective here, but the real fix to that is to either record the fully transformed glyph into the atlas or to use SDF/path based rendering.

Fixes https://github.com/flutter/flutter/issues/130476

I still have some concerns about how `EntityPass::Render` is a mix of mutations and constness but we can try to address that independently. 

I expect this to re-improve the regression noted in https://github.com/flutter/flutter/issues/130594

### DIFF
--- a/impeller/aiks/aiks_unittests.cc
+++ b/impeller/aiks/aiks_unittests.cc
@@ -2877,6 +2877,35 @@ TEST_P(AiksTest, CanCanvasDrawPicture) {
   ASSERT_TRUE(OpenPlaygroundHere(canvas.EndRecordingAsPicture()));
 }
 
+TEST_P(AiksTest, DrawPictureWithText) {
+  Canvas subcanvas;
+  ASSERT_TRUE(RenderTextInCanvas(
+      GetContext(), subcanvas,
+      "the quick brown fox jumped over the lazy dog!.?", "Roboto-Regular.ttf"));
+  subcanvas.Translate({0, 10});
+  subcanvas.Scale(Vector2(3, 3));
+  ASSERT_TRUE(RenderTextInCanvas(
+      GetContext(), subcanvas,
+      "the quick brown fox jumped over the very big lazy dog!.?",
+      "Roboto-Regular.ttf"));
+  auto picture = subcanvas.EndRecordingAsPicture();
+
+  Canvas canvas;
+  canvas.Scale(Vector2(.2, .2));
+  canvas.Save();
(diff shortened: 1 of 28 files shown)


impeller/entity/contents/text_contents.h
```diff
@@ -45,20 +45,26 @@ class TextContents final : public Contents {
...
+      Scalar scale) override;
+
   // |Contents|
   bool Render(const ContentContext& renderer,
               const Entity& entity,
               RenderPass& pass) const override;
 
  private:
   TextFrame frame_;
+  Scalar scale_;
```
bdero, line 60: Primitive default:
```suggestion
  Scalar scale_ = 1.0;
```

impeller/aiks/aiks_unittests.cc
```diff
@@ -2877,6 +2877,35 @@ TEST_P(AiksTest, CanCanvasDrawPicture) {
   ASSERT_TRUE(OpenPlaygroundHere(canvas.EndRecordingAsPicture()));
 }
 
+TEST_P(AiksTest, DrawPictureWithText) {
```
bdero, line 2880: Sorta nit: The text seems to either be invisible or hard to see in the goldens with the golden service's white background. Maybe throw in black backgrounds with `DrawPaint`?

bdero: Actually, seems like the new goldens aren't rendering anything for some reason.

dnfield: Did the gold test part have a chance to run?

The text is red, I thought that should render on either black or white background fine :\

dnfield: But before this change, the new tests would've had validation errors (the atlas would not contain the scale and the test would crash)

dnfield:
I'm really unclear about why the goldens show nothing.

bdero (approved):
If they're only broken on Skia gold, maybe land now and add a bug for investigating later?

dnfield:
Verified the golden harness produces an image locally, filed an issue to track for later.

PR #43697, opened 2023-07-14
[web] Cleanup touch and mouse event adapters

mdebbar:
This PR cleans up a big source of complexity in our pointer handling code. It also drops support for Safari versions prior to 13 ([caniuse](https://caniuse.com/mdn-api_pointerevent)).

Right now, we throw a hard error when running in a browser that doesn't support `PointerEvent`s. Should we turn it into a soft warning, and just disable gesture/pointer handling on those browsers?

For users who need to support older versions of Safari, they can try this polyfill: https://github.com/wessberg/pointer-events (thanks @ditman!)

Fixes https://github.com/flutter/flutter/issues/116141


lib/web_ui/test/engine/semantics/semantics_helper_test.dart
```diff
@@ -122,14 +119,7 @@ void testMain() {
 
       test('Non-relevant events should be forwarded to the framework',
           () async {
-        DomEvent event;
-        if (_defaultSupportDetector.hasPointerEvents) {
-          event = createDomPointerEvent('pointermove');
-        } else if (_defaultSupportDetector.hasTouchEvents) {
-          event = createDomTouchEvent('touchcancel');
```
ditman, line 129: `createDomTouchEvent` and `createDomMouseEvent` are now probably unused in `dom.dart`?

ditman (commented):
From what I can find, this drops support for: iPhone 6 and lower (5S, etc...) (those are stuck forever in iOS 12).

The oldest model that we'd support would be: iPhone 6S (iOS 15.7), released in September, 2018.

Should we deploy an app with this change applied so we can take a look at it with a bunch of devices?

lib/web_ui/lib/src/engine/pointer_binding.dart
```diff
@@ -80,9 +80,11 @@ class SafariPointerEventWorkaround {
 }
 
 class PointerBinding {
-  PointerBinding(this.flutterViewElement, this._keyboardConverter)
-    : _pointerDataConverter = PointerDataConverter(),
-      _detector = const PointerSupportDetector() {
+  PointerBinding(
+    this.flutterViewElement,
+    this._keyboardConverter, [
+    this._detector = const PointerSupportDetector(),
```
ditman, line 86: I don't think we need this `PointerSupportDetector` anymore?

mdebbar: This was (and still is) useful for testing

mdebbar:
> From what I can find, this drops support for: iPhone 6 and lower (5S, etc...) (those are stuck forever in iOS 12).
> 
> The oldest model that we'd support would be: iPhone 6S (iOS 15.7), released in September, 2018.

Bummer 😞 maybe we should wait a little longer?

lib/web_ui/test/engine/semantics/semantics_helper_test.dart
mdebbar: `createDomTouchEvent` yes, but `createDomMouseEvent` is used in other places.

yjbanov (approved):
LGTM if LGT @ditman

ditman:
> Bummer 😞 maybe we should wait a little longer?

I'm gathering some data about usage, maybe we can extract some guideline to know *when* something can be deprecated :P

ditman (approved):
(Approving because I didn't have other than nits)

PR #43698, opened 2023-07-14
Reland "add non-rendering operation culling to DisplayListBuilder" (#41463)

flar:
Fixes https://github.com/flutter/flutter/issues/129862

This reverts commit fc9fc93415578845c1df1b0bc85daba4f14b0233.

These changes were exposing an underlying bug in the DisplayListMatrixClipTracker that has now been fixed independently (https://github.com/flutter/engine/pull/43664). There are no changes to the commit being relanded here, it has been tested on the Wondrous app which demonstrated the regression before the NOP culling feature was reverted and it now works fine due to the fix of the underlying cause.


flar:
It would be nice to test against some of the internal failures as well - as reported in b/288542323

bdero (approved):
RSLGTM

PR #43701, opened 2023-07-14
Move ViewConfiguration ownership to FlutterView

goderbauer:
This makes the FlutterView object a little bit less brittle by not depending on the PlatformDispatcher so much.

### DIFF
--- a/lib/ui/platform_dispatcher.dart
+++ b/lib/ui/platform_dispatcher.dart
@@ -1368,16 +1368,15 @@ class _PlatformConfiguration {
 /// An immutable view configuration.
 class _ViewConfiguration {
   const _ViewConfiguration({
-    required this.devicePixelRatio,
-    required this.geometry,
-    required this.visible,
-    required this.viewInsets,
-    required this.viewPadding,,
-    required this.systemGestureInsets,,
-    required this.padding,,
-    required this.gestureSettings,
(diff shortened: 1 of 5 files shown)


gspencergoog (approved):

lib/ui/platform_dispatcher.dart
```diff
@@ -66,6 +66,11 @@ typedef ErrorCallback = bool Function(Object exception, StackTrace stackTrace);
 // A gesture setting value that indicates it has not been set by the engine.
 const double _kUnsetGestureSetting = -1.0;
 
+// The view ID of PlatformDispatcher.implicitView.
+//
+// Keep this in sync with kImplicitViewId in window/platform_configuration.cc.
```
loic-sharma, line 71: While we plan to bake the "implicit view has view ID 0" assumption throughout the engine, the framework and apps should not depend on this. Given this is Dart code and an enterprising user might read this, what do you think of expanding this comment with something like:

```suggestion
// The view ID of PlatformDispatcher.implicitView. This is an
// implementation detail that may change at any time. Apps
// should always use PlatformDispatcher.implicitView to determine
// the current implicit view, if any.
//
// Keep this in sync with kImplicitViewId in window/platform_configuration.cc.
```

loic-sharma (approved):
LGTM

PR #43708, opened 2023-07-14
[Impeller] Correct coverage hint usage in PipelineBlend

bdero:
Resolves https://github.com/flutter/flutter/issues/130411.

The remaining difference in appearance is due to https://github.com/flutter/flutter/issues/127770 -- Impeller's blurs skew too large.

Before:

Screen Shot 2023-07-14 at 3 36 02 PM

After:

Screen Shot 2023-07-14 at 3 22 25 PM

### DIFF
--- a/impeller/entity/contents/filters/blend_filter_contents.cc
+++ b/impeller/entity/contents/filters/blend_filter_contents.cc
@@ -498,7 +498,22 @@ static std::optional<Entity> PipelineBlend(
   auto dst_snapshot =
       inputs[0]->GetSnapshot("PipelineBlend(Dst)", renderer, entity);
   if (!dst_snapshot.has_value()) {
-    return std::nullopt;
+    return std::nullopt;  // Nothing to render.
+  }
+
+  Rect subpass_coverage = coverage;
+  if (entity.GetContents()) {
+    auto coverage_hint = entity.GetContents()->GetCoverageHint();
+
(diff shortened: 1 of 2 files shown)


impeller/entity/contents/filters/blend_filter_contents.cc
```diff
@@ -498,7 +498,22 @@ static std::optional<Entity> PipelineBlend(
...
+    return std::nullopt;  // Nothing to render.
+  }
+
+  Rect subpass_coverage = coverage;
+  if (entity.GetContents()) {
+    auto coverage_hint = entity.GetContents()->GetCoverageHint();
+
+    if (coverage_hint.has_value()) {
+      auto maybe_subpass_coverage =
+          subpass_coverage.Intersection(*coverage_hint);
```
bdero, line 510: This was the bug: Grabbing the minimum size isn't enough -- we need to intersect the rectangles and then render the blended result to the minimal subpass texture. That also means we need to offset both the geometry drawn to the subpass as well as the returned Entity's transform to account for this.

jason-simmons:
Does this also apply to the use of `subpass_size.Min(coverage_hint->size)` in `AdvancedBlend()`?

bdero:
Yeah, thanks. Pushed a fix for the advanced blend.

chinmaygarde (approved):

jason-simmons (approved):

PR #43711, opened 2023-07-15
Android 13 VSync presentation timing

moffatman:
Implements the Android 13+ APIs for Choreographer and SurfaceTransaction needed to get and set Vsync-IDs on each frame. This is an alternative to setting presentation timestamps on surfaces to present which shouldn't have the downside of requiring some magic/manual timing tweaks. When we set the vsync-ids on each surface, it can help deal with jank arriving from uneven rendering times in the engine. [More information from Android documentation.](https://developer.android.com/games/sdk/frame-pacing#short_game_frames_lead_to_stuttering) This is mainly because we may do an `eglSwapBuffers` more than once in the same vsync-period from the compositor's perspective, and it will only show the second one. I've found it leads to a big improvement in subjective performance on 120fps devices. 

It also contains a change to `VsyncWaiterAndroid`, where a frame callback is always requested following the previous frame, instead of waiting for `AwaitVsync()` to be called. This is because I observed some cases where Choreographer would not inform us of the next frame if we registered our callback soon (~1ms) before its arrival. This resulted in frames being missed for no reason, adding jank.

## ~~Open Question - Automatic Latency Selection~~

<details>
<summary>Resolved issue -- dropped the automatic latency selection</summary>
This PR contains automatic switching from +0-frame latency (frames sent to Android to be presented 1 frame after the vsync), and +1-frame latency (frames sent to Android to be presented 2 frames after the vsync signal). This would give the benefit of the lowest latency possible when the device is fast enough to never take more than 1 frame period to compose a frame. Being aware of our own expected latency also helps the engine make better decisions (e.g. how long to expect the UI thread to be idle for). But latency switching is kind of complex and this implementation is purely reactive (e.g. padding the pipeline following a slot frame). It would be a lot cleaner just to always accept +1 frame latency. Would appreciate some feedback on the Flutter team's standards and goals here for latency.
</details>

## Note - Buffer stuffing and instrumentation

When we run with +1-frame latency, we are providing two frames in advance to the Android compositor. Depending on the device, it might have a limited number of buffers available, so even if we finish rendering to the surface at 3ms, `eglSwapBuffers` will block until 8.33 ms, when the currently visible surface is swapped out. Basically, Android is applying backpressure to the engine, which is not a bad thing, but does mean the instrumentation/timeline looks weird. I moved some of the instrumentation around so that the "raster" ends before `eglSwapBuffers`. If we don't have this change, the bars will be pinned to the frame time, in a way that doesn't say much about the actual performance.

### Possible Follow-up

- Implement for Impeller on Android
- ~~Do some further testing on iOS. We can use some different calls in Metal to present with a timestamp (hopefully we shouldn't have the same timestamp issues as in Android (https://github.com/flutter/flutter/issues/112603)). But this problem doesn't seem to be as reported on iOS. Maybe iOS already has a backpressure/sequential presentation approach?~~

#### Fixes

https://github.com/flutter/flutter/issues/108979

#### May resolve

- https://github.com/flutter/flutter/issues/129150
- https://github.com/flutter/flutter/issues/130306
- https://github.com/flutter/flutter/issues/81563
- https://github.com/flutter/flutter/issues/120205

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/common/constants.h
+++ b/common/constants.h
@@ -33,6 +33,10 @@ constexpr double kMegaByteSizeInBytes = (1 << 20);
 // implicit view if the target ID is `kFlutterImplicitViewId`, unless specified
 // otherwise.
 constexpr int64_t kFlutterImplicitViewId = 0;
+
+// If this is the vsync ID recorded for a frame, it should not be applied as
+// part of any vsync transaction.
+constexpr int64_t kInvalidVSyncId = 0;
 }  // namespace flutter
 
 #endif  // FLUTTER_COMMON_CONSTANTS_H_
--- a/flow/frame_timings.cc
+++ b/flow/frame_timings.cc
@@ -50,16 +50,11 @@ FrameTimingsRecorder::FrameTimingsRecorder(uint64_t frame_number)
 
 FrameTimingsRecorder::~FrameTimingsRecorder() = default;
 
-fml::TimePoint FrameTimingsRecorder::GetVsyncStartTime() const {
+FrameTimingsRecorder::VSyncInfo FrameTimingsRecorder::GetCurrentVSyncInfo()
+    const {
   std::scoped_lock state_lock(state_mutex_);
   FML_DCHECK(state_ >= State::kVsync);
-  return vsync_start_;
-}
-
-fml::TimePoint FrameTimingsRecorder::GetVsyncTargetTime() const {
-  std::scoped_lock state_lock(state_mutex_);
-  FML_DCHECK(state_ >= State::kVsync);
-  return vsync_target_;
+  return vsync_info_;
 }
 
 fml::TimePoint FrameTimingsRecorder::GetBuildStartTime() const {
@@ -126,9 +121,8 @@ size_t FrameTimingsRecorder::GetPictureCacheBytes() const {
   return picture_cache_bytes_;
 }
 
-void FrameTimingsRecorder::RecordVsync(fml::TimePoint vsync_start,
-                                       fml::TimePoint vsync_target) {
-  fml::Status status = RecordVsyncImpl(vsync_start, vsync_target);
+void FrameTimingsRecorder::RecordVsync(VSyncInfo vsync_info) {
+  fml::Status status = RecordVsyncImpl(vsync_info);
   FML_DCHECK(status.ok());
   (void)status;
 }
@@ -151,16 +145,14 @@ void FrameTimingsRecorder::RecordRasterStart(fml::TimePoint raster_start) {
   (void)status;
 }
 
-fml::Status FrameTimingsRecorder::RecordVsyncImpl(fml::TimePoint vsync_start,
-                                                  fml::TimePoint vsync_target) {
+fml::Status FrameTimingsRecorder::RecordVsyncImpl(VSyncInfo vsync_info) {
   std::scoped_lock state_lock(state_mutex_);
   if (state_ != State::kUninitialized) {
     return fml::Status(fml::StatusCode::kFailedPrecondition,
                        "Check failed: state_ == State::kUninitialized.");
   }
   state_ = State::kVsync;
-  vsync_start_ = vsync_start;
-  vsync_target_ = vsync_target;
+  vsync_info_ = vsync_info;
   return fml::Status();
 }
 
@@ -218,7 +210,7 @@ FrameTiming FrameTimingsRecorder::RecordRasterEnd(const RasterCache* cache) {
     layer_cache_count_ = layer_cache_bytes_ = picture_cache_count_ =
         picture_cache_bytes_ = 0;
   }
-  timing_.Set(FrameTiming::kVsyncStart, vsync_start_);
+  timing_.Set(FrameTiming::kVsyncStart, vsync_info_.start);
   timing_.Set(FrameTiming::kBuildStart, build_start_);
   timing_.Set(FrameTiming::kBuildFinish, build_end_);
   timing_.Set(FrameTiming::kRasterStart, raster_start_);
@@ -245,8 +237,7 @@ std::unique_ptr<FrameTimingsRecorder> FrameTimingsRecorder::CloneUntil(
   recorder->state_ = state;
 
   if (state >= State::kVsync) {
-    recorder->vsync_start_ = vsync_start_;
-    recorder->vsync_target_ = vsync_target_;
+    recorder->vsync_info_ = vsync_info_;
   }
 
   if (state >= State::kBuildStart) {
--- a/flow/frame_timings.h
+++ b/flow/frame_timings.h
@@ -40,6 +40,25 @@ class FrameTimingsRecorder {
     kRasterStart,
     kRasterEnd,
   };
+  struct VSyncInfo {
+    /// Timestamp of the vsync signal.
+    fml::TimePoint start;
+    /// Timestamp of the deadline for presenting the frame to the platform
+    /// compositor to avoid missing frames.
+    ///
+    /// This is often the next vsync signal timestamp, but may also be
+    /// before or after that time, depending on platform compositor scheduling.
+    fml::TimePoint target;
+    /// Timestamp of when the next vsync signal will occur.
+    ///
+    /// A vsync signal will not occur at that time unless one is requested.
+    /// The purpose of this timestamp is to properly allocate the current frame
+    /// budget when multiple frames are in flight.
+    fml::TimePoint next_start;
+    // Opaque identifier to pass to the platform compositor to help it present
+    // this frame at the correct time.
+    int64_t id;
+  };
 
   /// Default constructor, initializes the recorder with State::kUninitialized.
   FrameTimingsRecorder();
@@ -49,13 +68,7 @@ class FrameTimingsRecorder {
 
   ~FrameTimingsRecorder();
 
-  /// Timestamp of the vsync signal.
-  fml::TimePoint GetVsyncStartTime() const;
-
-  /// Timestamp of when the frame was targeted to be presented.
-  ///
-  /// This is typically the next vsync signal timestamp.
-  fml::TimePoint GetVsyncTargetTime() const;
+  VSyncInfo GetCurrentVSyncInfo() const;
 
   /// Timestamp of when the frame building started.
   fml::TimePoint GetBuildStartTime() const;
@@ -88,7 +101,7 @@ class FrameTimingsRecorder {
   size_t GetPictureCacheBytes() const;
 
   /// Records a vsync event.
-  void RecordVsync(fml::TimePoint vsync_start, fml::TimePoint vsync_target);
+  void RecordVsync(VSyncInfo vsync_info);
 
   /// Records a build start event.
   void RecordBuildStart(fml::TimePoint build_start);
@@ -131,8 +144,7 @@ class FrameTimingsRecorder {
   FML_FRIEND_TEST(FrameTimingsRecorderTest,
                   ThrowWhenRecordRasterBeforeBuildEnd);
 
-  [[nodiscard]] fml::Status RecordVsyncImpl(fml::TimePoint vsync_start,
-                                            fml::TimePoint vsync_target);
+  [[nodiscard]] fml::Status RecordVsyncImpl(VSyncInfo vsync_info);
   [[nodiscard]] fml::Status RecordBuildStartImpl(fml::TimePoint build_start);
   [[nodiscard]] fml::Status RecordBuildEndImpl(fml::TimePoint build_end);
   [[nodiscard]] fml::Status RecordRasterStartImpl(fml::TimePoint raster_start);
@@ -145,8 +157,7 @@ class FrameTimingsRecorder {
   const uint64_t frame_number_;
   const std::string frame_number_trace_arg_val_;
 
-  fml::TimePoint vsync_start_;
-  fml::TimePoint vsync_target_;
+  VSyncInfo vsync_info_;
   fml::TimePoint build_start_;
   fml::TimePoint build_end_;
   fml::TimePoint raster_start_;
--- a/shell/common/animator.cc
+++ b/shell/common/animator.cc
@@ -112,8 +112,9 @@ void Animator::BeginFrame(
   // to service potential frame.
   FML_DCHECK(producer_continuation_);
   const fml::TimePoint frame_target_time =
-      frame_timings_recorder_->GetVsyncTargetTime();
-  dart_frame_deadline_ = frame_target_time.ToEpochDelta();
+      frame_timings_recorder_->GetCurrentVSyncInfo().target;
+  dart_frame_deadline_ =
+      frame_timings_recorder_->GetCurrentVSyncInfo().next_start.ToEpochDelta();
   uint64_t frame_number = frame_timings_recorder_->GetFrameNumber();
   delegate_.OnAnimatorBeginFrame(frame_target_time, frame_number);
 }
@@ -130,7 +131,7 @@ void Animator::EndFrame() {
     frame_timings_recorder_->RecordBuildEnd(fml::TimePoint::Now());
 
     delegate_.OnAnimatorUpdateLatestFrameTargetTime(
-        frame_timings_recorder_->GetVsyncTargetTime());
+        frame_timings_recorder_->GetCurrentVSyncInfo().target);
 
     // Commit the pending continuation.
     std::vector<std::unique_ptr<LayerTreeTask>> layer_tree_task_list;
@@ -195,7 +196,10 @@ void Animator::Render(int64_t view_id,
     // to render warm up frames.
     frame_timings_recorder_ = std::make_unique<FrameTimingsRecorder>();
     const fml::TimePoint placeholder_time = fml::TimePoint::Now();
-    frame_timings_recorder_->RecordVsync(placeholder_time, placeholder_time);
+    frame_timings_recorder_->RecordVsync({.start = placeholder_time,
+                                          .target = placeholder_time,
+                                          .next_start = placeholder_time,
+                                          .id = kInvalidVSyncId});
     frame_timings_recorder_->RecordBuildStart(placeholder_time);
   }
 
--- a/shell/common/rasterizer.cc
+++ b/shell/common/rasterizer.cc
@@ -515,7 +515,7 @@ Rasterizer::DoDrawResult Rasterizer::DoDraw(
   const fml::TimePoint raster_finish_time =
       frame_timings_recorder->GetRasterEndTime();
   fml::TimePoint frame_target_time =
-      frame_timings_recorder->GetVsyncTargetTime();
+      frame_timings_recorder->GetCurrentVSyncInfo().target;
   if (raster_finish_time > frame_target_time) {
     fml::TimePoint latest_frame_target_time =
         delegate_.GetLatestFrameTargetTime();
@@ -640,11 +640,13 @@ std::unique_ptr<FrameItem> Rasterizer::DrawToSurfacesUnsafe(
   // have been resubmitted. `presentation_time` on SubmitInfo is not set
   // in this case.
   {
-    const auto vsync_target_time = frame_timings_recorder.GetVsyncTargetTime();
+    const auto vsync_target_time =
+        frame_timings_recorder.GetCurrentVSyncInfo().target;
     if (vsync_target_time > fml::TimePoint::Now()) {
       presentation_time = vsync_target_time;
     }
   }
+  int64_t vsync_id = frame_timings_recorder.GetCurrentVSyncInfo().id;
 
   frame_timings_recorder.RecordRasterStart(fml::TimePoint::Now());
 
@@ -656,7 +658,7 @@ std::unique_ptr<FrameItem> Rasterizer::DrawToSurfacesUnsafe(
     float device_pixel_ratio = task->device_pixel_ratio;
 
     DrawSurfaceStatus status = DrawToSurfaceUnsafe(
-        view_id, *layer_tree, device_pixel_ratio, presentation_time);
+        view_id, *layer_tree, device_pixel_ratio, presentation_time, vsync_id);
     FML_DCHECK(status != DrawSurfaceStatus::kDiscarded);
 
     auto& view_record = EnsureViewRecord(task->view_id);
@@ -697,7 +699,8 @@ DrawSurfaceStatus Rasterizer::DrawToSurfaceUnsafe(
     int64_t view_id,
     flutter::LayerTree& layer_tree,
     float device_pixel_ratio,
-    std::optional<fml::TimePoint> presentation_time) {
+    std::optional<fml::TimePoint> presentation_time,
+    int64_t vsync_id) {
   FML_DCHECK(surface_);
 
   DlCanvas* embedder_root_canvas = nullptr;
@@ -780,6 +783,7 @@ DrawSurfaceStatus Rasterizer::DrawToSurfaceUnsafe(
 
     SurfaceFrame::SubmitInfo submit_info;
     submit_info.presentation_time = presentation_time;
+    submit_info.vsync_id = vsync_id;
     if (damage) {
       submit_info.frame_damage = damage->GetFrameDamage();
       submit_info.buffer_damage = damage->GetBufferDamage();
--- a/shell/common/shell.cc
+++ b/shell/common/shell.cc
@@ -2064,7 +2064,11 @@ bool Shell::OnServiceProtocolRenderFrameWithRasterStats(
     // set before build-end.
     auto frame_timings_recorder = std::make_unique<FrameTimingsRecorder>();
     const auto now = fml::TimePoint::Now();
-    frame_timings_recorder->RecordVsync(now, now);
+    // deadline and next_start will not be used.
+    frame_timings_recorder->RecordVsync({.start = now,
+                                         .target = now,
+                                         .next_start = now,
+                                         .id = kInvalidVSyncId});
     frame_timings_recorder->RecordBuildStart(now);
     frame_timings_recorder->RecordBuildEnd(now);
 
--- a/shell/gpu/gpu_surface_gl_delegate.h
+++ b/shell/gpu/gpu_surface_gl_delegate.h
@@ -46,6 +46,10 @@ struct GLPresentInfo {
   // The buffer damage refers to the region that needs to be set as damaged
   // within the frame buffer.
   const std::optional<SkIRect>& buffer_damage;
+
+  // Opaque identifier to pass to the platform compositor to help it present
+  // this frame at the correct time.
+  int64_t vsync_id = kInvalidVSyncId;
 };
 
 class GPUSurfaceGLDelegate {
--- a/shell/platform/android/android_surface_gl_skia.cc
+++ b/shell/platform/android/android_surface_gl_skia.cc
@@ -11,6 +11,7 @@
 #include "flutter/fml/memory/ref_ptr.h"
 #include "flutter/shell/platform/android/android_egl_surface.h"
 #include "flutter/shell/platform/android/android_shell_holder.h"
+#include "flutter/shell/platform/android/surface/android_surface_transaction.h"
 
 namespace flutter {
 
@@ -158,7 +159,16 @@ bool AndroidSurfaceGLSkia::GLContextPresent(const GLPresentInfo& present_info) {
   if (present_info.presentation_time) {
     onscreen_surface_->SetPresentationTime(*present_info.presentation_time);
   }
-  return onscreen_surface_->SwapBuffers(present_info.frame_damage);
+  auto& transaction = AndroidSurfaceTransaction::GetInstance();
+  if (present_info.vsync_id) {
+    transaction.Begin();
+    transaction.SetVsyncId(present_info.vsync_id);
+  }
+  bool ret = onscreen_surface_->SwapBuffers(present_info.frame_damage);
+  if (present_info.vsync_id) {
+    transaction.End();
+  }
+  return ret;
 }
 
 GLFBOInfo AndroidSurfaceGLSkia::GLContextFBO(GLFrameInfo frame_info) const {
--- a/shell/platform/android/external_view_embedder/external_view_embedder.cc
+++ b/shell/platform/android/external_view_embedder/external_view_embedder.cc
@@ -14,13 +14,15 @@ AndroidExternalViewEmbedder::AndroidExternalViewEmbedder(
     const AndroidContext& android_context,
     std::shared_ptr<PlatformViewAndroidJNI> jni_facade,
     std::shared_ptr<AndroidSurfaceFactory> surface_factory,
-    const TaskRunners& task_runners)
+    const TaskRunners& task_runners,
+    AndroidSurfaceTransaction& android_surface_transaction)
     : ExternalViewEmbedder(),
       android_context_(android_context),
       jni_facade_(std::move(jni_facade)),
       surface_factory_(std::move(surface_factory)),
       surface_pool_(std::make_unique<SurfacePool>()),
-      task_runners_(task_runners) {}
+      task_runners_(task_runners),
+      android_surface_transaction_(android_surface_transaction) {}
 
 // |ExternalViewEmbedder|
 void AndroidExternalViewEmbedder::PrerollCompositeEmbeddedView(
@@ -79,6 +81,18 @@ void AndroidExternalViewEmbedder::SubmitFlutterView(
     return;
   }
 
+  auto submit_info = frame->submit_info();
+  int64_t vsync_id = submit_info.vsync_id;
+  if (vsync_id != kInvalidVSyncId) {
+    // This frame should be presented within an AndroidSurfaceTransaction.
+    // Start the transaction now, and steal the vsync_id from the SurfaceFrame,
+    // so that it doesn't start its own transaction.
+    submit_info.vsync_id = kInvalidVSyncId;
+    frame->set_submit_info(submit_info);
+    android_surface_transaction_.Begin();
+    android_surface_transaction_.SetVsyncId(vsync_id);
+  }
+
   std::unordered_map<int64_t, SkRect> view_rects;
   for (auto platform_id : composition_order_) {
     view_rects[platform_id] = GetViewRect(platform_id);
@@ -130,6 +144,10 @@ void AndroidExternalViewEmbedder::SubmitFlutterView(
     if (should_submit_current_frame) {
       frame->Submit();
     }
+  }
+
+  if (vsync_id != kInvalidVSyncId) {
+    android_surface_transaction_.End();
   }
 }
 
--- a/shell/platform/android/external_view_embedder/external_view_embedder_unittests.cc
+++ b/shell/platform/android/external_view_embedder/external_view_embedder_unittests.cc
@@ -8,6 +8,7 @@
 
 #include "flutter/shell/platform/android/external_view_embedder/external_view_embedder.h"
 
+#include "flutter/common/constants.h"
 #include "flutter/flow/embedded_views.h"
 #include "flutter/flow/surface.h"
 #include "flutter/fml/raster_thread_merger.h"
@@ -66,6 +67,15 @@ class SurfaceMock : public Surface {
               (override));
 };
 
+class AndroidSurfaceTransactionMock final : public AndroidSurfaceTransaction {
+ public:
+  AndroidSurfaceTransactionMock() {}
+  ~AndroidSurfaceTransactionMock() {}
+  MOCK_METHOD0(Begin, void());
+  MOCK_METHOD1(SetVsyncId, void(int64_t));
+  MOCK_METHOD0(End, void());
+};
+
 fml::RefPtr<fml::RasterThreadMerger> GetThreadMergerFromPlatformThread(
     fml::Thread* rasterizer_thread = nullptr) {
   // Assume the current thread is the platform thread.
@@ -109,7 +119,8 @@ TaskRunners GetTaskRunnersForFixture() {
 TEST(AndroidExternalViewEmbedder, CompositeEmbeddedView) {
   auto android_context = AndroidContext(AndroidRenderingAPI::kSoftware);
   auto embedder = std::make_unique<AndroidExternalViewEmbedder>(
-      android_context, nullptr, nullptr, GetTaskRunnersForFixture());
+      android_context, nullptr, nullptr, GetTaskRunnersForFixture(),
+      AndroidSurfaceTransaction::GetInstance());
 
   ASSERT_EQ(nullptr, embedder->CompositeEmbeddedView(0));
   embedder->PrerollCompositeEmbeddedView(
@@ -125,7 +136,8 @@ TEST(AndroidExternalViewEmbedder, CompositeEmbeddedView) {
 TEST(AndroidExternalViewEmbedder, CancelFrame) {
   auto android_context = AndroidContext(AndroidRenderingAPI::kSoftware);
   auto embedder = std::make_unique<AndroidExternalViewEmbedder>(
-      android_context, nullptr, nullptr, GetTaskRunnersForFixture());
+      android_context, nullptr, nullptr, GetTaskRunnersForFixture(),
+      AndroidSurfaceTransaction::GetInstance());
 
   embedder->PrerollCompositeEmbeddedView(
       0, std::make_unique<EmbeddedViewParams>());
@@ -138,7 +150,8 @@ TEST(AndroidExternalViewEmbedder, RasterizerRunsOnPlatformThread) {
   auto jni_mock = std::make_shared<JNIMock>();
   auto android_context = AndroidContext(AndroidRenderingAPI::kSoftware);
   auto embedder = std::make_unique<AndroidExternalViewEmbedder>(
-      android_context, jni_mock, nullptr, GetTaskRunnersForFixture());
+      android_context, jni_mock, nullptr, GetTaskRunnersForFixture(),
+      AndroidSurfaceTransaction::GetInstance());
 
   fml::Thread rasterizer_thread("rasterizer");
   auto raster_thread_merger =
@@ -173,7 +186,8 @@ TEST(AndroidExternalViewEmbedder, RasterizerRunsOnRasterizerThread) {
   auto jni_mock = std::make_shared<JNIMock>();
   auto android_context = AndroidContext(AndroidRenderingAPI::kSoftware);
   auto embedder = std::make_unique<AndroidExternalViewEmbedder>(
-      android_context, jni_mock, nullptr, GetTaskRunnersForFixture());
+      android_context, jni_mock, nullptr, GetTaskRunnersForFixture(),
+      AndroidSurfaceTransaction::GetInstance());
 
   fml::Thread rasterizer_thread("rasterizer");
   auto raster_thread_merger =
@@ -194,7 +208,8 @@ TEST(AndroidExternalViewEmbedder, PlatformViewRect) {
 
   auto android_context = AndroidContext(AndroidRenderingAPI::kSoftware);
   auto embedder = std::make_unique<AndroidExternalViewEmbedder>(
-      android_context, jni_mock, nullptr, GetTaskRunnersForFixture());
+      android_context, jni_mock, nullptr, GetTaskRunnersForFixture(),
+      AndroidSurfaceTransaction::GetInstance());
   fml::Thread rasterizer_thread("rasterizer");
   auto raster_thread_merger =
       GetThreadMergerFromPlatformThread(&rasterizer_thread);
@@ -222,7 +237,8 @@ TEST(AndroidExternalViewEmbedder, PlatformViewRectChangedParams) {
 
   auto android_context = AndroidContext(AndroidRenderingAPI::kSoftware);
   auto embedder = std::make_unique<AndroidExternalViewEmbedder>(
-      android_context, jni_mock, nullptr, GetTaskRunnersForFixture());
+      android_context, jni_mock, nullptr, GetTaskRunnersForFixture(),
+      AndroidSurfaceTransaction::GetInstance());
   fml::Thread rasterizer_thread("rasterizer");
   auto raster_thread_merger =
       GetThreadMergerFromPlatformThread(&rasterizer_thread);
@@ -261,6 +277,7 @@ TEST(AndroidExternalViewEmbedder, SubmitFlutterView) {
   auto jni_mock = std::make_shared<JNIMock>();
   auto android_context =
       std::make_shared<AndroidContext>(AndroidRenderingAPI::kSoftware);
+  AndroidSurfaceTransactionMock surface_transaction_mock;
 
   auto window = fml::MakeRefCounted<AndroidNativeWindow>(nullptr);
   auto gr_context = GrDirectContext::MakeMock(nullptr);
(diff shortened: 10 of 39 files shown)


jonahwilliams:
Just some high level thoughts:

> . But latency switching is kind of complex and this implementation is purely reactive (e.g. padding the pipeline following a slot frame). It would be a lot cleaner just to always accept +1 frame latency. Would appreciate some feedback on the Flutter team's standards and goals here for latency.

This is really hard and I don't think there is a clean answer here. In general, I have found that over time the simpler solutions tend to prevail as they are easier to understand, whereas when complicated solutions fail in hard to predict ways. To that end, I would suggest starting with 1+ frame of latency and observing the effects. For the most part, flutter is already a few frames behind anyway.


> I moved some of the instrumentation around so that the "raster" ends before eglSwapBuffers. If we don't have this change, the bars will be pinned to the frame time, in a way that doesn't say much about the actual performance.

I suspect you're right that communicating that we are blocking on eglswapbuffers isn't very informative for most of the raster workload. That said it is good to know when there is back pressure. I'd recommend leaving changes to tracing out of this PR however, just because this will already be a tricky one to land and it does need to be any harder.

> Implement for Impeller on Android


I'm curious what the API usage looks like for Vulkan as that will be the renderer we focus on with Impeller. That said, we're a ways off from shipping that so I wouldn't worry about it in this PR.

> Do some further testing on iOS. We can use some different calls in Metal to present with a timestamp (hopefully we shouldn't have the same timestamp issues as in Android (https://github.com/flutter/flutter/issues/112603)). But this problem doesn't seem to be as reported on iOS. Maybe iOS already has a backpressure/sequential presentation approach?

on iOS we present with a core animation transaction and that seems to have solved most of the observable issues. https://developer.apple.com/documentation/quartzcore/cametallayer/1478157-presentswithtransaction?language=objc . The docs basically say not to do what we're doing, but we do it anyway and it seems to work great 🤷

moffatman:
Thx, I pulled out the variable latency. I never noticed any problem with 1-frame fixed latency, just wondered if it could be avoided on fast devices. But I think you're right to skip it, it was causing some thread safety problem which I couldn't figure out.

On iOS, I experimented with `presentAtTime:` on the `Drawable`, but I think it has the same problem with timestamps as on Android. And no real problem is observed on iOS anyway.

Still remaining: Figure out how to properly account for `NotifyIdle` when pipeline is depth 2. As each of the 2 frames in flight will overestimate available time on the UI thread since they are unaware of the other.

shell/common/animator.cc
```diff
@@ -99,7 +99,12 @@ void Animator::BeginFrame(
   FML_DCHECK(producer_continuation_);
   const fml::TimePoint frame_target_time =
       frame_timings_recorder_->GetVsyncTargetTime();
-  dart_frame_deadline_ = frame_target_time.ToEpochDelta();
+  // TODO(moffatman): kind of a hack to deal with pipeline depth 2
+  // It should be based on the latency of the frame
+  // Maybe frame recorder should know the latency?
+  dart_frame_deadline_ =
+      frame_target_time.ToEpochDelta() -
+      ((frame_target_time - frame_timings_recorder_->GetVsyncStartTime()) / 2);
```
jonahwilliams, line 107: Can you explain this ? Why does the dart target frame time need to change?

shell/common/rasterizer.cc
```diff
@@ -622,6 +623,10 @@ RasterStatus Rasterizer::DrawToSurfaceUnsafe(
 
     frame->set_submit_info(submit_info);
 
+    frame_timings_recorder.RecordRasterEnd(
```
jonahwilliams, line 626: This actually might be an issue. It looks like above we're recording the presentation time - but this is actually happening way before any of the engine workload is executed. So we have no idea if we're going to hit or miss it.

shell/gpu/gpu_surface_gl_delegate.h
```diff
@@ -48,6 +48,9 @@ struct GLPresentInfo {
   // The buffer damage refers to the region that needs to be set as damaged
   // within the frame buffer.
   const std::optional<SkIRect>& buffer_damage;
+
+  // For android pipeline TODO(moffatman): comment
+  int64_t vsync_id;
```
jonahwilliams, line 53: Please document, here and elsewhere, the properties of the vsync_id and/or link to documentation.

Is it monotonically increasing? What happens if we drop a frame, et cetera.

shell/platform/android/android_surfacetransaction.cc
```diff
@@ -0,0 +1,62 @@
...
+    ASurfaceTransaction_apply(pending_transaction_);
+    ASurfaceTransaction_delete(pending_transaction_);
+  }
+
+  if (ASurfaceTransaction_create) {
+    pending_transaction_ = ASurfaceTransaction_create();
+  }
+}
+
+}  // namespace flutter
```
jonahwilliams, line 62:
```suggestion
}  // namespace flutter

```

shell/platform/android/android_surfacetransaction.h
```diff
@@ -0,0 +1,49 @@
...
+  ASurfaceTransaction_apply_FPN ASurfaceTransaction_apply;
+  ASurfaceTransaction_delete_FPN ASurfaceTransaction_delete;
+  ASurfaceTransaction* pending_transaction_;
+
+  FML_DISALLOW_COPY_AND_ASSIGN(AndroidSurfaceTransaction);
+};
+
+}  // namespace flutter
+
+#endif  // FLUTTER_SHELL_PLATFORM_ANDROID_ANDROID_SURFACETRANSACTION_H_
```
jonahwilliams, line 49:
```suggestion
#endif  // FLUTTER_SHELL_PLATFORM_ANDROID_ANDROID_SURFACETRANSACTION_H_

```

shell/platform/android/android_surfacetransaction.h
```diff
@@ -0,0 +1,49 @@
...
+
+#include "flutter/fml/macros.h"
+
+#include <cstdint>
+
+namespace flutter {
+
+//------------------------------------------------------------------------------
+/// The Android SurfaceTransaction API is used by `AndroidSurface` to
+/// synchronize frame presentation times. It's only available on API 29+.
```
jonahwilliams, line 16: How is this used to sychronize frame presentation times?

shell/common/animator.cc
moffatman: Now frame_target_time will be t+2frames instead of t+1 frames. I was seeing NotifyIdle called by one frame and eating too much time out of the other frame (when depth=2) instead of splitting the work up. This fix isn't very good.... it should divide by frame pipeline depth instead of by 2.

shell/gpu/gpu_surface_gl_delegate.h
moffatman: Sure. In general FYI, I believe it is just monotonically increasing. But we are supposed to, and we can, just treat it as a magic token and never modify it or inspect it.

shell/platform/android/android_surfacetransaction.h
moffatman: It will turn all the changes applied during the transaction into one atomic update, that we can tag with the Vsync-ID.

shell/common/rasterizer.cc
moffatman: I guess you're right, it won't count any of the GPU time, right, just the preroll/paint in CPU? So either it can be reverted altogether, in which case the timings will include the swapBuffers call which might be backpressured, or I can try to find some way to move it deeper into the engine to call right before the swap?

jonahwilliams: Not even GPU time, at this point we haven't even encoded the frame. I don't think this API could take GPU time into account, as we will "present" after encoding is finished but not necessarily after execution is complete

shell/platform/android/android_surfacetransaction.h
jonahwilliams: which changes?

moffatman: Any SurfaceControl which is updated in that time. It works the same way as "[CATransaction begin]" / "[CATransaction commit]".

shell/common/rasterizer.cc
moffatman: I think how this worked before, is that the `ScopedFrame` destructor is what ended the raster time. And that wasn't destructed until after the frame was submitted and buffers swapped?

I don't know that much about the GPU terminology/model. What does encoded mean in this context?

With SurfaceTransaction there are some new APIs to get callbacks when the transaction is either committed and/or presented. Is it useful?

jonahwilliams:
I don't think the adjustments to raster time are working with Impeller enabled:

image

jonahwilliams:
On the Skia backend too, it looks like youve dropped almost all of the engine workload from being considered part of the raster frame time.

jonahwilliams:
Since the reported raster time is just a diagnostic aid, I would prefer if you left it as is for this PR.

moffatman:
The vsync-id and surfaceTransaction are not wired into Impeller, I can try it out, I guess just for GL backend.
Will revert the raster time change.

jonahwilliams:
I don't expect them to be wired into impeller, but adding them can't break impeller either.

shell/common/animator.cc
```diff
@@ -99,7 +99,10 @@ void Animator::BeginFrame(
   FML_DCHECK(producer_continuation_);
   const fml::TimePoint frame_target_time =
       frame_timings_recorder_->GetVsyncTargetTime();
-  dart_frame_deadline_ = frame_target_time.ToEpochDelta();
+  dart_frame_deadline_ =
+      frame_target_time.ToEpochDelta() -
+      ((frame_target_time - frame_timings_recorder_->GetVsyncStartTime()) /
+       layer_tree_pipeline_->inFlightCount());
```
jonahwilliams, line 105: Please comment the intention of this calcuation

jonahwilliams:
I've run through this a bit locally and everything seems to be working as expected. That said, I don't really notice a difference on my high frame rate phone.

How can we verify that this is doing what we expect?

shell/gpu/gpu_surface_gl_delegate.h
dnfield: (initialize this to something)
```suggestion
  int64_t vsync_id = 0;
```

shell/platform/android/vsync_waiter_android.cc
```diff
@@ -66,8 +90,35 @@ void VsyncWaiterAndroid::OnVsyncFromNDK(int64_t frame_nanos, void* data) {
...
+    void* data) {
+  auto frame_time = AndroidChoreographer::GetFrameTime(callback_data);
+  auto now = fml::TimePoint::Now();
+  if (frame_time > now) {
+    frame_time = now;
+  }
+  auto* weak_this = reinterpret_cast<std::weak_ptr<VsyncWaiterAndroid>*>(data);
+  auto shared_this = weak_this->lock();
+  if (!shared_this) {
+    delete weak_this;
```
dnfield, line 109: It'd be nice if we could avoid this in the NDK based code paths by making `ConsumePendingCallback` not do manual deletion and instead just do it in the JDK based implementation, but that can be done separately from this PR.

ci/licenses_golden/licenses_flutter
```diff
@@ -5006,6 +5008,8 @@ FILE: ../../../flutter/shell/platform/android/android_surface_software.cc
 FILE: ../../../flutter/shell/platform/android/android_surface_software.h
 FILE: ../../../flutter/shell/platform/android/android_surface_vulkan_impeller.cc
 FILE: ../../../flutter/shell/platform/android/android_surface_vulkan_impeller.h
+FILE: ../../../flutter/shell/platform/android/android_surfacetransaction.cc
```
dnfield, line 5011: nit on filenames:

not android_surface_transaction.cc/h?

dnfield:
From https://developer.android.com/ndk/reference/group/native-activity#asurfacetransaction_oncommit:

> THREADING The transaction committed callback can be invoked on any thread.

But from `apply`:

> Applies the updates accumulated in transaction.

> Note that the transaction is guaranteed to be applied atomically. The transactions which are applied on the same thread are also guaranteed to be applied in order.

Will this cause problems for us on Vulkan?

jonahwilliams:
I'm not sure what that means exactly. I'd think that a single swapchain image would constitute a single transaction?

moffatman:
FPS count is not necessarily helpful. Because we may be delivering enough frames to the display, but if paced unevenly, they don't get displayed well. Really the only objective way to check is slo-mo camera.

I reproduced it in this dummy app by adding an artificial delay every 5 frames which will push the combined build+raster time over 16.66ms. The real impact is on complex apps where build+raster time is hovering naturally near the frame time, especially on 120fps devices.

Before fix: https://www.youtube.com/watch?v=cz0snWqT6xs

After fix: https://www.youtube.com/watch?v=d6KVuDsN4G0

<details>
<summary>Demo code</summary>

```dart
import 'dart:math';
import 'dart:ui' as ui;

void main() {
  runApp(const MyApp());
}

class MyApp extends StatefulWidget {
  const MyApp({Key? key}) : super(key: key);

  @override
  createState() => MyAppState21();
}

class _SimplePainter extends CustomPainter {
  final double value;
  const _SimplePainter(this.value);

  //Color color(int b) => HSVColor.fromAHSV(1, (31 * b) % 360, 0.3, 1).toColor();
  Color color(int b) => b.isEven ? Colors.black : Colors.white;
  
  @override
  void paint(ui.Canvas canvas, ui.Size size) {
    canvas.drawRect(Offset.zero & size, Paint()..color = color(value.floor()));
    canvas.drawRect(Offset.zero & Size(size.width * (value % 1), size.height), Paint()..color = color(value.ceil()));
  }
  
  @override
  bool shouldRepaint(_SimplePainter oldDelegate) => value != oldDelegate.value;
}


class MyAppState21 extends State<MyApp> with TickerProviderStateMixin {
  static const _kSeconds = 10000;
  late final AnimationController controller;
  int frameNumber = 0;
  @override
  void initState() {
    super.initState();
    controller = AnimationController(vsync: this, duration: const Duration(seconds: _kSeconds));
  }
  
  @override
    return MaterialApp(
      home: Scaffold(
        body: AnimatedBuilder(
          animation: controller,
          builder: (context, _) {
            if (frameNumber % 5 == 0) {
              final start = DateTime.now();
              while (DateTime.now().difference(start) < const Duration(milliseconds: 5 + 8)) {

              }
            }
            return Stack(
              alignment: Alignment.center,
              children: [
                Column(
                  children: List.generate(10, (i) => Flexible(
                    fit: FlexFit.tight,
                    child: CustomPaint(
                      willChange: true,
                      painter: _SimplePainter(controller.value * (_kSeconds / pow(2, i))),
                      child: const SizedBox.expand()
                    )
                  ))
                ),
                Column(
                  children: [
                    Flexible(
                      child: Center(
                        child: Text('f=${frameNumber++}', style: TextStyle(
                          color: Colors.blue,
                          fontSize: MediaQuery.sizeOf(context).shortestSide / 5,
                          fontFeatures: const [
                            ui.FontFeature.tabularFigures()
                          ]
                        ))
                      )
                    ),
                    Flexible(
                      child: Center(
                        child: Text('t=${(DateTime.now().millisecond % 1000).toString().padLeft(3, '0')}', style: TextStyle(
                          color: Colors.red,
                          fontSize: MediaQuery.sizeOf(context).shortestSide / 5,
                          fontFeatures: const [
                            ui.FontFeature.tabularFigures()
                          ]
                        ))
                      )
                    )
                  ]
                )
              ]
            );
          }
        ),
        floatingActionButton: FloatingActionButton(
          onPressed: () {
            if (controller.isAnimating) {
              controller.stop();
            }
            else {
              controller.repeat();
            }
          }
        ),
      )
    );
  }

  @override
  void dispose() {
    super.dispose();
    controller.dispose();
  }
}
```

</details>

For the transaction question, it just means if we create and apply multiple transactions, they will go in-order. But here, we only do one transaction per frame, so it doesn't come into play.

shell/platform/android/android_surface_gl_skia.cc
```diff
@@ -155,10 +156,15 @@ void AndroidSurfaceGLSkia::GLContextSetDamageRegion(
 bool AndroidSurfaceGLSkia::GLContextPresent(const GLPresentInfo& present_info) {
   FML_DCHECK(IsValid());
   FML_DCHECK(onscreen_surface_);
+  TRACE_EVENT1("flutter", "AndroidSurfaceGLSkia::GLContextPresent", "vsync_id",
+               std::to_string(present_info.vsync_id).c_str());
```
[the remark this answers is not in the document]
moffatman: Sure, I was using it in development to match the presented frame to the vsync callback. But now it's working well, I can remove it.

jonahwilliams:
Platform views would cause there to be multiple transactions, right?

moffatman:
Any specific demo that would help me check that? I don't know much about platform view mechanics. I tried adding a print in my `AndroidSurfaceTransaction::PresentWithVsyncId` and it's only called once per frame when running with WebView plugin.

jonahwilliams:
It depends on how the specific plugin is rendered. i think you'd need a hybrid composition plugin like google maps (see flutter/packages).

That said, I'm still interested in how we can test that this works correctly and continues to work correctly.

moffatman:
From my reading so far, the only concern might be for Overlay views, they are displayed after the "background canvas" which contains other platform views + flutter content frame gets submitted, which is where the transaction stuff is. Any other surface won't have the proper submit_info set, so it will have vsync_id 0, and no transaction created/displayed. But since the main/only transaction is already applied, the overlay views would be part of the transaction for the following frame. So possibly presented 1 frame behind the rest of the app. There are some APIs to read which surfaces are part of a transaction in a onComplete callback. So I could have a test with overlay views and make sure they are included in the same transaction as the reset of the frame. But need to start with some app/plugin with overlay views to try this out. Anything you think I am missing? Just reading some of how this embedder works for the first time.

Also, tried google maps, but seemed to have some problem, it was not showing content and giving google api errors in the console. Any other hybrid composition plugin?

jonahwilliams:
I would ask on the ecosystem channel in the discord if there are any test apps that would be easy to use.

shell/common/animator.cc
moffatman: This way is still incorrect in most cases. I may need to add another timestamp returned from vsync callback or maybe some new method on the `VsyncWaiter` itself to differentiate when `frame_target_time` will be after the next vsync firing time, so the full time diff is not available for use.

moffatman: I have updated this to a new approach, the Vsync callback will now contain both `vsync_target_time` (wall clock budget) and `next_vsync_start_time` (single thread budget)

moffatman:
@jonahwilliams I've moved some of the transaction handling to the `ExternalViewEmbedder` so that we can be sure all surfaces updated in one frame are grouped in the same transaction.

jonahwilliams:
As an FYI from our end - we don't currently have any devices in the flutter devicelab that support the APIs used in this PR - but we will in the next few weeks.

chinmaygarde:
Can we revisit this now that we have received the extra devices? Or perhaps they are still being wired up.

jonahwilliams:
They are not wired up yet.

chinmaygarde:
Ping. The devices have been wired up. Can we get this over the finish line now?

jonahwilliams:
I've rebuilt this at ToT and tried the flutter gallery on a Pixel 7 Pro, using the Impeller Vulkan/GLES and Skia backends. I have to say, I can't see any differences in the performance of the rendering.

I'm OK taking this patch and throwing it at the devicelab to see what aggregate metrics show. But if there is no change in these metrics I'd want to revert it. If we don't have a slam dunk case to show this is improving performance or a way to verify that we're using this API correctly, it feels like taking on technical debt that may not pay off.

moffatman:
This patch is only for Skia/GL. I think I tried to put it in the impeller surfaces as well, but the way the swap callback is encapsulated made it non trivial. Although if you have Impeller + Platform view(s), it will be applied due to the changes in the ExternalViewEmbedder.

I don't know if it will change any of the existing metrics, which only measure throughput of the flutter pipeline, right? It is to fix an issue where we deliver all needed frames to the compositor, but only a subset of those frames actually get posted to the display. We need a demo app where total frame time hovers around 1/fps. Then we could check that the number of frames in an `adb screenrecord` matches the number of frames flutter built and painted.

jonahwilliams:
Its fine if its not wired up for Impeller. But yeah, We need _something_ we can measure, otherwise the likelyhood that this keeps working longterm is effectively 0

chinmaygarde:
Are we working on figuring out how to test this? cc @reidbaker or @johnmccutchan for guidance.

reidbaker:
> Are we working on figuring out how to test this? cc @reidbaker or @johnmccutchan for guidance. 

I don't have any offhand. If you want to dig into this together grab some time on my calendar.

moffatman:
My idea so far is to implement the `MediaProjection` API to self-screen record and get the real frame pacing from that. This would be as part of the Scenarios test app I guess. I haven't gotten far enough to see how feasible that is to use. E.G. How/when is permission auth required.

ds84182:
Screen recording to estimate dropped frames is going to be excessively flaky. dumpsys gfxinfo may give you the aggregate timing information you need.

moffatman:
@ds84182 Have you got `dumpsys gfxinfo` to work with Flutter apps? It only shows about 1-5 frames rendered total for the app even when that number should be in the hundreds. But yeah I agree based on the data shown there, it would cover the testing needs.

I tried the `MediaProjection` approach, but something about doing the self-screen recording seems to also have solved the lag too. So not useful for showing the difference. It is having some weird effect, SurfaceFlinger is showing rendering 240 fps when I do the recording.

ds84182:
I remember being able to get per-surface frametime stats from some part of dumpsys. Maybe `dumpsys surfaceflinger` has it?

ds84182:
It does have it, dunno the best way to get the name of the surface but `dumpsys SurfaceFlinger --latency [surface name here]` works for me. Not exactly sure what the columns mean, but they're definitely timestamps

ds84182:
`dumpsys SurfaceFlinger --list` contains the surface names, latency command outputs desired present time, actual present time, and frame ready time.

moffatman:
Opened a PR for a new metric that will show improvement from this change

While it's not a complete representation of the benefit of setting vsync-ids on each presented frame, it shows the improvement from some of my rewrite of the vsync waiter.

https://github.com/flutter/flutter/pull/135279

Using `scroll_perf_bad.dart` on 120fps device

| Metric | Before PR | After PR |
|--|--|--|
| average_frame_request_pending_lag | 7.33 ms | 5.45 ms
| 90th_percentile_frame_request_pending_lag | 9.12 ms | 8.39 ms
| 99th_percentile_frame_request_pending_lag | 42.73 ms | 10.34 ms

Although realized upon this, the `Xhz_frame_percentage` measurements are now broken with my split of current-frame-budget vs next-frame-start-time for better planning of pipeline-depth-2 in this PR.

shell/platform/android/android_choreographer.cc
```diff
@@ -58,4 +130,40 @@ void AndroidChoreographer::PostFrameCallback(OnFrameCallback callback,
...
 
+void AndroidChoreographer::PostFrameCallback33(OnFrameCallback33 callback,
+                                               void* data) {
+  AChoreographer* choreographer = AChoreographer_getInstance();
+  AChoreographer_postVsyncCallback(choreographer, callback, data);
+}
+
+int64_t AndroidChoreographer::GetFrameTimelineIndex(
+    AChoreographerFrameCallbackData* data,
+    size_t latency) {
```
johnmccutchan, line 141: what are the units of latency and why is it a size_t?

flow/frame_timings.h
```diff
@@ -51,11 +51,24 @@ class FrameTimingsRecorder {
   /// Timestamp of the vsync signal.
   fml::TimePoint GetVsyncStartTime() const;
 
-  /// Timestamp of when the frame was targeted to be presented.
+  /// Timestamp of the deadline for presenting the frame to the platform
+  /// compositor to avoid missing frames.
   ///
-  /// This is typically the next vsync signal timestamp.
+  /// This is often the next vsync signal timestamp, but may also be
+  /// before or after that time, depending on platform compositor scheduling.
```
johnmccutchan, line 58: Should this be named VsyncDeadline?

shell/platform/android/external_view_embedder/external_view_embedder.cc
```diff
@@ -73,6 +75,16 @@ void AndroidExternalViewEmbedder::SubmitFrame(
     return;
   }
 
+  auto submit_info = frame->submit_info();
+  int64_t vsync_id = submit_info.vsync_id;
+  if (vsync_id) {
+    // Steal the vsync_id so that the frame submission doesn't use it
```
johnmccutchan, line 81: Why don't we want the frame submission to use it?

shell/platform/android/external_view_embedder/external_view_embedder.cc
```diff
@@ -73,6 +75,16 @@ void AndroidExternalViewEmbedder::SubmitFrame(
     return;
   }
 
+  auto submit_info = frame->submit_info();
+  int64_t vsync_id = submit_info.vsync_id;
+  if (vsync_id) {
```
johnmccutchan, line 80: An integer isn't a boolean: replace this with vsync_id != 0

shell/platform/android/external_view_embedder/external_view_embedder.cc
```diff
@@ -180,6 +192,10 @@ void AndroidExternalViewEmbedder::SubmitFrame(
       frame->Submit();
     }
   }
+
+  if (vsync_id) {
```
johnmccutchan, line 196: same here. Probably better to have a named constant like: INVALID_VSYNC_ID

shell/platform/android/surface/android_surface_transaction.h
```diff
@@ -0,0 +1,55 @@
...
+
+namespace flutter {
+
+//------------------------------------------------------------------------------
+/// The Android SurfaceTransaction API is used by `AndroidSurface` to
+/// synchronize frame presentation times. It's only available on API 29+.
+///
+class AndroidSurfaceTransaction {
+ public:
+  static AndroidSurfaceTransaction& GetInstance() {
```
johnmccutchan, line 20: Move this out of the header file and into the source file.

shell/platform/android/surface/android_surface_transaction.cc
```diff
@@ -0,0 +1,76 @@
...
+  FML_DCHECK(!pending_transaction_);
+  if (ASurfaceTransaction_create) {
+    pending_transaction_ = ASurfaceTransaction_create();
+  }
+}
+
+void AndroidSurfaceTransaction::SetVsyncId(int64_t vsync_id) {
+  if (ASurfaceTransaction_create) {
+    FML_DCHECK(pending_transaction_);
+    if (vsync_id != 0) {
```
johnmccutchan, line 61: we should have a named constant for this invalid vsync id

shell/platform/android/vsync_waiter_android.cc
```diff
@@ -66,8 +90,41 @@ void VsyncWaiterAndroid::OnVsyncFromNDK(int64_t frame_nanos, void* data) {
...
+  ConsumePendingCallback(weak_this, frame_time, target_time, target_time, 0);
+}
+
+// static
+void VsyncWaiterAndroid::OnVsyncFromNDK33(
+    AndroidChoreographer::AChoreographerFrameCallbackData* callback_data,
+    void* data) {
+  auto frame_time = AndroidChoreographer::GetFrameTime(callback_data);
+  auto now = fml::TimePoint::Now();
+  if (frame_time > now) {
```
johnmccutchan, line 103: Please add a comment explaining why we are modifying the frame time here.

shell/platform/android/vsync_waiter_android.cc
```diff
@@ -66,8 +90,41 @@ void VsyncWaiterAndroid::OnVsyncFromNDK(int64_t frame_nanos, void* data) {
...
+  auto shared_this = weak_this->lock();
+  if (!shared_this) {
+    delete weak_this;
+    return;
+  }
+  // Target to present with +1 frame latency to avoid jitter from
+  // uneven frame pacing.
+  auto target_time = AndroidChoreographer::GetFrameDeadline(callback_data, 1);
+  auto next_frame_time = frame_time + fml::TimeDelta::FromNanoseconds(
+                                          1000000000.0 / g_refresh_rate_);
```
johnmccutchan, line 116: Please cache the nanoseconds value to avoid doing the divide every frame.

shell/platform/embedder/vsync_waiter_embedder.cc
```diff
@@ -42,7 +42,8 @@ bool VsyncWaiterEmbedder::OnEmbedderVsync(
         auto vsync_waiter = weak_waiter->lock();
         delete weak_waiter;
         if (vsync_waiter) {
-          vsync_waiter->FireCallback(frame_start_time, frame_target_time);
+          vsync_waiter->FireCallback(frame_start_time, frame_target_time,
```
johnmccutchan, line 45: Is there ever a case where we aren't passing frame_target_time twice?

johnmccutchan (commented):
First round: Lots of nits and requests for comments.

moffatman:
As an update here, was waiting to get the above metric in place, and it's now collected in the skia perf website after merging https://github.com/flutter/flutter/pull/135890.

[Example](https://flutter-flutter-perf.skia.org/e/?queries=device_type%3DPixel_7_Pro%26sub_result%3D90th_percentile_frame_request_pending_latency%26sub_result%3Daverage_frame_request_pending_latency%26sub_result%3D99th_percentile_frame_request_pending_latency&xbaroffset=36927)

So will be addressing the above feedback as well as bringing this branch up to date in the next few days

Razenpok:
Is there an option to directly integrate [Android Frame Pacing Library](https://developer.android.com/games/sdk/frame-pacing)? It looks like it's solving this exact issue, and it also has Vulkan support.

moffatman:
I looked at the Android frame pacing library but it replaces a lot of stuff flutter already does, but I did get some tips from there how to use the APIs. For status here, best estimate is in the next week to rebase this and move it, just some other stuff to finish first.

Razenpok:
AFAIK Frame Pacing Library works on any Android version (not just 13+). Is it doing something else there, and will it be possible to port the changes for earlier Android versions?

moffatman:
AFAIK it uses the approach which failed here https://github.com/flutter/flutter/issues/112603

shell/platform/android/vsync_waiter_android.cc
moffatman: I don't know why this is done, I just copied it from the previous code. And to be honest, it reminds me of some confusion all over the code base, where different timespecs are mixed. So it may be worth looking at all those and cleaning it up.

shell/platform/embedder/vsync_waiter_embedder.cc
moffatman: In Android, I am setting up 2-frame-depth-pipeline now. So the next_frame_start_time is before frame_target_time.

shell/platform/android/surface/android_surface_transaction.cc
moffatman: Sure, where do you think I should declare it? Not confident about any namespacing/importing rules in this repo, and this would be used in:

- `flow/surface_frame.h`
- `shell/gpu/gpu_surface_gl_delegate.h`
- `shell/platform/android/external_view_embedder/external_view_embedder.cc`
- `shell/platform/android/surface/android_surface_transaction.cc`

shell/platform/android/android_choreographer.cc
moffatman: Number of frames

shell/common/shell.cc
```diff
@@ -1950,7 +1950,7 @@ bool Shell::OnServiceProtocolRenderFrameWithRasterStats(
     // set before build-end.
     auto frame_timings_recorder = std::make_unique<FrameTimingsRecorder>();
     const auto now = fml::TimePoint::Now();
-    frame_timings_recorder->RecordVsync(now, now);
+    frame_timings_recorder->RecordVsync(now, now, now, 0);
```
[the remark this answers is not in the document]
moffatman: They are dummy values afaik

shell/platform/android/external_view_embedder/external_view_embedder.cc
moffatman: We need to wrap all updates within `AndroidSurfaceTransaction::Begin()` and `AndroidSurfaceTransaction::End()`. If vsync_id is passed to the frame, `AndroidSurfaceGLSkia` will do that wrapping. We need that not to happen so our transaction will also contain the updates to platform views.

shell/platform/android/external_view_embedder/external_view_embedder.cc
moffatman: Done

flow/frame_timings.cc
```diff
@@ -126,8 +139,11 @@ void FrameTimingsRecorder::RecordRasterStart(fml::TimePoint raster_start) {
   (void)status;
 }
 
-fml::Status FrameTimingsRecorder::RecordVsyncImpl(fml::TimePoint vsync_start,
-                                                  fml::TimePoint vsync_target) {
+fml::Status FrameTimingsRecorder::RecordVsyncImpl(
+    fml::TimePoint vsync_start,
```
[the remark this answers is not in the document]
moffatman: Done

shell/platform/android/surface/android_surface_transaction.h
moffatman: Done

shell/platform/android/vsync_waiter_android.cc
moffatman: Done

johnmccutchan:
So my big concern with this PR is the number of known unknowns (e.g. review questions with the answer of being "I'm not sure, I copied it from elsewhere") and lack of clear story around testing.

What can be done about these issues so we can feel confident merging?

shell/platform/android/vsync_waiter_android.cc
moffatman: Looked into it more and put a comment to explain why, sometimes Android may give you a timestamp which is a tiny bit in the future, and the rest of the engine can't handle that.

shell/platform/android/surface/android_surface_transaction.cc
moffatman: Added to `common/constants.h`

shell/platform/android/external_view_embedder/external_view_embedder.cc
moffatman: Done

flow/frame_timings.h
moffatman: I decided not to rename it as this turns into the timeline attribute "TargetTime" referred to in `flutter_driver`.

shell/common/shell.cc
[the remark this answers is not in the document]
moffatman: They're ignored, only `.start` matters which must be <= `.buildStart`. I've put a comment.

moffatman:
@johnmccutchan 
I looked into the unknowns in your review comments now, let me know if there's any more. 

I added a check that `AndroidSurfaceTransaction` is used properly by the `AndroidExternalViewEmbedder`. 

This PR is about pipelining better, optimizing throughput over latency, it will help the most on complex apps / low-end devices where build+frame time hovers near 1 frame period. The metric introduced above will show an improvement when merged. I could also create a demo app which has intentionally unstable build times, it will show a clear improvement 90 -> 120 FPS after merging this PR?

johnmccutchan:
> @johnmccutchan I looked into the unknowns in your review comments now, let me know if there's any more.
> 
> I added a check that `AndroidSurfaceTransaction` is used properly by the `AndroidExternalViewEmbedder`.
> 
> This PR is about pipelining better, optimizing throughput over latency, it will help the most on complex apps / low-end devices where build+frame time hovers near 1 frame period. The metric introduced above will show an improvement when merged. I could also create a demo app which has intentionally unstable build times, it will show a clear improvement 90 -> 120 FPS after merging this PR?

Thanks for the extra context. I'm still not sure how we know that this change doesn't cause regressions in existing apps?

moffatman:
Not sure whether you are asking about bug regression or performance regression, only thing to do for performance regression is to monitor the metrics / skia-perf. Although sadly the highest level metrics like cpu_usage, gpu_usage can't be captured on Android. 

Personally I have been shipping with this change in a few apps for the last 6 months, so I have good feedback across different devices / Android releases. But will certainly monitor devicelab stats after merge and we can revert it if there is problems.

flow/frame_timings.h
johnmccutchan: Why not fix the name here in C/C++ but preserve the timeline attribute? 

Or better, fix both and update flutter_driver (but could be done in a separate CL).

shell/platform/android/vsync_waiter_android.cc
```diff
@@ -7,67 +7,136 @@
...
 #include "flutter/fml/size.h"
 #include "flutter/fml/trace_event.h"
-#include "flutter/shell/platform/android/android_choreographer.h"
 
 namespace flutter {
 
 static fml::jni::ScopedJavaGlobalRef<jclass>* g_vsync_waiter_class = nullptr;
 static jmethodID g_async_wait_for_vsync_method_ = nullptr;
-static std::atomic_uint g_refresh_rate_ = 60;
+static std::atomic_uint g_refresh_period_ = 1000000000.0 / 60;
```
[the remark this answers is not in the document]
johnmccutchan: Why is the type a uint but the constant a double? You're storing the floor of (1000 / 60) either way.

johnmccutchan (commented):
More comments, mostly nits.

shell/platform/android/surface/android_surface_transaction.h
```diff
@@ -0,0 +1,40 @@
...
+
+//------------------------------------------------------------------------------
+/// The Android SurfaceTransaction API is used by `AndroidSurface` to
+/// synchronize frame presentation times. It's only available on API 29+.
+///
+class AndroidSurfaceTransaction {
+ public:
+  static AndroidSurfaceTransaction& GetInstance();
+
+  virtual void Begin();
```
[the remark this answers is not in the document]
moffatman: It's to allow mocking for tests

shell/platform/android/external_view_embedder/external_view_embedder.cc
```diff
@@ -74,6 +76,16 @@ void AndroidExternalViewEmbedder::SubmitFlutterView(
     return;
   }
 
+  auto submit_info = frame->submit_info();
+  int64_t vsync_id = submit_info.vsync_id;
+  if (vsync_id != kInvalidVsyncId) {
+    // Steal the vsync_id so that the frame submission doesn't use it
+    submit_info.vsync_id = 0;
```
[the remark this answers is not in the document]
moffatman: Done

common/constants.h
```diff
@@ -33,6 +33,10 @@ constexpr double kMegaByteSizeInBytes = (1 << 20);
 // implicit view if the target ID is `kFlutterImplicitViewId`, unless specified
 // otherwise.
 constexpr int64_t kFlutterImplicitViewId = 0;
+
+// If this is the vsync ID recorded for a frame, it should not be applied as
+// part of any vsync transaction.
+constexpr int64_t kInvalidVsyncId = 0;
```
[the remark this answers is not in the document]
moffatman: Done

shell/platform/android/external_view_embedder/external_view_embedder_unittests.cc
```diff
@@ -443,20 +469,34 @@ TEST(AndroidExternalViewEmbedder, SubmitFlutterView) {
                 FlutterViewDisplayOverlaySurface(0, 150, 150, 100, 100));
 
     auto did_submit_frame = false;
+    int64_t submitted_vsync_id = 0;
```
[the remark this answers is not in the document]
moffatman: Switched to kInvalidVSyncId

flow/frame_timings.h
```diff
@@ -40,6 +40,12 @@ class FrameTimingsRecorder {
     kRasterStart,
     kRasterEnd,
   };
+  struct VsyncTimings {
```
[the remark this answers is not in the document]
moffatman: Done, I changed all the `Vsync` -> `VSync` that you noted here, but FYI there are even more uses all over the code. After my renames there are 604 "Vsync"s and 298 "VSync"s.

chinmaygarde:
Looks like this PR is good for another review?

delfme:
@moffatman do you think this could help with this issue as well? https://github.com/flutter/flutter/issues/139447
On android in general we are hit by janks and jerking UI (ie. small listView with bunch of texts and small images). Things get very bad when running on mid-level phones (both 60 and 90hz).

delfme:
Since gathering some metrics about the improvements before merging was a point, could you run this sample code with your custom engine? https://github.com/flutter/flutter/issues/139447#issue-2022523716

johnmccutchan:
@moffatman can you refresh this PR so it doesn't have merge conflicts?

shell/platform/android/surface/android_surface_transaction.h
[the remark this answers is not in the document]
johnmccutchan: My other comment still applies though- please factor out a pure interface class and then make create an implementation class.

shell/platform/android/external_view_embedder/external_view_embedder.cc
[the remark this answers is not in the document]
johnmccutchan: I don't see the code updated here?

shell/platform/android/external_view_embedder/external_view_embedder.cc
johnmccutchan: Can you please add a comment here to explain that?

shell/platform/android/external_view_embedder/external_view_embedder.cc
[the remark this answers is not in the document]
moffatman: Whoops updated now, also I updated it all places the initial value of vsync_id in the structs to kInvalidVSyncId

shell/platform/android/surface/android_surface_transaction.h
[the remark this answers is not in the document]
moffatman: Sure, I updated it. I think I did it right, let me know if I broke the convention, I really don't use C++

shell/platform/android/external_view_embedder/external_view_embedder.cc
moffatman: Added a comment

moffatman:
@johnmccutchan  updated

johnmccutchan:
I tested this PR and didn't notice any breakage but also didn't notice any performance improvement.

Is there a good sample app I can run to observe the benefits?

delfme:
@johnmccutchan maybe this one? Couldnt test it https://github.com/flutter/flutter/issues/139447

We are experiencing janks when scrolling on android mid-entry devices. Even in our smoothest android scenario, there are always micro-janks. Same code on iOS is smooth.

If not already available, it would be great to see a sample from flutter team of a Sliver List where elements have some images and maybe different size (ie. social media feed). In this specific case, janks tend to be a real problem.

moffatman:
@johnmccutchan 

Here's a demo animation which intentionally wastes time to have unstable build time (while still remaining under budget), so that build+raster time for each frame toggles between 1 and 2 frame periods. After a few seconds of running, it settles into ~80 FPS. With my PR, it gets 120 FPS.

Maybe you need to tweak the values a little if you don't have the exact same performance timings and 120Hz display as I do (MI MIX 4).

```dart
import 'dart:async';
import 'dart:math' as math;
import 'dart:ui' as ui;

import 'package:flutter/scheduler.dart';

final _rng = math.Random();

int _lastFrameCount = 0;
int frameCount = 0;

void main() async {
  WidgetsFlutterBinding.ensureInitialized();
  SchedulerBinding.instance.addPersistentFrameCallback((timeStamp) {
    frameCount++;
  });
  Timer.periodic(const Duration(seconds: 1), (_) async {
    print('FPS: ${frameCount - _lastFrameCount}');
    _lastFrameCount = frameCount;
  });
  runApp(const MyApp());
}

class MyApp extends StatefulWidget {
  const MyApp({Key? key}) : super(key: key);

  @override
  createState() => MyAppState();
}

class MyAppState extends State<MyApp> with SingleTickerProviderStateMixin {
  late final AnimationController animation;

  @override
  void initState() {
    super.initState();
    animation = AnimationController(vsync: this, duration: const Duration(seconds: 1));
    animation.repeat();
  }

  @override
  void dispose() {
    super.dispose();
    animation.dispose();
  }

  @override
    return MaterialApp(
      showPerformanceOverlay: true,
      home: Scaffold(
        appBar: AppBar(),
        body: Center(
          child: Transform.scale(
            scale: 3,
            child: Stack(
              alignment: Alignment.center,
              children: [
                const CircularProgressIndicator(),
                AnimatedBuilder(
                  animation: animation,
                  builder: (context, _) {
                    final ms = _rng.nextInt(6);
                    final delay = Duration(milliseconds: ms);
                    final time = DateTime.now();
                    while (DateTime.now().difference(time) < delay) {
                      // Waste up to 5 ms out of 8.33 ms period
                    }
                    return Text(ms.toString(), style: const TextStyle(fontFeatures: [ui.FontFeature.tabularFigures()]));
                  }
                )
              ]
            )
          )
        )
      )
    );
  }
}
```

moffatman:
Bumped the build again. I am not completely confident how to expose and use some of the more complex types now that the NDK usage is in its own file. As we now `#include <android/surface_control.h>` and have the proper types when compiling on Android, my lazy `typedef void` usage no longer flies.

moffatman:
@johnmccutchan Any guidance from you here??

I just rebase it and now there is yet another rewrite of NDK handling...

johnmccutchan:
Hi,

I wanted to give an update on this PR from my point of view:

- I've patched this PR in and could not observe any difference in a number of sample applications I ran.
- AFAIK, there are no benchmarks demonstrating a clear performance win
- This PR touches a very sensitive part of the engine and does not come with extensive tests of this functionality.

Given that none of the above changes, I'm not keen on landing this PR.

Somewhat related, the Android team gave us feedback on our Vulkan swapchain implementation which resulted in a reimplementation. After the dust settles on the Vulkan swap chain changes, my recommendation for this PR would be: 

* Demonstrate in a repeatable/measurable way that this PR improves performance/functionality/something.
* Split this PR up into smaller PRs that are easy to review, have clear motivation and impact, and include extensive tests.

moffatman:
It's surprising there is no difference with my example app above, since it really clearly drops frames without my PR. Maybe it has to do with the phone, if it has built in backpressure it might not be seen. Or if the phone is never hitting 1-frame latency anyway and always has at least 2-frame. What device did you try it on?

I think it's a pretty useful change and disappointing it can't move forward easily. As I got some good feedback after starting to ship my app with it. Also I have added the `frame_request_pending_latency` metric which could show the improvement after landing in CI metrics.

If the Android/Impeller team decides to go forward with `SurfaceTransaction` for presentation, my additional tweaks here could be applied there too. (So far this patch is just for Skia). Is that the direction the team is going in (as opposed to Vulkan built-in swapping calls)? Then this PR could be split into the following

1. Start using choreographer33 APIs for VsyncWaiterAndroid
  - I haven't seen any performance improvement from this but it is a prerequisite for later things (vsyncId)
2.  Use SurfaceTransaction for presentation in Skia
3. Wire up vsyncId from VsyncWaiter -> Presentation
4. The [looping] tweak to VsyncWaiterAndroid from this PR
  - Basically, request the next frame callback immediately after receiving the last one, instead of in `AwaitVsync`. As I noticed Android may not give you the event if you await too soon. Also, that request needs the main thread which could be busy when Await is called, further delaying.
5. Intentionally opt into 2-frame pipeline (1-frame worse latency but better throughput)

chinmaygarde:
Triage: Spoke to @johnmccutchan and I think it makes sense to split this into multiple patches. I am reworking the swapchains in Impeller to work with transactions too and I'd need to page in the context here.

gmackall:
From android triage: What are the next steps for this PR? @moffatman do you plan to split into multiple patches, as @chinmaygarde recommended? Is there more action needed from reviewers?

moffatman:
I was waiting for all this pending work in impeller/vulkan to finish. @chinmaygarde what is the state of swapchain? i haven't followed this development.

chinmaygarde:
We had to temporarily back out the hardware buffer backed swapchains. I am working on fixing sync issues discovered in that implementation.

reidbaker:
From triage: Moved this pr to Draft to indicate the incomplete/blocked state of this work.

goderbauer:
(triage): @moffatman Just checking in, is this still on your radar - and is there a path forward?

moffatman:
I guess impeller/vk swapchain stuff is done now? @chinmaygarde If so I can give this a fresh look

reidbaker:
Fyi the engine has been migrated to flutter/flutter and prs will need to be moved to that repo. 

If this pr is reopened please link here so reviewers can keep their context.

PR #43727, opened 2023-07-15
[Impeller] fix some OpenGL Linux desktop issues.

jonahwilliams:
Fixes https://github.com/flutter/flutter/issues/130514

The context wasn't current so we couldn't get the GL version.

The extension lookup was throwing an invalid enum error, the desktop opengl docs say to use a different method.

### DIFF
--- a/impeller/renderer/backend/gles/description_gles.cc
+++ b/impeller/renderer/backend/gles/description_gles.cc
@@ -90,10 +90,10 @@ DescriptionGLES::DescriptionGLES(const ProcTableGLES& gl)
       extensions_.insert(extension);
     }
   } else {
-    size_t extension_count = 0u;
+    int extension_count = 0;
     gl.GetIntegerv(GL_NUM_EXTENSIONS, &extension_count);
-    for (auto i = 0u; i < extension_count; i++) {
-      extensions_.insert(GetGLStringi(GL_EXTENSIONS, i));
+    for (auto i = 0; i < extension_count; i++) {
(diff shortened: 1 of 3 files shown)


chinmaygarde (approved):

ardera:
Does this really work? This will call `gl_context_make_current` on the platform thread, AFAICT without calling `gl_context_clear_current` after, and the raster thread will also attempt to call `gl_context_make_current`.

I also came across this issue (that `FlutterEngineRunInitialized` needs a GL context) and I had to use a different context because the raster thread was also calling `gl_context_make_current`.  (While `FlutterEngineRunInitialized` was still running on the platform thread) See https://github.com/ardera/flutter-pi/commit/369a818db87365b2bdb5389b180a0c62f026f6da

I haven't tested yet with flutter-pi, so idk maybe it works fine. But if it doesn't work with flutter-pi, I'm wondering why it works with the desktop embedder.

jonahwilliams:
I mean maybe, but it really is just based off of a mirror of what we do on the Skia embedder: GPUSurfaceGLSkia

But in general the impeller backend doesn't work yet on desktop OpenGL.

ardera:
Okay I'll try, maybe it works differently than I expected

With the fixes flutter-pi actually ran with impeller, though it wasn't that smooth and it crashed after 1-2 minutes

jonahwilliams:
The OpenGL backend hasn't really seen a lot of focus, so I'd definitely expect poor performance. Through crashing shouldn't happen in general.  Though If you find particular crashes or scenes that are unusually slow please file issues on them.

PR #43747, opened 2023-07-17
Fix drawVertices documentation

brianosman:
Update the blend mode section of the documentation to specify the correct blend modes.

Fixes https://github.com/flutter/flutter/issues/130747


flar (approved):
LGTM - just to be sure, did we at last get the dst vs src behaviors correct?

brianosman:
Yup, the rest of the prose is correct (both in identifying which is the src vs. dst, and which blend mode selects each color).

flar:
The Skia gold changes must be flakes. Try rebasing and it should succeed second time...

PR #43748, opened 2023-07-17
[Impeller] Move glyph atlas state from context into lazy glyph atlas.

jonahwilliams:
No impact on behavior. Rather that separately cache the lazy glyph atlas and the underlying state (bitmap, atlas, texture), move the state into the lazy glyph atlas. To do make this behave correctly, we need to "reset" the lazy glyph atlas instead of deleting it on each frame. Reseting it clears out the text runs and cached atlas, which has the same impact as deleting it in the previous behavior.

This refactor is required in order to move to platform specific glyph atlases, to solve issues such as https://github.com/flutter/flutter/issues/126104


jonahwilliams:
Friendly ping @dnfield

chinmaygarde (approved):

dnfield (approved):

PR #43759, opened 2023-07-18
[Impeller] Avoid inserting additional save layers based on clip configuration.

jonahwilliams:
Fixes https://github.com/flutter/flutter/issues/130775

On the Skia backend, antiAliasWithSaveLayer is the highest fidelity clipping option. In the Impeller backend, there isn't any difference in how we clip, since the stencil buffer is always used. Nevertheless we were still inserting the save layer, which results in an extra offscreen texture and is wasteful.

Track if impeller is enabled in the diff/preroll/paint context and avoid inserting a save layer.

### DIFF
--- a/flow/compositor_context.cc
+++ b/flow/compositor_context.cc
@@ -13,13 +13,14 @@ namespace flutter {
 
 std::optional<SkRect> FrameDamage::ComputeClipRect(
     flutter::LayerTree& layer_tree,
-    bool has_raster_cache) {
+    bool has_raster_cache,
+    bool impeller_enabled) {
   if (layer_tree.root_layer()) {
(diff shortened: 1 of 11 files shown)


dnfield (approved):
LGTM.

nit: consider having some kind of enum rather than bools. https://google.github.io/styleguide/cppguide.html#Function_Argument_Comments

jonahwilliams:
I think we should keep it impelled named to facilitate an easier cleanup when we eventually delete all of this code.

flar:
This leads to a rendering problem - the saveLayer does more than just consolidate the AA edges: https://github.com/flutter/flutter/issues/134705

PR #43783, opened 2023-07-18
Apply the offset to the child bounds of an ImageFilterLayer with no filter

jason-simmons:
Fixes https://github.com/flutter/flutter/issues/130318


flar (approved):
LGTM

flar:
I don't think this fixes all of the issue since there is still the case that we null the filter if either sigma is 0 rather than if both are 0.

PR #43787, opened 2023-07-18
fix handling of clipped rendering inside a layer that applies a filter

flar:
When a saveLayer is rendered with an ImageFilter that modifies the bounds of the rendered pixels, and some of the content of that saveLayer did not intersect the clip, but the filtered output of that content did intersect the clip, we might not accumulate the bounds of those rendering operations into the DisplayList bounds.

This bug was not encountered during practical testing, but showed up on some testing with the new NOP culling code.

For now, this bug only affects the accuracy of the reported bounds of the DisplayList, but that can affect raster caching and potentially the layer culling done in the LayerTree. It might also affect the accuracy of the RTree associated with the DisplayList, which would only show up in rare circumstances when the indicated operation was used in the presence of an embedded view.


ColdPaleLight (approved):
LGTM

flar:
I'm at a loss for why there are crashes on every platform in CI testing, but they don't happen when I run the tests locally.

jason-simmons:
Some of the errors on CI are happening in a run that uses the address sanitizer.

Try reproducing it locally by building the host engine with the `--asan`/`--lsan` GN flags (see an example failing LUCI run at https://ci.chromium.org/ui/p/flutter/builders/try/Linux%20Engine%20Drone/1101808/overview)

jason-simmons:
The error is happening because `TestParameters` is holding the `sk_renderer_` and `dl_renderer_` by reference.  But in `SaveLayerClippedContentStillFilters` the renderer lambdas are temporaries that go out of scope after the `TestParameters` is constructed.

The renderer lambdas should be moved into variables within the test method's scope.

flar:
> The error is happening because `TestParameters` is holding the `sk_renderer_` and `dl_renderer_` by reference. But in `SaveLayerClippedContentStillFilters` the renderer lambdas are temporaries that go out of scope after the `TestParameters` is constructed.
> 
> The renderer lambdas should be moved into variables within the test method's scope.

Good catch and bad design on my part in creating these test mechanisms. Thanks for investigating!

flar:
> The renderer lambdas should be moved into variables within the test method's scope.

A simpler solution was just to change the fields to non-references. The reference is fine during the constructor, but the long-term storage should not be by reference. If there is copying there as a result, that's fine as none of that code is performance-sensitive. Ignoring the fact that it is part of a test, that constructor is only executed a couple dozen times.

PR #43790, opened 2023-07-18
Remove obsolete legacy engine builds from staging

zanderso:
These builds have had engine v2 variants running for nearly a quarter or more in most cases, and they are now not serving much of a purpose. Additionally, during MTV peak hours, queue times and therefore overall cycle times are getting too long, and we should be freeing up these resources.

### DIFF
--- a/.ci.yaml
+++ b/.ci.yaml
@@ -82,26 +82,6 @@ platform_properties:
       os: Windows-10
 
 targets:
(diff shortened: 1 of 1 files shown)


bdero (approved):
LGTM

keyonghan (approved):
LGTM

PR #43791, opened 2023-07-18
Fix the rules for determining whether a blur image filter is valid

jason-simmons:
See https://github.com/flutter/flutter/issues/130318


display_list/effects/dl_image_filter.h
```diff
@@ -229,11 +229,13 @@ class DlBlurImageFilter final : public DlImageFilter {
...
                                              SkScalar sigma_y,
                                              DlTileMode tile_mode) {
-    if (SkScalarIsFinite(sigma_x) && sigma_x > SK_ScalarNearlyZero &&
-        SkScalarIsFinite(sigma_y) && sigma_y > SK_ScalarNearlyZero) {
-      return std::make_shared<DlBlurImageFilter>(sigma_x, sigma_y, tile_mode);
+    if (!SkScalarIsFinite(sigma_x) || !SkScalarIsFinite(sigma_y)) {
+      return nullptr;
     }
-    return nullptr;
+    if (sigma_x < SK_ScalarNearlyZero && sigma_y < SK_ScalarNearlyZero) {
```
flar, line 235: Perhaps also clip the values against NearlyZero rather than just checking if they are smaller, so that if we do make a filter out of them, the filter won't have negative sigmas...?

display_list/effects/dl_image_filter_unittests.cc
```diff
@@ -209,6 +209,16 @@ TEST(DisplayListImageFilter, BlurBounds) {
...
 }
 
+TEST(DisplayListImageFilter, BlurZeroSigma) {
+  std::shared_ptr<DlImageFilter> filter =
+      DlBlurImageFilter::Make(0, 0, DlTileMode::kMirror);
+  ASSERT_EQ(filter, nullptr);
+  filter = DlBlurImageFilter::Make(3, SK_ScalarNaN, DlTileMode::kMirror);
+  ASSERT_EQ(filter, nullptr);
+  filter = DlBlurImageFilter::Make(3, 0, DlTileMode::kMirror);
+  ASSERT_NE(filter, nullptr);
```
flar, line 219: It couldn't hurt to check

- NaN,NaN
- Nan,3
- 0,3

as well...

flar (approved):
Some minor suggestions...

display_list/effects/dl_image_filter.h
jason-simmons: done

display_list/effects/dl_image_filter_unittests.cc
jason-simmons: done

display_list/effects/dl_image_filter.h
```diff
@@ -229,11 +229,15 @@ class DlBlurImageFilter final : public DlImageFilter {
...
+      return nullptr;
+    }
+    sigma_x = std::max(sigma_x, SK_ScalarNearlyZero);
```
flar, line 238: Actually, this is not correct. It basically takes a sigma of zero and makes it non-zero. I think the correct logic is:

`sigma_x = (sigma_x < NearlyZero) ? 0 : sigma_x;`

That's not really a "max" operation as the comparison and the assigned result are different.

flar (changes requested):

display_list/effects/dl_image_filter.h
jason-simmons: done

flar (approved):
LGTM

PR #43828, opened 2023-07-19
[web] Provide convenient default factories for platform views

mdebbar:
Convenient default factories for creating DOM element from a given tag name.

Required by https://github.com/flutter/flutter/pull/130513

Part of https://github.com/flutter/flutter/issues/127030

### DIFF
--- a/lib/web_ui/lib/src/engine/platform_views/content_manager.dart
+++ b/lib/web_ui/lib/src/engine/platform_views/content_manager.dart
@@ -10,6 +10,10 @@ import '../embedder.dart';
 import '../util.dart';
 import 'slots.dart';
 
+// Keep these values in sync with the ones used in the framework's `HtmlElementView`.
+const String _kDefaultVisibleViewType = '_default_document_create_element_visible';
+const String _kDefaultInvisibleViewType = '_default_document_create_element_invisible';
+
 /// This class handles the lifecycle of Platform Views in the DOM of a Flutter Web App.
 ///
 /// There are three important parts of Platform Views. This class manages two of
@@ -25,6 +29,19 @@ import 'slots.dart';
 /// This class keeps a registry of `factories`, `contents` so the framework can
 /// CRUD Platform Views as needed, regardless of the rendering backend.
 class PlatformViewManager {
+  PlatformViewManager() {
+    // Register some default factories.
+    registerFactory(
+      _kDefaultVisibleViewType,
+      _defaultFactory,
+    );
+    registerFactory(
+      _kDefaultInvisibleViewType,
+      _defaultFactory,
+      isVisible: false,
+    );
+  }
+
   // The factory functions, indexed by the viewType
   final Map<String, Function> _factories = <String, Function>{};
 
@@ -222,4 +239,19 @@ class PlatformViewManager {
     _viewIdToType.clear();
     return result;
   }
+}
+
+DomElement _defaultFactory(
+  int viewId, {
+  Object? params,
+}) {
+  params!;
+  params as _DefaultFactoryParams;
+  return domDocument.createElement(params.tagName);
+}
+
+typedef _DefaultFactoryParams = Map<Object?, Object?>;
+
+extension on _DefaultFactoryParams {
+  String get tagName => this['tagName']! as String;
 }
--- a/lib/web_ui/test/engine/platform_views/content_manager_test.dart
+++ b/lib/web_ui/test/engine/platform_views/content_manager_test.dart
@@ -6,7 +6,8 @@ import 'package:test/bootstrap/browser.dart';
 import 'package:test/test.dart';
 import 'package:ui/src/engine.dart';
 
-import '../../common/matchers.dart';
+const String _kDefaultVisibleViewType = '_default_document_create_element_visible';
+const String _kDefaultInvisibleViewType = '_default_document_create_element_invisible';
 
 void main() {
   internalBootstrapBrowserTest(() => testMain);
@@ -97,13 +98,14 @@ void testMain() {
       });
 
       test('refuse to render views for unregistered factories', () async {
-        try {
-          contentManager.renderContent(unregisteredViewType, viewId, null);
-          fail('renderContent should have thrown an Assertion error!');
-        } catch (e) {
-          expect(e, isAssertionError);
-          expect((e as AssertionError).message, contains(unregisteredViewType));
-        }
+        expect(
+          () => contentManager.renderContent(unregisteredViewType, viewId, null),
+          throwsA(const TypeMatcher<AssertionError>().having(
+            (AssertionError error) => error.message,
+            'assertion message',
+            contains(unregisteredViewType),
+          )),
+        );
       });
 
       test('rendered markup contains required attributes', () async {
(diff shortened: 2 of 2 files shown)


lib/web_ui/lib/src/engine/platform_views/content_manager.dart
```diff
@@ -223,3 +240,18 @@ class PlatformViewManager {
...
+DomElement _defaultFactory(
+  int viewId, {
+  Object? params,
+}) {
+  params!;
+  params as _DefaultFactoryParams;
+  return domDocument.createElement(params.tagName);
+}
+
+typedef _DefaultFactoryParams = Map<Object?, Object?>;
```
ditman, line 253:
```suggestion
typedef _DefaultFactoryParams = Map<Object, Object?>;
```

lib/web_ui/lib/src/engine/platform_views/content_manager.dart
```diff
@@ -223,3 +240,18 @@ class PlatformViewManager {
...
+
+extension on _DefaultFactoryParams {
```
ditman, line 255: What a wacky syntax we can get with extensions, I love it (in small doses) xD

lib/web_ui/lib/src/engine/platform_views/content_manager.dart
```diff
@@ -223,3 +240,18 @@ class PlatformViewManager {
     return result;
   }
 }
+
+DomElement _defaultFactory(
+  int viewId, {
+  Object? params,
+}) {
+  params!;
+  params as _DefaultFactoryParams;
```
ditman, line 249: Can this be one-lineized as?

```suggestion
  params! as _DefaultFactoryParams;
```

lib/web_ui/test/engine/platform_views/content_manager_test.dart
```diff
@@ -97,13 +98,14 @@ void testMain() {
...
-        try {
-          contentManager.renderContent(unregisteredViewType, viewId, null);
-          fail('renderContent should have thrown an Assertion error!');
-        } catch (e) {
-          expect(e, isAssertionError);
-          expect((e as AssertionError).message, contains(unregisteredViewType));
-        }
+        expect(
+          () => contentManager.renderContent(unregisteredViewType, viewId, null),
+          throwsA(const TypeMatcher<AssertionError>().having(
```
ditman, line 103: Much better!

lib/web_ui/lib/src/engine/platform_views/content_manager.dart
```diff
@@ -10,6 +10,10 @@ import '../embedder.dart';
 import '../util.dart';
 import 'slots.dart';
 
+// Keep these values in sync with the ones used in the framework's `HtmlElementView`.
+const String _kDefaultVisibleViewType = '_default_document_create_element_visible';
+const String _kDefaultInvisibleViewType = '_default_document_create_element_invisible';
```
ditman, line 15: Should these two consts be made public through the `PlatformViewRegistry` in package:ui_web, so the framework doesn't need to rely on having a copy of them in sync?

ditman (commented):
This looks great! I think that if we exposed the constants of the default factory names from package:ui_web, the implementation would be slightly more robust (one less thing to have to worry about I guess?)

lib/web_ui/lib/src/engine/platform_views/content_manager.dart
mdebbar: This is because the type sent from `HtmlElementView` is `Map<dynamic, dynamic>`, and doing `Map<dynamic, dynamic> as Map<Object, Object?>` fails.

But this is a good idea. I'll change it on the framework to `Map<Object, Object?>` instead.

mdebbar: Hmm I actually have a feeling this isn't going to work but I'll try it anyway.

`creationParams` go through a platform message and get serialized & deserialized, losing all their type information. So they are deserialized back as a `Map<dynamic, dynamic>`.

lib/web_ui/lib/src/engine/platform_views/content_manager.dart
mdebbar: I actually removed this because I was able to do the same thing using the existing `readX` extensions:
https://github.com/flutter/engine/blob/e590b24f3962fda3ec9144dcee3f7565b195839a/lib/web_ui/lib/src/engine/util.dart#L560

lib/web_ui/lib/src/engine/platform_views/content_manager.dart
mdebbar: Great idea!

lib/web_ui/lib/src/engine/platform_views/content_manager.dart
mdebbar: Yep, this didn't work. Even if I change `HtmlElementView` to send `creationParams` as `Map<Object, Object>`, it arrives to the engine as `Map<dynamic, dynamic>` (because of serialization & deserialization of platform messages).

mdebbar:
@ditman this is ready for a re-review!

lib/web_ui/lib/src/engine/platform_views/content_manager.dart
ditman: Good example with the `foo.toString()`, thanks!

lib/web_ui/lib/ui_web/src/ui_web/platform_view_registry.dart
```diff
@@ -4,6 +4,18 @@
...
+///
+/// There's no need to register this view type with [PlatformViewRegistry]
+/// because it is registered by default.
+const String kDefaultVisibleViewType = '_default_document_create_element_visible';
+
+/// The view type of the factory that creates invisible platform view DOM elements.
+///
+/// There's no need to register this view type with [PlatformViewRegistry]
+/// because it is registered by default.
+const String kDefaultInvisibleViewType = '_default_document_create_element_invisible';
```
ditman, line 17: My only change would be to make these `static final` of the `PlatformViewManager` so they're encapsulated within the class that's going to make use of them, rather than being some free globals of the library. they'd be used as `web_ui.PlatformViewManager.kDefaultInvisibleViewType` from the framework.

ditman (approved):
Small comment about encapsulating the currently global consts inside the class that "owns" them, as `static final`s, if possible!

lib/web_ui/lib/ui_web/src/ui_web/platform_view_registry.dart
mdebbar: Really good call!

I also found this section in our style guide that supports your comment: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#begin-global-constant-names-with-prefix-k

lib/web_ui/lib/ui_web/src/ui_web/platform_view_registry.dart
```diff
@@ -19,6 +19,20 @@ final PlatformViewRegistry platformViewRegistry = PlatformViewRegistry();
 
 /// A registry for factories that create platform views.
 class PlatformViewRegistry {
+  /// The view type of the built-in factory that creates visible platform view
+  /// DOM elements.
+  ///
+  /// There's no need to register this view type with [PlatformViewRegistry]
+  /// because it is registered by default.
+  static const String defaultVisibleViewType = '_default_document_create_element_visible';
```
ditman, line 27: ❤️ and without the "k" prefix.

tumblr_70976177f0cdc973a43fba42c9c04621_20bc424a_400

lib/web_ui/lib/src/engine/platform_views/content_manager.dart
ditman: Ah yes, `dynamic` is bad, and makes sense why you had to substitute by `Object?`... And I guess there's nothing "wrong" in using `null` as a possible key for a Map :/

PR #43854, opened 2023-07-20
[web] Preserve correct CanvasKit Variant during test initialization

mdebbar:
At some point, we started setting `useColorEmoji` to true in our tests, but we were doing it in a way that resets all other configurations to their defaults. This caused the `canvasKitVariant` config to be lost and always set to the default `auto`.

This PR fixes the issue and adds tests to:

1. Make sure that the CanvasKit suite always runs with a specific variant (not `auto`).
2. Make sure the given CanvasKit variant makes it all the way through to the tests.

The test harness uses a backdoor (a global JS property on `window`) to communicate which canvaskit variant it's using. The test then compares that with `configuration.canvasKitVariant` to make sure they match. If they don't match, then the configuration was lost somewhere on the way.

Fixes https://github.com/flutter/flutter/issues/130993

### DIFF
--- a/lib/web_ui/test/common/test_initialization.dart
+++ b/lib/web_ui/test/common/test_initialization.dart
@@ -22,11 +22,11 @@ void setUpUnitTests({
     }
 
     // Some of our tests rely on color emoji
-    final engine.FlutterConfiguration config = engine.FlutterConfiguration()
-      ..setUserConfiguration(
-        js_util.jsify(<String, Object?>{
-          'useColorEmoji': true,
-        }) as engine.JsFlutterConfiguration);
+    final engine.FlutterConfiguration config = engine.configuration.merge(
+      js_util.jsify(<String, Object?>{
+        'useColorEmoji': true,
+      }) as engine.JsFlutterConfiguration,
+    );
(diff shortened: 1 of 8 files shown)


hterkelsen (approved):
LGTM

lib/web_ui/test/common/test_initialization.dart
```diff
@@ -22,11 +22,11 @@ void setUpUnitTests({
...
 
     // Some of our tests rely on color emoji
-    final engine.FlutterConfiguration config = engine.FlutterConfiguration()
-      ..setUserConfiguration(
-        js_util.jsify(<String, Object?>{
-          'useColorEmoji': true,
-        }) as engine.JsFlutterConfiguration);
+    final engine.FlutterConfiguration config = engine.configuration.merge(
+      js_util.jsify(<String, Object?>{
+        'useColorEmoji': true,
```
ditman, line 27: I think it'd been much easier to pass `useColorEmoji: true` on the `window.flutterConfiguration` JS object of the `test_platform.dart`?

That way the value would be read as part of the default egine configuration (since it seems that it's being applied to all tests anyway?)

I can try a PR later if you want!

PR #43865, opened 2023-07-20
Add a breadcrumb for people who want to run the web tests

Hixie:


testing/run_tests.sh
```diff
@@ -6,3 +6,6 @@ BUILD_VARIANT="${1:-host_debug_unopt}"
 CURRENT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 
 python3 "${CURRENT_DIR}/run_tests.py" --variant="${BUILD_VARIANT}" --type=engine,dart,benchmarks
+
+# This does not run the web_ui tests. To run those, use:
+#   lib/web_ui/dev-felt test
```
yjbanov, line 11: nit: it's `lib/web_ui/dev/felt`

yjbanov (approved):

testing/run_tests.sh
Hixie: uh, that's not a nit, that's the difference between this being useful and it being completely wrong :sweat_smile:

Hixie: fixed

PR #43874, opened 2023-07-21
add ColorFilter.toString to web_ui

Hixie:


lib/web_ui/lib/src/engine/color_filter.dart
```diff
@@ -121,4 +121,18 @@ class EngineColorFilter implements SceneImageFilter, ui.ColorFilter {
   /// Color filters don't affect the image bounds
   @override
   ui.Rect filterBounds(ui.Rect inputBounds) => inputBounds;
+
+  @override
+  String toString() {
+    switch (type) {
```
kevmoo, line 127: could be switch expression!

Hixie: it could be an if chain too :-P

kevmoo: ...but a switch expression is better.

Hixie: why?

kevmoo: It's tighter. It makes it trivial to parse the intent. You KNOW every branch yields a value. Greatly increases scannability of the code IMHO

yjbanov (approved):

lib/web_ui/lib/src/engine/color_filter.dart
Hixie: I don't think there's a meaningful difference in readability between these two:

```dart
String toString() {
  switch (type) {
    case ColorFilterType.mode:
      return 'ColorFilter.mode($color, $blendMode)';
    case ColorFilterType.matrix:
      return 'ColorFilter.matrix($matrix)';
    case ColorFilterType.linearToSrgbGamma:
      return 'ColorFilter.linearToSrgbGamma()';
    case ColorFilterType.srgbToLinearGamma:
      return 'ColorFilter.srgbToLinearGamma()';
  }
}
```

```dart
String toString() {
  return switch (type) {
    ColorFilterType.mode => 'ColorFilter.mode($color, $blendMode)',
    ColorFilterType.matrix => 'ColorFilter.matrix($matrix)',
    ColorFilterType.linearToSrgbGamma => 'ColorFilter.linearToSrgbGamma()',
    ColorFilterType.srgbToLinearGamma => 'ColorFilter.srgbToLinearGamma()',
  };
}
```

In both cases the language enforces that there must be a return. In both cases it scans relatively easily (maybe the first is a little easier, in the second the `return` is not as obvious because of being on the same line as the `switch`, but maybe the second is easier because of the avoidance of line wrapping).

The former in general is better IMHO in this case just because it's more consistent with other code in the same package (it's literally just copied and pasted from the existing code elsewhere), and because it's easier to maintain if we ever have to add a branch that's more complicated than just a single string return.

PR #43878, opened 2023-07-21
[Impeller] Do not use clear color optimization if the subpass is being collapsed into its parent

jason-simmons:
The InlinePassContext will not apply the clear color within a parent pass that is already in progress.

Fixes https://github.com/flutter/flutter/issues/130824

### DIFF
--- a/impeller/entity/entity_pass.cc
+++ b/impeller/entity/entity_pass.cc
@@ -604,12 +604,10 @@ bool EntityPass::OnRender(
     return false;
   }
 
-  if (!(GetClearColor(root_pass_size) == Color::BlackTransparent())) {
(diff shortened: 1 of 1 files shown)


jonahwilliams (commented):
DBC: it would be preferable from a performance perspective to still collapse the parent pass and instead undo the clear color optimization. The pass collapse avoids an expensive allocation and texture fill while the clear color optimization only removes a draw call.

jason-simmons:
Changed this to disable the clear color optimization for subpasses that are collapsed into the parent

jonahwilliams (approved):
LGTM

bdero (approved):
LGTM!

PR #43905, opened 2023-07-21
[Impeller] Disable color attachment removal for clips on VK

bdero:
This is causing a validation failure on the Vulkan backend at ToT, as discovered by @jonahwilliams.

Need to figure out why this isn't getting hit in the Vulkan goldens.


jonahwilliams (approved):
LGTM

PR #43911, opened 2023-07-21
[Impeller] Capture SkFonts instead of reconstructing them.

bdero:
WIP.

I'll restructure the text backend to make this not a hack prior to landing if it works.


bdero:
After further debugging, I uncovered the real issues: https://github.com/flutter/engine/pull/43919

PR #43919, opened 2023-07-22
[Impeller] Fix text glyph bounds on Android.

bdero:
Resolves https://github.com/flutter/flutter/issues/128624.

It turns out that `SkFont::getBounds()` snaps results to integers on Android, but not iOS. By scaling the font up and scaling the resulting per-glyph bounds back down, we can ensure that the results are always precise enough.

I also found errors with our usage of the computed bounds, but those were comparatively minor fixes.

Before:

https://github.com/flutter/engine/assets/919017/2d376868-1009-4223-851f-5ff17d9994ed

After:

https://github.com/flutter/engine/assets/919017/9efdc82f-9c9b-43c2-be57-ad95b4d5c29c

Minimal repro app:

```dart
import 'dart:ui' as ui;

void main() {
  runApp(const MyApp());
}

class Painter extends CustomPainter {
  Painter(this.scale, this.textSize);

  double scale = 1;
  double textSize = 1;

  @override
    debugPrint("Scale: $scale  FontSize: $textSize");

    ui.ParagraphBuilder b = ui.ParagraphBuilder(
        ui.ParagraphStyle(fontSize: textSize, fontStyle: FontStyle.normal))
      ..pushStyle(ui.TextStyle(color: Colors.black))
      ..addText("i");
    ui.Paragraph p = b.build();
    p.layout(ui.ParagraphConstraints(width: size.width));

    canvas.scale(scale);
    canvas.drawParagraph(p, const Offset(0, 0));
  }

  @override
    return true;
  }
}

class MyApp extends StatefulWidget {
  const MyApp({super.key});

  @override
  State<MyApp> createState() => _MyAppState();
}

class _MyAppState extends State<MyApp> {
  double scale = 1;
  double textSize = 10;

  @override
  Widget build(BuildContext context) => MaterialApp(
        debugShowCheckedModeBanner: false,
        home: Scaffold(
          body: Flex(
            direction: Axis.vertical,
            children: [
              const SizedBox.square(dimension: 50),
              Slider(
                  min: 0,
                  max: 10,
                  value: scale,
                  onChanged: (double value) {
                    setState(() {
                      scale = value;
                    });
                  }),
              Slider(
                  min: 0,
                  max: 50,
                  value: textSize,
                  onChanged: (double value) {
                    setState(() {
                      textSize = value;
                    });
                  }),
              Expanded(child: CustomPaint(painter: Painter(scale, textSize))),
            ],
          ),
        ),
      );
}
```

### DIFF
--- a/impeller/display_list/dl_dispatcher.cc
+++ b/impeller/display_list/dl_dispatcher.cc
@@ -1104,8 +1104,7 @@ void DlDispatcher::drawDisplayList(
 void DlDispatcher::drawTextBlob(const sk_sp<SkTextBlob> blob,
                                 SkScalar x,
                                 SkScalar y) {
-  const auto text_frame = TextFrameFromTextBlob(
-      blob, canvas_.GetCurrentTransformation().GetMaxBasisLengthXY());
+  const auto text_frame = TextFrameFromTextBlob(blob);
   if (paint_.style == Paint::Style::kStroke) {
     auto path = skia_conversions::PathDataFromTextBlob(blob);
     auto bounds = text_frame.GetBounds();
--- a/impeller/display_list/dl_dispatcher.cc
+++ b/impeller/display_list/dl_dispatcher.cc
@@ -1104,7 +1104,8 @@ void DlDispatcher::drawDisplayList(
 void DlDispatcher::drawTextBlob(const sk_sp<SkTextBlob> blob,
                                 SkScalar x,
                                 SkScalar y) {
-  const auto text_frame = TextFrameFromTextBlob(blob);
+  const auto text_frame = TextFrameFromTextBlob(
+      blob, canvas_.GetCurrentTransformation().GetMaxBasisLengthXY());
   if (paint_.style == Paint::Style::kStroke) {
     auto path = skia_conversions::PathDataFromTextBlob(blob);
     auto bounds = text_frame.GetBounds();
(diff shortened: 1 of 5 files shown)


zanderso (approved):

impeller/display_list/dl_dispatcher.cc
```diff
@@ -1104,7 +1104,8 @@ void DlDispatcher::drawDisplayList(
 void DlDispatcher::drawTextBlob(const sk_sp<SkTextBlob> blob,
                                 SkScalar x,
                                 SkScalar y) {
-  const auto text_frame = TextFrameFromTextBlob(blob);
+  const auto text_frame = TextFrameFromTextBlob(
```
jonahwilliams, line 1107: I wonder if this will work against @dnfield s work to defer computing the text scale. Seems like this is strictly an improvement for now, but we should probably also defer the initial scale value?

bdero: Yeah, good catch. Using the scale here could indeed cause problems. Text frames drawn through DrawPicture might not end up having precise enough glyph bounds depending on how much the elements of the picture get scaled up.

We can get away with just using a decently large number here instead of the scale, so I went ahead and removed the usage and checked to make sure it's still working with unreasonably huge glyphs on Android.

bdero: After doing this, there are no longer any overlapping characters in the `MaybeHasOverlapping` test. The accompanying comment was:
> Characters probably have overlap due to low fidelity text metrics, but this could be fixed.

So if there happens to be an accompanying bug I guess we can close that too. :)

jonahwilliams: Ahh interesting. So we don't actually end up rasterizing at a higher scale, we're essentially just tricking Skia into giving higher fidelity metrics.

I don't think I filed a bug for that particular overlap issue, I would update the expectation and delete the TODO.

jonahwilliams (approved):
LGTM

bdero:
There are no golden changes listed. ¯\_(ツ)_/¯ Merging this.

PR #43946, opened 2023-07-24
[Impeller] Fix clip pipeline validation failure; add dedicated clip shaders.

bdero:
Turns out removing the color attachment is failing Metal validation on iOS due to the pipeline color attachments mismatching the framebuffer (even though it winds up rendering correctly).

So this patch replaces the mechanism with a better default config, including:
* dedicated shaders that are less expensive than Solid Fill
* destination blending

I'll take another swing at trying to get it to work if the iOS benchmarks pop back up a bit from this. For now, this fixes the validation failure and should make clips a tad faster on Vulkan and GLES.


jonahwilliams (approved):
LGTM

The color attachment issue would require a similar fix to the vulkan backend where the pipeline we use to create the prototype has the correct color attachment, right?

PR #43948, opened 2023-07-24
[Impeller] Skip rect clips that do nothing.

bdero:
Resolves https://github.com/flutter/flutter/issues/131100

* Skip rect clips that we know won't further restrict the clip shape.
* Don't append clips to enforce subpass bounds. We currently don't collapse passes when subpass clips are added anyhow, so just a restriction to subpass collapsing in order to keep the current behavior intact.
* Small refactor to reduce confusion: Just place the subpass bounds limit into `EntityPass` itself instead of the delegate.

Built on https://github.com/flutter/engine/pull/43946 (because iOS currently fails validation without it).

Gallery home screen before/after:

compare

### DIFF
--- a/impeller/aiks/canvas.cc
+++ b/impeller/aiks/canvas.cc
@@ -280,6 +280,18 @@ void Canvas::ClipPath(const Path& path, Entity::ClipOperation clip_op) {
 }
 
 void Canvas::ClipRect(const Rect& rect, Entity::ClipOperation clip_op) {
+  auto& cull_rect = xformation_stack_.back().cull_rect;
+  if (clip_op == Entity::ClipOperation::kIntersect &&               //
+      cull_rect.has_value() &&                                      //
+      xformation_stack_.back().xformation.IsTranslationScaleOnly()  //
+  ) {
+    auto transformed_rect =
+        rect.TransformBounds(xformation_stack_.back().xformation);
+    if (transformed_rect.Contains(*cull_rect)) {
+      return;  // This clip will do nothing, so skip it.
+    }
+  }
+
   ClipGeometry(Geometry::MakeRect(rect), clip_op);
   switch (clip_op) {
     case Entity::ClipOperation::kIntersect:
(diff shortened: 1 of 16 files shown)


dnfield (approved):

impeller/aiks/canvas.cc
```diff
@@ -280,6 +280,18 @@ void Canvas::ClipPath(const Path& path, Entity::ClipOperation clip_op) {
...
+  auto& cull_rect = xformation_stack_.back().cull_rect;
+  if (clip_op == Entity::ClipOperation::kIntersect &&               //
+      cull_rect.has_value() &&                                      //
+      xformation_stack_.back().xformation.IsTranslationScaleOnly()  //
+  ) {
+    auto transformed_rect =
+        rect.TransformBounds(xformation_stack_.back().xformation);
+    if (transformed_rect.Contains(*cull_rect)) {
+      return;  // This clip will do nothing, so skip it.
+    }
```
jonahwilliams, line 292: This is basically the same as Geometry::CoversArea right? If we use that method instead, then I imagine we could also handle ClipRRects. To do that we could give FilledPathGeometry an optional inner_rect parameter to do the computation (and false if not present)

That would handle the case where a rrect material is scrolled to entirely cover the screen.

I had a similar idea here: https://github.com/jonahwilliams/engine/commit/e3deb28680e9ebd181146689126529337ff688ea

(Though some of the cullrect math is wrong, its right in your PR though 😆 )

impeller/entity/entity_pass.h
```diff
@@ -54,6 +54,10 @@ class EntityPass {
 
   void SetDelegate(std::unique_ptr<EntityPassDelegate> delgate);
 
+  void SetBoundsLimit(std::optional<Rect> bounds_limit);
```
gaaclarke, line 57: This should have a docstring.  In other comments you've added in this PR mention "absorbed clips".  We should mention that this what we are talking about.

bdero: Done.

impeller/entity/entity_pass.h
```diff
@@ -54,6 +54,19 @@ class EntityPass {
...
+  /// @brief  Set the bounds limit, which is provided by the user when creating
+  ///         a SaveLayer. This is a hint that allows the user to communicate
+  ///         that it's OK to not render content outside of the bounds.
+  ///
+  ///         For consistency with Skia, we effectively treat this like a
+  ///         rectangle clip by forcing the subpass texture size to never exceed
+  ///         it.
+  void SetBoundsLimit(std::optional<Rect> bounds_limit);
+
+  /// @brief  Set the bounds limit, which is provided by the user when creating
```
gaaclarke, line 66:
```suggestion
  /// @brief  Get the bounds limit, which is provided by the user when creating
```

impeller/entity/geometry/geometry.h
```diff
@@ -1,4 +1,5 @@
 // Copyright 2013 The Flutter Authors. All rights reserved.
+// Copyright 2013 The Flutter Authors. All rights reserved.
```
jonahwilliams, line 2:
```suggestion
```

impeller/entity/geometry/fill_path_geometry.cc
```diff
@@ -147,4 +149,16 @@ std::optional<Rect> FillPathGeometry::GetCoverage(
   return path_.GetTransformedBoundingBox(transform);
 }
 
+bool FillPathGeometry::CoversArea(const Matrix& transform,
```
jonahwilliams, line 152: I would add a few sanity unit tests that cover the rrect cases, i.e. large rect, empty rect, et cetera.

jonahwilliams (approved):
LGTM with nits

impeller/entity/entity_pass.h
bdero: Fixed.

impeller/entity/geometry/geometry.h
bdero: Fixed.

impeller/entity/geometry/fill_path_geometry.cc
bdero: Good idea, done.

PR #43966, opened 2023-07-24
[do not merge] google testing verify early exit on neutral state

XilaiZhang:
similar to https://github.com/flutter/flutter/pull/130046 and https://github.com/flutter/flutter/pull/127895
this PR is used as an integration test to verify the logics we landed in cl/550076135

When the PR only includs file changes that do not need to be rolled into google 3, Google testing should report a neutral state and return the state early.


XilaiZhang:
just realized that we don't run google testing on engine prs. thank you casey!

PR #43972, opened 2023-07-24
Flutter iOS Interactive Keyboard: Take Screenshot and Handle Pointer Movement 

Matt2D:
This PR address the movement aspect of the flutter interactive keyboard. It handles pointer movement while a scroll view widget is visible, and the interactive behavior is chosen for keyboardDismissBehavior. This is a desired behavior of the keyboard that has not yet been implemented. 
Design Document:
https://docs.google.com/document/d/1-T7_0mSkXzPaWxveeypIzzzAdyo-EEuP5V84161foL4/edit?pli=1

Issues Address:
https://github.com/flutter/flutter/issues/57609

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.h
+++ b/shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.h
@@ -167,7 +167,7 @@ FLUTTER_DARWIN_EXPORT
 @end
 
 @interface UIView (FindFirstResponder)
-- (id)flt_firstResponder;
+  @property (nonatomic, readonly) id flutterFirstResponder;
 @end
 
 #endif  // SHELL_PLATFORM_IOS_FRAMEWORK_SOURCE_FLUTTERTEXTINPUTPLUGIN_H_
--- a/shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.h
+++ b/shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.h
@@ -165,4 +165,9 @@ FLUTTER_DARWIN_EXPORT
 - (instancetype)initWithOwner:(FlutterTextInputPlugin*)textInputPlugin NS_DESIGNATED_INITIALIZER;
 
 @end
+
+@interface UIView (FindFirstResponder)
+- (id)flt_firstResponder;
+@end
+
 #endif  // SHELL_PLATFORM_IOS_FRAMEWORK_SOURCE_FLUTTERTEXTINPUTPLUGIN_H_
--- a/shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.h
+++ b/shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.h
@@ -167,7 +167,7 @@ FLUTTER_DARWIN_EXPORT
 @end
 
 @interface UIView (FindFirstResponder)
-  @property (nonatomic, readonly) id flutterFirstResponder;
+@property(nonatomic, readonly) id flutterFirstResponder;
 @end
 
 #endif  // SHELL_PLATFORM_IOS_FRAMEWORK_SOURCE_FLUTTERTEXTINPUTPLUGIN_H_
--- a/shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
+++ b/shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
@@ -45,6 +45,8 @@ static NSString* const kDeprecatedSetSelectionRectsMethod = @"TextInput.setSelec
 static NSString* const kSetSelectionRectsMethod = @"Scribble.setSelectionRects";
 static NSString* const kStartLiveTextInputMethod = @"TextInput.startLiveTextInput";
 static NSString* const kUpdateConfigMethod = @"TextInput.updateConfig";
+static NSString* const kOnInteractiveKeyboardPointerMoveMethod =
+    @"TextInput.onPointerMoveForInteractiveKeyboard";
 
 #pragma mark - TextInputConfiguration Field Names
 static NSString* const kSecureTextEntry = @"obscureText";
@@ -761,6 +763,7 @@ static BOOL IsSelectionRectBoundaryCloserToPoint(CGPoint point,
 @property(nonatomic, assign) CGRect markedRect;
 @property(nonatomic) BOOL isVisibleToAutofill;
 @property(nonatomic, assign) BOOL accessibilityEnabled;
+@property(nonatomic, assign) int textInputClient;
 // The composed character that is temporarily removed by the keyboard API.
 // This is cleared at the start of each keyboard interaction. (Enter a character, delete a character
 // etc)
@@ -2214,6 +2217,11 @@ static BOOL IsSelectionRectBoundaryCloserToPoint(CGPoint point,
 @property(nonatomic, retain) FlutterTextInputView* activeView;
 @property(nonatomic, retain) FlutterTextInputViewAccessibilityHider* inputHider;
 @property(nonatomic, readonly, weak) id<FlutterViewResponder> viewResponder;
+
+@property(nonatomic, strong) UIView* keyboardViewContainer;
+@property(nonatomic, strong) UIView* keyboardView;
+@property(nonatomic, strong) UIView* firstResponder;
+@property(nonatomic, assign) CGRect keyboardRect;
 @end
 
 @implementation FlutterTextInputPlugin {
@@ -2222,16 +2230,27 @@ static BOOL IsSelectionRectBoundaryCloserToPoint(CGPoint point,
 
 - (instancetype)initWithDelegate:(id<FlutterTextInputDelegate>)textInputDelegate {
   self = [super init];
-
   if (self) {
     // `_textInputDelegate` is a weak reference because it should retain FlutterTextInputPlugin.
     _textInputDelegate = textInputDelegate;
     _autofillContext = [[NSMutableDictionary alloc] init];
     _inputHider = [[FlutterTextInputViewAccessibilityHider alloc] init];
     _scribbleElements = [[NSMutableDictionary alloc] init];
+    _keyboardViewContainer = [[UIView alloc] init];
+
+    [[NSNotificationCenter defaultCenter] addObserver:self
+                                             selector:@selector(handleKeyboardWillShow:)
+                                                 name:UIKeyboardWillShowNotification
+                                               object:nil];
   }
 
   return self;
+}
+
+- (void)handleKeyboardWillShow:(NSNotification*)notification {
+  NSDictionary* keyboardInfo = [notification userInfo];
+  NSValue* keyboardFrameEnd = [keyboardInfo valueForKey:UIKeyboardFrameEndUserInfoKey];
+  _keyboardRect = [keyboardFrameEnd CGRectValue];
 }
 
 - (void)dealloc {
@@ -2295,9 +2314,60 @@ static BOOL IsSelectionRectBoundaryCloserToPoint(CGPoint point,
   } else if ([method isEqualToString:kUpdateConfigMethod]) {
     [self updateConfig:args];
     result(nil);
+  } else if ([method isEqualToString:kOnInteractiveKeyboardPointerMoveMethod]) {
+    CGFloat pointerY = (CGFloat)[args[@"pointerY"] doubleValue];
+    [self handlePointerMove:pointerY];
+    result(nil);
   } else {
     result(FlutterMethodNotImplemented);
   }
+}
+
+- (void)handlePointerMove:(CGFloat)pointerY {
+  double screenHeight = [UIScreen mainScreen].bounds.size.height;
+  double keyboardHeight = _keyboardRect.size.height;
+  // Determines if pointer goes overtop of the keyboard.
+  if (screenHeight - keyboardHeight <= pointerY) {
+    // If no screenshot has been taken.
+    if (_keyboardView.superview != nil) {
+      [self takeScreenshot];
+      [self hideKeyboardWithoutAnimation];
+    } else {
+      [self setKeyboardContainerHeight:pointerY];
+    }
+  } else {
+    // If pointer is not overtop of the original keyboard keep screenshot in place.
+    if (_keyboardView.superview != nil) {
+      _keyboardViewContainer.frame = _keyboardRect;
+    }
+  }
+}
+
+- (void)setKeyboardContainerHeight:(CGFloat)pointerY {
+  CGRect frameRect = _keyboardRect;
+  frameRect.origin.y = pointerY;
+  _keyboardViewContainer.frame = frameRect;
+}
+
+- (void)hideKeyboardWithoutAnimation {
+  [UIView setAnimationsEnabled:NO];
+  _firstResponder = UIApplication.sharedApplication.keyWindow.flt_firstResponder;
+  [_firstResponder resignFirstResponder];
+  [UIView setAnimationsEnabled:YES];
+}
+
+- (void)takeScreenshot {
+  UIView* keyboardSnap = [[UIScreen mainScreen] snapshotViewAfterScreenUpdates:YES];
+  keyboardSnap = [keyboardSnap resizableSnapshotViewFromRect:_keyboardRect
+                                          afterScreenUpdates:YES
+                                               withCapInsets:UIEdgeInsetsZero];
+  _keyboardView = keyboardSnap;
+  [_keyboardViewContainer addSubview:_keyboardView];
+  if (_keyboardViewContainer.superview == nil) {
+    [UIApplication.sharedApplication.delegate.window.rootViewController.view
+        addSubview:_keyboardViewContainer];
+  }
+  _keyboardViewContainer.frame = _keyboardRect;
 }
 
 - (void)setEditableSizeAndTransform:(NSDictionary*)dictionary {
@@ -2762,5 +2832,23 @@ static BOOL IsSelectionRectBoundaryCloserToPoint(CGPoint point,
  */
 - (BOOL)handlePress:(nonnull FlutterUIPressProxy*)press API_AVAILABLE(ios(13.4)) {
   return NO;
+}
+@end
+
+/**
+ * Recursively searches the UIView's subviews to locate the First Responder
+ */
+@implementation UIView (FindFirstResponder)
+- (id)flt_firstResponder {
+  if (self.isFirstResponder) {
+    return self;
+  }
+  for (UIView* subView in self.subviews) {
+    UIView* firstResponder = subView.flt_firstResponder;
+    if (firstResponder) {
+      return firstResponder;
+    }
+  }
+  return nil;
 }
 @end
--- a/shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
+++ b/shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
@@ -2325,25 +2325,25 @@ static BOOL IsSelectionRectBoundaryCloserToPoint(CGPoint point,
 }
 
 - (void)handlePointerMove:(CGFloat)pointerY {
-  //View must be loaded at this point.
+  // View must be loaded at this point.
   UIScreen* screen = _viewController.flutterScreenIfViewLoaded;
   double screenHeight = screen.bounds.size.height;
   double keyboardHeight = _keyboardRect.size.height;
-    if (screenHeight - keyboardHeight <= pointerY) {
+  if (screenHeight - keyboardHeight <= pointerY) {
     // If the pointer is within the bounds of the keyboard.
-      if (_keyboardView.superview == nil) {
+    if (_keyboardView.superview == nil) {
       // If no screenshot has been taken.
-        [self takeKeyboardScreenshotAndDisplay];
-        [self hideKeyboardWithoutAnimation];
-      } else {
-        [self setKeyboardContainerHeight:pointerY];
-      }
+      [self takeKeyboardScreenshotAndDisplay];
+      [self hideKeyboardWithoutAnimation];
     } else {
-      if (_keyboardView.superview != nil) {
-        //Keeps keyboard at proper height.
-        _keyboardViewContainer.frame = _keyboardRect;
-      }
+      [self setKeyboardContainerHeight:pointerY];
     }
+  } else {
+    if (_keyboardView.superview != nil) {
+      // Keeps keyboard at proper height.
+      _keyboardViewContainer.frame = _keyboardRect;
+    }
+  }
 }
 
 - (void)setKeyboardContainerHeight:(CGFloat)pointerY {
@@ -2360,7 +2360,7 @@ static BOOL IsSelectionRectBoundaryCloserToPoint(CGPoint point,
 }
 
 - (void)takeKeyboardScreenshotAndDisplay {
-   //View must be loaded at this point
+  // View must be loaded at this point
   UIScreen* screen = _viewController.flutterScreenIfViewLoaded;
   UIView* keyboardSnap = [screen snapshotViewAfterScreenUpdates:YES];
   keyboardSnap = [keyboardSnap resizableSnapshotViewFromRect:_keyboardRect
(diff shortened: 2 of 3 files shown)


shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.h
```diff
@@ -164,4 +164,9 @@ FLUTTER_DARWIN_EXPORT
 - (instancetype)initWithOwner:(FlutterTextInputPlugin*)textInputPlugin NS_DESIGNATED_INITIALIZER;
 
 @end
+
+@interface UIView (FindFirstResponder)
+- (id)firstResponder;
```
hellohuanlin, line 169: Can you rename to `flt_firstResponder` in case of naming conflict? (ObjC doesn't support namespace)

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
```diff
@@ -45,6 +45,7 @@
 static NSString* const kSetSelectionRectsMethod = @"Scribble.setSelectionRects";
 static NSString* const kStartLiveTextInputMethod = @"TextInput.startLiveTextInput";
 static NSString* const kUpdateConfigMethod = @"TextInput.updateConfig";
+static NSString* const kOnPointerMoveMethod = @"TextInput.onPointerMove";
```
hellohuanlin, line 48: Nit maybe `kOnInteractiveKeyboardPointerMoveMethod = @"TextInput.onPointerMoveForInteractiveKeyboard"`, to be more specific.

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
```diff
@@ -2272,11 +2289,57 @@ - (void)handleMethodCall:(FlutterMethodCall*)call result:(FlutterResult)result {
   } else if ([method isEqualToString:kUpdateConfigMethod]) {
     [self updateConfig:args];
     result(nil);
+  } else if ([method isEqualToString:kOnPointerMoveMethod]) {
+    double pointerY = [args[@"pointerY"] doubleValue];
```
hellohuanlin, line 2293: use `CGRect` rather than `double` for UI stuff.

hellohuanlin (commented):
Overall looks great. I will ask @cbracken to also take a look.

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
Matt2D: With regards to the pointerY this only contains a single double, so I am not sure that this would classify as UI.

hellohuanlin: sorry i meant CGFloat

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
```diff
@@ -2272,11 +2290,57 @@ - (void)handleMethodCall:(FlutterMethodCall*)call result:(FlutterResult)result {
...
+    result(nil);
   } else {
     result(FlutterMethodNotImplemented);
   }
 }
 
+- (void)handlePointerMove:(double)pointerY {
+  double screenHeight = [UIScreen mainScreen].bounds.size.height;
+  double keyboardHeight = _keyboardRect.size.height;
+  // Determines if pointer goes overtop of the keyboard
```
cyanglaz, line 2305: nits:
```suggestion
  // Determines if pointer goes overtop of the keyboard.
```

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
```diff
@@ -2272,11 +2290,57 @@ - (void)handleMethodCall:(FlutterMethodCall*)call result:(FlutterResult)result {
...
+  if (screenHeight - keyboardHeight <= pointerY) {
+    // If no screenshot has been taken
+    if (_keyboardViewContainer.subviews.count == 0) {
```
cyanglaz, line 2308: Using `_keyboardViewContainer.subviews.count == 0` doesn't seems to be safe and future proof. For example, what if someone added a new subview for `_keyboardViewContainer ` and didn't know this logic exist? Could you use a different indicator?

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
```diff
@@ -2272,11 +2290,57 @@ - (void)handleMethodCall:(FlutterMethodCall*)call result:(FlutterResult)result {
...
+      [self takeScreenshot];
+      [self hideKeyboardWithoutAnimation];
+    } else {
+      CGRect frameRect = _keyboardRect;
+      frameRect.origin.y = pointerY;
+      _keyboardViewContainer.frame = frameRect;
```
cyanglaz, line 2314: nits: could you refactor these lines into a method to match what you did in `if`
Maybe something like:
`[self setKeyboardContainerHeight:pointerY]

shell/platform/darwin/ios/framework/Source/FlutterTextInputPluginTest.mm
```diff
@@ -60,6 +60,8 @@ @interface FlutterSecureTextInputView : FlutterTextInputView
 
 @interface FlutterTextInputPlugin ()
 @property(nonatomic, assign) FlutterTextInputView* activeView;
+@property(nonatomic, assign) UIView* keyboardViewContainer;
```
cyanglaz, line 63:
```suggestion
@property(nonatomic, weak) UIView* keyboardViewContainer;
```

cyanglaz (changes requested):

shell/platform/darwin/ios/framework/Source/FlutterTextInputPluginTest.mm
hellohuanlin: @cyanglaz i could be wrong, but iirc the standard practice is just `assign` when exposing the property in test header (similar to the `activeView` properly above). `weak` is confusing since it's actually a `strong` property.

cyanglaz: Oh. I didn't realize it is only exposing the property. I'm not aware of the standard practice tho but since activeView also used assign, I will take your word for it :)

hellohuanlin: @cyanglaz and I just discussed offline - it should be `readonly` here, since your unit test does not need the setter. 

```
@property(nonatomic, readonly) UIView* keyboardViewContainer;
```

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.h
Matt2D: Good idea! Went through and made the change.

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
Matt2D: Just updated, I'll make sure to try and add Interactive Keyboard more in the future, definitely would help with clarity!

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
Matt2D: Gotcha, just updated the code!

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
Matt2D: Will do! Just slipped by me.

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
Matt2D: Definitely a concern, the same thing occurs in the tests so there definitely is a need to update this approach. I went through and included a UIView that represents the screenshot. Then we can see if it the superview matches instead.

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
Matt2D: Done! Always helpful to simplify code into sections to increase understandability.

shell/platform/darwin/ios/framework/Source/FlutterTextInputPluginTest.mm
Matt2D: Gotcha! That works for me, just went through and made the changes.

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
```diff
@@ -2272,11 +2291,62 @@ - (void)handleMethodCall:(FlutterMethodCall*)call result:(FlutterResult)result {
...
+    CGFloat pointerY = (CGFloat)[args[@"pointerY"] doubleValue];
+    [self handlePointerMove:pointerY];
+    result(nil);
   } else {
     result(FlutterMethodNotImplemented);
   }
 }
 
+- (void)handlePointerMove:(CGFloat)pointerY {
+  double screenHeight = [UIScreen mainScreen].bounds.size.height;
```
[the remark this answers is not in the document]
cyanglaz: Is the pointerY based on the screen, or the flutter app?

Since it comes from the Framework, I assume it is based on the flutter app? So In scenarios (add to app, or app extension) when the flutter app does not fill the full screen, the calculation below that determines if the pointer goes overtop of the keyboard should be based on the flutter app's height instead of the screen height, correct?

Maybe instead of the screenHeight, we should use the height of the FlutterViewController's view?

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
```diff
@@ -2272,11 +2291,62 @@ - (void)handleMethodCall:(FlutterMethodCall*)call result:(FlutterResult)result {
...
+  double keyboardHeight = _keyboardRect.size.height;
+  // Determines if pointer goes overtop of the keyboard.
+  if (screenHeight - keyboardHeight <= pointerY) {
+    // If no screenshot has been taken.
+    if (_keyboardView.superview != nil) {
```
cyanglaz, line 2309: Did you mean `if (_keyboardView.superview == nil)`

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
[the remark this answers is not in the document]
mossmana: > We are moving away from using [UIScreen mainScreen]. I don't think we should block this PR because of this, but we will need to follow up to clean up later. FYI @mossmana

@Matt2D While usually I would agree with not blocking the PR, in this particular case, this is a new introduction of [UIScreen mainScreen]. I would prefer that we don't add new instances. You can refer to https://github.com/flutter/flutter/issues/125496 for how we are handling the deprecation in other places.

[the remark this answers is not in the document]
hellohuanlin: @cyanglaz and I were discussing this the other day, and it's actually a bit tricky here - we would need to refactor [this code](https://github.com/flutter/engine/pull/43690/files#diff-365b13daa5168d36d5824f220bc472a4a90f4685761947f8daf9dc76e7f8d962R480-R497) to make it re-usable. 

I think I can take this part over. Can make a PR today.

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.h
```diff
@@ -164,4 +164,9 @@ FLUTTER_DARWIN_EXPORT
 - (instancetype)initWithOwner:(FlutterTextInputPlugin*)textInputPlugin NS_DESIGNATED_INITIALIZER;
 
 @end
+
+@interface UIView (FindFirstResponder)
+- (id)flt_firstResponder;
```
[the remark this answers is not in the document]
cbracken: Avoid underscore in obj-c identifiers. We're doing something similar in another outstanding PR on the macOS side and using `flutter` as the prefix here:
https://github.com/flutter/engine/pull/43101/files#diff-edbbe0be10d99338c2496cad85a27776c79ec8b260825ee4d309b1e4d194d5b2R10

So we'd end up with `flutterFirstResponder` here.

@cyanglaz @hellohuanlin preferences? We should probably be consistent across embedders.

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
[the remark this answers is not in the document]
cbracken: @cyanglaz in this case, since this is being used to determine the pointer position on the screen, this is relative to the screen rather than the flutter view.

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
```diff
@@ -2272,11 +2291,62 @@ - (void)handleMethodCall:(FlutterMethodCall*)call result:(FlutterResult)result {
...
+    result(nil);
   } else {
     result(FlutterMethodNotImplemented);
   }
 }
 
+- (void)handlePointerMove:(CGFloat)pointerY {
+  double screenHeight = [UIScreen mainScreen].bounds.size.height;
+  double keyboardHeight = _keyboardRect.size.height;
+  // Determines if pointer goes overtop of the keyboard.
```
cbracken, line 2306: Maybe something like:
```suggestion
  // If the pointer is within the bounds of the keyboard.
```

cbracken: (And maybe move the comment inside the conditional block)

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
```diff
@@ -2272,11 +2291,62 @@ - (void)handleMethodCall:(FlutterMethodCall*)call result:(FlutterResult)result {
...
+}
+
+- (void)hideKeyboardWithoutAnimation {
+  [UIView setAnimationsEnabled:NO];
+  _firstResponder = UIApplication.sharedApplication.keyWindow.flt_firstResponder;
+  [_firstResponder resignFirstResponder];
+  [UIView setAnimationsEnabled:YES];
+}
+
+- (void)takeScreenshot {
```
cbracken, line 2336: Consider renaming to something more specific for readability -- e.g. `take KeyboardSnapshot`

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
```diff
@@ -2272,11 +2291,62 @@ - (void)handleMethodCall:(FlutterMethodCall*)call result:(FlutterResult)result {
...
+  UIView* keyboardSnap = [[UIScreen mainScreen] snapshotViewAfterScreenUpdates:YES];
+  keyboardSnap = [keyboardSnap resizableSnapshotViewFromRect:_keyboardRect
+                                          afterScreenUpdates:YES
+                                               withCapInsets:UIEdgeInsetsZero];
+  _keyboardView = keyboardSnap;
+  [_keyboardViewContainer addSubview:_keyboardView];
+  if (_keyboardViewContainer.superview == nil) {
+    [UIApplication.sharedApplication.delegate.window.rootViewController.view
+        addSubview:_keyboardViewContainer];
```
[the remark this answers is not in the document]
cbracken: Since this is positioning a keyboard relative to the full screen -- regardless of whether the FlutterView is full screen, or a subview covering only part of the screen in an add-to-app scenario -- I think we want the root view controller don't we?

shell/platform/darwin/ios/framework/Source/FlutterTextInputPluginTest.mm
```diff
@@ -2315,4 +2318,186 @@ - (void)testSetPlatformViewClient {
...
+  XCTAssertEqual(textInputPlugin.keyboardViewContainer.frame.origin.y, keyboardFrame.origin.y);
+
+  FlutterMethodCall* onPointerMoveCallMove =
+      [FlutterMethodCall methodCallWithMethodName:@"TextInput.onPointerMoveForInteractiveKeyboard"
+                                        arguments:@{@"pointerY" : @(600)}];
+  [textInputPlugin handleMethodCall:onPointerMoveCallMove
+                             result:^(id _Nullable result){
+                             }];
+  XCTAssert(textInputPlugin.keyboardView.superview != nil);
+  CGFloat newHeight = 600;
```
cbracken, line 2404: Consider just inlining this into the call below.

shell/platform/darwin/ios/framework/Source/FlutterTextInputPluginTest.mm
```diff
@@ -2315,4 +2318,186 @@ - (void)testSetPlatformViewClient {
...
```
cbracken, line 2441: Similar comment here.

shell/platform/darwin/ios/framework/Source/FlutterTextInputPluginTest.mm
```diff
@@ -2315,4 +2318,186 @@ - (void)testSetPlatformViewClient {
...
+  XCTAssertEqual(textInputPlugin.keyboardViewContainer.frame.origin.y, newHeight);
+
+  for (UIView* subView in textInputPlugin.keyboardViewContainer.subviews) {
+    [subView removeFromSuperview];
+  }
+}
+
+- (void)testInteractiveKeyboardScreenshotWillBeMovedUpAfterUserScroll {
```
cbracken, line 2412: Maybe rather than just "up" something like "back to original position"?

cbracken (commented):
Overall looks great - just a few comments aside from those left by others!

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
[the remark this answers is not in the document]
cyanglaz: Ah yeah that's true. We would need to make sure to position the keyboard screenshot on the very top. Set ZIndex to maxInt maybe?

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.h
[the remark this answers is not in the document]
cyanglaz: `flutterFirstResponder` SGTM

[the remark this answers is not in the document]
Matt2D: Will do! Thank you for the suggestion

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
Matt2D: Makes sense, thank you!

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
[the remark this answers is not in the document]
hellohuanlin: FYI this is the PR promised: https://github.com/flutter/engine/pull/44303

After that gets landed, you can do something like this here: 

```
UIScreen *screen = _viewController.flutterScreenIfViewLoaded;
if (!screen) {
  return;
}
CGFloat screenHeight = screen.bounds.size.height;
// ...
```

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
```diff
@@ -2272,11 +2291,62 @@ - (void)handleMethodCall:(FlutterMethodCall*)call result:(FlutterResult)result {
...
+    if (_keyboardView.superview != nil) {
+      _keyboardViewContainer.frame = _keyboardRect;
+    }
+  }
+}
+
+- (void)setKeyboardContainerHeight:(CGFloat)pointerY {
+  CGRect frameRect = _keyboardRect;
+  frameRect.origin.y = pointerY;
+  _keyboardViewContainer.frame = frameRect;
```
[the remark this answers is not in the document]
Matt2D: This works great! The one change I added is that we still want to check if the pointer moves over the keyboard before we take the screenshot.

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
Matt2D: Definitely, I also added add to screen to be even better for readability.

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
```diff
@@ -2272,11 +2291,62 @@ - (void)handleMethodCall:(FlutterMethodCall*)call result:(FlutterResult)result {
...
+
+- (void)hideKeyboardWithoutAnimation {
+  [UIView setAnimationsEnabled:NO];
+  _firstResponder = UIApplication.sharedApplication.keyWindow.flt_firstResponder;
+  [_firstResponder resignFirstResponder];
+  [UIView setAnimationsEnabled:YES];
+}
+
+- (void)takeScreenshot {
+  UIView* keyboardSnap = [[UIScreen mainScreen] snapshotViewAfterScreenUpdates:YES];
```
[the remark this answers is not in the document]
Matt2D: Updated with the fix provided by the new PR Huan did.

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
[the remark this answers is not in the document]
Matt2D: Definitely worth adding!

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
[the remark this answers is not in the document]
Matt2D: Thank you for the code snippet, I think the new api is much cleaner! Updated the code using the mentioned solution.

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
Matt2D: Yep you are right! Thanks for catching this.

shell/platform/darwin/ios/framework/Source/FlutterTextInputPluginTest.mm
Matt2D: Definitely!

shell/platform/darwin/ios/framework/Source/FlutterTextInputPluginTest.mm
Matt2D: Done

shell/platform/darwin/ios/framework/Source/FlutterTextInputPluginTest.mm
Matt2D: Good idea! Done

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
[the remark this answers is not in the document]
cyanglaz: Ah I missed that part. So the caller part can probably be
```
if (_keyboardView.superview != nil) {
   [self setKeyboardContainerHeight:pointerY];
} else if (screenHeight - keyboardHeight <= pointerY) {
   [self takeScreenshot];
   [self hideKeyboardWithoutAnimation];
}
```

cyanglaz (approved):
LGTM

hellohuanlin (approved):

hellohuanlin:
You got a build error in test: 

```
../../flutter/shell/platform/darwin/ios/framework/Source/FlutterTextInputPluginTest.mm:2515:72: error: C-style cast from 'NSNumber *' to 'CGFloat' (aka 'double') is not allowed
 2515 |   XCTAssertEqual(textInputPlugin.keyboardViewContainer.frame.origin.y, (CGFloat) @(600));
```

You may have to do 
```
XCTAssertEqual(textInputPlugin.keyboardViewContainer.frame.origin.y, 600.0);
```

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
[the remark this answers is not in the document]
hellohuanlin: Discussed offline - the duplicate logic of `screenHeight - keyboardHeight <= pointerY` can get error prune. 

Another idea is to pass in an additional BOOL to the helper function, but that also makes the code unnecessarily complicate. 

I think it makes sense to keep it as it is.

PR #43974, opened 2023-07-24
Embed a license readme in every zip archive

dnfield:
fixes https://github.com/flutter/flutter/issues/130682

This patch makes all zip files created via the zip_bundle rule have an embedded LICENSE that points to a permalink to the LICENSE for that SDK commit hash. Previously we only did this for a few tools, but it was pointed out that some of the zip files we generate that are meant for consumption in the tool are otherwise missing a stand-alone license file (the tool downloads the full license file separately from the individual zips, as it would be costly to include it with each and every one).


build/zip_bundle.gni
```diff
@@ -38,16 +39,39 @@ if (flutter_runtime_mode == "jit_release") {
...
+        rebase_path("//flutter/sky/packages/sky_engine/LICENSE", "//flutter")
+    git_url = "https://github.com/flutter/engine/tree/$engine_version"
+    outputs = [ license_readme ]
+    contents = [
+      "# $target_name",
+      "",
+      "This bundle is part of the the Flutter SDK.",
+      "",
+      "The source code is hosted at [flutter/engine/$source_path]($git_url/$source_path).",
+      "The for this bundle is hosted at [flutter/engine/sky/packages/sky_engine/LICENSE]($git_url/$license_path).",
```
zanderso, line 58:
```suggestion
      "The license for this bundle is hosted at [flutter/engine/sky/packages/sky_engine/LICENSE]($git_url/$license_path).",
```

Hixie, line 58: is there any way we can make this URL point to a file on the same host as the zip files are hosted?

My concern is that the lifetime of github.com and the lifetime of the host where we upload the zip files is not guaranteed to be correlated, so if we suddenly find github goes away and we have to move all the code, all these old files will no longer point to a license file.

dnfield: We'd have to modify the recipes to do that, but we could. @godofredoc WDYT?

dnfield: Actually, I think we can now control it from the build json.. I'll look into that tomorrow.

dnfield: Ok, I've added a link to the sky_engine.zip location as well, which should have a full copy of the license.

zanderso:
Looks like a couple of unrelated diffs got included in the PR.

dnfield:
> Looks like a couple of unrelated diffs got included in the PR.

I was wondering where those files went... removed.

Hixie:
LGTM but I'm not qualified to review the build file changes, I just don't have recent-enough experience with that.

zanderso (approved):
LGTM

PR #43976, opened 2023-07-24
[Impeller] presentKHR from background worker.

jonahwilliams:
Testing with validation enabled shows no crashes or validation warnings on Pixel 6 or S10. I believe this is safe since we are correctly synchronized in the acquireNextImage via the presentation semaphore. The only loss is that we will finish the raster task before we know that presentation is failed.

Having said that, we can end up waiting multiple ms for the layout transition and presentation, so that doesn't seem worth it.

### DIFF
--- a/impeller/renderer/backend/vulkan/swapchain_impl_vk.cc
+++ b/impeller/renderer/backend/vulkan/swapchain_impl_vk.cc
@@ -384,7 +384,7 @@ SwapchainImplVK::AcquireResult SwapchainImplVK::AcquireNextDrawable() {
       )};
 }
 
-bool SwapchainImplVK::Present(const std::shared_ptr<SwapchainImageVK> image,
+bool SwapchainImplVK::Present(const std::shared_ptr<SwapchainImageVK>& image,
                               uint32_t index) {
   auto context_strong = context_.lock();
   if (!context_strong) {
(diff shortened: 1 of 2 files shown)


jonahwilliams:
The clang tidy lint doesn't seem correct but I could be misunderstanding it. I need to capture the shared ptr and not a reference to the shared ptr for the closure, otherwise when the task executes the stack allocated reference will ahve disappeared. Should I just add an ignore or am I missing something else?

impeller/renderer/backend/vulkan/swapchain_impl_vk.cc
```diff
@@ -384,96 +384,105 @@ SwapchainImplVK::AcquireResult SwapchainImplVK::AcquireNextDrawable() {
...
+bool SwapchainImplVK::Present(const std::shared_ptr<SwapchainImageVK> image,
                               uint32_t index) {
   auto context_strong = context_.lock();
   if (!context_strong) {
     return false;
   }
 
   const auto& context = ContextVK::Cast(*context_strong);
+  context.GetConcurrentWorkerTaskRunner()->PostTask(
+      [&, index, image, current_frame = current_frame_] {
```
dnfield, line 396: The refcount on image will get bumped here. You can move the parameter back to a reference.

dnfield (approved):
LGTM once tidy is happy

impeller/renderer/backend/vulkan/swapchain_impl_vk.cc
jonahwilliams: Done

PR #44005, opened 2023-07-25
Change ifdef to if, remove nognchecks

dnfield:
I don't know why the define is happening, but we should be using if instead of ifdef here.

As long as fuchsia still builds this should be good.


dnfield:
From the GN manual:

```
  If you have conditional includes, make sure the build conditions and the
  preprocessor conditions match, and annotate the line with "nogncheck" (see
  "gn help nogncheck" for an example).
```

GN doesn't understand conditional includes. If we want to get rid of the ignores here we need to refactor this code, which may not be worth it.

PR #44021, opened 2023-07-25
Ignore the NOTICE.fuchsia file in the new license script

jason-simmons:
See https://github.com/flutter/engine/pull/34038


Hixie (approved):
LGTM

My understanding is that the licenses in this file are a mix of licenses that we pick up ourselves in our own detailed scan of the SDK files, and licenses that don't apply to us because we don't use the code covered by those licenses, and that that is why it's ok to ignore this file.

PR #44023, opened 2023-07-25
[web:a11y] fix interactions between scrolling and focus

yjbanov:
Fixes https://github.com/flutter/flutter/issues/130950

Some screen readers, in particular TalkBack on Android, want to scroll to the element that is programmatically focused (more details in the issue). Sometimes, the web engine calls `domElement.focus()` on elements that are scrolled out of the visible viewport, which causes undesired scrolls in the opposite direction from where the user wants to scroll. Additionally, when the currently focused element is removed from the page, TalkBack focuses on the entire `<body>` tag.

To fix the issue, this PR makes the DOM focus more stable as follows:

* When removing a semantics node from the tree, if that node is currently focused, leave a "tombstone", which is the original element with all its attributes intact, but with role managers disposed, the element shifted out of the view and sized to zero, and made `aria-hidden="true"`. Since the element remains focused, it does not cause TalkBack to scroll to anything new.
* When detaching an `AccessibilityFocusManager` from an element, do not `blur()` the element. It is not safe to blur it because, unless another element requests focus, no element will be focused, which will also cause talkback to focus on `<body>`.
* `AccessibilityFocusManager` is now stateful. If it requested the focus on an element before, it won't request it again, unless it was blurred and it is not already focused. This is because requesting focus on an already focused element causes the screen reader to scroll to that element, but the act of shifting focus is not the signal that drives scrolling in Flutter.

Role managers were updated to not do too much clean-up, because the tombstone should continue to look like the original element to the screen reader. Otherwise, changing the role or other ARIA attributes on a focused element risks the screen reader to read the changes out loud.

### DIFF
--- a/lib/web_ui/lib/src/engine/safe_browser_api.dart
+++ b/lib/web_ui/lib/src/engine/safe_browser_api.dart
@@ -1060,3 +1060,25 @@ class OffScreenCanvas {
       !isSafari
       && js_util.hasProperty(domWindow, 'OffscreenCanvas');
 }
+
+/// Determines the shadow scope of the `scopeSibling` then finds the currently
+/// focused (a.k.a. active) element inside it.
+///
+/// The shadow scope is either the subtree under the nearest ancestor shadow
+/// root of the `scopeSibling` element, or the entire document of the page.
+///
+/// If no element is currently active on the page, returns null.
+///
+/// If an element is currently active on the page but that element is in a
+/// different shadow scope, returns null.
+///
+/// If `scopeSibling` is detached from the document, returns null.
+DomElement? getActiveElementInNearestShadowScope(DomNode scopeSibling) {
+  final DomNode root = scopeSibling.getRootNode();
+  if (root is DomDocument) {
+    return root.activeElement;
+  } else if (root is DomShadowRoot) {
+    return root.activeElement;
+  }
+  return null;
+}
--- a/lib/web_ui/lib/src/engine/semantics/semantics.dart
+++ b/lib/web_ui/lib/src/engine/semantics/semantics.dart
@@ -2,6 +2,7 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+import 'dart:async';
 import 'dart:math' as math;
 import 'dart:typed_data';
 
@@ -16,6 +17,7 @@ import '../configuration.dart';
 import '../dom.dart';
 import '../embedder.dart';
 import '../platform_dispatcher.dart';
+import '../safe_browser_api.dart';
 import '../util.dart';
 import '../vector_math.dart';
 import 'checkable.dart';
@@ -521,7 +523,13 @@ abstract class PrimaryRoleManager {
   /// gesture mode changes.
   @mustCallSuper
   void dispose() {
-    semanticsObject.element.removeAttribute('role');
+    final List<RoleManager>? secondaryRoles = _secondaryRoleManagers;
+    if (secondaryRoles == null) {
+      return;
+    }
+    for (final RoleManager secondaryRole in secondaryRoles) {
+      secondaryRole.dispose();
+    }
     _isDisposed = true;
   }
 }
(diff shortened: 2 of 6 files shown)


lib/web_ui/lib/src/engine/safe_browser_api.dart
```diff
@@ -1060,3 +1060,25 @@ class OffScreenCanvas {
...
+/// different shadow scope, returns null.
+///
+/// If `scopeSibling` is detached from the document, returns null.
+DomElement? getActiveElementInNearestShadowScope(DomNode scopeSibling) {
+  final DomNode root = scopeSibling.getRootNode();
+  if (root is DomDocument) {
+    return root.activeElement;
+  } else if (root is DomShadowRoot) {
+    return root.activeElement;
+  }
```
ditman, line 1082: There used to be a `DocumentOrShadowRoot` mixin to not need this type of check, but I cannot find it anymore in MDN, so I guess it's gone! (I'm very outdated: https://github.com/mdn/content/pull/2395)

lib/web_ui/lib/src/engine/semantics/semantics.dart
```diff
@@ -1739,6 +1792,27 @@ class EngineSemanticsOwner {
     });
   }
 
+  final Map<int, Tombstone> _tombstones = <int, Tombstone>{};
+
+  DomElement createElementFor(int nodeId) {
+    final Tombstone? tombstone = _tombstones.remove(nodeId);
+    if (tombstone != null) {
+      // The semantics node has returned, so the tombstone is no longer
+      // necessary. As I (yjbanov) was testing this (July 19, 2023), I did not
```
ditman, line 1801: Can this `createElementFor` happen without the `nodeId` tombstone being ever blurred (and thus scheduled for cleanup anyway)?

lib/web_ui/lib/src/engine/semantics/semantics.dart
```diff
@@ -1676,6 +1685,42 @@ class SemanticsObject {
...
+    element.setAttribute('aria-hidden', true);
+    element.setAttribute('flt-tombstone', '');
+    element.style.transform = 'translate(-1px, -1px)';
+    element.style.top = '';
+    element.style.left = '';
+    element.style.width = '0px';
+    element.style.height = '0px';
+
+    final DomEventListener blurListener = createDomEventListener((_) {
+      owner._tombstones.remove(id);
```
ditman, line 1709: Once that a tombstone is left, it seems to never be re-used, so do we really need to keep a map of `_tombstones` by `id` in the semantics `owner`? I say that once you schedule them for deletion after onBlur, nobody else cares about those objects anymore, right? (so no need to keep track of the Tombstone typedef for it to be eventually deleted, we just let it go onBlur)

lib/web_ui/lib/src/engine/semantics/semantics.dart
```diff
@@ -1833,7 +1916,16 @@ class EngineSemanticsOwner {
...
-        removal.element.remove();
+
+        // This is the trick used to prevent the screen reader from focusing on
+        // the <body> tag
+        if (removal.element != focusedElement) {
+          // Element is not focused: safe to remove from DOM.
+          removal.element.remove();
+        } else {
+          // Element is focused: not safe to remove from DOM. Leave a tombstone.
+          removal.leaveTombstone();
```
ditman, line 1927: `leaveTombstone` doesn't describe very well what the method does. What about `.removeOnBlur()`?

ditman (commented):
This looks good to me, but I think the whole `tyedef Tombstone` can be removed? 

Since we never attempt to revive a Tombstone if focus for the same nodeId is received, I don't think there's a need to keep the Tombstone typedef around (after the removal of the actual node has already been scheduled onBlur).

lib/web_ui/lib/src/engine/semantics/semantics.dart
mdebbar: This is worth a try IMHO. Another aspect to consider here is whether it's even possible to have more than one tombstone at a time?

Consider the following sequence of events:
1. Node1 disposed ==> 1st tombstone.
2. Node2 focused ==> we can safely remove the 1st tombstone now.
3. Node2 disposed ==> 2nd tombstone (but at this point, the 1st tombstone is gone).

If my understanding is correct, we don't need a map of tombstones, we just need the `Tombstone? _lastTombstone` (if at all).

Hixie:
As of September 12, @yjbanov assures me this is still on his radar.

goderbauer:
@yjbanov again said this is still on his radar.

yjbanov:
On my radar it still is.

PR #44029, opened 2023-07-26
[Impeller] Only strip the translation from backdrop filter effect transforms.

bdero:
Resolves https://github.com/flutter/flutter/issues/131305.


jonahwilliams (approved):
LGTM

PR #44051, opened 2023-07-26
Change a raw string in non-breaking way to fix syntax highlighting.

matanlurey:
This has no change on the semantics of the Python programming, but fixes syntax highlighting in VSCode:

**Before**:
Screenshot 2023-07-26 at 3 22 51 PM

**After**:
Screenshot 2023-07-26 at 3 23 06 PM

Happy to revert if this isn't worth landing.

### DIFF
--- a/tools/gn
+++ b/tools/gn
@@ -608,7 +608,7 @@ def to_gn_args(args):
 
   if args.use_glfw_swiftshader:
     if get_host_os() == 'mac':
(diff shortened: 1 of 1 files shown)


matanlurey:
Actually, this looked like a bug with GNLS, and isn't related to Python. I don't think it's worth fixing.

gaaclarke:
Weird, yea.   I'm pretty sure the python linter complains about double quotes?  I might be getting it mixed up with dart.

PR #44073, opened 2023-07-27
Hide some log lines from release builds

Hixie:
These lines are appearing in framework tests but do not appear to have significant diagnostic value.


Hixie:
test-exempt: essentially this is code removal, i guess?

jonahwilliams (approved):
LGTM

chinmaygarde (approved):

PR #44089, opened 2023-07-27
[Impeller] Avoid culling when the current matrix has perspective.

dnfield:
Fixes https://github.com/flutter/flutter/issues/130613

Before this patch, the test would fail to render anything because the culling would decide it fell outside the cull rect when the transform has perspective sometimes. We should fix our Rect::TransformBounds implementation, but I can't quite seem to get it happy enough here so I'm just bailing out on culling if there is perspective instead, which is safe. This also makes the patch a bit easier/safer to cherry pick since it's a simple de-optimization when perspective is involved for the sake of fidelity, instead of a larger change that may have other side effects.

Filed https://github.com/flutter/flutter/issues/131445 to track the perspective issue.

### DIFF
--- a/impeller/display_list/dl_dispatcher.cc
+++ b/impeller/display_list/dl_dispatcher.cc
@@ -1074,6 +1074,8 @@ void DlDispatcher::drawDisplayList(
     canvas_.SaveLayer(save_paint);
   }
 
+  // TODO(131445): Remove this restriction if we can correctly cull with
+  // perspective transforms.
   if (display_list->has_rtree() && !initial_matrix_.HasPerspective()) {
     // The canvas remembers the screen-space culling bounds clipped by
     // the surface and the history of clip calls. DisplayList can cull
--- a/impeller/display_list/dl_dispatcher.cc
+++ b/impeller/display_list/dl_dispatcher.cc
@@ -1074,7 +1074,7 @@ void DlDispatcher::drawDisplayList(
     canvas_.SaveLayer(save_paint);
   }
 
-  if (display_list->has_rtree()) {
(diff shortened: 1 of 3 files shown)


impeller/display_list/dl_dispatcher.cc
```diff
@@ -1074,7 +1074,7 @@ void DlDispatcher::drawDisplayList(
     canvas_.SaveLayer(save_paint);
   }
 
-  if (display_list->has_rtree()) {
+  if (display_list->has_rtree() && !initial_matrix_.HasPerspective()) {
```
bdero, line 1077: Nit: Add todo comment nearby linking to the new bug?

bdero (approved):
LGTM

impeller/geometry/matrix.h
```diff
@@ -313,6 +313,10 @@ struct Matrix {
             m[9] == 0 && m[10] == 1 && m[11] == 0 && m[14] == 0 && m[15] == 1);
   }
 
+  constexpr bool HasPerspective() const {
+    return m[3] != 0 || m[7] != 0 || m[11] != 0 || m[15] != 1;
```
flar, line 317: That should be safe. We may not care about 11 for the case at hand since it is the Z component of W and Z=0, but it can't hurt to bail on that for simplicity/robustness sake.

flar (approved):
LGTM, would like to see a comment pointing to a filed issue as well...

impeller/display_list/dl_dispatcher.cc
dnfield: Done

PR #44106, opened 2023-07-28
[CP][Impeller] Avoid culling when the current matrix has perspective.

dnfield:
Cherry pick of #44089

Before this patch, the test would fail to render anything because the culling would decide it fell outside the cull rect when the transform has perspective sometimes. We should fix our Rect::TransformBounds implementation, but I can't quite seem to get it happy enough here so I'm just bailing out on culling if there is perspective instead, which is safe. This also makes the patch a bit easier/safer to cherry pick since it's a simple de-optimization when perspective is involved for the sake of fidelity, instead of a larger change that may have other side effects.


flar (approved):
LGTM

chinmaygarde (approved):

PR #44116, opened 2023-07-28
Set the required Android API level when Vulkan validation layers are automatically enabled for android_debug_unopt_arm64 builds

jason-simmons:
The Vulkan validation layer headers will not build if the Android API level is 23 or lower.

Currently the engine buildroot's default Android API level is 23.  The API level must be overridden for any configuration where Vulkan validation is enabled.

### DIFF
--- a/tools/gn
+++ b/tools/gn
@@ -563,14 +563,16 @@ def to_gn_args(args):
   if args.fstack_protector:
     gn_args['use_fstack_protector'] = True
 
-  if args.enable_vulkan_validation_layers:
-    if args.target_os == 'android':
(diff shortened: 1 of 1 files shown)


matanlurey:
Thanks! Do you mind editing the PR description and add 1-2 sentences on what problem this solves for posterity?

matanlurey (approved):

jason-simmons:
> Thanks! Do you mind editing the PR description and add 1-2 sentences on what problem this solves for posterity?

done

matanlurey:
That makes sense, thanks for fixing this!

PR #44172, opened 2023-07-31
[Impeller] moved to one staging buffer pool

gaaclarke:
I measured the pool usage while using the wonderous app after having consolidated the pools by printing capturing the pool usage after every call of `vmaCreateBuffer`.  The max number of allocations ever seen in this pool were 1288704 B.  Each of these pools are 32 MB.  So this change removes 64MB from the baseline memory usage of a Vulkan Flutter engine.  Worse case for a swapchain of 3 would be roughly 3 * 1.2 MB which is well within the limits of the 32 MB buffer.

Using one pool also increases the likelihood that a buffer will have been more recently used which leads to better memory performance.

This is safe to do since buffers will only be placed back into the pool after they are disposed of.

## patch used for measuring
```diff
--- a/impeller/renderer/backend/vulkan/allocator_vk.cc
+++ b/impeller/renderer/backend/vulkan/allocator_vk.cc
@@ -474,6 +474,15 @@ std::shared_ptr<DeviceBuffer> AllocatorVK::OnCreateBuffer(
                                              &buffer_allocation_info  //
                                              )};
 
+  VmaStatistics pool_stats;
+  ::vmaGetPoolStatistics(allocator_.get(), staging_buffer_pool_.get().pool,
+                         &pool_stats);
+  static uint64_t s_max = 0;
+  if (pool_stats.allocationBytes > s_max) {
+    s_max = pool_stats.allocationBytes;
+    FML_LOG(ERROR) << "pool max: " << s_max;
+  }
+
   if (result != vk::Result::eSuccess) {
     VALIDATION_LOG << "Unable to allocate a device buffer: "
                    << vk::to_string(result);
```

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/impeller/renderer/backend/vulkan/allocator_vk.cc
+++ b/impeller/renderer/backend/vulkan/allocator_vk.cc
@@ -73,8 +73,7 @@ static PoolVMA CreateBufferPool(VmaAllocator allocator) {
 
   VmaPoolCreateInfo pool_create_info = {};
   pool_create_info.memoryTypeIndex = memTypeIndex;
-  pool_create_info.flags = VMA_POOL_CREATE_IGNORE_BUFFER_IMAGE_GRANULARITY_BIT |
-                           VMA_POOL_CREATE_LINEAR_ALGORITHM_BIT;
+  pool_create_info.flags = VMA_POOL_CREATE_IGNORE_BUFFER_IMAGE_GRANULARITY_BIT;
 
   VmaPool pool = {};
   result = vk::Result{::vmaCreatePool(allocator, &pool_create_info, &pool)};
--- a/impeller/renderer/backend/vulkan/allocator_vk.cc
+++ b/impeller/renderer/backend/vulkan/allocator_vk.cc
@@ -161,10 +161,8 @@ AllocatorVK::AllocatorVK(std::weak_ptr<Context> context,
     VALIDATION_LOG << "Could not create memory allocator";
     return;
   }
-  for (size_t i = 0u; i < staging_buffer_pools_.size(); i++) {
-    staging_buffer_pools_[i].reset(CreateBufferPool(allocator));
-    created_buffer_pools_ &= staging_buffer_pools_[i].is_valid();
-  }
+  staging_buffer_pool_.reset(CreateBufferPool(allocator));
+  created_buffer_pool_ &= staging_buffer_pool_.is_valid();
   allocator_.reset(allocator);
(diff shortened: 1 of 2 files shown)


jonahwilliams (changes requested):
You need to remove the Linear allocted bit from the pool or else there is a chance to just endlessly leak memory (as allocation only happens at the end).

jonahwilliams:
Specifically: https://github.com/flutter/engine/blob/main/impeller/renderer/backend/vulkan/allocator_vk.cc#L77

gaaclarke:
> You need to remove the Linear allocted bit from the pool or else there is a chance to just endlessly leak memory (as allocation only happens at the end).

Done.  If we wanted to have a more optimal pool we could look into a ring buffer.  Since in practice i'm seeing about 1/32 of the pool being used it's probably not hard for the algorithm to find space for new allocations.

jonahwilliams:
ring buffer isn't something we can use for this, only VMA or Vulkan supported allocation strategies are reasonable.

jonahwilliams (approved):
LGTM

gaaclarke:
> ring buffer isn't something we can use for this, only VMA or Vulkan supported allocation strategies are reasonable.

The VMA documentation makes mention of using a ring buffer: https://gpuopen-librariesandsdks.github.io/VulkanMemoryAllocator/html/custom_memory_pools.html#linear_algorithm

I've read through it a few times and I'm guessing you have to implement it yourself on top of the pool?  It isn't very clear.

gaaclarke:
> That would be a ring buffer of VMA pools, each with a linear allocated bit set. That is the current implementation, which you are removing 😄

It's not that because it says the requirement is that the `blockCount == 1`, so it is talking about one pool acting as a ring buffer.

jonahwilliams:
Ahh you got me! I think that might work except, what happens if we run out of memory?

gaaclarke:
> Ahh you got me! I think that might work except, what happens if we run out of memory?

We wouldn't in the ring buffer case because we are limited by the swapchain images we have.  At worst case we would have swapchain_number of resources in memory.  When the last frame is rendered all of that space will be given back to the ring.  When writing overlaps reading in a ring buffer its a buffer overrun and since in practice it looks like we are using at most 1.2 MB, lets say thats per frame, we'd never get close to that point with a 32MB ring.

jonahwilliams:
There is no bound on the amount of memory we can use per frame. A developer could create a picture with thousands of paths that leads to many MB of geometry. Anything is possible.

PR #44184, opened 2023-07-31
[Windows] Delay enabling app lifecycle states until requested

yaakovschectman:
Await a platform message before sending lifecycle state updates so we are not sending messages that do not get consumed by the framework.

https://github.com/flutter/flutter/issues/131616

## Pre-launch Checklist

- [ ] I signed the [CLA].

### DIFF
--- a/shell/platform/windows/fixtures/main.dart
+++ b/shell/platform/windows/fixtures/main.dart
@@ -128,6 +128,21 @@ void exitTestCancel() async {
 }
 
 @pragma('vm:entry-point')
+void enableLifecycleTest() async {
+  final Completer<ByteData?> finished = Completer<ByteData?>();
+  ui.channelBuffers.setListener('flutter/lifecycle', (ByteData? data, ui.PlatformMessageResponseCallback callback) async {
+    if (data != null) {
+      ui.PlatformDispatcher.instance.sendPlatformMessage(
+        'flutter/unittest',
+        data,
+        (ByteData? reply) {
+        });
+    }
+  });
+  await finished.future;
+}
+
+@pragma('vm:entry-point')
 void customEntrypoint() {}
 
 @pragma('vm:entry-point')
--- a/shell/platform/windows/flutter_windows_engine.cc
+++ b/shell/platform/windows/flutter_windows_engine.cc
@@ -158,8 +158,7 @@ FlutterLocale CovertToFlutterLocale(const LanguageInfo& info) {
 
 FlutterWindowsEngine::FlutterWindowsEngine(const FlutterProjectBundle& project)
     : project_(std::make_unique<FlutterProjectBundle>(project)),
-      aot_data_(nullptr, nullptr),
-      lifecycle_manager_(std::make_unique<WindowsLifecycleManager>(this)) {
+      aot_data_(nullptr, nullptr) {
   embedder_api_.struct_size = sizeof(FlutterEngineProcTable);
   FlutterEngineGetProcAddresses(&embedder_api_);
 
@@ -230,6 +229,7 @@ FlutterWindowsEngine::FlutterWindowsEngine(const FlutterProjectBundle& project)
       std::make_unique<PlatformHandler>(messenger_wrapper_.get(), this);
   settings_plugin_ = std::make_unique<SettingsPlugin>(messenger_wrapper_.get(),
                                                       task_runner_.get());
+  lifecycle_manager_ = std::make_unique<WindowsLifecycleManager>(messenger_wrapper_.get(), this);
 }
 
 FlutterWindowsEngine::~FlutterWindowsEngine() {
--- a/shell/platform/windows/flutter_windows_engine.cc
+++ b/shell/platform/windows/flutter_windows_engine.cc
@@ -826,6 +826,9 @@ bool FlutterWindowsEngine::SendAppLifecycleStateUpdate(
 
 void FlutterWindowsEngine::SetAppLifecycleStateEnabled(bool enabled) {
   app_lifecycle_enabled_ = enabled;
+  if (enabled) {
+    SendAppLifecycleStateUpdate(lifecycle_manager_->GetLifecycleState());
+  }
 }
 
 }  // namespace flutter
--- a/shell/platform/windows/flutter_windows_engine.cc
+++ b/shell/platform/windows/flutter_windows_engine.cc
@@ -229,7 +229,8 @@ FlutterWindowsEngine::FlutterWindowsEngine(const FlutterProjectBundle& project)
       std::make_unique<PlatformHandler>(messenger_wrapper_.get(), this);
   settings_plugin_ = std::make_unique<SettingsPlugin>(messenger_wrapper_.get(),
                                                       task_runner_.get());
-  lifecycle_manager_ = std::make_unique<WindowsLifecycleManager>(messenger_wrapper_.get(), this);
+  lifecycle_manager_ =
+      std::make_unique<WindowsLifecycleManager>(messenger_wrapper_.get(), this);
 }
 
 FlutterWindowsEngine::~FlutterWindowsEngine() {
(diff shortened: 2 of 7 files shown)


shell/platform/windows/fixtures/main.dart
```diff
@@ -127,6 +127,21 @@ void exitTestCancel() async {
...
+  ui.channelBuffers.setListener('flutter/lifecycle', (ByteData? data, ui.PlatformMessageResponseCallback callback) async {
+    if (data != null) {
+      ui.PlatformDispatcher.instance.sendPlatformMessage(
+        'flutter/unittest',
+        data,
+        (ByteData? reply) {
+        });
+    }
+  });
+  await finished.future;
```
cbracken, line 142: It looks like you're never completing this future. If that's intentional, add a comment as to why.

cbracken: It might actually be useful to describe the back and forth being done over the channels to verify the handling/enabling in a comment in the C++ side of the unit tests, then maybe a one-liner here pointing readers to that test.

shell/platform/windows/windows_lifecycle_manager.cc
```diff
@@ -13,8 +13,38 @@
...
+            if (itr == arguments.MemberEnd() || !itr->value.IsBool()) {
+              result->Error(kLifecycleError, kInvalidArgumentMessage);
+              return;
+            }
+            engine_->SetAppLifecycleStateEnabled(itr->value.GetBool());
+            result->Success();
+            return;
+          }
+          result->NotImplemented();
+        });
```
loic-sharma, line 45: Should we consolidate all of the `flutter/lifecycle` platform channel stuffs to a lifecycle plugin? Specifically:

1. Listening to `flutter/lifecycle`'s `setLifecycleStateEnabled`
2. Sending `flutter/lifecycle` application states

shell/platform/windows/flutter_windows_engine.cc
```diff
@@ -812,4 +813,22 @@ std::optional<LRESULT> FlutterWindowsEngine::ProcessExternalWindowMessage(
   return std::nullopt;
 }
 
+bool FlutterWindowsEngine::SendAppLifecycleStateUpdate(
+    AppLifecycleState state) {
+  if (!app_lifecycle_enabled_) {
```
loic-sharma, line 818: What's the benefit of storing `app_lifecycle_enabled_` state in `FlutterWindowsEngine`?

Would it be preferrable to store it in the `WindowsLifecycleManager`? That already stores whether to handle application exits.

shell/platform/windows/flutter_windows_engine.h
```diff
@@ -274,6 +274,10 @@ class FlutterWindowsEngine {
                                                       WPARAM wparam,
                                                       LPARAM lparam);
 
+  bool SendAppLifecycleStateUpdate(AppLifecycleState state);
+
+  void SetAppLifecycleStateEnabled(bool enabled);
```
loic-sharma, line 279: Please add quick comments to any new members

loic-sharma:
Please don't merge this pull request until _after_ the corresponding framework changes have landed (doing the reverse would regress Windows's app lifecycle messages).

Also, how are other platforms affected? What's our plan to update them if needed?

yaakovschectman:
@loic-sharma Other platforms should not be affected if they have been working so far. However, merging the framework changes first would mean in the meantime we would introduce a message from the framework to the engine that is not consumed and result in a new error. Is this preferable?

shell/platform/windows/windows_lifecycle_manager.cc
yaakovschectman: What logic is part of the engine currently other than deferring to calls in the lifecycle manager that you are referring to?

cbracken:
> However, merging the framework changes first would mean in the meantime we would introduce a message from the framework to the engine that is not consumed and result in a new warning.

Keep in mind that there are also quite a few third-party embedders, so whichever approach is taken, please make sure that nothing breaks for embedders that haven't implemented lifecycle events, if at all possible. Otherwise, we'll want to send out a notification to embedder authors as a heads-up.

yaakovschectman:
One thing I have noticed is that the location in the framework where we would send the message to enable the lifecycle states is also where we already send a message to enable graceful app exit. We may be able to leave the framework as-is and enable the lifecycle states using the same message.

shell/platform/windows/windows_lifecycle_manager.cc
loic-sharma: I'm referring to just the message serialization/deserialization plus the platform channel  sending/receiving here, not the business logic that decides when to send messages or how to handle received messages.

Right now, serializing + sending `flutter/lifecycle` messages is done by the `FlutterWindowsEngine`, and receiving + deserializing `flutter/lifecycle` messages is done by the `WindowsLifecycleManager`. It would be preferable if this was consolidated. I'd be fine with consolidating this in `WindowsLifecycleManager`. If for whatever reason this isn't feasible, I'd consider creating a lifecycle plugin for the `flutter/lifecycle` message serialization/deserialization + platform channel sending/receiving.

loic-sharma:
> However, merging the framework changes first would mean in the meantime we would introduce a message from the framework to the engine that is not consumed and result in a new warning. Is this preferable?

Hm that's a good point. We don't want any warnings, so I agree this is not preferable. 

FYI, [Android, iOS, web, macOS, and Linux](https://github.com/search?q=repo%3Aflutter%2Fengine%20%22flutter%2Flifecycle%22&type=code) also send `flutter/lifecycle` messages. Our solution should consider these platforms as well.

> We may be able to leave the framework as-is and enable the lifecycle states using the same message.

I'm not opposed to this, but I think it'd be better if we didn't re-use `flutter/system`'s `System.initializationComplete` to enable `flutter/lifecycle` messages: an app that doesn't use the widgets layer might want to opt-in to only app lifecycle messages, but not system messages like low memory pressure or application exit.

A potential alternative idea: could embedders be notified when a listener is added to a platform channel? This would remove the need for sending initialized messages.

yaakovschectman:
If other platforms have app lifecycle working without needing to wait for explicit enabling like Windows (because of the warning message), they can just ignore the enable message from the framework. Alternatively, we could send that message only when on Windows.

>A potential alternative idea: could embedders be notified when a listener is added to a platform channel? This would remove the need for sending initialized messages.

Perhaps, though I'm unsure how off the top of my head, and I think this would be a broader and deeper change. Perhaps this can be a future goal?

loic-sharma:
> Alternatively, we could send that message only when on Windows.

What if a third-party embedder needs a similar workaround? Would we need to accept a patch to add similar logic for that third-party embedder? This feels hacky, I would avoid this if possible.

> If other platforms have app lifecycle working without needing to wait for explicit enabling like Windows (because of the warning message), they can just ignore the enable message from the framework.

How would you address [Chris's concern](https://github.com/flutter/engine/pull/44184#issuecomment-1660700419) of not breaking third-party embedders that haven't implemented the lifecycle plugin?

yaakovschectman:
Do you have some idea of how we would be able to notify the embedder of a platform channel being registered? The first thing that would come to my mind is sending a platform message when the framework calls `setMessage/MethodCallHandler`, but that seems to be what you're suggesting we ought to avoid.

loic-sharma:
> The "breaking" that would occur would be a warning message that there was an unhandled platform message, correct? Is this any different than any other platform messages we've added in the past?

It's different in that the impact is higher. Normally, these warnings happen if an app uses a feature that's not supported by the current embedder - however, in this case this warning happens for _all_ apps until the embedder is updated.

> Do you have some idea of how we would be able to notify the embedder of a platform channel being registered? The first thing that would come to my mind is sending a platform message when the framework calls `setMessage/MethodCallHandler`, but that seems to be what you're suggesting we ought to avoid.

If we were to go with this approach, I would recommend using the embedder API instead and adding a new callback to [`FlutterProjectArgs`](https://github.com/flutter/engine/blob/92125f5b4dc2ff8a76fd4d5b599d1f5645eb07d5/shell/platform/embedder/embedder.h#L2121).

Perhaps you, Chris, and I should have a quick call to discuss this problem?

PR #44185, opened 2023-07-31
Manual roll of Dart with fixups to zircon system.dart

a-siva:
https://dart.googlesource.com/sdk.git/+log/4fd35600b477..0e02245635ea

Merge to Dart Version 3.2.0-16.0.dev


shell/platform/fuchsia/dart-pkg/zircon/lib/src/system.dart
```diff
@@ -86,7 +86,8 @@ class ReadResult extends _Result {
   /// Returns the bytes as a Uint8List. If status != OK this will throw
   /// an exception.
   Uint8List bytesAsUint8List() {
-    return _bytes!.buffer.asUint8List(_bytes!.offsetInBytes, _numBytes!);
+    final lbytes = _bytes!;
```
jason-simmons, line 89: Use the `bytes` getter (which returns a non-nullable `ByteData`) that is already available in `ReadResult` and `ReadEtcResult`

a-siva: Done.

shell/platform/fuchsia/dart-pkg/zircon/lib/src/system.dart
```diff
@@ -86,7 +86,8 @@ class ReadResult extends _Result {
   /// Returns the bytes as a Uint8List. If status != OK this will throw
   /// an exception.
   Uint8List bytesAsUint8List() {
-    return _bytes!.buffer.asUint8List(_bytes!.offsetInBytes, _numBytes!);
+    final lbytes = bytes;
+    return lbytes.buffer.asUint8List(lbytes.offsetInBytes, _numBytes!);
```
jason-simmons, line 90: Remove `lbytes` and replace it with `bytes`

a-siva: wouldn't that be a little less efficient with two getter calls involving null checks?

jason-simmons: Either is fine - not sure if it matters for performance

jason-simmons (approved):

shell/platform/fuchsia/dart-pkg/zircon/lib/src/system.dart
a-siva: Added @stereotype441 for his take on which form would be better.

stereotype441: I don't have a lot to add about which form would be better. @a-siva is right that two getter calls involving null checks might in principle be slightly worse in performance. Then again, it's possible that due to inlining, the two null checks will be recognized by an optimization pass as redundant, in which case it might not matter at all. It's also possible that this is simply not a hot enough code path to have a noticeable effect.

So lgtm either way.

stereotype441 (approved):

PR #44188, opened 2023-07-31
[web:a11y] add platform view role

yjbanov:
Add `PlatformViewRoleManager`, the primary role manager for platform views. Currently, all it does is manager the `aria-owns` attribute that determines the screen reader traversal order of the platform view w.r.t. surrounding content.

This is a partial fix for https://github.com/flutter/flutter/issues/124765. While it does not address literally using the TAB key as a means for traversing widgets, it does address traversal via screen reader shortcuts.

### DIFF
--- a/lib/web_ui/lib/src/engine/platform_views/content_manager.dart
+++ b/lib/web_ui/lib/src/engine/platform_views/content_manager.dart
@@ -108,7 +108,7 @@ class PlatformViewManager {
   /// The resulting DOM for the `contents` of a Platform View looks like this:
   ///
   /// ```html
-  /// <flt-platform-view id="flt-platform-view-VIEW_ID" slot="...">
+  /// <flt-platform-view id="flt-pv-VIEW_ID" slot="...">
   ///   <arbitrary-html-elements />
   /// </flt-platform-view-slot>
   /// ```
--- a/lib/web_ui/lib/src/engine/platform_views/content_manager.dart
+++ b/lib/web_ui/lib/src/engine/platform_views/content_manager.dart
@@ -108,7 +108,7 @@ class PlatformViewManager {
   /// The resulting DOM for the `contents` of a Platform View looks like this:
   ///
   /// ```html
-  /// <flt-platform-view slot="...">
+  /// <flt-platform-view id="flt-platform-view-VIEW_ID" slot="...">
   ///   <arbitrary-html-elements />
   /// </flt-platform-view-slot>
   /// ```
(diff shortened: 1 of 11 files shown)


lib/web_ui/lib/src/engine/platform_views/content_manager.dart
```diff
@@ -108,7 +108,7 @@ class PlatformViewManager {
   /// The resulting DOM for the `contents` of a Platform View looks like this:
   ///
   /// ```html
-  /// <flt-platform-view slot="...">
+  /// <flt-platform-view id="flt-platform-view-VIEW_ID" slot="...">
```
ditman, line 111:
```suggestion
  /// <flt-platform-view id="flt-pv-VIEW_ID" slot="...">
```

(According to https://github.com/flutter/engine/pull/44188/files#diff-0ee07db46a6272991cd49c1bace8240dc5efaa8a0efabbbfc3f80722589653f1R19)

lib/web_ui/lib/src/engine/semantics/platform_view.dart
```diff
@@ -0,0 +1,42 @@
...
+///   * https://bugs.webkit.org/show_bug.cgi?id=223798
+class PlatformViewRoleManager extends PrimaryRoleManager {
+  PlatformViewRoleManager(SemanticsObject semanticsObject)
+      : super.withBasics(PrimaryRole.platformView, semanticsObject);
+
+  @override
+  void update() {
+    super.update();
+
+    if (semanticsObject.isPlatformView) {
```
ditman, line 31: So we already had `isPlatformView` but it was never used?

lib/web_ui/test/engine/platform_views/slots_test.dart
```diff
@@ -35,4 +35,12 @@ void testMain() {
...
     });
   });
+
+  test('getPlatformViewSlotName', () {
+    expect(getPlatformViewSlotName(42), 'flt-pv-slot-42');
+  });
+
+  test('getPlatformViewDomId', () {
+    expect(getPlatformViewDomId(42), 'flt-pv-42');
+  });
```
ditman, line 45: Good Boy Scout! Thanks!

ditman (approved):
Looks good! I hope this improves the navigation in most screen readers! (except for Safari)

lib/web_ui/lib/src/engine/platform_views/content_manager.dart
yjbanov: Good catch! I initially named it "platform-view", but then found the slot name and decided to be consistent with that, but forgot to update the docs.

lib/web_ui/lib/src/engine/semantics/platform_view.dart
yjbanov: It is used elsewhere to set proper pointer event handling, but that aspect is not handled by this role manager since the logic is applied to all roles.

PR #44228, opened 2023-08-01
Improve logging in the clang-tidy script

jason-simmons:
* Add timestamps to process pool status logs
* Log the remaining job names when only a few jobs are pending

See https://github.com/flutter/flutter/issues/131689

### DIFF
--- a/tools/clang_tidy/lib/clang_tidy.dart
+++ b/tools/clang_tidy/lib/clang_tidy.dart
@@ -89,8 +89,12 @@ class ClangTidy {
   final StringSink _outSink;
   final StringSink _errSink;
 
+  late final DateTime _startTime;
+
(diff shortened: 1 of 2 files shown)


tools/clang_tidy/lib/clang_tidy.dart
```diff
@@ -89,8 +89,12 @@ class ClangTidy {
   final StringSink _outSink;
   final StringSink _errSink;
 
+  DateTime? _startTime;
```
matanlurey, line 92: Could probably be `late DateTime _startTime` since that's how it's used.

jason-simmons: done

zanderso (approved):

PR #44248, opened 2023-08-02
[Impeller] DlCanvas implementation wrapping Aiks canvas

dnfield:
Fixes https://github.com/flutter/flutter/issues/130141

The primary goal of this patch is to move dispatching of `dart:ui` `Canvas` commands to the UI thread.

Before this patch, the architecture is something like:

## UI Thread
- `dart:ui` talks to `DisplayListBuilder`, a `DlCanvas` implementation.
- `DisplayListBuilder` does some clip/bounds tracking and creates a `DisplayList` object that is held by `dart:ui`'s `Picture` objects.
- `DisplayList`s are added to `DisplayListLayer`s in `flow`. 

## Raster Thread
- `flow` flattens the various operations into a single `DisplayList` via another `DisplayListBuilder`.
- A `DlOpReceiver`implementation converts that `DisplayList` into an `Aiks` `Canvas`/`Picture`.

After this patch, the architecture instead looks like:

## UI Thread

- No change for Skia.
- If Impeller, use a new `DlCanvasImplementation` that talks to `Aiks`'s `Canvas`.
- If Impeller, `dart:ui` Picture's now hold an `Aiks` `Picture`, which get shared into `AiksLayer`s in `flow`.

## Raster thread

- No change for Skia, but some light refactoring for places that assumed a `DisplayListBuilder` where they really just needed a `DlCanvas`.
- The `Aiks` `Picture`s are combined using new API on `DlCanvas` and still backed by `Aiks`.

These changes show significant improvement on raster times on Android and only very small regressions on UI times in local testing, see https://gist.github.com/dnfield/26528090194c9f5abdbac13cdcbf4f79 for old gallery transition perf numbers.

Many of the other changes in this patch are related to the following:

- Making `DlRTree` usable for Impeller.
  - It would be nice to have a version of DlRTree that speaks `impeller::Rect`.
- Creating the requisite classes to support `EmbeddedViews` so that Desktop works.

This patch does not remove the `impeller::DlDispatcher`, which now would only be used in tests.


dnfield:
Blocked on https://github.com/flutter/flutter/issues/131887

dnfield:
I have patched in the fix from https://github.com/flutter/engine/pull/44346, as this patch was causing those iOS tests to start crashing since it adds some DCHECKs around Impeller-related codepaths that were now firing due to the usage of software rendering in those tests.

dnfield:
I should also add that my performance numbers were taken from an earlier iteration of this patch and may look slightly different now, but should still be indicative.

flow/layers/aiks_layer.cc
```diff
@@ -0,0 +1,67 @@
...
+void AiksLayer::Diff(DiffContext* context, const Layer* old_layer) {
+  DiffContext::AutoSubtreeRestore subtree(context);
+  context->PushTransform(SkMatrix::Translate(offset_.x(), offset_.y()));
+  context->AddLayerBounds(bounds_);
+  context->SetLayerPaintRegion(this, context->CurrentSubtreeRegion());
+}
+
+void AiksLayer::Preroll(PrerollContext* frame) {
+  // TODO(dnfield): Handle group opacity checks.
+  set_paint_bounds(bounds_);
```
jonahwilliams, line 43: Link to an issue?

flow/layers/aiks_layer.cc
```diff
@@ -0,0 +1,67 @@
...
+  FML_DCHECK(!context.raster_cache);
+
+  SkScalar opacity = context.state_stack.outstanding_opacity();
+
+  if (context.enable_leaf_layer_tracing) {
+    // TODO(dnfield): Decide if we need to capture this for Impeller.
+    // We can't do this the same way as on the Skia backend, because Impeller
+    // does not expose primitives for flushing things down to the GPU without
+    // also allocating a texture.
+    FML_LOG(ERROR) << "Leaf layer tracing unsupported for Impeller.";
```
jonahwilliams, line 61: Lets file an issue to discuss, but my inclination is that we should not support this feature for Impeller and should instead direct folks to either native profiling tools or something based on archivist.

flow/layers/layer_tree.cc
```diff
@@ -155,6 +155,69 @@ void LayerTree::Paint(CompositorContext::ScopedFrame& frame,
...
+    const std::shared_ptr<TextureRegistry>& texture_registry) {
+#if IMPELLER_SUPPORTS_RENDERING
+  TRACE_EVENT0("flutter", "LayerTree::FlattenToImpellerPicture");
+
+  impeller::DlAiksCanvas canvas(bounds);
+
+  const FixedRefreshRateStopwatch unused_stopwatch;
+
+  LayerStateStack preroll_state_stack;
+  // No root surface transformation. So assume identity.
```
jonahwilliams, line 169: I think Vulkan prerotation may eventually give us a root surface transformation.

impeller/display_list/dl_aiks_canvas.cc
```diff
@@ -0,0 +1,1191 @@
...
+#include "impeller/entity/geometry/geometry.h"
+#include "impeller/geometry/path.h"
+#include "impeller/geometry/path_builder.h"
+#include "impeller/geometry/scalar.h"
+#include "impeller/geometry/sigma.h"
+#include "impeller/typographer/backends/skia/text_frame_skia.h"
+
+namespace impeller {
+
+#define UNIMPLEMENTED \
```
jonahwilliams, line 42: This is more or less the same implementation as dl_dispatcher right? if this sticks the landing we should clean that up quickly so they don't diverge.

impeller/tools/impeller.gni
```diff
@@ -12,13 +12,15 @@ declare_args() {
...
 
   # Whether the OpenGLES backend is enabled.
-  impeller_enable_opengles = is_linux || is_win || is_android
+  impeller_enable_opengles =
+      (is_linux || is_win || is_android) && target_os != "fuchsia"
 
   # Whether the Vulkan backend is enabled.
-  impeller_enable_vulkan = is_linux || is_win || is_android
+  impeller_enable_vulkan =
+      (is_linux || is_win || is_android) && target_os != "fuchsia"
```
jonahwilliams, line 24: Was this necessary? I don't remember where we landed on the embedder.h issue with fuchsia

lib/ui/compositing/scene_builder.cc
```diff
@@ -226,6 +227,15 @@ void SceneBuilder::addPicture(double dx,
...
         !!(hints & 1), !!(hints & 2));
     AddLayer(std::move(layer));
+    return;
+  }
+
+  auto impeller_picture = picture->impeller_picture();
+  if (impeller_picture) {
+    auto layer = std::make_unique<flutter::AiksLayer>(
+        SkPoint::Make(SafeNarrow(dx), SafeNarrow(dy)),
+        std::move(impeller_picture), !!(hints & 1), !!(hints & 2));
```
jonahwilliams, line 237: Lets just drop the hints since we don't use them for anything.

lib/ui/painting/picture.cc
```diff
@@ -79,66 +81,126 @@ static sk_sp<DlImage> CreateDeferredImage(
...
 size_t Picture::GetAllocationSize() const {
-  if (display_list_) {
-    return display_list_->bytes() + sizeof(Picture);
-  } else {
-    return sizeof(Picture);
+  auto size = sizeof(Picture);
+  if (display_list()) {
+    size += display_list()->bytes();
   }
+  // TODO(dnfield): Add support to EntityPass to get its allocation size.
```
jonahwilliams, line 115: What are we using the allocation size for? Is this necesary to support?

jonahwilliams (commented):
are there any features from the dlbuilder that we're losing from this switch?

flow/layers/aiks_layer.cc
dnfield: I'm going to delete this comment.

In `DisplayListLayer::Preroll`, there's opacity related checking that happens to make decisions about raster caching. Raster caching isn't a thing for Impeller, and Impeller internally handles its own opacity peepholing.

dnfield: (adding a comment about that and a dcheck that there's no raster cache)

flow/layers/aiks_layer.cc
dnfield: https://github.com/flutter/flutter/issues/131941

flow/layers/layer_tree.cc
dnfield: This comment is duplicated from `Flatten` below. But I guess we can handle that then? The idea is that this method would only ever be called on a `Scene` object that is not really attached to a particular surfacde.

jonahwilliams: Yeah, no action needed, just a comment

impeller/display_list/dl_aiks_canvas.cc
dnfield: It's similar but a little different. The two major differences are:

- It can't really draw a display list (and vice versa, a display list can't draw an impeller picture)
- It's not using RTrees to determine culling when concatenating Impeller pictures (but it probably could if needed).

If this patch lands and sticks we can go ahead and delete dl_dispatcher, which is only used in tests. I didn't want to make this already large patch bigger by doing that now though.

impeller/tools/impeller.gni
dnfield: If I don't do this, there are host artifacts built for linux or mac when building Fuchsia where the target_os is Fuchsia, and enable_impeller ends up being true in places where it really shouldn't be.

I'd be happy to separate this into another patch if it helps.

lib/ui/compositing/scene_builder.cc
dnfield: Done

impeller/display_list/dl_aiks_canvas.cc
jonahwilliams: Sounds good to me

lib/ui/painting/picture.cc
dnfield: It's exposed in https://api.flutter.dev/flutter/dart-ui/Picture/approximateBytesUsed.html

impeller/tools/impeller.gni
jonahwilliams: Nah thats fine.

lib/ui/painting/picture.cc
jonahwilliams: :(

entity count * 100?

jonahwilliams (approved):
overall LGTM, the devicelab will be the ultiamte arbiter of success thougj

jonahwilliams:
"ultimate" darn I can't spell

flow/layers/aiks_layer.cc
```diff
@@ -0,0 +1,61 @@
...
+#if IMPELLER_SUPPORTS_RENDERING
+  if (picture_) {
+    auto bounds = picture_->pass->GetElementsCoverage(std::nullopt)
+                      .value_or(impeller::Rect());
+    bounds_ = SkRect::MakeXYWH(bounds.origin.x + offset_.x(),
+                               bounds.origin.y + offset.y(), bounds.size.width,
+                               bounds.size.height);
+  }
+#endif
+}
```
flar, line 23: There is no override for IsReplacing. I forget the details, but sometimes a new <Picture>Layer is constructed with the same ui.Picture as was used in the previous frame and so we need to check if the pictures are the same, not just the layer, otherwise we lose a lot of partial repaint reductionl

flow/layers/aiks_layer.cc
flar: Why is this done on display_list_layer then?

flow/surface_frame.cc
```diff
@@ -31,9 +32,13 @@ SurfaceFrame::SurfaceFrame(sk_sp<SkSurface> surface,
     canvas_ = &adapter_;
   } else if (display_list_fallback) {
     FML_DCHECK(!frame_size.isEmpty());
-    dl_builder_ =
-        sk_make_sp<DisplayListBuilder>(SkRect::Make(frame_size), true);
-    canvas_ = dl_builder_.get();
+#if IMPELLER_SUPPORTS_RENDERING
```
flar, line 35: We should perhaps rename this flag if it really means "rendering_to_impeller" otherwise you are creating an impeller picture for a flag that is named "please create a display list"...

flar: This is also an odd combination of tests that implies that a given call can contain either or both of a surface and/or the flag. If it's "one or the other", then delete the flag and just assume it if there is no surface. If it is both, we aren't implementing it very well. If there are cases when neither is provided - why is that?

impeller/display_list/dl_aiks_canvas.cc
```diff
@@ -0,0 +1,1191 @@
...
+    accumulator_->save();
+  }
+}
+
+// |flutter::DlCanvas|
+void DlAiksCanvas::SaveLayer(const SkRect* bounds,
+                             const flutter::DlPaint* paint,
+                             const flutter::DlImageFilter* backdrop) {
+  if (!paint) {
+    return;
```
flar, line 689: This is a bug. Paint is optional, you can't just return. Perhaps we don't end up here with paint being null, but it doesn't invalidate the operation, it just means that the collected pixels are applied with default paint values.

impeller/display_list/dl_aiks_canvas.cc
```diff
@@ -0,0 +1,1191 @@
...
+}
+
+// |flutter::DlCanvas|
+void DlAiksCanvas::ClipRect(const SkRect& rect, ClipOp clip_op, bool is_aa) {
+  canvas_.ClipRect(skia_conversions::ToRect(rect), ToClipOperation(clip_op));
+}
+
+// |flutter::DlCanvas|
+void DlAiksCanvas::ClipRRect(const SkRRect& rrect, ClipOp clip_op, bool is_aa) {
+  if (rrect.isSimple()) {
```
flar, line 804: This is missing the "isRect" optimization.

impeller/display_list/dl_aiks_canvas.cc
```diff
@@ -0,0 +1,1191 @@
...
+namespace impeller {
+
+#define UNIMPLEMENTED \
+  FML_DLOG(ERROR) << "Unimplemented detail in " << __FUNCTION__;
+
+DlAiksCanvas::DlAiksCanvas(const SkRect& cull_rect, bool prepare_rtree)
+    : canvas_(skia_conversions::ToRect(cull_rect)) {
+  if (prepare_rtree) {
+    accumulator_ = std::make_unique<flutter::RTreeBoundsAccumulator>();
+  }
```
flar, line 49: else - a rect accumulator. We usually set the flag on most internal uses, but if anyone ever constructs the object with the flag set to false, it will just crash on the first Save call.

impeller/display_list/dl_aiks_canvas.cc
```diff
@@ -0,0 +1,1191 @@
...
+  if (rrect.isSimple()) {
+    canvas_.ClipRRect(skia_conversions::ToRect(rrect.rect()),
+                      rrect.getSimpleRadii().fX, ToClipOperation(clip_op));
+  } else {
+    canvas_.ClipPath(skia_conversions::ToPath(rrect), ToClipOperation(clip_op));
+  }
+}
+
+// |flutter::DlCanvas|
+void DlAiksCanvas::ClipPath(const SkPath& path, ClipOp clip_op, bool is_aa) {
```
flar, line 813: This is also missing isRect and isRRect optimizations.

impeller/display_list/dl_aiks_canvas.cc
```diff
@@ -0,0 +1,1191 @@
...
+#include "impeller/geometry/scalar.h"
+#include "impeller/geometry/sigma.h"
+#include "impeller/typographer/backends/skia/text_frame_skia.h"
+
+namespace impeller {
+
+#define UNIMPLEMENTED \
+  FML_DLOG(ERROR) << "Unimplemented detail in " << __FUNCTION__;
+
+DlAiksCanvas::DlAiksCanvas(const SkRect& cull_rect, bool prepare_rtree)
```
flar, line 45: This is all quite a bit different from what I thought was being done.

As originally described to me, I had thought that Impeller would continue to just have the dispatcher, but the conversion of DlCanvas to DlOpReceiver calls was going to happen independently and then that conversion would just talk to DlDispatcher directly.

As it happens, there are a few mistakes in this and duplication of code (or failure to duplicate important code) from DisplayListBuilder in addition to basically duplicating all of impeller's DlDispatcher.

What I though was going to happen was:

- currently DisplayListBuilder talks DlCanvas and optimizes the calls and then records them into a DL buffer it takes "DlCanvas" in, but it records "DlOpReceiver" calls on the back end
- Take the part of the Builder that does the optimizations and pull it out into a DisplayListCanvasOptimizer class that forwards the calls to a backend which implements DlOpReceiver
- Take the part of the Builder that records the Op calls into a buffer and pull that out into a DisplayListRecorder class that just records the calls into a buffer.
- DisplayListBuilder would simply link up a DLOptimizer with a DlRecorder and maintain the same functionality.
- For Impeller, DLOptimizer would talk to the existing impeller::DlDispatcher and it would get all of the optimizations for free without having to reinvent the wheel or create essentially a line-for-line replacement of DlDispatcher with, as we see here, missing optimizations and a couple of bugs.
- That's what I thought the effort was going to be, but I see now that I didn't hear that right.

impeller/display_list/dl_aiks_canvas.cc
flar: I would rather...

- Delete this new class which is a poor approximation of nearly identical code in DisplayListBuilder
- Break DisplayListBuilder out into a preprocessor that already does the optimizations, and a pluggable backend which already exists and is called impeller::DlDispatcher
- Basically the preprocessor output can either go to a recorder which is "OpReceiver in, OpReceiver out" or directly to an impeller DlDispatcher.

flow/layers/aiks_layer.cc
knopp: Indeed. The way it is now partial repaint will not work correctly with `AiksLayer`. With `DisplayListLayer` (and similarly `AiksLayer`) the diffing is moved one level up - to `ContainerLayer`.  The `DisplayListLayer::Diff` basically just reports the layer paint region.

The reason for this is that we want to be able to detect added / removed display list layers similar to how the framework diffs element children. (See `ContainerLayer::DiffChildren`).

knopp: The reason why picture layers are different than other layers in this regard is that picture layers don't exit in framework in a similar way other layers exist - so we don't get the old layer to link with the new layer when adding picture like we would get with a backdrop filter or a clip rect layer for example. So we relay on comparing the contents.

flar:
See also: https://github.com/flutter/engine/pull/44718 which can be attached to an impeller::DlDispatcher (which implements DlOpReceiver) to take in DlCanvas methods and more directly feed them to the Impeller renderer.

PR #44265, opened 2023-08-02
Relax JNI lookup for classes and methods that aren't available in API 22

johnmccutchan:
Relax JNI lookup for classes and methods that aren't available in API 22


chinmaygarde:
From Triage: @jason-simmons says this is now redundant. Closing. Please reopen if I messed up.

PR #44280, opened 2023-08-02
[Flutter GPU] Export symbols from engine, stub for testing on CI.

bdero:
Part of https://github.com/flutter/flutter/issues/131346

Stubs a minimal test of the FFI utilities that `dart:ui` uses, but using public symbols exported from the engine library. If this goes well, I'll move the stuff from `dart:ui` into here and begin landing parts of the API with test coverage.

### DIFF
--- a/testing/dart/gpu_test.dart
+++ b/testing/dart/gpu_test.dart
@@ -1,0 +1,20 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+// ignore_for_file: avoid_relative_lib_imports
+
+import 'dart:ui';
+
+import 'package:litetest/litetest.dart';
+import '../../lib/gpu/gpu.dart' as gpu;
+
+void main() {
+  test('no crash', () async {
+    final int result = gpu.testProc();
+    expect(result, 1);
+
+    final gpu.FlutterGpuTestClass a = gpu.FlutterGpuTestClass();
+    a.coolMethod(9847);
+  });
+}
--- a/testing/dart/gpu_test.dart
+++ b/testing/dart/gpu_test.dart
@@ -4,8 +4,6 @@
 
 // ignore_for_file: avoid_relative_lib_imports
 
-import 'dart:ui';
-
 import 'package:litetest/litetest.dart';
 import '../../lib/gpu/gpu.dart' as gpu;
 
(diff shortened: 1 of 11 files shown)


bdero:
@chinmaygarde @jonahwilliams @zanderso This is a silly PR with some exported symbols to make sure implementing the full API will actually be viable through public symbol exports. I'll probably move everything over in incremental PRs if this doesn't suck.

Definitely interested in feedback around patterns for exporting these symbols as well. We're separately solving the issue of getting the flutter_gpu code into the framework through this discussion: https://github.com/flutter/flutter/issues/131711

jonahwilliams:
I'm okay with exposing symbols as long as 1) we make a reasonable effort to ensure there won't be accidental collisions with other libraries 2) we document somewhere that these symbols do not constitute a public API

bdero:
@jonahwilliams I plan to prefix all of the exported symbols with `FlutterGpu` -- does that seem sufficient to avoid collisions? We could perhaps use more specialty names that look more dangerous to rely on. For example, prepending `Private`. Thoughts?

WRT documenting that they shouldn't be used, do you have any suggestions about where we could best do this? I don't see any perfect entry points. In addition to warnings around the C++ and Dart code, I could add a note to the custom embedder wiki page [linked here](https://docs.flutter.dev/embedded) since these users are likely to discover these symbols.

Perhaps starting a new wiki page for Dart GPU in general would be a good idea.

jonahwilliams:
FlutterGPU sounds like a good idea. You could go a bit further with the naming like InternalFlutterGPU or something that makes it more apparent it is not intended to be a public API (not sure what best practices are here). Starting a wiki page sounds like a good idea too.

jonahwilliams (approved):
LGTM

testing/dart/gpu_test.dart
```diff
@@ -0,0 +1,20 @@
...
+import 'dart:ui';
+
+import 'package:litetest/litetest.dart';
+import '../../lib/gpu/gpu.dart' as gpu;
+
+void main() {
+  test('no crash', () async {
+    final int result = gpu.testProc();
+    expect(result, 1);
+
```
zanderso, line 16: Does `testProcWithCallback` work?

zanderso (commented):
Overall approach seems fine. It might make sense to sprinkle some TODOs with links to GitHub issues in places where the code is temporary. Presubmit test failures are related.

testing/dart/gpu_test.dart
bdero: Ah yeah, added it in the test.

bdero:
Okay, almost there. Just need to figure out these remaining breakages.
- Sprinkled TODOs around.
- Fixed ref counting issue.
- Test the callback.
- Added a wiki page: https://github.com/flutter/flutter/wiki/Flutter-GPU

bdero:
Failures are looking unrelated.

zanderso (approved):

PR #44295, opened 2023-08-02
[CP] Remove windows_arm_host_engine from release builder

drewroengoogle:
Skip windows arm on release branches. This broke during the migration to the release builders.


XilaiZhang (approved):
images (1)

PR #44302, opened 2023-08-02
feat: support changing update behavior.

eseidel:
Fixes shorebirdtech/shorebird#716
Fixes shorebirdtech/shorebird#950

Depends on https://github.com/shorebirdtech/updater/pull/65 landing before I can include the DEPS change.


eseidel:
Sorry, wrong buttton.

PR #44303, opened 2023-08-02
[ios]make the screenIfViewLoaded and windowSceneIfLoaded helpers reusable

hellohuanlin:
The existing `screenIfViewLoaded` and `windowSceneIfLoaded` functions are private helpers of `FlutterViewController` class. This PR moves the logic to a category of the `UIViewController`, so it can be reused for any `UIViewController`s. 

https://github.com/flutter/engine/pull/43972#discussion_r1282214557

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/shell/platform/darwin/ios/framework/Source/UIViewController+FlutterScreenAndSceneIfLoaded.mm
+++ b/shell/platform/darwin/ios/framework/Source/UIViewController+FlutterScreenAndSceneIfLoaded.mm
@@ -1,0 +1,30 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#import "flutter/shell/platform/darwin/ios/framework/Source/UIViewController+FlutterScreenAndSceneIfLoaded.h"
+
+#include "flutter/fml/logging.h"
+
+@implementation UIViewController (FlutterScreenAndSceneIfLoaded)
+
+- (UIWindowScene*)flutterWindowSceneIfViewLoaded {
+  if (self.viewIfLoaded == nil) {
+    FML_LOG(WARNING) << "Trying to access the window scene before the view is loaded.";
+    return nil;
+  }
+  return self.viewIfLoaded.window.windowScene;
+}
+
+- (UIScreen*)flutterScreenIfViewLoaded {
+  if (@available(iOS 13.0, *)) {
+    if (self.viewIfLoaded == nil) {
+      FML_LOG(WARNING) << "Trying to access the screen before the view is loaded.";
+      return nil;
+    }
+    return [self flutterWindowSceneIfViewLoaded].screen;
+  }
+  return UIScreen.mainScreen;
+}
+
+@end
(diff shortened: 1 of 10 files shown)


shell/platform/darwin/ios/framework/Source/UIViewController+FlutterScreenAndSceneIfLoaded.mm
```diff
@@ -0,0 +1,30 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#import "flutter/shell/platform/darwin/ios/framework/Source/UIViewController+FlutterScreenAndSceneIfLoaded.h"
+
+#include "flutter/fml/logging.h"
+
+@implementation UIViewController (FlutterScreenAndSceneIfLoaded)
+
+- (UIWindowScene*)flutterWindowSceneIfViewLoaded {
```
mossmana, line 11: The name is long already and I'm not sure the value of adding `flutter`. Can we discuss leaving the name `windowSceneIfViewLoaded`? I think it would result in less change to FlutterViewController as well.

shell/platform/darwin/ios/framework/Source/UIViewController+FlutterScreenAndSceneIfLoaded.mm
```diff
@@ -0,0 +1,30 @@
...
+  if (self.viewIfLoaded == nil) {
+    FML_LOG(WARNING) << "Trying to access the window scene before the view is loaded.";
+    return nil;
+  }
+  return self.viewIfLoaded.window.windowScene;
+}
+
+- (UIScreen*)flutterScreenIfViewLoaded {
```
mossmana, line 19: Same question about naming convention.

mossmana (changes requested):
Thanks for doing this work. It was something I considered initially, but wanted to wait and see if there would be other cases where a shared utility could be used.

shell/platform/darwin/ios/framework/Source/UIViewController+FlutterScreenAndSceneIfLoaded.mm
hellohuanlin: Fun story about ObjC is that it has no namespace support. So if someone created a category with the same function name, and even if they import their own category header, it's indeterministic which function of the 2 categories will be called at run-time. (That's right, the category name is not a namespace!)

So to deal with this, the rule of thumb for defining categories of system types (e.g. UIViewController) is to avoid naming conflict of functions in the first place, by adding a prefix. 

I agree the name is a bit long. I was going to use flt_, but looks like flutter prefix is the convention.

shell/platform/darwin/ios/framework/Source/UIViewController+FlutterScreenAndSceneIfLoaded.mm
mossmana: I guess I'm not clear the difference between class method and global function in this case.

hellohuanlin: A global function would just work, but there are a few benefits of using categories in this particular case. Just from top of my head - 

1. it's more idiomatic objc code, and more consistent with Apple's convention (e.g. `vc.viewIfLoaded`, rather than `viewIfLoadedForVC(vc)`)

2. It can be stubbed for tests. Category methods use "message dispatch", meaning that the implementation can be swizzled out during run-time (basically how `OCMock` works internally)

mossmana (approved):
LGTM!

hellohuanlin:
> It was something I considered initially, but wanted to wait and see if there would be other cases where a shared utility could be used.

@mossmana Yep, I would do the same - previously we only needed them inside `FlutterViewController`, so it made sense to just keep them private helpers. 

Looks like there are some issues with license files update. Will fix them tomorrow.

cyanglaz (approved):
LGTM

PR #44304, opened 2023-08-02
Check whether the lookup of android.hardware.HardwareBuffer found a class

jason-simmons:


eyebrowsoffire (approved):
LGTM

jason-simmons:
One of the Firebase hosted test scenarios was still failing with this PR.

On the Firebase `griffin-24-en-portrait` device the `FindClass("android/hardware/HardwareBuffer")` call was returning success even though this is an API level 24 device.  The `GetMethodID("close")` call would then throw an exception.

The Firebase test passes with this change:
```
  if (!g_hardware_buffer_class->is_null()) {
    g_hardware_buffer_close_method =
        env->GetMethodID(g_hardware_buffer_class->obj(), "close", "()V");
  }
  if (g_hardware_buffer_close_method == nullptr) {
    FML_LOG(WARNING)
        << "Could not locate android.hardware.HardwareBuffer class";
    // Continue on as this class may not exist at API <= 26.
    fml::jni::ClearException(env);
  }
```

PR #44324, opened 2023-08-03
WIP: [impeller] Add tessellation cache

knopp:
Results for the [static_path_tessellation](https://github.com/flutter/flutter/pull/131837) macrobenchmark:

| average_frame_build_time_millis | Time |
|--|--
| Cache Disabled |  0.4710688405797106 |
| Cache Enabled | 0.46162815884476605 |

| average_frame_rasterizer_time_millis | Time |
|--|--
| Cache Disabled |  6.654328519855595 |
| Cache Enabled | 1.6611624548736454 |

I had to run the perf test with background thread spinning (see https://github.com/flutter/flutter/pull/131837) otherwise the variant with cache enabled actually regressed in `average_frame_build_time_millis` because iOS let the CPU run at lower speed.

The cache implementation is very straightforward, leverages the fact that `SkPath`s are retained by the framework and thus we can rely on `SkPath::getGenerationID()` to identify paths that are retained across frames. The cache size is bounded by  unused paths being evicted immediately at the end of each frame.

Both polylines and tessellation results are cached.

## Pre-launch Checklist

- [X] I signed the [CLA].

### DIFF
--- a/impeller/aiks/aiks_context.cc
+++ b/impeller/aiks/aiks_context.cc
@@ -5,6 +5,7 @@
 #include "impeller/aiks/aiks_context.h"
 
 #include "impeller/aiks/picture.h"
+#include "impeller/entity/contents/tessellation_cache.h"
 
 namespace impeller {
 
@@ -42,7 +43,11 @@ bool AiksContext::Render(const Picture& picture, RenderTarget& render_target) {
   }
 
   if (picture.pass) {
-    return picture.pass->Render(*content_context_, render_target);
+    auto res = picture.pass->Render(*content_context_, render_target);
+    // FIXME(knopp): This should be called for the last surface of the frame,
+    // but there's currently no way to do this.
+    content_context_->GetTessellationCache().FinishFrame();
+    return res;
   }
 
   return true;
--- a/impeller/display_list/skia_conversions.cc
+++ b/impeller/display_list/skia_conversions.cc
@@ -121,7 +121,9 @@ Path ToPath(const SkPath& path) {
   }
   builder.SetConvexity(path.isConvex() ? Convexity::kConvex
                                        : Convexity::kUnknown);
-  return builder.TakePath(fill_type);
+  auto result = builder.TakePath(fill_type);
+  result.SetOriginalGenerationId(path.getGenerationID());
+  return result;
 }
 
 Path ToPath(const SkRRect& rrect) {
--- a/impeller/display_list/skia_conversions.cc
+++ b/impeller/display_list/skia_conversions.cc
@@ -9,6 +9,12 @@
 namespace impeller {
 namespace skia_conversions {
 
+namespace {
+const uint32_t kTagSkiaPath = 'skpa';
+const uint32_t kTagSkiaRRect = 'skrr';
+const uint32_t kTagSkiaDRRect = 'skdr';
+}  // namespace
+
 Rect ToRect(const SkRect& rect) {
   return Rect::MakeLTRB(rect.fLeft, rect.fTop, rect.fRight, rect.fBottom);
 }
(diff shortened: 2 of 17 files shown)


impeller/aiks/aiks_context.cc
```diff
@@ -42,7 +43,11 @@ bool AiksContext::Render(const Picture& picture, RenderTarget& render_target) {
   }
 
   if (picture.pass) {
-    return picture.pass->Render(*content_context_, render_target);
+    auto res = picture.pass->Render(*content_context_, render_target);
+    // FIXME(knopp): This should be called for the last surface of the frame,
+    // but there's currently no way to do this.
+    content_context_->GetTessellationCache().FinishFrame();
+    return res;
```
jonahwilliams, line 50: There is no concept of a frame, but you could probably use the concept of a particular presentation. See context_vk, this tracks increments in the swapchain. We might need to bring a similar concept out of the platform specific backend.

impeller/entity/contents/tessellation_cache.cc
```diff
@@ -0,0 +1,70 @@
...
+
+#include "impeller/entity/contents/tessellation_cache.h"
+
+#define IMPELLER_ENABLE_TESSELLATION_CACHE 1
+
+namespace impeller {
+
+Path::Polyline TessellationCache::GetOrCreatePolyline(
+    const impeller::Path& path,
+    Scalar scale) {
```
jonahwilliams, line 13: If you take the exact scale as a cache key then animated transformations will thrash your cache each frame despite the vertices being almost unchanged.

impeller/display_list/skia_conversions.cc
```diff
@@ -121,7 +121,9 @@ Path ToPath(const SkPath& path) {
   }
   builder.SetConvexity(path.isConvex() ? Convexity::kConvex
                                        : Convexity::kUnknown);
-  return builder.TakePath(fill_type);
+  auto result = builder.TakePath(fill_type);
+  result.SetOriginalGenerationId(path.getGenerationID());
```
jonahwilliams, line 125: Where does this generation ID come from?

impeller/entity/contents/content_context.h
```diff
@@ -853,6 +858,7 @@ class ContentContext {
 
   bool is_valid_ = false;
   std::shared_ptr<Tessellator> tessellator_;
+  std::unique_ptr<TessellationCache> tessellation_cache_;
```
jonahwilliams, line 861: Can we make the tessellation cache own the tessellator?

impeller/entity/contents/tessellation_cache.cc
```diff
@@ -0,0 +1,70 @@
...
+          size_t indices_size) {
+        TessellatorData data;
+        data.vertices.assign(vertices, vertices + vertices_size);
+        data.indices.assign(indices, indices + indices_size);
+        tessellator_cache_.Set(key, data);
+        return callback(vertices, vertices_size, indices, indices_size);
+      });
+#else
+  return tesselator.Tessellate(fill_type, polyline, callback);
+#endif
```
jonahwilliams, line 67: It would be nice if the tessellation cache was a bit more general and could also cache data that was generated by compute. For an example, look at the draw points geometry.

For that to work, it would likely have to be able to vend device buffers instead.

On the other hand, we'll likely batch compute workloads in ways that make these buffers difficult to use. So perhaps only using the tessellation cache for the CPU bound workloads is the way forward.

jonahwilliams (commented):
You should also add a benchmark that intentionally does not hit the cache, perhaps by doing scale transformations. I'm technically OOO today, but I'd also like to see a CPU breakdown of where these costs are going.

Tessellation itself is unlikely to be improved, but certain kinds of paths have better algorithms we can use.

impeller/display_list/skia_conversions.cc
knopp: It's assigned to path on request from a global counter. Any time the page is mutated it is reset. That way if path has same generationId it is guaranteed to be same instance unmodified.

knopp:
I added a test commit (would not be part of PR) to remove allocations from path components extrema which seems which seems to speed up bound calculation considerably reducing rasterizer time from 1.66ms to 1.37ms.

But still there seem to be about 30% of rasterization time spent translating same skia paths over and again, and calculating bounds for every path every frame. I think the cache as it is implemented here might be a the wrong place.

knopp:
Closing in favor of https://github.com/flutter/engine/pull/44390 (better performance)

PR #44335, opened 2023-08-03
[darwin] Move common targets to common/BUILD.gn

cbracken:
Previously, some common Darwin framework targets existed in: 
//flutter/shell/platform/darwin/BUILD.gn

This moves all targets into:
//flutter/shell/platform/darwin/common/BUILD.gn

We also de-duplicate targets with existing targets.
* The `flutter_channels` target duplicates the existing `framework_shared` target.
* The `common` target already includes the buffer conversions translation units,
  which are only used by the iOS embedder.

No test changes since there are no semantic changes, just a restructuring of build targets.

## Pre-launch Checklist

- [X] I signed the [CLA].

### DIFF
--- a/shell/platform/darwin/common/BUILD.gn
+++ b/shell/platform/darwin/common/BUILD.gn
@@ -65,7 +65,7 @@ test_fixtures("flutter_channels_fixtures") {
 }
 
 # Unit tests for channels.
-executable("flutter_channels_unittests") {
+executable("framework_shared_unittests") {
   testonly = true
 
   sources = [
--- a/shell/platform/darwin/common/BUILD.gn
+++ b/shell/platform/darwin/common/BUILD.gn
@@ -6,7 +6,7 @@ assert(is_mac || is_ios)
 
 import("//flutter/common/config.gni")
 import("//flutter/testing/testing.gni")
-import("framework_shared.gni")
+import("framework_common.gni")
 
 source_set("common") {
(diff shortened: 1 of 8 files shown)


cbracken:
Taking a quick look to see how much pain would be involved in renaming the unittest executable to framework_shared_unittests.

cbracken:
Pushed a second commit that renames the test target so we can add more to it later.

shell/platform/darwin/common/BUILD.gn
```diff
@@ -56,3 +59,26 @@ source_set("framework_shared") {
 
   deps = [ "//flutter/fml" ]
 }
+
+test_fixtures("flutter_channels_fixtures") {
+  fixtures = []
+}
+
+# Unit tests for channels.
+executable("framework_shared_unittests") {
```
gaaclarke, line 68: Shouldn't this use the same nomenclature, "common" instead of "shared"?

gaaclarke (approved):

shell/platform/darwin/common/BUILD.gn
cbracken: That's a good point. I named it after the framework_shared target, but framework_common is probably the better name. Will update. Thanks!

chinmaygarde (approved):

cyanglaz (approved):
LGTM! Thanks

PR #44238, opened 2023-08-01
[Windows] Delay enabling app lifecycle states until requested

yaakovschectman:
Await a platform message before sending lifecycle state updates so we are not sending messages that do not get consumed by the framework.
In the _near_ future we hope to extend the embedder API to allow registering callbacks called upon the framework registering a listener to a channel, which would obviate this problem.

https://github.com/flutter/flutter/issues/131616

*If you had to change anything in the [flutter/tests] repo, include a link to the migration guide as per the [breaking change policy].*

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [Flutter Style Guide] and the [C++, Objective-C, Java style guides].
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I added new tests to check the change I am making or feature I am adding, or Hixie said the PR is test-exempt. See [testing the engine] for instructions on writing and running engine tests.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [ ] I signed the [CLA].
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/flutter/wiki/Tree-hygiene#overview
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[C++, Objective-C, Java style guides]: https://github.com/flutter/engine/blob/main/CONTRIBUTING.md#style
[testing the engine]: https://github.com/flutter/flutter/wiki/Testing-the-engine
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/wiki/Chat

PR #44344, opened 2023-08-03
Reintroduce Windows lifecycle with guard for posthumous `OnWindowStateEvent`

yaakovschectman:
Previously, destruction of `Window` called `DestroyWindow`, which may send `WM_KILLFOCUS` to the to-be-destroyed window. Because `Window`'s destructor is called after `FlutterWindow`'s, the `FlutterWindow` vtable was already destroyed at this point, and the subsequent call to the virtual method `OnWindowStateEvent` would cause a crash. This PR reintroduces the reverted changes for Windows lifecycle with a check before calling the virtual method that the `FlutterWindow` object has not yet been destructed.

https://github.com/flutter/flutter/issues/131872

## Pre-launch Checklist

- [ ] I signed the [CLA].


yaakovschectman:
Should I incorporate the changes in https://github.com/flutter/engine/pull/44238 into this PR? Doing so will resolve the startup warnings issue at the same time, but will make this a rather broad request.

shell/platform/windows/window.h
```diff
@@ -249,6 +253,9 @@ class Window : public KeyboardManager::WindowDelegate {
   // Accessibility node that represents an alert.
   std::unique_ptr<ui::AXPlatformNodeWin> alert_node_;
 
+  // Guard against posthumous vtable access;
+  bool vtable_is_alive = true;
```
loic-sharma, line 257: This is spooky. Some potential alternative options:

1. Merge `FlutterWindow` and `Window` to make destruction simpler
2. Expose a `Destroy` method to destroy the window manually. This would need to be called before the `Window` is destructed to ensure no messages are receive post destruction.
3. Switch from an inheritance pattern to a delegate pattern so that the `Window` is destructed before the `FlutterWindow`

What are your thoughts?

yaakovschectman: 1. I do not see any reason why they need to be separate classes, though I am a little nervous about trying it. As long as it's safe, and no one will complain about needing both classes for some 3rd party or future change, this seems like a valid option.
2. We could also try this, though do we know from where we could call it, or also importantly, if we have any way to ensure it's called before destruction?
3. If `FlutterWindow` is our only use case for `Window`, merging the classes seems like a better idea.

loic-sharma: > 1. I do not see any reason why they need to be separate classes, though I am a little nervous about trying it. As long as it's safe, and no one will complain about needing both classes for some 3rd party or future change, this seems like a valid option.

This type hierarchy is a common win32 C++ pattern: https://learn.microsoft.com/windows/win32/learnwin32/managing-application-state-#an-object-oriented-approach

I'm not in love with this pattern though 😅

> 2. ... though do we know from where we could call it, or also importantly, if we have any way to ensure it's called before destruction?

One potential place might `FlutterDesktopViewControllerDestroy`. We could manually close the window before destroying our objects:

https://github.com/flutter/engine/blob/138a1ea9a69235342f479d61399fe7394fc1f3cb/shell/platform/windows/flutter_windows.cc#L90-L93

> 3. ... merging the classes seems like a better idea.

Agreed. Right now I'm leaning towards option 2, it seems like the least intrusive option.

/cc @cbracken For thoughts

yaakovschectman: What do you think about moving the call to `Window::Destroy()` from `~Window` to `~FlutterWindow`? I just do not know how we can enforce within a test that window messages are received between the two destructors.

cbracken: In the interest of getting this landed, agreed that option 2 (make Destroy() protected, then move the destroy call down to the subclass) is probably the quickest fix. Once that's landed, let's follow up with a patch that implements option 1 (merge Window and FlutterWindow).

I don't think there's a strong case for keeping them separate. Now that the UWP embedder is gone let's use the [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) rule here and merge the two. If we ever find a good reason for a separation, we can look at that in the future.

cbracken, line 257: On a separate note, I'm surprised the linter didn't catch the missing underscore suffix on this field. Probably irrelevant given the change in approach proposed for this patch, but I'll file a bug to see why the linters aren't running on windows.

shell/platform/windows/client_wrapper/include/flutter/flutter_engine.h
```diff
@@ -84,6 +85,16 @@ class FlutterEngine : public PluginRegistry {
   // once on the platform thread.
   void SetNextFrameCallback(std::function<void()> callback);
 
+  // Called to pass an external window message to the engine for lifecycle
+  // state updates. This does not consume the window message. Non-Flutter
+  // windows must call this method in their WndProc in order to be included in
+  // the logic for application lifecycle state updates. Returns a result when
```
loic-sharma, line 91:
```suggestion
  // the logic for application lifecycle state updates. Returns a result if
```

shell/platform/windows/client_wrapper/include/flutter/flutter_engine.h
```diff
@@ -84,6 +85,16 @@ class FlutterEngine : public PluginRegistry {
   // once on the platform thread.
   void SetNextFrameCallback(std::function<void()> callback);
 
+  // Called to pass an external window message to the engine for lifecycle
+  // state updates. This does not consume the window message. Non-Flutter
```
loic-sharma, line 89:
```suggestion
  // state updates. Non-Flutter
```

shell/platform/windows/client_wrapper/testing/stub_flutter_windows_api.h
```diff
@@ -89,6 +89,17 @@ class StubFlutterWindowsApi {
   // FlutterDesktopPluginRegistrarUnregisterTopLevelWindowProcDelegate.
   virtual void PluginRegistrarUnregisterTopLevelWindowProcDelegate(
       FlutterDesktopWindowProcCallback delegate) {}
+
+  // Claled for FlutterDesktopEngineProcessExternalWindowMessage.
```
loic-sharma, line 93:
```suggestion
  // Called for FlutterDesktopEngineProcessExternalWindowMessage.
```

shell/platform/windows/public/flutter_windows.h
```diff
@@ -212,6 +212,19 @@ FLUTTER_EXPORT HWND FlutterDesktopViewGetHWND(FlutterDesktopViewRef view);
 FLUTTER_EXPORT IDXGIAdapter* FlutterDesktopViewGetGraphicsAdapter(
     FlutterDesktopViewRef view);
 
+// Called to pass an external window message to the engine for lifecycle
+// state updates. This does not consume the window message. Non-Flutter windows
```
loic-sharma, line 216:
```suggestion
// state updates. Non-Flutter windows
```

shell/platform/windows/public/flutter_windows.h
```diff
@@ -212,6 +212,19 @@ FLUTTER_EXPORT HWND FlutterDesktopViewGetHWND(FlutterDesktopViewRef view);
...
+// must call this method in their WndProc in order to be included in the logic
+// for application lifecycle state updates. Returns true when the message is
```
loic-sharma, line 218:
```suggestion
// for application lifecycle state updates. Returns true if the message should be
```

shell/platform/windows/client_wrapper/include/flutter/flutter_engine.h
```diff
@@ -84,6 +85,16 @@ class FlutterEngine : public PluginRegistry {
...
+  // windows must call this method in their WndProc in order to be included in
+  // the logic for application lifecycle state updates. Returns a result when
+  // the message has been consumed.
```
loic-sharma, line 92:
```suggestion
  // the logic for application lifecycle state updates. Returns a result if
  // the message should be consumed.
```

shell/platform/windows/client_wrapper/include/flutter/flutter_engine.h
```diff
@@ -84,6 +85,16 @@ class FlutterEngine : public PluginRegistry {
   // once on the platform thread.
   void SetNextFrameCallback(std::function<void()> callback);
 
+  // Called to pass an external window message to the engine for lifecycle
+  // state updates. Non-Flutter
+  // windows must call this method in their WndProc in order to be included in
```
loic-sharma, line 90: Please reformat this to fill the line

loic-sharma: Same comment for `FlutterDesktopEngineProcessExternalWindowMessage`

shell/platform/windows/window.h
```diff
@@ -236,6 +240,9 @@ class Window : public KeyboardManager::WindowDelegate {
   // Returns the root view accessibility node, or nullptr if none.
   virtual gfx::NativeViewAccessible GetNativeViewAccessible() = 0;
 
+  // Release OS resources associated with window.
```
loic-sharma, line 243:
```suggestion
  // Release OS resources associated with the window.
```

shell/platform/windows/windows_lifecycle_manager.cc
```diff
@@ -163,4 +183,108 @@ void WindowsLifecycleManager::BeginProcessingClose() {
   process_close_ = true;
 }
 
+void WindowsLifecycleManager::SetLifecycleState(AppLifecycleState state) {
```
loic-sharma, line 186: Please add a `TODO` that we shouldn't send the message until we receive the system initialized message.

loic-sharma:
@yaakovschectman Before merging this, please verify that this latest approach also works as expected with the internal app team.

yaakovschectman:
@loic-sharma Someone from said team confirmed they experience no crash when testing with this build.

shell/platform/windows/windows_lifecycle_manager.cc
```diff
@@ -163,4 +183,111 @@ void WindowsLifecycleManager::BeginProcessingClose() {
   process_close_ = true;
 }
 
+// TODO(schectman): Wait until the platform channel is registered to send
+// the platform message.
+// https://github.com/flutter/engine/pull/44238
```
loic-sharma, line 188: Please use the tracking issue instead of a pull request :)

```suggestion
// https://github.com/flutter/flutter/issues/131616
```

loic-sharma (approved):
LGTM, nice work! 🎉

cbracken (approved):

PR #44346, opened 2023-08-03
[Impeller] Fail loudly if --enable-software-rendering is used.

dnfield:
We have tests today that are trying to testwith impeller but using software rendering, so they're not actually testing with impeller.

In an unrelated patch, I'm causing the request to enable impeller to have additional checks that Impeller related objects/configurations are set when using Impeller, which isn't true if software rendering is enabled, and is causing the tests getting disabled here to fail.

The tests are not actually testing Impeller and thus are getting added to the skip list (they now crash with this change). https://github.com/flutter/flutter/issues/131888 tracks re-enabling them. They will almost certainly need different golden files than the Skia ones.

Fixes https://github.com/flutter/flutter/issues/131887


shell/common/shell.cc
```diff
@@ -419,6 +419,8 @@ Shell::Shell(DartVMRef vm,
       volatile_path_tracker_(std::move(volatile_path_tracker)),
       weak_factory_gpu_(nullptr),
       weak_factory_(this) {
+  FML_CHECK(!settings.enable_software_rendering || !settings.enable_impeller)
```
cyanglaz, line 422: nits: 

Would `settings.enable_software_rendering ^ settings.enable_impeller` be more readable?

cyanglaz (approved):
LGTM

bdero (approved):
LGTM

dnfield:
I'm not clear if the message on Fuchsia is just slightly different or if the test is messing up the Fuchsia harness, but Impeller isn't a thing on Fuchsia anyway so I'm skipping it there.

PR #44361, opened 2023-08-03
[Impeller] Replace Vulkan rotation checks with polling

matanlurey:
tl;dr: Reverts https://github.com/flutter/engine/pull/43214. closes https://github.com/flutter/flutter/issues/129459.

7v5zfd

---

In https://github.com/flutter/engine/pull/43214, @jonahwilliams used `vkQueuePresentKHR`[^1] [to check](https://github.com/jonahwilliams/engine/blob/a0df46add166202fe68a853391d379e69ddc7c18/impeller/renderer/backend/vulkan/swapchain_impl_vk.cc#L465-L467) if device orientations had changed, and if so, the swap-chain was torn down and replaced with a new instance (that would now have a correct rotation for ... compositing reasons I don't quite understand).

Unfortunately `vkQueuePresentKHR` is (a) only present on Android 10+ _and_ (b) ... isn't implemented consistently across Android devices. For example, the popular Samsung Galaxy S10 doesn't return `VK_SUBOPTIMAL_KHR`, leading to the same rotation bugs described in https://github.com/flutter/flutter/issues/129459 (i.e. even with https://github.com/flutter/engine/pull/43214).

This PR implements the polling technique recommended by Android[^2], in this case on every frame. We should expect this to have a performance penalty of ~`0.2`ms per frame, but should at least give us consistent fidelity on Vulkan + Android for the time-being.

flutter_04

<details>
<summary>
Some additional screenshots generated while debugging
</summary>

</details>

[^1]: https://developer.android.com/games/optimize/vulkan-prerotation#detect_device_orientation_changes
[^2]: https://developer.android.com/games/optimize/vulkan-prerotation#using_polling


impeller/renderer/backend/vulkan/swapchain_impl_vk.cc
```diff
@@ -470,10 +487,6 @@ bool SwapchainImplVK::Present(const std::shared_ptr<SwapchainImageVK>& image,
             // complete successfully.
             [[fallthrough]];
           case vk::Result::eSuccess:
-            is_rotated_ = false;
-            return;
-          case vk::Result::eSuboptimalKHR:
```
jonahwilliams, line 475: We still need to treat this case as a success otherwise we'll exit from the application with validations enabled.

impeller/renderer/backend/vulkan/swapchain_impl_vk.cc
```diff
@@ -10,10 +10,12 @@
 #include "impeller/renderer/backend/vulkan/formats_vk.h"
 #include "impeller/renderer/backend/vulkan/surface_vk.h"
 #include "impeller/renderer/backend/vulkan/swapchain_image_vk.h"
+#include "vulkan/vulkan_structs.hpp"
 
 namespace impeller {
 
 static constexpr size_t kMaxFramesInFlight = 3u;
+static constexpr size_t kPollFramesForOrientation = 1u;
```
jonahwilliams, line 18: nit: document what this means (i.e. 1 is every frame, 2 is every other frame)

jonahwilliams (approved):
LGTM with nits

impeller/renderer/backend/vulkan/swapchain_impl_vk.cc
matanlurey: Done!

impeller/renderer/backend/vulkan/swapchain_impl_vk.cc
matanlurey: Done!

impeller/renderer/backend/vulkan/swapchain_impl_vk.cc
```diff
@@ -469,11 +490,12 @@ bool SwapchainImplVK::Present(const std::shared_ptr<SwapchainImageVK>& image,
...
             // complete successfully.
             [[fallthrough]];
-          case vk::Result::eSuccess:
-            is_rotated_ = false;
-            return;
           case vk::Result::eSuboptimalKHR:
-            is_rotated_ = true;
+            // In theory we shouldn't receive this (because we're polling above)
+            // but we shouldn't handle it as a validation error if we do receive
+            // it.
```
jonahwilliams, line 496: I think this isn't correct - right. Won't we be presenting with suboptimalkhr on every frame when we are rotated? (give that we aren't doing vulkan prerotation). 

Don't try this on the S10 😄

matanlurey: Updated!

PR #44395, opened 2023-08-04
[Darwin] Move FlutterBinaryMessengerRelay to common

cbracken:
Migrates FlutterBinaryMessengerRelay to
//flutter/shell/platform/darwin/common so that it can be used in a followup patch on macOS, to fix a memory leak due to a retain cycle.

Migrates the unit tests to be compatible with gtest, which is used by the darwin/common unit tests.

Issue: https://github.com/flutter/flutter/issues/116445

## Pre-launch Checklist

- [X] I signed the [CLA].

### DIFF
--- a/shell/platform/darwin/common/framework/Source/FlutterBinaryMessengerRelayTest.mm
+++ b/shell/platform/darwin/common/framework/Source/FlutterBinaryMessengerRelayTest.mm
@@ -2,35 +2,31 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
-#import "flutter/shell/platform/darwin/ios/framework/Source/FlutterBinaryMessengerRelay.h"
+#import "flutter/shell/platform/darwin/common/framework/Source/FlutterBinaryMessengerRelay.h"
 
 #import <OCMock/OCMock.h>
-#import <XCTest/XCTest.h>
 
 #import "flutter/shell/platform/darwin/common/framework/Headers/FlutterMacros.h"
+#import "flutter/shell/platform/darwin/common/framework/Source/FlutterTestUtils.h"
+#import "flutter/testing/testing.h"
+#include "gtest/gtest.h"
 
 FLUTTER_ASSERT_ARC
 
 @protocol FlutterTaskQueue <NSObject>
 @end
 
-@interface FlutterBinaryMessengerRelayTest : XCTestCase
+@interface FlutterBinaryMessengerRelayTest : NSObject
 @end
 
 @implementation FlutterBinaryMessengerRelayTest
-
-- (void)setUp {
-}
-
-- (void)tearDown {
-}
 
 - (void)testCreate {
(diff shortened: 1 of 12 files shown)


cbracken:
This is currently stacked on top of https://github.com/flutter/engine/pull/44396.

shell/platform/darwin/common/framework/Source/FlutterBinaryMessengerRelayTest.mm
```diff
@@ -2,49 +2,51 @@
...
   NSString* channel = @"foobar";
   [relay sendOnChannel:channel message:message binaryReply:nil];
-  OCMVerify([messenger sendOnChannel:channel message:message binaryReply:nil]);
+  @try {
+    OCMVerify(  // NOLINT(google-objc-avoid-throwing-exception)
+        [messenger sendOnChannel:channel message:message binaryReply:nil]);
+  } @catch (...) {
+    return NO;
+  }
+  return YES;
```
gaaclarke, line 46: Instead of moving the try block into every method, wouldn't it be better just to make a helper function that can wrap these calls?

```objc
BOOL ThrowsObjcException(dispatch_block_t block) {
  @try {
    block();
  } @catch (...) {
    return NO;
  }
  return YES;
}
```

Doing try/catch in objc is so uncommon/unorthodox, it's worth limiting it to one instance.

cbracken: Seems reasonable; that would allow us to eliminate a lot of this in the macOS embedder as well.

cbracken: Done.

shell/platform/darwin/common/framework/Source/FlutterBinaryMessengerRelayTest.mm
```diff
@@ -2,49 +2,48 @@
...
   FlutterBinaryMessengerRelay* relay =
       [[FlutterBinaryMessengerRelay alloc] initWithParent:messenger];
   char messageData[] = {'a', 'a', 'r', 'o', 'n'};
   NSData* message = [NSData dataWithBytes:messageData length:sizeof(messageData)];
   NSString* channel = @"foobar";
   [relay sendOnChannel:channel message:message binaryReply:nil];
-  OCMVerify([messenger sendOnChannel:channel message:message binaryReply:nil]);
+  return !FLTThrowsObjcException(^{
+    OCMVerify([messenger sendOnChannel:channel message:message binaryReply:nil]);
+  });
```
gaaclarke, line 43: I was thinking moving `FLTThrowsObjcException` up one level higher, then all of these tests can stay unchanged.

gaaclarke (approved):

shell/platform/darwin/common/framework/Source/FlutterBinaryMessengerRelayTest.mm
cbracken: Ah yeah - much better!

PR #44401, opened 2023-08-04
Remove GFX branches from Flutter engine

caroqliu:
This change does the following:
- Refactors `engine.cc`/`.h` and `component_v2.cc`/`.h` to remove GFX ctors and callbacks.
- Removes `gfx_external_view_embedder`, `gfx_platform_view`, `gfx_session_connection`, their tests, and any supporting testing infrastructure (e.g. `fake_session`) from the build. This change does not delete the source files themselves in order to keep the change size smaller (as all these file deletions together are about another ~8K lines worth of deletions.) A separate follow up change will delete these files.

Related bug: fxbug.dev/64206

## Pre-launch Checklist

- [x] I signed the [CLA].


arbreng (commented):
Please rename all of the "flatland_" files by dropping the "flatland_" prefix

Flatland is the only implementation of these interfaces now, and the name will not be meaningful to other contributors in this repository who have to change these files.  Names like "platform_view" and "external_view_embedder" will however.  Dropping the prefix removes cognitive overhead for Flutter contributors.

For "platform_view", "flatland_platform_view.cc" should be collapsed back into "platform_view.cc" and the various callbacks removed.  They add complexity and overhead which is no longer necessary.

I would greatly prefer these changes be made as part of this PR, but an immediate follow-up is acceptable if you don't want to bloat the line count.  However, punting on the work is not acceptable to me at this time.

shell/platform/fuchsia/flutter/platform_view.cc
```diff
@@ -876,10 +874,9 @@ bool PlatformView::HandleFuchsiaChildViewChannelPlatformMessage(
   FML_DCHECK(message->channel() == kFuchsiaChildViewChannel);
 
   if (message->data().GetSize() != 1 ||
-      (message->data().GetMapping()[0] != '0' &&
-       message->data().GetMapping()[0] != '1')) {
+      (message->data().GetMapping()[0] != '1')) {
     FML_LOG(ERROR) << kFuchsiaChildViewChannel
-                   << " data must be '0' (for gfx) or '1' (for flatland).";
+                   << " data must be singularly '1' (for flatland).";
```
arbreng, line 879: Please remove the "(for flatland)"

caroqliu:
> Please rename all of the "flatland_" files by dropping the "flatland_" prefix
> 
> Flatland is the only implementation of these interfaces now, and the name will not be meaningful to other contributors in this repository who have to change these files. Names like "platform_view" and "external_view_embedder" will however. Dropping the prefix removes cognitive overhead for Flutter contributors.
> 
> For "platform_view", "flatland_platform_view.cc" should be collapsed back into "platform_view.cc" and the various callbacks removed. They add complexity and overhead which is no longer necessary.
> 
> I would greatly prefer these changes be made as part of this PR, but an immediate follow-up is acceptable if you don't want to bloat the line count. However, punting on the work is not acceptable to me at this time.

I've removed the `flatland_` prefixes as suggested, with the following exceptions:
- `flatland_` itself, which is referring to the composition API. (Let me know if this should be renamed as `composition_` or some such.
- `flatland_connection_`/`FlatlandConnection`, since this is referring to the connection with the composition API. I assume `Connection` on its own is too vague, but let me know if you prefer `CompositionConnection` or some such.

I've also added the `Callback` suffix to the existing callback types related to the `PlatformView`, since `OnCreateView` is overloaded with the refactor (both as the function and as the callback type). I added `Callback` to distinguish these two in particular, and added `Callback` to the other types so they would have a consistent naming style. LMK if you have any thoughts on this as well.

I'm sending an early unpolished version to CI before I head out for the day; if you have any suggestions for rearranging code blocks (e.g. with FlatlandPlatformView and PlatformView being merged together), or anything like that, please weigh in as these are still flexible. 

Thanks for the review!

arbreng (approved):
Most excellent!

chinmaygarde (approved):

PR #44445, opened 2023-08-07
[web] Silence `pub get` when it's successful

mdebbar:
In the spirit of keeping the happy path's output as clean as possible, let's hide the many lines printed by `pub get` even when it's successful.

If `pub get` fails, its output will be printed on the terminal.


eyebrowsoffire (approved):
LGTM!

PR #44469, opened 2023-08-07
Make toJS'd function use JS types

srujzs:
JSFunction's should only accept and return JS types.

Allows landing of external restrictions here: https://dart-review.googlesource.com/c/sdk/+/316867/9


eyebrowsoffire (approved):
LGTM!

PR #44505, opened 2023-08-08
[web] Remove some unused functions

mdebbar:


lib/web_ui/lib/painting.dart
```diff
@@ -5,18 +5,6 @@
...
-bool _offsetIsValid(Offset offset) {
-  assert(!offset.dx.isNaN && !offset.dy.isNaN, 'Offset argument contained a NaN value.');
-  return true;
-}
-
-// ignore: unused_element, Used in Shader assert.
-bool _matrix4IsValid(Float32List matrix4) {
-  assert(matrix4.length == 16, 'Matrix4 must have 16 entries.');
-  return true;
-}
```
mdebbar, line 18: @yjbanov is there a reason we need to keep these?

yjbanov (approved):
lgtm

PR #44526, opened 2023-08-09
[web] More efficient fallback font selection

rakudrama:
The PR improves the code size and runtime performance of fallback font selection.

### Performance improvements

Initialization of the data structures to support fallback font selection has been moved from creating the FallbackFontManager (first frame) to the first use, i.e. the first frame actually needing a fallback font.

The numbers reported below are for a lightly edited version of the counter demo that appends to the counter about ~300 missing code points that need ~25 fallback fonts to cover the missing code points. Timings taken from a few profiles on my performance workstation.

|  |  Before | After  | 
| --- | ---: | ---: |
| FallbackFontManager() |~100ms | <2ms |
| First need | 0ms | 12ms |
| Subsequent need | 20-30ms | <1ms |

### Size improvements

|  |  Before | After  | Δ |
| --- | ---: | ---: | ---: |
| main.dart.js | 1586405 | 1477319 | -109086 (-6.87%) |
| brotli -9 | 427304 | 401611 | -25693 (-6.01%) |

#### Startup

The old algorithm built an interval tree from the code point ranges of the ~140 fallback fonts and uses the interval tree to build a list of fonts that support each missing code point. The new algorithm uses a binary search map that directly produces the list of fonts. There are fewer binary search ranges (~22k) than the aggregate ranges for all the fonts (~26k).
Most of the startup time gain comes from using a data unpacks directly into a useful form rather than needing processing to build an interval tree (~12ms vs ~100ms).

#### Running

The runtime for font selection is greatly improved for several reasons

- The code point space is partitioned into components so that code point counting can be batched.
- When a font is selected, the counts are updated incrementally rather than being recomputed.
- The counts are held in fields of the NotoFont and component objects rather than in Maps or Sets.

Batching, incremental update and avoiding hash tables are roughly multiplicative in effect.

## Issues

- https://github.com/flutter/flutter/issues/131440

## Pre-launch Checklist

- [x] I signed the [CLA].


lib/web_ui/dev/roll_fallback_fonts.dart
```diff
@@ -509,3 +494,354 @@ bool _checkForLicenseAttribution(Uint8List fontBytes) {
...
+///  (1) A list of sets of fonts.
+///  (2) A list of code point ranges mapping to an index of list (1).
+///
+/// Each set of fonts is represented as a list of font indexes. The indexes are
+/// always increasing so the delta is stored. The stored value is biased by -1
+/// (i.e.  `delta - 1`) since a delta is never less than 1. The deltas are STMR
+/// encoded.
+///
+/// The list of code point ranges is complete, covering every code point. There
+/// are no gaps bewteen ranges. Instead there is a range mapping the missing
```
harryterkelsen, line 632: typo: bewteen -> between

lib/web_ui/dev/roll_fallback_fonts.dart
```diff
@@ -509,3 +494,354 @@ bool _checkForLicenseAttribution(Uint8List fontBytes) {
...
+  final List<_Range> ranges = <_Range>[];
+  final List<_FontSet> allSets = <_FontSet>[];
+
+  {
+    // The fonts have their supported code points provided as list of inclusive
+    // [start, end] ranges. We want to intersect all of these ranges and find
+    // the fonts that overlap each intersected range.
+    //
+    // It is easier to work with the boundaries of the ranges rather than the
+    // ranges themselves. The bounaries of the intersected ranges is the union
```
harryterkelsen, line 688: typo: bounaries -> boundaries

lib/web_ui/dev/roll_fallback_fonts.dart
```diff
@@ -509,3 +494,354 @@ bool _checkForLicenseAttribution(Uint8List fontBytes) {
...
+}
+
+/// A trie node [1] used to find the canonical _FontSet.
+///
+/// [1]: https://en.wikipedia.org/wiki/Trie
+class _TrieNode {
+  final Map<_Font, _TrieNode> _children = <_Font, _TrieNode>{};
+  _FontSet? fontSet;
+
+  _TrieNode insertAtRoot(Iterable<_Font> fonts) {
```
harryterkelsen, line 596: IIUC this adds the fonts (eg `[font1, font2, font3, ...]`) in a line, not all as children of the Trie that `insertAtRoot` is called on, which is what I assumed when I read the function name. Add docs explaining what this does and what it returns?

```
font1
  \
  font2
    \
    font3
    ...
```

lib/web_ui/dev/roll_fallback_fonts.dart
```diff
@@ -509,3 +494,354 @@ bool _checkForLicenseAttribution(Uint8List fontBytes) {
...
+///
+/// Each set of fonts is represented as a list of font indexes. The indexes are
+/// always increasing so the delta is stored. The stored value is biased by -1
+/// (i.e.  `delta - 1`) since a delta is never less than 1. The deltas are STMR
+/// encoded.
+///
+/// The list of code point ranges is complete, covering every code point. There
+/// are no gaps bewteen ranges. Instead there is a range mapping the missing
+/// code points to the empty set of fonts. Each range is encoded as the size
+/// (number of code points) in the range followed by the value which the index
```
harryterkelsen, line 634: which -> which is

lib/web_ui/dev/roll_fallback_fonts.dart
```diff
@@ -509,3 +494,354 @@ bool _checkForLicenseAttribution(Uint8List fontBytes) {
...
+      fontSet.rangeCount++;
+      final _Range range = _Range(start, end, fontSet);
+      ranges.add(range);
+    }
+
+    int start = 0;
+    for (final _Boundary boundary in boundaries) {
+      final int value = boundary.value;
+      if (start < value) {
+        newRange(start, value - 1);
```
harryterkelsen, line 727: nit: maybe document that `value == 0` is impossible?

lib/web_ui/dev/roll_fallback_fonts.dart
```diff
@@ -509,3 +494,354 @@ bool _checkForLicenseAttribution(Uint8List fontBytes) {
...
+///
+///  (1) A list of sets of fonts.
+///  (2) A list of code point ranges mapping to an index of list (1).
+///
+/// Each set of fonts is represented as a list of font indexes. The indexes are
+/// always increasing so the delta is stored. The stored value is biased by -1
+/// (i.e.  `delta - 1`) since a delta is never less than 1. The deltas are STMR
+/// encoded.
+///
+/// The list of code point ranges is complete, covering every code point. There
```
harryterkelsen, line 631: docs nit: this point wasn't clear to me until I started reading the algorithm and saw that the empty FontSet covered the uncovered ranges. Maybe change this paragraph to highlight the importance of the empty set of fonts in the first sentence?

lib/web_ui/dev/roll_fallback_fonts.dart
```diff
@@ -509,3 +494,354 @@ bool _checkForLicenseAttribution(Uint8List fontBytes) {
...
+      }
+    }
+    boundaries.sort(_Boundary.compare);
+
+    final _TrieNode trieRoot = _TrieNode();
+    final Set<_Font> currentElements = <_Font>{};
+
+    void newRange(int start, int end) {
+      // Ensure we are using the canonical font order.
+      final List<_Font> fonts = List<_Font>.of(currentElements)
```
harryterkelsen, line 711: maybe the algorithm should just always keep currentElements sorted when it inserts a new font into the set

harryterkelsen (approved):
LGTM with nits

Thanks so much for this contribution!

lib/web_ui/dev/roll_fallback_fonts.dart
rakudrama: I have renamed the function and added a comment to clarify that it follows a path trough the trie.

lib/web_ui/dev/roll_fallback_fonts.dart
rakudrama: I did have it that way but it was too 'clever'. Sorting the boundaries by increasing index for starts but decreasing index for ends, and sorting ends before starts, will give you two ordered subsequences that still need to be merged. When I decided to simply sort, it git rid of ~30 lines of code.

lib/web_ui/dev/roll_fallback_fonts.dart
rakudrama: Done.

rakudrama:
@eyebrowsoffire PTAL

eyebrowsoffire (approved):
Thank you so much for taking the time to do all this! This is going to be an amazing improvement.

lib/web_ui/dev/roll_fallback_fonts.dart
```diff
@@ -509,3 +494,364 @@ bool _checkForLicenseAttribution(Uint8List fontBytes) {
...
+
+  String description() {
+    return fonts.map((_Font font) => font.shortName).join(', ');
+  }
+}
+
+/// A trie node [1] used to find the canonical _FontSet.
+///
+/// [1]: https://en.wikipedia.org/wiki/Trie
+class _TrieNode {
```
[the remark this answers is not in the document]
rakudrama: Yes, this is somewhat vestigial of a more complicated scheme I had.

We could use a set of FontSets, but it would require a custom hashCode and equality, which is more code than the trie. (We would probably want a Map, rather than a Set so we have a distinct representative object for each set that has the index as a field.)

---

Before I simplified away a couple of pages of code, the parent structure of the trie was useful in differentially encoding one FontSet against the previous. We could save another 1kB of compiled JavaScript size with the more complicated scheme, but we also have to maintain the code. There are simpler ways of getting a similar saving (not implemented here), and the differential encoding _requires_ all the fontsets are decoded together. We don't decode them on need, but we could shave a few ms (<4ms)  from first-frame-with-uncovered-codepoints by decoding on demand.

lib/web_ui/dev/roll_fallback_fonts.dart
```diff
@@ -509,3 +494,364 @@ bool _checkForLicenseAttribution(Uint8List fontBytes) {
...
+///
+/// A code point with no fonts is mapped to an empty set of fonts. This allows
+/// the list of code point ranges to be complete, covering every code
+/// point. There are no gaps between ranges; instead there are some ranges that
+/// map to the empty set. Each range is encoded as the size (number of code
+/// points) in the range followed by the value which is the index of the
+/// corresponding set in the list of sets.
+///
+///
+/// STMR (Self terminating multiple radix) encoding
```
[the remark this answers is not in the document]
rakudrama: It is not dissimilar to VLQ encodings like those used by MIDI, sourcemaps, and protobuf varint encoding where there is a 'continuation' bit, which could be interpreted as partitioning the digits (bytes) into prefix (continuation=1) and terminating (continuation=0) digits.

I have added a sentence to reference VLQ.

lib/web_ui/dev/roll_fallback_fonts.dart
```diff
@@ -133,49 +160,26 @@ class RollFallbackFontsCommand extends Command<bool>
     sb.writeln();
     sb.writeln('List<NotoFont> getFallbackFontData(bool useColorEmoji) => <NotoFont>[');
```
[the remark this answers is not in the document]
rakudrama: Changed to `getFallbackFontList`.

PR #44527, opened 2023-08-09
[Impeller] Cache render target texture allocations across frames.

jonahwilliams:
When allocating a non-host visible texture for a render target, the allocator will hold onto a reference to this texture descriptor. On the immediate next frame, this texture is made available to replace requested allocations for new render target textures. if this texture is not used, then at the end of the next frame it is discarded.

This removes the vast majority of allocations in most flutter gallery and wonderous. This does not attempt to use different sized textures for the cache.

There are two caveats noted in the PR contents.

Fixes https://github.com/flutter/flutter/issues/131515


impeller/core/allocator.cc
```diff
@@ -54,13 +77,27 @@ std::shared_ptr<Texture> Allocator::CreateTexture(
     return nullptr;
   }
 
+  if (enable_render_target_texture_cache_ && !desc.ignore_cache &&
```
jonahwilliams, line 80: The ignore cache setting is required for this allocator cache to not fight with the swapchain caching we do for the MSAA textures. However, if this patch is succesful then we can remove the swapchain caching of MSAA textures and just rely on this.

I will make that change in a separate patch so that it is easier to understand the benchmark results.

impeller/core/texture_descriptor.h
```diff
@@ -46,6 +46,7 @@ struct TextureDescriptor {
       static_cast<TextureUsageMask>(TextureUsage::kShaderRead);
   SampleCount sample_count = SampleCount::kCount1;
   CompressionType compression_type = CompressionType::kLossless;
+  bool ignore_cache = false;
```
jonahwilliams, line 49: See note above, this is temporary

impeller/renderer/backend/metal/surface_mtl.mm
```diff
@@ -224,6 +225,7 @@
   if (!context) {
     return false;
   }
+  context->DidFinishSurfaceFrame();
```
bdero, line 228: Do we need to call these in the playground and between golden PNG generation in order to not leak/risk GPU OOM?

impeller/renderer/context.h
```diff
@@ -154,6 +154,20 @@ class Context {
...
+  ///        This may or may not indicate a new frame, or there may be multiple
+  ///        presents within a single frame due to platform view presence.
+  virtual void DidAcquireSurfaceFrame() const = 0;
+
+  //---------------------------------------------------------------------------
+  /// @brief Indicate the ending of a presentation cycle.
+  ///
+  ///        This may or may not indicate a new frame, or there may be multiple
+  ///        presents within a single frame due to platform view presence.
+  virtual void DidFinishSurfaceFrame() const = 0;
```
bdero, line 169: Should we call these `BeginFrame()` and `EndFrame()`, since surfaces aren't necessarily involved here?

We should also document that GPU resources may leak if they aren't called properly.

impeller/core/allocator.cc
```diff
@@ -35,6 +36,28 @@ std::shared_ptr<DeviceBuffer> Allocator::CreateBufferWithCopy(
...
+void Allocator::DidFinishSurfaceFrame() {
+  std::vector<TextureData> retain;
+
+  for (auto td : data_to_recycle_) {
+    if (td.used_this_frame) {
+      retain.push_back(td);
+    }
+  }
+  data_to_recycle_.clear();
+  data_to_recycle_.insert(data_to_recycle_.end(), retain.begin(), retain.end());
```
dnfield, line 54:
```suggestion
  data_to_recycle_.swap(retain);
```

impeller/core/allocator.cc
dnfield: Add a TODO?

impeller/core/texture_descriptor.h
dnfield: (Add TODO here as well linked to same issue)

impeller/core/texture_descriptor.h
```diff
@@ -66,6 +67,19 @@ struct TextureDescriptor {
...
+  ///        This is used to check compatibility of already allocated textures
+  ///        in the render target cache..
+  constexpr bool IsCompatibleWith(const TextureDescriptor& other) const {
+    return size == other.size &&                  //
+           storage_mode == other.storage_mode &&  //
+           format == other.format &&              //
+           usage == other.usage &&                //
+           sample_count == other.sample_count &&  //
+           type == other.type;                    //
+  }
```
dnfield, line 81: maybe just `constexpr bool operator==(`?

impeller/renderer/context.h
```diff
@@ -154,6 +154,20 @@ class Context {
   ///
   virtual std::shared_ptr<CommandBuffer> CreateCommandBuffer() const = 0;
 
+  //---------------------------------------------------------------------------
+  /// @brief Indicate the beginning of a presentation cycle.
+  ///
+  ///        This may or may not indicate a new frame, or there may be multiple
+  ///        presents within a single frame due to platform view presence.
+  virtual void DidAcquireSurfaceFrame() const = 0;
```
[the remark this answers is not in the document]
jonahwilliams: in terms of where does the cache live, we could definitely put it in entity pass. But I'm not sure how we live without the frame concept. I don't think its the case that 1 entity pass = 1 presentation cycle, or is it the case that the current cache is based around frames.

FYI @dnfield , does you AIKS canvas patch cause us to create multiple entity passes per frame (non-platform view)? or are impeller picutres always combined into a single entity pass?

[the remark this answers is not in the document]
jonahwilliams: Well, we can't store it in entity pass - but we could store it on the content context and access via entity pass.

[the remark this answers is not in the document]
bdero: > is the context the HAL?

Yeah, `impeller::Context` is the central resource of the HAL, since you use it to spawn all GPU resources and behaviors. By HAL I mean pretty much everything under `impeller/renderer`, or our equivalent or `gfx_hal` or WebGPU.

> and what is the application?

Renderers built atop the HAL. `EntityPass`, Flutter GPU renderers, and my stupid side projects 😛.

> But I'm not sure how we live without the frame concept. I don't think its the case that 1 entity pass = 1 presentation cycle, or is it the case that the current cache is based around frames.

One call to `EntityPass::Render()` should happen per-platform view per-frame.

> does you AIKS canvas patch cause us to create multiple entity passes per frame (non-platform view)?

Oh, do you mean the `DrawPicture` stuff? `DrawPicture` merges a cloned `EntityPass` from a picture into another and won't result in `EntityPass::Render()` being called more than once per-platform view per-frame.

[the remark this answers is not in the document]
bdero: Ah, you probably mean https://github.com/flutter/engine/pull/44248.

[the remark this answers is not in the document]
jonahwilliams: SGTM, I'll move the allocation cache to a structure on the content context. I think I'll adjust the render target constructors a bit so that they also accept an allocator - and I can wrap the context allocator in a caching content_context allocator?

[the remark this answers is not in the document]
bdero: TBH the `RenderTarget::CreateOffscreen[MSAA]` static factories wound up being super Entities-specific in the end and belong in something like `ContentContext` anyhow. Moving those factories up into Entities and passing in some optional Entities-specific cache through `EntityPass CreateRenderTarget` seems like a potentially good route.

jonahwilliams:
Marking WIP while I move the caching around.

jonahwilliams:
Based on @bdero 's feedback I've moved the caching to EntityPass::Render. This works as expected in all backends, and Ill follow up tomorrow to ensure that this keeps working with @dnfield s aiks canvas patches.

jonahwilliams:
There will be more than one entity pass per frame if there are platform views, that is fine as we're not attempting to recycle the swapchain textures.

impeller/core/texture_descriptor.h
jonahwilliams: Done

impeller/core/texture_descriptor.h
```diff
@@ -66,6 +66,28 @@ struct TextureDescriptor {
...
+
+  constexpr bool operator!=(const TextureDescriptor& other) const {
+    return size != other.size ||                          //
+           storage_mode != other.storage_mode ||          //
+           format != other.format ||                      //
+           usage != other.usage ||                        //
+           sample_count != other.sample_count ||          //
+           type != other.type ||                          //
+           compression_type != other.compression_type ||  //
+           mip_count != other.mip_count;
```
bdero, line 88: Could this just be `return !(*this == other)`?

bdero (approved):
LGTM, just minor nits!

impeller/core/texture_descriptor.h
jonahwilliams: I guess this is a struct ... does that work? I don't know C++ 😆

bdero: Yeah it should work; classes are just structs with private members by default. :p

jonahwilliams: Oh, duh. I thought you were saying to do a ptr comparison or something like an implict equal operator. I didn't realize you were specifically talking about using the `==` operation to define the `!=` operation. Makes sense!

PR #44559, opened 2023-08-09
[Impeller] External OpenGLES texture sampling.

bdero:
Allows for sampling from external textures in the OpenGLES backend. DrawImageRect should Just Work if external textures are wrapped with the new `kTextureExternalOES` type.

### DIFF
--- a/impeller/compiler/types.h
+++ b/impeller/compiler/types.h
@@ -16,6 +16,8 @@
 namespace impeller {
 namespace compiler {
 
+const std::string kExternalTexturePrefix = "SAMPLER_EXTERNAL_";
+
 enum class SourceType {
   kUnknown,
   kVertexShader,
--- a/impeller/compiler/types.h
+++ b/impeller/compiler/types.h
@@ -16,8 +16,6 @@
 namespace impeller {
 namespace compiler {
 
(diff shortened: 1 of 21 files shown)


impeller/compiler/types.h
```diff
@@ -16,6 +16,8 @@
 namespace impeller {
 namespace compiler {
 
+const std::string kExternalTexturePrefix = "SAMPLER_EXTERNAL_";
```
johnmccutchan, line 19: nit:

- This should be constexpr
- We call it SAMPLER_EXTERNAL_ here but in TextureType it is ExternalOES -- would be nice if they shared the same name "SAMPLER_EXTERNAL_OES"
- Does this constant belong in types.h It's not a type, maybe it should be format.h right near the definition of TextureType::kTextureExternalOES so folks reading the code know these two concepts are tightly coupled.

johnmccutchan (approved):
Thank you!

impeller/compiler/types.h
bdero: * Switched to constexpr char*.
* Done.
* Moved it to a new `constants.h`; I don't think impeller's format types are used by the compiler right now.

bdero: (Although we could probably pull some of the format stuff from `core` into `base` at some point)

PR #44566, opened 2023-08-10
[Impeller] Flutter GPU: Add context override.

bdero:
Adds a way to inject a context override, which allows us to test the API in the Dart playground without spinning up an Engine/Shell.

### DIFF
--- a/impeller/renderer/renderer_dart_unittests.cc
+++ b/impeller/renderer/renderer_dart_unittests.cc
@@ -10,6 +10,7 @@
 #include "flutter/common/task_runners.h"
 #include "flutter/fml/backtrace.h"
 #include "flutter/fml/command_line.h"
+#include "flutter/lib/gpu/context.h"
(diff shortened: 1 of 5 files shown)


impeller/renderer/renderer_dart_unittests.cc
```diff
@@ -72,22 +79,16 @@ INSTANTIATE_PLAYGROUND_SUITE(RendererDartTest);
 
 TEST_P(RendererDartTest, CanRunDartInPlaygroundFrame) {
   auto isolate = GetIsolate();
-
-  SinglePassCallback callback = [&](RenderPass& pass) {
```
zanderso, line 76: Any reason not to retain this test, and add the new one as a new test?

bdero: Not really, added it back.

jonahwilliams (approved):
RSLGTM

PR #44571, opened 2023-08-10
Support for Android Platform Views under Impeller/Vulkan

johnmccutchan:
- Introduce AndroidHardwareBufferTextureSourceVK.
- Enable the necessary Vulkan extensions to support AndroidHardwareBufferTextureSourceVK.
- Refactor HardwareBufferExternalTextureGL into a base class and a GL and VK specializations.
- Switch ImageReaderPlatformView to use the PRIVATE image format (enables DRM'd content to be captured in the texture).
- Add a AHardwareBuffer_describe to NDKHelpers.
- Misc cleanups.


shell/platform/android/hardware_buffer_external_texture_vk.cc
```diff
@@ -0,0 +1,60 @@
...
+
+  AHardwareBuffer_Desc hb_desc = {};
+  flutter::NDKHelpers::AHardwareBuffer_describe(latest_hardware_buffer,
+                                                &hb_desc);
+
+  impeller::TextureDescriptor desc;
+  desc.storage_mode = impeller::StorageMode::kDevicePrivate;
+  desc.size = {static_cast<int>(bounds.width()),
+               static_cast<int>(bounds.height())};
+  desc.format = impeller::PixelFormat::kR8G8B8A8UNormInt;
```
jonahwilliams, line 44: Do we need to worry about checking the pixel format at runtime?

shell/platform/android/io/flutter/plugin/platform/ImageReaderPlatformViewRenderTarget.java
```diff
@@ -53,9 +53,9 @@ public void onImageAvailable(ImageReader reader) {
   protected ImageReader createImageReader33() {
     final ImageReader.Builder builder = new ImageReader.Builder(bufferWidth, bufferHeight);
     // Allow for double buffering.
-    builder.setMaxImages(2);
-    // Assume that we will be producing 32-bit RGBA values.
-    builder.setDefaultHardwareBufferFormat(HardwareBuffer.RGBA_8888);
+    builder.setMaxImages(3);
```
jonahwilliams, line 56: Are there any downsides to unconditionally requesting private format here? i.e. does it stop us from capturing the bytes via readback on the framework side (Picture/Layer.toImage)

jonahwilliams (approved):
LGTM with two questions

shell/platform/android/io/flutter/plugin/platform/ImageReaderPlatformViewRenderTarget.java
johnmccutchan: This is a good point, it definitely will stop us from reading it back.

I'm going to leave a TODO here to consider this in the future.

shell/platform/android/hardware_buffer_external_texture_vk.cc
johnmccutchan: Yes we do, I'm going to leave this as a TODO for now though.

jonahwilliams: 👍

PR #44579, opened 2023-08-10
[Android] Add support for text processing actions

bleroux:
## Description

This Android related PR adds a channel buffer and a plugin to interact with Android 'process text' feature. It makes it possible to query text processing actions and to run those actions (for instance 'calling' Google translate).
Text actions that outputs a processed text are supported.

The implementation is based on the great sample provided by @gualse , see https://github.com/flutter/flutter/issues/107603#issuecomment-1646629158. 

In order to return a non empty list of text actions, the implementation will require adding a section to the Android manifest file (see https://github.com/flutter/flutter/issues/107603#issuecomment-1683487087).
Adding this section automatically to new or existing Flutter apps is not part of this PR but will be tackled in a future PR.

## Related Issue

Android engine side for https://github.com/flutter/flutter/issues/107603

## Tests

Adds 3 tests.


justinmc (commented):
This approach LGTM.  I think it's good to keep this separate in its own method channel. 
 CC @camsim99 in case you have any quick thoughts from the Android side of things.

shell/platform/android/io/flutter/plugin/text/ProcessTextPlugin.java
```diff
@@ -0,0 +1,175 @@
...
+
+  @NonNull
+  private Map<Integer, MethodChannel.Result> requestsByCode =
+      new HashMap<Integer, MethodChannel.Result>();
+
+  public ProcessTextPlugin(@NonNull ProcessTextChannel processTextChannel) {
+    this.processTextChannel = processTextChannel;
+    this.packageManager = processTextChannel.packageManager;
+
+    this.cacheResolveInfos();
```
camsim99, line 40: I'm wondering if we can do this elsewhere because this will happen immediately when a `ProcessTextPlugin` is created, and that will happen regardless of whether or not this functionality is used. Would it make sense to be part of `queryTextActions()` or `processTextAction(...)`?

camsim99 (commented):
The approach looks good to me! Just a question on some constructor logic.

shell/platform/android/io/flutter/plugin/text/ProcessTextPlugin.java
```diff
@@ -0,0 +1,175 @@
...
+      return;
+    }
+
+    Intent intent = new Intent().setAction(Intent.ACTION_PROCESS_TEXT).setType("text/plain");
+
+    List<ResolveInfo> infos;
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
+      infos = packageManager.queryIntentActivities(intent, PackageManager.ResolveInfoFlags.of(0));
+    } else {
+      infos = packageManager.queryIntentActivities(intent, 0);
```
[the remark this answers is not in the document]
bleroux: The previous version is deprecated, the new one allow using 'long' values instead of 'int' for the previous one.
I will probabl add a comment above the call to the old one to mention it is deprecated.

shell/platform/android/io/flutter/plugin/text/ProcessTextPlugin.java
bleroux: Done

shell/platform/android/io/flutter/plugin/text/ProcessTextPlugin.java
```diff
@@ -0,0 +1,192 @@
...
+    this.processTextChannel = processTextChannel;
+    this.packageManager = processTextChannel.packageManager;
+
+    processTextChannel.setMethodHandler(this);
+  }
+
+  @Override
+  public Map<String, String> queryTextActions() {
+    if (resolveInfosById == null) {
+      resolveInfosById = new HashMap<String, ResolveInfo>();
```
camsim99, line 52: Nit: can this happen in `cacheResolveInfos` versus here?

shell/platform/android/io/flutter/plugin/text/ProcessTextPlugin.java
```diff
@@ -0,0 +1,192 @@
...
+    Intent intent = new Intent().setAction(Intent.ACTION_PROCESS_TEXT).setType("text/plain");
+
+    List<ResolveInfo> infos;
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
+      infos = packageManager.queryIntentActivities(intent, PackageManager.ResolveInfoFlags.of(0));
+    } else {
+      infos = packageManager.queryIntentActivities(intent, 0);
+    }
+
+    resolveInfosById.clear();
```
camsim99, line 120: Is this ever called where `resolveInfosById` wouldn't be an empty array? Wondering if we need this call.

camsim99 (approved):
Looks good, just some nits!

shell/platform/android/io/flutter/plugin/text/ProcessTextPlugin.java
bleroux: Yes, great suggestion, and doing so we can remove the `resolveInfosById.clear()`call your mention below.

PR #44587, opened 2023-08-10
Allow macOS plugins to register as app delegates

stuartmorgan:
Adds `addApplicationDelegate:` to the macOS plugin registrar, following the corresponding iOS method, and wires it up to the existing app delegation forwarding that was recently added for use at the application level. (The actual delegate is non-trivially different between iOS and macOS, but that's not resolveable without a complex migration on the iOS side, so the APIs currently diverge after the level of the `addApplicationDelegate:` method itself.)

This doesn't add any new methods to the delegation; those will be added in a follow-up PR.

Also fixes a retain cycle in the termination handler that prevented the new test from working.

Most of https://github.com/flutter/flutter/issues/41471

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/shell/platform/darwin/common/framework/Source/FlutterBinaryMessengerRelay.mm
+++ b/shell/platform/darwin/common/framework/Source/FlutterBinaryMessengerRelay.mm
@@ -14,7 +14,7 @@ FLUTTER_ASSERT_ARC
 - (instancetype)initWithParent:(NSObject<FlutterBinaryMessenger>*)parent {
   self = [super init];
   if (self != nil) {
-    self.parent = parent;
+    _parent = parent;
   }
   return self;
 }
--- a/shell/platform/darwin/macos/framework/Headers/FlutterAppDelegate.h
+++ b/shell/platform/darwin/macos/framework/Headers/FlutterAppDelegate.h
@@ -11,6 +11,30 @@
 #import "FlutterMacros.h"
 
 /**
+ * A protocol to be implemented by the `NSApplicationDelegate` of an application to enable the
+ * Flutter framework and any Flutter plugins to register to receive application life cycle events.
+ *
+ * Implementers should forward all of the `NSApplicationDelegate` methods corresponding to the
+ * handlers in FlutterAppLifecycleDelegate to any registered delegates.
+ */
+FLUTTER_DARWIN_EXPORT
+@protocol FlutterAppLifecycleProvider <NSObject>
+
+/**
+ * Adds an object implementing |FlutterAppLifecycleDelegate| to the list of
+ * delegates to be informed of application lifecycle events.
+ */
+- (void)addApplicationLifecycleDelegate:(nonnull NSObject<FlutterAppLifecycleDelegate>*)delegate;
+
+/**
+ * Removes an object implementing |FlutterAppLifecycleDelegate| to the list of
+ * delegates to be informed of application lifecycle events.
+ */
+- (void)removeApplicationLifecycleDelegate:(nonnull NSObject<FlutterAppLifecycleDelegate>*)delegate;
+
+@end
+
+/**
  * |NSApplicationDelegate| subclass for simple apps that want default behavior.
  *
  * This class implements the following behaviors:
@@ -20,13 +44,14 @@
  *   * Updates the main Flutter window's title to match the name in the app's Info.plist.
  *     |mainFlutterWindow| must be set before the application finishes launching for this to take
  *     effect.
+ *   * Forwards `NSApplicationDelegate` callbacks to plugins that register for them.
  *
  * App delegates for Flutter applications are *not* required to inherit from
  * this class. Developers of custom app delegate classes should copy and paste
  * code as necessary from FlutterAppDelegate.mm.
  */
 FLUTTER_DARWIN_EXPORT
-@interface FlutterAppDelegate : NSObject <NSApplicationDelegate>
+@interface FlutterAppDelegate : NSObject <NSApplicationDelegate, FlutterAppLifecycleProvider>
 
 /**
  * The application menu in the menu bar.
@@ -38,18 +63,6 @@ FLUTTER_DARWIN_EXPORT
  * primarily intended for use in single-window applications.
  */
 @property(weak, nonatomic) IBOutlet NSWindow* mainFlutterWindow;
-
-/**
- * Adds an object implementing |FlutterAppLifecycleDelegate| to the list of
- * delegates to be informed of application lifecycle events.
- */
-- (void)addApplicationLifecycleDelegate:(NSObject<FlutterAppLifecycleDelegate>*)delegate;
-
-/**
- * Removes an object implementing |FlutterAppLifecycleDelegate| to the list of
- * delegates to be informed of application lifecycle events.
- */
-- (void)removeApplicationLifecycleDelegate:(NSObject<FlutterAppLifecycleDelegate>*)delegate;
 
 @end
 
--- a/shell/platform/darwin/macos/framework/Headers/FlutterPluginMacOS.h
+++ b/shell/platform/darwin/macos/framework/Headers/FlutterPluginMacOS.h
@@ -4,6 +4,7 @@
 
 #import <Foundation/Foundation.h>
 
+#import "FlutterAppLifecycleDelegate.h"
 #import "FlutterChannels.h"
 #import "FlutterCodecs.h"
 #import "FlutterMacros.h"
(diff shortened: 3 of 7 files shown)


stuartmorgan:
This is now tested and ready for review.

shell/platform/darwin/common/framework/Source/FlutterBinaryMessengerRelay.mm
```diff
@@ -14,7 +14,7 @@ @implementation FlutterBinaryMessengerRelay
 - (instancetype)initWithParent:(NSObject<FlutterBinaryMessenger>*)parent {
   self = [super init];
   if (self != nil) {
-    self.parent = parent;
+    _parent = parent;
```
stuartmorgan, line 17: This is unrelated, but as discussed in Discord I accidentally duplicated the recent work to adopt FlutterBinaryMessengerRelay on macOS as part of debugging my delloc test not working, and this was the only substantive diff between our versions. `init` methods should not call `self` (which we violate a lot unfortunately, but this one was trivial to fix), so I had done a drive-by fix to this as part of moving it.

I can revert if you'd rather.

shell/platform/darwin/macos/framework/Headers/FlutterAppDelegate.h
```diff
@@ -10,6 +10,30 @@
...
 
+/**
+ * A protocol to be implemented by the `NSApplicationDelegate` of an application to enable the
+ * Flutter framework and any Flutter plugins to register to receive application life cycle events.
+ *
+ * Implementers should forward all of the `NSApplicationDelegate` methods corresponding to the
+ * handlers in FlutterAppLifecycleDelegate to any registered delegates.
+ */
+FLUTTER_DARWIN_EXPORT
+@protocol FlutterAppLifecycleProvider <NSObject>
```
stuartmorgan, line 21: This change isn't strictly necessary, but it makes us match iOS, and it makes the engine code that uses these methods cleaner.

shell/platform/darwin/macos/framework/Headers/FlutterPluginMacOS.h
```diff
@@ -22,7 +23,7 @@ NS_ASSUME_NONNULL_BEGIN
  * expand over time to more closely match the functionality of the iOS FlutterPlugin.
  */
 FLUTTER_DARWIN_EXPORT
-@protocol FlutterPlugin <NSObject>
+@protocol FlutterPlugin <NSObject, FlutterAppLifecycleDelegate>
```
stuartmorgan, line 26: This is to match iOS. Personally I would have not done this, and had plugins that also wanted to be `FlutterAppLifecycleDelegate`s implement this themselves, but the ship has sailed for iOS and there's not really much downside, so I'm opting for consistency with iOS (as I've tried to do in plugin APIs in general, to minimize pain in code sharing in plugins).

shell/platform/darwin/macos/framework/Headers/FlutterPluginRegistrarMacOS.h
```diff
@@ -52,6 +52,13 @@ FLUTTER_DARWIN_EXPORT
 - (void)addMethodCallDelegate:(nonnull id<FlutterPlugin>)delegate
                       channel:(nonnull FlutterMethodChannel*)channel;
 
+/**
+ * Registers the plugin as a receiver of `NSApplicationDelegate` calls.
+ *
+ * @param delegate The receiving object, such as the plugin's main class.
+ */
+- (void)addApplicationDelegate:(nonnull NSObject<FlutterAppLifecycleDelegate>*)delegate;
```
stuartmorgan, line 60: This is a slight deviation from iOS, which uses `FlutterPlugin` as the argument type, but that shouldn't prevent any code sharing.

shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
```diff
@@ -165,7 +170,7 @@ - (void)handleMethodCall:(FlutterMethodCall*)call result:(FlutterResult)result;
 #pragma mark -
 
 @implementation FlutterEngineTerminationHandler {
-  FlutterEngine* _engine;
+  __weak FlutterEngine* _engine;
```
stuartmorgan, line 173: This fixes a retain cycle that will affect essentially all real-word, non-add-to-app cases (and was breaking my new test).

shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
```diff
@@ -448,12 +465,12 @@ - (instancetype)initWithName:(NSString*)labelPrefix
   [self setUpAccessibilityChannel];
   [self setUpNotificationCenterListeners];
   id<NSApplicationDelegate> appDelegate = [[NSApplication sharedApplication] delegate];
-  const SEL selector = @selector(addApplicationLifecycleDelegate:);
-  if ([appDelegate respondsToSelector:selector]) {
+  if ([appDelegate conformsToProtocol:@protocol(FlutterAppLifecycleProvider)]) {
```
stuartmorgan, line 468: This is a cleaner way to express the check, and also is what the iOS engine does.

shell/platform/darwin/macos/framework/Source/FlutterEngineTest.mm
```diff
@@ -515,7 +572,7 @@ - (NSApplicationTerminateReply)applicationShouldTerminate:(NSApplication* _Nonnu
                ICUDataPath:[fixtures stringByAppendingString:@"/icudtl.dat"]];
     FlutterEngine* engine = [[FlutterEngine alloc] initWithName:@"test"
                                                         project:project
-                                         allowHeadlessExecution:true];
+                                         allowHeadlessExecution:YES];
```
stuartmorgan, line 575: Drive-by, noticed while comparing your test to my non-working test (now fixed).

shell/platform/darwin/common/framework/Source/FlutterBinaryMessengerRelay.mm
cbracken: Ah good call. Thanks.

shell/platform/darwin/macos/framework/Headers/FlutterPluginMacOS.h
cbracken: Agreed - getting the headers as similar as possible both for consistency and with an eye to a future world in which we can merge the non-UIKit/AppKit specific parts implies that we've got to live with the union of past decisions across both that we might prefer to not have made in retrospect.

shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
cbracken: Thanks for spotting this.

shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
cbracken: Yeah less raw selector wrangling is always a plus.

shell/platform/darwin/macos/framework/Source/FlutterEngineTest.mm
```diff
@@ -59,6 +63,59 @@ - (NSApplicationTerminateReply)applicationShouldTerminate:(NSApplication* _Nonnu
...
+  std::vector<void*> _delegates;
+}
+
+- (void)addApplicationLifecycleDelegate:(nonnull NSObject<FlutterAppLifecycleDelegate>*)delegate {
+  _delegates.push_back((__bridge void*)delegate);
+}
+
+- (void)removeApplicationLifecycleDelegate:
+    (nonnull NSObject<FlutterAppLifecycleDelegate>*)delegate {
+  auto i = std::find(_delegates.begin(), _delegates.end(), (__bridge void*)delegate);
```
cbracken, line 95: Maybe `delegateIndex` for the minor readability benefit given the type is `auto`?

shell/platform/darwin/macos/framework/Source/FlutterEngineTest.mm
cbracken: Ugh - thanks. I thought I'd done a sweep of these at one point but clearly missed a few. Will re-check today; maybe I never committed that branch?

cbracken (approved):

shell/platform/darwin/macos/framework/Source/FlutterEngineTest.mm
stuartmorgan: Done.

PR #44176, opened 2023-07-31
[ios][autocorrection]disable auto-correction highlight in iOS 17

hellohuanlin:
This PR disables the "auto-correction highlight" feature introduced in https://github.com/flutter/flutter/pull/45354 and https://github.com/flutter/engine/pull/13959/ 

Note that auto-correction still works, it's only the "highlight" part is disabled. 

(CC: @LongCatIsLooong who was the original author)

## Why remove this feature?

The original PR uses `UITextInput::firstRectForRange` API for auto-correction, since Apple does not provide any other API when auto-correction should show up, so that PR used this API as a workaround. 

In iOS 17, Apple changed a few `UITextInput` behaviors:  
- UIKit does not query `UITextInput::firstRectForRange` for text range of the auto-corrected word. 
- But instead, it repeatedly queries every single character of the current word (after entering or deleting a character), regardless whether the word should be auto-corrected or not. 

I have tried all other `UITextInput` APIs that takes a text range, and none are suitable for auto-correction feature. As a result, I have to disable this feature for iOS 17 for now. 

I have created a new issue (TODO) to find other approaches to re-enable this feature. 

*List which issues are fixed by this PR. You must list at least one issue.*

Fixes https://github.com/flutter/flutter/issues/128406

*If you had to change anything in the [flutter/tests] repo, include a link to the migration guide as per the [breaking change policy].*

## Pre-launch Checklist

- [ ] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [ ] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [ ] I read and followed the [Flutter Style Guide] and the [C++, Objective-C, Java style guides].
- [ ] I listed at least one issue that this PR fixes in the description above.
- [ ] I added new tests to check the change I am making or feature I am adding, or Hixie said the PR is test-exempt. See [testing the engine] for instructions on writing and running engine tests.
- [ ] I updated/added relevant documentation (doc comments with `///`).
- [ ] I signed the [CLA].
- [ ] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/flutter/wiki/Tree-hygiene#overview
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[...]

PR #44604, opened 2023-08-10
[ios][iOS 17] fix auto-correction highlight on top left corner

hellohuanlin:
Fix native auto-correction highlight region on top left corner 

I did some github archaeology - some code changes in `firstRectForRange` are related to [this PR](https://github.com/flutter/flutter/issues/57914) (CC: @LongCatIsLooong). I have verified that it still works. 

Some investigation: 

1. In iOS 17, the size of the highlight region matches the size of the prefix of words to be auto-corrected, except for the last character. For example, in the following 2 screenshot, the width of letter `i` and `w` are different, but this observation still holds. This is likely due to internal implementation of UIKit, and it's hard to guess why exactly Apple does this. 

Screenshot 2023-08-10 at 1 03 15 PM

2. The highlight region is drawn with respect to the coordinate of our FlutterTextInputView. However, it's not a subview, so likely a draw call on its CALayer. 

3. As discussed in [this PR](https://github.com/flutter/engine/pull/44176), Apple does not provide any API to tell us when a word should be auto-corrected or not. This means under current API, it is unlikely possible to leverage the system highlight to replace the our custom highlight. We have to find other solutions. As a workaround, I put the FlutterTextInputView outside of the screen. 

4. This happens on previous iOS as well, but just not as noticeable (much smaller, see screenshot below). So we should have this workaround for both iOS 16 and iOS 17. 

ios 16 2

5. The same workaround also fixes another issue that shows the native auto-correction context menu (https://github.com/flutter/flutter/issues/130170 and https://github.com/flutter/flutter/issues/130818). But it only happens on iOS 17, and not iOS 16. In the future it's worth investigating whether we can leverage this native menu (tho at that time we still have to solve the highlight region problem). 

## Pre-launch Checklist

- [x] I signed the [CLA].

Fixes https://github.com/flutter/flutter/issues/131695
Fixes https://github.com/flutter/flutter/issues/130818

### DIFF
--- a/shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
+++ b/shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
@@ -1625,10 +1625,13 @@ static BOOL IsSelectionRectBoundaryCloserToPoint(CGPoint point,
   NSUInteger start = ((FlutterTextPosition*)range.start).index;
   NSUInteger end = ((FlutterTextPosition*)range.end).index;
   if (_markedTextRange != nil) {
+    UIView* hostView = _textInputPlugin.hostView;
+    NSAssert(hostView == nil || [self isDescendantOfView:hostView], @"%@ is not a descendant of %@",
+             self, hostView);
     // The candidates view can't be shown if the framework has not sent the
     // first caret rect.
     if (CGRectEqualToRect(kInvalidFirstRect, _markedRect)) {
-      return kInvalidFirstRect;
+      return hostView ? [hostView convertRect:kInvalidFirstRect toView:self] : kInvalidFirstRect;
     }
 
     if (CGRectEqualToRect(_cachedFirstRect, kInvalidFirstRect)) {
@@ -1642,9 +1645,6 @@ static BOOL IsSelectionRectBoundaryCloserToPoint(CGPoint point,
       _cachedFirstRect = [self localRectFromFrameworkTransform:rect];
     }
 
-    UIView* hostView = _textInputPlugin.hostView;
-    NSAssert(hostView == nil || [self isDescendantOfView:hostView], @"%@ is not a descendant of %@",
-             self, hostView);
     return hostView ? [hostView convertRect:_cachedFirstRect toView:self] : _cachedFirstRect;
   }
 
@@ -2461,6 +2461,16 @@ static BOOL IsSelectionRectBoundaryCloserToPoint(CGPoint point,
     _activeView.frame =
         CGRectMake(0, 0, [dictionary[@"width"] intValue], [dictionary[@"height"] intValue]);
     _activeView.tintColor = [UIColor clearColor];
+  } else {
+    // View must be loaded at this point.
+    UIScreen* screen = _viewController.flutterScreenIfViewLoaded;
+
+    // Position FlutterTextInputView outside of the screen (if scribble is disabled).
+    // This is to fix a bug where native auto-correction highlight is displayed on
+    // top left corner of the screen (See: https://github.com/flutter/flutter/issues/131695)
+    // and a bug where the native auto-correction suggestion menu displayed (See:
+    // https://github.com/flutter/flutter/issues/130818).
+    _inputHider.frame = CGRectMake(0, -screen.bounds.size.height, 0, 0);
   }
 }
 
--- a/shell/platform/darwin/ios/framework/Source/FlutterTextInputPluginTest.mm
+++ b/shell/platform/darwin/ios/framework/Source/FlutterTextInputPluginTest.mm
@@ -61,6 +61,7 @@ FLUTTER_ASSERT_ARC
 
 @interface FlutterTextInputPlugin ()
 @property(nonatomic, assign) FlutterTextInputView* activeView;
+@property(nonatomic, readonly) UIView* inputHider;
 @property(nonatomic, readonly) UIView* keyboardViewContainer;
 @property(nonatomic, readonly) UIView* keyboardView;
 @property(nonatomic, assign) UIView* cachedFirstResponder;
@@ -1426,44 +1427,53 @@ FLUTTER_ASSERT_ARC
     @(-6.0), @(3.0), @(9.0), @(1.0)
   ];
 
+  CGRect kInvalidFirstRectRelative =
+      [textInputPlugin.viewController.view convertRect:kInvalidFirstRect toView:inputView];
+
   // Invalid since we don't have the transform or the rect.
-  XCTAssertTrue(CGRectEqualToRect(kInvalidFirstRect, [inputView firstRectForRange:range]));
+  XCTAssertTrue(CGRectEqualToRect(kInvalidFirstRectRelative, [inputView firstRectForRange:range]));
 
   [inputView setEditableTransform:yOffsetMatrix];
   // Invalid since we don't have the rect.
-  XCTAssertTrue(CGRectEqualToRect(kInvalidFirstRect, [inputView firstRectForRange:range]));
+  XCTAssertTrue(CGRectEqualToRect(kInvalidFirstRectRelative, [inputView firstRectForRange:range]));
 
   // Valid rect and transform.
   CGRect testRect = CGRectMake(0, 0, 100, 100);
   [inputView setMarkedRect:testRect];
 
   CGRect finalRect = CGRectOffset(testRect, 0, 200);
-  XCTAssertTrue(CGRectEqualToRect(finalRect, [inputView firstRectForRange:range]));
+  CGRect finalRectRelative = [textInputPlugin.viewController.view convertRect:finalRect
+                                                                       toView:inputView];
+  XCTAssertTrue(CGRectEqualToRect(finalRectRelative, [inputView firstRectForRange:range]));
   // Idempotent.
-  XCTAssertTrue(CGRectEqualToRect(finalRect, [inputView firstRectForRange:range]));
+  XCTAssertTrue(CGRectEqualToRect(finalRectRelative, [inputView firstRectForRange:range]));
 
   // Use an invalid matrix:
   [inputView setEditableTransform:zeroMatrix];
   // Invalid matrix is invalid.
-  XCTAssertTrue(CGRectEqualToRect(kInvalidFirstRect, [inputView firstRectForRange:range]));
-  XCTAssertTrue(CGRectEqualToRect(kInvalidFirstRect, [inputView firstRectForRange:range]));
+  XCTAssertTrue(CGRectEqualToRect(kInvalidFirstRectRelative, [inputView firstRectForRange:range]));
+  XCTAssertTrue(CGRectEqualToRect(kInvalidFirstRectRelative, [inputView firstRectForRange:range]));
 
   // Revert the invalid matrix change.
   [inputView setEditableTransform:yOffsetMatrix];
   [inputView setMarkedRect:testRect];
-  XCTAssertTrue(CGRectEqualToRect(finalRect, [inputView firstRectForRange:range]));
+  XCTAssertTrue(CGRectEqualToRect(finalRectRelative, [inputView firstRectForRange:range]));
 
   // Use an invalid rect:
   [inputView setMarkedRect:kInvalidFirstRect];
   // Invalid marked rect is invalid.
-  XCTAssertTrue(CGRectEqualToRect(kInvalidFirstRect, [inputView firstRectForRange:range]));
-  XCTAssertTrue(CGRectEqualToRect(kInvalidFirstRect, [inputView firstRectForRange:range]));
+  XCTAssertTrue(CGRectEqualToRect(kInvalidFirstRectRelative, [inputView firstRectForRange:range]));
+  XCTAssertTrue(CGRectEqualToRect(kInvalidFirstRectRelative, [inputView firstRectForRange:range]));
 
   // Use a 3d affine transform that does 3d-scaling, z-index rotating and 3d translation.
   [inputView setEditableTransform:affineMatrix];
   [inputView setMarkedRect:testRect];
-  XCTAssertTrue(
-      CGRectEqualToRect(CGRectMake(-306, 3, 300, 300), [inputView firstRectForRange:range]));
+
+  CGRect relativeRect =
+      [textInputPlugin.viewController.view convertRect:CGRectMake(-306, 3, 300, 300)
+                                                toView:inputView];
+
+  XCTAssertTrue(CGRectEqualToRect(relativeRect, [inputView firstRectForRange:range]));
 
   NSAssert(inputView.superview, @"inputView is not in the view hierarchy!");
   const CGPoint offset = CGPointMake(113, 119);
@@ -1472,8 +1482,8 @@ FLUTTER_ASSERT_ARC
   inputView.frame = currentFrame;
   // Moving the input view within the FlutterView shouldn't affect the coordinates,
   // since the framework sends us global coordinates.
-  XCTAssertTrue(CGRectEqualToRect(CGRectMake(-306 - 113, 3 - 119, 300, 300),
-                                  [inputView firstRectForRange:range]));
+  CGRect target = CGRectOffset(relativeRect, -113, -119);
+  XCTAssertTrue(CGRectEqualToRect(target, [inputView firstRectForRange:range]));
 }
 
 - (void)testFirstRectForRangeReturnsCorrectSelectionRect {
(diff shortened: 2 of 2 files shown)


shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
```diff
@@ -1625,10 +1632,13 @@ - (CGRect)firstRectForRange:(UITextRange*)range {
...
   NSUInteger end = ((FlutterTextPosition*)range.end).index;
   if (_markedTextRange != nil) {
+    UIView* hostView = _textInputPlugin.hostView;
+    NSAssert(hostView == nil || [self isDescendantOfView:hostView], @"%@ is not a descendant of %@",
+             self, hostView);
     // The candidates view can't be shown if the framework has not sent the
     // first caret rect.
     if (CGRectEqualToRect(kInvalidFirstRect, _markedRect)) {
-      return kInvalidFirstRect;
+      return hostView ? [hostView convertRect:kInvalidFirstRect toView:self] : kInvalidFirstRect;
```
hellohuanlin, line 1641: @LongCatIsLooong I'm a bit iffy on this one and could use another pair of eyes: 

My understanding of [your previous PR](https://github.com/flutter/engine/pull/36707) is that this should return rects in FlutterTextInputView's coord space - not only valid rects, but also this dummy invalid rect (because otherwise an invalid rect could very well become a valid rect with some transforms of FlutterTextInputView).

shell/platform/darwin/ios/framework/Source/FlutterTextInputPluginTest.mm
```diff
@@ -1426,44 +1427,53 @@ - (void)testUpdateFirstRectForRange {
     @(-6.0), @(3.0), @(9.0), @(1.0)
   ];
 
+  CGRect kInvalidFirstRectRelative =
+      [textInputPlugin.viewController.view convertRect:kInvalidFirstRect toView:inputView];
```
hellohuanlin, line 1431: Wanna highlight changes to this test. Let me know if it makes sense @LongCatIsLooong

stuartmorgan:
I'm still unclear on why we can't make this work correctly. If the size is exactly correct, it's clearly based on rects that we are providing; why can't we ensure that we are providing correct location in addition to size so that the highlight actually works?

hellohuanlin:
> I'm still unclear on why we can't make this work correctly. If the size is exactly correct, it's clearly based on rects that we are providing; why can't we ensure that we are providing correct location in addition to size so that the highlight actually works?

@stuartmorgan Because the size is 1 character shorter, and there's no reliable API to tell whether the highlight is shown or not (If there were such API, we would be able to maybe manually highlight the last character).

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
LongCatIsLooong: Yeah I think `kInvalidFirstRect` is probably specified in the window's coordinate system so it needs to be transformed. The input view used to be placed at the top left corner of the window iirc.

stuartmorgan:
Leaving aside the issue of it missing the last character: do we know why the rect is in completely the place, and can we fix that?

Then for the last-character issue, have we made a reduced test case and filled an issue with Apple?

hellohuanlin:
> Leaving aside the issue of it missing the last character: do we know why the rect is in completely the wrong place, and can we fix that? Are we providing bogus offsets in our implementation of the query method?

To clarify, our implementation of the query method (`firstRectForRange`) is providing the correct rects. It's placed at the wrong place because we put `FlutterTextInputView` on top left corner. 

The reason I chose to hide it in this PR is because **the size** doesn't match (missing 1 char). Other than that, we can easily position the highlight at the right place (pretty much the same logic I wrote to position it outside screen). 

> Then for the last-character issue, have we made a reduced test case and filled an issue with Apple?

I have not filed a radar with Apple. It's a good idea.

hellohuanlin:
> The reason I chose to hide it in this PR is because the size doesn't match (missing 1 char). Other than that, we can easily position the highlight at the right place (pretty much the same logic I wrote to position it outside screen).

@stuartmorgan Actually, while I was writing this, it came to my mind that it might be better UX to show a **slightly** smaller highlight region, than not showing it at all. Let me try and see how it looks.

stuartmorgan:
> To clarify, our implementation of the query method (`firstRectForRange`) is providing the correct rects [...] The reason it is placed at the wrong place is because we put the `FlutterTextInputView` on top left corner.

If the view that we are telling UIKit the words are relative to is in the top left corner, and we're returning rects that don't have the correct offset for the word relative to the top left corner of the Flutter view, then we aren't returning the correct rects. Am I misunderstanding here?

hellohuanlin:
> If the view that we are telling UIKit the words are relative to is in the top left corner, and we're returning rects that don't have the correct offset for the word relative to the top left corner of the Flutter view, then we aren't returning the correct rects. Am I misunderstanding here?

Oh the API should return the rects wrt the coordinate space of UITextInput itself, so I think we did it right. 

> Actually, while I was writing this, it came to my mind that it might be better UX to show a slightly smaller highlight region, than not showing it at all. Let me try and see how it looks.

I take back my word - Just tried it, looks pretty bad: 

https://github.com/flutter/engine/assets/41930132/395e4b8f-d6a4-402c-9eff-712909e7899d

stuartmorgan:
> > If the view that we are telling UIKit the words are relative to is in the top left corner, and we're returning rects that don't have the correct offset for the word relative to the top left corner of the Flutter view, then we aren't returning the correct rects. Am I misunderstanding here?
> 
> Oh the API should return the rects wrt the coordinate space of UITextInput itself, so I think we did it right.

I'm not following. You're saying that we are returning the actual offsets of the queried text, relative to the upper-left corner of the Flutter view (where the UITextInput is positioned), and yet UIKit is drawing them in a random location instead?

hellohuanlin:
> I'm not following. You're saying that we are returning the actual offsets of the queried text, relative to the upper-left corner of the Flutter view (where the UITextInput is positioned), and yet UIKit is drawing them in a random location instead?

We are returning the offset of queried text, relative to the upper-left corner of **UITextInput**, which happens to be positioned at the origin of FlutterView. 

UIKit is drawing the highlight relative to **UITextInput** (not randomly). So in this PR, I moved this UITextInput outside screen.

stuartmorgan:
> We are returning the offset of queried text, relative to the upper-left corner of **UITextInput**, which happens to be positioned at the origin of FlutterView.
> 
> UIKit is drawing the highlight relative to **UITextInput** (not randomly).

I do not understand how both of these things can possibly be true.

If we were returning the actual offset of the words relative to the origin of the Flutter view (where the UITextInput is), and UIKit were drawing the rect using that same offset and origin, then the rectangle would be in the correct location. It's not, so either one of those things is not true, or something else is wrong that I still don't understand.

stuartmorgan:
Looking at the code here more closely, I see that the beginning of `firstRectForRange:` is assertions about the class types of the inputs being our own classes. That means that the values UIKit is passing us to query the rect are values *we* provided from some other part of the text input protocol. Given that, do we actually know that the missing last character is an iOS bug, and not a bug in our own implementation of some other part of the protocol, returning the wrong position (and thus leading to incorrect queries to `firstRectForRange:` using our own bad data)?

hellohuanlin:
> If we were returning the actual offset of the words relative to the origin of the Flutter view (where the UITextInput is), and UIKit were drawing the rect using that same offset and origin, then the rectangle would be in the correct location. It's not, so either one of those things is not true, or something else is wrong that I still don't understand.

Sorry I think I misread your previous message - we were not returning the offset of the actual rendered words (by flutter), because UITextInput is at the wrong place. In my mental model, UITextInput should have been placed at the same position as the actual rendered words, that's why I said UITextInput itself is at the wrong position, but it's returning the right rects in query APIs (which was indeed confusing). 

If we were to keep UITextInput to stay at the origin of FlutterView, then we would need to update all offsets to make it right. Tough based on my understanding of existing code, iPad scribble likely requires UITextInput to overlap with the actual rendered text (my mental model). 

> Looking at the code here more closely, I see that the beginning of firstRectForRange: is assertions about the class types of the inputs being our own classes. That means that the values UIKit is passing us to query the rect are values we provided from some other part of the text input protocol. Given that, do we actually know that the missing last character is an iOS bug, and not a bug in our own implementation of some other part of the protocol, returning the wrong position (and thus leading to incorrect queries to firstRectForRange: using our own bad data)?

We are querying the correct range (every letter of the last word, including the last letter). For example, in the following screenshot, the input value is `a texx`, and the range of misspelt word is [2, 6), and as you can see in the logging below, UIKit queried the range for `t`, `e`, `x`, `x` one by one. 

Screenshot 2023-08-11 at 10 49 15 AM

justinmc (commented):
I'm interested in understanding more about why the highlight lags by 1 character if you know any more details @hellohuanlin.  It would be great if we could fix that rather than hiding it off screen.  If that's not possible then this approach looks good to me, though.

I wonder if some state is out of date (`_markedRect`, `text`?) at the time that firstRectsForRange is called.

  > I do not understand how both of these things can possibly be true.

My understanding is (before this PR) we returned the position of the text relative to the UITextInput correctly, but the UITextInput was just stuck naively at 0,0 rather than at the location of the TextField.  We could easily put the UITextInput at the correct location of the TextField, but then the highlight would show up, and it looks wrong since it lags by 1 character.  This PR moves UITextInput offscreen to avoid this.

So the real problem we have here seems to be the 1 character lag in the autocorrect highlight, if I'm understanding correctly.

justinmc:
(I posted my review without seeing @hellohuanlin's reply above)

So firstRectsForRange is called one by one for each of the characters that should be highlighted, and we return the correct rects for all of those characters, but mysteriously the last character is not highlighted?

stuartmorgan:
Okay, so to make sure I'm understanding correctly: when you said:

> We are returning the offset of queried text, relative to the upper-left corner of **UITextInput**

what you actually meant is that we are returning the offset relative to the Flutter text field, *not* relative to the UITextInput?

(That was my assumption based on the visual behavior, but I was confused by your assertions to the contrary.)

> For example, in the following screenshot, the input value is `a texx`, and the range of misspelt word is `[2, 6)`, and as you can see in the logging below, UIKit queried the range for `t`, `e`, `x`, `x` one by one.

And you've validated that we are in fact returning the correctly sized and positioned rect for the final `x` in that case? We're not, e.g., end up in an invalid-rect-return codepath for that?

Also, is this specific to the very last character of the text field? I.e., if you type `a text field`, then go back and delete `text` then re-type it as `texx`, does it have the same behavior, or not?

hellohuanlin:
> I wonder if some state is out of date (_markedRect, text?) at the time that firstRectsForRange is called.

They look right to me. Note that `_markedRect` is actually for CJK languages' intermediate state (you can see example in [this PR's description](https://github.com/flutter/flutter/issues/57914)). 

> So firstRectsForRange is called one by one for each of the characters that should be highlighted, and we return the correct rects for all of those characters, but mysteriously the last character is not highlighted?

Yes, we return the correct rects (if UITextInput is overlapping the flutter rendered text, rather than at the origin of FlutterView).

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
```diff
@@ -34,6 +34,13 @@
 // returns kInvalidFirstRect, iOS will not show the IME candidates view.
 const CGRect kInvalidFirstRect = {{-1, -1}, {9999, 9999}};
 
+// Position FlutterTextInputView outside of the screen (if scribble is disabled).
+// This is to fix a bug where native auto-correction highlight is displayed on
+// top left corner of the screen (See: https://github.com/flutter/flutter/issues/131695)
+// and a bug where the native auto-correction suggestion menu displayed (See:
+// https://github.com/flutter/flutter/issues/130818).
+const CGRect kNonScribbleInputHiderOffScreenRect = {{0, -2000}, {0, 0}};
```
stuartmorgan, line 42: If we need to keep the off-screen hack, please use a dynamic value that is definitively off-screen based on iterating all `UIScreen` objects and uninioning their bounds.

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
```diff
@@ -34,6 +34,13 @@
 // returns kInvalidFirstRect, iOS will not show the IME candidates view.
 const CGRect kInvalidFirstRect = {{-1, -1}, {9999, 9999}};
 
+// Position FlutterTextInputView outside of the screen (if scribble is disabled).
```
stuartmorgan, line 37: What effect does this have on IME display? Or do we do all of that via Flutter UI?

LongCatIsLooong:
It sounds like what's happening could be, when the user types (btw are you using the software keyboard or the hardware keyboard?):

1. the key event is sent to the framework and the framework processes the platform channel message during the idle phase, sending back the updated text editing state without waiting for the text layout process to take place.
2. the ios text input plugin receives the updated text editing state and that triggers the software keyboard implementation to query the "firstRectsForRange" using the updated text editing state.
3. the framework starts to layout the text in TextPainter, sending back the bounding boxes for visible characters in composite phase, or in a post frame callback. In either case it will be too late since the software keyboard has already called the "firstRectsForRange" method and got the outdated bounding boxes

hellohuanlin:
> what you actually meant is that we are returning the offset relative to the Flutter text input widget, not relative to the UITextInput?

@stuartmorgan Well, I was actually thinking about `UITextInput`, not Flutter widget. 

In my mental model, `UITextInput` is an "invisible" native text on top left corner, of exactly the same size as the text drawn on screen. If we were to implement the `UITextInput` in native iOS, we would draw that invisible native text out, like we are implementing our own version of `UITextField` or `UITextView`. But since Flutter handles the text drawing, we do not need to draw it on native side. 

Previously, we did not need this "invisible native text" to overlap with the Flutter drawing, so we put it anywhere (conveniently, top left corner). But later, when iPad scribble is enabled, we put this "invisible native text" to overlap with the Flutter drawing on screen. In either scribble or non-scribble case, the range query API returns the same offset, relative to UITextInput (just the position of UITextInput itself is different). 

> And you've validated that we are in fact returning the correctly sized and positioned rect for the final x in that case? We're not, e.g., ending up in an invalid-rect-return codepath for that?

Yeah I have manually checked all rects. But let me try again (and post the result here in case I made terrible mistakes)

> Also, is this specific to the very last character of the text field? I.e., if you type a text field, then go back and delete text then re-type it as texx, does it have the same behavior, or not?

Let me try this too.

stuartmorgan:
> If we were to implement the `UITextInput` in native iOS, we would draw that invisible native text out, like we are implementing our own version of `UITextField` or `UITextView`. But since Flutter handles the text drawing, we do not need to draw it on native side.

Yes, which means the actual text only has one position on the screen: the position where Flutter draws it. I think it would help if any discussion of the position of text referred to the position of *the actual text* that has a position, not to the theoretical position of theoretical text that does not actually exist.

> In either scribble or non-scribble case, the range query API returns the same offset, relative to `UITextInput` (just the position of `UITextInput` itself is different in these 2 cases).

I understand that what we return needs to be the offset relative to the `UITextInput`. My point above was that referring to the position we are returning as "correct" in the case where what we are returning is **not** the position of the *the actual text on screen* relative to *the actual `UITextInput` location* is extremely confusing.

hellohuanlin:
> My point above was that referring to the position we are returning as "correct" in the case where what we are returning is not the position of the the actual text on screen relative to the actual UITextInput location is extremely confusing.

Yeah agree it was confusing. I missed a big "if" - it's correct **if** the UITextInput overlaps with the text drawn on screen (which is the iPad scribble case). Sorry I think i paid most attention on the `size` part and somehow overlooked the subtle `offset` part.

hellohuanlin:
> It would be great if we could fix that rather than hiding it off screen. If that's not possible then this approach looks good to me, though.

I think I will dig a few more days, and have this PR as the backup plan, in case no luck before the release branch cut.

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
hellohuanlin: I can confirm that IME still works. Below is IME with software and hardware keyboards:

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
hellohuanlin: Updated to use the current UIScreen that it is on (I think current screen is good enough unless im misunderstanding)

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
hellohuanlin: Ah I tested it again, and found iOS 16 and 17 have different behaviors for IME. Without the fix, iPhone shows IME on top ([issue here](https://github.com/flutter/flutter/issues/132551)). The current behavior is bad, but this PR will hide IME, which is even worse. 

Even after putting the UIInputView to overlap with text doesn't 100% resolve the problem - the IME is still slightly off. Will dig into this more.

hellohuanlin:
Closing this PR in favor of https://github.com/flutter/engine/pull/44779

PR #44617, opened 2023-08-11
Add support for HardwareBuffer backed Android Platform Views under Impeller/GLES

johnmccutchan:
- Implement Impeller/GLES code paths.
- A couple of fixes for EXTERNAL_OES textures via bdero@

### DIFF
--- a/shell/platform/android/hardware_buffer_external_texture_gl.cc
+++ b/shell/platform/android/hardware_buffer_external_texture_gl.cc
@@ -6,6 +6,10 @@
 
 #include <android/hardware_buffer_jni.h>
 #include <android/sensor.h>
+#include "flutter/common/graphics/texture.h"
+#include "impeller/core/formats.h"
+#include "impeller/display_list/dl_image_impeller.h"
+#include "impeller/renderer/backend/gles/texture_gles.h"
 #include "impeller/toolkit/egl/image.h"
 #include "impeller/toolkit/gles/texture.h"
 #include "shell/platform/android/ndk_helpers.h"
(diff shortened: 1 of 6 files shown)


shell/platform/android/hardware_buffer_external_texture_gl.cc
```diff
@@ -6,6 +6,10 @@
 
 #include <android/hardware_buffer_jni.h>
 #include <android/sensor.h>
+#include "common/graphics/texture.h"
```
bdero, line 9: nit
```suggestion
#include "flutter/common/graphics/texture.h"
```

impeller/entity/shaders/texture_fill_external.frag
```diff
@@ -10,7 +10,9 @@ in float v_alpha;
 out vec4 frag_color;
 
 void main() {
-  vec4 sampled =
-      texture(SAMPLER_EXTERNAL_OES_texture_sampler, v_texture_coords);
+  vec2 tc = v_texture_coords;
```
bdero, line 13: Nit: Add todo.

Filed an issue for follow-up investigation: https://github.com/flutter/flutter/issues/132360

```suggestion
  // TODO(132360): For some reason, the vertex shader Y invert isn't working
  //               even when setting the wrapped texture intent to
  //               `TextureIntent::kRenderToTexture` for external GL textures.
  //               In the meantime, performing the sampling invert here is safe
  //               because the Android embedder only tries to composite external
  //               textures that have been written to as a render target.
  vec2 tc = v_texture_coords;
```

bdero (approved):
LGTM! Just some nits.

bdero:
You'll also need to apply the malioc diff (visible here): https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8773102538526899121/+/u/test:_malioc_diff/stdout

impeller/entity/shaders/texture_fill_external.frag
johnmccutchan: This is no longer necessary after setting the texture intent to be kUploadFromHost.

johnmccutchan: Also note that the vertex shader is no longer modified.

shell/platform/android/hardware_buffer_external_texture_gl.cc
johnmccutchan: Done.

PR #44623, opened 2023-08-11
[Skwasm] Avoid a copy when laying out text.

eyebrowsoffire:


eyebrowsoffire:
Sadly, benchmarks have shown this to be slower. Closing without merging.

PR #44636, opened 2023-08-11
[Linux] Expose channel buffers 'resize' and 'overflow' control commands

bleroux:
## Description

This PR adds two helper functions to invoke the 'resize' and 'overflow' commands exposed by the control channel.
See:

## Related Issue

Linux implementation for https://github.com/flutter/flutter/issues/132386

## Tests

Adds two tests.

### DIFF
--- a/shell/platform/linux/BUILD.gn
+++ b/shell/platform/linux/BUILD.gn
@@ -158,6 +158,7 @@ source_set("flutter_linux_sources") {
   ]
 
   deps = [
+    "//flutter/fml:fml",
     "//flutter/shell/platform/common:common_cpp_enums",
     "//flutter/shell/platform/common:common_cpp_input",
     "//flutter/shell/platform/common:common_cpp_switches",
--- a/shell/platform/linux/BUILD.gn
+++ b/shell/platform/linux/BUILD.gn
@@ -158,7 +158,7 @@ source_set("flutter_linux_sources") {
   ]
 
   deps = [
-    "//flutter/fml:fml",
+    "//flutter/fml",
     "//flutter/shell/platform/common:common_cpp_enums",
     "//flutter/shell/platform/common:common_cpp_input",
     "//flutter/shell/platform/common:common_cpp_switches",
--- a/shell/platform/linux/fl_binary_messenger.cc
+++ b/shell/platform/linux/fl_binary_messenger.cc
@@ -5,6 +5,7 @@
 #include "flutter/shell/platform/linux/public/flutter_linux/fl_binary_messenger.h"
 #include "flutter/shell/platform/linux/fl_binary_messenger_private.h"
 
+#include "flutter/fml/logging.h"
 #include "flutter/shell/platform/linux/fl_engine_private.h"
 #include "flutter/shell/platform/linux/fl_method_codec_private.h"
 #include "flutter/shell/platform/linux/public/flutter_linux/fl_engine.h"
@@ -344,6 +345,7 @@ static void resize_channel_response_cb(GObject* object,
 static void resize_channel(FlBinaryMessenger* messenger,
                            const gchar* channel,
                            int64_t new_size) {
+  FML_DCHECK(new_size >= 0);
   g_autoptr(FlStandardMethodCodec) codec = fl_standard_method_codec_new();
   g_autoptr(FlValue) args = fl_value_new_list();
   fl_value_append_take(args, fl_value_new_string(channel));
--- a/shell/platform/linux/fl_binary_messenger.cc
+++ b/shell/platform/linux/fl_binary_messenger.cc
@@ -6,9 +6,14 @@
 #include "flutter/shell/platform/linux/fl_binary_messenger_private.h"
 
 #include "flutter/shell/platform/linux/fl_engine_private.h"
+#include "flutter/shell/platform/linux/fl_method_codec_private.h"
 #include "flutter/shell/platform/linux/public/flutter_linux/fl_engine.h"
+#include "flutter/shell/platform/linux/public/flutter_linux/fl_method_codec.h"
+#include "flutter/shell/platform/linux/public/flutter_linux/fl_standard_method_codec.h"
 
 #include <gmodule.h>
+
+static constexpr char kControlChannelName[] = "dev.flutter/channel-buffers";
 
 G_DEFINE_QUARK(fl_binary_messenger_codec_error_quark,
                fl_binary_messenger_codec_error)
@@ -311,6 +316,32 @@ static GBytes* send_on_channel_finish(FlBinaryMessenger* messenger,
   return fl_engine_send_platform_message_finish(engine, r, error);
 }
 
+static void resize_channel(FlBinaryMessenger* messenger,
+                           const gchar* channel,
+                           int new_size) {
+  g_autoptr(FlStandardMethodCodec) codec = fl_standard_method_codec_new();
+  g_autoptr(FlValue) args = fl_value_new_list();
+  fl_value_append_take(args, fl_value_new_string(channel));
+  fl_value_append_take(args, fl_value_new_int(new_size));
+  g_autoptr(GBytes) message = fl_method_codec_encode_method_call(
+      FL_METHOD_CODEC(codec), "resize", args, nullptr);
+  fl_binary_messenger_send_on_channel(messenger, kControlChannelName, message,
+                                      nullptr, nullptr, nullptr);
+}
+
+static void allow_channel_overflow(FlBinaryMessenger* messenger,
+                                   const gchar* channel,
+                                   bool allowed) {
+  g_autoptr(FlStandardMethodCodec) codec = fl_standard_method_codec_new();
+  g_autoptr(FlValue) args = fl_value_new_list();
+  fl_value_append_take(args, fl_value_new_string(channel));
+  fl_value_append_take(args, fl_value_new_bool(allowed));
+  g_autoptr(GBytes) message = fl_method_codec_encode_method_call(
+      FL_METHOD_CODEC(codec), "overflow", args, nullptr);
+  fl_binary_messenger_send_on_channel(messenger, kControlChannelName, message,
+                                      nullptr, nullptr, nullptr);
+}
+
 static void fl_binary_messenger_impl_class_init(
     FlBinaryMessengerImplClass* klass) {
   G_OBJECT_CLASS(klass)->dispose = fl_binary_messenger_impl_dispose;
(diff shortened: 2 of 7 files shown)


shell/platform/linux/public/flutter_linux/fl_binary_messenger.h
```diff
@@ -187,6 +197,32 @@ GBytes* fl_binary_messenger_send_on_channel_finish(FlBinaryMessenger* messenger,
                                                    GAsyncResult* result,
                                                    GError** error);
 
+/**
+ * fl_binary_messenger_resize_channel:
+ * @binary_messenger: an #FlBinaryMessenger.
```
robert-ancell, line 202: Should be '@messenger'

shell/platform/linux/public/flutter_linux/fl_binary_messenger.h
```diff
@@ -187,6 +197,32 @@ GBytes* fl_binary_messenger_send_on_channel_finish(FlBinaryMessenger* messenger,
...
+ *
+ * Sends a message to the control channel asking to resize a channel buffer.
+ */
+void fl_binary_messenger_resize_channel(FlBinaryMessenger* messenger,
+                                        const gchar* channel,
+                                        int new_size);
+
+/**
+ * fl_binary_messenger_allow_channel_overflow:
+ * @binary_messenger: an #FlBinaryMessenger.
```
robert-ancell, line 214: Should be '@messenger'

shell/platform/linux/public/flutter_linux/fl_binary_messenger.h
```diff
@@ -187,6 +197,32 @@ GBytes* fl_binary_messenger_send_on_channel_finish(FlBinaryMessenger* messenger,
...
+ * fl_binary_messenger_resize_channel:
+ * @binary_messenger: an #FlBinaryMessenger.
+ * @channel: channel to be resize.
+ * @new_size: the new size for the channel buffer.
+ *
+ * Sends a message to the control channel asking to resize a channel buffer.
+ */
+void fl_binary_messenger_resize_channel(FlBinaryMessenger* messenger,
+                                        const gchar* channel,
+                                        int new_size);
```
robert-ancell, line 210: Use int64_t as that's the size that's data type that's sent in the channel. I was also considering size_t, but it's probably better to match what is sent.

shell/platform/linux/public/flutter_linux/fl_binary_messenger.h
```diff
@@ -187,6 +197,32 @@ GBytes* fl_binary_messenger_send_on_channel_finish(FlBinaryMessenger* messenger,
...
+ * fl_binary_messenger_allow_channel_overflow:
+ * @binary_messenger: an #FlBinaryMessenger.
+ * @channel: channel to be allowed to overflow silently.
+ * @allowed: when true the channel is expected to overflow and warning messages
+ * will not be shown.
+ *
+ * Sends a message to the control channel asking to allow or disallow a channel
+ * to overflow silently.
+ */
+void fl_binary_messenger_allow_channel_overflow(FlBinaryMessenger* messenger,
```
robert-ancell, line 222: This seems like it should be called `fl_binary_messenger_set_allow_channel_overflow`

shell/platform/linux/fl_binary_messenger.cc
```diff
@@ -311,6 +316,32 @@ static GBytes* send_on_channel_finish(FlBinaryMessenger* messenger,
...
+static void resize_channel(FlBinaryMessenger* messenger,
+                           const gchar* channel,
+                           int new_size) {
+  g_autoptr(FlStandardMethodCodec) codec = fl_standard_method_codec_new();
+  g_autoptr(FlValue) args = fl_value_new_list();
+  fl_value_append_take(args, fl_value_new_string(channel));
+  fl_value_append_take(args, fl_value_new_int(new_size));
+  g_autoptr(GBytes) message = fl_method_codec_encode_method_call(
+      FL_METHOD_CODEC(codec), "resize", args, nullptr);
+  fl_binary_messenger_send_on_channel(messenger, kControlChannelName, message,
```
robert-ancell, line 328: Might be worth handling the result here in case there's an error which could be shown with g_warning. Otherwise you'll never know.

shell/platform/linux/fl_binary_messenger.cc
```diff
@@ -311,6 +316,32 @@ static GBytes* send_on_channel_finish(FlBinaryMessenger* messenger,
...
 
+static void resize_channel(FlBinaryMessenger* messenger,
+                           const gchar* channel,
+                           int new_size) {
+  g_autoptr(FlStandardMethodCodec) codec = fl_standard_method_codec_new();
+  g_autoptr(FlValue) args = fl_value_new_list();
+  fl_value_append_take(args, fl_value_new_string(channel));
+  fl_value_append_take(args, fl_value_new_int(new_size));
+  g_autoptr(GBytes) message = fl_method_codec_encode_method_call(
+      FL_METHOD_CODEC(codec), "resize", args, nullptr);
```
robert-ancell, line 327: To match existing code this should probably use `kResizeMethod`.

shell/platform/linux/fl_binary_messenger.cc
```diff
@@ -311,6 +316,32 @@ static GBytes* send_on_channel_finish(FlBinaryMessenger* messenger,
...
+
+static void allow_channel_overflow(FlBinaryMessenger* messenger,
+                                   const gchar* channel,
+                                   bool allowed) {
+  g_autoptr(FlStandardMethodCodec) codec = fl_standard_method_codec_new();
+  g_autoptr(FlValue) args = fl_value_new_list();
+  fl_value_append_take(args, fl_value_new_string(channel));
+  fl_value_append_take(args, fl_value_new_bool(allowed));
+  g_autoptr(GBytes) message = fl_method_codec_encode_method_call(
+      FL_METHOD_CODEC(codec), "overflow", args, nullptr);
```
robert-ancell, line 340: To match existing code this should probably use `kOverflowMethod`.

robert-ancell (changes requested):
Looks generally good, thanks!

This doesn't seem to be implemented for the other platforms that I can see, so I'm not 100% sure if these methods should be part of FlBinaryMessenger or another class. Also unsure if dev.fluter/channel-buffers should be a class of its own, or it's fine to hide it inside FlBinaryMessenger. @cbracken @gspencergoog @stuartmorgan  you might have thoughts on this.

shell/platform/linux/fl_binary_messenger.cc
bleroux: I'm not sure how to do this. I was not able to find a place where a similar handling is done, can you point out an example?

shell/platform/linux/public/flutter_linux/fl_binary_messenger.h
```diff
@@ -91,6 +93,14 @@ struct _FlBinaryMessengerInterface {
   GBytes* (*send_on_channel_finish)(FlBinaryMessenger* messenger,
                                     GAsyncResult* result,
                                     GError** error);
+
+  void (*resize_channel)(FlBinaryMessenger* messenger,
+                         const gchar* channel,
+                         int64_t new_size);
```
cbracken, line 99: Is int64_t the right type here? Should this be `size_t` or at least `uint64_t`, or is there a need to support a negative size?

I notice that the [Dart side of this](https://github.com/flutter/engine/blob/eba35482ccaaba7526a4c2a82383c8edc0e907a6/lib/ui/channel_buffers.dart#L78) doesn't seem to guard against negative sizes either, though the documentation seems to imply that capacity should always be >= 0.

cbracken (commented):
Just one comment regarding guarding against negative values for the new size in a resize.

shell/platform/linux/public/flutter_linux/fl_binary_messenger.h
robert-ancell: If we use size_t here the method will need to validate if it's greater than INT64_MAX. If we allow int64_t, then we let the server Dart side do the validation and return an error. I think in practice the user is unlikely to send a negative number and relying on the second method is probably less likely to cause an issue.

shell/platform/linux/fl_binary_messenger.cc
robert-ancell: See https://github.com/flutter/engine/blob/533670269c99a8f552edb1e5bf5a17b706e54e5b/shell/platform/linux/fl_text_input_plugin.cc#L164 and https://github.com/flutter/engine/blob/533670269c99a8f552edb1e5bf5a17b706e54e5b/shell/platform/linux/fl_text_input_plugin.cc#L116 for a case where a method call is only completed to check for the error. There's no other cases using `fl_binary_messenger_send_on_channel` directly so in this case you'll need to use `fl_binary_messenger_send_on_channel_finish` and `fl_method_codec_decode_response`.

shell/platform/linux/public/flutter_linux/fl_binary_messenger.h
cbracken: Perhaps worth adding a debug-mode assertion in that case via FML_DCHECK

stuartmorgan:
> This doesn't seem to be implemented for the other platforms that I can see, so I'm not 100% sure if these methods should be part of FlBinaryMessenger or another class. Also unsure if dev.fluter/channel-buffers should be a class of its own, or it's fine to hide it inside FlBinaryMessenger. @cbracken @gspencergoog @stuartmorgan you might have thoughts on this.

Putting it on the binary messenger seems reasonable to me. If they were likely to be frequently used we'd probably want to expose them on the channel classes themselves, but there's no reason to think that would be the case.

robert-ancell (approved):
LGTM

cbracken (approved):
lgtm; consider adding a debug-time FML_DCHECK assertion on size>=0 to help catch errors.

bleroux:
> lgtm; consider adding a debug-time FML_DCHECK assertion on size>=0 to help catch errors.

@cbracken I added it, can you check if I did it right (I had to update a build configuration file)?

shell/platform/linux/BUILD.gn
```diff
@@ -158,6 +158,7 @@ source_set("flutter_linux_sources") {
   ]
 
   deps = [
+    "//flutter/fml:fml",
```
cbracken, line 161:
```suggestion
    "//flutter/fml",
```

cbracken (approved):
Looks great - and yep, that's exactly how you do it. 

As a minor improvement, when the target is the same as the BUILD.gn directory, you can just use the directory -- added a suggestion that can be committed.

PR #44644, opened 2023-08-11
Update android_embedding_bundle CIPD instructions

zanderso:
This PR updates the gradle build file for the android_embedding_bundle cipd package to fetch license information about the dependencies listed under `tools/androidx/files.json`. The README file for the android_embedding_bundle now explains how to fetch the license information and make sure it's correct in the cipd package.

This PR also pulls a new version of android_embedding_dependencies that includes a `LICENSE` file.

It turns out that Flutter does not re-ship anything from the embedding bundle to Flutter end-users. These dependencies are only used to create a classpath against which we build the Android embedder and in-tree tests. I've updated docs/comments to clarify that since I was initially confused about how these were used.

I'm not adding a test that compares the license info fetched from maven against a checked-in golden, yet. Still thinking about the most effective way to do that in combination with the steps in the README.


tools/cipd/android_embedding_bundle/README.md
```diff
@@ -1,12 +1,34 @@
...
+   `src/third_party/gradle/bin/gradle`.
 
 ## Steps
 
-1. Update tools/androidx/files.json. (This file includes the Maven dependencies used to build Flutter apps).
-2. `cd` into this directory.
-3. Run `./generate.sh`.
+1. Update `tools/androidx/files.json`. (This file includes the Maven
+   dependencies used to build Flutter apps).
+1. `cd` into this directory.
```
reidbaker, line 23: Can you spell out the directory? It is nice when following instructions that have not been run in a while to be super specific. 

```suggestion
1. `cd` into `tools/androidx` directory.
```

reidbaker (approved):

Hixie:
This implementation matches my understanding of the advice we were given. I'll let @reidbaker give the actual approval since he has a better understanding of the subtleties of the relevant code.

tools/cipd/android_embedding_bundle/README.md
zanderso: Done.

PR #44690, opened 2023-08-14
Revert "Make run_tests.py assert that the ios test dylib is at least as new as libFlutter.dylib"

dnfield:
Reverts flutter/engine#41720

Fixes https://github.com/flutter/flutter/issues/131899


dnfield:
Someone (maybe @gaaclarke remembers who?) was saying they're going to fix the build rules so the test lib is just built as its own target and this won't be an issue anymore.

zanderso (approved):

PR #44693, opened 2023-08-14
[Web] Fix insertions/deletions at inverted selection for TextEditingDeltas

Renzo-Olivares:
Fixes issue where the delta range would be calculated wrong when the selection is inverted.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/lib/web_ui/test/engine/text_editing_test.dart
+++ b/lib/web_ui/test/engine/text_editing_test.dart
@@ -1829,7 +1829,8 @@ Future<void> testMain() async {
       spy.messages.clear();
 
       hideKeyboard();
-    });
+      // TODO(Renzo-Olivares): https://github.com/flutter/flutter/issues/134271
+    }, skip: isSafari);
 
     test('Supports new line at inverted selection', () async {
       final MethodCall setClient = MethodCall(
@@ -1903,7 +1904,8 @@ Future<void> testMain() async {
       spy.messages.clear();
 
       hideKeyboard();
-    });
+      // TODO(Renzo-Olivares): https://github.com/flutter/flutter/issues/134271
+    }, skip: isSafari);
 
     test('multiTextField Autofill sync updates back to Flutter', () async {
       // Create a configuration with an AutofillGroup of four text fields.
--- a/lib/web_ui/test/engine/text_editing_test.dart
+++ b/lib/web_ui/test/engine/text_editing_test.dart
@@ -1757,7 +1757,7 @@ Future<void> testMain() async {
       hideKeyboard();
     });
 
-    test('Supports deletion at inverted selection', () {
+    test('Supports deletion at inverted selection', () async {
       final MethodCall setClient = MethodCall(
           'TextInput.setClient', <dynamic>[123, createFlutterConfig('text', enableDeltaModel: true)]);
       sendFrameworkMessage(codec.encodeMethodCall(setClient));
@@ -1773,10 +1773,10 @@ Future<void> testMain() async {
       const MethodCall show = MethodCall('TextInput.show');
       sendFrameworkMessage(codec.encodeMethodCall(show));
 
-      // The "setSizeAndTransform" message has to be here before we call
-      // checkInputEditingState, since on some platforms (e.g. Desktop Safari)
-      // we don't put the input element into the DOM until we get its correct
-      // dimensions from the framework.
+      // // The "setSizeAndTransform" message has to be here before we call
+      // // checkInputEditingState, since on some platforms (e.g. Desktop Safari)
+      // // we don't put the input element into the DOM until we get its correct
+      // // dimensions from the framework.
       final MethodCall setSizeAndTransform =
           configureSetSizeAndTransformMethodCall(150, 50,
               Matrix4.translationValues(10.0, 20.0, 30.0).storage.toList());
@@ -1831,7 +1831,7 @@ Future<void> testMain() async {
       hideKeyboard();
     });
 
-    test('Supports new line at inverted selection', () {
+    test('Supports new line at inverted selection', () async {
       final MethodCall setClient = MethodCall(
           'TextInput.setClient', <dynamic>[123, createFlutterConfig('text', enableDeltaModel: true)]);
       sendFrameworkMessage(codec.encodeMethodCall(setClient));
--- a/lib/web_ui/test/engine/text_editing_test.dart
+++ b/lib/web_ui/test/engine/text_editing_test.dart
@@ -3,7 +3,6 @@
 // found in the LICENSE file.
 
 import 'dart:async';
-import 'dart:js_interop';
 import 'dart:js_util' as js_util;
 import 'dart:typed_data';
 
--- a/lib/web_ui/test/engine/text_editing_test.dart
+++ b/lib/web_ui/test/engine/text_editing_test.dart
@@ -1773,6 +1773,17 @@ Future<void> testMain() async {
       const MethodCall show = MethodCall('TextInput.show');
       sendFrameworkMessage(codec.encodeMethodCall(show));
 
+      // The "setSizeAndTransform" message has to be here before we call
+      // checkInputEditingState, since on some platforms (e.g. Desktop Safari)
+      // we don't put the input element into the DOM until we get its correct
+      // dimensions from the framework.
+      final MethodCall setSizeAndTransform =
+          configureSetSizeAndTransformMethodCall(150, 50,
+              Matrix4.translationValues(10.0, 20.0, 30.0).storage.toList());
+      sendFrameworkMessage(codec.encodeMethodCall(setSizeAndTransform));
+
+      await waitForDesktopSafariFocus();
+
       final DomHTMLInputElement input = textEditing!.strategy.domElement! as
           DomHTMLInputElement;
 
@@ -1835,6 +1846,17 @@ Future<void> testMain() async {
 
       const MethodCall show = MethodCall('TextInput.show');
       sendFrameworkMessage(codec.encodeMethodCall(show));
+
+      // The "setSizeAndTransform" message has to be here before we call
+      // checkInputEditingState, since on some platforms (e.g. Desktop Safari)
+      // we don't put the input element into the DOM until we get its correct
+      // dimensions from the framework.
+      final MethodCall setSizeAndTransform =
+          configureSetSizeAndTransformMethodCall(150, 50,
+              Matrix4.translationValues(10.0, 20.0, 30.0).storage.toList());
+      sendFrameworkMessage(codec.encodeMethodCall(setSizeAndTransform));
+
+      await waitForDesktopSafariFocus();
 
       final DomHTMLInputElement input = textEditing!.strategy.domElement! as
           DomHTMLInputElement;
--- a/lib/web_ui/test/engine/text_editing_test.dart
+++ b/lib/web_ui/test/engine/text_editing_test.dart
@@ -1773,10 +1773,10 @@ Future<void> testMain() async {
       const MethodCall show = MethodCall('TextInput.show');
       sendFrameworkMessage(codec.encodeMethodCall(show));
 
-      // // The "setSizeAndTransform" message has to be here before we call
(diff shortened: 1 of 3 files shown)


lib/web_ui/test/engine/text_editing_test.dart
```diff
@@ -1757,6 +1769,142 @@ Future<void> testMain() async {
...
+      sendFrameworkMessage(codec.encodeMethodCall(show));
+
+      final DomHTMLInputElement input = textEditing!.strategy.domElement! as
+          DomHTMLInputElement;
+
+      final Object jsInputEvent = js_util.getProperty<Object>(domWindow, 'InputEvent');
+      final List<dynamic> eventArgs = <dynamic>[
+        'beforeinput',
+        js_util.jsify(<String, dynamic>{
+          'inputType': 'insertLineBreak',
```
Renzo-Olivares, line 1863: @mdebbar This test currently fails because there is no `DomInputEvent` defined for `InputEvent` which contains `inputType`. In `handleBeforeInput` we check the `inputType` in the logic. Is there a way to test this?

Renzo-Olivares: Thanks for the fix! I was trying to avoid adding to the DOM API but if that's the only way to test this I think its okay. `InputEvent` is a basic building block like `KeyboardEvent` and other related events so it would be good to add.

knopp: Yeah. It seems to me that extending DOM bindings where needed is better than ad-hoc wrapper just for the test.

lib/web_ui/test/engine/text_editing_test.dart
```diff
@@ -3,6 +3,7 @@
 // found in the LICENSE file.
 
 import 'dart:async';
+import 'dart:js_interop';
```
knopp, line 6: Seems causing the CI check to fail.

knopp (approved):

Renzo-Olivares:
cc @htoor3 @mdebbar for feedback/thoughts.

htoor3:
Thank you for this change @Renzo-Olivares! 

There appear to be a few linting errors
```
warning - lib/src/engine/text_editing/text_editing.dart:564:96 - The '!' will have no effect because the receiver can't be null. Try removing the '!' operator. - unnecessary_non_null_assertion

warning - lib/src/engine/text_editing/text_editing.dart:567:91 - The '!' will have no effect because the receiver can't be null. Try removing the '!' operator. - unnecessary_non_null_assertion

warning - lib/src/engine/text_editing/text_editing.dart:567:125 - The '!' will have no effect because the receiver can't be null. Try removing the '!' operator. - unnecessary_non_null_assertion
```

Also it doesn't look like we're sending `selectionAffinity` information back to the framework ([ref](https://github.com/flutter/engine/blob/main/lib/web_ui/lib/src/engine/text_editing/text_editing.dart#L811)).  Is this something that might cause issues? Framework seems to just default to `downstream` ([ref](https://github.com/flutter/flutter/blob/master/packages/flutter/lib/src/services/text_input.dart#L782))

knopp:
> Thank you for this change @Renzo-Olivares!
> 
> There appear to be a few linting errors
> 
> ```
> warning - lib/src/engine/text_editing/text_editing.dart:564:96 - The '!' will have no effect because the receiver can't be null. Try removing the '!' operator. - unnecessary_non_null_assertion
> 
> warning - lib/src/engine/text_editing/text_editing.dart:567:91 - The '!' will have no effect because the receiver can't be null. Try removing the '!' operator. - unnecessary_non_null_assertion
> 
> warning - lib/src/engine/text_editing/text_editing.dart:567:125 - The '!' will have no effect because the receiver can't be null. Try removing the '!' operator. - unnecessary_non_null_assertion
> ```
> 
> Also it doesn't look like we're sending `selectionAffinity` information back to the framework ([ref](https://github.com/flutter/engine/blob/main/lib/web_ui/lib/src/engine/text_editing/text_editing.dart#L811)). Is this something that might cause issues in situations where `baseOffset` > `extentOffset`? Framework seems to just default to `downstream` ([ref](https://github.com/flutter/flutter/blob/master/packages/flutter/lib/src/services/text_input.dart#L782))

TextAffinity should be reported, but it's somewhat orthogonal to inverted selection (`baseOffset` > `extentOffset`). I have a PR to properly report inverted selection here: https://github.com/flutter/engine/pull/44806, it just waits for this one to land.

Renzo-Olivares:
@htoor3 Sorry about that must've missed these. They should be fixed now. `selectionAffinity` is important but piping it through has no affect on the fix in this PR. 

In general the framework does use `selectionAffinity` to calculate some text boundaries [(ref1)](https://github.com/flutter/flutter/blob/e1e47221e86272429674bec4f1bd36acc4fc7b77/packages/flutter/lib/src/rendering/editable.dart#L2097) and [(ref2)](https://github.com/flutter/flutter/blob/e1e47221e86272429674bec4f1bd36acc4fc7b77/packages/flutter/lib/src/widgets/editable_text.dart#L4297) so it would be good to pipe it through eventually. From looking into the other platforms `TextInputPlugin` it seems only iOS/macOS have non-default values for `selectionAffinity`, for reference: 
[(Android)](https://github.com/flutter/engine/blob/29085c5564e9e14c104f4ea39b1c44dc2b1c1d89/shell/platform/android/io/flutter/embedding/engine/systemchannels/TextInputChannel.java#L207-L236)
[(macOS)](https://github.com/flutter/engine/blob/29085c5564e9e14c104f4ea39b1c44dc2b1c1d89/shell/platform/darwin/macos/framework/Source/FlutterTextInputPlugin.mm#L313)
[(iOS)](https://github.com/flutter/engine/blob/29085c5564e9e14c104f4ea39b1c44dc2b1c1d89/shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm#L789)
[(Windows)](https://github.com/flutter/engine/blob/29085c5564e9e14c104f4ea39b1c44dc2b1c1d89/shell/platform/windows/text_input_plugin.cc#L398-L399)
[(Linux)](https://github.com/flutter/engine/blob/29085c5564e9e14c104f4ea39b1c44dc2b1c1d89/shell/platform/linux/fl_text_input_plugin.cc#L156-L160)

htoor3:
> @htoor3 Sorry about that must've missed these. They should be fixed now. `selectionAffinity` is important but piping it through has no affect on the fix in this PR.
> 
> In general the framework does use `selectionAffinity` to calculate some text boundaries [(ref1)](https://github.com/flutter/flutter/blob/e1e47221e86272429674bec4f1bd36acc4fc7b77/packages/flutter/lib/src/rendering/editable.dart#L2097) and [(ref2)](https://github.com/flutter/flutter/blob/e1e47221e86272429674bec4f1bd36acc4fc7b77/packages/flutter/lib/src/widgets/editable_text.dart#L4297) so it would be good to pipe it through eventually. From looking into the other platforms `TextInputPlugin` it seems only iOS/macOS have non-default values for `selectionAffinity`, for reference: [(Android)](https://github.com/flutter/engine/blob/29085c5564e9e14c104f4ea39b1c44dc2b1c1d89/shell/platform/android/io/flutter/embedding/engine/systemchannels/TextInputChannel.java#L207-L236) [(macOS)](https://github.com/flutter/engine/blob/29085c5564e9e14c104f4ea39b1c44dc2b1c1d89/shell/platform/darwin/macos/framework/Source/FlutterTextInputPlugin.mm#L313) [(iOS)](https://github.com/flutter/engine/blob/29085c5564e9e14c104f4ea39b1c44dc2b1c1d89/shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm#L789) [(Windows)](https://github.com/flutter/engine/blob/29085c5564e9e14c104f4ea39b1c44dc2b1c1d89/shell/platform/windows/text_input_plugin.cc#L398-L399) [(Linux)](https://github.com/flutter/engine/blob/29085c5564e9e14c104f4ea39b1c44dc2b1c1d89/shell/platform/linux/fl_text_input_plugin.cc#L156-L160)

Yep it didn't seem like it was a pressing TODO, but I was curious if it was related to this issue.  I think there's a fallback in `editable_text.dart` (`_checkNeedsAdjustAffinity()`) that just overwrites the engine value with whatever the framework already has. 

This change lgtm! There are just a few more breaking tests in what looks like Desktop Safari only - `HybridTextEditing Supports new line at inverted selection` and `HybridTextEditing Supports deletion at inverted selection` 

Desktop Safari can be a little weird to test.  Take a look at this comment [here](https://github.com/flutter/engine/blob/main/lib/web_ui/test/engine/text_editing_test.dart#L526-L535) - it's possible adding these to ensure the placement of the input element could resolve those tests.

htoor3 (approved):

knopp:
I can't reproduce the safari error locally (though I do get another failing - "text editing styles invisible element"). Could this be a problem with different Safari version?

knopp:
Right. It seems like the test that is failing for me (`element.style.background` is not transparent, but none) would pass in Safari 15, but fails in Safar 16.

Renzo-Olivares:
I don't think this PR is a result of those failures. I tried running tests locally on `main` without any changes and got two failures: `HybridTextEditing focus and connection with blur` and `text editing styles invisible element`. It looks like the linux web engine is failing so i'll try to build the engine on my linux machine and see if I can work this out.

knopp:
> I don't think this PR is a result of those failures. I tried running tests locally on `main` without any changes and got two failures: `HybridTextEditing focus and connection with blur` and `text editing styles invisible element`. It looks like the linux web engine is failing so i'll try to build the engine on my linux machine and see if I can work this out.

I think you're getting the failure because after setting `style.background="transparent"` on element and then getting the `style.background" returns `none` on Safari 15, but `transparent` on Safari 14. I think the CI might be running the tests on Safari 14, hence the discrepancy.

knopp:
I played with this a little bit. It turns out that Safari 15 doesn't support setting `inputType` on DOM input event.
```js
var ev = new InputEvent('beforeinput', {'inputType' : 'deleteContentBackward'})
ev.inputType // always prints ""
```

InputEventInit inputType has been only supported on Safari since this commit: https://github.com/WebKit/WebKit/pull/3990. which  didn't make it into Safari 15.

I don't know if it is reasonable to expect the CI to be updated to Safari 16 any time soon, but if not then these tests should maybe be disabled for Safari?

Renzo-Olivares:
@knopp Thank you for looking into that! Given that information I think its okay to disable these tests on Safari for the time being and a TODO explaining the issue. Our use of `inputType` here is not introduced by this PR this is just the first time we have tested this code path. Do you have any opinions on doing this @htoor3?

htoor3:
@Renzo-Olivares I think that makes sense. Let me doublecheck with the team to determine what version of Safari is being used to run our CI tests and also what the version rolling process looks like. 

Could you point me to the codepath where we're testing `inputType`?

knopp:
> @Renzo-Olivares I think that makes sense. Let me doublecheck with the team to determine what version of Safari is being used to run our CI tests and also what the version rolling process looks like.
> 
> Could you point me to the codepath where we're testing `inputType`?

[Here](https://github.com/flutter/engine/pull/44693/files#diff-35e49256f21edf28839244a4083d687b57a82c8bc45b3e8c145f7dafd2a3db64R1790) we're creating a `DomInputEvent` with `'inputType': 'deleteContentBackward'`, which only works in Safari 16. In Safari 15  querying the `inputType` after creating the event always returns empty string.

htoor3:
Thank you @knopp! Regarding the CI Safari version, it seems like it's tied to the CI server MacOS version and probably needs manual intervention to roll.  

I think in that case, it's fair to disable them for Safari so we can get this fix in.  I would add a TODO that links to a newly created issue outlining the Safari issue, mentioning to re-enable these after the Safari version rolls to 16 in CI.

knopp:
Looks like the tests pass, can this be merged?

Renzo-Olivares:
Yes I'll merge it. I created an issue to re-enable the tests https://github.com/flutter/flutter/issues/134271.

PR #44696, opened 2023-08-14
[Impeller] Flutter GPU: Add HostBuffer.

bdero:
Resolves https://github.com/flutter/flutter/issues/132516.

Add `impeller::HostBuffer` wrapper to Flutter GPU.
* Allows for lazy batch uploads of sparse host data to the GPU.
* Handles platform alignment requirements.
* API returns buffer view handles that will be fed to commands.

### DIFF
--- a/lib/gpu/lib/src/buffer.dart
+++ b/lib/gpu/lib/src/buffer.dart
@@ -32,11 +32,10 @@ class BufferView {
 /// This is useful for efficiently chunking sparse data uploads, especially
 /// ephemeral uniform data that needs to change from frame to frame.
 ///
-/// Note: Different platforms have different data alignment requirements for
-///       accessing device buffer data. The [HostBuffer] takes these
-///       requirements into account and automatically inserts padding between
-///       emplaced data if necessary.
-class HostBuffer extends NativeFieldWrapperClass1 {
+/// Different platforms have different data alignment requirements for accessing
+/// device buffer data. The [HostBuffer] takes these requirements into account
+/// and automatically inserts padding between emplaced data if necessary.
+base class HostBuffer extends NativeFieldWrapperClass1 {
   /// Creates a new HostBuffer.
   HostBuffer() {
     _initialize();
--- a/lib/gpu/lib/src/buffer.dart
+++ b/lib/gpu/lib/src/buffer.dart
@@ -8,30 +8,31 @@ import 'dart:typed_data';
 
 import 'context.dart';
 
-/// A buffer data range.
+/// A reference to a byte range within a GPU-resident [Buffer].
 class BufferView {
   /// The buffer of this view.
   final Buffer buffer;
+
   /// The start of the view, in bytes starting from the beginning of the
   /// [buffer].
   final int offsetInBytes;
+
   /// The end of the view
   final int lengthInBytes;
 
   /// Create a new view into a buffer on the GPU.
-  const BufferView(
-      {required this.buffer,
-      required this.offsetInBytes,
-      required this.lengthInBytes});
+  const BufferView(this.buffer,
+      {required this.offsetInBytes, required this.lengthInBytes});
 }
 
 /// A buffer that can be referenced by commands on the GPU.
 mixin Buffer {}
 
-/// [HostBuffer] is a [Buffer] which is allocated on the host and lazily
-/// uploaded to the GPU. A [HostBuffer] can be safely mutated or extended at
-/// any time on the host, and will be automatically re-uploaded to the GPU
-/// the next time a GPU operation needs to access it.
+/// [HostBuffer] is a [Buffer] which is allocated on the host (native CPU
+/// resident memory) and lazily uploaded to the GPU. A [HostBuffer] can be
+/// safely mutated or extended at any time on the host, and will be
+/// automatically re-uploaded to the GPU the next time a GPU operation needs to
+/// access it.
 ///
 /// This is useful for efficiently chunking sparse data uploads, especially
 /// ephemeral uniform data that needs to change from frame to frame.
@@ -54,17 +55,16 @@ class HostBuffer extends NativeFieldWrapperClass1 with Buffer {
   /// Append byte data to the end of the [HostBuffer] and produce a [BufferView]
   /// that references the new data in the buffer.
   ///
-  /// This method automatically inserts padding in-between emplaced data if
-  /// necessary to abide by platform-specific uniform alignment requirements.
+  /// This method automatically inserts padding in-between emplace calls in the
+  /// buffer if necessary to abide by platform-specific uniform alignment
+  /// requirements.
   ///
   /// The updated buffer will be uploaded to the GPU if the returned
   /// [BufferView] is used by a rendering command.
-  BufferView emplaceBytes({required ByteData bytes}) {
+  BufferView emplace(ByteData bytes) {
     int resultOffset = _emplaceBytes(bytes);
-    return BufferView(
-        buffer: this,
-        offsetInBytes: resultOffset,
-        lengthInBytes: bytes.lengthInBytes);
+    return BufferView(this,
+        offsetInBytes: resultOffset, lengthInBytes: bytes.lengthInBytes);
   }
 
   @Native<Uint64 Function(Pointer<Void>, Handle)>(
--- a/lib/gpu/lib/src/buffer.dart
+++ b/lib/gpu/lib/src/buffer.dart
@@ -1,0 +1,73 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'dart:ffi';
+import 'dart:nativewrappers';
+import 'dart:typed_data';
+
+import 'context.dart';
+
+/// A buffer data range.
+class BufferView {
+  /// The buffer of this view.
+  final Buffer buffer;
+  /// The start of the view, in bytes starting from the beginning of the
+  /// [buffer].
+  final int offsetInBytes;
+  /// The end of the view
+  final int lengthInBytes;
+
+  /// Create a new view into a buffer on the GPU.
+  const BufferView(
+      {required this.buffer,
+      required this.offsetInBytes,
+      required this.lengthInBytes});
+}
+
+/// A buffer that can be referenced by commands on the GPU.
+mixin Buffer {}
+
+/// [HostBuffer] is a [Buffer] which is allocated on the host and lazily
+/// uploaded to the GPU. A [HostBuffer] can be safely mutated or extended at
+/// any time on the host, and will be automatically re-uploaded to the GPU
+/// the next time a GPU operation needs to access it.
+///
+/// This is useful for efficiently chunking sparse data uploads, especially
+/// ephemeral uniform data that needs to change from frame to frame.
+///
+/// Note: Different platforms have different data alignment requirements for
+///       accessing device buffer data. The [HostBuffer] takes these
+///       requirements into account and automatically inserts padding between
+///       emplaced data if necessary.
+class HostBuffer extends NativeFieldWrapperClass1 with Buffer {
+  /// Creates a new HostBuffer.
+  HostBuffer() {
+    _initialize();
+  }
+
+  /// Wrap with native counterpart.
+  @Native<Void Function(Handle)>(
+      symbol: 'InternalFlutterGpu_HostBuffer_Initialize')
+  external void _initialize();
+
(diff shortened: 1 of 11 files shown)


lib/gpu/lib/src/buffer.dart
```diff
@@ -0,0 +1,73 @@
...
+
+  /// Create a new view into a buffer on the GPU.
+  const BufferView(
+      {required this.buffer,
+      required this.offsetInBytes,
+      required this.lengthInBytes});
+}
+
+/// A buffer that can be referenced by commands on the GPU.
+mixin Buffer {}
```
jonahwilliams, line 29: Generally speaking, a completely empty interface is a pretty big code smell.

What other kinds of buffers are we going to have?  Are device buffers going to be usable in the same context as host buffers?

lib/gpu/lib/src/buffer.dart
```diff
@@ -0,0 +1,73 @@
...
+
+  /// Append byte data to the end of the [HostBuffer] and produce a [BufferView]
+  /// that references the new data in the buffer.
+  ///
+  /// This method automatically inserts padding in-between emplaced data if
+  /// necessary to abide by platform-specific uniform alignment requirements.
+  ///
+  /// The updated buffer will be uploaded to the GPU if the returned
+  /// [BufferView] is used by a rendering command.
+  BufferView emplaceBytes({required ByteData bytes}) {
```
jonahwilliams, line 62: Are there going to be any other arguments here? Can we make this a positional parameter instead so there is less ceremony?

lib/gpu/lib/src/buffer.dart
```diff
@@ -0,0 +1,73 @@
...
+
+  /// Wrap with native counterpart.
+  @Native<Void Function(Handle)>(
+      symbol: 'InternalFlutterGpu_HostBuffer_Initialize')
+  external void _initialize();
+
+  /// Append byte data to the end of the [HostBuffer] and produce a [BufferView]
+  /// that references the new data in the buffer.
+  ///
+  /// This method automatically inserts padding in-between emplaced data if
```
jonahwilliams, line 57: in-between - can you clarify that this is in between subsequent emplace calls?

lib/gpu/lib/src/buffer.dart
jonahwilliams, line 62: Nit: can we call this "emplace"?

lib/gpu/lib/src/context.dart
```diff
@@ -22,6 +24,12 @@ class GpuContext extends NativeFieldWrapperClass1 {
   @Native<Handle Function(Handle)>(
       symbol: 'InternalFlutterGpu_Context_InitializeDefault')
   external String? _initializeDefault();
+
+  /// Create a new [HostBuffer] that can be used for staging and transferring
+  /// data from the host to the GPU.
+  HostBuffer createHostBuffer() {
```
jonahwilliams, line 30: Does this need to be on the GPUContext? All it does is call the the HostBuffer constructor, which developers could just do themselves.

lib/gpu/lib/src/buffer.dart
```diff
@@ -0,0 +1,73 @@
...
+  const BufferView(
+      {required this.buffer,
+      required this.offsetInBytes,
+      required this.lengthInBytes});
+}
+
+/// A buffer that can be referenced by commands on the GPU.
+mixin Buffer {}
+
+/// [HostBuffer] is a [Buffer] which is allocated on the host and lazily
```
jonahwilliams, line 31: I would clarify what "host" means in this context, specifically: CPU resident memory native heap, not GPU memory or Dart heap.

lib/gpu/lib/src/buffer.dart
```diff
@@ -0,0 +1,73 @@
...
+  final Buffer buffer;
+  /// The start of the view, in bytes starting from the beginning of the
+  /// [buffer].
+  final int offsetInBytes;
+  /// The end of the view
+  final int lengthInBytes;
+
+  /// Create a new view into a buffer on the GPU.
+  const BufferView(
+      {required this.buffer,
```
jonahwilliams, line 23: nit: make the buffer argument positional, leave the others are required named parameters.

lib/gpu/lib/src/buffer.dart
```diff
@@ -0,0 +1,73 @@
...
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'dart:ffi';
+import 'dart:nativewrappers';
+import 'dart:typed_data';
+
+import 'context.dart';
+
+/// A buffer data range.
```
jonahwilliams, line 11: This is just restating the name of the class.

lib/gpu/lib/src/buffer.dart
bdero: Yeah, `DeviceBuffer` is the other kind of buffer we need to support, and they will always be usable in the same context as `HostBuffer`.

lib/gpu/lib/src/buffer.dart
bdero: Sure, done.

lib/gpu/lib/src/buffer.dart
bdero: Made this comment a bit more informative.

lib/gpu/lib/src/buffer.dart
bdero: Done. Your suggestion is nicer IMO, but do we happen to have any guidelines for this kind of thing I should seek to align myself with?

lib/gpu/lib/src/buffer.dart
bdero: Done.

lib/gpu/lib/src/context.dart
bdero: Removed for now. I added this here in the API draft because the other resource types (`Texture` and `DeviceBuffer`) need to be built with constructor methods.

lib/gpu/lib/src/buffer.dart
bdero: Sure, done.

lib/gpu/lib/src/buffer.dart
bdero: Done.

lib/gpu/lib/src/buffer.dart
bdero: I don't really know if this is how the final API is going to end up. My guess is that I'll work out something better when we add actual commands.

jonahwilliams: If you don't need it yet then don't add the interface yet.

lib/gpu/lib/src/buffer.dart
jonahwilliams: I'm not sure that there is a hard and fast rule for this, but I would think about named parameters in cases where the literal value or variable name itself does not give enough clarirty on what the argument is. The canonical example would be something like:

`void foo(true, false)`

What the heck are true / false doing here? So similar to how we might add a comment in the c++ code like 

`void foo(/*arg_name=*/true, /*arg_name=*/false)`, then we can use named arguments to make the Dart code clearer:

`void foo(argName: true, argName2, false)`.

So using the buffer constructor argument above, we've got something like:

`BufferView(buffer, 0, 12)`. The buffer itself is obvious, but the numbers benefit from clarification.

lib/gpu/lib/src/buffer.dart
bdero: Done.

bdero:
FYI @Hixie, I'm starting to land prototype patches for Flutter GPU this week. Roping you in since you mentioned you'd like to review the API. :)

Hixie:
Indeed! Is there a general overview of the whole intended API surface? Is this PR the place to start? Provide me with context. :-)

lib/gpu/lib/src/context.dart
```diff
@@ -5,6 +5,8 @@
 import 'dart:ffi';
 import 'dart:nativewrappers';
 
+import 'buffer.dart';
+
```
jonahwilliams, line 9: nit: unused

lib/gpu/lib/src/buffer.dart
```diff
@@ -0,0 +1,68 @@
...
+  final HostBuffer buffer;
+
+  /// The start of the view, in bytes starting from the beginning of the
+  /// [buffer].
+  final int offsetInBytes;
+
+  /// The length of the view.
+  final int lengthInBytes;
+
+  /// Create a new view into a buffer on the GPU.
```
jonahwilliams, line 21: its not really "on the gpu though"? For more functional pieces like the buffer view it may be easier to document once there are things that can be done with it. for example, "a view into a buffer that can be used for binding uniform data yada yada"

jonahwilliams (approved):
LGTM

bdero:
@Hixie Sure, the intention is for the API to just be a subset of Impeller's hardware abstraction layer (located under [`impeller/renderer`](https://github.com/flutter/engine/tree/main/impeller/renderer)).

A good jumping off point would be the [Detailed Design/Discussion](https://docs.google.com/document/d/1Sh1BAC5c_kkuMVreo7ymBzPoMzb7lamZRPsI7GBXv5M/edit?resourcekey=0-5w8u2V-LS41tCHeoE8bDTQ#heading=h.91qbbbrk2rmb) part of the doc. I started appending some more detail above the usage example today.

The entire initial API surface needed for implementing a renderer like the one built for the [FF Demo](https://github.com/bdero/flutter-scene-example/tree/master#impeller-scene-3d-demo) is covered by the [usage example](https://docs.google.com/document/d/1Sh1BAC5c_kkuMVreo7ymBzPoMzb7lamZRPsI7GBXv5M/edit?resourcekey=0-5w8u2V-LS41tCHeoE8bDTQ#heading=h.pdad6p7b9wb), although it has a couple of lingering TODOs floating around for minor tricky bits.

I also added a few designs for compositing Flutter GPU textures into the widget tree [here](https://docs.google.com/document/d/1Sh1BAC5c_kkuMVreo7ymBzPoMzb7lamZRPsI7GBXv5M/edit?resourcekey=0-5w8u2V-LS41tCHeoE8bDTQ#heading=h.xy89nizcq40m). Higher up is more preferred.

lib/gpu/lib/src/context.dart
bdero: Removed.

lib/gpu/lib/src/buffer.dart
bdero: I think `BufferView`s are effectively GPU-only because they'll only be used for binding buffers to `Command`s. I don't think we'd use a `BufferView` for updating a range on the host buffer, for example.

Hixie:
I'm in meetings non-stop tomorrow but I've scheduled some time on Wednesday to study this in depth.

bdero:
Ah, I just noticed the block of time on your cal for this. I'll leave it be for now

lib/gpu/lib/src/buffer.dart
```diff
@@ -0,0 +1,68 @@
...
+/// [HostBuffer] is a [Buffer] which is allocated on the host (native CPU
+/// resident memory) and lazily uploaded to the GPU. A [HostBuffer] can be
+/// safely mutated or extended at any time on the host, and will be
+/// automatically re-uploaded to the GPU the next time a GPU operation needs to
+/// access it.
+///
+/// This is useful for efficiently chunking sparse data uploads, especially
+/// ephemeral uniform data that needs to change from frame to frame.
+///
+/// Note: Different platforms have different data alignment requirements for
```
Hixie, line 35: trivial nit: per https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#avoid-empty-prose, avoid the `Note` bit. (Also I suspect indenting things like this will have a very different effect in the output than you might expect.)

lib/gpu/lib/src/buffer.dart
```diff
@@ -0,0 +1,68 @@
...
+///       accessing device buffer data. The [HostBuffer] takes these
+///       requirements into account and automatically inserts padding between
+///       emplaced data if necessary.
+class HostBuffer extends NativeFieldWrapperClass1 {
```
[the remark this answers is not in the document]
bdero: Oh, do you mean mark native classes as `base` to avoid the problems mentioned in: https://github.com/flutter/flutter/issues/123756? I think the weird class hierarchy solution that `ui.Picture` moved to is something we could safely migrate to if we wind up having the same testing needs.
Marked `HostBuffer` as `base`.

bdero:
Merging this to unblock the rest of this prototyping surface. This is experimental and not accessible by users.

PR #44711, opened 2023-08-15
Roll clang with fix for ABI change

zanderso:
In the change here https://github.com/llvm/llvm-project/commit/b653a2823fe4b4c9c6d85cfe119f31d8e70c2fa0, an intentional ABI breaking change was introduced to the clang runtime library for macOS and iOS. That change caused a symbol requiring dynamic linkage to be exposed that triggers iOS App Store checks for usage of private API.

This PR resolves that issue by rolling clang forward and introducing a definition of `_availability_version_check`. The declaration with weak linkage in the clang runtime library [here](https://github.com/llvm/llvm-project/blob/b653a2823fe4b4c9c6d85cfe119f31d8e70c2fa0/compiler-rt/lib/builtins/os_version_check.c#L89) will then be resolved against the definition introduced in this PR. Since the declaration in the clang runtime library will now be resolved by static linking, the Flutter dylib will no longer require it to be dynamically linked, and will therefore not trigger the App Store check for using private API.

The definition of `_availability_version_check` is implemented using the `dlsym` strategy used by the old version of clang [here](https://github.com/llvm/llvm-project/blob/f9ac5575675e4117c2e097a71ad0f02cab92aa97/compiler-rt/lib/builtins/os_version_check.c#L97).

Fixes https://github.com/flutter/flutter/issues/132130

### DIFF
--- a/shell/platform/darwin/common/availability_version_check.cc
+++ b/shell/platform/darwin/common/availability_version_check.cc
@@ -1,0 +1,48 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include <dispatch/dispatch.h>
+#include <dlfcn.h>
+#include <cstdint>
+
+#include "flutter/fml/logging.h"
+
+// See context in https://github.com/flutter/flutter/issues/132130 and
+// https://github.com/flutter/engine/pull/44711.
+
+// TODO(zanderso): Remove this after Clang 18 rolls into Xcode.
+// https://github.com/flutter/flutter/issues/133203
+
+namespace {
+
+typedef uint32_t dyld_platform_t;
+
+typedef struct {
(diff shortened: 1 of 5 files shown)


zanderso:
@dnfield A trial app with this fix made it through to testflight, so I think we're good to go.

shell/platform/darwin/common/availability_version_check.cc
```diff
@@ -0,0 +1,42 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
```
dnfield, line 1: Please add a comment somewhere about why we need this and what it's doing.

dnfield (approved):
LGTM but please add the comment so we can remember this

shell/platform/darwin/common/availability_version_check.cc
zanderso: D'oh! I added the comments below locally, and then forgot to upload them...

PR #44720, opened 2023-08-15
[web] add support for AppLifecycleState changes

maRci002:
closes flutter/flutter#53107

This PR introduces support for `AppLifecycleState` on Web, aligning the web's lifecycle events with those of the mobile platforms. This ensures a more consistent developer experience and better lifecycle management across all platforms.

**PR includes:**

- Page Visibility Handling: Integrated the `visibilitychange` DOM event to determine if the app is in a `resumed` or `paused` state based on the visibility state of the document.
- Page Transition Handling: Used `beforeunload` events to better manage the `detached` state.

## Pre-launch Checklist

- [x] I signed the [CLA].


maRci002:
@yjbanov @ditman, can you please add a reviewer?

lib/web_ui/lib/src/engine/platform_dispatcher.dart
```diff
@@ -60,6 +63,56 @@ class HighContrastSupport {
   }
 }
 
+/// Determines the [ui.AppLifecycleState] by listening to 'focus', 'blur', and
+/// 'beforeunload' events on the [DomWindow].
+class AppLifecycleStateSupport {
```
yjbanov, line 68: This implementation assumes we have a single global `FlutterView` that owns the entire page. Coincidentally, we are now introducing multi-view support where one "app" can have multiple `FlutterView` objects and the page's lifecycle may be owned by the host page/app, which may not even be written in Flutter or Dart (e.g. the host app can be a ReactJS app). The full-page mode isn't going away, however, so this functionality is reasonable to have. But maybe @ditman has some recommendations about how to add it in a way that won't make it more difficult to introduce the multi-view mode.

ditman: The biggest concern I have with this implementation is that the `focus`/`blur` events on window will trigger more often than expected. For example, when opening a "file selector", it is common practice to rely on window focus/blur to detect the moment that users selects (or doesn't) a file, or when opening another browser window (while the current one is still visible) probably shouldn't stop the execution of the current flutter app.

I'd recommend using the Page Visibility API[^1] (`onvisibilitychange` and `visibilityState`) rather than `onfocus`/`onblur`.

From a multi-view perspective, we might want to expose some JS API for host pages to programmatically control the lifecycle of the embedded Flutter App, (so it can be torn down properly when a ReactJS component is unmounting, for example), but as a default implementation, the global `visibilitychange` and `beforeunload` events are so low level that should be enough in the most common cases.

[^1]: https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API

lib/web_ui/lib/src/engine/platform_dispatcher.dart
```diff
@@ -86,6 +140,8 @@ class EnginePlatformDispatcher extends ui.PlatformDispatcher {
     accessibilityFeatures: computeAccessibilityFeatures(),
   );
 
+  ui.AppLifecycleState appLifecycleState = ui.AppLifecycleState.resumed;
```
ditman, line 143: This global should belong in the instance of the `AppLifecycleStateSupport` class (it is in fact, the state that the class is supporting).

lib/web_ui/lib/src/engine/platform_dispatcher.dart
```diff
@@ -971,6 +1028,15 @@ class EnginePlatformDispatcher extends ui.PlatformDispatcher {
...
   }
 
+  /// Updates [appLifecycleState] and invokes [_setAppLifecycleState] if
+  /// [appLifecycleState] changed.
+  void _updateAppLifecycleState(ui.AppLifecycleState value) {
+    if (value != appLifecycleState) {
+      appLifecycleState = value;
+      _setAppLifecycleState(value);
+    }
+  }
```
ditman, line 1038: Instead of having this loose function here, please modify the `AppLifecycleStateSupport` class so its instance requires a "broadcastLifecycleState" function when it's created, then pass the `_setAppLifecycleState` function when creating the `instance`.

lib/web_ui/lib/src/engine/platform_dispatcher.dart
```diff
@@ -60,6 +63,56 @@ class HighContrastSupport {
   }
 }
 
+/// Determines the [ui.AppLifecycleState] by listening to 'focus', 'blur', and
+/// 'beforeunload' events on the [DomWindow].
+class AppLifecycleStateSupport {
+  static AppLifecycleStateSupport instance = AppLifecycleStateSupport();
```
ditman, line 69: info: Creation of this class cannot be done inline now that it needs a `_setAppLifecycleState` function tear-off

lib/web_ui/lib/src/engine/platform_dispatcher.dart
```diff
@@ -60,6 +63,56 @@ class HighContrastSupport {
   }
 }
 
+/// Determines the [ui.AppLifecycleState] by listening to 'focus', 'blur', and
+/// 'beforeunload' events on the [DomWindow].
```
ditman, line 67: As you've seen, this platform_dispatcher file is enormous. Do you mind creating a `platform_dispatcher` directory next to it, and add a new `app_lifecycle_state.dart` file defining this class (and the typedef on line 23) that we import and use here?

That helps us keep the file small, and gives us a new place to start moving some of the other classes away from here!

lib/web_ui/test/engine/platform_dispatcher/platform_dispatcher_test.dart
```diff
@@ -37,6 +37,21 @@ void testMain() {
...
+      final MockAppLifecycleStateSupport mockAppLifecycleState =
+          MockAppLifecycleStateSupport();
+      AppLifecycleStateSupport.instance = mockAppLifecycleState;
+      final EnginePlatformDispatcher engineDispatcher =
+          EnginePlatformDispatcher();
+
+      expect(engineDispatcher.appLifecycleState, ui.AppLifecycleState.resumed);
+      mockAppLifecycleState.state = ui.AppLifecycleState.paused;
+      mockAppLifecycleState.invokeListeners(mockAppLifecycleState.state);
+      expect(engineDispatcher.appLifecycleState, ui.AppLifecycleState.paused);
```
ditman, line 50: It'd be nice if this tested that the `state` is propagated through the `flutter/lifecycle` method channel. Also that if the `state` is the same twice, the second time the call is skipped.

ditman: (If the class gets reimplemented as above, one could just `verify` that the function that is being passed to the created instance is being called with the correct `state` value.)

ditman (commented):
I think this is a nice addition. For now we can use only the Page Visibility API.

If we ever need an imperative version, we'd add pause/resume/unload methods to the state helper that would "inject" those events as if they came from the DOM.

I've added a few minor suggestions to improve encapsulation, and also to keep the platform_dispatcher.dart file small.

Thanks for this contribution!

maRci002:
Whenever I update my `main` branch from the `upstream` and then run `git merge main` on my `feat` branch, Git chooses the `ort strategy`. As a result, I end up with a bunch of commits from `main` appearing in this PR. These commits probably have different commit hashes.
@ditman, are you experiencing the same issue? I could use GitHub's `Update branch` button and/or resolve conflicts through GitHub's UI, but `git merge --ff-only` also fails.

Anyway I updated the PR to reflect your requested changes.

> It'd be nice if this tested that the state is propagated through the `flutter/lifecycle` method channel.

Currently, `EnginePlatformDispatcher` has an instance member `late AppLifecycleStateSupport appLifecycleStateSupport`, but it is set to a non-mocked version. Can you suggest some ideas on how to approach this test?

ditman:
> Whenever I update my main branch from the upstream and then run git merge main on my feat branch, Git chooses the ort strategy. As a result, I end up with a bunch of commits from main appearing in this PR. These commits probably have different commit hashes.
@ditman, are you experiencing the same issue?

@maRci002 nope, but I normally don't do `merge` to update `main` or my workr branches.

For my `main` branch, I `git fetch upstream` and then on `main` I do `git reset --hard upstream/main` (and then `git push origin/main` to keep my fork updated).

For work branches, after doing the above, I `git checkout work-branch` and then `git rebase main`. Then I have to `git push origin work-branch --force` for github to take the rebase (since commit hashes change)

maRci002:
> For my `main` branch, I `git fetch upstream` and then on `main` I do `git reset --hard upstream/main` (and then `git push origin/main` to keep my fork updated).
> 
> For work branches, after doing the above, I `git checkout work-branch` and then `git rebase main`. Then I have to `git push origin work-branch --force` for github to take the rebase (since commit hashes change)

I did that too, thanks.

lib/web_ui/test/engine/platform_dispatcher/platform_dispatcher_test.dart
maRci002: > It'd be nice if this tested that the `state` is propagated through the `flutter/lifecycle` method channel. Also that if the `state` is the same twice, the second time the call is skipped.

Can you suggest how to test this?

lib/web_ui/lib/src/engine/platform_dispatcher.dart
```diff
@@ -73,7 +73,10 @@ class EnginePlatformDispatcher extends ui.PlatformDispatcher {
     _addFontSizeObserver();
     _addLocaleChangedListener();
     registerHotRestartListener(dispose);
-    _setAppLifecycleState(ui.AppLifecycleState.resumed);
+    appLifecycleStateSupport = DomAppLifecycleStateSupport(
+      broadcastLifecycleState: _setAppLifecycleState,
+    );
+    _setAppLifecycleState(appLifecycleStateSupport.appLifecycleState);
```
ditman, line 79: Minor suggestion: broadcast this initial state from the constructor of the `DomAppLifecycleStateSupport`, so we don't need to manually call the setter here.

(If we ever want to prevent this behavior, we can add a `broadcastInitialState` or similar to the constructor to enable/disable it, but for now, it seems that we always want to broadcast the `appLifecycleState` next to the creation of the DomAppLifecycleStateSupport object, so I'd move it to the constructor).

ditman (commented):
This looks good to me now. Some comments:

It seems that `flutter/lifecycle` is not implemented at all in the `_sendPlatformMessage` method. Other than tracking the state of the App Lifecycle, this PR is not really introducing a concept of a lifecycle into the engine, right? (we seem to not be doing anything with the `flutter/lifecycle` message).

About the full-blown test, after moving the lifecycle to its own class, I think it's enough that you're logging that the broadcastLifecycleState is being called as needed, and not duplicated. Since the engine is still not responding to the actual lifecycle changes, the more complex integration test can wait.

(Small comment about the first broadcast of the lifecycle though)

ditman:
PS: To fix the "licenses" check, you need to apply the following diff to the `flutter/ci/licenses_golden/licenses_flutter` file:

```diff
--- flutter/ci/licenses_golden/licenses_flutter	2023-09-01 00:42:21.537516151 -0700
+++ out/license_script_output/licenses_flutter	2023-09-01 00:48:39.022751344 -0700
@@ -2042,12 +2042,13 @@
 ORIGIN: ../../../flutter/lib/web_ui/lib/src/engine/navigation/history.dart + ../../../flutter/LICENSE
 ORIGIN: ../../../flutter/lib/web_ui/lib/src/engine/noto_font.dart + ../../../flutter/LICENSE
 ORIGIN: ../../../flutter/lib/web_ui/lib/src/engine/noto_font_encoding.dart + ../../../flutter/LICENSE
 ORIGIN: ../../../flutter/lib/web_ui/lib/src/engine/onscreen_logging.dart + ../../../flutter/LICENSE
 ORIGIN: ../../../flutter/lib/web_ui/lib/src/engine/picture.dart + ../../../flutter/LICENSE
 ORIGIN: ../../../flutter/lib/web_ui/lib/src/engine/platform_dispatcher.dart + ../../../flutter/LICENSE
+ORIGIN: ../../../flutter/lib/web_ui/lib/src/engine/platform_dispatcher/app_lifecycle_state.dart + ../../../flutter/LICENSE
 ORIGIN: ../../../flutter/lib/web_ui/lib/src/engine/platform_views.dart + ../../../flutter/LICENSE
 ORIGIN: ../../../flutter/lib/web_ui/lib/src/engine/platform_views/content_manager.dart + ../../../flutter/LICENSE
 ORIGIN: ../../../flutter/lib/web_ui/lib/src/engine/platform_views/message_handler.dart + ../../../flutter/LICENSE
 ORIGIN: ../../../flutter/lib/web_ui/lib/src/engine/platform_views/slots.dart + ../../../flutter/LICENSE
 ORIGIN: ../../../flutter/lib/web_ui/lib/src/engine/plugins.dart + ../../../flutter/LICENSE
 ORIGIN: ../../../flutter/lib/web_ui/lib/src/engine/pointer_binding.dart + ../../../flutter/LICENSE
@@ -4793,12 +4794,13 @@
 FILE: ../../../flutter/lib/web_ui/lib/src/engine/navigation/history.dart
 FILE: ../../../flutter/lib/web_ui/lib/src/engine/noto_font.dart
 FILE: ../../../flutter/lib/web_ui/lib/src/engine/noto_font_encoding.dart
 FILE: ../../../flutter/lib/web_ui/lib/src/engine/onscreen_logging.dart
 FILE: ../../../flutter/lib/web_ui/lib/src/engine/picture.dart
 FILE: ../../../flutter/lib/web_ui/lib/src/engine/platform_dispatcher.dart
+FILE: ../../../flutter/lib/web_ui/lib/src/engine/platform_dispatcher/app_lifecycle_state.dart
 FILE: ../../../flutter/lib/web_ui/lib/src/engine/platform_views.dart
 FILE: ../../../flutter/lib/web_ui/lib/src/engine/platform_views/content_manager.dart
 FILE: ../../../flutter/lib/web_ui/lib/src/engine/platform_views/message_handler.dart
 FILE: ../../../flutter/lib/web_ui/lib/src/engine/platform_views/slots.dart
 FILE: ../../../flutter/lib/web_ui/lib/src/engine/plugins.dart
 FILE: ../../../flutter/lib/web_ui/lib/src/engine/pointer_binding.dart
=================================================================
```

(It's complaining that we added a new file with a Flutter license that is not catalogued in the golden file. It should be added in alphabetical order in a couple of places!)

ditman (approved):
This LGTM, but I'd like @mdebbar to take a second look. I know he's knee deep refactoring the platform_dispatcher and friends!

ditman:
@mdebbar this is a new singleton/global that monitors the visibility API to decide if an app should be paused/resumed, and when it initializes. PTAL and see if it jives well with what you've been doing so far / plan on doing later! (Thanks!)

lib/web_ui/lib/src/engine/platform_dispatcher/app_lifecycle_state.dart
```diff
@@ -0,0 +1,61 @@
...
+    onAppLifecycleStateChange(ui.AppLifecycleState.detached);
+  });
+
+  late final DomEventListener _visibilityChangeListener =
+      createDomEventListener((DomEvent event) {
+    if (domDocument.visibilityState == 'visible') {
+      onAppLifecycleStateChange(ui.AppLifecycleState.resumed);
+    } else if (domDocument.visibilityState == 'hidden') {
+      onAppLifecycleStateChange(ui.AppLifecycleState.paused);
+    }
```
mdebbar, line 59: Reading through [mdn](https://developer.mozilla.org/en-US/docs/Web/API/Document/visibilityState) and Flutter [docs](https://api.flutter.dev/flutter/dart-ui/AppLifecycleState.html), it's not 100% clear to me whether this should be `paused` or `hidden`. But I think `hidden` might be better, @gspencergoog?

lib/web_ui/lib/src/engine/platform_dispatcher/app_lifecycle_state.dart
```diff
@@ -0,0 +1,61 @@
...
+import 'package:meta/meta.dart';
+import 'package:ui/ui.dart' as ui;
+
+import '../../engine.dart';
+
+/// Signature of functions added as a listener to [ui.AppLifecycleState] changes
+typedef AppLifecycleStateListener = void Function(ui.AppLifecycleState state);
+
+/// Determines the [ui.AppLifecycleState].
+abstract class AppLifecycleStateSupport {
```
mdebbar, line 14: If this is designed to be an app-level instance (as opposed to view-level), then let's make it a clear singleton by following the same pattern we use for other singletons. The listener callback doesn't have to be provided to the constructor, it can instead be added later. Here's a rough idea:
```dart
abstract class AppLifecycleState {
  static final AppLifecycleState instance = _BrowserAppLifecycleState();

  ui.AppLifecycleState get appLifecycleState => _appLifecycleState;
  ui.AppLifecycleState _appLifecycleState = ui.AppLifecycleState.resumed;

  final List<AppLifecycleStateListener> _listeners = <AppLifecycleStateListener>[];

  void addListener(AppLifecycleStateListener listener) {
    if (_listeners.isEmpty) {
      activate();
    }
    _listeners.add(listener);
    listener(_appLifecycleState);
  }

  void removeListener(AppLifecycleStateListener listener) {
    _listeners.remove(listener);
    if (_listeners.isEmpty) {
      deactivate();
    }
  }

  @protected
  void activate();

  @protected
  void deactivate();

  @visibleForTesting
  void onAppLifecycleStateChange(ui.AppLifecycleState newState) {
    if (newState != _appLifecycleState) {
      _appLifecycleState = newState;
      for (final AppLifecycleStateListener listener in _listeners) {
        listener(newState);
      }
    }
  }
}

class _BrowserAppLifecycleState extends AppLifecycleState {
  @override
  void activate() {
    domWindow.addEventListener('beforeunload', _beforeUnloadListener);
    domDocument.addEventListener('visibilitychange', _visibilityChangeListener);
  }

  @override
  void deactivate() {
    domWindow.removeEventListener('beforeunload', _beforeUnloadListener);
    domDocument.removeEventListener('visibilitychange', _visibilityChangeListener);
  }

  // ...
}
```

mdebbar (commented):
This is looking good! I just have a few suggestions to improve naming and code structure a bit.

lib/web_ui/lib/src/engine/platform_dispatcher/app_lifecycle_state.dart
maRci002: I'm hesitant to implement the singleton version as it may compromise the existing multi-view support.
https://github.com/flutter/engine/pull/44720#discussion_r1312215165

ditman: @maRci002 since this is using the visibility observer (which is a global browser attribute), this definitely can be a singleton.

In the future we could receive more signals that may depend on multi-view basis, for example, if we were to use an IntersectionObserver to decide if *any* flutter view is visible in the viewport of the browser or not to mark the app as "invisible"; but for these global browser properties, a singleton is fine!

PS: **Thanks for caring** about the multi-view migration, btw!!

mdebbar:
This looks good to me! Thanks for the contribution!

I just want to hear from @gspencergoog about the `paused` vs `hidden` states (comment [here](https://github.com/flutter/engine/pull/44720#discussion_r1357288754)).

lib/web_ui/lib/src/engine/platform_dispatcher/app_lifecycle_state.dart
gspencergoog: Yes, I think `AppLifecycleState.hidden` makes more sense here, since that matches what the DOM is telling you. Paused is typically only entered on non-web mobile platforms.

lib/web_ui/lib/src/engine/platform_dispatcher/app_lifecycle_state.dart
```diff
@@ -0,0 +1,86 @@
...
+    }
+  }
+}
+
+/// Determines the [ui.AppLifecycleState] by listening to 'beforeunload',
+/// 'visibilitychange' events on the [DomWindow] / [DomHTMLDocument].
+class _BrowserAppLifecycleState extends AppLifecycleState {
+  @override
+  void activate() {
+    domWindow.addEventListener('beforeunload', _beforeUnloadListener);
```
gspencergoog, line 60: This does have the possibility of making Flutter apps less efficient on Firefox (and probably other browsers). See https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event#usage_notes

We might need to hook up a way to know if someone is listening to the application lifecycle state in the framework so that we can only register for this event if someone is listening.

In the meantime, I think this is OK to check in, but you might add a TODO to make this improvement.

lib/web_ui/lib/src/engine/platform_dispatcher/app_lifecycle_state.dart
```diff
@@ -0,0 +1,86 @@
...
+    domDocument.addEventListener('visibilitychange', _visibilityChangeListener);
```
gspencergoog, line 61: Why isn't there also a listener here for `windows.onFocusChanged` that sets the lifecycle state to `inactive`? This is a useful state for applications, since they can unfocus widgets, close menus, or otherwise indicate that they no longer have input focus.

gspencergoog (commented):
Thanks for this patch!  It's the last piece in the app lifecycle puzzle!

Just some comments: It seems like the web should handle both an `inactive` (unfocused) state, and a `hidden` state, but I'm not seeing those here. The detached and paused states are generally only used on non-web mobile apps, but I'm not opposed to using `detached` for an unloaded state.  If we do that, we need to change the documentation for the states to reflect that.

maRci002:
@gspencergoog, I've implemented the changes you suggested.
To illustrate the browser's behavior with these adjustments, I've prepared a small HTML + JS demo.

<details><summary>Code sample</summary>

```html
<!DOCTYPE html>
<html lang="en">
<head>
<meta charset="UTF-8">
<meta name="viewport" content="width=device-width, initial-scale=1.0">
<title>App Lifecycle State Tracker</title>
<script>
document.addEventListener('DOMContentLoaded', function () {
    var appStateList = document.getElementById('appStateList');

    function addState(newState) {
        var stateItem = document.createElement('li');
        stateItem.textContent = newState;
        appStateList.appendChild(stateItem);
    }

    // Listen for visibility changes
    document.addEventListener('visibilitychange', function () {
        if (document.visibilityState === 'visible') {
            addState('AppLifecycleState.resumed - (visible)');
        } else if (document.visibilityState === 'hidden') {
            addState('AppLifecycleState.hidden');
        }
    });

    // Listen for the page unloading
    window.addEventListener('beforeunload', function () {
        addState('AppLifecycleState.detached');
    });

    // Listen for when the page gains focus
    window.addEventListener('focus', function () {
        addState('AppLifecycleState.resumed - (focus)');
    });

    // Listen for when the page loses focus
    window.addEventListener('blur', function () {
        addState('AppLifecycleState.inactive');
    });

    document.getElementById('fileSelector').addEventListener('click', function() {
        // Open the file selector
        document.getElementById('dummyFileInput').click();
    });

    document.getElementById('newTab').addEventListener('click', function() {
        // Open a new tab
        window.open('https://www.example.com', '_blank');
    });

    document.getElementById('newWindow').addEventListener('click', function() {
        // Open a new window
        window.open('https://www.example.com', '_blank', 'noopener,noreferrer');
    });
});
</script>
</head>
<body>
<h1>App Lifecycle State Tracker</h1>

<input type="file" id="dummyFileInput" style="display: none;">
<button id="fileSelector">Open File Selector</button>
<button id="newTab">Open New Tab</button>
<button id="newWindow">Open New Window</button>

<ul id="appStateList"></ul>
</body>
</html>
```

https://github.com/flutter/engine/assets/8436039/c81ea902-d0f0-44c0-a1fb-f08f1d1f5678

</details>

gspencergoog (approved):

Thanks so much!

lib/web_ui/lib/src/engine/platform_dispatcher/app_lifecycle_state.dart
```diff
@@ -0,0 +1,106 @@
...
+  });
+
+  late final DomEventListener _blurListener =
+      createDomEventListener((DomEvent event) {
+    onAppLifecycleStateChange(ui.AppLifecycleState.inactive);
+  });
+
+  late final DomEventListener _beforeUnloadListener =
+      createDomEventListener((DomEvent event) {
+    onAppLifecycleStateChange(ui.AppLifecycleState.detached);
```
gspencergoog, line 95: Since this uses detached on all web platforms you'll need to change the documentation for the AppLifecycleState class, both in the engine and for the framework.

https://github.com/gspencergoog/engine/blob/main/shell/platform/common/app_lifecycle_state.h#L36-L37

https://github.com/gspencergoog/engine/blob/main/lib/ui/platform_dispatcher.dart#L1794-L1795

maRci002:
> auto label is removed for flutter/engine/44720, due to - The status or check suite [Linux Web Framework tests](https://ci.chromium.org/b/8766959546866979937) has failed. Please fix the issues identified (or deflake) before re-applying this label.

I've noticed many [platform-web PR](https://github.com/flutter/engine/labels/platform-web)s fail the 'Linux Web Framework tests'. Is there a specific action I can take to address this?

ditman:
@maRci002 : I haven't had the time to locally test this on my development environment, but it seems that some tests are timing out, like this one:

https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8766953075319353121/+/u/test:_run_suite_chrome-full-dart2wasm-canvaskit-ui/stdout

Can you try and `felt test` [`draw_atlas_golden_test.dart`](https://github.com/flutter/engine/blob/main/lib/web_ui/test/ui/draw_atlas_golden_test.dart)

From the test log above, it seems to be hanging somewhere in [this common initialization method](https://github.com/flutter/engine/blob/main/lib/web_ui/test/common/test_initialization.dart#L17) (so probably other tests that utilize that initializating might time out as well)

(Maybe we need to "fake" a different app lifecycle state for the tests? Or disable the detection altogether?)

gspencergoog:
Where are we on this?  I'd like to land it soon if possible. @maRci002 if you aren't able to, or don't have time to address the remaining issues, please let us know.

maRci002:
Hey @gspencergoog, I was able to resolve the earlier timeout issue.

However, I've now come across a new challenge while working with the `flutter_test` package, specifically the [`lifecycle_test.dart`](https://github.com/flutter/flutter/blob/master/packages/flutter/test/services/lifecycle_test.dart) file. The code expects `expect(ServicesBinding.instance.lifecycleState, isNull)`, but in my implementation, an initial 'resumed' state is being sent right at the start, which seems to be causing conflicts.

I'm not sure how to best handle this in `lifecycle_test.dart`. Do you have any suggestions or workarounds for managing the lifecycle state in this scenario?

maRci002:
> I don't think that expectation is needed anyhow. I'm going to remove it, see this PR: [flutter/flutter#141822](https://github.com/flutter/flutter/pull/141822)

Thanks for the fast pull request; now every check passes.

PR #44730, opened 2023-08-15
[Skia] Only respect ui.Paint.dither when the colorSource is a gradient

matanlurey:
In the Impeller backend, we **only** support dithering of _gradients_. In addition, it will be the default (and only option).

In the [process of enabling dithering by default](https://github.com/flutter/engine/pull/44705), i.e.
```diff
class Paint {
-  static bool enableDithering = false;
+  static bool enableDithering = true;
}
```

... we realized with internal Google testing this will now apply dithering on more than just gradients, i.e. images in the Skia backend. Since we won't support dithering of images in the Impeller backend, this PR gives a "hint" on whether the `colorSource` (if one is set) can be dithered by the contrived rules we've created.


jonahwilliams (approved):
LGTM

display_list/skia/dl_sk_canvas.cc
```diff
@@ -23,11 +23,21 @@ constexpr float kInvertColorMatrix[20] = {
...
-  sk_paint.setDither(paint.isDither());
+
+  // On the Impeller backend, we will only support dithering of *gradients*,
+  // and it will be enabled by default (without the option to disable it). Until
+  // Skia support is completely removed, we only want to respect the dither flag
+  // for gradients (otherwise it will also apply to, for example, images, which
+  // is not supported in Impeller).
+  //
+  // See https://github.com/flutter/flutter/issues/112498.
+  if (paint.isDitherHintForSkBackend()) {
```
flar, line 38: A simpler and more direct solution would be to check ColorSource for null down below - which will skip both the call to ToSk and the need to deal with dithering for the very common case where it is null. Down below:

```
  auto color_source = paint.getColorSourcePtr();
  if (color_source) {
    sk_paint.setDither(color_source->isDither...());
    sk_paint.setShader(ToSk(color_source));
  }
```

display_list/dl_paint.h
```diff
@@ -66,6 +66,20 @@ class DlPaint {
...
+  // *never* apply dithering otherwise, and there will be no user-facing feature
+  // to change this behavior.
+  //
+  // To temporarily match this behavior in the Skia backend, we tag gradients
+  // with a special flag, and then check for that flag when converting from
+  // DlPaint to SkPaint.
+  //
+  // TODO(matanl): Remove this flag when the Skia backend is removed,
+  // https://github.com/flutter/flutter/issues/112498.
+  bool isDitherHintForSkBackend() const {
```
flar, line 79: Simpler name? "requiresDithering"? "shouldBeDithered"? Not clear on why we need to name a backend for this property. It's true that only the Skia backend will need to do something with this information, but it's not like the information is only valid under Skia - it's just a new fundamental implicit property of a dither.

flar: Or, "isGradient" and leave it at that. Skia knows what it must do in that case. Impeller does it anyway.

flar (commented):
Suggested changes.

(Though I really dislike the concept of "ForSuchAndSuchBackend" when describing a basic property of an object. Really, it's just identifying that a color source is some kind of gradient...)

matanlurey:
> Though I really dislike the concept of "ForSuchAndSuchBackend" when describing a basic property of an object. Really, it's just identifying that a color source is some kind of gradient...

Fair enough! I wanted to make sure I (and nobody else) felt like it was anything more than a hack for a specific backend, but I don't have a strong feeling about the name specifically.

display_list/skia/dl_sk_canvas.cc
matanlurey: Nice suggestion, done!

display_list/dl_paint.h
matanlurey: Fair enough! Done and updated.

display_list/skia/dl_sk_canvas.cc
```diff
@@ -49,6 +47,19 @@ static SkPaint ToSk(const DlPaint& paint, bool force_stroke = false) {
...
   sk_paint.setColorFilter(color_filter);
+
+  // On the Impeller backend, we will only support dithering of *gradients*,
+  // and it will be enabled by default (without the option to disable it). Until
+  // Skia support is completely removed, we only want to respect the dither flag
+  // for gradients (otherwise it will also apply to, for example, images, which
+  // is not supported in Impeller).
+  //
+  // See https://github.com/flutter/flutter/issues/112498.
+  auto color_source = paint.getColorSourcePtr();
```
flar, line 58: Combine this with the code up above that calls setShader. No need to get the ColorSourcePtr twice...

display_list/dl_paint.h
```diff
@@ -66,6 +66,20 @@ class DlPaint {
...
 
+  // In the Impeller backend, we'll *always* apply dithering to gradients, and
+  // *never* apply dithering otherwise, and there will be no user-facing feature
+  // to change this behavior.
+  //
+  // To temporarily match this behavior in the Skia backend, we tag gradients
+  // with a special flag, and then check for that flag when converting from
+  // DlPaint to SkPaint.
+  //
+  // TODO(matanl): Remove this flag when the Skia backend is removed,
```
flar, line 77: Can't this be removed now? This method isn't called any more.

flar (commented):
Should remove dead code.

display_list/skia/dl_sk_canvas.cc
matanlurey: Done.

display_list/dl_paint.h
matanlurey: Done.

ci/licenses_golden/excluded_files
```diff
@@ -41,6 +41,7 @@
 ../../../flutter/display_list/effects/dl_path_effect_unittests.cc
 ../../../flutter/display_list/geometry/dl_region_unittests.cc
 ../../../flutter/display_list/geometry/dl_rtree_unittests.cc
+../../../flutter/display_list/skia/dl_sk_canvas_unittests.cc
```
flar, line 44: Just noticed this - these tests can go in the existing dl_sk_conversions_unittests.cc which combines all of the tests for all of the conversions in all of the skia files.

display_list/skia/dl_sk_canvas.cc
```diff
@@ -49,6 +45,20 @@ static SkPaint ToSk(const DlPaint& paint, bool force_stroke = false) {
...
+  sk_paint.setShader(ToSk(color_source));
```
flar, line 57: Could move this inside the if. What was wrong with my original pseudo-code?

Also, sk_paint.setDither should be based on whether or not the color source is a gradient and not on paint.isDither()...

```
auto color_source = paint.getColorSourcePtr();
if (color_source) {
  sk_paint.setDither(color_source->isGradient());
  sk_paint.setShader(ToSk(color_source));
}
```

flar (changes requested):

ci/licenses_golden/excluded_files
matanlurey: Sure thing - I was trying to follow the (inferred) convention of `foo.cc` has `foo_unittests.cc`.

display_list/skia/dl_sk_canvas.cc
matanlurey: Done (inside the `if`).

For your second question, no, that's not the behavior we want. It's a bit convoluted, but we want to set dithering IFF:
1. the source is a gradient
1. dithering is requested

Dithering will soon be enabled by _default_ (it's currently opt-in), but we will still allow users to opt-out for a period fo time.

display_list/skia/dl_sk_conversions_unittests.cc
```diff
@@ -2,13 +2,14 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
-#include "flutter/display_list/skia/dl_sk_conversions.h"
-
+#include "display_list/effects/dl_color_source.h"
```
flar, line 5: This should be grouped with the other DL includes below and with the proper include path (`flutter/display_list/`etc.)

flar (changes requested):

display_list/skia/dl_sk_canvas.cc
```diff
@@ -49,6 +45,22 @@ static SkPaint ToSk(const DlPaint& paint, bool force_stroke = false) {
...
+  if (color_source) {
+    // On the Impeller backend, we will only support dithering of *gradients*,
+    // and it will be enabled by default (without the option to disable it).
+    // Until Skia support is completely removed, we only want to respect the
+    // dither flag for gradients (otherwise it will also apply to, for example,
+    // images, which is not supported in Impeller).
+    //
+    // See https://github.com/flutter/flutter/issues/112498.
+    if (color_source->isGradient()) {
+      sk_paint.setDither(paint.isDither());
```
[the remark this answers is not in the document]
matanlurey: > I thought the purpose of this fix was that dithering happens when gradients are involved 

Yeah that's the end goal, but as written here:

> // On the Impeller backend, we will only support dithering of *gradients*,
> // and it will be enabled by default (without the option to disable it).
> // Until Skia support is completely removed ...

... that's an aspirational state. The current behavior (see issue https://github.com/flutter/engine/pull/44705) is that users *are* allowed to disable dithering, so we don't want to _unconditionally_ enable or disable dithering. I'm open to making this more clear - and given the back and forth on the PR, have done so already.

display_list/skia/dl_sk_canvas.cc
flar: Ah, future tense. I think I missed that...

So, first pass, only set dither inside color_source handling.
Second pass, if making it the default sticks, eliminate the option and the method...

flar (approved):
LGTM

(technically, ToSk(paint) could be moved to the conversions file. I'm not sure why that appears on its own (and I wrote this))

PR #44734, opened 2023-08-15
Add support for SurfaceTexture based external textures on Android under Impeller/GLES.

johnmccutchan:
- Refactor AndroidExternalTextureGL into SurfaceTextureExternalTexture base class with two specialized implementations:
- Add SurfaceTextureExternalTextureGL for legacy GL rendering.
- Add SurfaceTextureExternalTextureImpellerGL for Impeller GLES rendering.

NOTE: There is still no support for SurfaceTextures under Vulkan rendering.


shell/platform/android/surface_texture_external_texture.h
```diff
@@ -1,9 +1,10 @@
+
```
jonahwilliams, line 1:
```suggestion
```

shell/platform/android/surface_texture_external_texture_gl.cc
```diff
@@ -0,0 +1,115 @@
...
+
+void SurfaceTextureExternalTextureImpellerGL::ProcessFrame(
+    PaintContext& context,
+    const SkRect& bounds) {
+  if (state_ == AttachmentState::kUninitialized) {
+    // Generate the texture handle.
+    impeller::TextureDescriptor desc;
+    desc.type = impeller::TextureType::kTextureExternalOES;
+    desc.storage_mode = impeller::StorageMode::kDevicePrivate;
+    desc.format = impeller::PixelFormat::kR8G8B8A8UNormInt;
```
jonahwilliams, line 86: Same comment as last PR: we need to look this up at runtime right?

shell/platform/android/platform_view_android.cc
```diff
@@ -301,10 +301,20 @@ void PlatformViewAndroid::RegisterExternalTexture(
     int64_t texture_id,
     const fml::jni::ScopedJavaGlobalRef<jobject>& surface_texture) {
   if (android_context_->RenderingApi() == AndroidRenderingAPI::kOpenGLES) {
-    RegisterTexture(std::make_shared<AndroidExternalTextureGL>(
-        texture_id, surface_texture, jni_facade_));
+    if (android_context_->GetImpellerContext() != nullptr) {
```
jonahwilliams, line 304: I think we'd probably just write it like
```suggestion
    if (android_context_->GetImpellerContext()) {
```

jonahwilliams (approved):
LGTM

shell/platform/android/surface_texture_external_texture_gl.cc
jonahwilliams: Interesting! right, I guess its GL. I think we also do something similar in the GLES Impeller backend.

shell/platform/android/surface_texture_external_texture.h
johnmccutchan: Done.

shell/platform/android/platform_view_android.cc
johnmccutchan: Done.

PR #44776, opened 2023-08-16
[Impeller] unconditionally present with transaction.

jonahwilliams:
Testing the potential fix for iOS scheduling issues.


jonahwilliams:
@dnfield @chinmaygarde @zanderso

zanderso (changes requested):
Not actually requesting changes. Just noting that this should not land until we've discussed this further with all the stakeholders.

zanderso (approved):

PR #44779, opened 2023-08-16
[ios][ios17]fix auto correction highlight on top left corner

hellohuanlin:
Fix native auto-correction highlight region on top left corner. 

This PR uses the system auto-correction highlight on iOS 17, which was disabled by https://github.com/flutter/engine/pull/44354

Screenshot 2023-08-16 at 1 19 39 PM

Fixes https://github.com/flutter/flutter/issues/131622
Fixes https://github.com/flutter/flutter/issues/131695
Fixes https://github.com/flutter/flutter/issues/130818

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
+++ b/shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
@@ -2462,7 +2462,7 @@ static BOOL IsSelectionRectBoundaryCloserToPoint(CGPoint point,
         CGRectMake(0, 0, [dictionary[@"width"] intValue], [dictionary[@"height"] intValue]);
     _activeView.tintColor = [UIColor clearColor];
   } else {
-    // TODO: Also need to handle iOS 16 case, where the auto-correction highlight does
+    // TODO(hellohuanlin): Also need to handle iOS 16 case, where the auto-correction highlight does
     // not match the size of text.
     // See https://github.com/flutter/flutter/issues/131695
     if (@available(iOS 17, *)) {
--- a/shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
+++ b/shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
@@ -2451,8 +2451,8 @@ static BOOL IsSelectionRectBoundaryCloserToPoint(CGPoint point,
 - (void)setEditableSizeAndTransform:(NSDictionary*)dictionary {
   NSArray* transform = dictionary[@"transform"];
   [_activeView setEditableTransform:transform];
-  int leftIndex = 12;
-  int topIndex = 13;
+  const int leftIndex = 12;
+  const int topIndex = 13;
   if ([_activeView isScribbleAvailable]) {
     // This is necessary to set up where the scribble interactable element will be.
     _inputHider.frame =
@@ -2462,6 +2462,9 @@ static BOOL IsSelectionRectBoundaryCloserToPoint(CGPoint point,
         CGRectMake(0, 0, [dictionary[@"width"] intValue], [dictionary[@"height"] intValue]);
     _activeView.tintColor = [UIColor clearColor];
   } else {
+    // TODO: Also need to handle iOS 16 case, where the auto-correction highlight does
+    // not match the size of text.
+    // See https://github.com/flutter/flutter/issues/131695
     if (@available(iOS 17, *)) {
       // Move auto-correction highlight to overlap with the actual text.
       // This is to fix an issue where the system auto-correction highlight is displayed at
@@ -2498,16 +2501,20 @@ static BOOL IsSelectionRectBoundaryCloserToPoint(CGPoint point,
                                                          : NSWritingDirectionRightToLeft]];
   }
 
+  BOOL shouldNotifyTextChange = NO;
   if (@available(iOS 17, *)) {
     // Force UIKit to query the selectionRects again on iOS 17+
     // This is to fix a bug on iOS 17+ where UIKit queries the outdated selectionRects after
     // entering a character, resulting in auto-correction highlight region missing the last
     // character.
+    shouldNotifyTextChange = YES;
+  }
+  if (shouldNotifyTextChange) {
     [_activeView.inputDelegate textWillChange:_activeView];
-    _activeView.selectionRects = rectsAsRect;
+  }
+  _activeView.selectionRects = rectsAsRect;
+  if (shouldNotifyTextChange) {
     [_activeView.inputDelegate textDidChange:_activeView];
-  } else {
-    _activeView.selectionRects = rectsAsRect;
   }
 }
 
--- a/shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
+++ b/shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
@@ -2449,18 +2449,27 @@ static BOOL IsSelectionRectBoundaryCloserToPoint(CGPoint point,
 }
 
 - (void)setEditableSizeAndTransform:(NSDictionary*)dictionary {
-  [_activeView setEditableTransform:dictionary[@"transform"]];
+  NSArray* transform = dictionary[@"transform"];
+  [_activeView setEditableTransform:transform];
+  int leftIndex = 12;
+  int topIndex = 13;
   if ([_activeView isScribbleAvailable]) {
     // This is necessary to set up where the scribble interactable element will be.
-    int leftIndex = 12;
-    int topIndex = 13;
     _inputHider.frame =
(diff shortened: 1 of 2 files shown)


shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
```diff
@@ -2449,18 +2449,27 @@ - (void)takeKeyboardScreenshotAndDisplay {
 }
 
 - (void)setEditableSizeAndTransform:(NSDictionary*)dictionary {
-  [_activeView setEditableTransform:dictionary[@"transform"]];
+  NSArray* transform = dictionary[@"transform"];
+  [_activeView setEditableTransform:transform];
+  int leftIndex = 12;
+  int topIndex = 13;
```
stuartmorgan, line 2455: Nit: these should be `const`.

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
```diff
@@ -2488,7 +2497,18 @@ - (void)setSelectionRects:(NSArray*)encodedRects {
...
+  if (@available(iOS 17, *)) {
+    // Force UIKit to query the selectionRects again on iOS 17+
+    // This is to fix a bug on iOS 17+ where UIKit queries the outdated selectionRects after
+    // entering a character, resulting in auto-correction highlight region missing the last
+    // character.
+    [_activeView.inputDelegate textWillChange:_activeView];
+    _activeView.selectionRects = rectsAsRect;
+    [_activeView.inputDelegate textDidChange:_activeView];
+  } else {
+    _activeView.selectionRects = rectsAsRect;
```
stuartmorgan, line 2510: Rather than duplicate this line, which could cause drift over time, we should refactor to look more like the other method where there's a bool for sending updates.:

```
BOOL triggerTextUpdates = NO;
if (@available(iOS 17, *)) {
  // move explanatory comment here.
  triggerTextUpdates = YES;
}
if (triggerTextUpdates) {
  [_activeView.inputDelegate textWillChange:_activeView];
}
_activeView.selectionRects = rectsAsRect;
if (triggerTextUpdates) {
  [_activeView.inputDelegate textDidChange:_activeView];
}
```

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
```diff
@@ -2449,18 +2449,27 @@ - (void)takeKeyboardScreenshotAndDisplay {
...
-        CGRectMake([dictionary[@"transform"][leftIndex] intValue],
-                   [dictionary[@"transform"][topIndex] intValue], [dictionary[@"width"] intValue],
-                   [dictionary[@"height"] intValue]);
+        CGRectMake([transform[leftIndex] intValue], [transform[topIndex] intValue],
+                   [dictionary[@"width"] intValue], [dictionary[@"height"] intValue]);
     _activeView.frame =
         CGRectMake(0, 0, [dictionary[@"width"] intValue], [dictionary[@"height"] intValue]);
     _activeView.tintColor = [UIColor clearColor];
+  } else {
+    if (@available(iOS 17, *)) {
```
stuartmorgan, line 2465: Please add a TODO to simplify all of this logic. It looks like post cherry-pick we can probably just make the `_inputHider.frame` logic unconditional?

stuartmorgan (changes requested):

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
hellohuanlin: The iOS 16 behavior is different - the highlight region does not match the size of text (see screenshot in https://github.com/flutter/flutter/issues/131695 for iOS 16). But will add a TODO to explain this.

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
```diff
@@ -2449,18 +2449,30 @@ - (void)takeKeyboardScreenshotAndDisplay {
...
     _activeView.tintColor = [UIColor clearColor];
+  } else {
+    // TODO(hellohuanlin): Also need to handle iOS 16 case, where the auto-correction highlight does
+    // not match the size of text.
+    // See https://github.com/flutter/flutter/issues/131695
+    if (@available(iOS 17, *)) {
+      // Move auto-correction highlight to overlap with the actual text.
+      // This is to fix an issue where the system auto-correction highlight is displayed at
+      // the top left corner of the screen on iOS 17+.
+      // See https://github.com/flutter/flutter/issues/131695
```
justinmc, line 2472: The autocorrect highlight is also displayed in the top left corner on earlier versions, it's just smaller, right?  This sentence seems misleading, maybe remove it but keep the "See...".  Really this is just positioning the input in the correct place, the place as Flutter's input.

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
```diff
@@ -2449,18 +2449,30 @@ - (void)takeKeyboardScreenshotAndDisplay {
...
+      _inputHider.frame =
+          CGRectMake([transform[leftIndex] intValue], [transform[topIndex] intValue], 0, 0);
```
justinmc, line 2474: What happens if you pass the real width and height here like in the scribble case above instead of 0,0?
`[dictionary[@"width"] intValue], [dictionary[@"height"] intValue]`

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
```diff
@@ -2488,7 +2500,22 @@ - (void)setSelectionRects:(NSArray*)encodedRects {
                                                          ? NSWritingDirectionLeftToRight
                                                          : NSWritingDirectionRightToLeft]];
   }
+
+  BOOL shouldNotifyTextChange = NO;
+  if (@available(iOS 17, *)) {
+    // Force UIKit to query the selectionRects again on iOS 17+
+    // This is to fix a bug on iOS 17+ where UIKit queries the outdated selectionRects after
```
justinmc, line 2507: Similar to above, I think this bug exists on all versions of iOS, right?

justinmc (commented):
Some comments to make sure I understand how this bug affects versions before iOS 17, but looks good overall.

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
hellohuanlin: That's right. How about instead of deleting it, I add a new line saying "This problem also happens on iOS 16, but the size of highlight does not match the text"?

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
hellohuanlin: Not in this case - in iOS 16, we do not use system highlight, so no such bug. 

I also tried to notify the change on iOS 16 and nothing seems to break, but just to be safe here.

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
stuartmorgan: (This is one of the things I was thinking would be part of unifying the these two codepaths in a follow-up, master-only PR.)

stuartmorgan (approved):
LGTM

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
hellohuanlin: added a todo there.

cyanglaz:
I checked out this fix, now the system popup is not showing at all for https://github.com/flutter/flutter/issues/132594

justinmc:
Hmm that's not what I would expect for this change... @cyanglaz You tried it on iOS 17?

cyanglaz:
yes on simulator tho

cyanglaz:
@justinmc I saw that the fix is also cherry-picked. Can you or someone else confirm the issue that I saw and maybe we need to revert this fix and the cp if the issue does exist?

hellohuanlin:
Hmmm strange. When I tried it the system menu shows right next to the text field (the correct position) rather than the top left corner. But it did show up

hellohuanlin:
When i tried it on the release candidate branch (with the fix), the button does show up. But looks like the buttons are tappable. Here's iPhone simulator + iOS 17: 

https://github.com/flutter/engine/assets/41930132/9589af4d-8bb0-4679-b627-3d82bb46fcfb

Hmmm, I may have remembered incorrectly previously. Could be a false alarm. Sorry :(

PR #44784, opened 2023-08-16
Switch some ERROR logs to WARNING logs

johnmccutchan:


jonahwilliams (commented):
You might consider leaving out the log message altogether, if its not directly actionable.

johnmccutchan:
> You might consider leaving out the log message altogether, if its not directly actionable.

It's useful information for someone working on the engine. Is there not a way

jonahwilliams:
In that case an INFO or ERROR seems reasonable, and I would probably update the test to ignore this message

johnmccutchan:
Oops, did not mean to close the pull request.

johnmccutchan:
> In that case an INFO or ERROR seems reasonable, and I would probably update the test to ignore this message

I would do that but I have no idea how to do that nor is there anything printed in the test failure on how one might go about making an exception for some logs.

jonahwilliams (approved):
LGTM

johnmccutchan:
Thanks!

PR #44811, opened 2023-08-17
Adds new builders for partial clang-tidy checks.

zanderso:
This is the first in a sequence of PRs which will lint only changed files on presubmit. The process is the following:
1. Add new `bringup: true` builders that lint only changed files. `bringup: true` builders don't run in presubmit, and since after landing this no files will be detected as changed, these builders will not do anything yet.
1. Remove `bringup: true`. This will allow the new builders to run in presubmit.
1. Ask the team to monitor the new checks on PRs to make sure they are doing the right thing.
1. When we're reasonably sure that the checks are doing the right thing, add `presubmit: false` to `mac_clang_tidy` and `linux_clang_tidy` so that we no longer needlessly lint all files in presubmit.

Related https://github.com/flutter/flutter/issues/105068

### DIFF
--- a/ci/builders/linux_clang_tidy_presubmit.json
+++ b/ci/builders/linux_clang_tidy_presubmit.json
@@ -1,0 +1,95 @@
+{
+    "builds": [
+        {
+            "name": "android_debug_arm64",
+            "archives": [],
+            "drone_dimensions": [
+                "device_type=none",
+                "os=Linux"
+            ],
+            "gn": [
+                "--android",
+                "--android-cpu",
+                "arm64",
+                "--no-lto"
+            ],
+            "ninja": {
+                "config": "android_debug_arm64"
+            }
+        },
+        {
+            "name": "host_debug",
+            "archives": [],
+            "drone_dimensions": [
+                "device_type=none",
+                "os=Linux"
+            ],
+            "gn": [
+                "--runtime-mode",
+                "debug",
+                "--prebuilt-dart-sdk",
+                "--no-lto"
+            ],
+            "ninja": {
+                "config": "host_debug"
+            }
+        }
+    ],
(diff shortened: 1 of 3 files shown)


ci/builders/linux_clang_tidy_presubmit.json
```diff
@@ -0,0 +1,95 @@
...
+                {
+                    "name": "test: lint android_debug_arm64",
+                    "parameters": [
+                        "--variant",
+                        "android_debug_arm64",
+                        "--shard-id=1",
+                        "--shard-variants=host_debug"
+                    ],
+                    "max_attempts": 1,
+                    "script": "flutter/ci/clang_tidy.sh"
```
godofredoc, line 90: Was clang_tidy script modified in a different PR to include the filtering logic?

ci/builders/linux_clang_tidy_presubmit.json
```diff
@@ -0,0 +1,95 @@
...
+                "device_type=none",
+                "os=Linux",
+                "cores=32"
+            ],
+            "gclient_variables": {
+                "download_android_deps": false
+            },
+            "dependencies": [
+                "host_debug",
+                 "android_debug_arm64"
```
godofredoc, line 52: tab vs space?

ci/builders/linux_clang_tidy_presubmit.json
zanderso: It's existed for awhile for local development.

ci/builders/linux_clang_tidy_presubmit.json
zanderso: Fixed.

godofredoc (approved):

PR #44816, opened 2023-08-17
Implement 2 suggested Clang Tidy fixes we don't look for yet.

matanlurey:
I haven't investigated if there are more occurrences or if it's worth enforcing turning the check on generally.

(They were flagged on the Google roll, ironically)

---

[`readability-redundant-smartptr-get`](https://clang.llvm.org/extra/clang-tidy/checks/readability/redundant-smartptr-get.html)

> Find and remove redundant calls to smart pointer’s `.get()` method

[`performance-for-range-copy`](https://clang.llvm.org/extra/clang-tidy/checks/performance/for-range-copy.html)

> Finds C++11 for ranges where the loop variable is copied in each iteration but it would suffice to obtain it by reference.


jonahwilliams (approved):
LGTM

gaaclarke (approved):

PR #44827, opened 2023-08-17
Add callback to Embedder API to respond to new channel listeners, and use for Windows lifecycle

yaakovschectman:
Supplant the temporary solution https://github.com/flutter/engine/pull/44238 with a more elegant solution with am embedder API callback. The windows engine provides a callback that enables graceful exit and app lifecycle when the platform and lifecycle channels are listened to, respectively.

https://github.com/flutter/flutter/issues/131616

## Pre-launch Checklist

- [ ] I signed the [CLA].

### DIFF
--- a/lib/ui/channel_buffers.dart
+++ b/lib/ui/channel_buffers.dart
@@ -380,6 +380,7 @@ class ChannelBuffers {
     assert(!name.contains('\u0000'), 'Channel names must not contain U+0000 NULL characters.');
     final _Channel channel = _channels.putIfAbsent(name, () => _Channel());
     channel.setListener(callback);
+    channelListenedTo(name, true);
   }
 
   /// Clears the listener for the specified channel.
@@ -392,8 +393,14 @@ class ChannelBuffers {
     final _Channel? channel = _channels[name];
     if (channel != null) {
       channel.clearListener();
+      channelListenedTo(name, false);
     }
   }
+
+  @Native<Void Function(Handle, Bool)>(symbol: 'PlatformConfigurationNativeApi::PlatformChannelListenedTo')
+  external static void _channelListenedTo(String name, bool listening);
+
+  void channelListenedTo(String name, bool listening) => _channelListenedTo(name, listening);
 
   /// Deprecated. Migrate to [setListener] instead.
   ///
--- a/lib/ui/window/platform_configuration.h
+++ b/lib/ui/window/platform_configuration.h
@@ -217,6 +217,15 @@ class PlatformConfigurationClient {
   ///
   virtual void RequestDartDeferredLibrary(intptr_t loading_unit_id) = 0;
 
+  //--------------------------------------------------------------------------
+  /// @brief      Invoked when a listener is registered on a platform channel.
+  ///
+  /// @param[in]  name             The name of the platform channel to which a
+  ///                              listener has been registered or cleared.
+  /// 
+  /// @param[in]  listening        Whether the listener has been set (true) or
+  ///                              cleared (false).
+  ///
   virtual void ChannelListenedTo(const std::string& name, bool listening) = 0;
 
  protected:
--- a/lib/ui/window/platform_configuration.h
+++ b/lib/ui/window/platform_configuration.h
@@ -222,7 +222,7 @@ class PlatformConfigurationClient {
   ///
   /// @param[in]  name             The name of the platform channel to which a
   ///                              listener has been registered or cleared.
-  /// 
+  ///
   /// @param[in]  listening        Whether the listener has been set (true) or
   ///                              cleared (false).
   ///
@@ -532,7 +532,8 @@ class PlatformConfigurationNativeApi {
   static void RespondToPlatformMessage(int response_id,
                                        const tonic::DartByteData& data);
 
-  static void PlatformChannelListenedTo(const std::string& name, bool listening);
+  static void PlatformChannelListenedTo(const std::string& name,
+                                        bool listening);
 
   //--------------------------------------------------------------------------
   /// @brief      Requests the Dart VM to adjusts the GC heuristics based on
--- a/lib/ui/window/platform_configuration.h
+++ b/lib/ui/window/platform_configuration.h
@@ -521,7 +521,7 @@ class PlatformConfigurationNativeApi {
   static void RespondToPlatformMessage(int response_id,
                                        const tonic::DartByteData& data);
 
-  static void PlatformChannelListenedTo(const std::string& name);
+  static void PlatformChannelListenedTo(const std::string& name, bool listening);
 
   //--------------------------------------------------------------------------
   /// @brief      Requests the Dart VM to adjusts the GC heuristics based on
--- a/lib/ui/window/platform_configuration.h
+++ b/lib/ui/window/platform_configuration.h
@@ -521,6 +521,8 @@ class PlatformConfigurationNativeApi {
   static void RespondToPlatformMessage(int response_id,
                                        const tonic::DartByteData& data);
 
+  static void PlatformChannelListenedTo(const std::string& name);
+
   //--------------------------------------------------------------------------
   /// @brief      Requests the Dart VM to adjusts the GC heuristics based on
   ///             the requested `performance_mode`. Returns the old performance
--- a/lib/ui/window/platform_configuration.h
+++ b/lib/ui/window/platform_configuration.h
@@ -217,6 +217,8 @@ class PlatformConfigurationClient {
   ///
   virtual void RequestDartDeferredLibrary(intptr_t loading_unit_id) = 0;
 
+  virtual void ChannelListenedTo(const std::string& name, bool listening) = 0;
+
  protected:
   virtual ~PlatformConfigurationClient();
 };
--- a/runtime/runtime_controller.cc
+++ b/runtime/runtime_controller.cc
@@ -388,6 +388,11 @@ RuntimeController::ComputePlatformResolvedLocale(
   return client_.ComputePlatformResolvedLocale(supported_locale_data);
 }
 
+// |PlatformConfigurationClient|
+void RuntimeController::ChannelListenedTo(const std::string& name, bool listening) {
+  client_.ChannelListenedTo(name, listening);
+}
+
 Dart_Port RuntimeController::GetMainPort() {
   std::shared_ptr<DartIsolate> root_isolate = root_isolate_.lock();
   return root_isolate ? root_isolate->main_port() : ILLEGAL_PORT;
--- a/runtime/runtime_controller.cc
+++ b/runtime/runtime_controller.cc
@@ -389,7 +389,8 @@ RuntimeController::ComputePlatformResolvedLocale(
 }
 
 // |PlatformConfigurationClient|
-void RuntimeController::ChannelListenedTo(const std::string& name, bool listening) {
+void RuntimeController::ChannelListenedTo(const std::string& name,
+                                          bool listening) {
   client_.ChannelListenedTo(name, listening);
 }
 
--- a/shell/common/platform_view.h
+++ b/shell/common/platform_view.h
@@ -466,6 +466,9 @@ class PlatformView {
                                CustomAccessibilityActionUpdates actions);
 
   //----------------------------------------------------------------------------
+  virtual void ChannelListenedTo(const std::string& name, bool listening);
+
+  //----------------------------------------------------------------------------
   /// @brief      Used by embedders to specify the updated viewport metrics for
   ///             a view. In response to this call, on the raster thread, the
   ///             rasterizer may need to be reconfigured to the updated viewport
--- a/shell/common/shell.cc
+++ b/shell/common/shell.cc
@@ -1332,6 +1332,11 @@ void Shell::OnEngineHandlePlatformMessage(
   }
 }
 
+void Shell::OnEngineChannelListenedTo(const std::string& name, bool listening) {
+  FML_LOG(ERROR) << "Shell received indication of listening to " << name;
+  // TODO forward to platform view
+}
+
 void Shell::HandleEngineSkiaMessage(std::unique_ptr<PlatformMessage> message) {
   const auto& data = message->data();
 
--- a/shell/common/shell.cc
+++ b/shell/common/shell.cc
@@ -1333,8 +1333,15 @@ void Shell::OnEngineHandlePlatformMessage(
 }
 
 void Shell::OnEngineChannelListenedTo(const std::string& name, bool listening) {
-  FML_LOG(ERROR) << "Shell received indication of listening to " << name;
-  // TODO forward to platform view
+  FML_DCHECK(is_set_up_);
+  FML_DCHECK(task_runners_.GetUITaskRunner()->RunsTasksOnCurrentThread());
+
+  task_runners_.GetPlatformTaskRunner()->PostTask(
+      [view = platform_view_->GetWeakPtr(), name, listening] {
+        if (view) {
+          view->ChannelListenedTo(name, listening);
+        }
+      });
 }
 
 void Shell::HandleEngineSkiaMessage(std::unique_ptr<PlatformMessage> message) {
--- a/shell/platform/embedder/embedder.h
+++ b/shell/platform/embedder/embedder.h
@@ -1327,10 +1327,10 @@ typedef void (*FlutterUpdateSemanticsCallback2)(
     void* /* user data*/);
 
 typedef void (*FlutterChannelListenedToCallback)(
-  const char* data /* name */,
-  size_t data_len /* name.size */,
-  bool /* listening */,
-  void* /* user data */);
+    const char* data /* name */,
+    size_t data_len /* name.size */,
+    bool /* listening */,
+    void* /* user data */);
 
 typedef struct _FlutterTaskRunner* FlutterTaskRunner;
 
--- a/shell/platform/embedder/embedder.h
+++ b/shell/platform/embedder/embedder.h
@@ -1326,6 +1326,12 @@ typedef void (*FlutterUpdateSemanticsCallback2)(
     const FlutterSemanticsUpdate2* /* semantics update */,
     void* /* user data*/);
 
+typedef void (*FlutterChannelListenedToCallback)(
+  const char* data /* name */,
+  size_t data_len /* name.size */,
+  bool /* listening */,
+  void* /* user data */);
+
 typedef struct _FlutterTaskRunner* FlutterTaskRunner;
 
 typedef struct {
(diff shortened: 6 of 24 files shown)


shell/platform/embedder/embedder.h
```diff
@@ -1326,6 +1326,12 @@ typedef void (*FlutterUpdateSemanticsCallback2)(
     const FlutterSemanticsUpdate2* /* semantics update */,
     void* /* user data*/);
 
+typedef void (*FlutterChannelListenedToCallback)(
+    const char* data /* name */,
+    size_t data_len /* name.size */,
+    bool /* listening */,
+    void* /* user data */);
```
loic-sharma, line 1333: I would use the info struct pattern. For example: https://github.com/flutter/engine/blob/83fdadaf12f00ef0ca3aa9401918a2486fcb89a2/shell/platform/embedder/embedder.h#L486-L503

This allows us to add new arguments to the callback if we ever need to in the future.

shell/platform/embedder/embedder.h
```diff
@@ -1326,6 +1326,12 @@ typedef void (*FlutterUpdateSemanticsCallback2)(
     const FlutterSemanticsUpdate2* /* semantics update */,
     void* /* user data*/);
 
+typedef void (*FlutterChannelListenedToCallback)(
+    const char* data /* name */,
+    size_t data_len /* name.size */,
```
loic-sharma, line 1331: Is a length necessary? I'd assume the name is null terminated?

yaakovschectman: It appears that is a guarantee as of C++11, and I forgot about `c_str`.

shell/common/shell.cc
```diff
@@ -1332,6 +1332,18 @@ void Shell::OnEngineHandlePlatformMessage(
   }
 }
 
+void Shell::OnEngineChannelListenedTo(const std::string& name, bool listening) {
+  FML_DCHECK(is_set_up_);
+  FML_DCHECK(task_runners_.GetUITaskRunner()->RunsTasksOnCurrentThread());
+
+  task_runners_.GetPlatformTaskRunner()->PostTask(
```
loic-sharma, line 1339: I'm a bit worried about posting a task to the platform thread that captures a `const std::string &` that was created by Dart FFI.

`PlatformConfigurationNativeApi::SendPlatformMessage` has a similar problem, and I see that it copies the channel's `std::string` ([here](https://github.com/flutter/engine/blob/711ebac54d095717f3448cda5cf73fd99153e4a1/lib/ui/window/platform_configuration.cc#L376) and [here](https://github.com/flutter/engine/blob/83fdadaf12f00ef0ca3aa9401918a2486fcb89a2/lib/ui/window/platform_message.cc#L11))

Should we also copy the channel's `std::string`? Also, should we `std::move` this copied string down instead of passing it by reference?

shell/platform/embedder/embedder.h
```diff
@@ -2118,6 +2124,10 @@ typedef struct {
   /// and `update_semantics_callback2` may be provided; the others must be set
   /// to null.
   FlutterUpdateSemanticsCallback2 update_semantics_callback2;
+
+  /// The callback invoked by the engine in response to a channel listener
+  /// being registered on the framework side.
+  FlutterChannelListenedToCallback channel_listened_to_callback;
```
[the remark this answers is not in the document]
loic-sharma: (If you update this, please make sure to be consistent with the name everywhere)

shell/platform/embedder/embedder.h
```diff
@@ -2194,6 +2203,10 @@ typedef struct {
...
```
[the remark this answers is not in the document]
yaakovschectman: I am currently attempting to implement this. However, it seems that when running one of the embedder's fixtures from `embedder_unittests`, in `Shell:OnEngineChannelListenedTo`, while that method is called, the lambda passed to `PostTask` is not. I have yet to determine why.

lib/ui/channel_buffers.dart
```diff
@@ -380,6 +380,7 @@ class ChannelBuffers {
     assert(!name.contains('\u0000'), 'Channel names must not contain U+0000 NULL characters.');
     final _Channel channel = _channels.putIfAbsent(name, () => _Channel());
     channel.setListener(callback);
+    channelListenedTo(name, true);
```
loic-sharma, line 383: To be consistent with the embedder API, can we rename `channelListenedTo` to `sendChannelUpdate` and `onChannelListenedTo` to `onChannelUpdate` everywhere?

lib/ui/channel_buffers.dart
```diff
@@ -380,7 +380,7 @@ class ChannelBuffers {
     assert(!name.contains('\u0000'), 'Channel names must not contain U+0000 NULL characters.');
     final _Channel channel = _channels.putIfAbsent(name, () => _Channel());
     channel.setListener(callback);
-    channelListenedTo(name, true);
+    sendChannelUpdate(name, true);
```
loic-sharma, line 383: I'm not sure what our style guide is here, but I'd consider making this a named argument for clarity:

```suggestion
    sendChannelUpdate(name, listening: true);
```

lib/ui/window/platform_configuration.h
```diff
@@ -582,8 +582,7 @@ class PlatformConfigurationNativeApi {
   static void RespondToPlatformMessage(int response_id,
                                        const tonic::DartByteData& data);
 
-  static void PlatformChannelListenedTo(const std::string& name,
-                                        bool listening);
+  static void SendChannelUpdate(const std::string& name, bool listening);
```
loic-sharma, line 585: Instead of using `const std::string&`, could we use `std::string` instead and `std::move` it down?

shell/platform/embedder/tests/embedder_unittests.cc
```diff
@@ -2562,6 +2562,36 @@ TEST_F(EmbedderTest, CanSendPointer) {
...
+  auto& context = GetEmbedderContext(EmbedderTestContextType::kSoftwareContext);
+
+  fml::AutoResetWaitableEvent latch;
+  fml::AutoResetWaitableEvent latch2;
+  bool listening = false;
+  context.AddNativeCallback(
+      "SignalNativeTest",
+      CREATE_NATIVE_ENTRY([&](Dart_NativeArguments) { latch.Signal(); }));
+  context.SetChannelUpdateCallback([&](const FlutterChannelUpdate* update) {
+    if (strcmp(update->channel, "test/listen") == 0) {
```
loic-sharma, line 2575: Can this be a `EXPECT_EQ` instead?

Also, could we add a `EXPECT_TRUE` on `update->listening`?

shell/platform/embedder/embedder.h
```diff
@@ -1402,6 +1402,15 @@ typedef void (*FlutterUpdateSemanticsCallback2)(
     const FlutterSemanticsUpdate2* /* semantics update */,
     void* /* user data*/);
 
+typedef struct {
+  const char* channel;
+  bool listening;
+} FlutterChannelUpdate;
```
loic-sharma, line 1408: Please add a quick comment on the struct and each of its members.

shell/platform/windows/flutter_windows_engine.cc
```diff
@@ -819,4 +815,13 @@ std::optional<LRESULT> FlutterWindowsEngine::ProcessExternalWindowMessage(
   return std::nullopt;
 }
 
+void FlutterWindowsEngine::OnChannelUpdate(const std::string& name,
+                                           bool listening) {
+  if (name.compare("flutter/platform") == 0) {
```
loic-sharma, line 820: Can we use `==` instead?

loic-sharma, line 820: Consider checking that `listening` is `true`. This makes this logic a bit more resilient if the engine ever misbehaved and sent updates `listening` of `false` erroneously.

shell/platform/windows/flutter_windows_engine_unittests.cc
```diff
@@ -1111,5 +1113,37 @@ TEST_F(FlutterWindowsEngineTest, LifecycleStateToFrom) {
...
+
+  static bool listened_to = false;
+  modifier.embedder_api().Run = MOCK_ENGINE_PROC(
+      Run, ([old_run](size_t version, const FlutterRendererConfig* config,
+                      const FlutterProjectArgs* args, void* user_data,
+                      FLUTTER_API_SYMBOL(FlutterEngine) * engine_out) {
+        FlutterProjectArgs new_args = *args;
+        new_args.channel_update_callback =
+            [](const FlutterChannelUpdate* update, void* user_data) {
+              listened_to = true;
```
loic-sharma, line 1137: Can we add `EXPECT_X` for the channel name and listening status boolean?

shell/common/shell.cc
```diff
@@ -1328,6 +1328,18 @@ void Shell::OnEngineHandlePlatformMessage(
   }
 }
 
+void Shell::OnEngineChannelUpdate(std::string name, bool listening) {
+  FML_DCHECK(is_set_up_);
+  FML_DCHECK(task_runners_.GetUITaskRunner()->RunsTasksOnCurrentThread());
+
+  task_runners_.GetPlatformTaskRunner()->PostTask(
+      [view = platform_view_->GetWeakPtr(), name = name, listening] {
```
loic-sharma, line 1336: Can we move this too to avoid a copy?

```suggestion
      [view = platform_view_->GetWeakPtr(), name = std::move(name), listening] {
```

shell/platform/embedder/tests/embedder_unittests.cc
```diff
@@ -2562,6 +2562,36 @@ TEST_F(EmbedderTest, CanSendPointer) {
...
+  auto& context = GetEmbedderContext(EmbedderTestContextType::kSoftwareContext);
+
+  fml::AutoResetWaitableEvent latch;
+  fml::AutoResetWaitableEvent latch2;
+  bool listening = false;
+  context.AddNativeCallback(
+      "SignalNativeTest",
+      CREATE_NATIVE_ENTRY([&](Dart_NativeArguments) { latch.Signal(); }));
+  context.SetChannelUpdateCallback([&](const FlutterChannelUpdate* update) {
+    EXPECT_EQ(strcmp(update->channel, "test/listen"), 0);
```
loic-sharma, line 2575: Can this be `EXPECT_STREQ`?

shell/platform/windows/flutter_windows_engine_unittests.cc
```diff
@@ -1111,5 +1113,39 @@ TEST_F(FlutterWindowsEngineTest, LifecycleStateToFrom) {
...
+              EXPECT_EQ(strcmp(update->channel, "flutter/lifecycle"), 0);
```
loic-sharma, line 1138: Can this be `EXPECT_STREQ`?

shell/common/shell.cc
yaakovschectman: Putting `std::move` there produced an error for an unnecessary `std::move` in one of the postsub tests when I tried it.

shell/platform/windows/flutter_windows_engine_unittests.cc
```diff
@@ -1111,5 +1113,39 @@ TEST_F(FlutterWindowsEngineTest, LifecycleStateToFrom) {
...
+  modifier.embedder_api().RunsAOTCompiledDartCode = []() { return false; };
+  auto old_run = modifier.embedder_api().Run;
+
+  static bool listened_to = false;
+  modifier.embedder_api().Run = MOCK_ENGINE_PROC(
+      Run, ([old_run](size_t version, const FlutterRendererConfig* config,
+                      const FlutterProjectArgs* args, void* user_data,
+                      FLUTTER_API_SYMBOL(FlutterEngine) * engine_out) {
+        FlutterProjectArgs new_args = *args;
+        new_args.channel_update_callback =
```
loic-sharma, line 1135: Correct me if I'm wrong, but this test is effectively the same as `EmbedderTest.RegisterChannelListener`? Both check that the embedder's `channel_update_callback` receives the expected values.

Perhaps a better test for this layer would be that `WindowsLifecycleManager::BeginProcessingLifecycle` is called when the framework adds the `flutter/lifecycle` listener. This could be done by using the `EngineModifier` to set a mocked lifecycle manager.

Please let me know if I'm missing something!

shell/common/shell.cc
loic-sharma: Oh ok. I don't understand C++ 😅

shell/platform/windows/flutter_windows_engine_unittests.cc
```diff
@@ -1111,5 +1113,30 @@ TEST_F(FlutterWindowsEngineTest, LifecycleStateToFrom) {
...
+  builder.SetDartEntrypoint("enableLifecycleToFrom");
+
+  auto window_binding_handler =
+      std::make_unique<::testing::NiceMock<MockWindowBindingHandler>>();
+  MockFlutterWindowsView view(std::move(window_binding_handler));
+  view.SetEngine(builder.Build());
+  FlutterWindowsEngine* engine = view.GetEngine();
+  EngineModifier modifier(engine);
+  modifier.embedder_api().RunsAOTCompiledDartCode = []() { return false; };
+  auto old_run = modifier.embedder_api().Run;
```
loic-sharma, line 1127: Can we remove this?

loic-sharma (approved):
Looks great, excellent work! Please get an approval from @cbracken or @chinmaygarde on the changes in `//flutter/shell/platform/embedder/embedder.h` before merging.

shell/platform/embedder/embedder.h
```diff
@@ -1402,6 +1402,18 @@ typedef void (*FlutterUpdateSemanticsCallback2)(
     const FlutterSemanticsUpdate2* /* semantics update */,
     void* /* user data*/);
 
+/// An update to whether a message channel has a listener set or not.
+typedef struct {
+  /// The name of the channel.
+  const char* channel;
+  /// True if a listener has been set, false if one has been cleared.
+  bool listening;
+} FlutterChannelUpdate;
```
cbracken, line 1411: Are we 100% certain that we'll never  want to add more members to this struct? If not, add a `size_t struct_size` member as the first field in the struct and use safe dereference macros when reading from it.

shell/platform/embedder/embedder.h
```diff
@@ -2194,6 +2206,10 @@ typedef struct {
   /// and `update_semantics_callback2` may be provided; the others must be set
   /// to null.
   FlutterUpdateSemanticsCallback2 update_semantics_callback2;
+
+  /// The callback invoked by the engine in response to a channel listener
+  /// being registered on the framework side.
```
cbracken, line 2211: Document which thread this callback is invoked on.

lib/ui/window/platform_configuration.h
```diff
@@ -207,6 +207,17 @@ class PlatformConfigurationClient {
...
+  //--------------------------------------------------------------------------
+  /// @brief      Invoked when a listener is registered on a platform channel.
+  ///
+  /// @param[in]  name             The name of the platform channel to which a
+  ///                              listener has been registered or cleared.
+  ///
+  /// @param[in]  listening        Whether the listener has been set (true) or
+  ///                              cleared (false).
+  ///
+  virtual void SendChannelUpdate(std::string name, bool listening) = 0;
```
cbracken, line 219: This signature doesn't match the implementation. `std::string` vs `const std::string& name`

shell/platform/embedder/embedder.h
loic-sharma: Oh derp I completely forgot this 🤦. Yup let's add a `struct_size` to be safe

runtime/runtime_controller.cc
```diff
@@ -401,6 +401,11 @@ RuntimeController::ComputePlatformResolvedLocale(
   return client_.ComputePlatformResolvedLocale(supported_locale_data);
 }
 
+// |PlatformConfigurationClient|
+void RuntimeController::SendChannelUpdate(std::string name, bool listening) {
```
cbracken, line 405: Similarly - this one also doesn't match.

shell/common/platform_view.h
```diff
@@ -465,6 +465,17 @@ class PlatformView {
   virtual void UpdateSemantics(SemanticsNodeUpdates updates,
                                CustomAccessibilityActionUpdates actions);
 
+  //----------------------------------------------------------------------------
+  /// @brief      Used by the framework to tell the embedder that it has
+  ///             registered a listener on a given channel
```
cbracken, line 470: Super pedantic nit.
```suggestion
  ///             registered a listener on a given channel.
```

shell/platform/embedder/tests/embedder_unittests.cc
```diff
@@ -2562,6 +2562,36 @@ TEST_F(EmbedderTest, CanSendPointer) {
...
+
+  EmbedderConfigBuilder builder(context);
+  builder.SetSoftwareRendererConfig();
+  builder.SetDartEntrypoint("channel_listener_response");
+
+  auto engine = builder.LaunchEngine();
+  ASSERT_TRUE(engine.is_valid());
+
+  latch.Wait();
+  fml::MessageLoop::GetCurrent().RunExpiredTasksNow();
```
cbracken, line 2589: Since we do this relatively rarely in our tests, maybe just for clarity add a comment. It's pretty clear already, and we probably don't comment on this elsewhere, but probably not a bad idea give it was a pain point during development.
```suggestion
  // Drain tasks posted to platform thread task runner.
  fml::MessageLoop::GetCurrent().RunExpiredTasksNow();
```

cbracken (commented):
Overall, really nice work! Just a few small nits!

lib/ui/window/platform_configuration.h
yaakovschectman: What non-matching implementation to this signature are you looking at? It is implemented in `runtime_controller.cc:405`, where the signatures match.

shell/platform/embedder/embedder.h
yaakovschectman: By this, do you mean the `SAFE_ACCESS` macro? Do we currently use this at all in the `FlutterWindowsEngine`? It does not appear so, as far as I can tell.

lib/ui/window/platform_configuration.h
yaakovschectman: Apparently, it seems this may not be an option. I tried that and pushed the change, but ironically, it seems doing so actually results in errors reported by the [clang tidy tests](https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8771483322002799041/+/u/test:_test:_lint_host_debug/stdout) explicitly saying to make the parameters const references. This is only the case for some of the lambdas/methods, but the signature of `PlatformView::SendChannelUpdate` must necessarily match that of `PlatformViewEmbedder::SendChannelUpdate`.

cbracken (approved):

cbracken:
Looks great -- the lint would prefer const ref for the strings, which seems fine to me. I don't think we're mutating them anywhere.

PR #44836, opened 2023-08-18
Remove explicit mac cpu dimensions

godofredoc:
Running builds and tests on x64 machines is causing long queue times even when there is plenty of arm64 machines. This PR is removing most of the cpu pinned dimensions and it is separating some of the test to global tests to allow to select x64 machines only for the tests part.

### DIFF
--- a/ci/builders/mac_android_aot_engine.json
+++ b/ci/builders/mac_android_aot_engine.json
@@ -15,7 +15,7 @@
             "drone_dimensions": [
                 "device_type=none",
                 "os=Mac-12",
-                "cpu=x86"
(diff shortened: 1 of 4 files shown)


zanderso:
Hmm. Just to be sure that the artifacts built on arm64 hosts work, should this PR first hardcode the build configs to arm64 hosts?

godofredoc:
> Hmm. Just to be sure that the artifacts built on arm64 hosts work, should this PR first hardcode the build configs to arm64 hosts?

Sure, let me update it.

zanderso (approved):

PR #44848, opened 2023-08-18
[macOS,iOS] Expose channel buffers 'resize' and 'overflow' control co…

bleroux:
## Description

This PR update the helper function that invokes the control channel 'resize' command (previous implementation relied on a deprecated format). It also adds a similar helper function for the 'overflow' commands exposed by the control channel.

See:

## Related Issue

iOS and macOS implementation for https://github.com/flutter/flutter/issues/132386

Similar implementations:
- Android: https://github.com/flutter/engine/pull/44434
- Linux: https://github.com/flutter/engine/pull/44636

## Tests

Adds two tests.

### DIFF
--- a/shell/platform/darwin/common/framework/Headers/FlutterChannels.h
+++ b/shell/platform/darwin/common/framework/Headers/FlutterChannels.h
@@ -143,8 +143,45 @@ FLUTTER_DARWIN_EXPORT
  * Adjusts the number of messages that will get buffered when sending messages to
  * channels that aren't fully set up yet.  For example, the engine isn't running
  * yet or the channel's message handler isn't set up on the Dart side yet.
+ *
+ * @param name The channel name.
+ * @param messenger The binary messenger.
+ * @param newSize The new size of the buffer.
+ */
++ (void)resizeChannelWithName:(NSString*)name
+              binaryMessenger:(NSObject<FlutterBinaryMessenger>*)messenger
+                      newSize:(NSInteger)newSize;
+
+/**
+ * Adjusts the number of messages that will get buffered when sending messages to
+ * channels that aren't fully set up yet.  For example, the engine isn't running
+ * yet or the channel's message handler isn't set up on the Dart side yet.
+ *
+ * @param newSize The new size of the buffer.
  */
 - (void)resizeChannelBuffer:(NSInteger)newSize;
+
+/**
+ * Toggles whether the channel should show warning messages when discarding messages
+ * due to overflow.
+ *
+ * @param name The channel name.
+ * @param messenger The binary messenger.
+ * @param allowed When true, the channel is expected to overflow and warning messages
+ *                will not be shown.
+ */
++ (void)setAllowOverflowChannelWithName:(NSString*)name
+                        binaryMessenger:(NSObject<FlutterBinaryMessenger>*)messenger
+                                allowed:(BOOL)allowed;
+
+/**
+ * Toggles whether the channel should show warning messages when discarding messages
+ * due to overflow.
+ *
+ * @param allowed When true, the channel is expected to overflow and warning messages
+ *                will not be shown.
+ */
+- (void)setAllowOverflow:(BOOL)allowed;
 
 @end
 
--- a/shell/platform/darwin/common/framework/Headers/FlutterChannels.h
+++ b/shell/platform/darwin/common/framework/Headers/FlutterChannels.h
@@ -146,42 +146,42 @@ FLUTTER_DARWIN_EXPORT
  *
  * @param name The channel name.
  * @param messenger The binary messenger.
- * @param newSize The new size of the buffer.
+ * @param newSize The number of messages that will get buffered.
  */
 + (void)resizeChannelWithName:(NSString*)name
               binaryMessenger:(NSObject<FlutterBinaryMessenger>*)messenger
-                      newSize:(NSInteger)newSize;
+                         size:(NSInteger)newSize;
 
 /**
  * Adjusts the number of messages that will get buffered when sending messages to
  * channels that aren't fully set up yet.  For example, the engine isn't running
  * yet or the channel's message handler isn't set up on the Dart side yet.
  *
- * @param newSize The new size of the buffer.
+ * @param newSize The number of messages that will get buffered.
  */
 - (void)resizeChannelBuffer:(NSInteger)newSize;
 
 /**
- * Toggles whether the channel should show warning messages when discarding messages
+ * Defines whether the channel should show warning messages when discarding messages
  * due to overflow.
  *
+ * @param allowed When true, the channel is expected to overflow and warning messages
+ *                will not be shown.
  * @param name The channel name.
  * @param messenger The binary messenger.
- * @param allowed When true, the channel is expected to overflow and warning messages
- *                will not be shown.
  */
-+ (void)setAllowOverflowChannelWithName:(NSString*)name
-                        binaryMessenger:(NSObject<FlutterBinaryMessenger>*)messenger
-                                allowed:(BOOL)allowed;
++ (void)setWarnsOnOverflow:(BOOL)allowed
+        forChannelWithName:(NSString*)name
+           binaryMessenger:(NSObject<FlutterBinaryMessenger>*)messenger;
 
 /**
- * Toggles whether the channel should show warning messages when discarding messages
+ * Defines whether the channel should show warning messages when discarding messages
  * due to overflow.
  *
  * @param allowed When true, the channel is expected to overflow and warning messages
  *                will not be shown.
  */
-- (void)setAllowOverflow:(BOOL)allowed;
+- (void)setWarnsOnOverflow:(BOOL)allowed;
 
 @end
 
--- a/shell/platform/darwin/common/framework/Headers/FlutterChannels.h
+++ b/shell/platform/darwin/common/framework/Headers/FlutterChannels.h
@@ -165,12 +165,12 @@ FLUTTER_DARWIN_EXPORT
  * Defines whether the channel should show warning messages when discarding messages
  * due to overflow.
  *
- * @param allowed When true, the channel is expected to overflow and warning messages
- *                will not be shown.
+ * @param warns When false, the channel is expected to overflow and warning messages
+ *              will not be shown.
  * @param name The channel name.
  * @param messenger The binary messenger.
  */
-+ (void)setWarnsOnOverflow:(BOOL)allowed
++ (void)setWarnsOnOverflow:(BOOL)warns
         forChannelWithName:(NSString*)name
            binaryMessenger:(NSObject<FlutterBinaryMessenger>*)messenger;
 
@@ -178,10 +178,10 @@ FLUTTER_DARWIN_EXPORT
  * Defines whether the channel should show warning messages when discarding messages
  * due to overflow.
  *
- * @param allowed When true, the channel is expected to overflow and warning messages
- *                will not be shown.
+ * @param warns When false, the channel is expected to overflow and warning messages
+ *              will not be shown.
  */
-- (void)setWarnsOnOverflow:(BOOL)allowed;
+- (void)setWarnsOnOverflow:(BOOL)warns;
 
 @end
 
--- a/shell/platform/darwin/common/framework/Source/FlutterChannels.mm
+++ b/shell/platform/darwin/common/framework/Source/FlutterChannels.mm
@@ -15,7 +15,8 @@ static NSString* const kOverflowMethod = @"overflow";
 static void ResizeChannelBuffer(NSObject<FlutterBinaryMessenger>* binaryMessenger,
                                 NSString* channel,
                                 NSInteger newSize) {
-  NSArray* args = @[ channel, [NSNumber numberWithInt:newSize] ];
+  NSCAssert(newSize >= 0, @"Channel buffer size must be non-negative");
+  NSArray* args = @[ channel, @(newSize) ];
   FlutterMethodCall* resizeMethodCall = [FlutterMethodCall methodCallWithMethodName:kResizeMethod
                                                                           arguments:args];
   NSObject<FlutterMethodCodec>* codec = [FlutterStandardMethodCodec sharedInstance];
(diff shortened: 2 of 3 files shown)


shell/platform/darwin/common/framework/Source/FlutterChannels.mm
```diff
@@ -9,12 +9,28 @@
...
 static NSString* const kFlutterChannelBuffersChannel = @"dev.flutter/channel-buffers";
+static NSString* const kResizeMethod = @"resize";
+static NSString* const kOverflowMethod = @"overflow";
 
 static void ResizeChannelBuffer(NSObject<FlutterBinaryMessenger>* binaryMessenger,
                                 NSString* channel,
                                 NSInteger newSize) {
-  NSString* messageString = [NSString stringWithFormat:@"resize\r%@\r%@", channel, @(newSize)];
-  NSData* message = [messageString dataUsingEncoding:NSUTF8StringEncoding];
+  NSArray* args = @[ channel, [NSNumber numberWithInt:newSize] ];
```
cbracken, line 18: Given that NSInteger supports negative values, we should
```cc
NSAssert(newSize >= 0, @"Channel buffer size must be non-negative");
```
at the top of this method to help catch bugs.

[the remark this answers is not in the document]
stuartmorgan: IIRC the Google style guide used to mention it; most likely it was culled under https://google.github.io/styleguide/objcguide.html#style-rules-should-pull-their-weight because there's no reason *not* to use the modern literal syntax.

Boxing with NSNumber should definitely use modern boxing syntax.

shell/platform/darwin/common/framework/Headers/FlutterChannels.h
```diff
@@ -143,9 +143,46 @@ FLUTTER_DARWIN_EXPORT
...
  * channels that aren't fully set up yet.  For example, the engine isn't running
  * yet or the channel's message handler isn't set up on the Dart side yet.
+ *
+ * @param name The channel name.
+ * @param messenger The binary messenger.
+ * @param newSize The new size of the buffer.
+ */
++ (void)resizeChannelWithName:(NSString*)name
+              binaryMessenger:(NSObject<FlutterBinaryMessenger>*)messenger
+                      newSize:(NSInteger)newSize;
```
stuartmorgan, line 153: Calling the variable `newSize` is fine, but the signature should just be `size:`; the value passed to a setter is always the new value.

shell/platform/darwin/common/framework/Headers/FlutterChannels.h
```diff
@@ -143,9 +143,46 @@ FLUTTER_DARWIN_EXPORT
  * Adjusts the number of messages that will get buffered when sending messages to
  * channels that aren't fully set up yet.  For example, the engine isn't running
  * yet or the channel's message handler isn't set up on the Dart side yet.
+ *
+ * @param name The channel name.
+ * @param messenger The binary messenger.
+ * @param newSize The new size of the buffer.
```
stuartmorgan, line 149: "size" needs a clear description here. Bytes? kb? Messages?

Even better would be a parameter name that actually captures that information as well.

shell/platform/darwin/common/framework/Headers/FlutterChannels.h
```diff
@@ -143,9 +143,46 @@ FLUTTER_DARWIN_EXPORT
...
++ (void)resizeChannelWithName:(NSString*)name
+              binaryMessenger:(NSObject<FlutterBinaryMessenger>*)messenger
+                      newSize:(NSInteger)newSize;
+
+/**
+ * Adjusts the number of messages that will get buffered when sending messages to
+ * channels that aren't fully set up yet.  For example, the engine isn't running
+ * yet or the channel's message handler isn't set up on the Dart side yet.
+ *
+ * @param newSize The new size of the buffer.
```
stuartmorgan, line 160: Same here.

shell/platform/darwin/common/framework/Headers/FlutterChannels.h
```diff
@@ -143,9 +143,46 @@ FLUTTER_DARWIN_EXPORT
...
  */
 - (void)resizeChannelBuffer:(NSInteger)newSize;
```
stuartmorgan, line 162: This is really unfortunate naming, but I guess we're stuck with it :(

shell/platform/darwin/common/framework/Headers/FlutterChannels.h
```diff
@@ -143,9 +143,46 @@ FLUTTER_DARWIN_EXPORT
...
 
+/**
+ * Toggles whether the channel should show warning messages when discarding messages
```
stuartmorgan, line 165: If all this controls is warnings, the naming of the method seems very confusing. I would expect allowing overflow to mean it actually changes the behavior of handling extra messages.

shell/platform/darwin/common/framework/Headers/FlutterChannels.h
```diff
@@ -143,9 +143,46 @@ FLUTTER_DARWIN_EXPORT
...
+ * due to overflow.
+ *
+ * @param name The channel name.
+ * @param messenger The binary messenger.
+ * @param allowed When true, the channel is expected to overflow and warning messages
+ *                will not be shown.
+ */
++ (void)setAllowOverflowChannelWithName:(NSString*)name
```
stuartmorgan, line 173: Repetition of parameter names should be avoided in signatures when possible; if we were to keep this naming I would call it `setOverflowAllowed:forChannelWithName:binaryMessenger:`. Combined with the comment above I would call it something more like `setWarnsOnOverflow:forChannelWithName:binaryMessenger:`.

shell/platform/darwin/common/framework/Headers/FlutterChannels.h
stuartmorgan, line 165: "Toggle" means that it flips the state to the opposite, which isn't what this does; it sets it to a specific value. A toggle method wouldn't take a bool at all.

shell/platform/darwin/common/framework/Headers/FlutterChannels.h
```diff
@@ -143,9 +143,46 @@ FLUTTER_DARWIN_EXPORT
...
+                                allowed:(BOOL)allowed;
+
+/**
+ * Toggles whether the channel should show warning messages when discarding messages
+ * due to overflow.
+ *
+ * @param allowed When true, the channel is expected to overflow and warning messages
+ *                will not be shown.
+ */
+- (void)setAllowOverflow:(BOOL)allowed;
```
stuartmorgan, line 184: All the same notes here.

shell/platform/darwin/common/framework/Headers/FlutterChannels.h
```diff
@@ -143,9 +143,46 @@ FLUTTER_DARWIN_EXPORT
...
+
```
stuartmorgan, line 185: If the Obj-C API has already set a precedent of putting these controls on the channel, what's the reason for duplicating them as static methods? Can't clients just use the channel versions?

shell/platform/darwin/common/framework/Source/FlutterChannels.mm
```diff
@@ -9,12 +9,28 @@
...
+  FlutterMethodCall* resizeMethodCall = [FlutterMethodCall methodCallWithMethodName:kResizeMethod
+                                                                          arguments:args];
+  NSObject<FlutterMethodCodec>* codec = [FlutterStandardMethodCodec sharedInstance];
+  NSData* message = [codec encodeMethodCall:resizeMethodCall];
+  [binaryMessenger sendOnChannel:kFlutterChannelBuffersChannel message:message];
+}
+
+static void SetAllowChannelOverflow(NSObject<FlutterBinaryMessenger>* binaryMessenger,
```
stuartmorgan, line 26: All method and function declarations need declaration comments, per the style guide.

stuartmorgan (changes requested):
(Wrong button, sorry; that was meant to be a "request changes")

shell/platform/darwin/common/framework/Headers/FlutterChannels.h
bleroux: I stuck to the naming from channels_buffer.dart. 
https://github.com/flutter/engine/blob/93e8901490e78c7ba7e319cce4470d9c6478c6dc/lib/ui/channel_buffers.dart#L574

I agree that it is confusing. Is it ok to use a different naming? If yes, I would switch to `expectOverflow` or `warnOnOverflow`.

bleroux: I copied/pasted the comment from channel_buffers.dart
https://github.com/flutter/engine/blob/93e8901490e78c7ba7e319cce4470d9c6478c6dc/lib/ui/channel_buffers.dart#L562-L564
I will try to rephrase it.

shell/platform/darwin/common/framework/Headers/FlutterChannels.h
bleroux: When trying to use the existing API on a real use case (configuring `flutter/keydata` buffer), I realized that it was somewhat confusing to have to instanciate a method channel (which requires specifying a codec). I think for use cases where the creation of the channel and the configuration of  its buffer are separated will benefit from using the static methods.

shell/platform/darwin/common/framework/Source/FlutterChannels.mm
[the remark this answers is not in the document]
bleroux: When switching to the modern boxing syntax, for this particular code, it does not produce the same result at runtime.
As you can see in the change I pushed, the test is failing because the serialization result is not the same (and the result is wrong with the modern boxing syntax). I don't have enough ObjectiveC knowledge to understand why.

Test failure: https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8770883490087443713/+/u/test:_Tests_for_ios_debug_sim_arm64_extension_safe__3_/stdout

Failure message: 
```console
error: -[FlutterChannelsTest testResize] : OCProtocolMockObject(FlutterBinaryMessenger): unexpected method invoked: sendOnChannel:@"dev.flutter/channel-buffers" message:{length = 33, bytes = 0x07067265 73697a65 0c02070c 666c7574 ... 03000000 00000000 } 
expected:	sendOnChannel:@"dev.flutter/channel-buffers" message:{length = 29, bytes = 0x07067265 73697a65 0c02070c 666c7574 ... 65737403 03000000 } (NSInternalInconsistencyException)
```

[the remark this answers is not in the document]
stuartmorgan: The failing test says that the expected value was created with Dart code; why? It's not clear to me why you are asserting a specific representation here that comes from a different implementation with different integer semantics.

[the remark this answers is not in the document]
bleroux: The encoded message is deserialized in the Dart side, the deserialization code is there:

https://github.com/flutter/engine/blob/93e8901490e78c7ba7e319cce4470d9c6478c6dc/lib/ui/channel_buffers.dart#L459-L537

The test checks that the message that will be send to Dart as an array of bytes is exactly the expected one.
With the old boxing syntax (`NSArray* args = @[ channel, [NSNumber numberWithInt:newSize] ];`), calling `encodeMethodCall` returns the exact byte array. 

The problem is not specific to the test, when running a Flutter app with a modified engine that calls the resize API, the Dart side throws the following exception:
```console
[ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: Exception: Invalid arguments for 'resize' method sent to dev.flutter/channel-buffers (second argument must be an integer in the range 0 to 2147483647)
#0      ChannelBuffers.handleMessage (dart:ui/channel_buffers.dart:496:13)
#1      PlatformDispatcher._dispatchPlatformMessage (dart:ui/platform_dispatcher.dart:674:24)
#2      _dispatchPlatformMessage (dart:ui/hooks.dart:257:31)
```

Based on the Dart exception, I figured out that the modern boxing syntax is treating the NSInteger as a long on 64bits platforms. I updated the PR with a cast to int and the encoded message is ok now.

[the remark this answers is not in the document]
bleroux: I have updated the PR with a comment on the cast.

shell/platform/darwin/common/framework/Headers/FlutterChannels.h
```diff
@@ -143,9 +143,46 @@ FLUTTER_DARWIN_EXPORT
...
+/**
+ * Defines whether the channel should show warning messages when discarding messages
+ * due to overflow.
+ *
+ * @param allowed When true, the channel is expected to overflow and warning messages
+ *                will not be shown.
+ * @param name The channel name.
+ * @param messenger The binary messenger.
+ */
++ (void)setWarnsOnOverflow:(BOOL)allowed
```
stuartmorgan, line 173: Nit: `(BOOL)warns`

shell/platform/darwin/common/framework/Headers/FlutterChannels.h
```diff
@@ -143,9 +143,46 @@ FLUTTER_DARWIN_EXPORT
...
+           binaryMessenger:(NSObject<FlutterBinaryMessenger>*)messenger;
+
+/**
+ * Defines whether the channel should show warning messages when discarding messages
+ * due to overflow.
+ *
+ * @param allowed When true, the channel is expected to overflow and warning messages
+ *                will not be shown.
+ */
+- (void)setWarnsOnOverflow:(BOOL)allowed;
```
stuartmorgan, line 184: `(BOOL)warns`

shell/platform/darwin/common/framework/Source/FlutterChannels.mm
```diff
@@ -9,12 +9,41 @@
...
 static void ResizeChannelBuffer(NSObject<FlutterBinaryMessenger>* binaryMessenger,
                                 NSString* channel,
                                 NSInteger newSize) {
-  NSString* messageString = [NSString stringWithFormat:@"resize\r%@\r%@", channel, @(newSize)];
-  NSData* message = [messageString dataUsingEncoding:NSUTF8StringEncoding];
+  NSCAssert(newSize >= 0, @"Channel buffer size must be non-negative");
+  // Cast newSize to int because the deserialization logic handles only 32 bits values,
+  // see
+  // https://github.com/flutter/engine/blob/93e8901490e78c7ba7e319cce4470d9c6478c6dc/lib/ui/channel_buffers.dart#L495.
+  NSArray* args = @[ channel, @((int)newSize) ];
```
stuartmorgan, line 22: `@(static_cast<int>(newSize))`

(We should never use C-style casts in a `.mm` file as they are strictly worse than C++-style casts.)

shell/platform/darwin/common/framework/Source/FlutterChannels.mm
```diff
@@ -9,12 +9,41 @@
...
+ * due to overflow.
+ *
+ * @param binaryMessenger The binary messenger.
+ * @param channel The channel name.
+ * @param allowed When true, the channel is expected to overflow and warning messages
+ *                will not be shown.
+ */
+static void SetWarnsOnOverflow(NSObject<FlutterBinaryMessenger>* binaryMessenger,
+                               NSString* channel,
+                               BOOL allowed) {
```
stuartmorgan, line 41: `warns`

shell/platform/darwin/common/framework/Source/FlutterChannels.mm
```diff
@@ -9,12 +9,41 @@
...
+  NSArray* args = @[ channel, @(allowed) ];
```
stuartmorgan, line 42: Why not just inline this into the call?

stuartmorgan (approved):
LGTM with nits

PR #44855, opened 2023-08-18
Change Doxyfile options to not create subdirs

gspencergoog:
## Description

This turns off the Doxyfile option `CREATE_SUBDIRS` because it prevents deep link URLs from being stable. It means that thousands of files will be in the main directory, but the filesystem should be able to handle that.

I got rid of the header/footer files for doxygen (because Doxygen will generate them anyhow) and fixed where it was looking for the Flutter logo.

I also ran Doxygen on the Doxyfile template and Doxyfile and updated all the options to correspond with the latest Doxygen.


christopherfujino (approved):
LGTM

PR #44861, opened 2023-08-18
[Impeller] Started recycling vkRenderPass and vkFramebuffer

gaaclarke:

## Pre-launch Checklist

- [ ] I signed the [CLA].


jonahwilliams (commented):
You can't recycle the objects like this, you need to use the resource manager and wait for the fence waiter to be signal

gaaclarke:
> You can't recycle the objects like this, you need to use the resource manager and wait for the fence waiter to be signal

It's getting tracked in the tracked resources so it won't be reused until the RingBufferItem is deleted (when the fence waiter signals).  There is a bug in it right now which i'm fixing.

impeller/renderer/backend/vulkan/render_pass_vk.cc
```diff
@@ -592,27 +592,42 @@ bool RenderPassVK::OnEncodeCommands(const Context& context) const {
...
+    RenderPassFrameBufferPair pair = RenderPassFrameBufferPair{
+        .render_pass = render_pass,
+        .framebuffer = framebuffer,
+    };
+
+    render_pass_frame_buffer_pair.GetValue() = std::move(pair);
   }
 
-  if (!encoder->Track(framebuffer) || !encoder->Track(render_pass)) {
+  if (!encoder->Track(
```
gaaclarke, line 619: @jonahwilliams right here we are making sure it isn't placed back into the pool until the frame is done.

jonahwilliams (commented):
Ahh I see!

gaaclarke:
This isn't quite right yet.  I hadn't noticed that the vkrenderpass and vkframebuffer creation depends on the render target.  So we will need something like a ring buffer per render target (while making sure we clean up old render targets).

gaaclarke:
Closing this.  After further investigation we've found this wasn't the culprit.

jonahwilliams:
There are multiple culprits, here is a run where despite not being the slowest operation, creating the renderpass still took 3ms:

image

gaaclarke:
Current state: Disabling the cache has no validation errors, turning it on leads to validation errors surrounding the resolve texture:
```
```

gaaclarke:
It tried:
1) setting the layout transition for the resolve texture from coloroptimal->present to undefined->present and that gets us past the first frame, but eventually leads to problems.
2) Introducing a barrier that transitions current->general->current for the resolve texture, like we do for the other color attachments.

My theory is that the synchronization that we are doing for the other textures has to happen for the resolve texture.  But what we are doing for the other textures won't work for the resolve texture since it is swapping between coloroptimal->present so we can't knowingly swap it from general back to whatever.  We need to find a new way.  Given that theory though, I'm at a lose to explain why it currently works.

gaaclarke:
Here is the output now with the new logs: https://gist.github.com/gaaclarke/437c0753782601d0f78300977b7f2c46

It seems like the initialLayout for the resolve texture should be color optimal, or there should be a barrier to make it happen.

jonahwilliams:
Okay I feel dumb, the problem seems to be that the framebuffer objects have references to the actual attachments. So if you recycle them and the attachments aren't the same its just straight up invalid - meaning we always have to recreate them. IN theory the RenderPass is only referencing the attachments in the framebuffer by index, so that may work. Still looking

so basically you can only reuse the framebuffer if you have a static layout with a fixed number of objects per frame

gaaclarke:
My thought process was that render passes should just have a layout which considering the rendertargets are equal, should be the same.  Based on the trace statements we can see that the rendertargets are the same so I didn't implement that part of the caching yet.  I had considered framebufferrs will be harder to recycle since they may have links to swapchain images.  But since the errors I was seeing were related to resolve texture image layouts, which is the province of the renderpass and pipeline barriers I didn't spend much time looking into the framebuffer.

However, allocating the framebuffer is the one that I measured as being slow so if we can't easily recycle that, the value of this PR decreases.

gaaclarke:
Closing this PR to clear out the triage process.  I requested that this policy be reviewed since my hope would be that drafts are not reviewed.

PR #44864, opened 2023-08-18
Allow optional codepoints to be expressed to the font subset generator.

eyebrowsoffire:
This is the engine side change to fix https://github.com/flutter/flutter/issues/132711. There will be a subsequent framework change to express the space character as an "optional" character.


Hixie:
test-exempt: is tested

bonus points for teaching the bot about this form of testing

jason-simmons (approved):

PR #44865, opened 2023-08-18
FlutterFragment predictive back

justinmc:
Following up on https://github.com/flutter/engine/pull/39208 and https://github.com/flutter/engine/pull/42789, which added predictive back support for FlutterActivity.

My test app for this is: https://github.com/justinmc/flutter-add-to-app

I tried this both with the Flutter app as the main activity, and with it as a secondary activity after an initial Android route. Predictive back worked in both cases.

Related: https://github.com/flutter/flutter/issues/109558, and my comment https://github.com/flutter/flutter/issues/109558#issuecomment-1683099477.

b/242216228

### DIFF
--- a/shell/platform/android/io/flutter/embedding/android/FlutterFragmentActivity.java
+++ b/shell/platform/android/io/flutter/embedding/android/FlutterFragmentActivity.java
@@ -521,6 +521,7 @@ public class FlutterFragmentActivity extends FragmentActivity
             ? TransparencyMode.opaque
             : TransparencyMode.transparent;
     final boolean shouldDelayFirstAndroidViewDraw = renderMode == RenderMode.surface;
+    final boolean shouldAutomaticallyHandleOnBackPressed = Build.VERSION.SDK_INT >= 33;
 
     if (getCachedEngineId() != null) {
       Log.v(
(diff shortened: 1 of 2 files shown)


shell/platform/android/io/flutter/embedding/android/FlutterFragmentActivity.java
```diff
@@ -521,6 +521,7 @@ protected FlutterFragment createFlutterFragment() {
             ? TransparencyMode.opaque
             : TransparencyMode.transparent;
     final boolean shouldDelayFirstAndroidViewDraw = renderMode == RenderMode.surface;
+    final boolean shouldAutomaticallyHandleOnBackPressed = Build.VERSION.SDK_INT >= 33;
```
math1man, line 524: With this change, you'll need to update the implementation of `onBackPressed()` below. On Android 33, the `onBackPressed()` method is still relevant, but it won't behave correctly without a call to `super.onBackPressed()`. 

Alternatively, you could make this condition always true and fully remove the `onBackPressed()` override below. That way you have less conditional behavior.

math1man: I suppose removing the `onBackPressed()` override could impact behavior for any clients that subclass `FlutterFragmentActivity` and themselves override the method, but given that Android now formally discourages that anyways, that seems like a reasonable breaking change.

goderbauer:
I spoke to @justinmc last week and he told me that he will get back to this one soon.

justinmc:
I'll re-evaluate this after https://github.com/flutter/engine/pull/49093 is merged.

justinmc:
Spiritual successor: https://github.com/flutter/engine/pull/52302

PR #44881, opened 2023-08-20
[Impeller] Sync presentation when rendering into FlutterImageView.

jonahwilliams:
Fixes https://github.com/flutter/flutter/issues/131730

When the Android embedder starts rendering into a FlutterImageView, notify the Impeller context to block on submitKHR.


jonahwilliams:
This isn't really the  "right" way to do this, but I'm not sure how much effort we should invest in the right way.

I think the correct way would be something like having the FlutterEngine render into an external memory hardware buffer?

shell/platform/android/platform_view_android.h
```diff
@@ -120,6 +120,12 @@ class PlatformViewAndroid final : public PlatformView {
     return platform_message_handler_;
   }
 
+  void SetIsRenderingToImageView(bool value) {
```
johnmccutchan, line 123: If it is possible (I think it is) to have more than one FlutterImageViews active concurrently this should probably keep track of an integer count and only update the context when it hits 0 or != 0

jonahwilliams: I think I could get that to happen if I overlay something on the platform view. Will try that out and see if I can repro.

jonahwilliams: I played aroud with this a bit and I wasn't able to get this to happen in a simple example with an overlay over a platform view. In that case it looks like the overlay view is added as a child of the FlutterView and so their lifecycle is tied.

Perhaps this could happen in an add2app scenario with multiple flutter views sharing the same engine? Extermely cursed to think about those having platform views in them - but I'll make it a count anyway since that should be perfectly safe (TM).

jonahwilliams:
I moved the state tracking into the FlutterImageView as there seemed to be fewer potential entrypoints to worry about. Something odd about this style of platform view is that we seem to immediately switch to imageview rendering when the platform view is added to the framework layer tree, even if it is offscreen.

jonahwilliams:
Which is to say this change unfortunately regresses the performance of the entire wonderous editorial screen.

shell/platform/android/platform_view_android.h
johnmccutchan: Thanks!

johnmccutchan (approved):

PR #44891, opened 2023-08-21
Add `Image.approximateBytesUsed`

fzyzcjy:
Suggested by @dnfield in https://github.com/flutter/flutter/pull/129054#pullrequestreview-1514854917. Please follow that issue for why this is needed.

Partially handle https://github.com/flutter/flutter/issues/128163

I will update the framework side to use this API after this PR is merged.

Given that there are no unit tests for image.cc (neither for picture.cc which I mimic), maybe I should utilize the tests in the framework layer instead?

## Pre-launch Checklist

- [x] I signed the [CLA].


lib/ui/painting.dart
```diff
@@ -1839,6 +1839,15 @@ class Image {
     return stacks;
   }
 
+  /// Returns the approximate byte size of the allocation of this image.
+  ///
+  /// This takes into account details such as mip-mapping. The allocation
+  /// is usually resident in the device memory.
```
dnfield, line 1845: We should document this as being accurate for the storage format and mipmapping of the image, but not document where the allocation is resident - for most images, they're actually CPU resident.

lib/web_ui/lib/painting.dart
```diff
@@ -344,6 +344,7 @@ abstract class Image {
   Future<ByteData?> toByteData({ImageByteFormat format = ImageByteFormat.rawRgba});
   void dispose();
   bool get debugDisposed;
+  int get approximateBytesUsed => width * height * 4;
```
dnfield, line 347: We should leave this as unimplemented here.

lib/web_ui/lib/src/engine/canvaskit/image.dart
```diff
@@ -350,6 +350,9 @@ class CkImage implements ui.Image, StackTraceDebugger {
     }
   }
 
+  @override
+  int get approximateBytesUsed => width * height * 4;
```
dnfield, line 354: We should ask CanvasKit what the size is, it knows a much better value than this.

lib/web_ui/lib/src/engine/skwasm/skwasm_impl/image.dart
```diff
@@ -74,6 +74,9 @@ class SkwasmImage extends SkwasmObjectWrapper<RawImage> implements ui.Image {
   @override
   List<StackTrace>? debugGetOpenHandleStackTraces() => null;
 
+  @override
+  int get approximateBytesUsed => width * height * 4;
```
dnfield, line 78: @eyebrowsoffire - we should make a call into Skia for this value but I'm not sure how that works in the skwasm backend.

eyebrowsoffire: We'll have to pipe a binding through `raw_image.dart` and `image.cpp` to get the value from skia in the skwasm renderer. This should look very similar to the `approximateBytesUsed` call in `picture.dart`/`raw_picture.dart`/`picture.cpp` we have already implemented in the skwasm renderer.

lib/web_ui/lib/src/engine/canvaskit/image.dart
eyebrowsoffire: It does not appear CanvasKit actually exposes this on `Picture` or `Image`, so this would require a separate change to the skia repo to add this (right now the `Picture` implementation for CanvasKit just returns 0 for `approximateBytesUsed`)

lib/ui/painting.dart
fzyzcjy: done

fzyzcjy:
I see, I will do that later, thank you!

lib/ui/painting.dart
fzyzcjy: Btw this looks interesting: DlImage says "Represents an image whose allocation is (usually) resident on device memory", while the CanvasImage (as a simple wrapper of DlImage) is on CPU.

lib/ui/painting.dart
```diff
@@ -1839,6 +1839,14 @@ class Image {
     return stacks;
   }
 
+  /// Returns the approximate byte size of the allocation of this image.
+  ///
+  /// This is accurate for the storage format and mipmapping of the image.
+  int get approximateBytesUsed {
```
jonahwilliams, line 1845: I don't believe dlimage actually has storage format correct size estimations. We'd need to know about lossy or lossless texture compression as well.

jonahwilliams (changes requested):
I would prefer that we didn't add this at all. I struggle to think of scenarios where this information is useful, and there are too many caveats to image sizes (lossy/lossless compression) as well as storage (host visible/device local). All impeller images are device local for example.

The simple estimation the framework provides is 90% of the value for no ongoing cost to engine complexity.

lib/ui/painting.dart
jonahwilliams: The current impeller implemenation is here: https://github.com/flutter/engine/blob/main/impeller/display_list/dl_image_impeller.cc#L88

While it correctly includes the format, it does not include mips. These should be added.

Additionally we should be able to check if texture compression is enabled. and include the compression ratios. approximately 0.5 for iOS backends.

fzyzcjy:
> @jonahwilliams I struggle to think of scenarios where this information is useful

So the history is like:

1. https://github.com/flutter/flutter/issues/128163 - this definitely needs to be fixed, because no end user will expect two functions getting image size will have a 1/3 error
2. https://github.com/flutter/flutter/pull/129054 - my first attempt to fix it (seems like your suggestion of simple estimation), then @dnfield suggests me to create this PR touching the engine
3. this PR - touch the engine

So... shall I just simply reopen the old PR?

jonahwilliams:
My opinion is that if we're going to go through the trouble of exposing this through the engine, then the "approximateBytesUsed" should be ... approximately correct. If not, why bother?

fzyzcjy:
@jonahwilliams I agree. So to do something actionable, maybe I should reopen #129054?

fzyzcjy:
Btw I also agree it is a bit of complexity, especially it needs to be maintained across so many different backends...

lib/web_ui/lib/src/engine/html_image_codec.dart
```diff
@@ -214,6 +214,9 @@ class HtmlImage implements ui.Image {
   @override
   ui.ColorSpace get colorSpace => ui.ColorSpace.sRGB;
 
+  @override
+  int get approximateBytesUsed => width * height * 4;
```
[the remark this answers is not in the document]
harryterkelsen: For `approximateBytesUsed` we should probably use `naturalWidth` and `naturalHeight`

lib/web_ui/lib/src/engine/canvaskit/image.dart
harryterkelsen: I added `Picture.approximateBytesUsed` to CanvasKit: https://github.com/google/skia/commit/46d3564c1e1b1814a12c6cdeeda5eccf636a216f

lib/web_ui/lib/src/engine/html_image_codec.dart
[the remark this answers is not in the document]
dnfield: @fzyzcjy - can you update this patch to account for that?

fzyzcjy:
Thank you, but I am not sure what I should do. Maybe I should wait until Googlers agree with one solution before I continue 👀

dnfield:
Let's close this to get it off review queues, I'llt ry to talk with Jonah about what we can do here.

PR #44904, opened 2023-08-21
[Impeller] Added recycled command buffers.

gaaclarke:
Recycles command buffers instead of deleting them when they are done being used.

@jonahwilliams profiled the Gallery on the Pixel 7 pro and saw that sometimes `RenderPassVK::OnEncodeCommands` takes a long time in https://github.com/flutter/flutter/issues/132690.  After adding adding more tracing it was found that the culprit was deleting the CommandBuffers on the raster thread.  This PR removes the deletion of the CommandBuffers and instead resets and recycles them.

unnamed

> where GARBO2 is
> ```c++
> void CommandPoolVK::GarbageCollectBuffersIfAble() {
>   TRACE_EVENT0("impeller", "CommandPoolVK::GARBO2");
>    if (std::this_thread::get_id() != owner_id_) {    
>      return;  
>    }
>    buffers_to_collect_.clear();
> }
> ```

This approach matches the guidances from ["Vulkan Best Practice for Mobile Developers"](https://arm-software.github.io/vulkan_best_practice_for_mobile_developers/samples/performance/command_buffer_usage/command_buffer_usage_tutorial.html) where it is preferable to recycle command buffers than delete them.  Also, the recommendation is that if you are not reusing command buffers that you should periodically reset the pool, which we weren't doing.

issue: https://github.com/flutter/flutter/issues/132690

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/impeller/renderer/backend/vulkan/command_pool_vk.cc
+++ b/impeller/renderer/backend/vulkan/command_pool_vk.cc
@@ -14,6 +14,10 @@
 
 namespace impeller {
 
+// This number comes from observing the recycled_buffers_ size while running
+// Wonderous and seeing it cap out at 7.
+static constexpr size_t kMaxRecycledBufferSize = 10;
+
 using CommandPoolMap = std::map<uint64_t, std::shared_ptr<CommandPoolVK>>;
 FML_THREAD_LOCAL fml::ThreadLocalUniquePtr<CommandPoolMap> tls_command_pool;
 
@@ -91,15 +95,22 @@ bool CommandPoolVK::IsValid() const {
 }
 
 void CommandPoolVK::Reset() {
-  Lock lock(buffers_to_collect_mutex_);
-  graphics_pool_.reset();
+  {
+    Lock lock(buffers_to_collect_mutex_);
+    graphics_pool_.reset();
 
-  // When the command pool is destroyed, all of its command buffers are freed.
-  // Handles allocated from that pool are now invalid and must be discarded.
-  for (vk::UniqueCommandBuffer& buffer : buffers_to_collect_) {
+    // When the command pool is destroyed, all of its command buffers are freed.
+    // Handles allocated from that pool are now invalid and must be discarded.
+    for (vk::UniqueCommandBuffer& buffer : buffers_to_collect_) {
+      buffer.release();
+    }
+    buffers_to_collect_.clear();
+  }
+
+  for (vk::UniqueCommandBuffer& buffer : recycled_buffers_) {
     buffer.release();
   }
-  buffers_to_collect_.clear();
+  recycled_buffers_.clear();
 
   is_valid_ = false;
 }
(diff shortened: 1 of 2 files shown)


impeller/renderer/backend/vulkan/command_pool_vk.cc
```diff
@@ -14,6 +14,10 @@
 
 namespace impeller {
 
+// This number comes from observing the recycled_buffers_ size while running
+// Wonderous and seeing it cap out at 7.
+static constexpr size_t kMaxRecycledBufferSize = 10;
```
gaaclarke, line 19: This may be moot since they will still probably live in the pool.

impeller/renderer/backend/vulkan/command_pool_vk.cc
```diff
@@ -148,6 +164,14 @@ void CommandPoolVK::GarbageCollectBuffersIfAble() {
   if (std::this_thread::get_id() != owner_id_) {
     return;
   }
+  for (auto& buffer : buffers_to_collect_) {
+    if (recycled_buffers_.size() >= kMaxRecycledBufferSize) {
+      // Don't grow boundlessly.
+      break;
+    }
+    buffer->reset();
+    recycled_buffers_.emplace_back(std::move(buffer));
```
gaaclarke, line 173: We might be able to get away with calling `reset` and `release`, letting the pool handle the recycling.

gaaclarke: I tried that and it didn't seem to be working as I expected.  Actually keeping the CommandBuffer around is a better way to assert that it is recycling at least.

impeller/renderer/backend/vulkan/command_pool_vk.cc
jonahwilliams: I'm not sure this is necessary since like you mention, its the pool itself that has memory. reseting/freeing should be more or less the same right?

gaaclarke: Yea, `kMaxRecycledBufferSize` isn't going to really work because it seems like the pool should be holding onto the command buffers anyways.  I removed it.

jonahwilliams (approved):
LGTM

impeller/renderer/backend/vulkan/command_pool_vk.cc
jonahwilliams: command buffers themselves don't own any memory, the command pool is where the pooled memory is.

gaaclarke:
This was a 10% drop in worst frame build time for `new_gallery_impeller__transition_perf` on `SM-G973U1`.

[link](https://flutter-flutter-perf.skia.org/e/?begin=1692601200&end=1692687600&keys=X2a7fcc91253cc635dea8a2aec4f71598&requestType=0&selected=commit%3D36415%26name%3D%252Carch%253Dintel%252Cbranch%253Dmaster%252Cconfig%253Ddefault%252Cdevice_type%253DSM-G973U1%252Cdevice_version%253Dnone%252Chost_type%253Dlinux%252Csub_result%253Dworst_frame_build_time_millis%252Ctest%253Dnew_gallery_impeller__transition_perf%252C)

PR #44912, opened 2023-08-21
Further filter/clear `<SkPaint>.setDither(true)`, this time in `DlSkPaintDispatchHelper`

matanlurey:
Closes https://github.com/flutter/flutter/issues/132860.

- If `setDither(true)` is called, and an existing `setColorSource` is a gradient, it is ignored.
- If `setColorSource(...)` is called, and it is a gradient, and dithering was previously set, it is cleared.

I'm not sure this is fool proof.


matanlurey:
This doesn't actually work as written :(

It's possible for code to run the following (otherwise valid) branch:

```
paint.setDither(true);
...
paint.setColorSource(aGradient);
```

With this PR, `setDither` is discarded (the last known colorSource was not a gradient).

Let me try to understand the lifecycle of this class better, maybe there is a better place to do this kind of check?

jonahwilliams:
Ahh that makes sense. You'd probably need to intercept the dithering call when it is being converted into an SkPaint.

matanlurey:
Appears to be all good now! Yay!

display_list/skia/dl_sk_paint_dispatcher.h
```diff
@@ -37,7 +37,19 @@ class DlSkPaintDispatchHelper : public virtual DlOpReceiver {
...
+  const SkPaint& paint() {
+    // On the Impeller backend, we will only support dithering of *gradients*,
+    // and it will be enabled by default (without the option to disable it).
+    // Until Skia support is completely removed, we only want to respect the
+    // dither flag for gradients (otherwise it will also apply to, for
+    // example, images, which is not supported in Impeller).
+    //
+    // See https://github.com/flutter/flutter/issues/112498.
+    if (!color_source_gradient_) {
+      paint_.setDither(false);
```
flar, line 49: Should this hinge on setDither(true) having been called?

So, 2 flags
- is_gradient, set from setColorSource as now
- is_dither, set from setDither (in lieu of setting sk_paint.setDither())

In this method, just paint_.setDither(is_gradient && is_dither);

flar (changes requested):

display_list/skia/dl_sk_paint_dispatcher.h
flar: Keep in mind that we might inherit a color source and/or a dither setting across multiple rendering calls. Those methods are only called when the associated attribute *changes*, they aren't called on every single rendering op.

So, you have to track the "last set values" for both and make your sk_paint dithering decision based on those last values, just on "a recently seen value" for one of them.

flar: I think where this will likely fail is
- dither true
- color_source non-gradient
- render something
- color_source gradient
- render something - your code will have the flag set to false in this case I think...?

matanlurey: Done, and done, thanks for the suggestion and added the last case as a test!

display_list/skia/dl_sk_paint_dispatcher.h
```diff
@@ -37,7 +37,19 @@ class DlSkPaintDispatchHelper : public virtual DlOpReceiver {
...
-  const SkPaint& paint() { return paint_; }
+  const SkPaint& paint() {
+    // On the Impeller backend, we will only support dithering of *gradients*,
+    // and it will be enabled by default (without the option to disable it).
+    // Until Skia support is completely removed, we only want to respect the
+    // dither flag for gradients (otherwise it will also apply to, for
+    // example, images, which is not supported in Impeller).
+    //
+    // See https://github.com/flutter/flutter/issues/112498.
+    if (color_source_gradient_ && dither_) {
```
flar, line 48: I don't think you want this inside an if, as it will get turned on once and never turned off. Just call `paint_.setDither(this && that)` in all cases. paint_ is reused across multiple rendering ops.

flar (changes requested):

display_list/skia/dl_sk_paint_dispatcher_unittests.cc
```diff
@@ -31,5 +43,44 @@ TEST(DisplayListUtils, OverRestore) {
...
+  MockDispatchHelper helper;
+
+  // "Render" a dithered non-gradient.
+  helper.setDither(true);
+  EXPECT_FALSE(helper.paint().isDither());
+
+  // "Render" a gradient.
+  helper.setColorSource(kTestLinearGradient.get());
+  EXPECT_TRUE(helper.paint().isDither());
+}
```
flar, line 83: Add a test that does:

dither=true
source=gradient
(true)
source=non-gradient
(false)
source=gradient
(true)
dither=false
(false)

display_list/skia/dl_sk_paint_dispatcher.h
matanlurey: Done! And thanks for catching this, I didn't realize it was re-used like that!

display_list/skia/dl_sk_paint_dispatcher_unittests.cc
matanlurey: Added!

flar (approved):
LGTM

PR #44918, opened 2023-08-21
Render only during callback scopes

goderbauer:
Enforces documented behavior in preparation for multi view rendering. Quoting from https://master-api.flutter.dev/flutter/dart-ui/FlutterView/render.html:

> This function must be called within the scope of the [PlatformDispatcher.onBeginFrame](https://master-api.flutter.dev/flutter/dart-ui/PlatformDispatcher/onBeginFrame.html) or [PlatformDispatcher.onDrawFrame](https://master-api.flutter.dev/flutter/dart-ui/PlatformDispatcher/onDrawFrame.html) callbacks being invoked.
>
> If this function is called a second time during a single [PlatformDispatcher.onBeginFrame](https://master-api.flutter.dev/flutter/dart-ui/PlatformDispatcher/onBeginFrame.html)/[PlatformDispatcher.onDrawFrame](https://master-api.flutter.dev/flutter/dart-ui/PlatformDispatcher/onDrawFrame.html) callback sequence or called outside the scope of those callbacks, the call will be ignored.

Fixes https://github.com/flutter/flutter/issues/125326.

This change is passing a TGP in g3: tap/OCL:558891137:BASE:558891211:1692660347578:e108b061

### DIFF
--- a/lib/ui/platform_dispatcher.dart
+++ b/lib/ui/platform_dispatcher.dart
@@ -341,7 +341,7 @@ class PlatformDispatcher {
     assert(_renderedViews == null);
     assert(_renderedViewsBetweenCallbacks == null);
 
-    _renderedViews = Set<FlutterView>();
+    _renderedViews = <FlutterView>{};
     _invoke1<Duration>(
       onBeginFrame,
       _onBeginFrameZone,
--- a/lib/ui/platform_dispatcher.dart
+++ b/lib/ui/platform_dispatcher.dart
@@ -315,7 +315,13 @@ class PlatformDispatcher {
   // calls to [FlutterView.render] must be ignored. Furthermore, if a given
   // [FlutterView] is already present in this set when its [FlutterView.render]
   // is called again, that call must be ignored as a duplicate.
+  //
+  // Between [onBeginFrame] and [onDrawFrame] the properties value is
+  // temporarily stored in `_renderedViewsBetweenCallbacks` so that it survives
+  // the gap between the two callbacks.
   Set<FlutterView>? _renderedViews;
+  // Temporary storage of the `_renderedViews` value between `_beginFrame` and
+  // `_drawFrame`.
   Set<FlutterView>? _renderedViewsBetweenCallbacks;
 
   /// A callback invoked when any view begins a frame.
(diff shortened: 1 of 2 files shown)


lib/ui/platform_dispatcher.dart
```diff
@@ -308,6 +308,15 @@ class PlatformDispatcher {
...
 
+  // [FlutterView]s for which [FlutterView.render] has already been called
+  // during the current [onBeginFrame]/[onDrawFrame] callback sequence.
+  //
+  // The field is null outside the scope of those callbacks indicating that
+  // calls to [FlutterView.render] must be ignored. Furthermore, if a given
+  // [FlutterView] is already present in this set when its [FlutterView.render]
+  // is called again, that call must be ignored as a duplicate.
+  Set<FlutterView>? _renderedViews;
+  Set<FlutterView>? _renderedViewsBetweenCallbacks;
```
dkwingsmt, line 319:
```suggestion
  // [FlutterView]s for which [FlutterView.render] has already been called
  // during the last `_beginFrame` call.
  //
  // The field holds the result of `_renderedViews` temporarily between `_beginFrame` and `_drawFrame`, 
  // so that `_renderedViews` can be null during the gap, but still gets the previous value 
  // at the at the beginning of `_drawFrame`.
  Set<FlutterView>? _renderedViewsInPastCalls;
```

lib/ui/platform_dispatcher.dart
```diff
@@ -308,6 +308,15 @@ class PlatformDispatcher {
...
   }
 
+  // [FlutterView]s for which [FlutterView.render] has already been called
+  // during the current [onBeginFrame]/[onDrawFrame] callback sequence.
+  //
+  // The field is null outside the scope of those callbacks indicating that
+  // calls to [FlutterView.render] must be ignored. Furthermore, if a given
+  // [FlutterView] is already present in this set when its [FlutterView.render]
+  // is called again, that call must be ignored as a duplicate.
+  Set<FlutterView>? _renderedViews;
```
dkwingsmt, line 318:
```suggestion
  // [FlutterView]s for which [FlutterView.render] has already been called
  // during the current `_beginFrame`/`_drawFrame` callback sequence.
  //
  // The field is null outside the scope of those callbacks; specifically, its value is
  // temporarily stored in `_renderedViewsInPastCalls` at the end of `_beginFrame`
  // so that the value can survive through the gap. In this way,
  // calls to [FlutterView.render] out of those callbacks are ignored, and
  // one view can only be rendered once in a callback sequence.
  Set<FlutterView>? _renderedViews;
```

goderbauer:
Closing this as it has been taken over by @dkwingsmt.

PR #44930, opened 2023-08-21
Disallow using ./tools/gn --enable-unittests --android (or --ios)

matanlurey:
Closes https://github.com/flutter/flutter/issues/132611.

As seen on https://github.com/flutter/flutter/issues/132611, this will never succeed compiling.

I tried to update the Wiki as well:
https://github.com/flutter/flutter/wiki/Setting-up-the-Engine-development-environment#vscode-with-cc-intellisense-cc


tools/gn
```diff
@@ -299,6 +299,17 @@ def to_gn_args(args):
   gn_args['full_dart_sdk'] = args.full_dart_sdk
 
   if args.enable_unittests:
+    # Ensure that Android is *not* enabled for unit tests.
+    # https://github.com/flutter/flutter/issues/132611
+    if args.target_os == 'android':
```
zanderso, line 304: IIRC, iOS will also not work.

zanderso (approved):
Thanks!

PR #44936, opened 2023-08-22
Enable clang-tidy for pre-push (opt-out), exclude `performance-unnecessary-value-param`

matanlurey:
Closes https://github.com/flutter/flutter/issues/132687.

Zach, this is a pretty naive attempt, so feel free to suggest alternatives. I considered for example just using `--checks=-{{LINT}}` as well, but figured that might be less discoverable (especially for non-core folks) than just a file called `.clang-tidy-for-githooks`.

/cc @jonahwilliams


matanlurey:
Whomever reviews this, I have no idea how to pass the LICENSES check and could use an adult!

tools/clang_tidy/lib/src/options.dart
```diff
@@ -70,10 +71,21 @@ class Options {
     required List<io.File> shardCommandsPaths,
     int? shardId,
   }) {
+    io.File? configPath;
+    if (options.wasParsed('config-file')) {
```
zanderso, line 75: Looks like the argument validation is happening further down in `_checkArguments` on line 284.

tools/githooks/lib/src/pre_push_command.dart
```diff
@@ -81,6 +77,22 @@ class PrePushCommand extends Command<bool> {
     return true;
   }
 
+  io.File? _findCompileCommands(String flutterRoot) {
+    for (final String dir in _checkForHostTargets) {
```
zanderso, line 81: Maybe choose the one with the most recent file modification time? (Not sure if modification or creation time would be better...) Which one was chosen should probably also be printed to the terminal.

.clang-tidy-for-githooks
```diff
@@ -0,0 +1,41 @@
+# Keep in sync with .clang-tidy.
```
zanderso, line 1: We should try to think of some ideas to avoid the duplication. Here's one:

The `.clang-tidy` file is just a yaml file. If `clang-tidy` ignores fields it doesn't understand, we could add one in the existing `.clang-tidy` file, like `IgnoredCheckForFlutterGitHook` (or some better name). Then when running in the context of the pre-push hook, our `clang-tidy` wrapper can create a new `.clang-tidy` file under `/tmp`, and pass that to the `clang-tidy` `--config-file` flag.

zanderso:
> Whomever reviews this, I have no idea how to pass the LICENSES check and could use an adult!

There are instructions at the bottom of the logs for the failing check: https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8772102403923285345/+/u/test:_licenses_check/stdout?format=raw.

tools/clang_tidy/lib/clang_tidy.dart
```diff
@@ -292,6 +299,10 @@ class ClangTidy {
     List<Command> commands,
     Options options,
   ) async {
+    io.File? configPath = options.configPath;
+    if (configPath != null && options.excludeSlowChecks) {
+      configPath = _createClangTidyConfigExcludingSlowLints(configPath);
```
zanderso, line 304: The temp directory should probably be deleted sometime after `_runJobs` finishes up on line 171.

zanderso (approved):
lgtm w/ temp dir cleanup.

matanlurey:
Done!

PR #44947, opened 2023-08-22
[iOS] Don't call App lifecycle resumed when device is locked

zhongwuzw:
Fixed https://github.com/flutter/flutter/issues/130055.
We can cancel the `resumed` life cycle notification if we received `inactive` notification in the same runloop.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/shell/platform/darwin/ios/framework/Source/FlutterViewController.mm
+++ b/shell/platform/darwin/ios/framework/Source/FlutterViewController.mm
@@ -1035,10 +1035,15 @@ static void SendFakeTouchEvent(UIScreen* screen,
   if (_viewportMetrics.physical_width) {
     [self surfaceUpdated:YES];
   }
-  [self goToApplicationLifecycle:@"AppLifecycleState.resumed"];
+  [self performSelector:@selector(goToApplicationLifecycle:)
+             withObject:@"AppLifecycleState.resumed"
+             afterDelay:0.0f];
 }
 
 - (void)appOrSceneWillResignActive {
+  [NSObject cancelPreviousPerformRequestsWithTarget:self
+                                           selector:@selector(goToApplicationLifecycle:)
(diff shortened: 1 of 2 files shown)


chinmaygarde (approved):
lgtm but perhaps @stuartmorgan can also find a reviewer.

shell/platform/darwin/ios/framework/Source/FlutterViewController.mm
```diff
@@ -1035,10 +1035,15 @@ - (void)appOrSceneBecameActive {
   if (_viewportMetrics.physical_width) {
     [self surfaceUpdated:YES];
   }
-  [self goToApplicationLifecycle:@"AppLifecycleState.resumed"];
+  [self performSelector:@selector(goToApplicationLifecycle:)
+             withObject:@"AppLifecycleState.resumed"
+             afterDelay:0.0f];
```
stuartmorgan, line 1040: This seems dangerous; what prevents us from mis-ordering lifecycle events because just one of them is delayed?

zhongwuzw: > This seems dangerous; what prevents us from mis-ordering lifecycle events because just one of them is delayed?

When the device like iPhone14 pro is locked, the system mistakenly sends an "active" event, so we may need to cancel it if we receive an inactive notification in the same runloop.

stuartmorgan: I understand what the goal of the PR is; that's not what I asked. I'm asking why we should have confidence that this will not mis-order the event relative to other lifestyle events in the cases where it's not cancelled.

zhongwuzw: active/resumed is the final event of application lifecycle events when app changed from inactive to active.

PR #44961, opened 2023-08-22
Re-write docs for DlSkCanvas{Adapter|Dispatcher}.

matanlurey:
Closes https://github.com/flutter/flutter/issues/132994.

I suspect this is imperfect, so please suggest any changes or additions (either using GitHub's tooling or as a review comment).

/cc @dnfield


flar:
It's a start. It might be worth mentioning the odd stateful nature of the paint attributes which I don't think DlOpReceiver really describes in much detail. There is a short description in that interface right before all of the setFoo() attribute setters, but it should be called out in the "class comment" and we should also point out that they persist through and beyond save/restore calls unlike the clip and transform state.

flar (changes requested):
Submitting review based on previous comments

matanlurey:
Added a note!

display_list/skia/dl_sk_canvas.h
```diff
@@ -10,8 +10,12 @@
...
 namespace flutter {
 
-// An adapter to receive DlCanvas calls and dispatch them into
-// an SkCanvas.
+// -----------------------------------------------------------------------------
+/// @brief      A Canvas implementation that delegates to an Skia-backed Canvas.
+///
+/// This class is intended to be used an internal implementation detail for the
+/// DisplayList package. @see DlSkCanvasDispatcher for the developer-facing
+/// interface to Skia.
```
flar, line 18: Whoops, I just noticed this is backwards.

DlCanvas is the public facing API.
DlOpReceiver is internal.

We want the framework, layers, embedders, other engine code to all be using the DlCanvas API. This class implements DlCanvas for Skia. While this class isn't itself the public facing API, it is implementing the public-facing API.

display_list/skia/dl_sk_dispatcher.h
```diff
@@ -14,11 +14,20 @@
 namespace flutter {
 
 //------------------------------------------------------------------------------
-/// Can be fed to the dispatch() method of a DisplayList to feed the resulting
-/// rendering operations to an SkCanvas instance.
+/// @brief      A Dispatcher that feeds rendering operations to an SkCanvas.
 ///
-/// Receives all methods on Dispatcher and sends them to an SkCanvas
+/// This class is intended to be used as the developer-facing interface to Skia
```
flar, line 19: Similarly, this is *NOT* the developer-facing API.

DlOpReceiver is the internal API that only backends implement, and this is Skia's implementation of it.

flar (changes requested):
These comments are backwards and misplaced.

These files are the Skia-specific implementation of DL interfaces. None of them are developer facing as is, but one of them does at least implement a developer-facing API and that is the DlSkCanvasAdapter which implements the public DlCanvas interface.

The comments here should be on DlCanvas and DlOpReceiver (and they need to be swapped for the most part, except that the comment about state is appropriate for the Receiver implementation (named Dispatcher because that is what it used to be called)).

Mostly these 2 Skia class should just say "the Skia-specific implementation of Dl..." and the comments about whether they are developer-facing or not should be on the DlFoo interfaces.

flar:
Class comment for DlCanvas in `dl_canvas.h` should be something like:

-------
This is the engine developer facing API for rendering anything within the engine. It should be the API used to render anything in the framework classes (lib/ui), flow and flow/layers, embedders, shell, etc.

The only state carried by implementations of this interface are the clip and transform which are saved and restored by the |save|, |saveLayer|, and |restore| calls. (possibly explain that in more detail if you think it needs it)

The interface resembles closely the familiar SkCanvas interface that was used throughout the engine.

@see DlSkAdapter(sp?)
@see impeller::<class that Dan is working on>
-------

Then DlOpReceiver in `dl_op_receiver.h` should be something like:

---
This is an internal implementation interface for rendering backends that need to interact with recorded display lists. The `DisplayList` object will play back recorded operations in this format. Most developers should not needed to deal with this interface unless they are writing a utility that needs to examine the contents of a display list.

[Mention that it also has the same clip/transform state as DlCanvas, and...]
[Description of attribute state]

@see DlSkDispatcher(sp?)
@see impeller::DlDispatcher(sp?)
@see DlOpSpy
---

DlSk* are only implementations, they shouldn't be the place where we document these interfaces. They should simply point people to the DlCanvas or DlOpReceiver interfaces and mention that they are the Skia-specific implementation of those interface classes...

display_list/skia/dl_sk_canvas.h
matanlurey: Simplified to point at the interface.

display_list/skia/dl_sk_dispatcher.h
matanlurey: Simplified to point at the interface.

matanlurey:
Tried my best :), feel free to point out anything additional!

display_list/dl_op_receiver.h
```diff
@@ -23,11 +23,19 @@ namespace flutter {
...
-///             invoked through the DisplayList::dispatch() method.
+/// @brief      Internal API for rendering recorded display lists to backends.
 ///
+/// The |DisplayList| object will play back recorded operations in this format.
+/// Most developers should not needed to deal with this interface unless they
+/// are writing a utility that needs to examine the contents of a display list.
+///
+/// Similar to |DlCanvas|, the only state carried by implementations of this
+/// interface are the clip and transform which are saved and restored by the
+/// |save|, |saveLayer|, and |restore| calls.
```
flar, line 34: Unlike DlCanvas, this interface has attribute state which is global across an entire DisplayList (not affected by save/restore).

flar (changes requested):

display_list/dl_op_receiver.h
flar: Basically at the start of processing any given DL, the attributes should all have default state and then modified each time the receiver gets a setAttribute() call.

[Side note: This gets tricky for the renderers that implement this. Utilities that are simply scanning a DL can just skip over the `drawDisplayList` calls, but renderers have to immediately process those DisplayLists in situ and so they have to maintain separate attribute states for the current DL and the new DL that they recurse into. For Skia I went with just creating a brand new dispatcher and feeding the new DL into that. For Impeller, I reuse "this" and just do a lot of saving state and restoring of state before/after...]

matanlurey: Done!

flar: You should remove the "the only state" as it isn't the only state. Something like "similar to ... this interface carries clip and transform state which ..."

display_list/dl_op_receiver.h
```diff
@@ -23,11 +23,22 @@ namespace flutter {
...
 
 //------------------------------------------------------------------------------
-/// @brief      The pure virtual interface for interacting with a display list.
-///             This interface represents the methods used to build a list
-///             through the DisplayListBuilder and also the methods that will be
-///             invoked through the DisplayList::dispatch() method.
+/// @brief      Internal API for rendering recorded display lists to backends.
 ///
+/// The |DisplayList| object will play back recorded operations in this format.
+/// Most developers should not needed to deal with this interface unless they
```
flar, line 29: "should not need~ed~ to deal with"

flar (approved):
Just some minor grammar typos to fix...

display_list/dl_op_receiver.h
matanlurey: Done!

display_list/dl_op_receiver.h
matanlurey: Done!

PR #43750, opened 2023-07-17
FontVariation.lerp, custom FontVariation constructors, and more documentation

Hixie:
This should aid with implementing the framework side of https://github.com/flutter/flutter/issues/105120 This should also address https://github.com/flutter/flutter/issues/28543.

PR #44996, opened 2023-08-23
FontVariation.lerp, custom FontVariation constructors, and more documentation

Hixie:
This should aid with implementing the framework side of https://github.com/flutter/flutter/issues/105120.
This should also address https://github.com/flutter/flutter/issues/28543.

This is a reland of https://github.com/flutter/engine/pull/43750 with two changes, one to fix a typo mentioned in https://github.com/flutter/engine/pull/43750#issuecomment-1689299114, and one to fix the analysis error found when rolling this to the framework (https://ci.chromium.org/ui/p/flutter/builders/try/Linux%20analyze/62845/overview). The latter change is temporary and can be relaxed when FontVariations is reexported from dart:ui. I plan to do that when fixing https://github.com/flutter/flutter/issues/105120.

### DIFF
--- a/lib/ui/text.dart
+++ b/lib/ui/text.dart
@@ -3,16 +3,51 @@
 // found in the LICENSE file.
 part of dart.ui;
 
-/// Whether to slant the glyphs in the font
+/// Whether to use the italic type variation of glyphs in the font.
+///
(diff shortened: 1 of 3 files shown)


eyebrowsoffire (approved):
LGTM!

PR #45021, opened 2023-08-23
[Impeller] combine uniform metadata and buffer slots.

jonahwilliams:
Follow up from https://github.com/flutter/engine/pull/44990

We're spending a ton of time deallocating mostly empty maps, by reducing the number of maps we reduce the amount of memory used, and reduce map lookups.


impeller/renderer/command.cc
```diff
@@ -61,13 +61,13 @@ bool Command::DoBindResource(ShaderStage stage,
 
   switch (stage) {
     case ShaderStage::kVertex:
-      vertex_bindings.uniforms[slot.ext_res_0] = slot;
-      vertex_bindings.buffers[slot.ext_res_0] = BufferResource(metadata, view);
```
gaaclarke, line 65: Note one has versus 2.

gaaclarke (approved):

gaaclarke:
You know, thinking about this I think it might be preferable to go down the assoc vector route instead or in addition of this route.  Let's chat you me and Dan.

jonahwilliams:
If you want to refactor that later, thats fine, but we still need to reduce the number of containers.

PR #45024, opened 2023-08-23
[Impeller] Dat rvalue reference (fix engine head)

bdero:
Not sure why we didn't catch this on presubmits. I think there was a merge order problem.


jonahwilliams (approved):
LGTM

PR #45025, opened 2023-08-23
[Impeller] Fix stencil buffer format selection on Vulkan backend, add support for D24UnormS8Uint

dnfield:
Vulkan was checking for D24 and using D32. This is part of the problem on the Samsung A02. There is still some other problem.

Fixes https://github.com/flutter/flutter/issues/133188


jonahwilliams (approved):
LGTM

chinmaygarde (approved):

PR #44993, opened 2023-08-23
Revert "FontVariation.lerp, custom FontVariation constructors, and more documentation"

zanderso:
Reverts flutter/engine#43750

Looks like this is having trouble rolling into the framework: https://ci.chromium.org/ui/p/flutter/builders/try/Linux%20analyze/62845/overview

PR #45030, opened 2023-08-23
FontVariation.lerp, custom FontVariation constructors, and more documentation

Hixie:
This should aid with implementing the framework side of https://github.com/flutter/flutter/issues/105120.
This should also address https://github.com/flutter/flutter/issues/28543.

Original commit in https://github.com/flutter/engine/pull/43750 was reverted in https://github.com/flutter/engine/pull/44993.
Second attempt in https://github.com/flutter/engine/pull/44996 was reverted in https://github.com/flutter/engine/pull/45023.

### DIFF
--- a/lib/ui/text.dart
+++ b/lib/ui/text.dart
@@ -3,16 +3,51 @@
 // found in the LICENSE file.
 part of dart.ui;
 
-/// Whether to slant the glyphs in the font
+/// Whether to use the italic type variation of glyphs in the font.
+///
+/// Some modern fonts allow this to be selected in a more fine-grained manner.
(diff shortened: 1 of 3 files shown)


Hixie:
I actually ran the framework `analyze.dart` tests (locally) this time, which increases the odds that this is correct...

eyebrowsoffire (approved):
LGTM. I ran a test run here, and the only failures are the expected ones (to block us from committing an engine realm file, and some of the Fuchsia CI) and some other flake that doesn't look related. Hopefully this should roll into the framework smoothly🤞

PR #45040, opened 2023-08-24
[Impeller] Don't place vertex buffer bindings in the Binding map.

jonahwilliams:
The metal backend is the only backend that doesn't treat binding the vertex buffer specially. For the GLES and Vulkan backend, we can instead pull it out of the cmd map entirely.

This will make it easier to partially inline the map into the bindings object.

https://github.com/flutter/flutter/issues/133199


chinmaygarde (approved):
Good call. Metal is the odd one out here.

PR #45047, opened 2023-08-24
Revert "[Impeller] add trace events for VkRenderPass and VkFrameBuffe…

jonahwilliams:
…r creation. (#44837)"

This reverts commit ca4ea82258a661f81aa3dec9c6e668c686f0e820.

These trace events aren't super useful, since we've tracked down the issues as either 1) phone is overloaded due to image decoding at large size or 2) https://github.com/flutter/flutter/issues/133198


chinmaygarde (approved):
I don't know why but.... ok?

PR #45056, opened 2023-08-24
[Linux] Fix channel buffers control commands error handling

bleroux:
## Description

This PR fixes a mistake I made on https://github.com/flutter/engine/pull/44636 where the error handling code wrongly relied on `fl_method_channel_invoke_method_finish` instead of  `fl_binary_messenger_send_on_channel_finish`.
The error handling code was not called when running the tests added in https://github.com/flutter/engine/pull/44636 so this mistake did not pop up.

@robert-ancell I added a test that simulates an error response and I had to rely on `g_idle_add` to make it works. Is this approach ok?

## Related Issue

Linux implementation for https://github.com/flutter/flutter/issues/132386

## Tests

Adds one test.

### DIFF
--- a/shell/platform/linux/fl_binary_messenger_test.cc
+++ b/shell/platform/linux/fl_binary_messenger_test.cc
@@ -501,6 +501,55 @@ TEST(FlBinaryMessengerTest, AllowOverflowChannel) {
   EXPECT_TRUE(called);
 }
 
+static gboolean quit_main_loop_cb(gpointer user_data) {
+  g_main_loop_quit(static_cast<GMainLoop*>(user_data));
+  return FALSE;
+}
+
+// Checks if error returned when invoking a command on the control channel
+// are handled.
+TEST(FlBinaryMessengerTest, ControlChannelErrorResponse) {
+  g_autoptr(GMainLoop) loop = g_main_loop_new(nullptr, 0);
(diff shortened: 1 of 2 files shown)


shell/platform/linux/fl_binary_messenger_test.cc
```diff
@@ -501,6 +501,55 @@ TEST(FlBinaryMessengerTest, AllowOverflowChannel) {
...
+          return old_handler(engine, message);
+        }
+
+        called = true;
+
+        // Simulates an internal error.
+        return kInvalidArguments;
+      }));
+
+  // Register a callbalk to quit the main loop when binary messenger work ends.
```
robert-ancell, line 541: typo in callback.

shell/platform/linux/fl_binary_messenger_test.cc
```diff
@@ -501,6 +501,55 @@ TEST(FlBinaryMessengerTest, AllowOverflowChannel) {
...
+  g_idle_add(quit_main_loop_cb, loop);
```
robert-ancell, line 542: Can this be put inside the mock function, i.e. after the `called = true`? Although it works here it seems less clear than if it occurs after the error is detected.

robert-ancell (changes requested):

robert-ancell (approved):
Looks great, thanks!

PR #45070, opened 2023-08-24
[Impeller] avoid hashing and std::vector growth when binding descriptor sets.

jonahwilliams:
We call allocateAndBindDescriptorSets once for every cmd rendered, so this time can bubble up quite a bit. From the traces I've gathered map emplacement (emplace_unique_key_args) and std::vector growth (__push_back_slow_path) contribute the most to this trace.

I think we have physical limits on the number of descriptors we can bind, though I'm not sure if we enforce that anywhere?

### DIFF
--- a/impeller/renderer/backend/vulkan/render_pass_vk.cc
+++ b/impeller/renderer/backend/vulkan/render_pass_vk.cc
@@ -321,20 +321,20 @@ static bool AllocateAndBindDescriptorSets(const ContextVK& context,
 
   auto& allocator = *context.GetResourceAllocator();
 
-  vk::DescriptorImageInfo images[32];
-  vk::DescriptorBufferInfo buffers[32];
-  size_t image_count = 0u;
-  size_t buffer_count = 0u;
+  std::vector<vk::DescriptorImageInfo> images;
+  std::vector<vk::DescriptorBufferInfo> buffers;
   std::vector<vk::WriteDescriptorSet> writes;
   writes.reserve(command.vertex_bindings.buffers.size() +
                  command.fragment_bindings.buffers.size() +
                  command.fragment_bindings.sampled_images.size());
+  images.reserve(command.fragment_bindings.sampled_images.size());
+  buffers.reserve(command.vertex_bindings.buffers.size() +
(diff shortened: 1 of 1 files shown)


jonahwilliams:
Can/should we add a hard limit to the binding count somewhere?

impeller/renderer/backend/vulkan/render_pass_vk.cc
```diff
@@ -353,20 +358,23 @@ static bool AllocateAndBindDescriptorSets(const ContextVK& context,
       write_set.dstBinding = slot.binding;
       write_set.descriptorCount = 1u;
       write_set.descriptorType = vk::DescriptorType::eCombinedImageSampler;
-      write_set.pImageInfo = &(images[slot.binding] = image_info);
+      FML_DCHECK(image_count < 32u);
+      write_set.pImageInfo = &(images[image_count] = image_info);
```
jason-simmons, line 362: The old implementation would only store one entry if multiple images used the same `slot.binding` index.  Is that behavior desirable?

@chinmaygarde

impeller/renderer/backend/vulkan/render_pass_vk.cc
```diff
@@ -322,14 +321,20 @@ static bool AllocateAndBindDescriptorSets(const ContextVK& context,
 
   auto& allocator = *context.GetResourceAllocator();
 
-  std::unordered_map<uint32_t, vk::DescriptorBufferInfo> buffers;
-  std::unordered_map<uint32_t, vk::DescriptorImageInfo> images;
+  vk::DescriptorImageInfo images[32];
```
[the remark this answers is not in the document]
jonahwilliams: No I think that is probably a better idea. We should be able to assert on the size to make sure the ptrs are valid

impeller/renderer/backend/vulkan/render_pass_vk.cc
jonahwilliams: The bindings shouldn't be duplicated, but even if they were the image/buffer map we are processing should have already dedicated (I think)

jason-simmons (approved):

PR #45079, opened 2023-08-24
[Impeller] Fix mask blurs and the Gaussian blur coverage hint.

bdero:
Resolves https://github.com/flutter/flutter/issues/132936.

* In some cases, mask blurs were getting applied twice.
* Coverage hint clipping was too aggressive on the second pass of the gaussian blur.
* Rework the mask blur applicator to support gracefully falling back to GPU-based color filters when necessary.

This fix requires us to cherry-pick: https://github.com/flutter/engine/pull/43519
Going a revert-based route to fix all of these problems would also be pretty complex.

I've been testing this against 3.13 and still need to cross reference the other issues. I haven't rigorously tested this against HEAD yet.

https://github.com/flutter/engine/assets/919017/9ee5e856-af7a-42b9-b135-ea268c2ba53f

### DIFF
--- a/impeller/aiks/canvas.cc
+++ b/impeller/aiks/canvas.cc
@@ -492,7 +492,7 @@ void Canvas::DrawImageRect(const std::shared_ptr<Image>& image,
   Entity entity;
   entity.SetBlendMode(paint.blend_mode);
   entity.SetStencilDepth(GetStencilDepth());
-  entity.SetContents(paint.WithFilters(contents, false));
+  entity.SetContents(paint.WithFilters(contents));
   entity.SetTransformation(GetCurrentTransformation());
 
   GetCurrentPass().AddEntity(entity);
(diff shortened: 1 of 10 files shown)


impeller/aiks/canvas.cc
```diff
@@ -561,8 +561,13 @@ void Canvas::DrawTextFrame(const TextFrame& text_frame,
     color_text_contents->SetColorSourceContents(
         paint.color_source.GetContents(paint));
 
-    entity.SetContents(
-        paint.WithFilters(std::move(color_text_contents), false));
+    // TODO(bdero): This mask blur application is a hack. It will always wind up
```
jonahwilliams, line 564: Do we have bugs tracking these?

jonahwilliams (approved):
LGTM

impeller/aiks/canvas.cc
bdero: Whoops, thanks for the reminder. Filed https://github.com/flutter/flutter/issues/133297.

bdero:
Where are the dang goldens?

jonahwilliams:
Yeah this isn't working for me either in https://github.com/flutter/engine/pull/45090

bdero:
Trying to rebase to see if the goldens will pop out this time.

bdero:
I went through the goldens manually and things are looking mostly good. 

`CanRenderForegroundAdvancedBlendWithMaskBlur` is rendering wrong -- but I think that's a negative test I landed to track progress and am double-checking that this patch doesn't introduce a regression.

bdero:
Yeah we're not handling that case correctly before/after. This patch doesn't introduce a regression. Filed a bug to track it -- I thought I had already fixed this problem: https://github.com/flutter/flutter/issues/133354

PR #45080, opened 2023-08-24
[Impeller] fix validation warning on iOS/macOS when compiling external texture shader.

jonahwilliams:
```
[VERBOSE-2:validation.cc(49)] Break on 'ImpellerValidationBreak' to inspect point of failure: Could not create render pipeline for TextureFillExternal Pipeline :Fragment input(s) `user(locn1)` mismatching vertex shader output type(s) or not written by vertex shader
```

Fixes https://github.com/flutter/flutter/issues/133286
Fixes https://github.com/flutter/flutter/issues/133268


bdero (approved):
We should exclude this shader and pipeline cache on non-GL backends. Filed an issue: https://github.com/flutter/flutter/issues/133286

jonahwilliams:
I'll just fix that now

impeller/renderer/backend/metal/pipeline_library_mtl.mm
```diff
@@ -112,7 +112,8 @@
       ^(id<MTLRenderPipelineState> _Nullable render_pipeline_state,
         NSError* _Nullable error) {
         if (error != nil) {
-          VALIDATION_LOG << "Could not create render pipeline: "
```
jonahwilliams, line 115: This change would have made the error a lot easier to find for someone that isn't building the engine.

jonahwilliams:
Updated this to ifdef out the shader on non android platforms, and check if the context is GLES before initializing on Android.

impeller/renderer/backend/metal/pipeline_library_mtl.mm
bdero: This is a nice add. 👍

impeller/entity/contents/content_context.cc
```diff
@@ -312,7 +310,13 @@ ContentContext::ContentContext(
       CreateDefaultPipeline<YUVToRGBFilterPipeline>(*context_);
   porter_duff_blend_pipelines_[default_options_] =
       CreateDefaultPipeline<PorterDuffBlendPipeline>(*context_);
-
+  // GLES Android only shader.
+#ifdef FML_OS_ANDROID
```
bdero, line 314: Nit: Gate on building the GLES backend, not Android? This will probably come in handy for other desktop/embedded GL contexts. (Even in the case that we don't ship desktop GL for Flutter)

bdero (approved):

impeller/entity/contents/content_context.cc
jonahwilliams: Done

PR #45029, opened 2023-08-23
Turn on the `skia_enable_optimize_size` flag to save a bit of binary size

eyebrowsoffire:
This flag removes some code from CanvasKit to reduce size by a little bit. I went ahead and did a run of the benchmarks (https://github.com/flutter/flutter/pull/133208) to see if it negatively affected anything, and there was no difference beyond noise between the current benchmark numbers and the benchmarks with this flag enabled.

The size differences are as follows:
Before the change:
```
total 30616
drwxr-xr-x  11 jacksongardner  primarygroup      352 Aug 23 14:33 .
drwxr-xr-x   7 jacksongardner  primarygroup      224 Aug 10 18:14 ..
-rw-r--r--@  1 jacksongardner  primarygroup     6148 May 12 17:41 .DS_Store
-rw-r--r--   2 jacksongardner  primarygroup    94899 Aug 23 14:23 canvaskit.js
-rwxr-xr-x   2 jacksongardner  primarygroup  6631693 Aug 23 14:23 canvaskit.wasm
-rwxr-xr-x   1 jacksongardner  primarygroup  2102151 Aug 23 14:23 canvaskit.wasm.br
drwxr-xr-x   5 jacksongardner  primarygroup      160 Aug 23 14:33 chromium
-rw-r--r--   2 jacksongardner  primarygroup   161478 Aug 23 14:28 skwasm.js
-rwxr-xr-x   2 jacksongardner  primarygroup  3296038 Aug 23 14:28 skwasm.wasm
-rwxr-xr-x   1 jacksongardner  primarygroup  1101502 Aug 23 14:28 skwasm.wasm.br
-rw-r--r--   2 jacksongardner  primarygroup     3095 Aug 23 14:28 skwasm.worker.js

./chromium:
total 15520
drwxr-xr-x   5 jacksongardner  primarygroup      160 Aug 23 14:33 .
drwxr-xr-x  11 jacksongardner  primarygroup      352 Aug 23 14:33 ..
-rw-r--r--   2 jacksongardner  primarygroup    94545 Aug 23 14:25 canvaskit.js
-rwxr-xr-x   2 jacksongardner  primarygroup  5223378 Aug 23 14:25 canvaskit.wasm
-rwxr-xr-x   1 jacksongardner  primarygroup  1492433 Aug 23 14:25 canvaskit.wasm.br
```

After the change:
```
total 28568
drwxr-xr-x  11 jacksongardner  primarygroup      352 Aug 23 14:42 .
[...]

PR #45082, opened 2023-08-24
Revert "Turn on the `skia_enable_optimize_size` flag to save a bit of binary size"

eyebrowsoffire:
Reverts flutter/engine#45029

It appears this is causing a regression in the length of time some of our integration tests are taking, causing them to take twice as long, which is causing timeouts in CI. We should revert while we investigate.


flar (approved):
LGTM

PR #45088, opened 2023-08-24
[Impeller] Refactor: Create attachment descriptions without setting layouts

gaaclarke:
This refactor makes recycling render passes easier since it starts to split out setting the layout of textures with creating the render pass description.

I didn't fully implement the split since it would technically be slower.  Until we get caching of render passes working it would require us to calculate the attachment descriptions twice, once to make the render pass and once to set the texture layout.

teases out some of the work from: https://github.com/flutter/engine/pull/44527
in the service of implementing: https://github.com/flutter/engine/pull/44861
issue: https://github.com/flutter/flutter/issues/133182

The final split will have this function for recycled render passes:
```c++
void SetTextureLayouts(const RenderTarget& render_target,
                       const std::shared_ptr<CommandBufferVK>& command_buffer) {
  for (const auto& [bind_point, color] : render_target.GetColorAttachments()) {
    SetTextureLayout(color,
                     CreateAttachmentDescription(color, &Attachment::texture),
                     command_buffer, &Attachment::texture);
    if (color.resolve_texture) {
      SetTextureLayout(
          color,
          CreateAttachmentDescription(color, &Attachment::resolve_texture),
          command_buffer, &Attachment::resolve_texture);
    }
  }

  if (auto depth = render_target.GetDepthAttachment(); depth.has_value()) {
    SetTextureLayout(
        depth.value(),
        CreateAttachmentDescription(depth.value(), &Attachment::texture),
        command_buffer, &Attachment::texture);
  }

  if (auto stencil = render_target.GetStencilAttachment();
      stencil.has_value()) {
    SetTextureLayout(
        stencil.value(),
        CreateAttachmentDescription(stencil.value(), &Attachment::texture),
        command_buffer, &Attachment::texture);
  }
}
```

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/impeller/renderer/backend/vulkan/render_pass_vk.cc
+++ b/impeller/renderer/backend/vulkan/render_pass_vk.cc
@@ -29,8 +29,9 @@
 
 namespace impeller {
 
-static vk::AttachmentDescription foo(const Attachment& attachment,
-                                     bool resolve_texture = false) {
+static vk::AttachmentDescription CreateAttachmentDescription(
+    const Attachment& attachment,
+    bool resolve_texture = false) {
   const auto& texture =
       resolve_texture ? attachment.resolve_texture : attachment.texture;
   if (!texture) {
@@ -67,17 +68,17 @@ static vk::AttachmentDescription foo(const Attachment& attachment,
   );
 }
 
-static vk::AttachmentDescription CreateAttachmentDescription(
+static void SetTextureLayout(
     const Attachment& attachment,
+    const vk::AttachmentDescription& attachment_desc,
     const std::shared_ptr<CommandBufferVK>& command_buffer,
     bool resolve_texture = false) {
   const auto& texture =
       resolve_texture ? attachment.resolve_texture : attachment.texture;
   if (!texture) {
-    return {};
+    return;
   }
   const auto& texture_vk = TextureVK::Cast(*texture);
-  vk::AttachmentDescription attachment_desc = foo(attachment, resolve_texture);
 
   if (attachment_desc.initialLayout == vk::ImageLayout::eGeneral) {
     BarrierVK barrier;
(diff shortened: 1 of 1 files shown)


gaaclarke:
@jonahwilliams I wanted to land this groundwork for recycling alone to make sure I'm not missing anything.  It also cleans things up a bit anyway.

impeller/renderer/backend/vulkan/render_pass_vk.cc
```diff
@@ -31,10 +31,8 @@ namespace impeller {
 
 static vk::AttachmentDescription CreateAttachmentDescription(
     const Attachment& attachment,
-    const std::shared_ptr<CommandBufferVK>& command_buffer,
-    bool resolve_texture = false) {
-  const auto& texture =
-      resolve_texture ? attachment.resolve_texture : attachment.texture;
+    const std::shared_ptr<Texture> Attachment::*texture_ptr) {
```
jonahwilliams, line 34: What is this syntax?

gaaclarke: That's a pointer to a field.  It allows us to specify if we want to grab `Attachment::texture` or `Attachment::resolve_texture` instead of using an optional bool argument.  The pointers are const static.  Its easier to check out the calling site.

jonahwilliams (approved):
I think this looks right, so as long as its validating fine LGTM

gaaclarke:
I don't get validation layer errors when scrolling around in the gallery, let me know if you think there is something else we should test, like advanced blends?

jonahwilliams:
The issue I saw with barriers was scrolling through the pull quote on wonderous would occassionally show incorrect tiling.

jonahwilliams:
But I'm trying to do the math in my head here. I think this looks right.

I don't really understand why we need to split this into two functions, I would probably implement the recycling in the part of this method that actually constructs the object description and leave the side effect heavy wrapper?

gaaclarke:
Wonderous looks good.  The logic will eventually look something like this:

```c++
std::optional<vk::RenderPass> render_pass = cache.Fetch(vk_context, render_target_);
if (!render_pass) {
  render_pass = CreateVKRenderPass(vk_context, render_target_);
  cache.add(vk_context, render_target_, render_pass.value());
}
SetTextureLayouts(render_target_, command_buffer);
```

PR #45090, opened 2023-08-25
[Impeller] Simplify color source + text with new Skia API.

jonahwilliams:
The original implementation of arbitrary color source + text was quite complicated due to the fact that I had to use blending to emulate this functionality. Now that we can get paths from a given text data, I can instead make non solid color color sources and text just use the regular geometry + color source pipelines.

Fixes https://github.com/flutter/flutter/issues/132928
Fixes https://github.com/flutter/flutter/issues/132972


chinmaygarde (approved):

bdero:
Can we merge https://github.com/flutter/engine/pull/45079 first to simplify the cherry-pick?

jonahwilliams:
We can't CP this change, it depends on an API that was added after the branch cut.

bdero:
I mean simplify the cherry-pick of https://github.com/flutter/engine/pull/45079

jonahwilliams:
Ahh yeah no rush!

PR #45093, opened 2023-08-25
Remove --enable-software-rendering from iOS scenario tests

zanderso:


chinmaygarde (approved):

zanderso:
@dnfield This PR turns off software rendering except for a bunch of the platform view tests, where neither Skia nor Impeller GPU backends work. It also corrects the usage of the `-skip-testing` flag to `xcodebuild`, and detects Impeller so that we can compare against different goldens.

PR #45095, opened 2023-08-25
Revert ios cpu changes

zanderso:
https://github.com/flutter/flutter/issues/133301


christopherfujino (approved):
LGTM

godofredoc:
We may want to hold landing the change the fix to change a single property in the config. Sending the fix in min.

zanderso:
@godofredoc Oh, the autosubmit got to this before I saw your message. We can verify that the fix works on the presubmit checks when this relands.

PR #45131, opened 2023-08-25
[Impeller] DlAiksCanvas as a DlCanvas wrapper for impeller::Canvas

dnfield:
This is a reland of https://github.com/flutter/engine/pull/44248

Fixes https://github.com/flutter/flutter/issues/132416

Changes from last time:

- The `drawPoints` benchmark was failing to render anything meaningful because of an error in `AiksLayer` that resulted in an infinitely sized bounding rectangle poisoning the bounds with `NaN` (this happens, for example, with a `drawPaint` call, which that benchmark happens to use). Added a test covering this and filed https://github.com/flutter/flutter/issues/132770 to explore ways to avoid this in the future.
- There was a bug in `DlAiksCanvas::SaveLayer` where a `nullptr` `paint` but non-`nullptr` `backdrop` was failing to actually save the layer. This resulted in incorrect rendering.
- There was a bug in `impeller::Canvas::DrawPicture` that resulted in incorrect stencil depth counting. That was fixed separately by @bdero, but was the cause of incorrect rendering in some Wonderous screens.
- I've added a simple implementation for `AiksLayer::IsReplacing`. It does not currently compare as deeply as the `DisplayListLayer` version potentially does, but it is good enough to avoid the regression noted in https://github.com/flutter/flutter/issues/132071. That regression was on a benchmark that greatly benefits from partial repaint. With the new implementation, it still gains partial repaint where it previously did not. There is more work that can be done here, filed https://github.com/flutter/flutter/issues/133361 to track that work.

I merged but did not fully integrate the `DisplayListBuilder`/`CanvasToReceiver` work that @flar has done. I have a local experiment with that, but would prefer to see this land and run through the device lab so we get some better comparison numbers for which one performs better.


jonahwilliams (approved):
LGTM

jonahwilliams:
Seems like there are some GN config issues with flatbuffers tho

impeller/aiks/canvas.cc
```diff
@@ -502,6 +504,19 @@ Picture Canvas::EndRecordingAsPicture() {
   Picture picture;
   picture.pass = std::move(base_pass_);
 
+  std::vector<SkRect> rects;
+  picture.pass->IterateAllEntities([&rects](const Entity& entity) {
```
jonahwilliams, line 508: This is showing up as being pretty hot. This might need to be fixed to see a larger difference.

jonahwilliams: Perhaps we could append to the Rtree instead of temp storage? I think that is how it was designed to work.

flow/layers/aiks_layer.cc
```diff
@@ -0,0 +1,67 @@
...
+#include <utility>
+
+namespace flutter {
+
+AiksLayer::AiksLayer(const SkPoint& offset,
+                     const std::shared_ptr<const impeller::Picture>& picture)
+    : offset_(offset), picture_(picture) {
+#if IMPELLER_SUPPORTS_RENDERING
+  if (picture_) {
+    auto bounds = picture_->pass->GetElementsCoverage(std::nullopt)
```
jonahwilliams, line 16: Thie one is also really slow.

jonahwilliams: Looks like the TextFrame bounds are slow, at least in ComplexLayout (This is UI thread)

image

jonahwilliams:
I suspect you'll see fewer improvements on this than you'd like due to the naïve coverage implementation. Lets talk next week about a strategy to make this faster.

jonahwilliams:
We also still seem to be spending time populating raster cache entries?

image

impeller/aiks/canvas.cc
dnfield: The current RTree impl expects to get an array of `SkRect` objects. We could probably refactor it to make it append.

But this should be happening on the UI thread now.

flow/layers/aiks_layer.cc
dnfield: We end up needing to get these multiple times per frame AFAICT, we should probably cache them once we've calculated them.

dnfield:
> We also still seem to be spending time populating raster cache entries?
> 
> image

Let's fix that in a different patch. Looks like flow is missing a raster cache enabled check for this.

flow/layers/aiks_layer.cc
jonahwilliams: See also: https://github.com/flutter/flutter/issues/133388 . We should just cache the value we get from Skia. They're computing the bounds anyway.

flar: Not just the text bounds - why isn't the picture itself remembering its own bounds? These will get used frame after frame and we have to iterate the elements each time to save ... 16 bytes in the object?

impeller/aiks/canvas.cc
flar: How would that help anything? The rects have to be accumulated at some point, you are just pushing it from your own code into code that you don't own - it won't actually save anything.

You still have to do the iteration no matter how you create the RTree. The bounds still have to be accumulated into a tree. If Rtree is going to accumulate the bounds for you before it makes them into a tree then accumulation is still happening, it's just not in "your" code. If you want RTree to rebuild the tree on every append then not only will it have to still accumulate the bounds, but you will also have to add rebuilding the tree overhead on each rect.

Meanwhile, the dl_aiks_canvas was already creating an RTree accumulator that does *all of this for you* plus it will give you the bounds at the end and then the class wasn't even using it...

dnfield: We could just ask the rtree for the bounds.

flar: Correct, but I though this discussion was about creating the RTree, not using it to get the bounds...?

flow/layers/aiks_layer.cc
dnfield: This is where we could just ask the rtree for the bounds :)

flar: Does that code always produce an rtree? Then, yes, it's a very cheap way to get the bounds. We had been making that mistake in DL for a long time until I fixed it recently in the "split DL builder" PR.

PR #45137, opened 2023-08-26
[Impeller] always present drawable on main thread.

jonahwilliams:
WIP

### DIFF
--- a/impeller/renderer/backend/metal/surface_mtl.mm
+++ b/impeller/renderer/backend/metal/surface_mtl.mm
@@ -236,9 +236,9 @@ static void PresentToMainThread(id<MTLCommandBuffer> command_buffer,
 #ifdef FML_OS_IOS
   dispatch_async(dispatch_get_main_queue(), ^{
     TRACE_EVENT0("flutter", "PresentToMainThread");
-    [CATransaction begin];
     [command_buffer commit];
     [command_buffer waitUntilScheduled];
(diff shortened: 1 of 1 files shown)


luckysmg:
Tested this patch in profile mode.

1. I see some jitter when scrolling the ScrollView (Maybe due to the dispatch to main thread to present...)
2. The raster thread blocked by nextDrawable. I think it is due to the drawable doesn't present successfully (Although we call `drawable.present`, but seems that drawable is pending and doesn't present on screen instantaneously, so nextDrawable will be blocked). From the test, the nextDrawable is blocked by about 1s

luckysmg:
Also, I'm not sure we should do this or no. Because move present to main thread seems to be a big change. If currently some work is on main thread, the present will be blocked and delay, also will impact next frame to get drawable....

Or we should just do this for iOS 16.6? Because on iOS17 all things is ok without any changes
(means `presentsWithTransacation = ![NSThread isMainThread]` and present on raster thread.....) I think maybe apple do something bug in IOS16.6 SDK and fixed on iOS17....

jonahwilliams:
Yeah I agree this sort of change is extremely risky, and the blocking of main thread workloads is probably unavoidable. The safest change is probably to go back to conditionally presenting with transaction - however I'm not sure if we can conditionally detect this based on versions - because this will still have the issue of potentially triggering CATransaction assertions.

PR #45148, opened 2023-08-26
Use `SkTextBlob::bounds` instead of `TextFrame::getBounds()`.

matanlurey:
Closes https://github.com/flutter/flutter/issues/133157.


jonahwilliams (approved):
LGTM

If our text frame bounds computation is wrong, that is going to bubble up to more places too.
Can we either 1) make impeller::TextFrame use the bounds data from the SkTextFrame 2) fix the impeller::TextFrame computation?

matanlurey:
I'm still learning the words, so I'm not sure what difference between a `TextBlob` and a `TextFrame` is, but I'm not sure the `TextFrame` bounds are _wrong_, just potentially not appropriate in this case? Note that `blob->bounds()` is probably not the same thing as `frame->bounds()`?

jonahwilliams:
The impeller::TextFrame is the analog of the SkTextBlob

jonahwilliams:
(Not necessarily in this PR, FWIW)

jonahwilliams:
Actually in general TextFrame::GetBounds is super hot, so we might want to take the SkiaBounds in the constructor and use them cached, see https://github.com/flutter/engine/pull/45131#discussion_r1306202240

matanlurey:
Ah, sure! Filed https://github.com/flutter/flutter/issues/133388.

PR #45150, opened 2023-08-26
[Impeller] Cache Skia text bounds computation.

jonahwilliams:
Fixes https://github.com/flutter/flutter/issues/133388

Just use the TextBounds that Skia gives us. These are slightly larger but already computed. If the larger bounds are a problem we can go back to Impeller computed bounds, but using a cached computation.

### Skia
flutter_02

### Impeller
flutter_03


jonahwilliams:
Its easier for this to land after https://github.com/flutter/engine/pull/45090

impeller/typographer/backends/skia/text_frame_skia.cc
```diff
@@ -57,12 +56,6 @@ TextFrame MakeTextFrameFromTextBlobSkia(const sk_sp<SkTextBlob>& blob) {
     const auto glyph_count = run.glyphCount();
     const auto* glyphs = run.glyphs();
     switch (run.positioning()) {
-      case SkTextBlobRunIterator::kDefault_Positioning:
-        FML_DLOG(ERROR) << "Unimplemented.";
-        break;
-      case SkTextBlobRunIterator::kHorizontal_Positioning:
-        FML_DLOG(ERROR) << "Unimplemented.";
-        break;
```
jonahwilliams, line 65: Since we're never going to use these, I deleted them and left it for the default.

impeller/typographer/backends/skia/text_frame_skia.cc
```diff
@@ -75,32 +68,32 @@ TextFrame MakeTextFrameFromTextBlobSkia(const sk_sp<SkTextBlob>& blob) {
         font.setSize(kScaleSize);
         font.getBounds(glyphs, glyph_count, glyph_bounds.data(), nullptr);
 
+        std::vector<TextRun::GlyphPosition> positions;
+        positions.reserve(glyph_count);
```
jonahwilliams, line 72: Allocate the right amount of storage once rather than appending.

impeller/typographer/text_frame.cc
```diff
@@ -8,30 +8,13 @@ namespace impeller {
 
 TextFrame::TextFrame() = default;
```
jonahwilliams, line 9: The default constructor is no glyphs, empty bounds, no color. Which should be perfectly safe.

impeller/typographer/text_run.cc
```diff
@@ -37,8 +44,4 @@ const Font& TextRun::GetFont() const {
   return font_;
 }
 
-bool TextRun::HasColor() const {
```
jonahwilliams, line 40: We don't need to track this per run.

impeller/typographer/backends/skia/text_frame_skia.h
```diff
@@ -10,6 +10,7 @@
 
 namespace impeller {
 
-TextFrame MakeTextFrameFromTextBlobSkia(const sk_sp<SkTextBlob>& blob);
+std::optional<TextFrame> MakeTextFrameFromTextBlobSkia(
```
jonahwilliams, line 13: I made this optional so that we can avoid creating TextContents at all if the textblob is nullptr for some reason.

impeller/typographer/backends/skia/text_frame_skia.cc
matanlurey: Put this as a comment in the code instead of in the PR review

impeller/typographer/backends/skia/text_frame_skia.h
matanlurey: Ditto above, comment in the code :)

matanlurey (approved):
LGTM. Thanks!

impeller/typographer/backends/skia/text_frame_skia.cc
jonahwilliams: feels like

```
/// the int
int a = 0;
```

matanlurey: 🤷🏼. My point is if you think it's important enough to call out in the review, it's probably better as a comment :)

Defer to you, definitely a micro-nit.

impeller/typographer/backends/skia/text_frame_skia.cc
dnfield: I have a slight preference for removing the default so that we know if Skia adds some other value to this that we could be handling.

dnfield (approved):

PR #45154, opened 2023-08-27
A minimal `engine_tools_lib` to use for local-repo Dart tooling

matanlurey:
Partial work towards re-landing #44936.

Both the `clang_tidy` and `githooks` passage could benefit from being able to automatically find the latest `compile_commands.json` output, which means that some common code should exist in the `tools/` directory.

This is a very minimal (but tested) library for doing exactly that.


tools/engine_tools_lib/lib/engine_tools_lib.dart
```diff
@@ -0,0 +1,184 @@
...
+    // Check for the existence of a `flutter` directory within `src`.
+    final io.Directory flutterDir = io.Directory(p.join(srcPath, 'flutter'));
+    if (!flutterDir.existsSync()) {
+      throw ArgumentError(
+        'The path "$srcPath" does not contain a "flutter" directory.'
+      );
+    }
+
+    // Check for the existence of a `out` directory within `src`.
+    final io.Directory outDir = io.Directory(p.join(srcPath, 'out'));
```
zanderso, line 73: Could this be optional? `outputs()` below could return an empty list, and `latestOutput()` could return `null`.

zanderso (commented):
I like this idea, but wondering if we shouldn't put this under a new subdirectory like `tools/pkg` (`pkg` to mirror subdirectories containing several Dart packages). Also wondering about the name. What do you think of `engine_repo_tools`?

tools/engine_tools_lib/lib/engine_tools_lib.dart
matanlurey: Yeah, very fair. Updated.

tools/engine_tools_lib/lib/engine_tools_lib.dart
```diff
@@ -0,0 +1,184 @@
...
+
+    if (!maybeSrcDir.existsSync()) {
+      throw ArgumentError(
+        'The path "$path" does not exist or is not a directory.'
+      );
+    }
+
+    do {
+      if (p.basename(maybeSrcDir.path) == 'src') {
+        return Engine.fromSrcPath(maybeSrcDir.path);
```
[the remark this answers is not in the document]
matanlurey: Great idea, updated.

matanlurey:
> I like this idea, but wondering if we shouldn't put this under a new subdirectory like `tools/pkg` (`pkg` to mirror subdirectories containing several Dart packages). Also wondering about the name. What do you think of `engine_repo_tools`?

Works for me! Moved and renamed.

tools/pkg/engine_repo_tools/lib/engine_repo_tools.dart
```diff
@@ -0,0 +1,228 @@
...
+final class Engine {
+  /// Creates an [Engine] from a path such as `/Users/.../flutter/engine/src`.
+  ///
+  /// ```dart
+  /// final Engine engine = Engine.findWithin('/Users/.../engine/src');
+  /// print(engine.srcDir.path); // /Users/.../engine/src
+  /// ```
+  ///
+  /// Throws a [InvalidEngineException] if the path is not a valid engine root.
+  factory Engine.fromSrcPath(String srcPath) {
```
zanderso, line 46: Our conversation from lunch reminded me that passing a `FileSystem` from `package:file` here and to `findWithin` will let this library and its clients test with a `MemoryFileSystem` rather than the real thing.

matanlurey: Yeah that's fair.

Here is a question though, is that particularly useful? For example, where this would be used (`clang_tidy`, `githooks`) there isn't a real way to make them work on a virtual file system - which means we'd be adding the abstraction to `Engine` for ... engine to test itself?

Dunno, seems sus - I'm not against it but I'd prefer to add it when it provides value?

zanderso: Sure, we can add it later if we need it.

zanderso (approved):

PR #45182, opened 2023-08-28
[Impeller] its not safe to presentWithTransaction from a background thread.

jonahwilliams:
Sad Trombone. Because Flutter runs from a background thread, if we touch a CATransaction while there is a main thread transaction occuring then we'll trigger a framework error.

Fixes https://github.com/flutter/flutter/issues/133147

### DIFF
--- a/shell/platform/darwin/ios/ios_surface_metal_impeller.mm
+++ b/shell/platform/darwin/ios/ios_surface_metal_impeller.mm
@@ -58,8 +58,11 @@ GPUCAMetalLayerHandle IOSSurfaceMetalImpeller::GetCAMetalLayer(const SkISize& fr
 
   // When there are platform views in the scene, the drawable needs to be presented in the same
   // transaction as the one created for platform views. When the drawable are being presented from
-  // the raster thread, participate with the implicit CA transaction.
-  layer.presentsWithTransaction = YES;
+  // the raster thread, we may not be able to use a transaction as it will dirty the UIViews being
+  // presented. If there is a non-Flutter UIView active, such as in add2app or a
+  // presentViewController page transition, then this will cause CoreAnimation assertion errors and
+  // exit the app.
+  layer.presentsWithTransaction = [[NSThread currentThread] isMainThread];
 
   return layer;
 }


jonahwilliams:
There are a number of alternative solutions to this, all of which are fairly risky, so instead we should start with the safe option.

shell/platform/darwin/ios/ios_surface_metal_impeller.mm
```diff
@@ -58,8 +58,11 @@
 
   // When there are platform views in the scene, the drawable needs to be presented in the same
   // transaction as the one created for platform views. When the drawable are being presented from
-  // the raster thread, participate with the implicit CA transaction.
-  layer.presentsWithTransaction = YES;
+  // the raster thread, we may not be able to use a transaction as it will dirty the UIViews being
+  // presented. If there is a non-Flutter UIView active, such as in add2app or a
+  // presentViewController page transition, then this will cause CoreAnimation assertion errors and
+  // exit the app.
+  layer.presentsWithTransaction = [[NSThread currentThread] isMainThread];
```
dnfield, line 65: Wont' this re-introduce the bug we cherry picked a fix for into 3.10?

jonahwilliams: Yup

dnfield: Is there any way we could add some tests in this? Explicitly start a transaction on the main thread, then try to commit one on the background?

jonahwilliams: We can add tests in the framework repo. The assertion error cannot be triggered on Simulators IIRC

jonahwilliams:
The actual error is:

> [LayoutConstraints] Unsupported layout off the main thread for UITextEffectsWindow with no associated or ancestor view controller
This application is modifying the autolayout engine from a background thread after the engine was accessed from the main thread. This can lead to engine corruption and weird crashes.

If you trigger animations on the main thread like pushing a new view controller with an animation (what the video player does). when we set presentsWithTransaction, that implicitly creates a CATransaction for the background thread (though we can see the same effect by manually creating the transaction). When we [present drawable], that marks the UIViews up tot the root dirty, which triggers a layout, and then triggers the layout off main thread error.

luckysmg:
I think this LGTM.

Also, find some old context maybe can be some informations.
https://github.com/flutter/flutter/issues/66647
https://github.com/flutter/flutter/issues/66647#issuecomment-701053003

jonahwilliams:
Going to land this now while we work on figuring out a better fix going forward.

PR #45187, opened 2023-08-28
[skwasm] encode PNGs using browser APIs

eyebrowsoffire:
This allows us to remove libpng from skia entirely, which saves us about 25kb brotli compressed.

Note, this should not change any functionality. The existing functionality is covered by the unit tests here: https://github.com/flutter/engine/blame/bd2132a0814f6df95471e97f7d5efaca515506bd/lib/web_ui/test/ui/image_golden_test.dart#L197-L197

### DIFF
--- a/lib/web_ui/lib/src/engine/skwasm/skwasm_impl/image.dart
+++ b/lib/web_ui/lib/src/engine/skwasm/skwasm_impl/image.dart
@@ -72,6 +72,11 @@ class SkwasmImage extends SkwasmObjectWrapper<RawImage> implements ui.Image {
       context.transferFromImageBitmap(bitmap);
       final DomBlob blob = await offscreenCanvas.convertToBlob();
       final JSArrayBuffer arrayBuffer = (await blob.arrayBuffer().toDart)! as JSArrayBuffer;
+
+      // Zero out the contents of the canvas so that resources can be reclaimed
+      // by the browser.
+      offscreenCanvas.width = 0;
+      offscreenCanvas.height = 0;
       return ByteData.view(arrayBuffer.toDart);
     } else {
       return (renderer as SkwasmRenderer).surface.rasterizeImage(this, format);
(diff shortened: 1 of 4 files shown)


lib/web_ui/lib/src/engine/skwasm/skwasm_impl/image.dart
```diff
@@ -55,8 +56,26 @@ class SkwasmImage extends SkwasmObjectWrapper<RawImage> implements ui.Image {
...
+        await (renderer as SkwasmRenderer).surface.renderPicture(
+          recorder.endRecording() as SkwasmPicture,
+        );
+      final DomOffscreenCanvas offscreenCanvas =
+        createDomOffscreenCanvas(bitmap.width.toDartInt, bitmap.height.toDartInt);
+      final DomCanvasRenderingContextBitmapRenderer context =
+        offscreenCanvas.getContext('bitmaprenderer')! as DomCanvasRenderingContextBitmapRenderer;
+      context.transferFromImageBitmap(bitmap);
+      final DomBlob blob = await offscreenCanvas.convertToBlob();
+      final JSArrayBuffer arrayBuffer = (await blob.arrayBuffer().toDart)! as JSArrayBuffer;
```
yjbanov, line 74: Once we have the bytes of the image, should we set `offscreenCanvas.height/width` to zero so the browser can eagerly reclaim the bitmap memory? Otherwise, the browser may have to wait for the GC cycle to kick in. In the past, Safari would run out of memory if we didn't zero out the canvas, and return `null` when creating a new canvas.

yjbanov (approved):

Hixie:

lib/web_ui/lib/src/engine/skwasm/skwasm_impl/image.dart
eyebrowsoffire: Good idea. I'll do that

PR #45196, opened 2023-08-28
Rename `instrumentation` to `stopwatch`

matanlurey:
... since that's what the code is/does.

Partial work towards https://github.com/flutter/flutter/issues/126009.


flow/stopwatch_unittests.cc
```diff
@@ -2,8 +2,8 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
-#include "flutter/flow/instrumentation.h"
-#include "gmock/gmock.h"
+#include "flutter/flow/stopwatch.h"
+#include "gmock/gmock.h"  // IWYU pragma: keep
```
jonahwilliams, line 6: @dnfield was the issue that IWYU didnt understand all of the gn options and so tests didn't work well with it?

Anyway, unsure if we want to start adding IWYU pragmas until we understand how many we'd need

jonahwilliams (approved):
LGTM w/ question

flow/stopwatch_unittests.cc
matanlurey: Ack. This was more of a "hmm, what's going on here".

From my understanding, gmock specifically needs a pragma (I see the Chromium codebase doing the same thing).

I'm going to merge, but if we end up wanting to remove this I have no qualms.

jonahwilliams: its its only for gmock and only for a few places I think thats fine. I just dont want IWYU pragmas on half of our includes 😄

matanlurey: Ack. I'll run that experiment later this week to see how bad it would be.

matanlurey:
Merging ignoring `flutter-gold` as this doesn't impact goldens anyway.

flow/stopwatch_unittests.cc
dnfield: I do not think we're using IWYU , are we?

jonahwilliams: I was asking about the previous issue with trying to use IWYU that you mentioned.

PR #45199, opened 2023-08-29
Fix scene view canvas/platform view placement.

eyebrowsoffire:
I wasn't adjusting the `left` and `top` css styles based on the device pixel ratio. This fixes that and also sets up unit tests for `SceneView`.


lib/web_ui/lib/src/engine/scene_view.dart
```diff
@@ -163,11 +163,13 @@ final class PictureSliceContainer extends SliceContainer {
       final DomCSSStyleDeclaration style = canvas.style;
       final double logicalWidth = roundedOutBounds.width / window.devicePixelRatio;
```
yjbanov, line 164: Do we lose the sub-pixel delta between the rounded values and the actual values? I wonder of this is why we get https://github.com/flutter/flutter/issues/122541 in CanvasKit, and maybe we have the same issue in Skwasm?

yjbanov (approved):

lib/web_ui/lib/src/engine/scene_view.dart
eyebrowsoffire: When I first implemented the platform views change (when it was a draft), I actually did have a ton of golden diffs because I had blurring due to what you're talking about. The rounding out of the rectangle fixed it, and got rid of the extra blurring and the goldens were consistent with previous behavior.

That being said, I am not 100% sure the existing behavior is totally correct. This change at least gets us back to the previous rendering behavior though. I think I will follow up and look a little deeper into this, just to make sure that we aren't ending up with any fractional weirdness here.

pedromassango:
Does this fixes the resize blur issue https://github.com/flutter/flutter/issues/122541?

eyebrowsoffire:
> Does this fixes the resize blur issue [flutter/flutter#122541](https://github.com/flutter/flutter/issues/122541)?

It does not, this change only addresses the experimental skwasm renderer, not canvaskit.

PR #45211, opened 2023-08-29
Lazily allocate RasterCacheItems only when caching is enabled

flar:
Fixes https://github.com/flutter/flutter/issues/133377

The default allocation of RasterCacheItems in the layer tree was showing up in profiles of apps running on Impeller which don't actually use the raster cache.

In order to eliminate the overhead of those allocations, RasterCacheItems are now lazily allocated only when the layers encounter an actual raster_cache during the Preroll phase.

### DIFF
--- a/flow/layers/cacheable_layer.cc
+++ b/flow/layers/cacheable_layer.cc
@@ -6,31 +6,39 @@
 
 namespace flutter {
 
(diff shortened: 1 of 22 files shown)


dnfield (approved):
LGTM, thank you!

PR #45232, opened 2023-08-29
[Impeller] Reland DlAiksCanvas

dnfield:
Relands https://github.com/flutter/engine/pull/45131

Fixes https://github.com/flutter/flutter/issues/132416

Differences from last time:

- Some minor merge conflict fixes
- Use the RTree to get the bounds instead of recalculating the bounds
- Make the iOS platform view controller implementation use the impeller-aware slices instead of the display list ones. This has been fixed for Android and the desktop embedding, but I missed iOS. The unit tests weren't actually running before I branched for my PR, @zanderso fixed them up separately and this resulted in catching the failures on post submit last time.


jonahwilliams (approved):
Lets get this moving!

dnfield:
Failures are due to a missing compensating change after merging, will fix...

chinmaygarde (approved):

flow/layers/aiks_layer.cc
```diff
@@ -0,0 +1,64 @@
...
+  context->PushTransform(SkMatrix::Translate(offset_.x(), offset_.y()));
+  context->AddLayerBounds(bounds_);
+  context->SetLayerPaintRegion(this, context->CurrentSubtreeRegion());
+}
+
+void AiksLayer::Preroll(PrerollContext* frame) {
+  // There is no opacity peepholing to do here because Impeller handles that
+  // in the Entities, and this layer will never participate in raster caching.
+  FML_DCHECK(!frame->raster_cache);
+  set_paint_bounds(bounds_);
```
flar, line 39: Missing the offset here. Notice that DLLayer translates the `bounds_` field.

flow/layers/aiks_layer.cc
```diff
@@ -0,0 +1,64 @@
...
+
+void AiksLayer::Paint(PaintContext& context) const {
+  FML_DCHECK(needs_painting(context));
+
+  auto mutator = context.state_stack.save();
+  mutator.translate(offset_.x(), offset_.y());
+
+  FML_DCHECK(!context.raster_cache);
+
+  SkScalar opacity = context.state_stack.outstanding_opacity();
```
flar, line 50: You never set the flags to indicate that you could render your contents with opacity, so the state stack will do a protective saveLayer around you instead and this value will always be 1.0.

If AiksCanvas.DrawImpellerPicture can distribute opacity, then you will need to set the appropriate flag in Preroll to be able to do this.

flow/surface_frame.cc
```diff
@@ -31,9 +32,13 @@ SurfaceFrame::SurfaceFrame(sk_sp<SkSurface> surface,
...
     FML_DCHECK(!frame_size.isEmpty());
-    dl_builder_ =
-        sk_make_sp<DisplayListBuilder>(SkRect::Make(frame_size), true);
-    canvas_ = dl_builder_.get();
+#if IMPELLER_SUPPORTS_RENDERING
+    aiks_canvas_ =
+        std::make_shared<impeller::DlAiksCanvas>(SkRect::Make(frame_size));
+    canvas_ = aiks_canvas_.get();
+#else
+    FML_DCHECK(false);
```
flar, line 40: The flag is kind of misnamed now. It really means "we're rendering in Impeller mode, so do the Impeller thing". It was named after the (former) solution rather than the trigger that caused it...

impeller/display_list/dl_aiks_canvas.cc
```diff
@@ -0,0 +1,1204 @@
...
+    accumulator_ = std::make_unique<flutter::RTreeBoundsAccumulator>();
+  }
+}
+
+DlAiksCanvas::DlAiksCanvas(const SkIRect& cull_rect, bool prepare_rtree)
+    : DlAiksCanvas(SkRect::Make(cull_rect), prepare_rtree) {}
+
+DlAiksCanvas::~DlAiksCanvas() = default;
+
+static Paint::Style ToStyle(flutter::DlDrawStyle style) {
```
flar, line 61: For many of these, if they are enum classes with the same ordering then you can get away with a cast like dl_sk_conversions.h - though that class has unit tests to verify that the conversions produce the right values so that we'll catch if there is ever a mismatch in the enum values.

impeller/display_list/dl_aiks_canvas.cc
```diff
@@ -0,0 +1,1204 @@
...
+}
+
+// |flutter::DlCanvas|
+void DlAiksCanvas::ClipRect(const SkRect& rect, ClipOp clip_op, bool is_aa) {
+  canvas_.ClipRect(skia_conversions::ToRect(rect), ToClipOperation(clip_op));
+}
+
+// |flutter::DlCanvas|
+void DlAiksCanvas::ClipRRect(const SkRRect& rrect, ClipOp clip_op, bool is_aa) {
+  if (rrect.isSimple()) {
```
flar, line 802: Simple allows different radx and rady. When it says "equal radii" it means all 4 corners are equal, not that the horizontal and vertical radii are the same.

Also, it has an isRect() that could be trapped and converted into a ClipRect()

impeller/display_list/dl_aiks_canvas.cc
```diff
@@ -0,0 +1,1204 @@
...
+void DlAiksCanvas::DrawCircle(const SkPoint& center,
+                              SkScalar radius,
+                              const flutter::DlPaint& paint) {
+  canvas_.DrawCircle(skia_conversions::ToPoint(center), radius, ToPaint(paint));
+}
+
+// |flutter::DlCanvas|
+void DlAiksCanvas::DrawRRect(const SkRRect& rrect,
+                             const flutter::DlPaint& paint) {
+  if (rrect.isSimple()) {
```
flar, line 873: isSimple != horizontal and vertical radii are the same.

impeller/display_list/dl_aiks_canvas.cc
```diff
@@ -0,0 +1,1204 @@
...
+}
+
+// |flutter::DlCanvas|
+void DlAiksCanvas::DrawImpellerPicture(
+    const std::shared_ptr<const impeller::Picture>& picture,
+    SkScalar opacity) {
+  if (!picture) {
+    return;
+  }
+  FML_DCHECK(opacity == 1);
```
flar, line 1061: Might want to DCHECK this in aiks_layer as well.

flow/layers/aiks_layer.cc
dnfield: It can't

impeller/display_list/dl_aiks_canvas.cc
dnfield: This was copy pasted from dl_dispatcher

flow/layers/aiks_layer.cc
dnfield: What's a test case I could use for this?

flow/layers/aiks_layer.cc
dnfield: (but we could certainly update that later, or see about using the new split you created to handle that right?)

flow/layers/aiks_layer.cc
flar: opacity layer under an offset layer

impeller/display_list/dl_aiks_canvas.cc
flar: Filed: https://github.com/flutter/flutter/issues/133793

PR #45243, opened 2023-08-29
Add a build_bucket_golden_scraper tool.

matanlurey:
As discussed offline, this is best deleted when Skia-gold is used for all of our engine tests.

However, this will be useful for unblocking some PRs until then :)

See README.md for details!


tools/build_bucket_golden_scraper/lib/build_bucket_golden_scraper.dart
```diff
@@ -0,0 +1,205 @@
...
+    }
+    if (maybeUri.hasScheme) {
+      file = await _downloadFile(maybeUri);
+    } else {
+      file = io.File(pathOrUrl);
+    }
+
+    // Read the file as a string.
+    final String contents = await file.readAsString();
+
```
zanderso, line 111: Can the file and temp directory be cleaned up after this point? I guess if you read the contents in `_downloadFile()` you can do all the cleanup there and make sure the temp dir gets cleaned up if the http client throws an exception.

matanlurey: I can do you one better, why am I creating a temp file at all? :P

matanlurey:
Gentle ping @zanderso.

Not urgent to merge, but trying to avoid this getting stale since I want to start using `engine_repo_tools` in the other Dart tooling (i.e `clang_tidy` and friends).

zanderso (approved):

PR #45252, opened 2023-08-29
[Impeller] Fixes stroke path geometry that can draw outside of path if the path ends at sharp turn.

chunhtai:
Fixes https://github.com/flutter/flutter/issues/133032

The issue is that when we draw a line we always try to draw a complete rectangle from point 1 to point2 before drawing the join and the next line
Untitled drawing (7)

This becomes a problem if there is a sharp turn
Untitled drawing (6)

Untitled drawing (5)

Notice there will be a slight over drawing at the bottom.

The solution then is to not draw the rect first before drawing the join. It will be the join's responsibility to draw the complete the rect from the line start to the join.

This should also save a lot of repeatedly drawing during the join
rect

## Pre-launch Checklist

- [ ] I signed the [CLA].

### DIFF
--- a/impeller/entity/geometry/stroke_path_geometry.cc
+++ b/impeller/entity/geometry/stroke_path_geometry.cc
@@ -253,7 +253,7 @@ StrokePathGeometry::CreateSolidStrokeVertices(
   for (size_t contour_i = 0; contour_i < polyline.contours.size();
        contour_i++) {
     auto contour = polyline.contours[contour_i];
-    size_t contour_section = 0;
+    size_t contour_component_i = 0;
     size_t contour_start_point_i, contour_end_point_i;
     std::tie(contour_start_point_i, contour_end_point_i) =
         polyline.GetContourPointBounds(contour_i);
@@ -309,11 +309,11 @@ StrokePathGeometry::CreateSolidStrokeVertices(
     // Generate contour geometry.
     for (size_t point_i = contour_start_point_i + 1;
          point_i < contour_end_point_i; point_i++) {
-      if ((contour_section + 1 >= contour.sections.size()) &&
-          contour.sections[contour_section + 1].section_start_index <=
+      if ((contour_component_i + 1 >= contour.components.size()) &&
+          contour.components[contour_component_i + 1].component_start_index <=
               point_i) {
-        // The point_i has entered the next section in this contour.
-        contour_section += 1;
+        // The point_i has entered the next component in this contour.
+        contour_component_i += 1;
       }
       // Generate line rect.
       vtx.position = polyline.points[point_i - 1] + offset;
@@ -323,9 +323,9 @@ StrokePathGeometry::CreateSolidStrokeVertices(
 
       auto is_end_of_contour = point_i == contour_end_point_i - 1;
 
-      if (!contour.sections[contour_section].is_curve) {
-        /// Add two end points of the line rect to draw a solid
-        /// straight line.
+      if (!contour.components[contour_component_i].is_curve) {
+        // For line components, two additional points need to be appended prior
+        // to appending a join connecting the next component.
         vtx.position = polyline.points[point_i] + offset;
         vtx_builder.AppendVertex(vtx);
         vtx.position = polyline.points[point_i] - offset;
@@ -338,9 +338,8 @@ StrokePathGeometry::CreateSolidStrokeVertices(
                     offset, scaled_miter_limit, scale);
         }
       } else {
-        // Curve sections don't require the two end points of the line rects
-        // assuming the angles between of a continous two points are close
-        // enough.
+        // For curve components, the polyline is detailed enough such that
+        // it can avoid worrying about joins altogether.
         if (!is_end_of_contour) {
           compute_offset(point_i + 1);
         } else {
--- a/impeller/entity/geometry/stroke_path_geometry.cc
+++ b/impeller/entity/geometry/stroke_path_geometry.cc
@@ -356,7 +356,7 @@ StrokePathGeometry::CreateSolidStrokeVertices(
     if (!contour.is_closed) {
       auto cap_offset =
           Vector2(-contour.end_direction.y, contour.end_direction.x) *
-          stroke_width * 0.5; // Clockwise normal
+          stroke_width * 0.5;  // Clockwise normal
       cap_proc(vtx_builder, polyline.points[contour_end_point_i - 1],
                cap_offset, scale, false);
     } else {
--- a/impeller/entity/geometry/stroke_path_geometry.cc
+++ b/impeller/entity/geometry/stroke_path_geometry.cc
@@ -309,7 +309,9 @@ StrokePathGeometry::CreateSolidStrokeVertices(
     // Generate contour geometry.
     for (size_t point_i = contour_start_point_i + 1;
          point_i < contour_end_point_i; point_i++) {
-      if ((contour_section + 1 >= contour.sections.size()) && contour.sections[contour_section + 1].section_start_index <= point_i) {
+      if ((contour_section + 1 >= contour.sections.size()) &&
+          contour.sections[contour_section + 1].section_start_index <=
+              point_i) {
         // The point_i has entered the next section in this contour.
         contour_section += 1;
       }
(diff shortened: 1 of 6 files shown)


impeller/entity/geometry/stroke_path_geometry.cc
```diff
@@ -308,15 +308,11 @@ StrokePathGeometry::CreateSolidStrokeVertices(
     // Generate contour geometry.
     for (size_t point_i = contour_start_point_i + 1;
          point_i < contour_end_point_i; point_i++) {
-      // Generate line rect.
+      // Generate vertical line to fill the stroke width at point_i - 1
       vtx.position = polyline.points[point_i - 1] + offset;
       vtx_builder.AppendVertex(vtx);
       vtx.position = polyline.points[point_i - 1] - offset;
       vtx_builder.AppendVertex(vtx);
-      vtx.position = polyline.points[point_i] + offset;
```
chunhtai, line 316: Reasoning https://github.com/flutter/flutter/issues/133032#issuecomment-1698114342

chunhtai:
oops looks like the golden has failed

chunhtai:
The circle golden is missing a tiny tiny slide of pixels which i think should be fine. 

@bdero This is ready for review now

chinmaygarde (approved):
LGTM but @bdero is more familiar with this code and may have thoughts.

Thanks so much!

bdero (changes requested):
This patch introduces regressions with most strokes (visible in EntityTest.StrokeCapAndJoinTest).

bdero:
https://github.com/flutter/engine/assets/919017/2776c7b9-9f93-4dc2-afa6-f1f7f81b2155

chunhtai:
@bdero will take a look. Should we include all the playground test into goldens?

chunhtai:
I think I know the problem is, so the point order matters. for a complete retangle 
```
tl -------tr
|               |
bl-------br
```
the point needs to be provided in either of this order
tl, bl, tr, br
or
bl, tl, br, tr

bdero:
We should add a new golden that draws some stroke paths similar to this in `aiks_unittests.cc` to make this easier catch.

BTW, to help debug, you can also use the Z key when viewing a playground. 
It looks like there's also a bug in the miter joint logic:

https://github.com/flutter/engine/assets/919017/33e1c870-856a-4173-bf6e-a1b243335fa2

bdero:
I think the thesis of this patch is not really correct/the full picture? I wrote a Skia fiddle to render the given problem stroke and Skia renders it the way that we do:

https://fiddle.skia.org/c/0c12e8b2a46224c97b84bb6a0519d580

image

bdero:
Looking at the original issue, I suspect the problem is related to how we're managing the forward direction/normals for caps: https://github.com/flutter/flutter/issues/133032#issuecomment-1707243488

chunhtai:
I did the frame capturing, I did see the small overdraw of the original issue was due to the line I removed in https://github.com/flutter/engine/pull/45252#discussion_r1310829933 (there are at least 3 or 4 points before the end of the line that draws outside of the end cap)

It doesn't seem to related to cap drawing?

bdero:
Talked about this offline a bit, summarizing here:

I believe the root problem is that we always use the same normal for the start and end of the line rects. This works great for line contours, but is not ideal for cubics/quadratics, especially for tight curves.

For cubic/quadratic contours specifically, we need to skip the last two points when drawing polyline segments.

A couple of possible approaches for this:

1. The "perfect" approach: `impeller::PolylineContour` can track a `bool is_curve` member that is populated when generating the polyline, and whenever it's true, assume the angle between the two points if small enough such that we don't need to append the last two points for each polyline rectangle segment.
2. Imperfect quick n' dirty approach: if the next line segment normal `n` is very close to the angle of the current line segment normal `c`, then don't draw the 3rd/4th point of the polyline segment rectangle. `c.Dot(n) > 1 - small_number`.

(Although I think approach 2 is probably more trouble than it's worth TBH :))

chunhtai:
The circle golden has just a tiny difference, I assume it is ok

impeller/entity/geometry/stroke_path_geometry.cc
```diff
@@ -308,15 +309,39 @@ StrokePathGeometry::CreateSolidStrokeVertices(
...
-      vtx_builder.AppendVertex(vtx);
-      vtx.position = polyline.points[point_i] - offset;
-      vtx_builder.AppendVertex(vtx);
+
+      // Curve sections don't require the two end points of the line rects
+      // assuming the angles between of a continous two points are close enough.
```
bdero, line 325:
```suggestion
      // For line components, two additional points need to be appended prior
      // to appending a join connecting the next component. But for curve
      // components, we assume that the polyline is detailed enough such that
      // we can avoid worrying about joins altogether.
```

impeller/entity/geometry/stroke_path_geometry.cc
```diff
@@ -328,10 +353,10 @@ StrokePathGeometry::CreateSolidStrokeVertices(
...
     // Generate end cap or join.
-    if (!polyline.contours[contour_i].is_closed) {
+    if (!contour.is_closed) {
       auto cap_offset =
           Vector2(-contour.end_direction.y, contour.end_direction.x) *
-          stroke_width * 0.5;  // Clockwise normal
+          stroke_width * 0.5;
       cap_proc(vtx_builder, polyline.points[contour_end_point_i - 1],
                cap_offset, scale, false);
     } else {
```
bdero: Now that we know when we're in a curve or not, we should also be able to get away with removing joins (not calling `join_proc` right below) for curves.

impeller/geometry/path_component.h
```diff
@@ -47,9 +47,13 @@ struct LinearPathComponent {
   std::optional<Vector2> GetEndDirection() const;
 };
 
+// A component represets a Quadratic Bézier curve.
```
bdero, line 50:
```suggestion
// A component that represents a Quadratic Bézier curve.
```

impeller/geometry/path_component.h
```diff
@@ -87,10 +91,15 @@ struct QuadraticPathComponent {
   std::optional<Vector2> GetEndDirection() const;
 };
 
+// A component represets a Cubic Bézier curve.
```
bdero, line 94:
```suggestion
// A component that represents a Cubic Bézier curve.
```

bdero (commented):
I'm seeing issues with curves. It appears that all the triangles are there, but the offset is winding up 0 length sometimes.

https://github.com/flutter/engine/assets/919017/76354fb5-27e1-438b-97e6-1a63b261aed2

impeller/entity/geometry/stroke_path_geometry.cc
```diff
@@ -308,15 +309,39 @@ StrokePathGeometry::CreateSolidStrokeVertices(
...
+      // This also prevents overdrawing the line rect if the contour ends at a
+      // sharp curve with thick stroke width.
+      if (!contour.sections[contour_section].is_curve) {
+        vtx.position = polyline.points[point_i] + offset;
+        vtx_builder.AppendVertex(vtx);
+        vtx.position = polyline.points[point_i] - offset;
+        vtx_builder.AppendVertex(vtx);
+      } else if (point_i == contour_end_point_i - 1) {
+        // If this is a curve and is the end of the contour, two end points
+        // need to be drawn with the contour end_direction.
```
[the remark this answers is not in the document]
chunhtai: This is in case for contour is_closed, the join assume the two endpoints are drawn

chunhtai:
I also added a new golden test that similar to the cubic component playground

bdero (commented):
`EntityTest.StrokeCapAndJoinTest` still doesn't look right. This is the line I use to build and run it:

```
autoninja -C ../out/host_debug_unopt/ -j 1000 flutter/impeller:impeller_unittests && ../out/host_debug_unopt/impeller_unittests --timeout=0 --gtest_filter="Play/EntityTest.StrokeCapAndJoinTest/Metal" --enable_playground
```

Screenshot 2023-09-13 at 10 11 29 AM

bdero:
Ah hold on one sec, I think I screwed up the checkout locally.

bdero (approved):
LGTM! Thanks a lot for your work on this! :)

bdero:
The goldens are looking good and I marked them as positive.

impeller/entity/geometry/stroke_path_geometry.cc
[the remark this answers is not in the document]
bdero: Ah right, that makes sense.

PR #45255, opened 2023-08-29
[Impeller] transform text path offsets so color sources match expected offsets for gradients.

jonahwilliams:
Gradient offsets are computed based on path offsets + gradient positions, so we need to shift the points instead of just adding a concat to the canvas. Fixes https://github.com/flutter/flutter/issues/132972

### Skia

flutter_03

### Impeller (No Patch)

flutter_04

### Impeller (W/ Patch)

flutter_05

### DIFF
--- a/impeller/geometry/path.cc
+++ b/impeller/geometry/path.cc
@@ -66,23 +66,22 @@ void Path::Shift(Point shift) {
   for (const auto& component : components_) {
     switch (component.type) {
       case ComponentType::kLinear:
-        linears_[component.index].p1 = linears_[component.index].p1 + shift;
-        linears_[component.index].p2 = linears_[component.index].p2 + shift;
+        linears_[component.index].p1 += shift;
+        linears_[component.index].p2 += shift;
         break;
       case ComponentType::kQuadratic:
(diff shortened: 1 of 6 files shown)


impeller/geometry/path.cc
```diff
@@ -61,6 +61,34 @@ void Path::SetConvexity(Convexity value) {
   convexity_ = value;
 }
 
+void Path::Shift(Point shift) {
+  size_t currentIndex = 0;
+  for (const auto& component : components_) {
+    switch (component.type) {
+      case ComponentType::kLinear:
+        linears_[component.index].p1 = linears_[component.index].p1 + shift;
```
matanlurey, line 69: I don't know C++.

That being said, can you write:

```cc
liners[component.index].p1 += shift
```

...? or does that not work with vectors/arrays?

If not, at least pull `index` into `auto const index = component.index` :)

matanlurey (approved):

impeller/geometry/path.cc
jonahwilliams: Done

jonahwilliams: Fun fact, += is a different operator than + (see point.h)

PR #45256, opened 2023-08-29
Add an API in `ui_web` to create a `ui.Image` from an `ImageBitmap`

eyebrowsoffire:
This API will help with situations in which the user has a browser resource that they want to transform into a `ui.Image` and render directly into the layer tree. Most browser resources can be converted to an `ImageBitmap` via the `createImageBitmap` API.

### DIFF
--- a/lib/web_ui/lib/src/engine/canvaskit/canvaskit_api.dart
+++ b/lib/web_ui/lib/src/engine/canvaskit/canvaskit_api.dart
@@ -219,14 +219,31 @@ extension CanvasKitExtension on CanvasKit {
   ) => _MakeImage(info, pixels.toJS, bytesPerRow.toJS);
 
   @JS('MakeLazyImageFromTextureSource')
-  external SkImage? _MakeLazyImageFromTextureSource(
+  external SkImage? _MakeLazyImageFromTextureSource1(
     JSAny src,
     SkPartialImageInfo info,
   );
-  SkImage? MakeLazyImageFromTextureSource(
+
+  @JS('MakeLazyImageFromTextureSource')
+  external SkImage? _MakeLazyImageFromTextureSource2(
+    JSAny src,
+    JSNumber zeroSecondArgument,
+    JSBoolean srcIsPremultiplied,
+  );
+
+  SkImage? MakeLazyImageFromTextureSourceWithInfo(
     Object src,
     SkPartialImageInfo info,
-  ) => _MakeLazyImageFromTextureSource(src.toJSAnyShallow, info);
+  ) => _MakeLazyImageFromTextureSource1(src.toJSAnyShallow, info);
+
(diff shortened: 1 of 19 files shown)


dnfield:
svg_image_bitmap_fragment_shader_sampler looks like it has artifacts to me, but I'm not sure if that's just how that test is supposed to work.

eyebrowsoffire:
> svg_image_bitmap_fragment_shader_sampler looks like it has artifacts to me, but I'm not sure if that's just how that test is supposed to work.

That's actually the expected output of the shader. It's the "glitch" shader lifted from the `next_gen_ui` demo here https://github.com/flutter/samples/tree/main/next_gen_ui_demo

lib/web_ui/lib/src/engine/canvaskit/canvaskit_api.dart
```diff
@@ -219,14 +219,31 @@ extension CanvasKitExtension on CanvasKit {
   ) => _MakeImage(info, pixels.toJS, bytesPerRow.toJS);
 
   @JS('MakeLazyImageFromTextureSource')
-  external SkImage? _MakeLazyImageFromTextureSource(
+  external SkImage? _MakeLazyImageFromTextureSource1(
```
yjbanov, line 222: Not sure if there's value in reusing the Dart SDK convention, but there the convention is for the number to refer to the argument count, so in this case it would be "2" and "3", instead of "1" and "2".

lib/web_ui/lib/ui_web/src/ui_web/images.dart
```diff
@@ -25,3 +28,18 @@ Future<ui.Codec> createImageCodecFromUrl(
     chunkCallback: chunkCallback,
   );
 }
+
+/// Creates a [ui.Image] from an ImageBitmap object.
```
yjbanov, line 32: nit: dartdoc likes it when there's one empty line between the first and the second paragraph (or between paragraphs in general).

lib/web_ui/lib/ui_web/src/ui_web/images.dart
```diff
@@ -25,3 +28,18 @@ Future<ui.Codec> createImageCodecFromUrl(
...
+/// The contents of the ImageBitmap must have a premultiplied alpha.
+/// The engine will take ownership of the ImageBitmap object and consume its
+/// contents.
+///
+/// See https://developer.mozilla.org/en-US/docs/Web/API/ImageBitmap
+FutureOr<ui.Image> createImageFromImageBitmap(JSAny imageSource) {
+  if (!domInstanceOfString(imageSource, 'ImageBitmap')) {
+    throw Exception('Image source $imageSource is not an ImageBitmap!');
```
yjbanov, line 40: This should probably be a `ArgumentError`.

nit: s/!/./

yjbanov (approved):

PR #45259, opened 2023-08-30
Add a `DlStopwatchVisualizer` and conditionally use it for Impeller

matanlurey:
Closes https://github.com/flutter/flutter/issues/126009.

One major change, as-per @jonahwilliams's feedback, is that I created a helper class called `DlVertexPainter`, which provides a `DrawRect`-like API, but just builds a buffer of vertices, resulting into just a single:

```cc
// Actually draw.
canvas->DrawVertices(painter.IntoVertices(), DlBlendMode::kSrc, paint);
```

Also, added test for it, since there is no way to screenshot test to Impeller overlay yet.

# Impeller

Screenshot 2023-08-30 at 2 08 51 PM


flow/stopwatch_dl.cc
```diff
@@ -0,0 +1,123 @@
...
+    paint.setColor(0x99FFFFFF);
+    canvas->DrawRect(rect, paint);
+  }
+
+  // Prepare a path for the data; we start at the height of the last point so
+  // it looks like we wrap around.
+  {
+    SkPath path;
+    path.setIsVolatile(true);
+    path.moveTo(x, rect.bottom());
```
jonahwilliams, line 44: Rather than using a path, which we'll end up tessellating per frame, I would create a vertices object and add the rectangles to this.

Here is an example of turing a bunch of rectangles into a single vertices call: https://github.com/gskinnerTeam/flutter-wonderous-app/blob/d369891d1eef36b2a7dbefc1566dd61c644890ca/lib/ui/screens/artifact/artifact_search/time_range_selector/time_range_painter.dart#L16-L46

The line can also be approximated with a skinny rectangle. 

The goal of this is to have the most minimally invasive rendering for the performance monitor.

flow/stopwatch_dl.cc
```diff
@@ -0,0 +1,123 @@
...
+    path.lineTo(rect.right(), rect.bottom());
+    path.close();
+
+    paint.setColor(0xAA0000FF);
+    canvas->DrawPath(path, paint);
+  }
+
+  // Draw horizontal frame markers.
+  {
+    paint.setStrokeWidth(0);
```
jonahwilliams, line 78: Same here, use skinny rectangles all packed into the same vertices instead of a path/stroke.

flow/stopwatch_dl.cc
matanlurey: Thanks, this was really interesting!

flow/stopwatch_dl.cc
matanlurey: Ditto!

jonahwilliams (commented):
Its not as much of a help if you do a DlVertices per rectangle. instead you should pack all of your rectangles into a single vertices.

matanlurey:
> if you do a DlVertices per rectangle. instead you should pack all of your rectangles into a single vertices

I think you might have seen an earlier commit (I did that just to make sure I was triangle-ing correctly, before refactoring into a single buffer). PTAL again?

flow/stopwatch_dl.cc
```diff
@@ -0,0 +1,151 @@
...
+
+std::shared_ptr<DlVertices> DlVertexPainter::IntoVertices() {
+  auto const result = DlVertices::Make(
+      /*mode=*/DlVertexMode::kTriangles,
+      /*vertex_count=*/vertices_.size(),
+      /*vertices=*/vertices_.data(),
+      /*texture_coordinates=*/nullptr,
+      /*colors=*/colors_.data());
+  vertices_.clear();
+  colors_.clear();
```
jonahwilliams, line 147: So I'm not sure if this is actually valud. Does the DlVertices constructor copy the data out, or is it being stored by reference? If you clear the data here won't the vertices buffers end up filled with junk?

jonahwilliams: Looks like it does copy: https://github.com/flutter/engine/blob/main/display_list/dl_vertices.cc#L39

So this is fine!

matanlurey: A good question.

I did write a unit test verifying that it doesn't _appear_ to be junk. My understanding is it's probably copied (it's just `SkPoint` and `DlColor`), but maybe I shouldn't expect that to always be true.

I could either:

1. Explore moving the data (which probably is better given the name `IntoVertices()`?)
2. Copy the data explicitly before passing it
3. Do nothing, i.e. just mark with a boolean that this class is no longer valid and `FML_DCHECK`.

I'd prefer to do 1 > 2 > 3, wdut though?

flow/stopwatch_dl.cc
```diff
@@ -0,0 +1,148 @@
...
+        x + width * (static_cast<double>(stopwatch_.GetCurrentSample()) /
+                     kMaxSamples);
+    auto const t = y;
+    auto const r = l + width * sample_unit_width;
+    auto const b = rect.bottom();
+    painter.DrawRect(SkRect::MakeLTRB(l, t, r, b), color);
+  }
+
+  // Actually draw.
+  canvas->DrawVertices(painter.IntoVertices(), DlBlendMode::kSrc, paint);
```
jonahwilliams, line 104: This is a bit odd - it looks like the colors you're using have some opacity on them. But drawing with blend mode src with blow away whatever is "under" the vertices even with opacity. I'd think you'd want to either draw 1) fully opaque and blend mode src or 2) draw partially opaque and use the default blend mode srcOver.

flow/stopwatch_dl.cc
jonahwilliams: I don't think any action is necessary.

matanlurey: Ack, thanks for the discussion though!

flow/stopwatch_dl.cc
matanlurey: It sounds like `srcOver` will be more understandable to other folks, so I'll go with that and make a note.

flow/stopwatch_dl.cc
```diff
@@ -0,0 +1,149 @@
...
+                     kMaxSamples);
+    auto const t = y;
+    auto const r = l + width * sample_unit_width;
+    auto const b = rect.bottom();
+    painter.DrawRect(SkRect::MakeLTRB(l, t, r, b), color);
+  }
+
+  // Actually draw.
+  // Note we use kSrcOut, because some of the colors above have opacity < 1.0.
+  canvas->DrawVertices(painter.IntoVertices(), DlBlendMode::kSrcOut, paint);
```
jonahwilliams, line 105: srcOver :)

jonahwilliams (approved):
LGTM

PR #45271, opened 2023-08-30
Adds an --rbe option to tools/gn that works on Linux hosts

zanderso:
This PR lands changes to `tools/gn` to use Flutter's RBE instance instead of Goma. It is initially supported only for builds from a Linux host.

1. Authenticate with `gcloud`
```
engine/src$ ./buildtools/linux-x64/gcloud/bin/gcloud auth application-default login --disable-quota-project
```

2. GN
```
engine/src$ ./flutter/tools/gn --android --android-cpu arm64 --no-lto --runtime-mode debug --rbe
```

3. Build
```
engine/src$ ninja -C out/android_debug_arm64 -j200
```

### DIFF
--- a/tools/gn
+++ b/tools/gn
@@ -214,8 +214,69 @@ def buildtools_dir():
   return '%s-%s' % (host_os, host_cpu)
 
 
+def setup_rbe(args):
+  rbe_gn_args = {}
+  # RBE is default-off. If it is not asked for, then silently keep all default
+  # flag values.
+  if not args.rbe:
+    return rbe_gn_args
+
+  if get_host_os() not in ['linux']:
+    print(
+        'The --rbe flag has no effect. RBE is currently only supported on Linux.'
+    )
+    return rbe_gn_args
+
+  rbe_gn_args['use_rbe'] = True
+
+  # When running in CI, the recipes use their own rbe install, and take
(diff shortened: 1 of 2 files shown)


tools/gn
```diff
@@ -214,8 +214,69 @@ def buildtools_dir():
...
+    reproxy_path = os.path.join(cipd_reclient_dir, 'reproxy')
+    rbe_cfg_path = os.path.join(SRC_ROOT, 'build', 'rbe.cfg')
+    bootstrap_cmd = [
+        bootstrap_path,
+        '--re_proxy=' + reproxy_path,
+        '--automatic_auth=true',
+        '--cfg=' + rbe_cfg_path,
+    ]
+    try:
+      subprocess.call(bootstrap_cmd, cwd=SRC_ROOT)
```
keyonghan, line 254: From LUCI side, we expect to have a cleanup/shutdown step after task finishes. You may want to mention somewhere that users are expected to cleanup/shutdown the bootstrap later.

keyonghan (approved):
LGTM with nit.

tools/gn
zanderso: There isn't a good spot to put that in the local dev workflow, and the command is too long to ask folks to run by hand, so that will need some more thought.

PR #45295, opened 2023-08-30
[web] De-singletonize MouseCursor for multi-view

mdebbar:
`MouseCursor` is a singleton that works by accessing `flutterViewEmbedder` directly. After this PR, `MouseCursor` becomes a non-singleton that takes a `FlutterView` and controls the mouse cursor for said view.

### DIFF
--- a/lib/web_ui/lib/src/engine/window.dart
+++ b/lib/web_ui/lib/src/engine/window.dart
@@ -16,6 +16,8 @@ import 'package:ui/ui_web/src/ui_web.dart' as ui_web;
 import '../engine.dart' show DimensionsProvider, registerHotRestartListener, renderer;
 import 'display.dart';
 import 'dom.dart';
+import 'embedder.dart';
+import 'mouse_cursor.dart';
 import 'navigation/history.dart';
 import 'platform_dispatcher.dart';
 import 'services.dart';
@@ -29,8 +31,17 @@ const bool debugPrintPlatformMessages = false;
 /// The view ID for the implicit flutter view provided by the platform.
 const int kImplicitViewId = 0;
 
+/// Represents all views in the Flutter Web Engine.
+///
+/// In addition to everything defined in [ui.FlutterView], this class adds
+/// a few web-specific properties.
+abstract class EngineFlutterView extends ui.FlutterView {
+  MouseCursor get mouseCursor;
+  DomElement get rootElement;
+}
+
 /// The Web implementation of [ui.SingletonFlutterWindow].
-class EngineFlutterWindow extends ui.SingletonFlutterWindow {
+class EngineFlutterWindow extends ui.SingletonFlutterWindow implements EngineFlutterView {
   EngineFlutterWindow(this.viewId, this.platformDispatcher) {
     platformDispatcher.viewData[viewId] = this;
     platformDispatcher.windowConfigurations[viewId] = const ViewConfiguration();
@@ -52,6 +63,12 @@ class EngineFlutterWindow extends ui.SingletonFlutterWindow {
 
   @override
   final EnginePlatformDispatcher platformDispatcher;
+
+  @override
+  late final MouseCursor mouseCursor = MouseCursor(this);
+
+  @override
+  DomElement get rootElement => flutterViewEmbedder.flutterViewElement;
 
   /// Handles the browser history integration to allow users to use the back
   /// button, etc.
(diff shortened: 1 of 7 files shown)


ditman (commented):
Minor nits to simplify this a little bit.

lib/web_ui/lib/src/engine/window.dart
```diff
@@ -53,6 +64,12 @@ class EngineFlutterWindow extends ui.SingletonFlutterWindow {
   @override
   final EnginePlatformDispatcher platformDispatcher;
 
+  @override
+  late final MouseCursor mouseCursor = MouseCursor(this);
```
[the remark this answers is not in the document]
mdebbar: That's a good idea!

lib/web_ui/lib/src/engine/window.dart
```diff
@@ -29,8 +31,17 @@ const bool debugPrintPlatformMessages = false;
 /// The view ID for the implicit flutter view provided by the platform.
 const int kImplicitViewId = 0;
 
+/// Represents all views in the Flutter Web Engine.
+///
+/// In addition to everything defined in [ui.FlutterView], this class adds
+/// a few web-specific properties.
+abstract class EngineFlutterView extends ui.FlutterView {
```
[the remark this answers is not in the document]
mdebbar: My thought here was that `EngineFlutterView` will become the base for all window and non-window views. Other non-window views will directly extend `EngineFlutterView` since they don't need `ui.SingletonFlutterWindow`.

So the hierarchy becomes:
```
                         ui.FlutterView
               (dart:ui interface across platforms)
                                |
                                |
                       EngineFlutterView (new)
           (web-specific additions to the interface)
                /                               \
               /                                 \
     NonWindowFlutterView (future)          EngineFlutterWindow
(non-sintleton, non-window views)        (legacy singleton window)

```

> Also, I think this should be `abstract interface class EngineFlutterView`, if we can use that [feature of Dart](https://dart.dev/language/class-modifiers#interface).

We can! We already use it in `platform_location.dart` :)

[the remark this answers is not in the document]
ditman: Ahhh, I gotcha. The advantage of the current version:

`ui.View -> ui.Window -> WebWindow implements WebOnlyAccessors`

Is that we're only extending the base classes defined by the core of the engine. In my mind that means that if there's an upcoming change in the base structure of classes in package:ui, it should be easier for `ui` to do changes to the base class, and for us to adapt to those changes?

(Of course, it depends on the changes, so our class extensions might not matter too much!)

lib/web_ui/lib/src/engine/mouse/cursor.dart
```diff
@@ -61,15 +51,12 @@ class MouseCursor {
...
     return _kindToCssValueMap[kind] ?? 'default';
   }
 
   void activateSystemCursor(String? kind) {
-    setElementStyle(
-      flutterViewEmbedder.flutterViewElement,
-      'cursor',
-      _mapKindToCssValue(kind),
-    );
+    element.style.cursor = _mapKindToCssValue(kind);
```
ditman, line 60: The only problem I can imagine with this is that if the `element` gets detached/deleted from the DOM, and then a system cursor event message comes, this might attempt to operate on a null instance. Since this is speculation and not a bug that we've seen so far, I say we don't worry about it for now.

lib/web_ui/lib/src/engine/window.dart
```diff
@@ -29,8 +31,17 @@ const bool debugPrintPlatformMessages = false;
 /// The view ID for the implicit flutter view provided by the platform.
 const int kImplicitViewId = 0;
 
+/// Represents all views in the Flutter Web Engine.
```
ditman, line 34: I think that the class that represents all views in the Flutter Web Engine is `EngineFlutterWindow` (that's the type that was used to define the `implicitView`, for example), right?

ditman (approved):

lib/web_ui/lib/src/engine/mouse/cursor.dart
mdebbar: That's not a problem at all! The element is non-nullable and cannot become null. Yes, it can be detached/removed from the DOM tree, but it is still a valid element. In fact, in `cursor_test.dart`, we are using detached elements to test `MouseCursor`.

lib/web_ui/lib/src/engine/window.dart
mdebbar: `EngineFlutterWindow` is a misnomer. It should be called `EngineSingletonFlutterWindow` (it actually extends `ui.SingletonFlutterWindow`). Views in the multi-view world shouldn't inherit from this singleton class. They should inherit directly from `ui.FlutterView` (and implement `EngineFlutterView` to satisfy all the web-specific needs).

mdebbar: Here's a follow up PR to rename this class to make it clear that it's a singleton: https://github.com/flutter/engine/pull/45981

PR #45298, opened 2023-08-30
[Impeller] Started tracking the pool with the command buffer.

gaaclarke:
issue: https://github.com/flutter/flutter/issues/133506

This fixes one of the issues seen in https://github.com/flutter/flutter/issues/133506.  There were multiple ones though.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/impeller/renderer/backend/vulkan/command_encoder_vk.cc
+++ b/impeller/renderer/backend/vulkan/command_encoder_vk.cc
@@ -34,13 +34,7 @@ class TrackedObjectsVK {
     if (!buffer_) {
       return;
     }
-    auto pool = pool_.lock();
-    if (!pool) {
-      // The buffer can not be freed if its command pool has been destroyed.
-      buffer_.release();
-      return;
-    }
-    pool->CollectGraphicsCommandBuffer(std::move(buffer_));
+    pool_->CollectGraphicsCommandBuffer(std::move(buffer_));
   }
 
   bool IsValid() const { return is_valid_; }
@@ -86,7 +80,7 @@ class TrackedObjectsVK {
 
  private:
   DescriptorPoolVK desc_pool_;
-  std::weak_ptr<CommandPoolVK> pool_;
+  std::shared_ptr<CommandPoolVK> pool_;
   vk::UniqueCommandBuffer buffer_;
   std::set<std::shared_ptr<SharedObjectVK>> tracked_objects_;
   std::set<std::shared_ptr<const Buffer>> tracked_buffers_;
--- a/impeller/renderer/backend/vulkan/command_encoder_vk.cc
+++ b/impeller/renderer/backend/vulkan/command_encoder_vk.cc
@@ -80,6 +80,7 @@ class TrackedObjectsVK {
 
  private:
   DescriptorPoolVK desc_pool_;
+  // `shared_ptr` since command buffers have a link to the command pool.
   std::shared_ptr<CommandPoolVK> pool_;
   vk::UniqueCommandBuffer buffer_;
   std::set<std::shared_ptr<SharedObjectVK>> tracked_objects_;
--- a/impeller/renderer/backend/vulkan/command_encoder_vk_unittests.cc
+++ b/impeller/renderer/backend/vulkan/command_encoder_vk_unittests.cc
@@ -14,6 +14,8 @@ namespace impeller {
 namespace testing {
 
 TEST(CommandEncoderVKTest, DeleteEncoderAfterThreadDies) {
+  // Tests that when a CommandEncoderVK is deleted that it will clean up its
+  // command buffers before it cleans up its command pool.
   std::shared_ptr<std::vector<std::string>> called_functions;
   {
     auto context = CreateMockVulkanContext();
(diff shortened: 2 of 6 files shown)


impeller/renderer/backend/vulkan/command_encoder_vk.cc
```diff
@@ -86,7 +80,7 @@ class TrackedObjectsVK {
 
  private:
   DescriptorPoolVK desc_pool_;
-  std::weak_ptr<CommandPoolVK> pool_;
+  std::shared_ptr<CommandPoolVK> pool_;
```
gaaclarke, line 83: Our command buffers always have a link to the command pool so it is never safe to have a dangling reference to a command buffer who is referring to a potentially dead pool.  This makes that guarentee.

impeller/renderer/backend/vulkan/resource_manager_vk.h
```diff
@@ -71,11 +71,13 @@ class ResourceManagerVK
...
   using Reclaimables = std::vector<std::unique_ptr<ResourceVK>>;
 
-  std::thread waiter_;
   std::mutex reclaimables_mutex_;
   std::condition_variable reclaimables_cv_;
   Reclaimables reclaimables_;
   bool should_exit_ = false;
+  // This should be initialized last since it references the other instance
+  // variables.
+  std::thread waiter_;
```
gaaclarke, line 80: This is just a nasty race condition I was running into.

impeller/renderer/backend/vulkan/command_encoder_vk.cc
matanlurey: I would add this in a comment in the code (versus just here which will get lost to the sands of time)

matanlurey (commented):
Defer explicit approval to someone else since I'm just getting started here, but thanks!

impeller/renderer/backend/vulkan/command_encoder_vk.cc
gaaclarke: done

gaaclarke:
@chinmaygarde friendly ping

impeller/renderer/backend/vulkan/command_encoder_vk_unittests.cc
```diff
@@ -0,0 +1,76 @@
...
+#include "flutter/fml/synchronization/count_down_latch.h"
+#include "flutter/testing/testing.h"
+#include "impeller/renderer/backend/vulkan/command_encoder_vk.h"
+#include "impeller/renderer/backend/vulkan/fence_waiter_vk.h"
+#include "impeller/renderer/backend/vulkan/test/mock_vulkan.h"
+
+namespace impeller {
+namespace testing {
+
+TEST(CommandEncoderVKTest, DeleteEncoderAfterThreadDies) {
```
jonahwilliams, line 16: I have sort of mixed feelings about these tests, because its sort disjointed from how these classes are used. I would feel better if there was substantially more commentary on specifically what these tests are covering.

jonahwilliams: Or at least for one of them, and then subsequent tests could have less commentary.

jonahwilliams (commented):
So today the current implementation with a weak_ptr is safe because we actually never actually clean up a cmd pool, we just recycle the same one over and over again.

Of course, if we start trying to recycle the cmd pools this might become a problem, but that problem might be different?

if the cmd encoder has a strong reference to the cmd pool,  this won't cause retention cycles right? because the pool doesn't have a strong reference to the impeller::vk classes.

So thinking out loud I think this is safe

impeller/renderer/backend/vulkan/command_encoder_vk_unittests.cc
gaaclarke: Unit tests are supposed to be testing how a class can be used, not how they are supposed to be used.  Maybe it will help to understand why I implemented this fix.  The pool change is a fix for a real bug that I was running into when using the playgrounds test harness which Chinmay was trying to get working.  So, it may be true that this bug shouldn't present itself in production, it has demonstrably happened in our test harness and may present itself accidentally in production.

The second fix with the resourcemanager is definitely something that can manifest in production code, but was showing up regularly in these tests.

We typically don't annotate unit tests that are using the public api of a method.  I'm happy to add any comments you think will help.

jonahwilliams: I understand that, what I'm saying is that I'm having a hard time following the test and so I'd appreciate more commentary on how/why the particular test successfully exercises the assertions.

gaaclarke:
> So today the current implementation with a weak_ptr is safe because we actually never actually clean up a cmd pool, we just recycle the same one over and over again.

This was manifesting in the playgrounds test harness, not in production apps.  (mentioned above)

> Of course, if we start trying to recycle the cmd pools this might become a problem, but that problem might be different?
> 
> if the cmd encoder has a strong reference to the cmd pool, this won't cause retention cycles right? because the pool doesn't have a strong reference to the impeller::vk classes.

The other `shared_ptr` for the pool lives on the static thread_local storage, the pool doesn't have a strong reference to that so that shouldn't be a cycle.

> So thinking out loud I think this is safe

Yep, I believe so.

jonahwilliams (approved):
LGTM. Ask for comments is just a nit.

impeller/renderer/backend/vulkan/command_encoder_vk_unittests.cc
gaaclarke: I added a comment to the first test, expanded the second comment and removed one piece of vestigial code.

PR #45300, opened 2023-08-30
Enforce the rule of calling `FlutterView.Render`

dkwingsmt:
This PR enforces the rules as documented in `FlutterView.Render`, where calls in illegal situations should be ignored - but have never been enforced.

```
  /// This function must be called within the scope of the
  /// [PlatformDispatcher.onBeginFrame] or [PlatformDispatcher.onDrawFrame]
  /// callbacks being invoked.
  ///
  /// If this function is called a second time during a single
  /// [PlatformDispatcher.onBeginFrame]/[PlatformDispatcher.onDrawFrame]
  /// callback sequence or called outside the scope of those callbacks, the call
  /// will be ignored.
```

This rule is very important to implementing multi-view without having to introduce new APIs. However, currently these illegal calls are not ignored, and historically many tests (especially integration tests) were unknowingly running based on this fact. @goderbauer did great work by eliminating these cases in g3, and it's time for us to make sure these calls are ignored.

Most effort of this PR goes to unit testing the changes. Some part of `Shell::Create` is extracted into a static function to avoid duplicate code.

## Pre-launch Checklist

- [ ] I signed the [CLA].


lib/ui/fixtures/ui_test.dart
```diff
@@ -1076,3 +1076,39 @@ external void _callHook(
   Object? arg20,
   Object? arg21,
 ]);
+
+Scene CreateRedBoxScene(Size size) {
```
loic-sharma, line 1080:
```suggestion
Scene _createRedBoxScene(Size size) {
```

lib/ui/fixtures/ui_test.dart
```diff
@@ -1076,3 +1076,39 @@ external void _callHook(
...
+  canvas.drawRect(Rect.fromLTRB(0.0, 0.0, size.width, size.height), paint);
+  final Picture picture = baseRecorder.endRecording();
+  builder.addPicture(Offset(0.0, 0.0), picture);
+  builder.pop();
+  return builder.build();
+}
+
+@pragma('vm:entry-point')
+@pragma('vm:external-name', 'NotifyNative')
+external void notifyNative();
```
loic-sharma, line 1097: Is this used?

lib/ui/fixtures/ui_test.dart
```diff
@@ -1076,3 +1076,39 @@ external void _callHook(
...
+void incorrectDoubleRender() {
+  PlatformDispatcher.instance.onBeginFrame = (Duration value) {
+    PlatformDispatcher.instance.views.first.render(CreateRedBoxScene(Size(2, 2)));
+    PlatformDispatcher.instance.views.first.render(CreateRedBoxScene(Size(3, 3)));
+  };
+  PlatformDispatcher.instance.onDrawFrame = () {
+    PlatformDispatcher.instance.views.first.render(CreateRedBoxScene(Size(4, 4)));
+    PlatformDispatcher.instance.views.first.render(CreateRedBoxScene(Size(5, 5)));
+  };
+}
```
loic-sharma, line 1114: Please request a frame after registering the frame callbacks. `dart:ui` apps should not rely on the engine to schedule a frame.

dkwingsmt: I don't think it works in this case. A requested frame goes to the engine and the animator, which don't exist in this test case.

lib/ui/window/platform_configuration_unittests.cc
```diff
@@ -15,10 +15,166 @@
...
+          configuration.GetEntrypointLibrary(),       //
+          configuration.GetEntrypointArgs(),          //
+          configuration.TakeIsolateConfiguration());  //
+      ASSERT_TRUE(launch_success);
+      post_launch();
+    });
+  }
+
+  // Get the runtime controller.
+  RuntimeController& Controller() { return *runtime_controller_; }
```
loic-sharma, line 141:
```suggestion
  RuntimeController& controller() { return *runtime_controller_; }
```

lib/ui/window/platform_configuration_unittests.cc
```diff
@@ -332,5 +488,68 @@ TEST_F(PlatformConfigurationTest, SetDartPerformanceMode) {
...
 }
 
+TEST_F(PlatformConfigurationTest, OutOfScopeRenderCallsAreIgnored) {
+  Settings settings = CreateSettingsForFixture();
+  TaskRunners task_runners = GetTaskRunnersForFixture();
+
+  MockRuntimeDelegate client;
+  auto platform_message_handler =
+      std::make_shared<MockPlatformMessageHandler>();
+  EXPECT_CALL(client, GetPlatformMessageHandler())
```
loic-sharma, line 498:
```suggestion
  EXPECT_CALL(client, GetPlatformMessageHandler)
```

lib/ui/window/platform_configuration_unittests.cc
```diff
@@ -332,5 +488,68 @@ TEST_F(PlatformConfigurationTest, SetDartPerformanceMode) {
...
+      .WillOnce(Return(platform_message_handler));
+  // Render should not be called.
+  EXPECT_CALL(client, Render(_, _)).Times(Exactly(0));
```
loic-sharma, line 501:
```suggestion
  EXPECT_CALL(client, Render).Times(0);
```

lib/ui/window/platform_configuration_unittests.cc
```diff
@@ -332,5 +488,68 @@ TEST_F(PlatformConfigurationTest, SetDartPerformanceMode) {
...
+    runtime_controller_context->Controller().AddView(
+        kImplicitViewId, ViewportMetrics(
+                             /*pixel_ratio=*/1.0, /*width=*/20, /*height=*/20,
+                             /*touch_slop=*/2, /*display_id=*/0));
+  });
+
+  // Wait for the Dart main function to end.
+  fml::AutoResetWaitableEvent latch;
+  PostSync(task_runners.GetUITaskRunner(), [&]() { latch.Signal(); });
+  latch.Wait();
```
loic-sharma, line 518: Do you need a latch here? Could you post an empty callback and let `PostSync`'s latch do the waiting?

lib/ui/window/platform_configuration_unittests.cc
```diff
@@ -332,5 +488,68 @@ TEST_F(PlatformConfigurationTest, SetDartPerformanceMode) {
...
+                             /*touch_slop=*/2, /*display_id=*/0));
+  });
+
+  // Wait for the Dart main function to end, which means the callbacks have been
+  // registered.
+  fml::AutoResetWaitableEvent latch;
+  PostSync(task_runners.GetUITaskRunner(), [&]() { latch.Signal(); });
+  latch.Wait();
+
+  runtime_controller_context->Controller().BeginFrame(fml::TimePoint::Now(), 0);
```
loic-sharma, line 551: Should this test wait until `onDrawFrame` exits? Perhaps `onDrawFrame` should call `_finish` and this should wait for that?

lib/ui/window/platform_configuration_unittests.cc
```diff
@@ -332,5 +488,68 @@ TEST_F(PlatformConfigurationTest, SetDartPerformanceMode) {
...
+        kImplicitViewId, ViewportMetrics(
+                             /*pixel_ratio=*/1.0, /*width=*/20, /*height=*/20,
+                             /*touch_slop=*/2, /*display_id=*/0));
+  });
+
+  // Wait for the Dart main function to end, which means the callbacks have been
+  // registered.
+  fml::AutoResetWaitableEvent latch;
+  PostSync(task_runners.GetUITaskRunner(), [&]() { latch.Signal(); });
+  latch.Wait();
```
loic-sharma, line 549: This currently relies on UI task ordering. Could we have `incorrectImmediateRender` call `_signal` and then wait on that?

lib/ui/window/platform_configuration_unittests.cc
loic-sharma: On second thought, that's not necessary. `BeginFrame` calls the callbacks synchronously. Could you add a quick comment that this will call `onBeginFrame` and `onDrawFrame` synchronously?

lib/ui/fixtures/ui_test.dart
loic-sharma: Ahh right, gotcha. Could add a quick comment that the test calls these callbacks "directly" without using the full frame lifecycle?

lib/ui/window/platform_configuration_unittests.cc
dkwingsmt: I think this rule is rarely applied in the engine repo and is redeemed weird. We've been using the CamelCase for all methods, such as in https://github.com/flutter/engine/blob/main/shell/common/engine.h#L952C28-L952C37

loic-sharma: Hm it looks like the engine is inconsistent here, but it looks like `PlatformConfiguration` uses variable naming for accessors. For example:

https://github.com/flutter/engine/blob/494fd7fe85a30ce4dfc28058961bb71920f514cc/lib/ui/window/platform_configuration.h#L283

I would be stay consistent with `PlatformConfiguration`'s style given this is what this code tests.

lib/ui/window/platform_configuration_unittests.cc
dkwingsmt: Done. Thanks for this, it actually found a bug of my code.

lib/ui/window/platform_configuration_unittests.cc
dkwingsmt: But `PlatformConfiguration` also has `GetAssetManager` `GetFontCollection` `GetPersistentIsolateData`...

dkwingsmt: I've refactored the code to some methods with callbacks that provide the controller. This is mostly to ensure that the tasks are performed on the UI thread, but also voids this concern.

dkwingsmt:
@loic-sharma I should have addressed all your comments. Can you take another look?

lib/ui/window/platform_configuration_unittests.cc
```diff
@@ -332,5 +494,84 @@ TEST_F(PlatformConfigurationTest, SetDartPerformanceMode) {
...
+  Settings settings = CreateSettingsForFixture();
+  TaskRunners task_runners = GetTaskRunnersForFixture();
+
+  MockRuntimeDelegate client;
+  auto platform_message_handler =
+      std::make_shared<MockPlatformMessageHandler>();
+  EXPECT_CALL(client, GetPlatformMessageHandler)
+      .WillOnce(Return(platform_message_handler));
+  // Render should not be called.
+  EXPECT_CALL(client, Render).Times(Exactly(0));
```
loic-sharma, line 507: Flutter engine code base doesn't appear to use `Exactly`: https://github.com/search?q=repo%3Aflutter%2Fengine+%22Exactly%28%22&type=code

```suggestion
  EXPECT_CALL(client, Render).Times(0);
```

Please update other locations too

loic-sharma (approved):
Excellent work, LGTM!

chinmaygarde (approved):

lib/ui/window/platform_configuration_unittests.cc
dkwingsmt: Sorry I missed this comment. I've removed it. Didn't know that Exactly is the default (I thought the default was AtLeast. :)

dkwingsmt:
Status: I'm testing this PR on g3 and it seems there are still a few integrations broken. I'm investigating them.

dkwingsmt:
I forgot that engine changes wouldn't affected integrated tests without landing... let's merge and try.

dkwingsmt:
For the record: This PR has been relanded in several PRs:
* https://github.com/flutter/engine/pull/47062
* https://github.com/flutter/engine/pull/47095
* https://github.com/flutter/engine/pull/47323

PR #45302, opened 2023-08-30
fix: Mark pages as read-only on iOS to avoid code signing exceptions

eseidel:
Fixes https://github.com/shorebirdtech/shorebird/issues/1183.

Some versions of iOS were sometimes killing apps using Shorebird due to codesigning.  We were loading the Dart snapshot (unsigned) as executable (since Dart normally does) even though we never execute code from it on iOS.

This updates to use a version of or Dart fork which allows you to make Dart not mark those pages as executable, which seems to fix the issue.


eseidel:
Sorry, wrong repo.

PR #45313, opened 2023-08-31
[Impeller] EntityPass::Clone needs to clone harder

jonahwilliams:
Fixes https://github.com/flutter/flutter/issues/133731

### DIFF
--- a/impeller/entity/entity_pass.cc
+++ b/impeller/entity/entity_pass.cc
@@ -59,7 +59,7 @@ EntityPass::EntityPass() = default;
 
 EntityPass::~EntityPass() = default;
 
-void EntityPass::SetDelegate(std::unique_ptr<EntityPassDelegate> delegate) {
+void EntityPass::SetDelegate(std::shared_ptr<EntityPassDelegate> delegate) {
   if (!delegate) {
     return;
   }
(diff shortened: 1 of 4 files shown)


jonahwilliams:
These goldens look really weird...

dnfield (approved):
LGTM, would appreciate a review from @bdero as well

jonahwilliams:
I'm not sure if this is entirely right or if I'm missing something, goldens look very different locally...

jonahwilliams:
Goldens look like the pictures keep building up

impeller/entity/entity_pass.cc
```diff
@@ -1023,6 +1023,16 @@ std::unique_ptr<EntityPass> EntityPass::Clone() const {
 
   auto pass = std::make_unique<EntityPass>();
   pass->SetElements(std::move(new_elements));
+  pass->backdrop_filter_reads_from_pass_texture_ =
+      backdrop_filter_reads_from_pass_texture_;
+  pass->advanced_blend_reads_from_pass_texture_ =
+      advanced_blend_reads_from_pass_texture_;
```
bdero, line 1029: (These read-counts are what's important for `DrawPicture`.)

bdero (approved):
LGTM. Haven't checked the mentioned discrepancies w/ the local playgrounds.

jonahwilliams:
This still isn't right though, here is what the backdrop blur output is giving me on Vulkan:

image

impeller/entity/entity_pass.cc
jonahwilliams: You need the readcounts, the delegate, and the backdrop proc or wonderous will crash.

bdero: That shouldn't be the case if the cloned pass is being used for `DrawPicture` (the elements are copied inline and these values should get discarded).

bdero: For the delegate and backdrop proc, that is. The read counts are important.

jonahwilliams: I can confirm that wonderous does not work correctly without copying the backdrop prop and entity delegate. Maybe the entity delegate _doesn't_ need to be copied but it does need to be set to a non-default value.

Let me double check that.

bdero: Ah nevermind, we don't need these for the inline merge, but we definitely need to copy them for the recursive clone, otherwise Very Bad Things will happen: https://github.com/flutter/engine/blob/494fd7fe85a30ce4dfc28058961bb71920f514cc/impeller/entity/entity_pass.cc#L1015

PR #45333, opened 2023-08-31
[macOS] Link __availability_version_check

cbracken:
Links in an implementation of _availability_version_check on macOS.                
This is required due to an upstream compiler builtin (runtime) change              
that marked this function as weakly-linked via `__attribute__((weak import))`   
As such, no linking failure occurs when the function is unavailable at             
link time.                                                                         
                                                                                   
By providing an implementation, the linker picks up our implementation,            
which looks up symbol in question at runtime (via dlsym) on the first              
invocation, caches it for later invocations, then invokes it. This is,             
in fact, precisely what the original clang builtin implementation did.             
                                                                                   
Upstream clang change: https://reviews.llvm.org/D150397                            
                                                                                   
Issue: https://github.com/flutter/flutter/issues/133777 

## Pre-launch Checklist

- [X] I signed the [CLA].

### DIFF
--- a/shell/platform/darwin/common/BUILD.gn
+++ b/shell/platform/darwin/common/BUILD.gn
@@ -13,12 +13,37 @@ source_set("common") {
   cflags_objcc = flutter_cflags_objcc
 
   sources = [
-    "availability_version_check.cc",
     "buffer_conversions.h",
     "buffer_conversions.mm",
     "command_line.h",
     "command_line.mm",
   ]
+
+  deps = [
+    ":availability_version_check",
+    "//flutter/fml",
+  ]
+
+  public_configs = [ "//flutter:config" ]
+}
+
+# Provides an implementation for _availability_version_check.
+#
+# This is required due to an upstream compiler builtin (runtime) change
+# that marked this function as weakly-linked via __attribute__((weak import))
+# As such, no linking failure occurs when the function is unavailable at
+# link time. Since the symbol is no longer linked in, App Store review blocks
+# publishing due to relying on a private symbol. Instead we link in our own
+# implementation, which provides the exact implementation used in clang prior
+# to the upstream change.
+#
+# See: Upstream clang change: https://reviews.llvm.org/D150397
+# See: https://github.com/flutter/flutter/issues/133777
+source_set("availability_version_check") {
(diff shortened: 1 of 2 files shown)


zanderso (approved):
Thanks!

Hixie:
Is "Calling dlsym with RTLD_DEFAULT can be very slow as all images in the process are searched for the symbol." (the reason for the upstream change) a concern for Flutter?

shell/platform/darwin/common/BUILD.gn
```diff
@@ -13,13 +13,29 @@ source_set("common") {
...
+    ":availability_version_check",
+    "//flutter/fml",
+  ]
+
+  public_configs = [ "//flutter:config" ]
+}
+
+# Provides an implementation for _availability_version_check.
+#
+# See: https://github.com/flutter/flutter/issues/133777
```
Hixie, line 32: this source code is likely to outlive github, it would be ideal if we could inline the explanation.

cbracken: Sounds good; will do.

cbracken: Done - copied in the text from the commit log (which hopefully also outlives GitHub!).

cbracken:
> Is "Calling dlsym with RTLD_DEFAULT can be very slow as all images in the process are searched for the symbol." (the reason for the upstream change) a concern for Flutter?

This is a single lookup that we cache on first call and never do again so the impact is low. Agreed that ideally we wouldn't even do that one lookup, but it is identical to the behaviour that we had all along -- i.e. not a regression, though not an improvement.

An alternative might be to go back to dealing with two clang toolchains and use Apple's Xcode toolchain rather than our own.

zanderso:
> An alternative might be to go back to dealing with two clang toolchains and use Apple's Xcode toolchain rather than our own.

IMO, that approach would cause an unacceptable regression in CI build times.

Hixie:
Oh I'm definitely not saying we should have two toolchains. The implication of the upstream comment is that we could save some time on startup if we avoided the lookup. Hopefully we'll get that when we remove this workaround when Xcode updates (i.e. when everyone else does).

PR #45350, opened 2023-08-31
[Impeller] turned on validations for all debug builds

gaaclarke:
fixes https://github.com/flutter/flutter/issues/133794

This adds the following shared libraries to the flutter.jar for debug builds on arm64:
```
lib/arm64-v8a/libVkLayer_khronos_validation.so
lib/arm64-v8a/libc++_shared.so
```

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/shell/platform/android/BUILD.gn
+++ b/shell/platform/android/BUILD.gn
@@ -490,9 +490,7 @@ action("android_jar") {
     ":pom_libflutter",
   ]
 
-  if (enable_vulkan_validation_layers ||
-      (impeller_enable_vulkan && current_cpu == "arm64" &&
-       flutter_runtime_mode == "debug")) {
+  if (enable_vulkan_validation_layers) {
     assert(impeller_enable_vulkan)
     apilevel26_toolchain = "//build/toolchain/android:clang_arm64_apilevel26"
     validation_layer_target =
--- a/shell/platform/android/BUILD.gn
+++ b/shell/platform/android/BUILD.gn
@@ -494,12 +494,15 @@ action("android_jar") {
       (impeller_enable_vulkan && current_cpu == "arm64" &&
        flutter_runtime_mode == "debug")) {
     assert(impeller_enable_vulkan)
-    deps += [ "//third_party/vulkan_validation_layers" ]
+    apilevel26_toolchain = "//build/toolchain/android:clang_arm64_apilevel26"
+    validation_layer_target =
+        "//third_party/vulkan_validation_layers($apilevel26_toolchain)"
+    validation_layer_out_dir =
+        get_label_info(validation_layer_target, "root_out_dir")
+    deps += [ validation_layer_target ]
     args += [
       "--native_lib",
-      rebase_path("libVkLayer_khronos_validation.so",
-                  root_build_dir,
-                  root_build_dir),
+      rebase_path("$validation_layer_out_dir/libVkLayer_khronos_validation.so"),
     ]
     if (current_cpu == "arm64") {
       args += [
--- a/shell/platform/android/BUILD.gn
+++ b/shell/platform/android/BUILD.gn
@@ -490,7 +490,9 @@ action("android_jar") {
     ":pom_libflutter",
   ]
 
-  if (enable_vulkan_validation_layers) {
+  if (enable_vulkan_validation_layers ||
+      (impeller_enable_vulkan && current_cpu == "arm64" &&
+       flutter_runtime_mode == "debug")) {
     assert(impeller_enable_vulkan)
     deps += [ "//third_party/vulkan_validation_layers" ]
     args += [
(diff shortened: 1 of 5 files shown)


jonahwilliams (approved):
LGTM if licensing is good.

We can also run a debug build smoke test (later) in the devicelab to make sure we're not hitting validation errors.

gaaclarke:
This runs fine locally but fails fails to build on CI.  Presumably the api level is coming from the buildroot (https://github.com/flutter/buildroot/blob/6b0243e97d94836767ea8b94a477d7d15ee3bb91/build/config/android/config.gni#L47).

It's API guard isn't even apropos for us since we don't load the dylib unless we are using the vulkan backend (and thus are on a platform that supports vulkan, ie >= 23).

I'll have to find a way to override this API setting, probably editing the buildroot tomorrow.

shell/platform/android/BUILD.gn
```diff
@@ -490,7 +490,9 @@ action("android_jar") {
     ":pom_libflutter",
   ]
 
-  if (enable_vulkan_validation_layers) {
+  if (enable_vulkan_validation_layers ||
+      (impeller_enable_vulkan && current_cpu == "arm64" &&
+       flutter_runtime_mode == "debug")) {
     assert(impeller_enable_vulkan)
     deps += [ "//third_party/vulkan_validation_layers" ]
```
zanderso, line 497: What's the difference between this and `//third_party/vulkan-deps/vulkan-validation-layers` ?

zanderso: Oh, I see. They're the same thing https://github.com/flutter/buildroot/blob/master/build/secondary/third_party/vulkan_validation_layers/BUILD.gn.

gaaclarke:
@zanderso @jonahwilliams actually this may not be possible to implement because of limitations in GN.  So the problem is that we need android >= 23 to build the validation layers.  The engine supports android 22.  It's not a problem at runtime because we can check the android version before loading the dylib.

The way we get around this build problem today is that if you are building a local engine with validation layers enabled we globally set the android api level to 26.  We can't do the same trick when building the engine for everyone.  There doesn't seem to be a way to have different android api levels for different targets in GN.  I'll verify this.

gaaclarke:
The other option would be to split our build process into multiple ninja invocations then bundle everything together.  I'm not sure how we'd do that, but at face value sounds involved.

jonahwilliams:
I think that if this is too more work, instead we should just point folks and the triagers to https://developer.android.com/ndk/guides/graphics/validation-layer

zanderso:
I'd like to study this a bit to see if I can make a suggestion. I can sometimes bend GN to my will.

gaaclarke:
> I'd like to study this a bit to see if I can make a suggestion. I can sometimes bend GN to my will.

Here is the override we currently use for reference: https://github.com/flutter/engine/blob/069354fb694c016c3c994a2256f8513be1bdae88/tools/gn#L652

Our options as I see them:
1) Bump the engine to only support android `>= 23`
1) Split our build into multiple GN builds and rely on a build script that runs ninja twice and combines the result
1) (**forbidden**) ~~package prebuilt validation layers~~
1) Rely on showing users how to inject the validation layers.  Maybe there is something we can do to grease the wheels here?  A script that modifies an apk?

gaaclarke:
Heads up I'm, putting into the autosubmit queue.  I've been in contact with the reviewers about changes we made since approval.  I just wanted to notify every since it has technically changed.

jonahwilliams (approved):
Still LGTM, thanks @gaaclarke and @zanderso !

tools/gn
```diff
@@ -646,7 +646,9 @@ def to_gn_args(args):
   if args.unoptimized and args.target_os == 'android' and args.android_cpu == 'arm64':
     enable_vulkan_validation = True
 
-  if enable_vulkan_validation:
+  if enable_vulkan_validation or (args.enable_impeller_vulkan and
```
zanderso, line 649: With the changes to the GN build for the validation layers .so, does the API level for the whole build still need to be bumped up to 26 down on line 654?

gbak:
@gaaclarke is there a way to exclude this library from the Android build? The Vulkan library doubles the size of our debug apks.  I've seen the suggestion of a jniLibs exclude, but didn't know if that would have any adverse behaviors.
`excludes.add("**/libVkLayer_khronos_validation.so")`

gaaclarke:
> @gaaclarke is there a way to exclude this library from the Android build? The Vulkan library doubles the size of our debug apks. I've seen the suggestion of a jniLibs exclude, but didn't know if that would have any adverse behaviors.
> `excludes.add("**/libVkLayer_khronos_validation.so")`

Yes, you can just remove it.  I'm not familiar with android build systems so I'm not sure if that will invalidate your codesigning.  Otherwise, there should be some gradle command to strip it.  I believe that if you remove the dynamic libraries that flutter will still function correctly, it should just print out an error message that it couldn't find the validation layers.  If you find that isn't the case I'd file a bug for it.

PR #45355, opened 2023-08-31
[Impeller] copy data out of DlVertices.

jonahwilliams:
The implementation of DLVerticesGeometry is holding onto a raw ptr that it doesn't own nor does it know that the lifecycle is deterministic. My bad!

This mostly worked (ehhh) until the stopwatch visualizer change. I've confirmed that going back to copying the data fixes the issue with the stopwatch visualizer.

Fixes https://github.com/flutter/flutter/issues/133813

### DIFF
--- a/impeller/entity/geometry/vertices_geometry.cc
+++ b/impeller/entity/geometry/vertices_geometry.cc
@@ -4,8 +4,304 @@
 
 #include "impeller/entity/geometry/vertices_geometry.h"
 
+#include <utility>
+
+#include <utility>
+#include "impeller/core/formats.h"
+
 namespace impeller {
 
+// Fan mode isn't natively supported. Unroll into triangle mode by
+// manipulating the index array.
 //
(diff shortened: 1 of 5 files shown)


jonahwilliams:
I manually verified the existing golden tests are still valid, and I believe these run on CI too.

impeller/entity/geometry/vertices_geometry.cc
matanlurey, line 1: I wonder if it makes sense to mark this branched from dl_dispatcher?

matanlurey (approved):
Thanks for fixing this!

flar:
Ah, my fix was just going to be to have a shared_ptr to DlVertices in the object and to use DlVertices::Make() to make a new copy. I didn't look too closely, but this looks like it is saving an extra step as well.

impeller/entity/geometry/vertices_geometry.cc
jonahwilliams: I'm going to pass on doing this.

PR #45357, opened 2023-08-31
Do not decode PNG images into a wide gamut color space

jason-simmons:
Skia's PNG codec does not support wide gamut output formats for some images (such as palette-based images)

See https://github.com/flutter/flutter/issues/133013


gaaclarke (changes requested):
Looks like you hit our `ImpellerWideGamutDisplayP3` test.  I was starting to sweat, thinking that wasn't running.  We should fix this bug on the Skia side or limit disabling wide gamut to the subset of PNG files that aren't supported somehow until we can address it in Skia.

PR #45358, opened 2023-08-31
Cull the RTree bounds when they are forwarded in DrawDisplayList

flar:
Fixes https://github.com/flutter/flutter/issues/133437

With this fix, the [benchmarks](https://github.com/flutter/flutter/pull/133434) for pictures with huge bounds become runnable under more conditions. Currently those benchmarks only run on platforms where we don't run into this issue. The performance on those platforms/conditions will not change much, but with this fix we can run them on Impeller and add a variant that has a platform view on the display.

### DIFF
--- a/display_list/dl_builder.cc
+++ b/display_list/dl_builder.cc
@@ -1225,7 +1225,7 @@ void DisplayListBuilder::DrawDisplayList(const sk_sp<DisplayList> display_list,
       auto rtree = display_list->rtree();
       if (rtree) {
         std::list<SkRect> rects =
-            rtree->searchAndConsolidateRects(bounds, false);
+            rtree->searchAndConsolidateRects(GetLocalClipBounds(), false);
         accumulated = false;
         for (const SkRect& rect : rects) {
           // TODO (https://github.com/flutter/flutter/issues/114919): Attributes


flar:
It would be hard to write a test case to verify this fix because the end result of the problem was not visible - it was a lot of wasted effort under the covers. In particular, the code spent a large amount of time consolidating rectangles that were off screen only to have the code that iterated the rectangles clip them out anyway - resulting in no change to the final rtrees.

jonahwilliams:
If it's just a performance improving change then I think the benchmarks you added should be sufficient.

zanderso:
Yeah, this is tested in the framework by the benchmark added in https://github.com/flutter/flutter/pull/133434, right?

flar:
Yes and no. The benchmarks are currently only running on configurations that use code paths that do not go through this particular line of code. They test/track rtree culling in general, but they don't end up running into this line of code.

But, when I land this I can add a new variant of the benchmark that does run on a code path that goes through this code to check for regressions. I can't add that until this lands because the benchmark will error out until this perf issue is fixed.

(Since it would be landed with bring_up=true, it could be landed without "breaking" the tree, but it takes 5 minutes to annoy the test apparatus enough to have it stop trying to run the benchmark and then it retries it just chewing on the bad data over and over for no useful purpose and wasting time on our CI machines.)

So, yes, a benchmark variant I plan to land soon after this rolls in will test for this failure. It's just not in this PR itself.

zanderso (approved):

PR #45360, opened 2023-09-01
Dl geometry classes

flar:
Currently a WIP. Creating a PR to check how it builds and tests on multiple platforms.

This PR is for testing only, eventually this code will need an API review and a plan on how to integrate it without creating a huge regression/revert risk.


flar:
Will revisit this when it becomes a priority again

PR #45326, opened 2023-08-31
Revert dl split

zanderso:
Reverting for https://github.com/flutter/flutter/issues/133525
fixes: https://github.com/flutter/engine/pull/45326

PR #45386, opened 2023-09-01
[Impeller] Reland DlAiksCanvas, again.

dnfield:
This is a partial revert of #45326.

The DlAiksCanvas implementation was reverted there because it caused merge conflicts. I've fixed those up.

This did introduce some regressions that are fixed by https://github.com/flutter/engine/pull/45313

This also depends on #45355 to avoid dangling pointer issues with `DrawVertices`.


jonahwilliams (approved):
LGTM, but also building locally to test on wonderous

jonahwilliams:
This needs to pull in latest main

jonahwilliams:
Something funky is going on with the performance overlay. Even though it has opacity and is being drawn srcOver, it is showing up opaque in this patch

flutter_09

jonahwilliams:
It looks like opacity isn't working in a number of cases. This might be a paint pass delegate issue? or some case where the subpass expects the parent to apply the opacity but it doesn't?

Also see the gallery app, the shrine transitions and page transitions don't show any opacity.

jonahwilliams:
Something funky with the page transitions too:


https://github.com/flutter/engine/assets/8975114/3046469f-24e1-4cbb-8c8b-42fe0bbbb724

jonahwilliams:
The Perf overlay was unrelated, I'll help look at the opacity issues next week.

dnfield:
One possibility is that the picture is supposed to be drawn iwth opacity and right now we're always treating it as one in the DrawImpellerPicture call..

jonahwilliams:
There were some other regressions on the tessellation benchmarks on this change too. I think this is generally related to us computing the bounds of paths many times, so I opened https://github.com/flutter/engine/pull/45456 to address that.

jonahwilliams (changes requested):
Please don't land this until we verify all of the correctness issues

dnfield:
Ah, misunderstood what correctness issues remained. Are there any issues filed?

jonahwilliams:
Still trying to figure it out, but I think entitypass clone is still buggy

jonahwilliams:
I verified this locally and found it might have been due to an interaction with the DL split, but trying again on ToT I don't see any issues.

jonahwilliams (approved):
LGTM

jonahwilliams:
Actually we just found one issue. Investigating.

dnfield:
Looking at wonderous, the icon in the top left is fading out the text too soon/too late on the main page - the advanced blends also look off. This could be an issue with `DrawPicture` still. I will be out for about a week so marking this draft.

dnfield:
This pretty stale, going to close for now.

PR #45388, opened 2023-09-01
[Impeller] Better demonstrate blur and draw picture?

jonahwilliams:
The current goldens are changing a lot, this should be stabler. though not sure if it looks quite right.


bdero (approved):
LGTM

PR #45390, opened 2023-09-01
Adds a Dart library for loading and parsing build configs

zanderso:
Also adds a test that the build configs in the repo are valid json that matches the spec.

Fleshed out @christopherfujino's code from here https://github.com/christopherfujino/flutter-engine-runner/blob/main/main.dart

### DIFF
--- a/tools/pkg/engine_build_configs/bin/check.dart
+++ b/tools/pkg/engine_build_configs/bin/check.dart
@@ -1,0 +1,67 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'dart:io' as io;
+
+import 'package:engine_build_configs/engine_build_configs.dart';
+import 'package:engine_repo_tools/engine_repo_tools.dart';
+import 'package:path/path.dart' as p;
+
+// Usage:
+// $ dart bin/check.dart [/path/to/engine/src]
+
+void main(List<String> args) {
+  final String? engineSrcPath;
+  if (args.isNotEmpty) {
+    engineSrcPath = args[0];
+  } else {
+    engineSrcPath = null;
+  }
+
+  // Find the engine repo.
+  final Engine engine;
+  try {
+    engine = Engine.findWithin(engineSrcPath);
+  } catch (e) {
+    io.stderr.writeln(e);
+    io.exitCode = 1;
+    return;
+  }
+
+  // Find and parse the engine build configs.
+  final io.Directory buildConfigsDir = io.Directory(p.join(
+    engine.flutterDir.path, 'ci', 'builders',
+  ));
+  final BuildConfigLoader loader = BuildConfigLoader(
+    buildConfigsDir: buildConfigsDir,
+  );
+
+  // Treat it as an error if no build configs were found. The caller likely
+  // expected to find some.
+  final Map<String, BuildConfig> configs = loader.configs;
+  if (configs.isEmpty) {
+    io.stderr.writeln(
+      'Error: No build configs found under ${buildConfigsDir.path}',
+    );
+    io.exitCode = 1;
+    return;
+  }
+  if (loader.errors.isNotEmpty) {
+    loader.errors.forEach(io.stderr.writeln);
+    io.exitCode = 1;
+  }
+
+  // Check the parsed build configs for validity.
+  for (final String name in configs.keys) {
+    final BuildConfig buildConfig = configs[name]!;
+    final List<String> buildConfigErrors = buildConfig.check(name);
+    if (buildConfigErrors.isNotEmpty) {
+      io.stderr.writeln('Errors in ${buildConfig.path}:');
+      io.exitCode = 1;
+    }
+    for (final String error in buildConfigErrors) {
+      io.stderr.writeln('    $error');
+    }
+  }
+}
--- a/tools/pkg/engine_build_configs/lib/src/build_config.dart
+++ b/tools/pkg/engine_build_configs/lib/src/build_config.dart
@@ -1,0 +1,843 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'package:meta/meta.dart';
+
+// This library parses Engine build config data out of the "Engine v2" build
+// config JSON files with the format described at:
+//   https://github.com/flutter/engine/blob/main/ci/builders/README.md
+
+/// Base class for all nodes in the build config.
+sealed class BuildConfigBase {
+  BuildConfigBase(this.errors);
+
+  /// Accumulated errors. Non-null and non-empty when a node is invalid.
+  final List<String>? errors;
+
+  /// Whether there were errors when loading the data for this node.
+  late final bool valid = errors == null;
+
+  /// Returns an empty list when the object is valid, and errors when it is not.
+  /// Subclasses with more data to check for validity should override this
+  /// method and add `super.check(path)` to the returned list.
+  @mustCallSuper
+  List<String> check(String path) {
+    if (valid) {
+      return <String>[];
+    }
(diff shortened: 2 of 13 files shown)


tools/pkg/engine_build_configs/lib/src/build_config.dart
```diff
@@ -0,0 +1,820 @@
...
+    return errors;
+  }
+}
+
+/// "builds" -> "ninja" contains a map with fields like:
+/// {
+///   "config": "",
+///   "targets": [""]
+/// },
+final class BuildNinja  extends BuildConfigBase {
```
christopherfujino, line 316:
```suggestion
final class BuildNinja extends BuildConfigBase {
```

christopherfujino (approved):
LGTM other than the spacing nit

tools/pkg/engine_build_configs/bin/check.dart
```diff
@@ -0,0 +1,69 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'dart:io' as io;
+
+import 'package:engine_build_configs/engine_build_configs.dart';
+import 'package:engine_repo_tools/engine_repo_tools.dart';
```
matanlurey, line 8: Nice! 😀

tools/pkg/engine_build_configs/bin/check.dart
```diff
@@ -0,0 +1,69 @@
...
+import 'package:path/path.dart' as p;
+
+// Usage:
+// $ dart bin/check.dart [/path/to/engine/src]
+
+void main(List<String> args) {
+  late final String? engineSrcPath;
```
matanlurey, line 15: You can actually leave off "late", definitive assignment will allow you to make it "regular" final!

tools/pkg/engine_build_configs/bin/check.dart
```diff
@@ -0,0 +1,69 @@
...
+  if (args.isNotEmpty) {
+    engineSrcPath = args[0];
+  } else {
+    engineSrcPath = null;
+  }
+
+  // Find the engine repo.
+  late final Engine engine;
```
matanlurey, line 23: Same here I believe.

Please also apply elsewhere in the code (many but not all of your late usages elsewhere in the PR).

tools/pkg/engine_build_configs/lib/src/build_config.dart
```diff
@@ -0,0 +1,820 @@
...
+// found in the LICENSE file.
+
+import 'package:meta/meta.dart';
+
+// This library parses Engine build config data out of the "Engine v2" build
+// config JSON files with the format described at:
+//   https://github.com/flutter/engine/blob/main/ci/builders/README.md
+
+/// Base class for all nodes in the build config.
+base class BuildConfigBase {
```
matanlurey, line 12: optional (maybe file an issue to do it later): We could make this `sealed base` to allow exhaustive checking/pattern matching of the different kinds of BuildConfigs! It wouldn't have any other changes to your code I believe.

tools/pkg/engine_build_configs/lib/src/build_config.dart
```diff
@@ -0,0 +1,820 @@
...
+  BuildConfigBase(this.errors);
+
+  /// Accumulated errors. Non-null and non-empty when a node is invalid.
+  final List<String>? errors;
+
+  /// Whether there were errors when loading the data for this node.
+  late final bool valid = errors == null;
```
matanlurey, line 19: not a comment: This is so cool.

tools/pkg/engine_build_configs/lib/src/build_config.dart
```diff
@@ -0,0 +1,820 @@
...
+
+  /// Returns the list of errors when the object is not valid. Subclasses
```
matanlurey, line 21: nit: Perhaps it would read better as:

"Returns an empty list when the object is valid, and errors when it is not.

Subclasses..."

tools/pkg/engine_build_configs/lib/src/build_config.dart
```diff
@@ -0,0 +1,820 @@
...
+    }
+    if (ninja == null) {
+      errors.add('"ninja" field is malformed.');
+    }
+
+    late final List<BuildTask>? generators;
+    if (map['generators'] == null) {
+      generators = <BuildTask>[];
+    } else if (map['generators'] is! Map<String, Object?>) {
+      errors.add('"generators" field is malformed.');
```
matanlurey, line 202: Is it worth perhaps adding the runtime type of like "expected: Map, actual: bool" to make debugging easier (since this is a debugging/testing tool)?

If so maybe pull out a helper function too to re-use with 'ninja' and  related.

matanlurey (changes requested):

tools/pkg/engine_build_configs/bin/check.dart
```diff
@@ -0,0 +1,69 @@
...
+    return;
+  }
+
+  // Find and parse the engine build configs.
+  final io.Directory buildConfigsDir = io.Directory(p.join(
+    engine.flutterDir.path, 'ci', 'builders',
+  ));
+  final List<String> errors = <String>[];
+  final BuildConfigLoader loader = BuildConfigLoader(
+    errors: errors,
```
mossmana, line 38: Just curious why errors needs to be passed in. Is that for dependency injection? If that's the case, why just errors and not configs for example?

tools/pkg/engine_build_configs/test/build_config_test.dart
```diff
@@ -0,0 +1,366 @@
...
+      ]
+    }
+  ]
+}
+''';
+
+int main() {
+  test('BuildConfig parser works', () {
+    final BuildConfig buildConfig = BuildConfig.fromJson(
+      path: 'linux_test_config',
```
mossmana, line 93: Does the linux case cover all the bases or should other platforms also be included?

mossmana (commented):
This looks great so far. Just had a couple questions.

tools/pkg/engine_build_configs/test/build_config_test.dart
christopherfujino: it looks like `check_build_configs.sh` is essentially an integration test that will verify ALL the real configs, and run for every commit in the linux_unopt shard.

tools/pkg/engine_build_configs/lib/src/build_config.dart
zanderso: Done. Thanks!

tools/pkg/engine_build_configs/bin/check.dart
zanderso: Done. Thanks!

tools/pkg/engine_build_configs/lib/src/build_config.dart
zanderso: Changed to `sealed`.

tools/pkg/engine_build_configs/lib/src/build_config.dart
zanderso: Good call. Changed.

tools/pkg/engine_build_configs/lib/src/build_config.dart
zanderso: Good idea. Pulled out a helper `appendTypeError`.

tools/pkg/engine_build_configs/bin/check.dart
zanderso: Oh! This is the result of an incomplete refactor. It was previously a `StringSink` that needed to be injected for tests, but as a `List<String>` it doesn't need to be injected anymore. Thanks!

tools/pkg/engine_build_configs/test/build_config_test.dart
zanderso: Right. This file is just some unit tests, so the string on this line doesn't matter.

matanlurey (approved):
LGTM

PR #45406, opened 2023-09-01
[impeller] premultiply vertices colors.

jonahwilliams:
Lost this after the transition.

### After

image

### Before

image


jonahwilliams:
Without doing this, if you have fully saturated colors with opacity, then they show as opaque: For example 0x99FFFFFF will look solid, hence the rendering bug in the perf overlay.

matanlurey (approved):

PR #45418, opened 2023-09-03
[Impeller] construct text frames on UI thread.

jonahwilliams:
Conversion of SkTextBlobs to impeller::TextFrame objects is one of the most expensive operations in display list dispatching. While the rest of the engine and framework makes a reasonable attempt to cache the SkTextBlobs generated during paragraph construction, the design of the dl dispatcher means that these the Impeller backend will always reconstruct all text frames on each frame - even if the display list/picture that contained those text frames was unchanged.

Removing this overhead is one of the goals of https://github.com/flutter/engine/pull/45386 , however this patch is also fairly risky and will be difficult to land. As a more incremental solution, we can instead construct the impeller::TextFrame objects when performing paragraph painting and record them in the display list. This both moves the text frame construction to the UI thread and allows the framework/engine to cache unchanged text frames.

This also does not conflict with the dl_aiks_canvas patch directly, and is fine to land before or after it does. (though I'd argue we should land this first).

To compare the current performance levels, I ran the complex_layout_scroll perf test, since this is fairly text filled. On a Pixel 6 pro.   Across several runs this is a fairly consistent ~1ms raster time improvement.

### Skia
```
  "average_frame_build_time_millis": 1.497333333333333,
  "90th_percentile_frame_build_time_millis": 2.038,
  "99th_percentile_frame_build_time_millis": 17.686,
  "worst_frame_build_time_millis": 23.095,
  "missed_frame_build_budget_count": 3,
  "average_frame_rasterizer_time_millis": 5.5078589743589745,
  "stddev_frame_rasterizer_time_millis": 2.226343414420338,
  "90th_percentile_frame_rasterizer_time_millis": 7.481,
  "99th_percentile_frame_rasterizer_time_millis": 19.11,
  "worst_frame_rasterizer_time_millis": 79.799,
  "missed_frame_rasterizer_budget_count": 7,
  "frame_count": 234,
  "frame_rasterizer_count": 234,
  "new_gen_gc_count": 10,
  "old_gen_gc_count": 2,
 ```
 
 ### Impeller (ToT)
 
 ```
   "average_frame_build_time_millis": 1.431575000000001,
  "90th_percentile_frame_build_time_millis": 2.196,
  "99th_percentile_frame_build_time_millis": 14.486,
  "worst_frame_build_time_millis": 23.728,
  "missed_frame_build_budget_count": 2,
  "average_frame_rasterizer_time_millis": 6.536087499999999,
  "stddev_frame_rasterizer_time_millis": 1.9902712500000004,
  "90th_percentile_frame_rasterizer_time_millis": 9.705,
  "99th_percentile_frame_rasterizer_time_millis": 14.727,
  "worst_frame_rasterizer_time_millis": 17.838,
  "missed_frame_rasterizer_budget_count": 1,
  "frame_count": 240,
  "frame_rasterizer_count": 240,
  "new_gen_gc_count": 10,
  "old_gen_gc_count": 2,
  ```
  
  ### Impeller (Patched)
  
  ```
    "average_frame_build_time_millis": 1.4500167364016743,
  "90th_percentile_frame_build_time_millis": 2.478,
  "99th_percentile_frame_build_time_millis": 14.883,
  "worst_frame_build_time_millis": 18.782,
  "missed_frame_build_budget_count": 1,
  "average_frame_rasterizer_time_millis": 5.023033333333336,
  "stddev_frame_rasterizer_time_millis": 1.6445388888888894,
  "90th_percentile_frame_rasterizer_time_millis": 7.814,
  "99th_percentile_frame_rasterizer_time_millis": 13.497,
  "worst_frame_rasterizer_time_millis": 15.008,
  "missed_frame_rasterizer_budget_count": 0,
  "frame_count": 239,
  "frame_rasterizer_count": 240,
  "new_gen_gc_count": 8,
  "old_gen_gc_count": 0,
  ```

### CODE REVIEW

**Comment by jonahwilliams** [comment]
A second component to this change is that the public API for getting bitmap/emoji font requires a Skia paragraph object, and so moving the text frame construction allows us to (later) fix that too

File: display_list/dl_builder.cc [hunk 1]
```diff
@@ -1302,6 +1302,48 @@ void DisplayListBuilder::DrawTextBlob(const sk_sp<SkTextBlob>& blob,
   SetAttributesFromPaint(paint, DisplayListOpFlags::kDrawTextBlobFlags);
   drawTextBlob(blob, x, y);
 }
+
+void DisplayListBuilder::drawTextFrame(
```
matanlurey, line 1306: Does this mean that `drawTextBlob` will go unused in the Impeller backend?

If so, would you want to (ifdef'd?) make sure it's not used, i.e. `UNIMPLEMENTED`, similar to the dispatcher?

display_list/dl_builder.cc
```diff
@@ -1302,6 +1302,48 @@ void DisplayListBuilder::DrawTextBlob(const sk_sp<SkTextBlob>& blob,
...
+    const std::shared_ptr<impeller::TextFrame>& text_frame,
+    SkScalar x,
+    SkScalar y) {
+  DisplayListAttributeFlags flags = kDrawTextBlobFlags;
+  OpResult result = PaintResult(current_, flags);
+  if (result == OpResult::kNoEffect) {
+    return;
+  }
+  impeller::Rect bounds = text_frame->GetBounds();
+  SkRect sk_bounds = SkRect::MakeLTRB(bounds.GetLeft(), bounds.GetTop(),
```
matanlurey, line 1316: I'm surprised we don't have a conversion for this? If we don't, feel free to resolve.

third_party/txt/src/skia/paragraph_skia.cc
matanlurey, line 1: I wonder if we would want to just do a separate class at some point versus having all these switches?

(No changes requested in this PR, more of an open question/possible TODO)

third_party/txt/src/skia/paragraph_skia.cc
```diff
@@ -180,6 +202,14 @@ class DisplayListParagraphPainter : public skt::ParagraphPainter {
     return path;
   }
 
+  bool ShouldRenderAsPath(const DlPaint& paint) const {
+    // Text with non-trivial color sources or stroke paint mode should be
```
matanlurey, line 206: Thanks for adding context.

Is this for performance or correctness or both? Could you note that?

third_party/txt/src/skia/paragraph_skia.cc
```diff
@@ -79,6 +82,20 @@ class DisplayListParagraphPainter : public skt::ParagraphPainter {
     }
     size_t paint_id = std::get<PaintID>(paint);
     FML_DCHECK(paint_id < dl_paints_.size());
+
+    if (ShouldRenderAsPath(dl_paints_[paint_id])) {
```
matanlurey, line 86: Consider putting this inside the `impeller_enabled_` block below, and FML_DCHECK'ing instead.

matanlurey (approved):
LGTM with a few comments.

third_party/txt/src/skia/paragraph_skia.cc
jonahwilliams: I think that ultimately depends on how many (if any) additional checks we need to add versus how quickly we move through preview for the remaining platforms. I wouldn't think so at this time, but if we have to do much more than splitting into two might make sense. however is still 80% the same code.

display_list/dl_builder.cc
jonahwilliams: We do, its just all over the place.

flar:
How expensive are the TextBlobs themselves? If we kept both around then we'd retain the ability to render a DL under both Skia and Impeller, but at the cost of storing both. That isn't our long-term goal, but it keeps some ability to remain backend-agnostic while we get there.

third_party/txt/src/skia/paragraph_skia.cc
dnfield: I think you also want to guard this with `IMPELLER_SUPPORTS_RENDERING`?

jonahwilliams:
The text blobs themselves won't be expensive to store, once we have them we have them. But DLs isn't backend agnostic, as both images/textures and shaders are backend specific now.

display_list/dl_builder.cc
jonahwilliams: I'm not sure if we use DisplayListBuilder on Skia ever. FYI @flar

third_party/txt/src/skia/paragraph_skia.cc
jonahwilliams: Done

third_party/txt/src/skia/paragraph_skia.cc
jonahwilliams: Added correctness information.

third_party/txt/src/skia/paragraph_skia.cc
flar: At one point DL was supposed to be an agnostic storage format that could be used for any backend, allowing common operations at the "picture" level. There was potential for more value added operations that we never pursued. Meanwhile performance considerations related to conversions have been driving it into custom backend-specific cases.

display_list/dl_builder.cc
flar: Yes, we do. That was its original purpose and is iteratively becoming its only purpose.

PR #45459, opened 2023-09-05
Makes Skia's vkQueueSubmit threadsafe.

gaaclarke:
fixes https://github.com/flutter/flutter/issues/133933

## Pre-launch Checklist

- [x] I signed the [CLA].


shell/platform/embedder/tests/embedder_unittests.cc
```diff
@@ -995,6 +1010,133 @@ TEST_F(EmbedderTest, NoLayerCreatedForTransparentOverlayOnTopOfPlatformLayer) {
   ASSERT_EQ(context.GetSurfacePresentCount(), 0u);
 }
 
+#if SHELL_ENABLE_VULKAN
+TEST_F(EmbedderTest,
```
gaaclarke, line 1014: @bdero please advice how you want these tests to be.  They are just straight copies of the software rendering now.  It was the easiest way I could find to test this.

bdero: Oh interesting -- does this test reproduce the problem and fail without your change? And if not, is `CanRenderGradientWithCompositorOnNonRootLayer` sufficient for covering the code in this PR?

If the test doesn't repro the problem, I would just remove it. Otherwise, tossing it into `embedder_gl_unittests` and writing it using `EmbedderTestMultiBackend` to cover both GL and Vulkan seems like a good approach for the moment.
Later on we should add a software context support to the multi-backend class, but that's perhaps a job for another day.

Basically, anything that's Vulkan-specific has a switchboard method:
- `auto& context = GetEmbedderContext(EmbedderTestContextType::kVulkanContext)` becomes `auto& context = GetEmbedderContext(GetParam());`
- `builder.SetRenderTargetType(EmbedderTestBackingStoreProducer::RenderTargetType::kVulkanImage);` becomes `builder.SetRenderTargetType(GetRenderTargetFromBackend(backend, false));`
- etc

bdero (approved):
LGTM

shell/platform/embedder/tests/embedder_unittests.cc
gaaclarke: The problem is a race condition which I can't force to happen.  So this doesn't assert that the race condition is gone, but it does exercise the code to make sure that it functions.  It's possible that the failure would happen in this test infrequently without the change.

Running `embedder_unittests` this is the only way I could get this code to be hit.  The `BUILD.gn` file only compiles `embedder_gl_unittests.cc` if `test_enable_gl == true` which isn't possible to enable on macos.  Let me see if I can detangle some of this.

gaaclarke: I pulled the new tests into a `embedder_vk_unittests.cc` to mirror the ones in `embedder_gl_unittest.cc` and `embedder_metal_unittests.cc`.  It has a comment in there that says some of the tests are in `embedder_gl_unittests.cc` and I added an issue to clean them up more later (noted in comment).

gaaclarke:
I'll try to rebase this to see if that flushes out the failure.

gaaclarke:
I'm going to revert this.  It works fine in tests, but it's not going to work for real usage of the embedder since the user has free access to the vkQueue and thus has no way to ensure that the vkQueue isn't used concurrently.

PR #45502, opened 2023-09-06
Add macOS support for plugin value publishing

stuartmorgan:
These directly copy the iOS APIs, to minimize the branching needed in plugins with shared implementation code, and to facilitate the long-term goal of merging the iOS and macOS plugin headers. This does mean replicating the unfortunately non-idiomatic behavior of having `valuePublishedByPlugin:` sometimes return `nil` and sometimes return `NSNull`, instead of distinguishing between `nil` cases (if that's actually even necessary here) via a more specific API. In isolation I would definitely not design the API with this behavior, but consistency with iOS is the more important factor.

(Eventually I think we'll need a sort of "v2" of iOS plugin APIs since there are a number of strange behaviors that we're currently stuck with, but migrating iOS and macOS together to a new set of APIs won't be any harder than doing just iOS, and in the short to medium term consistency will help the ecosystem more that trying to pre-create better APIs as macOS-only.)

Also fixes `FlutterEngineRegistrar` to have a weak pointer to the engine. This should really already have been the case since plugins can retain the registrar, creating a likely cycle; it's now a guaranteed cycle (and failed unit tests designed to find cycles) without that since the engine itself is now keeping references to them.

Fixes https://github.com/flutter/flutter/issues/124721

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
+++ b/shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
@@ -25,6 +25,8 @@
 #import "flutter/shell/platform/darwin/macos/framework/Source/FlutterViewController_Internal.h"
 #import "flutter/shell/platform/darwin/macos/framework/Source/FlutterViewEngineProvider.h"
 
+@class FlutterEngineRegistrar;
+
 NSString* const kFlutterPlatformChannel = @"flutter/platform";
 NSString* const kFlutterSettingsChannel = @"flutter/settings";
 NSString* const kFlutterLifecycleChannel = @"flutter/lifecycle";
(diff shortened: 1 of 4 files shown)


shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
```diff
@@ -274,12 +282,19 @@ @interface FlutterEngineRegistrar : NSObject <FlutterPluginRegistrar>
...
+ *
+ * The unusual NSNull is for the documented behavior of valuePublishedByPlugin:.
+ */
+@property(nonatomic, readonly, nonnull) NSObject* publishedValue;
 @end
 
 @implementation FlutterEngineRegistrar {
   NSString* _pluginKey;
-  FlutterEngine* _flutterEngine;
+  __weak FlutterEngine* _flutterEngine;
```
cbracken, line 297: Was this intended to be in this patch? Either way, nice catch. If you wanted to test this, there's a similar test for the channels you could optionally clone and tweak.

cbracken (approved):
lgtm!

shell/platform/darwin/macos/framework/Source/FlutterEngine.mm
cbracken: Ah I see the comment at the end of the description. Answer is yes :)

PR #45512, opened 2023-09-06
Multi-view Rasterizer 

dkwingsmt:
This PR refactors `Rasterizer` so that it's more suitable for multi-view.

Design doc: [flutter.dev/go/multi-view-pipeline-and-rasterizer](http://flutter.dev/go/multi-view-pipeline-and-rasterizer)

With this change, `Rasterizer::DrawToSurfaces` has a structure that can handle drawing to multiple views, although the lack of some functionality still blocks it, mostly related to `ExternalViewEmbedder` and `RasterCache`. 

## Pre-launch Checklist

- [ ] I signed the [CLA].


shell/common/rasterizer.h
```diff
@@ -180,6 +229,7 @@ class Rasterizer final : public SnapshotDelegate,
   ///             collects associated resources. No more rendering may occur
   ///             till the next call to `Rasterizer::Setup` with a new render
   ///             surface. Calling a teardown without a setup is user error.
+  ///             Calling this method for multiple times is safe.
```
loic-sharma, line 232:
```suggestion
  ///             Calling this method multiple times is safe.
```

shell/common/rasterizer.h
```diff
@@ -48,6 +49,54 @@ class AiksContext;
...
+  kDiscarded,
+};
+
+// The information to draw to all views of a frame.
+struct FrameItem {
+  FrameItem(std::list<LayerTreeTask> tasks,
+            std::unique_ptr<FrameTimingsRecorder> frame_timings_recorder)
+      : layer_tree_tasks(std::move(tasks)),
+        frame_timings_recorder(std::move(frame_timings_recorder)) {}
+  std::list<LayerTreeTask> layer_tree_tasks;
```
loic-sharma, line 94: Could you expand on why you chose a `std::list` instead of a `std::vector<std::unique_ptr<T>>`? Generally I'd lean towards the `std::vector` approach.

shell/common/rasterizer.h
```diff
@@ -48,6 +49,54 @@ class AiksContext;
...
+  // The drawing was done without any specified status.
+  kDone,
+  // Failed to rasterize the frame because the Rasterizer is not set up.
+  kNotSetUp,
+  // Nothing was done, because the call was not on the raster thread. Yielded to
+  // let this frame be serviced on the right thread.
+  kYielded,
+  // Nothing was done, because pipeline was empty.
+  kPipelineEmpty,
+  // Nothing was done, because GPU was unavailable.
```
loic-sharma, line 63:
```suggestion
  // The drawing was done without any specified status.
  kDone,
  // Failed to rasterize the frame because the Rasterizer is not set up.
  kNotSetUp,
  // Nothing was done because the call was not on the raster thread. Yielded to
  // let this frame be serviced on the right thread.
  kYielded,
  // Nothing was done because the pipeline was empty.
  kPipelineEmpty,
  // Nothing was done because the GPU was unavailable.
```

shell/common/rasterizer.h
```diff
@@ -511,18 +561,45 @@ class Rasterizer final : public SnapshotDelegate,
   ///
   void DisableThreadMergerIfNeeded();
 
+  //----------------------------------------------------------------------------
+  /// @brief      Returns whether TearDown has been called.
+  ///
+  ///             This method is only used only in unit tests.
```
loic-sharma, line 567:
```suggestion
  ///             This method is used only in unit tests.
```

shell/common/rasterizer.h
```diff
@@ -511,18 +561,45 @@ class Rasterizer final : public SnapshotDelegate,
...
+  ///
+  bool IsTornDown();
+
+  //----------------------------------------------------------------------------
+  /// @brief      Returns the last status of drawing the specific view.
+  ///
+  ///             This method is only used only in unit tests.
```
loic-sharma, line 574:
```suggestion
  ///             This method is used only in unit tests.
```

shell/common/rasterizer.cc
```diff
@@ -176,74 +189,71 @@ GrDirectContext* Rasterizer::GetGrContext() {
...
-  return last_layer_tree_.get();
+flutter::LayerTree* Rasterizer::GetLastLayerTree(int64_t view_id) {
+  auto found = last_successful_tasks_.find(view_id);
+  if (found == last_successful_tasks_.end()) {
+    return nullptr;
+  }
+  return found->second.layer_tree.get();
 }
 
 void Rasterizer::DrawLastLayerTree(
```
loic-sharma, line 200: Should we update this method's name?

shell/common/rasterizer.cc
```diff
@@ -176,74 +189,71 @@ GrDirectContext* Rasterizer::GetGrContext() {
...
     std::unique_ptr<FrameTimingsRecorder> frame_timings_recorder) {
-  if (!last_layer_tree_ || !surface_) {
+  if (last_successful_tasks_.empty() || !surface_) {
     return;
   }
-  RasterStatus raster_status = DrawToSurface(
-      *frame_timings_recorder, *last_layer_tree_, last_device_pixel_ratio_);
+  std::list<LayerTreeTask> tasks;
+  for (auto& [view_id, task] : last_successful_tasks_) {
+    tasks.push_back(std::move(task));
```
loic-sharma, line 207: Is moving correct here? `LayerTreeTask` does not have a move constructor, I would expect this to copy instead.

shell/common/rasterizer.cc
```diff
@@ -270,12 +280,30 @@ RasterStatus Rasterizer::Draw(
...
-         raster_status == RasterStatus::kSkipAndRetry;
+DrawStatus Rasterizer::ToDrawStatus(DoDrawStatus status) {
+  switch (status) {
+    case DoDrawStatus::kEnqueuePipeline:
+      return DrawStatus::kDone;
+    case DoDrawStatus::kNotSetUp:
+      return DrawStatus::kNotSetUp;
+    case DoDrawStatus::kGpuUnavailable:
+      return DrawStatus::kGpuUnavailable;
+    default:
```
loic-sharma, line 302: Please add a case for `DoDrawStatus::kDone` and use `FML_UNREACHABLE` to crash on an unhandled status.

For example:

https://github.com/flutter/engine/blob/d64abdb88ee56888185443dccbffe159d9b18e67/impeller/renderer/backend/vulkan/pipeline_library_vk.cc#L142-L150

shell/common/rasterizer.cc
```diff
@@ -393,42 +421,31 @@ fml::Milliseconds Rasterizer::GetFrameBudget() const {
...
-    std::unique_ptr<flutter::LayerTree> layer_tree,
-    float device_pixel_ratio) {
+    std::list<LayerTreeTask> tasks) {
   TRACE_EVENT_WITH_FRAME_NUMBER(frame_timings_recorder, "flutter",
                                 "Rasterizer::DoDraw", /*flow_id_count=*/0,
                                 /*flow_ids=*/nullptr);
   FML_DCHECK(delegate_.GetTaskRunners()
                  .GetRasterTaskRunner()
                  ->RunsTasksOnCurrentThread());
+  frame_timings_recorder->AssertInState(FrameTimingsRecorder::State::kBuildEnd);
```
loic-sharma, line 431: Should this check be before `TRACE_EVENT_WITH_FRAME_NUMBER` as that mutates the `frame_timings_recorder`?

flow/compositor_context.h
```diff
@@ -23,18 +23,19 @@ namespace flutter {
 
 class LayerTree;
 
+// The result status of CompositorContext::ScopedFrame::Raster.
 enum class RasterStatus {
   // Frame has been successfully rasterized.
   kSuccess,
-  // Frame is submitted twice. This is only used on Android when
-  // switching the background surface to FlutterImageView.
+  // Frame has been submited, but should be submitted again. This is only used
```
dnfield, line 30: Here and below, avoid using `should`.

Instead, describe what `must`/`must not` happen to make it clearer about what developers reading this code can actually expect.

flow/frame_timings.h
```diff
@@ -116,6 +116,12 @@ class FrameTimingsRecorder {
   /// Returns the recorded time from when `RecordRasterEnd` is called.
   FrameTiming GetRecordedTime() const;
 
+  /// Asserts in debug mode that the recorder is current at the specified state.
```
dnfield, line 119:
```suggestion
  /// Asserts in unopt builds that the recorder is current at the specified state.
```

shell/common/animator.cc
```diff
@@ -161,12 +161,15 @@ void Animator::Render(std::unique_ptr<flutter::LayerTree> layer_tree,
   delegate_.OnAnimatorUpdateLatestFrameTargetTime(
       frame_timings_recorder_->GetVsyncTargetTime());
 
-  auto layer_tree_item = std::make_unique<LayerTreeItem>(
-      std::move(layer_tree), std::move(frame_timings_recorder_),
-      device_pixel_ratio);
+  // TODO(dkwingsmt): Currently only supports a single window.
+  int64_t view_id = kFlutterImplicitViewId;
```
dnfield, line 165: File a bug and link it here.

flow/layers/layer_tree.h
```diff
@@ -94,6 +94,22 @@ class LayerTree {
...
+      : view_id(view_id),
+        layer_tree(std::move(layer_tree)),
+        device_pixel_ratio(device_pixel_ratio) {}
+  /// The target view to drawn to.
+  int64_t view_id;
+  /// The target layer tree to be drawn.
+  std::unique_ptr<LayerTree> layer_tree;
+  /// The pixel ratio of the target view.
+  float device_pixel_ratio;
+};
```
dnfield, line 111: You should disable copy and assign on this struct

shell/common/rasterizer.cc
dnfield: Copying a unique pointer won't do what you want - moving should be right - but we should delete the copy/assign functionality of this struct.

shell/common/rasterizer.cc
```diff
@@ -176,74 +189,71 @@ GrDirectContext* Rasterizer::GetGrContext() {
...
-          draw_result.raster_status = RasterStatus::kDiscarded;
-        } else {
-          draw_result = DoDraw(std::move(frame_timings_recorder),
-                               std::move(layer_tree), device_pixel_ratio);
-        }
-      };
+  FramePipeline::Consumer consumer = [&draw_result,
+                                      this](std::unique_ptr<FrameItem> item) {
+    draw_result = DoDraw(std::move(item->frame_timings_recorder),
+                         std::move(item->layer_tree_tasks));
```
dnfield, line 238: Something should probably clear out the tasks after this right?

Otherwise we have a list of tasks with nullptr...

shell/common/rasterizer.cc
```diff
@@ -497,73 +514,157 @@ Rasterizer::DoDrawResult Rasterizer::DoDraw(
...
-RasterStatus Rasterizer::DrawToSurfaceUnsafe(
+std::unique_ptr<FrameItem> Rasterizer::DrawToSurfacesUnsafe(
     FrameTimingsRecorder& frame_timings_recorder,
-    flutter::LayerTree& layer_tree,
-    float device_pixel_ratio) {
-  FML_DCHECK(surface_);
+    std::list<LayerTreeTask> tasks) {
+  // TODO(dkwingsmt): The rasterizer only supports rendering a single view
+  // and that view must be the implicit view. Properly support multi-view
+  // in the future.
```
dnfield, line 562: Link a bug.

shell/common/rasterizer.cc
```diff
@@ -497,73 +514,157 @@ Rasterizer::DoDrawResult Rasterizer::DoDraw(
...
+  }
+
   if (external_view_embedder_) {
     FML_DCHECK(!external_view_embedder_->GetUsedThisFrame());
     external_view_embedder_->SetUsedThisFrame(true);
     external_view_embedder_->BeginFrame(
-        layer_tree.frame_size(), surface_->GetContext(), device_pixel_ratio,
-        raster_thread_merger_);
-    embedder_root_canvas = external_view_embedder_->GetRootCanvas();
+        // TODO(dkwingsmt): Add all views here.
```
dnfield, line 590: Link a bug.

shell/common/rasterizer.cc
```diff
@@ -497,73 +514,157 @@ Rasterizer::DoDrawResult Rasterizer::DoDraw(
...
+    if (status == DrawSurfaceStatus::kSuccess) {
+      last_successful_tasks_.insert_or_assign(
+          view_id,
+          LayerTreeTask(view_id, std::move(layer_tree), device_pixel_ratio));
+    } else if (status == DrawSurfaceStatus::kRetry) {
+      resubmitted_tasks.emplace_back(view_id, std::move(layer_tree),
+                                     device_pixel_ratio);
+    }
+  }
+  // TODO(dkwingsmt): Pass in all raster caches
```
dnfield, line 629: File a bug.

What's not clear to me is whether we'd actually want multiple raster caches or just one raster cache per layer tree. There are also ongoing explorations to see if we can just remove the raster cache entirely as we migrate to Impeller.

dnfield: (@jonahwilliams fyi)

shell/common/rasterizer.cc
```diff
@@ -760,7 +850,9 @@ sk_sp<SkData> Rasterizer::ScreenshotLayerTreeAsImage(
 Rasterizer::Screenshot Rasterizer::ScreenshotLastLayerTree(
     Rasterizer::ScreenshotType type,
     bool base64_encode) {
-  auto* layer_tree = GetLastLayerTree();
+  // TODO(dkwingsmt): Support screenshotting all last layer trees
+  // when the shell protocol supports multi-views.
```
dnfield, line 854: File a bug.

dnfield (commented):
No major concerns, mostly nits

shell/common/rasterizer.cc
jonahwilliams: I don't think its possible for the rasterizer to know if it is safe to share any of this information. i.e. if these are from different windows with a different GlContext, or the different windows with the same GlContext, et cetera. The backends are not designed to handle this.

shell/common/rasterizer.h
dkwingsmt: I chose it mostly because it is pushed one after another to avoid reallocating. I'm not hard on it. But why do you think vector is better? And do you specifically recommend a unique_ptr as well?

shell/common/rasterizer.cc
dkwingsmt: It seems to me that the `TRACE` only gets `recorder->GetFrameNumberTraceArg()` and does not mutate the recorder at all.

shell/common/rasterizer.cc
dkwingsmt: I've changed it to `DrawLastLayerTrees`.

shell/common/rasterizer.cc
dkwingsmt: The tasks are moved all the way into `DrawToSurfacesUnsafe`, where some are discarded, some moved into `resubmitted_tasks`, and some stored in `last_successful_tasks_`. Since the vector is moved, there shouldn't be a list of nullptrs.

dkwingsmt:
@loic-sharma @dnfield I should have addressed all review comments. Can you take another look?

shell/common/rasterizer.h
loic-sharma: > But why do you think vector is better?

There's trade-offs so I wouldn't say one is always better than the other. The `std::list` has different performance characteristics than `std::vector`, but in this case you're appending/iterating which is ideal for `std::vector`.

> And do you specifically recommend a unique_ptr as well?

More trade-offs, but the benefits are:

1. If appending the `std::vector` requires reallocating the underlying array, `std::unique_ptr` enables the items to be moved instead of copied
2. You can move items in/out of the vector instead of copying (not needed here)
3. Using `std::unique_ptr` avoids issues if items have different sizes (not needed here)

flow/layers/layer_tree.h
```diff
@@ -94,6 +94,27 @@ class LayerTree {
...
+struct LayerTreeTask {
+ public:
+  LayerTreeTask(int64_t view_id,
+                std::unique_ptr<LayerTree> layer_tree,
+                float device_pixel_ratio)
+      : view_id(view_id),
+        layer_tree(std::move(layer_tree)),
+        device_pixel_ratio(device_pixel_ratio) {}
+
+  /// The target view to drawn to.
```
loic-sharma, line 107:
```suggestion
  /// The target view to draw to.
```

flow/layers/layer_tree.h
```diff
@@ -94,6 +94,27 @@ class LayerTree {
...
+  int64_t view_id;
+  /// The target layer tree to be drawn.
+  std::unique_ptr<LayerTree> layer_tree;
+  /// The pixel ratio of the target view.
+  float device_pixel_ratio;
+
+ private:
+  FML_DISALLOW_COPY(LayerTreeTask);
```
dnfield, line 115: _AND_ASSIGN

shell/common/rasterizer.cc
dnfield: Oh, I think I see now, this is the last spot they're used.

dnfield (approved):
LGTM

PR #45520, opened 2023-09-06
[Impeller] Gaussian blur: Remove the current blur style implementation.

bdero:
Resolves https://github.com/flutter/flutter/issues/122658.
Resolves remaining blur issues visible in: https://github.com/flutter/flutter/issues/132839

The current blur style implementation hasn't been working since we added the downscaling optimization, and it will continue not working after we switch to https://github.com/flutter/flutter/issues/131580.

In it's current state, it at best causes awful looking undesirable results that users are forced to work around anyhow. We should just remove it in the meantime.

Before:
image

After:
image


jonahwilliams (approved):
RIP in Pieces, LGTM

I'm currently working on a patch that will jam the two filter contents together, this change will significantly simplify that.

Do we have an issue tracking relanding/fixing the blur styles?

bdero:
SGTM. I went ahead and split the reimplementation for https://github.com/flutter/flutter/issues/122658 out into a separate issue here: https://github.com/flutter/flutter/issues/134178

bdero:
I need to remember to delete the gen directory when testing shader removals locally.

zanderso:
Looks like this patch broke impeller-cmake-example.

bdero:
Yup, already on it. ;)

PR #45522, opened 2023-09-07
Fix: On the Web, cannot support multiline inputting when registering customized TextInputControl

Satsrag:
For [flutter/flutter/125875](https://github.com/flutter/flutter/issues/125875)

When registering customized TextInputControl, the _PlatformTextInputControl sends inputType = TextInputType.none to the engine. After receiving TextInputType.none, the engine on the Web will create a `<input>` element instead of  `<textarea>`. So there is no way to input \n(multiline).

This is my solution. I tested in Android Chrome, iOS Safari, and macOS Chrome. It works for me. But I'm not sure about other use cases. I pleasure, If someone gives me suggestions.

### DIFF
--- a/lib/web_ui/lib/src/engine/text_editing/input_type.dart
+++ b/lib/web_ui/lib/src/engine/text_editing/input_type.dart
@@ -112,7 +112,7 @@ class NoTextInputType extends EngineInputType {
 /// For [NoTextInputType] (mapped to [TextInputType.none] with
 /// [isMultiline] = false), it creates an <input> element with the
 /// inputmode="none" attribute.
-class MultilineNoTextInputType extends EngineInputType {
+class MultilineNoTextInputType extends MultilineInputType {
   const MultilineNoTextInputType();
 
   @override
--- a/lib/web_ui/lib/src/engine/text_editing/input_type.dart
+++ b/lib/web_ui/lib/src/engine/text_editing/input_type.dart
@@ -14,7 +14,7 @@ import '../dom.dart';
 abstract class EngineInputType {
   const EngineInputType();
 
-  static EngineInputType fromName(String name, {bool isDecimal = false}) {
+  static EngineInputType fromName(String name, {bool isDecimal = false, bool forceMultiline = false}) {
     switch (name) {
       case 'TextInputType.number':
         return isDecimal ? decimal : number;
@@ -27,7 +27,7 @@ abstract class EngineInputType {
       case 'TextInputType.multiline':
         return multiline;
       case 'TextInputType.none':
-        return none;
+        return forceMultiline ? multilineNone : none;
       case 'TextInputType.text':
       default:
         return text;
@@ -36,6 +36,9 @@ abstract class EngineInputType {
 
   /// No text input.
   static const NoTextInputType none = NoTextInputType();
+
+  /// Multi-line no text input.
+  static const MultilineNoTextInputType multilineNone = MultilineNoTextInputType();
 
   /// Single-line text input type.
   static const TextInputType text = TextInputType();
@@ -92,6 +95,14 @@ class NoTextInputType extends EngineInputType {
 
   @override
   String get inputmodeAttribute => 'none';
+}
+
+/// Multi-line no text input.
+class MultilineNoTextInputType extends EngineInputType {
+  const MultilineNoTextInputType();
+
+  @override
+  String? get inputmodeAttribute => 'none';
 
   @override
   DomHTMLElement createDomElement() => createDomHTMLTextAreaElement();
--- a/lib/web_ui/lib/src/engine/text_editing/input_type.dart
+++ b/lib/web_ui/lib/src/engine/text_editing/input_type.dart
@@ -92,6 +92,9 @@ class NoTextInputType extends EngineInputType {
 
   @override
   String get inputmodeAttribute => 'none';
+
+  @override
+  DomHTMLElement createDomElement() => createDomHTMLTextAreaElement();
 }
 
 /// Single-line text input type.
--- a/lib/web_ui/lib/src/engine/text_editing/input_type.dart
+++ b/lib/web_ui/lib/src/engine/text_editing/input_type.dart
@@ -97,6 +97,7 @@ class NoTextInputType extends EngineInputType {
   String get inputmodeAttribute => 'none';
 }
 
+/// See: https://github.com/flutter/flutter/issues/125875
 /// Multi-line no text input from system virtual keyboard.
 ///
 /// Use this for inputting multiple lines with a customized keyboard.
@@ -104,12 +105,12 @@ class NoTextInputType extends EngineInputType {
 /// When Flutter uses a custom virtual keyboard, it sends [TextInputType.none]
 /// with a [forceMultiline] flag to block the system virtual keyboard.
 ///
-/// For [MultilineNoTextInputType] (mapped to [TextInputType.none] with 
-/// [forceMultiline] = true), it creates a <textarea> element with the 
+/// For [MultilineNoTextInputType] (mapped to [TextInputType.none] with
+/// [forceMultiline] = true), it creates a <textarea> element with the
 /// inputmode="none" attribute.
 ///
-/// For [NoTextInputType] (mapped to [TextInputType.none] with 
-/// [forceMultiline] = false), it creates an <input> element with the 
+/// For [NoTextInputType] (mapped to [TextInputType.none] with
+/// [forceMultiline] = false), it creates an <input> element with the
 /// inputmode="none" attribute.
 class MultilineNoTextInputType extends EngineInputType {
   const MultilineNoTextInputType();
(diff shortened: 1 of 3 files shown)


lib/web_ui/lib/src/engine/text_editing/input_type.dart
```diff
@@ -92,6 +92,9 @@ class NoTextInputType extends EngineInputType {
 
   @override
   String get inputmodeAttribute => 'none';
+
+  @override
+  DomHTMLElement createDomElement() => createDomHTMLTextAreaElement();
```
mdebbar, line 97: So `TextInputType.none` is always treated as multiline? I don't think that's correct. cc @Renzo-Olivares for ideas on how to distinguish between single-line and multi-line modes when using `TextInputType.none`.

lib/web_ui/lib/src/engine/text_editing/text_editing.dart
```diff
@@ -1438,7 +1438,7 @@ abstract class DefaultTextEditingStrategy with CompositionAwareMixin implements
       if (event.keyCode == _kReturnKeyCode) {
         onAction!(inputConfiguration.inputAction);
         // Prevent the browser from inserting a new line when it's not a multiline input.
-        if (inputConfiguration.inputType is! MultilineInputType) {
+        if (inputConfiguration.inputType is! MultilineInputType && inputConfiguration.inputType is! NoTextInputType) {
```
mdebbar, line 1441: If we end up doing this, we will need a test to make sure it doesn't break in the future. You can take a look here for an example: https://github.com/flutter/engine/blob/4a14ee8072a3d3b12b830ca996fd656a6bca774b/lib/web_ui/test/engine/text_editing_test.dart#L421

mdebbar (commented):
Thanks for the contribution!

It makes sense to fix this bug, but I'm not sure if we should always treat `TextInputType.none` as multiline.

lib/web_ui/lib/src/engine/text_editing/input_type.dart
htoor3: Agreed here.  What's the reasoning behind setting `none` as an input type for [custom text input controls](https://github.com/flutter/flutter/blob/master/packages/flutter/lib/src/services/text_input.dart#L2229) in the framework?  Could we just make the determination at the framework level (via `maxLines > 1` or something similar) and just send an explicit input type to the engine?

Satsrag: I agreed to solve via framework level.
However, I think, the main reason for sending `none` as an input type is to disable the soft keyboard display. 
So, we may add a property (such as disableSoftKeyboard) to the class `InputConfiguration`. Then we will add code to all platforms to handle this property to disable soft keyboard display.

Renzo-Olivares: Agreeing with @Satsrag, I think we need to continue setting the `inputMode` of the web dom element (input/textarea) to `none` to block the virtual keyboard when a custom text input control is provided. This means we can't conditionally send an explicit input type based on `maxLines` (unless we add more input types).

Adding a flag to `TextInputConfiguration` as suggested by @Satsrag sounds reasonable, but `disableSoftKeyboard` sounds like it would accomplish the same thing as passing [`TextInputType.none`](https://api.flutter.dev/flutter/services/TextInputType/none-constant.html). Maybe what we need is a `forceMultiline/isMultiline` parameter which sets its value based on the `maxLines` field. Then the web engine can use this to decide whether to use the  `input/textarea`.

Satsrag: Yes, I agree with @Renzo-Olivares.

lib/web_ui/lib/src/engine/text_editing/text_editing.dart
Satsrag: Added test for `EngineInputType.multilineNone`. Is there a need for more tests? Thank you.

lib/web_ui/lib/src/engine/text_editing/input_type.dart
```diff
@@ -94,6 +97,17 @@ class NoTextInputType extends EngineInputType {
   String get inputmodeAttribute => 'none';
 }
 
+/// Multi-line no text input.
```
htoor3, line 100: We should probably have a clear comment description of why this is needed or in what case this type might be used

lib/web_ui/lib/src/engine/text_editing/input_type.dart
```diff
@@ -94,6 +97,30 @@ class NoTextInputType extends EngineInputType {
...
+/// When Flutter uses a custom virtual keyboard, it sends [TextInputType.none]
+/// with a [forceMultiline] flag to block the system virtual keyboard.
+///
+/// For [MultilineNoTextInputType] (mapped to [TextInputType.none] with 
+/// [forceMultiline] = true), it creates a <textarea> element with the 
+/// inputmode="none" attribute.
+///
+/// For [NoTextInputType] (mapped to [TextInputType.none] with 
+/// [forceMultiline] = false), it creates an <input> element with the 
+/// inputmode="none" attribute.
```
Satsrag, line 113: @htoor3 
Is the comment clear now? I'm sorry, I'm not good at English. If it's still unclear, could you please write it for me to copy? Thank you.

htoor3: Thank you @Satsrag! It looks good to me, I would just add another line that points to the issue this is fixing like
```
/// See: https://github.com/flutter/flutter/issues/125875
```

Satsrag:
I'm sorry about asking for review again and again. But this time, I think that all work is done. Ask for a review again. 
In [framework PR](https://github.com/flutter/flutter/pull/140356), `forceMultiline` was renamed to `isMultiline`. So here I was renamed to `isMultiline` and added tests.

lib/web_ui/lib/src/engine/text_editing/input_type.dart
```diff
@@ -14,7 +14,7 @@ import '../dom.dart';
 abstract class EngineInputType {
   const EngineInputType();
 
-  static EngineInputType fromName(String name, {bool isDecimal = false}) {
+  static EngineInputType fromName(String name, {bool isDecimal = false, bool isMultiline= false}) {
```
htoor3, line 17: nit: change `isMultiline= false` to `isMultiline = false`

htoor3 (approved):
Thank you for this! This lgtm.  Before merging, let's verify the framework approach is the correct one with @Renzo-Olivares

lib/web_ui/lib/src/engine/text_editing/input_type.dart
Satsrag: Sorry about this. Now I have fixed it. Thank you!

Satsrag: @htoor3, Happy New Year! [The framework PR](https://github.com/flutter/flutter/pull/140356) has been merged. Is it time to merge this PR?

htoor3:
lgtm, thank you @Satsrag for your contribution!

Satsrag:
@mdebbar, Please have a review.

Satsrag:
@htoor3 when this PR will be merged?

htoor3:
Let's get a review from @mdebbar and then we can get this merged

lib/web_ui/lib/src/engine/text_editing/text_editing.dart
```diff
@@ -1280,7 +1281,7 @@ abstract class DefaultTextEditingStrategy with CompositionAwareMixin implements
       activeDomElement.setAttribute('type', 'password');
     }
 
-    if (config.inputType == EngineInputType.none) {
+    if (config.inputType == EngineInputType.none || config.inputType == EngineInputType.multilineNone) {
       activeDomElement.setAttribute('inputmode', 'none');
     }
```
mdebbar, line 1286: Small suggestion to simplify this a bit:
```suggestion
    if (config.inputType.inputmodeAttribute == 'none') {
      activeDomElement.setAttribute('inputmode', 'none');
    }
```

lib/web_ui/lib/src/engine/text_editing/text_editing.dart
```diff
@@ -1462,7 +1463,7 @@ abstract class DefaultTextEditingStrategy with CompositionAwareMixin implements
       if (event.keyCode == _kReturnKeyCode) {
         onAction!(inputConfiguration.inputAction);
         // Prevent the browser from inserting a new line when it's not a multiline input.
-        if (inputConfiguration.inputType is! MultilineInputType) {
+        if (inputConfiguration.inputType is! MultilineInputType && inputConfiguration.inputType is! MultilineNoTextInputType) {
```
mdebbar, line 1466: Small suggestion to simplify this a bit:
```suggestion
        if (inputConfiguration.inputType is! MultilineInputType) {
```
and make it `class MultilineNoTextInputType extends MultilineInputType { ... }`.

mdebbar (approved):
Thanks for helping make Flutter better!

The PR looks good to me with some minor suggestions.

Satsrag:
> Thanks for helping make Flutter better!
> 
> The PR looks good to me with some minor suggestions.

@mdebbar Thank you! Fixed according to your suggestion.

PR #45524, opened 2023-09-07
[Impeller] Gaussian blur: Remove lingering BlurStyle vertex data/uniforms.

bdero:
Follow-up for https://github.com/flutter/engine/pull/45520.

Missed a few things... These were getting defaulted thanks to impeller's header gen but don't have any purpose now that we've removed the old blur style impl.


jonahwilliams (approved):
Woops, LGTM

PR #45530, opened 2023-09-07
[Impeller] Aiks image filters; allow setting effect transforms after FilterContents instantiation.

bdero:
These are API improvements on the way to solving https://github.com/flutter/flutter/issues/131182.

* Remove `effect_transform` and `is_subpass` from the static `FilterContents` factories. These are for internal use only.
* Replace `impeller::Paint` filter procs with Aiks `ImageFilter` factories similar to `ColorFilter`. This gives Aiks a simple interface for constructing filters and prevents possible state cloning bugs when copying filters into `EntityPass`.
* Allow for setting filter inputs, setting the effect transform, and enabling subpass mode on a filter chain after it has been instantiated. This will allow us to sample coverage in `EntityPass` without rendering any snapshots using rect filter inputs.

### DIFF
--- a/impeller/entity/contents/filters/inputs/filter_input.h
+++ b/impeller/entity/contents/filters/inputs/filter_input.h
@@ -77,6 +77,12 @@ class FilterInput {
   ///         of `inputs`.
   /// @see    `FilterInput::IsLeaf`
   virtual void SetLeafInputs(const FilterInput::Vector& inputs);
+
+  /// @brief  Sets the effect transform of filter inputs.
+  virtual void SetEffectTransform(const Matrix& matrix);
+
+  /// @brief  Turns on subpass mode for filter inputs.
+  virtual void SetIsForSubpass(bool is_for_subpass);
 };
 
 }  // namespace impeller
--- a/impeller/entity/entity_pass.cc
+++ b/impeller/entity/entity_pass.cc
@@ -604,13 +604,15 @@ EntityPass::EntityResult EntityPass::GetEntityForElement(
       // right thing.
       return EntityPass::EntityResult::Failure();
     }
-
-    element_entity.SetCapture(capture.CreateChild("Entity (Subpass texture)"));
+    Capture subpass_texture_capture =
+        capture.CreateChild("Entity (Subpass texture)");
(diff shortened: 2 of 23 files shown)


impeller/entity/contents/filters/inputs/filter_input.h
```diff
@@ -77,6 +77,12 @@ class FilterInput {
   ///         of `inputs`.
   /// @see    `FilterInput::IsLeaf`
   virtual void SetLeafInputs(const FilterInput::Vector& inputs);
+
+  /// @brief  Sets the effect transform of filter inputs.
+  virtual void SetEffectTransform(const Matrix& matrix);
+
+  /// @brief  Turns on subpass mode for filter inputs.
+  virtual void SetIsForSubpass(bool is_for_subpass);
```
jonahwilliams, line 85: Nit: can you add more context on what subpass mode implies?

jonahwilliams (approved):
LGTM

impeller/entity/entity_pass.cc
```diff
@@ -589,9 +589,7 @@ EntityPass::EntityResult EntityPass::GetEntityForElement(
 
     auto offscreen_texture_contents =
         subpass->delegate_->CreateContentsForSubpassTarget(
-            subpass_texture,
-            Matrix::MakeTranslation(Vector3{-global_pass_position}) *
-                subpass->xformation_);
+            subpass_texture, subpass->xformation_.Basis());
```
jonahwilliams, line 592: Is this a different bug or a new bug?

bdero: Investigating this a bit to remind myself about how we got here. It's kind of peculiar that this was working before.

impeller/entity/contents/filters/inputs/filter_input.h
bdero: Done.

impeller/entity/entity_pass.cc
bdero: I reverted this little change -- it turns out the test was actually behaving correctly and just needed to be adjusted. Before this PR, the effect_transform wasn't being set by the ImageFilter proc defined in the test.

PR #45531, opened 2023-09-07
[fuchsia] Restructure Flatland vsync loop

uysalere:
This CL restructures Flatland vsync loop to fire for each vsync instead of each OnNextFrameBegin. As shown in the traces attached to the bug, the current implementation of firing callbacks on each OnNextFrameBegin causes skips when Flutter has longer draw calls. By scheduling frames in between, we are increasing the chance of sending one before the latch point. OnNextFrameBegin is now used to keep track of present credits and future presentation times as well as when to start frame, replacing the need for max_frames_in_flight and vsync_offset fields.

Bug: b/296272449

### DIFF
--- a/shell/platform/fuchsia/flutter/engine.cc
+++ b/shell/platform/fuchsia/flutter/engine.cc
@@ -354,8 +354,6 @@ void Engine::Initialize(
        view_protocols = std::move(view_protocols),
        request = parent_viewport_watcher.NewRequest(),
        view_ref_pair = std::move(view_ref_pair),
-       max_frames_in_flight = product_config.get_max_frames_in_flight(),
-       vsync_offset = product_config.get_vsync_offset(),
        software_rendering = product_config.software_rendering()]() mutable {
         if (software_rendering) {
           surface_producer_ = std::make_shared<SoftwareSurfaceProducer>();
@@ -365,8 +363,7 @@ void Engine::Initialize(
 
         flatland_connection_ = std::make_shared<FlatlandConnection>(
             thread_label_, std::move(flatland),
-            std::move(session_error_callback), [](auto) {},
-            max_frames_in_flight, vsync_offset);
+            std::move(session_error_callback), [](auto) {});
 
         fuchsia::ui::views::ViewIdentityOnCreation view_identity = {
             .view_ref = std::move(view_ref_pair.second),
--- a/shell/platform/fuchsia/flutter/flatland_connection.cc
+++ b/shell/platform/fuchsia/flutter/flatland_connection.cc
@@ -13,11 +13,10 @@ namespace flutter_runner {
 
 namespace {
 
-fml::TimePoint GetNextPresentationTime(fml::TimePoint now,
-                                       fml::TimePoint next_presentation_time) {
-  return now > next_presentation_time
-             ? now + flutter_runner::kDefaultFlatlandPresentationInterval
-             : next_presentation_time;
+// Helper function for traces.
+double DeltaFromNowInNanoseconds(const fml::TimePoint& now,
+                                 const fml::TimePoint& time) {
+  return (time - now).ToNanoseconds();
 }
 
 }  // namespace
@@ -26,9 +25,7 @@ FlatlandConnection::FlatlandConnection(
     std::string debug_label,
     fuchsia::ui::composition::FlatlandHandle flatland,
     fml::closure error_callback,
-    on_frame_presented_event on_frame_presented_callback,
-    uint64_t max_frames_in_flight,
-    fml::TimeDelta vsync_offset)
+    on_frame_presented_event on_frame_presented_callback)
     : flatland_(flatland.Bind()),
       error_callback_(error_callback),
       on_frame_presented_callback_(std::move(on_frame_presented_callback)) {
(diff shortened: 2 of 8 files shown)


shell/platform/fuchsia/flutter/flatland_connection.cc
```diff
@@ -13,22 +13,22 @@ namespace flutter_runner {
...
-             : next_presentation_time;
+// Helper function for traces.
+double DeltaFromNowInNanoseconds(const fml::TimePoint& now,
+                                 const fml::TimePoint& time) {
+  return (time - now).ToNanoseconds();
 }
 
 }  // namespace
 
+// This function takes in all relevant information to determining when should
```
arbreng, line 24: Nit: Does this comment belong on the constructor or somewhere else?

shell/platform/fuchsia/flutter/flatland_connection.cc
```diff
@@ -127,30 +132,55 @@ void FlatlandConnection::OnError(
...
-    if (threadsafe_state_.fire_callback_) {
-      threadsafe_state_.fire_callback_(
-          /*frame_start=*/fml::TimePoint::Now(),
-          /*frame_target=*/next_presentation_time);
-      threadsafe_state_.fire_callback_ = nullptr;
-    } else {
-      threadsafe_state_.on_next_frame_pending_ = true;
-    }
-    threadsafe_state_.next_presentation_time_ = next_presentation_time;
+  // Update vsync_interval_ by calculating.
```
arbreng, line 150: Calculating how?  For future readers

shell/platform/fuchsia/flutter/flatland_connection.cc
```diff
@@ -127,30 +132,55 @@ void FlatlandConnection::OnError(
...
+  FML_CHECK(values.has_future_presentation_infos() &&
```
arbreng, line 151: Should we CHECK here?  Is it possible to warn instead and then use an approximation?

shell/platform/fuchsia/flutter/engine.cc
```diff
@@ -354,8 +354,6 @@ void Engine::Initialize(
        view_protocols = std::move(view_protocols),
        request = parent_viewport_watcher.NewRequest(),
        view_ref_pair = std::move(view_ref_pair),
-       max_frames_in_flight = product_config.get_max_frames_in_flight(),
```
arbreng, line 357: Nice, I think this means we can get rid of the product config file soon.

Please remove these fields from the flutter_runner_product_config.h

shell/platform/fuchsia/flutter/flatland_connection.h
```diff
@@ -89,10 +93,13 @@ class FlatlandConnection final {
...
   struct {
     std::mutex mutex_;
-    FireCallbackCallback fire_callback_;
-    bool first_present_called_ = false;
-    bool on_next_frame_pending_ = false;
-    fml::TimePoint next_presentation_time_;
+    bool first_feedback_received_ = false;
+    FireCallbackCallback pending_fire_callback_;
+    uint32_t present_credits_ = 1;
+    std::queue<fml::TimePoint> next_presentation_times_;
```
arbreng, line 99: Nit: queue first, then other fields, then boolean last (more readable, and fields pack into structure better)

shell/platform/fuchsia/flutter/flatland_connection.cc
```diff
@@ -127,30 +132,55 @@ void FlatlandConnection::OnError(
...
+  // Thread contention may result in OnNextFrameBegin() being called after the
+  // presentation time. Ignore these outliers.
+  if (vsync_offset > fml::TimeDelta::Zero()) {
+    threadsafe_state_.vsync_offset_ = vsync_offset;
+  }
+
+  // Throttle case.
+  if (threadsafe_state_.pending_fire_callback_ &&
+      threadsafe_state_.present_credits_ > 0) {
+    RunVsyncCallback(now, threadsafe_state_.pending_fire_callback_);
```
arbreng, line 182: Make a note here or below (at the definition of RunVsyncCallback) this is run on the Raster thread, but it is safe b/c the Vsync callback will be posted to the UI thread

arbreng (approved):
LGTM % comments

shell/platform/fuchsia/flutter/engine.cc
uysalere: Done.

shell/platform/fuchsia/flutter/flatland_connection.cc
uysalere: Ahh leftover comment from an old helper function that is removed now. Erased it.

shell/platform/fuchsia/flutter/flatland_connection.cc
uysalere: Added it.

shell/platform/fuchsia/flutter/flatland_connection.cc
uysalere: Added a warning in case.

shell/platform/fuchsia/flutter/flatland_connection.cc
uysalere: Added to the definition of RunVsyncCallback.

shell/platform/fuchsia/flutter/flatland_connection.h
uysalere: Done.

PR #45558, opened 2023-09-07
Document and provide small cleanups to `CommandPoolVk`.

matanlurey:
I was reviewing this code for the first time before implementing https://github.com/flutter/flutter/issues/133198.

- Removed unused public method `GetGraphicsCommandPool()`.
- Removed forward references by including the right references [^1].
- Tried to document the contracts of various methods that return invalid or disconnected objects.

If I made any mistakes, please feel free to point them out as this is how I'm learning how this stuff works 😄 

/cc @jonahwilliams 

[^1]: [Avoid using forward declarations where possible](https://google.github.io/styleguide/cppguide.html#Forward_Declarations). Instead, include the headers you need.

### DIFF
--- a/impeller/renderer/backend/vulkan/command_pool_vk.h
+++ b/impeller/renderer/backend/vulkan/command_pool_vk.h
@@ -76,7 +76,7 @@ class CommandPoolVK final {
   ///
   /// @param[in]  buffer  The |vk::CommandBuffer| to collect.
   ///
-  /// @note       This method is a NOP if a different thread created the pool.
+  /// @note       This method is a noop if a different thread created the pool.
   ///
   /// @see        |GarbageCollectBuffersIfAble|
   void CollectGraphicsCommandBuffer(vk::UniqueCommandBuffer buffer);
--- a/impeller/renderer/backend/vulkan/command_pool_vk.h
+++ b/impeller/renderer/backend/vulkan/command_pool_vk.h
@@ -104,7 +104,7 @@ class CommandPoolVK final {
   /// then they are reset and made available to future calls to
   /// |CreateGraphicsCommandBuffer|.
   ///
-  /// @note       This method is a NOP if a different thread created the pool.
+  /// @note       This method is a noop if a different thread created the pool.
   void GarbageCollectBuffersIfAble() IPLR_REQUIRES(buffers_to_collect_mutex_);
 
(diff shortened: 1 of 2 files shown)


impeller/renderer/backend/vulkan/command_pool_vk.h
```diff
@@ -5,36 +5,80 @@
...
   vk::UniqueCommandBuffer CreateGraphicsCommandBuffer();
 
+  /// @brief      Collects the given |vk::CommandBuffer| for recycling.
+  ///
+  /// The given |vk::CommandBuffer| will be recycled (reused) in the future when
+  /// |CreateGraphicsCommandBuffer| is called.
+  ///
+  /// @param[in]  buffer  The |vk::CommandBuffer| to collect.
+  ///
+  /// @note       This method is a NOP if a different thread created the pool.
```
gaaclarke, line 79: Nit: we use "noop" elsewhere, plus it's more fun to say.

gaaclarke (approved):

impeller/renderer/backend/vulkan/command_pool_vk.h
matanlurey: Agreed :). Done.

Hixie:
test-exempt: documentation and code removal

impeller/renderer/backend/vulkan/command_pool_vk.h
```diff
@@ -5,36 +5,80 @@
 #pragma once
 
 #include <memory>
-#include <set>
 #include <thread>
 
 #include "flutter/fml/macros.h"
 #include "impeller/base/thread.h"
+#include "impeller/renderer/backend/vulkan/context_vk.h"
```
jonahwilliams, line 12: The reason we generally do forward declarations is to avoid circular dependencies in header files (and sometimes to reduce compilation time but that is marginal). As long as it compiles that should be okay, but you might hit cases where you can't remove a forward reference.

jonahwilliams (approved):
Lgtm

impeller/renderer/backend/vulkan/command_pool_vk.h
matanlurey: A good to know. I think these are all in the same source set but I'll have CI double check me :)

matanlurey:
@chinmaygarde I'm going to merge, but if you have any post-submit comments or suggestions please leave then and I'll get to it in post!

PR #45566, opened 2023-09-08
[Impeller] Proof of concept: `CommandPoolRecyclerVK`

matanlurey:
I am pretty sure this is (a) not thread-safe and (b) won't actually work as written.

In particular I want to discuss `::Recycle()` and differences with https://github.com/flutter/engine/pull/45179/files.

Mostly adding as a draft to discuss with @jonahwilliams tomorrow :)

### DIFF
--- a/fml/closure.h
+++ b/fml/closure.h
@@ -32,6 +32,7 @@ using closure = std::function<void()>;
 class ScopedCleanupClosure {
  public:
   ScopedCleanupClosure() = default;
+  ScopedCleanupClosure(ScopedCleanupClosure&&) = default;
 
   explicit ScopedCleanupClosure(const fml::closure& closure)
       : closure_(closure) {}
--- a/impeller/renderer/backend/vulkan/BUILD.gn
+++ b/impeller/renderer/backend/vulkan/BUILD.gn
@@ -39,6 +39,8 @@ impeller_component("vulkan") {
     "command_buffer_vk.h",
     "command_encoder_vk.cc",
     "command_encoder_vk.h",
+    "command_pool_recycler_vk.cc",
+    "command_pool_recycler_vk.h",
     "command_pool_vk.cc",
     "command_pool_vk.h",
     "compute_pass_vk.cc",
--- a/impeller/renderer/backend/vulkan/command_pool_recycler_vk.cc
+++ b/impeller/renderer/backend/vulkan/command_pool_recycler_vk.cc
@@ -1,0 +1,62 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+#include "impeller/renderer/backend/vulkan/command_pool_recycler_vk.h"
+#include "fml/closure.h"
+#include "impeller/renderer/backend/vulkan/command_pool_vk.h"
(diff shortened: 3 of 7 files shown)


jonahwilliams (commented):
A few notes:

CommandPoolVK::Reset isn't actually what you want to use, its this driver method on the device holder: https://github.com/flutter/engine/pull/45179/files#diff-b0fef1c24c7391b60f10ced0a691353e0d1c677a1dd028ae5750367645da3d8fR149

You need to specifically keep the CommandPool alive by giving a strong reference to the TrackedObjectsVK: https://github.com/flutter/engine/blob/main/impeller/renderer/backend/vulkan/command_encoder_vk.cc#L15

Rather than using the ScopedCleanupClosure, you should use the destructor of some object wrapping the cmdpool, if possible.

To ensure that destructor will run, at the end of a frame (or start of the next one), you will need to remove the cmd pool wrapping object from the TLS map: 

https://github.com/flutter/engine/pull/45179/files#diff-b0fef1c24c7391b60f10ced0a691353e0d1c677a1dd028ae5750367645da3d8fR30

matanlurey:
> CommandPoolVK::Reset isn't actually what you want to use, its this driver method on the device holder: https://github.com/flutter/engine/pull/45179/files#diff-b0fef1c24c7391b60f10ced0a691353e0d1c677a1dd028ae5750367645da3d8fR149

Ah, the _other_ reset :). Thanks!

> You need to specifically keep the CommandPool alive
> ...
> To ensure that destructor will run

Good tips, thanks. I'll give you a ring when I have the next iteration up!

jonahwilliams:
So you probably want to not introduce a new class like CommandPoolRecyclerVK, instead the current object the wraps the Vulkan cmd pool should be responsible it.

matanlurey:
> So you probably want to not introduce a new class like CommandPoolRecyclerVK, instead the current object the wraps the Vulkan cmd pool should be responsible it.

I was initially trying to not modify the existing code at all so in the same branch I could try the old code and new code (i.e. without switching branches and ninja building).

It also feels .. weird (?) to provide tons of contextual information to `CommandPoolVK`, which makes the class do a _lot_ of things (it's responsible for wrapping and creating buffers, which makes sense) and need references to `ContextVK`.

That being said, no strong opinion!

jonahwilliams:
But keep in mind we don't need to recycle CommandPoolVK, that is an impeller wrapper class. The entire goal of that class is to manage access and the lifecycle of the actual cmd pool, which is a field on that class. In fact, if you do this right, CommandPoolVK will literally have less code in it when you're done.

jonahwilliams (commented):
What I would encourage you to do is 

1. Make the existing class (or wrap the actual cmd pool) in a resource type.
2. Create some mutable storage std::vector<vk::UniqueCommandPool> on context_vk
3. Have the destructor of the resource type call reset on the cmd pool and push it into context vk.
4. Have the existing CmdPool type ask the context if there are existing cmd pools before creating a new one.
5. (Tricky) Give each cmd buffer created from the cmd pool a strong reference to the cmd pool.
6. At the end of the frame, remove the CommandPoolVK from the TLS map.

What should happen is, once removed from the TLS map - the number of active shared_ptrs should correspond to the number of uncompleted buffers. Once all the buffers are done, all of their references to the cmd pool will be dropped, and the CmdPoolVK destructor will run. This allows the reset code to run on the resource manager thread.

PR #45579, opened 2023-09-08
[Impeller] use public Skia API to look up glyph emoji/colr status

jonahwilliams:
Fixes https://github.com/flutter/flutter/issues/112005

Now that the text frames are constructed closer to where the SkParagraph is, we can use the public API for lookup up emoji/colr font status.

I moved the old functionality into a "test only" method that we can use for now to look up this data in testing. Not sure what we should do there long term.


jonahwilliams:
Looks like this doesn't actually work. A number of SkTextBlobs containing nothing but "regular" glyphs are claiming they contain bitmap or colr fonts, and as a result we don't render them correctly:

flutter_04

flutter_05

PR #45591, opened 2023-09-08
Remove some of our hacks around JSPromise now that we have better APIs.

eyebrowsoffire:
Our JSPromise hackery is causing some issues with the new dart roll. We should just use the `JSPromise` and `JSFunction` support to simplify this. Note that I still have to do *some* hackery to construct `JSPromise` objects and to invoke `JSFunction` objects, and eventually we'll probably be able to simplify this even more once those APIs are baked into `dart:js_interop`

### DIFF
--- a/lib/web_ui/lib/src/engine/app_bootstrap.dart
+++ b/lib/web_ui/lib/src/engine/app_bootstrap.dart
@@ -2,7 +2,7 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
-import 'package:js/js_util.dart' show allowInterop;
+import 'dart:js_interop';
 
 import 'configuration.dart';
 import 'js_interop/js_loader.dart';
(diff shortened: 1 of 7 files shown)


lib/web_ui/lib/src/engine/app_bootstrap.dart
```diff
@@ -40,33 +40,26 @@ class AppBootstrap {
       // This is a convenience method that lets the programmer call "autoStart"
       // from JavaScript immediately after the main.dart.js has loaded.
       // Returns a promise that resolves to the Flutter app that was started.
-      autoStart: allowInterop(() => futureToPromise(() async {
+      autoStart: (() => futureToPromise(() async {
         await autoStart();
         // Return the App that was just started
-        return _prepareFlutterApp();
-      }())),
+        return _prepareFlutterApp() as JSAny;
```
srujzs, line 46: nit: This (and similar casts elsewhere) is unneeded if the return type is a `@staticInterop` type.

lib/web_ui/lib/src/engine/js_interop/js_loader.dart
```diff
@@ -45,22 +42,11 @@ extension FlutterLoaderExtension on FlutterLoader {
...
 abstract class FlutterEngineInitializer{
   external factory FlutterEngineInitializer({
-    required InitializeEngineFn initializeEngine,
-    required ImmediateRunAppFn autoStart,
+    required JSFunction initializeEngine,
+    required JSFunction autoStart,
   });
 }
 
-/// Typedef for the function that initializes the flutter engine.
```
srujzs, line 53: It's up to you if you want to make this typed a bit better by having a non-`external` function that takes in the typedefs, does the `toJS` conversions, and then forwards them to a (private) `external` factory function.

lib/web_ui/lib/src/engine/js_interop/js_promise.dart
```diff
@@ -11,40 +11,27 @@ import 'package:js/js_util.dart' as js_util;
...
-/// Type-safe JS Promises
 @JS('Promise')
-@staticInterop
-abstract class Promise<T extends Object?> {
-  /// A constructor for a JS promise
-  external factory Promise(PromiseExecutor<T> executor);
-}
+external JSAny get _promiseConstructor;
+
+JSPromise createPromise(JSFunction executor) =>
```
srujzs, line 21: Why not keep the old interface and have `Promise` take in a `JSFunction` instead?

srujzs (approved):

lib/web_ui/lib/src/engine/app_bootstrap.dart
srujzs: Nevermind, `futureToPromise` returns a `T extends JSAny` so this is fine.

lib/web_ui/lib/src/engine/js_interop/js_promise.dart
eyebrowsoffire: I really was annoyed by our redundant declarations of types that already exist in `dart:js_interop`. We actually redeclared the promise type twice. Blech. Ideally I would just declare an extension method, but you can't do that for factories, so I either have to use `js_util` to call the constructor directly (as I've done here), or I have to redeclare the class and then cast it to `JSPromise`. Both seem kind of gross.

lib/web_ui/lib/src/engine/app_bootstrap.dart
eyebrowsoffire: Looks like I could solve this by making `FlutterApp` and `FlutterAppRunner` extend `JSObject`.

lib/web_ui/lib/src/engine/js_interop/js_loader.dart
eyebrowsoffire: Hmm, I guess maybe it's better to keep a little bit of type safety here.

lib/web_ui/lib/src/engine/js_interop/js_promise.dart
srujzs: Makes sense. The right fix for this is for me to add a constructor to `JSPromise`. If we could use extension types, then we could subtype `JSPromise` and use that to avoid the cast, but still not quite as good as the constructor.

srujzs (approved):

PR #45595, opened 2023-09-08
Adds a GN template that invokes run_tests.py

zanderso:


goderbauer:
(triage): I spoke to @zanderso last week and he said the previous comment is still accurate.

johnmccutchan:
Could you dump the output of "gn desc" for these new rules somewhere? I'm asking because we will be using "gn desc" to enumerate all build targets and we need some way to identify the targets that use these new templates/rules.

PR #45598, opened 2023-09-08
Fix script

godofredoc:
Makes mac framework readable and executable by all.

Bug: https://github.com/flutter/flutter/issues/132725

## Pre-launch Checklist

- [X] I added new tests to check the change I am making or feature I am adding, or the PR is [test-exempt]. See [testing the engine] for instructions on writing and running engine tests.
- [X] I signed the [CLA].

### DIFF
--- a/sky/tools/create_macos_framework.py
+++ b/sky/tools/create_macos_framework.py
@@ -92,7 +92,6 @@ def main():
   subprocess.check_call([
       'lipo', arm64_dylib, x64_dylib, '-create', '-output', fat_framework_binary
   ])
-  process_framework(dst, args, fat_framework, fat_framework_binary)
 
   # Add group and other readability to all files.
   versions_path = os.path.join(fat_framework, 'Versions')
@@ -107,6 +106,8 @@ def main():
                                       stdin=find_subprocess.stdout)
   find_subprocess.wait()
   xargs_subprocess.wait()
+
+  process_framework(dst, args, fat_framework, fat_framework_binary)
 
   return 0
 
--- a/sky/tools/create_macos_framework.py
+++ b/sky/tools/create_macos_framework.py
@@ -95,13 +95,11 @@ def main():
   process_framework(dst, args, fat_framework, fat_framework_binary)
 
   # Add group and other readability to all files.
-  subprocess.check_call([
-      'chmod', '-R', 'og+r',
-      os.path.join(fat_framework, 'Versions')
-  ])
+  versions_path = os.path.join(fat_framework, 'Versions')
+  subprocess.check_call(['chmod', '-R', 'og+r', versions_path])
   # Find all the files below the target dir with owner execute permission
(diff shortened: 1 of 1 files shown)


sky/tools/create_macos_framework.py
```diff
@@ -93,6 +94,16 @@ def main():
   ])
   process_framework(dst, args, fat_framework, fat_framework_binary)
 
+  # Set fat_framework file attributes to read and execute for all.
+  subprocess.check_call([
```
gspencergoog, line 98: It's not really great to make non-script/binary files executable. It could be a security hole if one of them starts with `#!`, for instance

Ideally, you'd take whatever permissions the file already has for owner and give it the same permissions for other and group.

I'd suggest replacing these two calls with something like this:

```python
# Add group and other readability to all files.
subprocess.check_call([
  'chmod', '-R', 'og+r',
  os.path.join(fat_framework, 'Versions')
])
# Find all the files below the target dir with owner execute permission
find_subprocess = subprocess.Popen(["find", os.path.join(fat_framework, 'Versions', 'A', 'FlutterMacOS'), "-type", "f", "-perm", "-100", "-print0"], stdout=subprocess.PIPE)
# Add execute permission for other and group for all files that had it for owner.
xargs_subprocess = subprocess.Popen(["xargs", "-0", "chmod", "og+x"], stdin=find_subprocess.stdout)
find_subprocess.wait()
xargs_subprocess.wait()
```

You might also consider using the "Versions" path in the `find` subprocess instead of just the `FlutterMacOS` subdir, to make sure all the paths get the same treatment.

sky/tools/create_macos_framework.py
```diff
@@ -94,15 +94,21 @@ def main():
...
+      'chmod', '-R', 'og+r',
       os.path.join(fat_framework, 'Versions')
   ])
-  subprocess.check_call([
-      'chmod', '-R', '755',
-      os.path.join(fat_framework, 'Versions', 'A', 'FlutterMacOS')
-  ])
+  # Find all the files below the target dir with owner execute permission
+  find_subprocess = subprocess.Popen([
+      "find", fat_framework, "-perm", "-100", "-print0"
```
gspencergoog, line 104: Actually, I guess to be consistent with the earlier version of the code, it should be `os.path.join(fat_framework, 'Versions')` instead of just `fat_framework`, unless it's OK that it do this for all the files.

If it is OK, then (if it's OK) you could also change the one on line 100 to be just `fat_framework` too.

Just trying to make them consistent.  As it is now, it will add read permission to everything under `Versions`, but execute permission to everything that also has owner execute permission under the parent directory, `FlutterMacOS.framework`, which might be a larger group.

sky/tools/create_macos_framework.py
godofredoc: done

gspencergoog (approved):

PR #45651, opened 2023-09-11
[web] Use DOM to render paragraphs with letter spacing

mdebbar:
<**_Only applies to the HTML renderer_**>

When a paragraph has letter spacing that's greater than 0, we attempt to render each character individually on the canvas, while adding extra gaps between the characters (because the browser's canvas API doesn't support letter spacing).

This works well-ish in English but causes issues with international text. It also makes text rendering very slow.

With this PR, the idea is to force all paragraphs with letter spacing to be rendered to a DOM element.

Fixes https://github.com/flutter/flutter/issues/51234
Fixes https://github.com/flutter/flutter/issues/71220


yjbanov (approved):
LGTM! Prepare for Gold/Skuba failures.

mdebbar:
> LGTM! Prepare for Gold/Skuba failures.

Yeah, I was expecting to see Gold changes, but there's none. So I'm adding a new golden test to cover the issue that this PR is fixing. Test coming soon..

PR #45657, opened 2023-09-11
[iOS] Remove selectionDidChange call in UndoManager

LongCatIsLooong:
Fixes https://github.com/flutter/flutter/issues/133424

The `-[TextInputDelegate selectionDidChange:]` call actually triggers some unwanted keyboard NLP actions that generate a bunch of candidates and automatically accept the first candidate. This causes `-[UITextInput setMarkedText:selection]` to be called with the first candidate and that inserts extraneous characters after the user types certain characters on the iPad software keyboard.

## Pre-launch Checklist

- [ ] I signed the [CLA].

### DIFF
--- a/shell/platform/darwin/ios/framework/Source/FlutterUndoManagerPlugin.mm
+++ b/shell/platform/darwin/ios/framework/Source/FlutterUndoManagerPlugin.mm
@@ -106,9 +106,7 @@ static NSString* const kCanRedo = @"canRedo";
     // will not show up until the next keystroke (or other trigger).
     UITextInputAssistantItem* assistantItem =
         _viewController.engine.textInputPlugin.textInputView.inputAssistantItem;
-    NSArray<UIBarButtonItemGroup*>* leadingGroups = assistantItem.leadingBarButtonGroups;
-    assistantItem.leadingBarButtonGroups = @[];
-    assistantItem.leadingBarButtonGroups = leadingGroups;
+    assistantItem.leadingBarButtonGroups = assistantItem.leadingBarButtonGroups;
   }
   [self undoManager].groupsByEvent = groupsByEvent;
 }
--- a/shell/platform/darwin/ios/framework/Source/FlutterUndoManagerPlugin.mm
+++ b/shell/platform/darwin/ios/framework/Source/FlutterUndoManagerPlugin.mm
@@ -104,9 +104,11 @@ static NSString* const kCanRedo = @"canRedo";
     // This is needed to notify the iPadOS keyboard that it needs to update the
     // state of the UIBarButtons. Otherwise, the state changes to NSUndoManager
     // will not show up until the next keystroke (or other trigger).
-    id<UITextInputDelegate> inputDelegate =
-        _viewController.engine.textInputPlugin.textInputView.inputDelegate;
-    [inputDelegate selectionDidChange:_viewController.engine.textInputPlugin.textInputView];
+    UITextInputAssistantItem* assistantItem =
+        _viewController.engine.textInputPlugin.textInputView.inputAssistantItem;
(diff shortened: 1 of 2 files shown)


hellohuanlin:
> actually triggers some unwanted keyboard NLP actions that generate a bunch of candidates and automatically accept the first candidate. 

Could you explain more about this? Did you see a bar showing up and immediately dismissed?

shell/platform/darwin/ios/framework/Source/FlutterUndoManagerPlugin.mm
```diff
@@ -104,9 +104,11 @@ - (void)setUndoState:(NSDictionary*)dictionary API_AVAILABLE(ios(9.0)) {
     // This is needed to notify the iPadOS keyboard that it needs to update the
```
hellohuanlin, line 104: is this comment still relevant?

shell/platform/darwin/ios/framework/Source/FlutterUndoManagerPlugin.mm
```diff
@@ -104,9 +104,11 @@ - (void)setUndoState:(NSDictionary*)dictionary API_AVAILABLE(ios(9.0)) {
...
     // state of the UIBarButtons. Otherwise, the state changes to NSUndoManager
     // will not show up until the next keystroke (or other trigger).
-    id<UITextInputDelegate> inputDelegate =
-        _viewController.engine.textInputPlugin.textInputView.inputDelegate;
-    [inputDelegate selectionDidChange:_viewController.engine.textInputPlugin.textInputView];
+    UITextInputAssistantItem* assistantItem =
+        _viewController.engine.textInputPlugin.textInputView.inputAssistantItem;
+    NSArray<UIBarButtonItemGroup*>* leadingGroups = assistantItem.leadingBarButtonGroups;
+    assistantItem.leadingBarButtonGroups = @[];
+    assistantItem.leadingBarButtonGroups = leadingGroups;
```
hellohuanlin, line 111: why is it setting to empty, and then setting back? Some comments here will be helpful.

fbcouch: This was a workaround to notify the iPadOS keyboard to update the state of these buttons. It's been awhile, but I believe the behavior I was seeing was that state changes to canUndo/canRedo would not show up until the next keypress – so if I press a letter, the undo button should light up, but it won't until I press another key.

hellohuanlin: Worth trying if it works if we simply do `assistantItem.leandingBarButtonGroups = assistantItem.leandingBarButtonGroups`, in case apple doesn't check for no-ops in the setter.

hellohuanlin: BTW if it's iOS 17 related, feel free to contribute to [the design doc](https://docs.google.com/document/d/1sM3HMv-SQin39yX1aPUU7vtGv7Hcef1Quc3QhRXBl6A/edit?resourcekey=0-SFYD8vmOIkXiXCZvB1Wlcw) which already discusses about several iOS 17 text input issues. It can be a good reference doc for future developers.

LongCatIsLooong: the original issue is reproduced on an iOS 16 device so probably not ios specific. 

I tested removing the intermediate empty array step it appears the setter indeed doesn't skip even when given the same array instance. Updated. I'll file a radar later since the keyboard could just rely on notifications to properly update the undo/redo UI.

hellohuanlin (approved):
LGTM. thanks for the update.

PR #45660, opened 2023-09-11
Revert "Remove some of our hacks around JSPromise now that we have better APIs."

eyebrowsoffire:
Reverts flutter/engine#45591

This is somehow causing some issues with the hot reload tests and blocking engine -> framework rolls. See https://github.com/flutter/flutter/pull/134455


jonahwilliams (approved):
LGTM

PR #45668, opened 2023-09-11
Fix JS interop signatures to use only JS types.

eyebrowsoffire:
This prepares for some upcoming changes to dart2js which will be more strict about what types can be used in a JS interop declaration.

### DIFF
--- a/lib/web_ui/lib/src/engine/js_interop/js_loader.dart
+++ b/lib/web_ui/lib/src/engine/js_interop/js_loader.dart
@@ -8,9 +8,7 @@ library js_loader;
 import 'dart:js_interop';
 
 import 'package:js/js_util.dart' as js_util;
-
-import '../configuration.dart';
-import 'js_promise.dart';
+import 'package:ui/src/engine.dart';
 
 @JS()
 @staticInterop
@@ -34,6 +32,17 @@ extension FlutterLoaderExtension on FlutterLoader {
   bool get isAutoStart => !js_util.hasProperty(this, 'didCreateEngineInitializer');
 }
 
+/// Typedef for the function that initializes the flutter engine.
+/// ///
+/// [JsFlutterConfiguration] comes from `../configuration.dart`. It is the same
+/// object that can be used to configure flutter "inline", through the
+/// (to be deprecated) `window.flutterConfiguration` object.
+typedef InitializeEngineFn = Future<FlutterAppRunner> Function([JsFlutterConfiguration?]);
+
+/// Typedef for the `autoStart` function that can be called straight from an engine initializer instance.
+/// (Similar to [RunAppFn], but taking no specific "runApp" parameters).
+typedef ImmediateRunAppFn = Future<FlutterApp> Function();
+
 // FlutterEngineInitializer
 
 /// An object that allows the user to initialize the Engine of a Flutter App.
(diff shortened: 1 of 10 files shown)


lib/web_ui/lib/src/engine/js_interop/js_loader.dart
```diff
@@ -44,34 +53,34 @@ extension FlutterLoaderExtension on FlutterLoader {
 @anonymous
 @staticInterop
 abstract class FlutterEngineInitializer{
-  external factory FlutterEngineInitializer({
+  factory FlutterEngineInitializer({
     required InitializeEngineFn initializeEngine,
     required ImmediateRunAppFn autoStart,
+  }) => FlutterEngineInitializer._(
+      initializeEngine: (() => futureToPromise(initializeEngine())).toJS,
```
ditman, line 60: Is this swallowing the parameters passed to the initializeEngine fn from JS?

I suspect this needs to be:

```
initializeEngine: (([JsFlutterConfiguration?] args) => futureToPromise(initializeEngine(args))).toJS,
```

(testing)

ditman (commented):
Testing small nit about the configuration of the engine, otherwise LGTM!

lib/web_ui/lib/src/engine/js_interop/js_loader.dart
eyebrowsoffire: Oh dang, you're right. Let me fix that.

lib/web_ui/test/canvaskit/canvaskit_api_test.dart
```diff
@@ -1861,8 +1861,8 @@ void _paragraphTests() {
     });
 
     test('in Chromium-based browsers', () {
-      v8BreakIterator = Object(); // Any non-null value.
-      intlSegmenter = Object(); // Any non-null value.
+      v8BreakIterator = Object().toJSBox; // Any non-null value.
```
mdebbar, line 1864: What does `toJSBox` do?

mdebbar (approved):

lib/web_ui/test/canvaskit/canvaskit_api_test.dart
eyebrowsoffire: In dart2wasm, all dart objects are wasm structs. There is a wasm opcode to externalize these structs and create an externref, and that externref can be passed to the JS host environment. To JS, that object is a sort of special opaque object with an immutable prototype, and that can be passed back to wasm and internalized again if needed.

In dart2js, I think this actually creates some sort of secondary wrapper around the dart object, but that also might be changing soon to do nothing and just pass the dart object (which is just a JS object) around, and only really changes the static type of the object from the dart language perspective.

In any case, it doesn't matter much in this scenario, just that it creates a non-null, non-falsey JS value.

srujzs: Just +1'ing here. The desire was to avoid users attempting to interface with Dart objects on the JS side or accidentally passing it to different runtimes. For the JS backends, this does mean that we need an actual wrapper, which in practice is just an object literal with a runtime-specific symbol to store the object in. This may indeed change based on some internal feedback, but shouldn't affect how this code is being used.

PR #45673, opened 2023-09-11
[Android] Add support for setting thread affinity based on core speed.

jonahwilliams:
https://github.com/flutter/flutter/issues/134452

This patch parses the speed of all CPU data out of /proc and constructs a table that allows us to request high level CPU affinities: performance, efficiency, and not performance. These affinties are applied where appropriate during Android thread construction.


fml/cpu_affinity.cc
```diff
@@ -0,0 +1,119 @@
...
+      return efficiency_;
+    case CpuAffinity::kNotPerformance:
+      return not_performance_;
+  }
+}
+
+// Get the size of the cpuinfo file by reading it until the end. This is
+// required because files under /proc do not always return a valid size
+// when using fseek(0, SEEK_END) + ftell(). Nor can they be mmap()-ed.
+std::optional<int32_t> ReadIntFromFile(const std::string& path) {
```
jonahwilliams, line 63: This is scary looking but we're running lsan and asan over the tests. Unfortunately cant use the fml file APIs due to the note above.

jonahwilliams: (also lsan and asan helped me fix several bugs here 😆 )

jonahwilliams:
TODO: verify on some non-pixel phones, verify on an emulator

fml/cpu_affinity.cc
zanderso: Can `std::ifstream` read from `/proc` and `/sys`? If it can, I suspect that would simplify this code.

If it can't, I'll do a close review of this code =)

zanderso: Yeah, `std::` was/is banned from the Dart VM, but for new code in the Engine, I believe it is not banned =) cc @chinmaygarde in case he has different advice.

jonahwilliams: ifstream appears to work!

fml/cpu_affinity.cc
```diff
@@ -0,0 +1,78 @@
...
+  // size_t data_length = 0u;
+  std::ifstream file;
+  file.open(path.c_str());
+
+  // Dont use stoi because if this data isnt a parseable number then it
+  // will abort, as we compile with exceptions disabled.
+  int speed = 0;
+  file >> speed;
+  if (speed > 0) {
```
zanderso, line 72: If I'm reading the docs right, then `file.fail()` will be `true` if an `int` couldn't be read.

https://cplusplus.com/reference/istream/istream/operator%3E%3E/

fml/cpu_affinity.cc
```diff
@@ -0,0 +1,78 @@
...
+  }
+}
+
+// Get the size of the cpuinfo file by reading it until the end. This is
+// required because files under /proc do not always return a valid size
+// when using fseek(0, SEEK_END) + ftell(). Nor can they be mmap()-ed.
+std::optional<int32_t> ReadIntFromFile(const std::string& path) {
+  // size_t data_length = 0u;
+  std::ifstream file;
+  file.open(path.c_str());
```
zanderso, line 66: `file.fail()` will be true if the open fails.

https://cplusplus.com/reference/fstream/ifstream/open/

fml/cpu_affinity.cc
jonahwilliams: hmm, the tests I added, which include tests for missing files and non-numbers still pass without checking - I assume because in these cases we don't read anything out of the stream?

zanderso: Ah, yeah, if the open or read fails, then `speed` will just stay `0` and this check will fail.

fml/cpu_affinity.h
```diff
@@ -0,0 +1,65 @@
...
+  /// @brief Request CPU affinity for the efficiency cores.
+  kEfficiency,
+
+  /// @brief Request affinity for all non-performance cores.
+  kNotPerformance,
+};
+
+struct CpuIndexAndSpeed {
+  size_t index;
+  int32_t speed;
```
zanderso, line 32: How about 64 bits here?

fml/cpu_affinity.cc
```diff
@@ -0,0 +1,78 @@
...
+
+  // Dont use stoi because if this data isnt a parseable number then it
+  // will abort, as we compile with exceptions disabled.
+  int speed = 0;
```
zanderso, line 70: Should this be int64_t?

zanderso: Or maybe `long long`.

zanderso (commented):
The value of `cpuinfo_max_freq` is spec'd to be in kHz, but I'd switch it to an `int64_t` anyway.

Might also add a link somewhere to https://www.kernel.org/doc/Documentation/cpu-freq/user-guide.txt

zanderso (approved):
lgtm w/ small fixes.

fml/cpu_affinity.cc
jonahwilliams: long long lints that we should use int64_t

fml/cpu_affinity.h
jonahwilliams: done

jonahwilliams:
I updated to use ifstream. fml:: UniqueFD doesn't seem useful as the only way to read data out of them is FileMapping, which is mmap only and non-functional on these files.

fml/platform/android/cpu_affinity.h
```diff
@@ -0,0 +1,17 @@
...
+#pragma once
+
+#include "flutter/fml/cpu_affinity.h"
+
+namespace fml {
+
+/// @brief Request the given affinity for the current thread.
+///
+///        Returns true if successfull, or if it was a no-op. This function is
+///        only supported on Android devices.
```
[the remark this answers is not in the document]
jonahwilliams: Technically the underlying APIs, but I only added support to Android platform code as testing the linux embedding as well as arbitrary 3rd party linux embedders seems like a bit of a nightmare.

fml/cpu_affinity.h
```diff
@@ -0,0 +1,65 @@
...
+/// @brief A class that computes the correct CPU indices for a requested CPU
+/// affinity.
+///
+///        Note: this is visible for testing.
+class CPUSpeedTracker {
+ public:
+  explicit CPUSpeedTracker(std::vector<CpuIndexAndSpeed> data);
+
+  /// @brief The class is valid if it has more than one CPU index and a distinct
+  /// set of
```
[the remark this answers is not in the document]
jonahwilliams: Done

fml/platform/android/cpu_affinity.cc
```diff
@@ -0,0 +1,59 @@
...
+#include <mutex>
+#include <optional>
+#include <thread>
+
+namespace fml {
+
+/// The CPUSpeedTracker is initialized once the first time a thread affinity is
+/// requested.
+std::once_flag cpu_tracker_flag_;
+static CPUSpeedTracker* tracker_;
```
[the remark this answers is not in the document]
jonahwilliams: Done

fml/cpu_affinity.h
```diff
@@ -0,0 +1,65 @@
...
+
+struct CpuIndexAndSpeed {
+  size_t index;
+  int32_t speed;
+};
+
+/// @brief A class that computes the correct CPU indices for a requested CPU
+/// affinity.
+///
+///        Note: this is visible for testing.
```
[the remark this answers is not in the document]
jonahwilliams: Done

fml/cpu_affinity.cc
```diff
@@ -0,0 +1,78 @@
...
+      return efficiency_;
+    case CpuAffinity::kNotPerformance:
+      return not_performance_;
+  }
+}
+
+// Get the size of the cpuinfo file by reading it until the end. This is
+// required because files under /proc do not always return a valid size
+// when using fseek(0, SEEK_END) + ftell(). Nor can they be mmap()-ed.
+std::optional<int32_t> ReadIntFromFile(const std::string& path) {
```
zanderso, line 63:
```suggestion
std::optional<int64_t> ReadIntFromFile(const std::string& path) {
```

fml/cpu_affinity.h
```diff
@@ -0,0 +1,68 @@
...
+ private:
+  bool valid_ = false;
+  std::vector<CpuIndexAndSpeed> cpu_speeds_;
+  std::vector<size_t> efficiency_;
+  std::vector<size_t> performance_;
+  std::vector<size_t> not_performance_;
+};
+
+/// @note Visible for testing.
+std::optional<int32_t> ReadIntFromFile(const std::string& path);
```
zanderso, line 66:
```suggestion
std::optional<int64_t> ReadIntFromFile(const std::string& path);
```

fml/cpu_affinity.cc
jonahwilliams: oops

fml/cpu_affinity.cc
```diff
@@ -0,0 +1,78 @@
...
+    case CpuAffinity::kNotPerformance:
+      return not_performance_;
+  }
+}
+
+// Get the size of the cpuinfo file by reading it until the end. This is
+// required because files under /proc do not always return a valid size
+// when using fseek(0, SEEK_END) + ftell(). Nor can they be mmap()-ed.
+std::optional<int64_t> ReadIntFromFile(const std::string& path) {
+  // size_t data_length = 0u;
```
dnfield, line 64: Commented out code

PR #45675, opened 2023-09-11
Roll libwebp to 1.3.1

jason-simmons:


zanderso (approved):

zanderso:
Link errors look like there are some source files possibly missing from the list in the buildroot patch.

PR #45682, opened 2023-09-11
[Impeller] moved validation layers on by default logic to gni scripts

gaaclarke:
fixes https://github.com/flutter/flutter/issues/134460

## Pre-launch Checklist

- [ ] I signed the [CLA].


gaaclarke:
According to this build step the validation layers are now building, matching the behavior of the impeller vulkan backend: https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8770207159619809201/+/u/build_android_debug_arm64_flutter_flutter_shell_platform_android:abi_jars/stdout

jonahwilliams (approved):
LGTM

So the CI bots aren't provided enable_impeller_vulkan while building for Android? I guess this is working though 🤔

tools/gn
```diff
@@ -646,10 +646,9 @@ def to_gn_args(args):
   if args.unoptimized and args.target_os == 'android' and args.android_cpu == 'arm64':
     enable_vulkan_validation = True
 
-  if enable_vulkan_validation or (args.enable_impeller_vulkan and
-                                  runtime_mode == 'debug' and
-                                  gn_args['target_cpu'] == 'arm64'):
+  if enable_vulkan_validation:
     gn_args['enable_vulkan_validation_layers'] = True
+    gn_args['impeller_enable_vulkan_validation_layers'] = True
```
zanderso, line 651: Why do we have both `enable_vulkan_validation_layers` and `impeller_enable_vulkan_validation_layers`?

Or rather, what's the difference between `impeller_enable_vulkan_validation_layers` and `(enable_impeller_vulkan && enable_vulkan_validation_layers)`

(Not sure that `enable_impeller_vulkan` is the right spelling, but there's probably something we already have that could go there?)

gaaclarke:
> LGTM
> 
> So the CI bots aren't passing enable_impeller_vulkan while building for Android? I guess this is working though 🤔

Yep, that appears to be the case.  I looked at the post build bots and didn't see one that turned on enable_impeller_vulkan, the bot is just relying on the default value from the gni file.  I'm not 100% certain these are the source of the artifacts that are used by flutter/flutter, but it is likely that fixing this fixes it for whoever is making the artifacts if it isn't those bots.

tools/gn
gaaclarke: `enable_vulkan_validation_layers` is for the fuchsia code, `impeller_enable_vulkan_validation_layers` is for impeller.  There is also a `angle_enable_vulkan_validation_layers`.  `enable_vulkan_validation_layers` predates impeller.  I tried to share the variables.  That's how it was previously, but we need a different default value.  That's why we need a different variable here.

PR #45686, opened 2023-09-12
[Impeller] Fix thread leak in ResourceManagerVK.

matanlurey:
Fixes https://github.com/flutter/flutter/issues/134482.

I wish I knew how to create the thread in `::Create()`, but given the constructor is private I feel less bad about reverting part of https://github.com/flutter/engine/pull/45474. Added a test that consistently fails before this PR and passes after.

(Unblocks https://github.com/flutter/flutter/issues/133198)


jonahwilliams (approved):
LGTM!

PR #45687, opened 2023-09-12
[Impeller] Make `CreateMockVulkanContext()` thread-safe

matanlurey:
I started using `CreateMockVulkanContext()` in https://github.com/flutter/engine/pull/45654, and discovered that it was not thread-safe. This PR does a little bit of refactoring/TLC to ensure that functions and buffers can be accessed across threads (like the _real_ `ContextVK`).

The attached test (_test for a test fixture, hoorah!_) fails consistently before the changes, and passes after. Note that if you try testing this with `--gtest_repeat=10000` (at least that's the threshold for me), it'll fail - that's actually due to https://github.com/flutter/flutter/issues/134482 which in turn is being fixed in https://github.com/flutter/engine/pull/45686.

(Unblocks https://github.com/flutter/flutter/issues/133198)


impeller/renderer/backend/vulkan/test/mock_vulkan.cc
```diff
@@ -16,16 +18,34 @@ struct MockCommandBuffer {
...
+    Lock lock(command_buffers_mutex_);
     command_buffers_.emplace_back(std::move(buffer));
     return result;
   }
+
+  std::shared_ptr<std::vector<std::string>>& GetCalledFunctions() {
+    return called_functions_;
+  }
+
+  void was_called(const std::string& function) {
```
gaaclarke, line 37:
```suggestion
  void AddCalledFunction(const std::string& function) {
```

The method should be pascal case.  Nit: `was*` sounds like a method that should return a bool, "Add" matches better what it is doing.

impeller/renderer/backend/vulkan/test/mock_vulkan.cc
```diff
@@ -16,16 +18,34 @@ struct MockCommandBuffer {
...
-    std::unique_ptr<MockCommandBuffer> buffer =
-        std::make_unique<MockCommandBuffer>(called_functions_);
+    auto buffer = std::make_unique<MockCommandBuffer>(called_functions_);
     MockCommandBuffer* result = buffer.get();
+    Lock lock(command_buffers_mutex_);
     command_buffers_.emplace_back(std::move(buffer));
     return result;
   }
+
+  std::shared_ptr<std::vector<std::string>>& GetCalledFunctions() {
```
gaaclarke, line 33:
```suggestion
  std::shared_ptr<const std::vector<std::string>>& GetCalledFunctions() {
```

Now that someone has to use `was_called` it's worth making this return a const pointer so someone doesn't accidentally do the wrong thing.

impeller/renderer/backend/vulkan/test/mock_vulkan.cc
```diff
@@ -16,16 +18,34 @@ struct MockCommandBuffer {
...
+
+  void was_called(const std::string& function) {
+    Lock lock(called_functions_mutex_);
+    called_functions_->push_back(function);
+  }
+
+ private:
+  FML_DISALLOW_COPY_AND_ASSIGN(MockDevice);
+
+  Mutex called_functions_mutex_;
```
gaaclarke, line 45: Can we use `IPLR_GUARDED_BY` with these instance variables?  I'm not sure if it will work with that Mutex.

impeller/renderer/backend/vulkan/test/mock_vulkan_unittests.cc
```diff
@@ -0,0 +1,37 @@
...
+
+#include "flutter/testing/testing.h"  // IWYU pragma: keep
+#include "gtest/gtest.h"
+#include "impeller/renderer/backend/vulkan/command_pool_vk.h"
+#include "impeller/renderer/backend/vulkan/test/mock_vulkan.h"
+
+namespace impeller {
+namespace testing {
+
+TEST(MockVulkanContextTest, IsThreadSafe) {
```
gaaclarke, line 13: Is the idea that randomly this would crash if it wasn't threadsafe?  By it's nature MockVulkanContext only lives to be run on tests so there is no need to add tests to it.  If it was helpful to developing the change I'm fine with keeping the test if you want.

gaaclarke (approved):
Thanks, looks good to me.  I just have a few minor comments.

impeller/renderer/backend/vulkan/test/mock_vulkan.cc
matanlurey: Yup. Done.

impeller/renderer/backend/vulkan/test/mock_vulkan_unittests.cc
matanlurey: Yeah it was entirely to isolate the problems and probably useful to keep around.

PR #45731, opened 2023-09-12
Add warning about Vulkan queue usage in embedder.h

bdero:
Add warning that suggests the workaround inline and points to https://github.com/flutter/flutter/issues/134573.

Aside from this, we could also update our GLFW example embedder to employ this workaround (providing an explicit code sample that users may reference/copy).

### DIFF
--- a/shell/platform/embedder/embedder.h
+++ b/shell/platform/embedder/embedder.h
@@ -757,6 +757,19 @@ typedef struct {
   /// The queue family index of the VkQueue supplied in the next field.
   uint32_t queue_family_index;
   /// VkQueue handle.
+  ///
+  /// @warning  There is a synchronization bug where Flutter will submit
+  ///           to this queue on both the raster and IO threads. This is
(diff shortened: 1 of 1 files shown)


gaaclarke (changes requested):
I'm already working on a PR that includes the unit test that does the swapping out so the docstring can reference it.

bdero:
Does the docstring look good? This just adds a docstring and doesn't need a unittest. I'm OK to hold off though, no strong opinions here.

gaaclarke:
> Does the docstring look good? This just adds a docstring and doesn't need a unittest. I'm OK to hold off though, no strong opinions here.

I'm minutes from posting the PR as we discussed in our meeting.  I have something different written up that documents queue and the proc callback, plus it has the test that asserts it works.

gaaclarke:
Here is the PR with the test (https://github.com/flutter/engine/pull/45732).

bdero:
Closing in favor of https://github.com/flutter/engine/pull/45732.

PR #45734, opened 2023-09-12
Revert "Lazily allocate RasterCacheItems only when caching is enabled"

zanderso:
Reverts flutter/engine#45211

b/298583505

### DIFF
--- a/flow/layers/cacheable_layer.cc
+++ b/flow/layers/cacheable_layer.cc
@@ -6,39 +6,31 @@
 
 namespace flutter {
 
-AutoCache::AutoCache(CacheableLayer& cacheable_layer,
+AutoCache::AutoCache(RasterCacheItem* raster_cache_item,
                      PrerollContext* context,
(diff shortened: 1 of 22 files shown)


zanderso:
Prospective revert. Not sure this is right. Need input from @jason-simmons and/or @flar

jason-simmons:
Confirmed that the b/298583505 issue does not occur if https://github.com/flutter/engine/pull/45211 is reverted in a local engine build.

jason-simmons (approved):

flar:
I confirmed that it does reproduce in a local engine build on the commit *JUST PRIOR* to that PR commit...?

flar (changes requested):
My attempts to reproduce the problem indicated that this change was not suspect.

flar (approved):
It looks like my reproducing attempts weren't correct so I'm going to defer to Jason on this one.

PR #45748, opened 2023-09-13
Do not run real processes in `clang_tidy_test.dart`

matanlurey:
Partial work towards https://github.com/flutter/flutter/issues/133190.

Some context, from @zanderso on chat:

@matanlurey:

> For [`clang_tidy_test.dart`](https://github.com/flutter/engine/blob/c020936303cb0646cd10593e6e427d8c5aa6ce52/tools/clang_tidy/test/clang_tidy_test.dart), what is your vision there? I remember we ran into problems with it actually executing `clang.run()` in too many of the test cases, but I don't remember (or seem to have written down) what we decided to do 🙂

@zanderso:

> The existing tests are calling `computeFilesOfInterest` and `getLintCommandsForFiles` and inspecting the results for positive tests instead of hitting a real clang-tidy invocation. Tests of command line flags that call `clangTidy.run()` are also testing cases that short-circuit or fail before making a real invocation. Making it harder to do a real invocation from a unit test probably requires plumbing a fake-able ProcessRunner or ProcessManager object through to the ClangTidy constructor.

---

All this PR does is allow providing a `ProcessManager` (defaults to `LocalProcessManager`, i.e. `dart:io`) throughout the tool. Then, for tests, I've implemented a _very_ minimal fake of both `ProcessManager` and `Process` (it would be nice to share code w/ say, `flutter_tool`, but lots of work/overhead for very little surface area).

After this CL, no tests in `clang_tidy_test.dart` actually run a process. This should unblock adding new features (i.e. original intent of https://github.com/flutter/flutter/issues/133190) without causing headaches for others/CI?

### DIFF
--- a/tools/clang_tidy/test/process_fakes.dart
+++ b/tools/clang_tidy/test/process_fakes.dart
@@ -1,0 +1,79 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'dart:convert';
+import 'dart:io' as io;
+
+import 'package:process/process.dart';
+
+/// A fake implementation of [ProcessManager] that allows control for testing.
+final class FakeProcessManager implements ProcessManager {
+  FakeProcessManager({
+    io.ProcessResult Function(List<String> command) onRun = unhandledRun,
+    io.Process Function(List<String> command) onStart = unhandledStart,
+  }) : _onRun = onRun, _onStart = onStart;
+
+  static io.ProcessResult unhandledRun(List<String> command) {
+    throw UnsupportedError('Unhandled run: ${command.join(' ')}');
+  }
+
+  static io.Process unhandledStart(List<String> command) {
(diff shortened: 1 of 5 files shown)


zanderso (approved):
Thanks!

tools/clang_tidy/test/process_fakes.dart
```diff
@@ -0,0 +1,117 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
```
zanderso, line 1: I don't see an implementation of something like this anywhere else in the engine repo. At some point, I'd guess that we'll want to pull this out to its own package in the engine repo to share among other packages, or upstream something like https://github.com/flutter/flutter/blob/master/packages/flutter_tools/test/src/fake_process_manager.dart to https://github.com/google/process.dart

matanlurey: Yeah I agree with that! I'd probably say:

1. Pull this into `tools/pkg/fake_process_manager` as soon as we need it somewhere else.
2. Chat with Chris about moving `flutter_tools`' implementation

PR #45749, opened 2023-09-13
[web] support manual loading of manifest fonts

yjbanov:
Add support for manual loading of manifest fonts (i.e. those specified in `pubspec.yaml`):

- Add a new option `loadManifestFontsBeforeAppMain` in the configuration object (defaults to `true`). If not specified or set to true, font loading works like it does today: manifest fonts are loaded prior to calling the `main()` function. If set to `false`, the engine does not load manifest fonts automatically, but does allow rendering UI. Attempting to render text without fonts will result in blank space in place of the text.
- Add a new function `loadManifestFonts`, which can be called explicitly to load manifest fonts. After this function completes successfully, text can be rendered using manifest fonts.

Fixes https://github.com/flutter/flutter/issues/134594

### DIFF
--- a/lib/web_ui/lib/src/engine/configuration.dart
+++ b/lib/web_ui/lib/src/engine/configuration.dart
@@ -312,6 +312,19 @@ class FlutterConfiguration {
   /// The font used to render color emojis is large (~24MB). This configuration
   /// gives developers the ability to decide for their app.
   bool get useColorEmoji => _configuration?.useColorEmoji ?? false;
+
+  /// Whether to load fonts from the font manifest prior to calling the app's
+  /// `main` function.
+  ///
+  /// If set to true, manifest fonts will load as part of engine initialization,
+  /// delaying the invocation of `main`, if necessary.
+  ///
+  /// If set to false, engine initialization does not include manifest fonts.
+  /// Instead, the developer can load manifest fonts using the
+  /// `loadManifestFonts` function from `dart:ui_web`.
+  ///
+  /// Defaults to true.
+  bool get loadManifestFontsBeforeAppMain => _configuration?.loadManifestFontsBeforeAppMain ?? true;
 }
 
 @JS('window.flutterConfiguration')
@@ -320,7 +333,22 @@ external JsFlutterConfiguration? get _jsConfiguration;
 /// The JS bindings for the object that's set as `window.flutterConfiguration`.
 @JS()
 @staticInterop
-class JsFlutterConfiguration {}
+@anonymous
+class JsFlutterConfiguration {
+  external factory JsFlutterConfiguration({
+    JSString? assetBase,
+    JSString? canvasKitBaseUrl,
+    JSString? canvasKitVariant,
+    JSBoolean? canvasKitForceCpuOnly,
+    JSNumber? canvasKitMaximumSurfaces,
+    JSBoolean? debugShowSemanticsNodes,
+    DomElement? hostElement,
+    JSString? nonce,
+    JSString? renderer,
+    JSBoolean? useColorEmoji,
+    JSBoolean? loadManifestFontsBeforeAppMain,
+  });
+}
 
 extension JsFlutterConfigurationExtension on JsFlutterConfiguration {
   @JS('assetBase')
@@ -360,6 +388,10 @@ extension JsFlutterConfigurationExtension on JsFlutterConfiguration {
   @JS('useColorEmoji')
   external JSBoolean? get _useColorEmoji;
   bool? get useColorEmoji => _useColorEmoji?.toDart;
+
+  @JS('loadManifestFontsBeforeAppMain')
+  external JSBoolean? get _loadManifestFontsBeforeAppMain;
+  bool? get loadManifestFontsBeforeAppMain => _loadManifestFontsBeforeAppMain?.toDart;
 }
 
 /// A JavaScript entrypoint that allows developer to set rendering backend
--- a/lib/web_ui/lib/ui_web/src/ui_web/initialization.dart
+++ b/lib/web_ui/lib/ui_web/src/ui_web/initialization.dart
@@ -55,3 +55,22 @@ Future<void> bootstrapEngine({
     loader.didCreateEngineInitializer(bootstrap.prepareEngineInitializer());
   }
 }
+
+/// Loads manifest fonts.
+///
+/// By default manifest fonts are loaded automatically as part of engine
+/// initialization. However, that may delay the invocation of the app's `main`
+/// function. If the app needs to start doing useful work before the fonts are
+/// fully loaded, such as render UI without text (e.g. a loading indicator), or
+/// perform non-UI work (e.g. make early HTTP calls), then the configuration
+/// option `loadManifestFontsBeforeAppMain` can be set to `false`. This will
+/// make the engine skip font loading. Instead, the app can call this function
+/// directly when appropriate.
+///
+/// Beware that attempting to render text before fonts are loaded may lead to
+/// improper text rendering. It is expected that the developer has enough
+/// control over the app's code to ensure the correct loading sequence.
+Future<void> loadManifestFonts() async {
+  await downloadAssetFonts();
+  sendFontChangeMessage();
+}
--- a/lib/web_ui/test/canvaskit/initialization/load_manifest_fonts_manually_test.dart
+++ b/lib/web_ui/test/canvaskit/initialization/load_manifest_fonts_manually_test.dart
@@ -1,0 +1,94 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import 'dart:async';
+import 'dart:js_interop';
+import 'dart:typed_data';
+
+import 'package:test/bootstrap/browser.dart';
+import 'package:test/test.dart';
+import 'package:ui/src/engine.dart';
+import 'package:ui/ui.dart' as ui;
+import 'package:ui/ui_web/src/ui_web.dart' as ui_web;
+import 'package:web_engine_tester/golden_tester.dart';
+
+@JS('window.flutterConfiguration')
+external set _jsConfiguration(JsFlutterConfiguration? configuration);
+
+void main() {
+  internalBootstrapBrowserTest(() => testMain);
+}
+
+void testMain() {
+  test('fonts can be loaded manually', () async {
+    _jsConfiguration = null;
+
+    // Fully initialize engine services and UI without manifest fonts. Test that
+    // the engine can still render everything, albeit without text.
+    await initializeEngineServices(jsConfiguration: JsFlutterConfiguration(
+      loadManifestFontsBeforeAppMain: false.toJS,
+      canvasKitBaseUrl: '/canvaskit/'.toJS,
+    ));
+    await initializeEngineUi();
+
+    String? messageChannel;
+    Object? messageData;
+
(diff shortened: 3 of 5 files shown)


lib/web_ui/lib/src/engine/configuration.dart
```diff
@@ -312,6 +312,19 @@ class FlutterConfiguration {
...
+  ///
+  /// If set to true, manifest fonts will load as part of engine initialization,
+  /// delaying the invocation of `main`, if necessary.
+  ///
+  /// If set to false, engine initialization does not include manifest fonts.
+  /// Instead, the developer can load manifest fonts using the
+  /// `loadManifestFonts` function from `dart:ui_web`.
+  ///
+  /// Defaults to true.
+  bool get loadManifestFontsBeforeAppMain => _configuration?.loadManifestFontsBeforeAppMain ?? true;
```
jacobsimionato, line 327: Is the concept of the "manifest" exposed to app developers? Looking at the external documentation e.g. https://docs.flutter.dev/cookbook/design/fonts, it seems like developers don't usually touch it, and 1P developers ideally shouldn't either.

How about just calling this 'loadFontsBeforeAppMain' and the other function 'loadFonts' ? Or is that too generic?

lib/web_ui/lib/ui_web/src/ui_web/initialization.dart
```diff
@@ -55,3 +55,22 @@ Future<void> bootstrapEngine({
...
+/// fully loaded, such as render UI without text (e.g. a loading indicator), or
+/// perform non-UI work (e.g. make early HTTP calls), then the configuration
+/// option `loadManifestFontsBeforeAppMain` can be set to `false`. This will
+/// make the engine skip font loading. Instead, the app can call this function
+/// directly when appropriate.
+///
+/// Beware that attempting to render text before fonts are loaded may lead to
+/// improper text rendering. It is expected that the developer has enough
+/// control over the app's code to ensure the correct loading sequence.
+Future<void> loadManifestFonts() async {
```
jacobsimionato, line 73: Can we just call this method before engine initialization and then plumb the future through to runApp and await it there? To me, this seems like it'd be optimal for the vast majority of apps and avoid them adding more complexity in main.dart? Could we even change the default to do the above? I assume there would be minor consequences e.g. if people do some kind of custom prerendering things before runApp, they could break.

lib/web_ui/lib/src/engine/configuration.dart
yjbanov: I wanted to qualify this with _something_ because it does not affect all fonts. For example, fallback fonts will still attempt to load if you try rendering a piece of text. You can also load a font using the `FontLoader`. This change only affects the fonts declared in [pubspec](https://docs.flutter.dev/cookbook/design/fonts#2-declare-the-font-in-the-pubspec). There's no special term used in pubspec for those fonts. I've heard them called "asset fonts" and "custom fonts", none of which imply "fonts that are loaded prior to the app". As of today calling them "pubspec fonts" may actually be accurate, but I'm not sure if it would be future-proof. We could add fonts to the manifest from other sources in the future. For example, we discussed generating a minimal fallback font bundle based on predeclared strings that appear in the app.

lib/web_ui/lib/ui_web/src/ui_web/initialization.dart
yjbanov: This method includes fetching the fonts _and_ loading them into CanvasKit (or any renderer in general). So we need at least `initializeEngineServices` to complete, and that's a big chunk of engine initialization. I imagine, since this API is already for advanced users, those users will appreciate having more control despite a little more complexity. In particular, some users may want to skip manifest fonts entirely and rely purely on fonts loaded via `FontLoader`.

lib/web_ui/lib/src/engine/configuration.dart
```diff
@@ -320,7 +333,22 @@ external JsFlutterConfiguration? get _jsConfiguration;
...
+    JSString? canvasKitVariant,
+    JSBoolean? canvasKitForceCpuOnly,
+    JSNumber? canvasKitMaximumSurfaces,
+    JSBoolean? debugShowSemanticsNodes,
+    DomElement? hostElement,
+    JSString? nonce,
+    JSString? renderer,
+    JSBoolean? useColorEmoji,
+    JSBoolean? loadManifestFontsBeforeAppMain,
+  });
```
ditman, line 350: Please, do not expose a Dart constructor for this object. We already have a way to override JS Configuration values from Dart (without messing with the config set by the test runner): `debugOverrideJsConfiguration`.

lib/web_ui/test/canvaskit/initialization/load_manifest_fonts_manually_test.dart
```diff
@@ -0,0 +1,94 @@
...
+@JS('window.flutterConfiguration')
+external set _jsConfiguration(JsFlutterConfiguration? configuration);
+
+void main() {
+  internalBootstrapBrowserTest(() => testMain);
+}
+
+void testMain() {
+  test('fonts can be loaded manually', () async {
+    _jsConfiguration = null;
```
ditman, line 25: This is removing any configuration set by the test runner. Please use this instead:

```dart
debugOverrideJsConfiguration(
  <String, Object?>{
    'loadManifestFontsBeforeAppMain': false,
  }.jsify() as JsFlutterConfiguration?
);
```

Then `initializeEngineServices` can be called without any params, IIRC.

lib/web_ui/test/canvaskit/initialization/load_manifest_fonts_manually_test.dart
```diff
@@ -0,0 +1,94 @@
...
+
+    // The notification message is sent in the next frame so there's only one
+    // notification for a batch of fonts, so wait for 2 frames before checking
+    // for the message.
+    final Completer<void> waitForFontMessage = Completer<void>();
+    domWindow.requestAnimationFrame((_) {
+      domWindow.requestAnimationFrame((_) {
+        waitForFontMessage.complete();
+      });
+    });
```
ditman, line 58: Why not make the `onPlatformMessage` a function that funnels the data to a Stream, and here you just `await` for that?

Relying on X numbers of frames seems flaky. For example: Why 2 if we've already awaited for the fonts to be loaded above?

lib/web_ui/test/canvaskit/initialization/load_manifest_fonts_manually_test.dart
```diff
@@ -0,0 +1,94 @@
...
+    paragraph.layout(const ui.ParagraphConstraints(width: 1000));
+    canvas.drawRect(
+      ui.Rect.fromLTWH(0, 0, paragraph.width, paragraph.height).inflate(10),
+      CkPaint()..color = const ui.Color(0xFF00FF00)
+    );
+    canvas.drawParagraph(paragraph, ui.Offset.zero);
+    picture = recorder.endRecording();
+  }
+
+  {
```
ditman, line 84: Do these brackets add any semantics to the code, or is it visual separation of the different parts of this method?

ditman (commented):
Small comments on the way we override JSConfig on tests.

lib/web_ui/lib/ui_web/src/ui_web/initialization.dart
yjbanov: IIUC `customer: quake` does want to render UI before fonts are downloaded (I'll double-check with them), so we may already have a use-case for not blocking `runApp`.

More generally, I feel like for the use-case of fine-tuning your app start-up sequence this kind of logic should not be a hidden implementation detail, but be part of an explicit API. I think there's more complexity in hiding it than exposing it, because instead of seeing a function invocation in your code and deducing its effect on that line (and also deciding for yourself where to call this function), you have to reason about how the font loading process changes behind-the-scenes inside the engine and in parallel with your logic.

lib/web_ui/lib/src/engine/configuration.dart
jacobsimionato: I see this isn't as simple as I thought! I like Mouad's "loadManifestFontsManually" suggestion.

yjbanov: I like `loadManifestFontsManually`! I'm willing to remove the reference to the "manifest" if that feels off, but I still like that the word "manifest" (aside from its specific usages in Flutter today) implies that you have declared a list of fonts somewhere and it's a list that's known to the Flutter runtime and has special meaning that's different from other fonts that exist in your app. LMK.

lib/web_ui/lib/ui_web/src/ui_web/initialization.dart
jacobsimionato: Re advanced users: Yeah, this makes sense... I take the point that there are advanced use cases that require full manual control, and so just disabling automatic font loading completely is best for them. So this CL seems ideal to unblock that case.

I guess what I'm wondering is whether there's an opportunity to boost the performance of every app that needs to do some asynchronous work in main() very cheaply by pushing people towards a default of fonts not blocking main() execution as Mouad was saying. Remember, John's font loading optimization approach (CL/560854521) improved overall loading time by ~9% overall for Earth for slower connections. What if you could give all new apps an improvement of this scale with a relatively small change to the contract of main() vs runApp() that would probably not affect most apps at all? Surely this is a cheaper optimization than trying to shave another 10% off the binary size of canvaskit etc?

I actually didn't even realise there was a distinction between loading fonts at the start or end of engine initialization, though if it's possible to kick off the RPC as early as possible and then load the fonts into the engine when the engine is ready, that would be great.

jacobsimionato: Sorry I wrote my comment before I'd read yours, Yegor - that answers most of my concerns. Fair point that often having access to the actual futures is the easiest to understand. My resistance to embracing the explicit API here is partially due to the fact that we see so many apps await futures sequentially in main() when they can be parallelized. But this can probably be solved with documentation etc rather than withdrawing control.

That's interesting re `customer: quake`! I'm not aware of any google3 customers who do any kind of rendering before runApp. My perspective is skewed towards the handful of apps that I'm familiar with who basically load some data and initialize some services in main() and then render a conventional UI in runApp().

lib/web_ui/lib/src/engine/configuration.dart
jacobsimionato: I'm fine with 'manifest' now I understand more context.

jacobsimionato (approved):

lib/web_ui/lib/ui_web/src/ui_web/initialization.dart
yjbanov: I confirmed with `customer: quake` that they do indeed want to render non-text content prior to loading the fonts. How does that change your view?

goderbauer:
@yjbanov said this is still on his radar.

yjbanov:
@jacobsimionato Would this still be valuable? Or have users found other ways to achieve what this aims to fix?

jacobsimionato:
I think this would still be useful, though I'd like to revisit it with an additional data point - another internal Google app that I'm going to optimize over the next couple of weeks where font loading is currently an issue. Maybe I can get back to you in a few days once I've taken a look at this second app?

Though it does sound like there is demand for this outside Google also!

yjbanov:
Closing for now. The demand seems to have waned for this feature. We might come back to it at some point.

josh-burton:
> Closing for now. The demand seems to have waned for this feature. We might come back to it at some point.

I'm still really keen for this. I've had to implement a custom font loader for use on the web to improve our app startup time.

Would love to see this continued

yjbanov:
If someone would like to take this over, go ahead. I'm unlikely to get to it any time soon.

PR #45777, opened 2023-09-13
Revert "[Impeller] Patch the compiler to account for subpass inputs and PSO metadata."

gaaclarke:
Reverts flutter/engine#45739
fixes https://github.com/flutter/flutter/issues/134666


chinmaygarde (approved):

gaaclarke:
I ran this revert locally to verify that it fixes the issue even though it isn't apparent how it causes the failure.

PR #45797, opened 2023-09-13
Compile a platform dill for dart2wasm

eyebrowsoffire:
This continues work on https://github.com/flutter/flutter/issues/133467

We still need a change on the tool side to consume this platform dill file after this lands.

### DIFF
--- a/web_sdk/BUILD.gn
+++ b/web_sdk/BUILD.gn
@@ -229,6 +229,11 @@ template("_compile_platform") {
     if (defined(invoker.null_environment) && invoker.null_environment) {
       args += [ "--null-environment" ]
     }
+    skwasm_library = "dart:_skwasm_stub"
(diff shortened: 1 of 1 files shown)


web_sdk/BUILD.gn
```diff
@@ -306,6 +311,14 @@ _compile_platform("flutter_dart2js_kernel_sdk_full_sound") {
   null_environment = true
 }
 
+_compile_platform("flutter_dart2wasm_kernel_sdk_full_sound") {
+  sound_null_safety = true
```
yjbanov, line 315: Do we still need to spell it out?

yjbanov (approved):

web_sdk/BUILD.gn
eyebrowsoffire: Yes, insofar as I need to pass it as a parameter to `kernel_worker`

PR #45803, opened 2023-09-13
[Impeller] Set impeller_enabled flag consistently in Preroll and Paint.

flar:
Fixes https://github.com/flutter/flutter/issues/134689


jonahwilliams (approved):
LGTM

flar:
This PR will become obsolete (if/)when https://github.com/flutter/engine/pull/43759 is reverted.

jonahwilliams:
We should still land this if we can, as the preroll flag is used elsewhere too.

jonahwilliams:
nvm, you were right. I backed the change out manually and the flag could be removed entirely

PR #45804, opened 2023-09-13
Switch goma to reclient fro standalone targets

keyonghan:
Part of https://github.com/flutter/flutter/issues/132701.

### DIFF
--- a/.ci.yaml
+++ b/.ci.yaml
@@ -86,6 +86,7 @@ targets:
     enabled_branches:
       - main
     recipe: engine_v2/builder
+    presubmit: false
     properties:
       config_name: linux_android_emulator
     timeout: 60
@@ -152,6 +153,7 @@ targets:
     enabled_branches:
       - main
     recipe: engine_v2/builder
+    presubmit: false
     properties:
       config_name: linux_benchmarks
     timeout: 60
--- a/.ci.yaml
+++ b/.ci.yaml
@@ -86,7 +86,6 @@ targets:
     enabled_branches:
       - main
     recipe: engine_v2/builder
-    presubmit: false
     properties:
       config_name: linux_android_emulator
     timeout: 60
@@ -153,7 +152,6 @@ targets:
     enabled_branches:
       - main
     recipe: engine_v2/builder
-    presubmit: false
     properties:
       config_name: linux_benchmarks
     timeout: 60
--- a/ci/builders/standalone/linux_android_emulator.json
+++ b/ci/builders/standalone/linux_android_emulator.json
@@ -6,7 +6,9 @@
     "gn": [
          "--android",
          "--android-cpu=x64",
-         "--no-lto"
+         "--no-lto",
+         "--rbe",
+         "--no-goma"
     ],
     "name": "android_debug_x64",
     "ninja": {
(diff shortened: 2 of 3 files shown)


keyonghan:
First run of benchmark: https://ci.chromium.org/ui/p/flutter/builders/try/Linux%20linux_benchmarks/1/overview, which took 57 mins without hitting any cache;
Second run of benchmark: https://ci.chromium.org/ui/p/flutter/builders/try/Linux%20linux_benchmarks/2/overview, which took 10 mins.

The androit_emulator_tests: https://ci.chromium.org/ui/p/flutter/builders/try/Linux%20linux_android_emulator_tests/2/overview, which took 13 mins with cache hit.

zanderso:
In what situations will the build take 57 minutes?

keyonghan:
> In what situations will the build take 57 minutes?

Where [we do not have a single cache hit](https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket/8770028718994326977/+/u/teardown_remote_execution/read_rbe_metrics.txt/rbe_metrics.txt):
```
num_records: 5510
stats: {
	name: "CompletionStatus"
	counts_by_value: {
		name: "STATUS_LOCAL_FALLBACK"
		count: 43
	}
	counts_by_value: {
		name: "STATUS_REMOTE_EXECUTION"
		count: 5467
	}
}
```

zanderso:
My question is in what situations can we expect to not have a single cache hit. Is this something that will happen on every commit? Once a day? Once a week?

keyonghan:
> My question is in what situations can we expect to not have a single cache hit. Is this something that will happen on every commit? Once a day? Once a week?

I expect this happens only the first time of the build, where the cache is empty ([Run your builds and tests using the remote cache](https://developers.google.com/remote-build-execution/docs/set-up/set-up-caching#run_your_builds_and_tests_using_the_remote_cache)).

zanderso:
> I expect this happens only the first time of the build, where the cache is empty ([Run your builds and tests using the remote cache](https://developers.google.com/remote-build-execution/docs/set-up/set-up-caching#run_your_builds_and_tests_using_the_remote_cache)).

What does "the first time of the build" mean? The first time on any given commit? Or the first time with the repo?

I'm also confused by the documentation at that link. It says, "the cache is empty, so all actions are executed locally". That is very different from goma, which I believe uses the remote workers to run the compile commands even when there isn't a cache hit?

keyonghan:
> > I expect this happens only the first time of the build, where the cache is empty ([Run your builds and tests using the remote cache](https://developers.google.com/remote-build-execution/docs/set-up/set-up-caching#run_your_builds_and_tests_using_the_remote_cache)).
> 
> What does "the first time of the build" mean? The first time on any given commit? Or the first time with the repo?
> 
> I'm also confused by the documentation at that link. It says, "the cache is empty, so all actions are executed locally". That is very different from goma, which I believe uses the remote workers to run the compile commands even when there isn't a cache hit?

The first execution time with the task/builder, which should be the same with the repo.

I should have pointed out that the link focuses on using remote cache for local runs (the topic right after is `Setting up remote execution`) . Since we have set up remote execution, if no cache hit, it will try remote execution first. This is aligned with the stats: https://github.com/flutter/engine/pull/45804#issuecomment-1719898300

zanderso (approved):
Awesome. That all sounds hunky-dory to me, then.

keyonghan:
I will monitor the coming builds to make sure things go well as expected. Landing now.

PR #45832, opened 2023-09-14
[Impeller] Release a texture during Playground teardown

matanlurey:
Closes https://github.com/flutter/flutter/issues/134678.

Partial work towards https://github.com/flutter/flutter/issues/133708, https://github.com/flutter/flutter/issues/133506.

This gets us closer to being able to run `aiks_unittests` with validations enabled, removing a VMA allocation error:

## Before

```txt
Note: Google Test filter = Play/AiksTest.ReleasesTextureOnTeardown/Vulkan
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from Play/AiksTest
[ RUN      ] Play/AiksTest.ReleasesTextureOnTeardown/Vulkan
[INFO:capabilities_vk.cc(50)] Vulkan validations are enabled.
[FATAL:third_party/vulkan_memory_allocator/include/vk_mem_alloc.h(6179)] Check failed: (false && "Unfreed dedicated allocations found!"). Vulkan Memory Allocator Failure!
```

## After

```txt
Note: Google Test filter = Play/AiksTest.ReleasesTextureOnTeardown/Vulkan
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from Play/AiksTest
[ RUN      ] Play/AiksTest.ReleasesTextureOnTeardown/Vulkan
[INFO:capabilities_vk.cc(50)] Vulkan validations are enabled.
[       OK ] Play/AiksTest.ReleasesTextureOnTeardown/Vulkan (1192 ms)
[----------] 1 test from Play/AiksTest (1192 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (1192 ms total)
[  PASSED  ] 1 test.
```

---

Added a test that will catch regressions.


impeller/aiks/aiks_playground_inspector.h
```diff
@@ -21,6 +21,17 @@ class AiksInspector {
       AiksContext& aiks_context,
       const std::function<std::optional<Picture>()>& picture_callback);
 
+  // Resets (releases) the underlying |Picture| object.
+  //
+  // Underlying issue: <https://github.com/flutter/flutter/issues/134678>.
+  //
+  // The tear-down code is not running in the right order; it's torn down after
+  // we shut down the |ContextVK|, which means that the Vulkan allocator still
```
jonahwilliams, line 29: Its not that VMA still has a reference, its that when shutting down VMA we have retained dedicated allocations without freeing them

impeller/aiks/aiks_unittests.cc
```diff
@@ -3362,5 +3362,31 @@ TEST_P(AiksTest, CaptureInactivatedByDefault) {
   ASSERT_FALSE(GetContext()->capture.IsActive());
 }
 
+// Regression test for https://github.com/flutter/flutter/issues/134678.
+TEST_P(AiksTest, ReleasesTextureOnTeardown) {
```
jonahwilliams, line 3366: Does this actually work on a "real" backend - does that matter? In theory if we've submitted a frame the fence waiter should keep the weak texture alive.

impeller/aiks/aiks_playground_inspector.h
matanlurey: Done.

impeller/aiks/aiks_unittests.cc
matanlurey: Does (a) this test work on a real backend, (b) do we avoid the underlying issue on a real backend?

(a) No, at least I've not tried it. I just want to make sure we can enable validation layers for `aiks_unittests`.
     (https://github.com/flutter/flutter/issues/133506)

(b) This probably is, as you've mentioned, show a weakness in the fence waiter that needs to be fixed as well.

jonahwilliams: LGTM, though we may need to adjust this further

jonahwilliams (approved):
LGTM

impeller/aiks/aiks_unittests.cc
matanlurey: Oh, wait - your point is if the fence waiter was working properly we wouldn't need this :)?

jonahwilliams: I think that if the fence waiter was working this may not be released by the end of the scope

jonahwilliams: it would only be after Shutdown. So maybe manually call contextvk->Shutdown() and then check?

matanlurey: Just to clarify, this test does fail (100% of the time) before our patch as-is.

Do you think I should add _another_ test or tweak this tests behavior?

jonahwilliams: or, will become flaky after we fix fence waiter

matanlurey: Gotcha, added a comment and ran 1000 times to ensure it's fine right now.

PR #45848, opened 2023-09-14
[Impeller] Assert IMPELLER_ENABLE_3D when importing scene_contents.h.

bdero:
Noticed we had a residual unguarded import of scene_contents.h while investigating https://github.com/flutter/flutter/issues/134745.


matanlurey (approved):
Thanks!

PR #45850, opened 2023-09-14
Revert "[Impeller] Avoid inserting additional save layers based on clip configuration."

jonahwilliams:
Reverts flutter/engine#43759


jonahwilliams:
The save layer introduced by selecting the AAwithSaveLayer is observable, so optimizing it out is not generally safe.

jonahwilliams:
Looks like we can't do a straight revert, will put together a limited revert next week

PR #45895, opened 2023-09-15
[Impeller] dont cache failed render target allocations.

jonahwilliams:
Otherwise we'll null de-ref when trying to create a texture with the same descriptor later on.


matanlurey (approved):
LGTM!

PR #45910, opened 2023-09-16
Revert "[Impeller] construct text frames on UI thread."

jonahwilliams:
Reverts flutter/engine#45418

Some google3 tests are hitting the CHECK I added in the DlSkCanvasDispatcher::drawTextFrame, which indicates that the SkParagraph code likely thinks impeller is enabled, whereas other code might be running with Skia.

Perhaps this could happen if its software rendering? It should be a fatal error on startup so we can track this down.


jason-simmons (approved):

XilaiZhang (approved):
images (1)

PR #45981, opened 2023-09-18
[web] Rename EngineFlutterWindow => EngineSingletonFlutterWindow

mdebbar:
Make it clear that this class is a singleton.


ditman (approved):
If you want to land this rename, land it, but I think its utility is going to be limited (or short lived).

Take into account that `ui.SingletonFlutterWindow` is now deprecated:

* https://github.com/flutter/engine/blob/10c480310926d7e04a77ba8a04d321f9d225a545/lib/ui/window.dart#L397-L402

Maybe a better approach here is to make `class EngineFlutterView extends ui.FlutterView`? (that's probably a much bigger change, because we're going to be missing a lot of getters!)

mdebbar:
@ditman you are right, this class is going away.

What prompted me to do this rename was [your comment](https://github.com/flutter/engine/pull/45295#discussion_r1325130751). I felt maybe it was not clear that this is a singleton so I wanted to make it clear.

ditman:
Ahhh, I got you! I think the current name is "correct", except it should eventually `extends ui.FlutterView + our web interface` rather than the `ui.SingletonFlutterWindow`, apologies for miscomunicating this (x_x)

PR #45994, opened 2023-09-18
[Impeller] Update CAMetalLayer to be opaque

jonahwilliams:
this setting improves the performance of the "CPU TO Display Latency" metric in metal system trace. On high frame rate devices, this reduces the time we spend waiting for drawable availibility. This only applies if the CAMetalLayer is fullscreen, as it allows a direct to display optimization that reduces "CPU TO Display Latency" from about 21 ms down to 14ms.

Xref: https://github.com/flutter/flutter/issues/134959

### DIFF
--- a/impeller/renderer/backend/metal/surface_mtl.mm
+++ b/impeller/renderer/backend/metal/surface_mtl.mm
@@ -218,6 +218,14 @@ IRect SurfaceMTL::coverage() const {
   return IRect::MakeSize(resolve_texture_->GetSize());
 }
 
+bool ShouldPresentSync() {
+  #ifdef FML_OS_IOS
+    return [[NSThread currentThread] isMainThread];
+  #else
+    return true;
+  #endif // FML_OS_IOS
+}
+
 // |Surface|
 bool SurfaceMTL::Present() const {
   auto context = context_.lock();
@@ -252,9 +260,14 @@ bool SurfaceMTL::Present() const {
     id<MTLCommandBuffer> command_buffer =
         ContextMTL::Cast(context.get())
             ->CreateMTLCommandBuffer("Present Waiter Command Buffer");
-    [command_buffer commit];
-    [command_buffer waitUntilScheduled];
-    [drawable_ present];
+    if (ShouldPresentSync()) {
+      [command_buffer commit];
+      [command_buffer waitUntilScheduled];
+      [drawable_ present];
+    } else {
+      [command_buffer presentDrawable:drawable_];
+      [command_buffer commit];
+    }
   }
 
   return true;
--- a/impeller/renderer/backend/metal/surface_mtl.mm
+++ b/impeller/renderer/backend/metal/surface_mtl.mm
@@ -219,11 +219,11 @@ IRect SurfaceMTL::coverage() const {
 }
 
 bool ShouldPresentSync() {
-  #ifdef FML_OS_IOS
-    return [[NSThread currentThread] isMainThread];
-  #else
-    return true;
-  #endif // FML_OS_IOS
+#ifdef FML_OS_IOS
+  return [[NSThread currentThread] isMainThread];
+#else
+  return true;
+#endif  // FML_OS_IOS
 }
 
 // |Surface|
--- a/impeller/renderer/backend/metal/surface_mtl.mm
+++ b/impeller/renderer/backend/metal/surface_mtl.mm
@@ -218,14 +218,6 @@ IRect SurfaceMTL::coverage() const {
   return IRect::MakeSize(resolve_texture_->GetSize());
 }
 
-bool ShouldPresentSync() {
-#ifdef FML_OS_IOS
-  return [[NSThread currentThread] isMainThread];
-#else
-  return true;
-#endif  // FML_OS_IOS
-}
-
 // |Surface|
 bool SurfaceMTL::Present() const {
   auto context = context_.lock();
@@ -260,14 +252,9 @@ bool SurfaceMTL::Present() const {
     id<MTLCommandBuffer> command_buffer =
         ContextMTL::Cast(context.get())
             ->CreateMTLCommandBuffer("Present Waiter Command Buffer");
-    if (ShouldPresentSync()) {
-      [command_buffer commit];
-      [command_buffer waitUntilScheduled];
-      [drawable_ present];
-    } else {
-      [command_buffer presentDrawable:drawable_];
-      [command_buffer commit];
-    }
+    [command_buffer commit];
+    [command_buffer waitUntilScheduled];
+    [drawable_ present];
   }
 
   return true;
--- a/shell/platform/darwin/ios/framework/Source/FlutterView.mm
+++ b/shell/platform/darwin/ios/framework/Source/FlutterView.mm
@@ -73,12 +73,6 @@
                            "supported, downgrading the color gamut to sRGB.";
     }
     self.layer.opaque = opaque;
-
-    // This line is necessary. CoreAnimation(or UIKit) may take this to do
-    // something to compute the final frame presented on screen, if we don't set this,
-    // it will make it take long time for us to take next CAMetalDrawable and will
-    // cause constant junk during rendering.
-    self.backgroundColor = UIColor.clearColor;
   }
 
   return self;
--- a/shell/platform/darwin/ios/framework/Source/FlutterView.mm
+++ b/shell/platform/darwin/ios/framework/Source/FlutterView.mm
@@ -73,6 +73,17 @@
                            "supported, downgrading the color gamut to sRGB.";
     }
     self.layer.opaque = opaque;
+
+    // This line is necessary. CoreAnimation(or UIKit) may take this to do
+    // something to compute the final frame presented on screen, if we don't set this,
+    // it will make it take long time for us to take next CAMetalDrawable and will
+    // cause constant junk during rendering. Note: setting any background color will
+    // disable opaque, which on full screen flutter apps removes the direct to display
+    // optimization. This can increase presentation time by a frame interval, which
+    // is problematic on high frame rate devices.
+    if (!opaque) {
+      self.backgroundColor = UIColor.whiteColor;
+    }
   }
 
   return self;
--- a/shell/platform/darwin/ios/framework/Source/FlutterView.mm
+++ b/shell/platform/darwin/ios/framework/Source/FlutterView.mm
@@ -72,7 +72,6 @@
       FML_DLOG(WARNING) << "Rendering wide gamut colors is turned on but isn't "
                            "supported, downgrading the color gamut to sRGB.";
     }
-    self.layer.opaque = opaque;
 
     // This line is necessary. CoreAnimation(or UIKit) may take this to do
     // something to compute the final frame presented on screen, if we don't set this,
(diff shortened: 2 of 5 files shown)


impeller/renderer/backend/metal/surface_mtl.mm
```diff
@@ -252,9 +260,14 @@
...
-    [command_buffer waitUntilScheduled];
-    [drawable_ present];
+    if (ShouldPresentSync()) {
+      [command_buffer commit];
+      [command_buffer waitUntilScheduled];
+      [drawable_ present];
+    } else {
+      [command_buffer presentDrawable:drawable_];
+      [command_buffer commit];
+    }
```
jonahwilliams, line 270: This should be safe if we're not presenting with transaction, but needs more testing from me.

shell/platform/darwin/ios/framework/Source/FlutterView.mm
```diff
@@ -77,8 +77,13 @@ - (instancetype)initWithDelegate:(id<FlutterViewEngineDelegate>)delegate
     // This line is necessary. CoreAnimation(or UIKit) may take this to do
     // something to compute the final frame presented on screen, if we don't set this,
     // it will make it take long time for us to take next CAMetalDrawable and will
-    // cause constant junk during rendering.
-    self.backgroundColor = UIColor.clearColor;
+    // cause constant junk during rendering. Note: setting any background color will
+    // disable opaque, which on full screen flutter apps removes the direct to display
+    // optimization. This can increase presentation time by a frame interval, which
+    // is problematic on high frame rate devices.
+    if (!opaque) {
```
jonahwilliams, line 84: @luckysmg , I found that if we set the background color to any value, it resets the value of layer.opaque. If you have a fullscreen flutter view and opaque is true, then we get a direct to display optimization and the presentation delay decreases from ~21ms to ~14ms, which is particularly important for high frame rate devices.

See some findings in https://github.com/flutter/flutter/issues/134959

How are you using a FlutterView? Is it an add2app case where it is not the only view in the scene?

luckysmg: Yes. We will change the layer's opaque to make the flutter view to be transparent and present it. The scene is that the new page is the transparent page, and it is presented on current page to do some thing like modal bottom sheet.

See https://github.com/alibaba/flutter_boost/blob/13eb9aed1dff9fcdfa6066666ba08a11b231af9f/ios/Classes/container/FBFlutterViewContainer.m#L137

luckysmg: Also see https://github.com/alibaba/flutter_boost/blob/13eb9aed1dff9fcdfa6066666ba08a11b231af9f/ios/Classes/container/FBFlutterViewContainer.m#L221

luckysmg: `If !opaque` and set a white color, will it make the `FlutterView` not transparent?

jonahwilliams: ahh ok. Right now the problem is that we're unconditionally setting this clear color here. But setting the background color resets the value of opaque to NO, even if we requested it to be YES. So I think as long as your code is setting opaque = NO, this should keep working for you?

luckysmg: Seems not working... Here init with delegate is too early to set bg color to white.

See
https://github.com/alibaba/flutter_boost/blob/13eb9aed1dff9fcdfa6066666ba08a11b231af9f/ios/Classes/container/FBFlutterViewContainer.m#L220

If we need the view is transparent, we will set the `opaque` to `NO` and don't set bg color(in above link's code)

Before this patch: the result FlutterView is will be not opaque and with clear color, so it will be transparent ✅
After this patch: the result FlutterView is will be not opaque and with white color, so it will not be transparent ❌

shell/platform/darwin/ios/ios_surface_metal_impeller.mm
```diff
@@ -55,6 +55,10 @@
   // Flutter needs to read from the color attachment in cases where there are effects such as
   // backdrop filters. Flutter plugins that create platform views may also read from the layer.
   layer.framebufferOnly = NO;
+  // Note: this setting improves the performance of the "CPU TO Display Latency" metric in
+  // metal system trace. On high frame rate devices, this reduces the time we spend waiting
+  // for drawable availibility.
+  layer.opaque = YES;
```
luckysmg, line 61: Emm If we set this `opaque=YES`, means our the change to `opaque` is invalid? Since this will be called every frame.

shell/platform/darwin/ios/framework/Source/FlutterView.mm
jonahwilliams: So your code is starting out with opaque = YES, then you're setting it to opaque = NO later? I'm not really familiar enough with flutter_boost to follow the rendering paths there. I can change this so that if you later set opaque = NO, then we also set the background color to the clearColor?

luckysmg: <  I can change this so that if you later set opaque = NO, then we also set the background color to the clearColor?

I think this can work. But I m not sure this is good or not... Because this has a implicit operation. Change `opaque`also change the bg color, maybe sometimes will confusing the developer..

jonahwilliams: Yeah good point. I need to test if setting opaque _after_ we set the background color also resets the background color too. If not, maybe I just change the order and leave this as is.

luckysmg: ^_^

shell/platform/darwin/ios/ios_surface_metal_impeller.mm
jonahwilliams: I'll remove this

jonahwilliams: Just testing 😄

luckysmg: Yeah ^_^

shell/platform/darwin/ios/framework/Source/FlutterView.mm
```diff
@@ -96,7 +101,7 @@ - (void)layoutSubviews {
     layer.allowsGroupOpacity = YES;
     layer.contentsScale = screenScale;
     layer.rasterizationScale = screenScale;
-    layer.framebufferOnly = flutter::Settings::kSurfaceDataAccessible ? NO : YES;
```
jonahwilliams, line 99: This value should match what we set in https://github.com/flutter/engine/blob/3303a3e040b1ea27a4a732612eae32c0eaf5d21f/shell/platform/darwin/ios/ios_surface_metal_impeller.mm#L57

luckysmg:
Have a locally test for current patch. The constant junk is regression. I think the patch will also impact device not on iOS16.6. 😭

https://github.com/flutter/engine/assets/49340347/afc55803-bb9f-44ba-a7e5-3d8ec165b4eb

jonahwilliams:
Are you setting your FlutterViews opaque to YES or are you relying on the side effect of the bgcolor setting to make it opaque?

jonahwilliams:
I wouldn't consider this a fix for 16.6, the fix for that is to update to 17. But this does help, and fixes a bug since right now we're ignoring the opaque property entirely

luckysmg:
I test with a pure flutter example without add to app. I didn't change any code oof flutter view

jonahwilliams:
Can you run metal system trace and share a screenshot of the LCD category? For me, this is showing ~8ms faster presentations with this change.

luckysmg:
Will have a check later

jonahwilliams:
I'm testing on 17, but even on 16.6 I don't see this causing regressions - if anything its stabilizing things and I'm seeing more throughput. But event with a fast app we'll always hit a point where we're limited by drawables, given that there are only 3.

I'm not sure what to make of this right now.

jonahwilliams:
If this is the case, then I suspect that setting the background color to the clearColor in https://github.com/flutter/engine/pull/39359 fixed things for you by way of reseting the opaque bit. But I'm not sure why that would improve performance, since that seems to disable direct to display.

I'm not sure if we use that opaque bit elsewhere, but we might be in a bit of an inconsistent state because while the layer is no opaque NO, the viewcontroller things opaque is YES 🤔

luckysmg:
Yes. Confusing..... I will have a recheck to check whether my code is correct.

jonahwilliams:
You might also try patching in a change that remove the backgroundColor but makes the default opaque to NO, to see if it has consistent performance. That should verify if it is the opaque field or the background color that is helping your case.

luckysmg:
Have a retest. 
1. The patch on my iPhone indeed regress the constant junk issue.
2. After I set the code like this.

```
// This line is necessary. CoreAnimation(or UIKit) may take this to do
    // something to compute the final frame presented on screen, if we don't set this,
    // it will make it take long time for us to take next CAMetalDrawable and will
    // cause constant junk during rendering. Note: setting any background color will
    // disable opaque, which on full screen flutter apps removes the direct to display
    // optimization. This can increase presentation time by a frame interval, which
    // is problematic on high frame rate devices.
//    self.backgroundColor = UIColor.clearColor;
    // Make sure to set this value after backgroundColor above, otherwise it will
    // reset the value of opaque.
    self.layer.opaque = NO;
```

The screen will have jitter, but behavior is diffrent from 1 point. I think this is because my iPhone is 16.6, I think maybe this code have no jitter on other version of iOS

luckysmg:
@jonahwilliams I have a project to fire on later, not sure how long to have next response.. (just ensure you are not waiting for my resp because it is mid night in your time haha ^_^)

jonahwilliams:
No worries, I'm marking this as a draft for now while I investigate other things!

luckysmg:
@jonahwilliams I retest using iPhone13 Pro Max on iOS 16.0

1. Using the code below:
```
//    self.backgroundColor = UIColor.clearColor;
    self.layer.opaque = YES;
```
Will regress the constant junk issue fixed in https://github.com/flutter/engine/pull/39359

2. Using the code in this patch: The same as 1, will also regress constant junk.
3. Using the code below:
```
//    self.backgroundColor = UIColor.clearColor;
    self.layer.opaque = NO;
```
The app is smooth without any junk.

Also I see the CALayer's `opaque` default value is `NO`, I m not sure whether that is apple's suggestion and indicates the performance will be better.
image

jonahwilliams:
OK, I think we can work with this. I suspect this explains why setting the background color improved performance then, because we were defaulting to Opaque = YES, and then changing the color reset it to NO?

jonahwilliams:
I'll need to do some follow ups and figure out what the official story is on Layer.opaque.

luckysmg:
> because we were defaulting to Opaque = YES, and then changing the color reset it to NO?

Yes. I think so..

jonahwilliams:
Sorry to ping on this again @luckysmg , but what is the CPU to display latency look like if you don't have this change patched in?

luckysmg:
I think without this change, it is the same in `main` channel, maybe you could have a try. Because I only have a iOS16.6 device, I m afraid the behavior on my  device can not be referred to....

jonahwilliams:
Makes sense, no worries. Still researching this 💻 🧑‍🔬 !

jonahwilliams:
Closing as this seems to be a dead end.

PR #45781, opened 2023-09-13
[ios] use python script to generate extension safe frameworks and code sign them

cyanglaz:
Generates the fat framework for extension safe variants and add them to the code sign list.

Also reverts part of https://github.com/flutter/engine/pull/45664 where the variants are uploaded in separate archive. Now with this change, the extension safe is inside a "extension_safe" folder under existing archive. 

fixes https://github.com/flutter/flutter/issues/124291

The destination folder after this change:
- An extension_safe folder is added, which includes the extension safe Flutter.frameworks and dsym
- The artifacts.zip now includes the `extension_safe` except the dysm
- A extension_safe_Flutter.dsym.zip is created (under ios/release/) , which includes the extension_safe/Flutter.dsym

Screenshot 2023-09-13 at 3 26 23 PM

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [Flutter Style Guide] and the [C++, Objective-C, Java style guides].
- [x] I listed at least one issue that this PR fixes in the description above.
- [ ] I added new tests to check the change I am making or feature I am adding, or the PR is [test-exempt]. See [testing the engine] for instructions on writing and running engine tests.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I signed the [CLA].
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/flutter/wiki/Tree-hygiene#overview
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[test-exempt]: https://github.com/flutter/flutter/wiki/Tree-hygiene#tests
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[C++, Objective-C, Java style guides]: https://github.com/flutter/engine/blob/main/CONTRIBUTING.md#style
[testing the engine]: https://github.com/flutter/flutter/wiki/Testing-the-engine
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/wiki/Chat

PR #46004, opened 2023-09-18
Revert "[ios] use python script to generate extension safe frameworks and code sign them"

cyanglaz:
Reverts flutter/engine#45781

Issue with engine roll https://github.com/flutter/flutter/pull/134964.
We need to update the conductor and performa manual roll.
We also need to remove the Flutter.framework from the artifact.zip

### DIFF
--- a/ci/builders/mac_ios_engine.json
+++ b/ci/builders/mac_ios_engine.json
@@ -290,6 +290,53 @@
                 "language": "python3"
             },
             {
+                "name": "Debug-ios-Flutter-Extension-Safe.xcframework",
+                "parameters": [
+                    "--dst",
+                    "out/debug_extension_safe",
+                    "--arm64-out-dir",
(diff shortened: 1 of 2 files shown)


christopherfujino (commented):
so this revert will make the conductor test green because the affected binaries are in a completely different archive that the tool doesn't download?

christopherfujino (approved):
RSLGTM to unblock engine roll

cyanglaz:
> so this revert will make the conductor test green because the affected binaries are in a completely different archive that the tool doesn't download?

After the revert, the binaries are not uploaded anymore.

christopherfujino:
> > so this revert will make the conductor test green because the affected binaries are in a completely different archive that the tool doesn't download?
> 
> After the revert, the binaries are not uploaded anymore.

Ahh ok, cool

fluttergithubbot (approved):

PR #46013, opened 2023-09-18
[Impeller] Hold the CommandPoolVK at a higher scope.

matanlurey:
Fixes https://github.com/flutter/flutter/issues/134982.

I could never reproduce the flakes locally, but this should prevent `pool1 == pool2`.


bdero (approved):
LGTM

PR #46023, opened 2023-09-19
[Impeller] Use final cmd buffer to present drawable.

jonahwilliams:
If we're not running with a transaction then we don't need to block on waitUntilScheduled on any platform. I think this should always work, as the problems we had before were due to always using drawable present. But this helper method schedules the drawable presentation after the cmd buffer is scheduled - which is what we're doing with waitUntilScheduled anyway - just non blocking from our perspective.

Fixes https://github.com/flutter/flutter/issues/131520


impeller/renderer/backend/metal/surface_mtl.mm
```diff
@@ -248,13 +248,17 @@
...
             ->CreateMTLCommandBuffer("Present Waiter Command Buffer");
-    [command_buffer commit];
-    [command_buffer waitUntilScheduled];
-    [drawable_ present];
+    if ([[NSThread currentThread] isMainThread]) {
+      [command_buffer commit];
+      [command_buffer waitUntilScheduled];
+      [drawable_ present];
+    } else {
+      [command_buffer presentDrawable:drawable_];
```
jonahwilliams, line 259: Oh wait, I need to double check if this messes up frame capture. If so we need to check if we're capturing a frame and block.

jonahwilliams: Done

impeller/renderer/backend/metal/surface_mtl.mm
```diff
@@ -248,13 +248,21 @@
...
     id<MTLCommandBuffer> command_buffer =
         ContextMTL::Cast(context.get())
             ->CreateMTLCommandBuffer("Present Waiter Command Buffer");
-    [command_buffer commit];
-    [command_buffer waitUntilScheduled];
-    [drawable_ present];
+    // If the threads have been merged, or there is a pending frame capture,
+    // then block on cmd buffer scheduling to ensure that the
+    // transaction/capture work correctly.
+    if ([[NSThread currentThread] isMainThread] ||
```
bdero, line 257: Would this check work if a custom embedder launched the engine in a thread other than the main thread?

bdero (approved):
LGTM modulo some questions.

FWIW, I tried this with Wondrous on my 12 mini and I'm not seeing any platform view flickering going on.

impeller/renderer/backend/metal/surface_mtl.mm
jonahwilliams: I think technically this only needs to match wherever we set the layer to presentsWithTransaction. But I suspect that is only for iOS? I'll have to check how macOS desktop is doing.

chinmaygarde (approved):

PR #46037, opened 2023-09-19
[Android] Fix enableSuggestions set to false not honored

bleroux:
## Description

This PR fixes an issue where setting `TextField.enableSuggestions` to false was not honored on Android.

Several Android devices (Samsung) and/or IMEs (including GBoard) does not disabled suggestions when `InputType.TYPE_TEXT_FLAG_NO_SUGGESTIONS `flag is set. The common solution is to rely on the following flag:  ~~`InputType.TYPE_TEXT_VARIATION_PASSWORD`~~ `InputType.TYPE_TEXT_VARIATION_VISIBLE_PASSWORD`.

Reference:
- https://issuetracker.google.com/issues/36934423#comment4
- https://stackoverflow.com/questions/33148168/inputtype-type-text-flag-no-suggestions-in-samsung/33227237#33227237
- Existing comment on the codebase:
https://github.com/flutter/engine/blob/53430c7c96a3ef1fc78d63ea6b0c35e5ff46c45e/shell/platform/android/io/flutter/plugin/editing/TextInputPlugin.java#L270

## Related Issue

Fixes https://github.com/flutter/flutter/issues/71679.

## Tests

Adds 1 test.

### DIFF
--- a/shell/platform/android/io/flutter/plugin/editing/TextInputPlugin.java
+++ b/shell/platform/android/io/flutter/plugin/editing/TextInputPlugin.java
@@ -275,7 +275,7 @@ public class TextInputPlugin implements ListenableEditingState.EditingStateWatch
       if (!enableSuggestions) {
         // Note: both required. Some devices ignore TYPE_TEXT_FLAG_NO_SUGGESTIONS.
         textType |= InputType.TYPE_TEXT_FLAG_NO_SUGGESTIONS;
-        textType |= InputType.TYPE_TEXT_VARIATION_PASSWORD;
+        textType |= InputType.TYPE_TEXT_VARIATION_VISIBLE_PASSWORD;
       }
     }
 
--- a/shell/platform/android/io/flutter/plugin/editing/TextInputPlugin.java
+++ b/shell/platform/android/io/flutter/plugin/editing/TextInputPlugin.java
@@ -272,7 +272,11 @@ public class TextInputPlugin implements ListenableEditingState.EditingStateWatch
       textType |= InputType.TYPE_TEXT_VARIATION_PASSWORD;
     } else {
       if (autocorrect) textType |= InputType.TYPE_TEXT_FLAG_AUTO_CORRECT;
-      if (!enableSuggestions) textType |= InputType.TYPE_TEXT_FLAG_NO_SUGGESTIONS;
+      if (!enableSuggestions) {
+        // Note: both required. Some devices ignore TYPE_TEXT_FLAG_NO_SUGGESTIONS.
+        textType |= InputType.TYPE_TEXT_FLAG_NO_SUGGESTIONS;
+        textType |= InputType.TYPE_TEXT_VARIATION_PASSWORD;
+      }
     }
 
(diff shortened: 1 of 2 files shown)


shell/platform/android/io/flutter/plugin/editing/TextInputPlugin.java
```diff
@@ -272,7 +272,11 @@ private static int inputTypeFromTextInputType(
       textType |= InputType.TYPE_TEXT_VARIATION_PASSWORD;
     } else {
       if (autocorrect) textType |= InputType.TYPE_TEXT_FLAG_AUTO_CORRECT;
-      if (!enableSuggestions) textType |= InputType.TYPE_TEXT_FLAG_NO_SUGGESTIONS;
+      if (!enableSuggestions) {
+        // Note: both required. Some devices ignore TYPE_TEXT_FLAG_NO_SUGGESTIONS.
+        textType |= InputType.TYPE_TEXT_FLAG_NO_SUGGESTIONS;
+        textType |= InputType.TYPE_TEXT_VARIATION_PASSWORD;
```
camsim99, line 278: Why this flag versus `TYPE_TEXT_VARIATION_VISIBLE_PASSWORD`? The links you mentioned point to that flag, the concern I assume being that this may obscure what the user is typing.

shell/platform/android/test/io/flutter/plugin/editing/TextInputPluginTest.java
```diff
@@ -273,7 +273,7 @@ public void textEditingDelta_TestUpdateEditingValueWithDeltasIsNotInvokedWhenDel
             true,
             false, // Delta model is disabled.
             TextInputChannel.TextCapitalization.NONE,
-            new TextInputChannel.InputType(TextInputChannel.TextInputType.TEXT, false, false),
+            new TextInputChannel.InputType(TextInputChannel.TextInputType.MULTILINE, false, false),
```
camsim99, line 276: Why this change?

shell/platform/android/io/flutter/plugin/editing/TextInputPlugin.java
bleroux: You are perfectly right that the flag should be `TYPE_TEXT_VARIATION_VISIBLE_PASSWORD`. While testing on my device (Pixel 7 / GBoard), I did not notice this mistake because the keyboard is the expected one in both cases.

> this may obscure what the user is typing.

This won't have such an effect because the obscure logic is on the Dart side. Nonetheless, this is a really good catch because using the right value makes this PR coherent with the links listed in its description.

shell/platform/android/test/io/flutter/plugin/editing/TextInputPluginTest.java
bleroux: Oops my bad... that's the test I copy/pasted to bootstrap the new one, it looks like I updated it by mistake at one point. :facepalm:

camsim99 (approved):
LGTM!

PR #46044, opened 2023-09-19
[web] Make `PlatformViewManager` a clear singleton

mdebbar:
Make it clear that `PlatformViewManager` is a singleton and follow the patterns that we use for other singleton classes.


lib/web_ui/lib/src/engine/platform_views/content_manager.dart
```diff
@@ -38,6 +38,11 @@ class PlatformViewManager {
     );
   }
 
+  /// The shared instance of PlatformViewManager shared across the engine to handle
+  /// rendering of PlatformViews into the web app.
+  // TODO(dit): How to make this overridable from tests?
+  static final PlatformViewManager instance = PlatformViewManager();
```
ditman, line 44: This is now very easy to override for tests! When we need that, this just needs to be non-final and have a setter for the testing environment. (Same as all the other singletons!)

ditman (approved):
LGTM! Great change! I love how everything is converging to either SingletonClassName.instance or something that lives in the new WebEngineView interface.

PR #46046, opened 2023-09-19
[web] Stop using `flutterViewEmbedder` for platform views

mdebbar:
As of right now, there is an indirect link between `flutter/platform_views` messages and `flutterViewEmbedder`, expressed by the callback that's being passed to the `PlatformViewMessageHandler`.

This PR proposed making this relationship clear between `flutter/platform_views` and `implicitView` (i.e. the singleton `window`).

This PR also opens the path for each view to have its own `PlatformViewMessageHandler` instance.

Depends on https://github.com/flutter/engine/pull/46044


lib/web_ui/test/engine/platform_views/message_handler_test.dart
```diff
@@ -89,27 +90,37 @@ void testMain() {
...
-                  'The element created by the factory should be present in the created view.');
-          expect(codec.decodeEnvelope(response!), isNull,
-              reason:
-                  'The response should be a success envelope, with null in it.');
+          expect(
+            platformViewsContainer.children.single,
+            isNotNull,
+            reason: 'The container has a single child, the created view.',
+          );
+          final DomElement platformView = platformViewsContainer.lastElementChild!;
```
ditman, line 113: No need to add `lastElementChild` for this test, right? can this be:

```suggestion
          final DomElement platformView = platformViewsContainer.children.single!;
```

?

lib/web_ui/lib/src/engine/platform_views/message_handler.dart
```diff
@@ -98,7 +99,7 @@ class PlatformViewMessageHandler {
 
     // For now, we don't need anything fancier. If needed, this can be converted
     // to a PlatformViewStrategy class for each web-renderer backend?
-    _contentHandler?.call(content);
+    _platformViewsContainer.append(content);
```
ditman, line 102: Much simpler, this is fantastic!

ditman (approved):
The `data` from this call must contain the viewId that is the container of the platform view. I guess some of the "handlePlatformViewCall" will need to be moved to the platform_dispatcher, to know what view is being used (at least).

((We can probably have a generic handler that extracts the `viewId` from any message, and reuse that, but that's for the future!))

mdebbar:
> The `data` from this call must contain the viewId that is the container of the platform view.

Good point. I'm preparing a framework PR that sends `flutterViewId` when creating a platform view. Then we can start using it in the engine.

> I guess some of the "handlePlatformViewCall" will need to be moved to the platform_dispatcher, to know what view is being used (at least).

The way I was thinking of doing it was: `PlatformDispatcher` extracts `flutterViewId` from the platform message, and based on that, it picks the right `view` from the `viewData` map. The `view` instance should have its own `platformViewMessageHandler` (added by this PR) that can handle the platform message.

ditman:
> PlatformDispatcher extracts flutterViewId from the platform message, and based on that, it picks the right view from the viewData map. The view instance should have its own platformViewMessageHandler (added by this PR) that can handle the platform message.

Agree. Also the bit that extracts the `flutterViewId`, if we settle in a standard (maybe @goderbauer has an idea for this) can be shared across the whole engine (instead of having an ad-hoc decoder every time we need to extract a viewId from a platform message)

Still LGTM!

wanjm:
when will it be merged?

PR #46047, opened 2023-09-19
[web] DOM objects implement JS object

mdebbar:
Instead of doing `as JSAny` for DOM objects that are defined in static JS interop, let's make them `implements JSObject`?

cc @eyebrowsoffire @srujzs thoughts?


srujzs (approved):
LGTM! At some point we may migrate all JS types to extension types for readability, but that'll be further down the road and we'll probably migrate this entire file to extension types before then.

PR #46054, opened 2023-09-19
Rename `layoutGoals` to `kLayoutGoals` to enforce lints on headers.

matanlurey:
This is required to eventually land https://github.com/flutter/flutter/issues/134969 (lint header files).


cbracken (approved):
Nice catch! I thought I'd landed enforcement of this already - apparently not!

lgtm.

PR #46058, opened 2023-09-19
Conform to clang_tidy in `client_wrapper` headers.

matanlurey:
Partial work towards https://github.com/flutter/flutter/issues/134969.

All of these were auto-suggested by Clang, and mostly avoid unnecessary copies.


shell/platform/common/client_wrapper/include/flutter/basic_message_channel.h
```diff
@@ -77,7 +79,7 @@ class BasicMessageChannel {
     BinaryMessageHandler binary_handler = [handler, codec, channel_name](
                                               const uint8_t* binary_message,
                                               const size_t binary_message_size,
-                                              BinaryReply binary_reply) {
+                                              const BinaryReply& binary_reply) {
```
loic-sharma, line 82: Is `BinaryReply` not movable? It looks like it should be: https://github.com/flutter/engine/blob/28f14e6eec4f4dbe6d678f758ca7ff4654473f54/shell/platform/common/client_wrapper/include/flutter/binary_messenger.h#L16-L17

matanlurey: A const reference is not copied, it essentially accomplishes the same thing

shell/platform/common/client_wrapper/include/flutter/event_stream_handler.h
```diff
@@ -20,8 +20,8 @@ struct StreamHandlerError {
   const std::string error_message;
   const std::unique_ptr<T> error_details;
 
-  StreamHandlerError(const std::string error_code,
-                     const std::string error_message,
+  StreamHandlerError(const std::string& error_code,
+                     const std::string& error_message,
```
loic-sharma, line 24: From my understanding, lines 26 and 27 will copy. Could we use `std::move` instead on lines 26 and 27 to allow the caller to avoid copies entirely?

matanlurey: Same Loic, this is the result of clang_tidy --fix. A reference is not copied.

shell/platform/common/client_wrapper/testing/test_codec_extensions.h
```diff
@@ -32,7 +32,7 @@ class Point {
 // variable-length type that includes types handled by the core standard codec.
 class SomeData {
  public:
-  SomeData(const std::string label, const std::vector<uint8_t>& data)
+  SomeData(const std::string& label, const std::vector<uint8_t>& data)
```
loic-sharma, line 35: Could we use `std::move` on line 36 as well here?

matanlurey: Ditto above.

shell/platform/common/client_wrapper/include/flutter/basic_message_channel.h
loic-sharma: Copying my message from: https://discord.com/channels/608014603317936148/608021010377080866/1153752411614220310

> The reference argument doesn't copy, but doesn't assigning the reference argument to a non-reference instance field result in a copy?

shell/platform/common/client_wrapper/include/flutter/event_stream_handler.h
stuartmorgan: I think changing the params here was a mistake when we fixed a serious bug with the ivars in https://github.com/flutter/engine/pull/40311.

While we could keep this signature to allow [copy elision](https://abseil.io/tips/117), it seems very unlikely that the performance of copying error strings (which are probably short, and shouldn't be frequent) is worth having non-standard argument types.

shell/platform/common/client_wrapper/testing/test_codec_extensions.h
stuartmorgan: We don't need to micro-optimize test code performance.

shell/platform/common/client_wrapper/include/flutter/event_stream_handler.h
matanlurey: Stuart, just to clarify, do you want this change reverted/changed to something else?

shell/platform/common/client_wrapper/include/flutter/basic_message_channel.h
stuartmorgan: There's more potential argument for supporting copy elision here and sprinkling in some `move`s, but setting handlers isn't a common operation, so absent evidence that we need it we should probably just change to the reference. I don't think I did this intentionally.

shell/platform/common/client_wrapper/include/flutter/event_stream_handler.h
stuartmorgan: I'm saying I think the removal of references on these two lines was probably a just mistake (cc @yaakovschectman in case I'm missing something looking back at that PR and review), and that the potential optimization Loic is suggesting here shouldn't be necessary here, and am thus voting for making the change you have here.

matanlurey: Ack, thanks.

matanlurey:
Loic, I think we should move forward with these changes as-is, as at _worst_ it's not worse performance or characteristics than the current code, and it unblocks https://github.com/flutter/flutter/issues/134969, which will catch back-slides or new additions.

loic-sharma (approved):
LGTM

stuartmorgan:
> as at _worst_ it's not worse performance or characteristics than the current code

This is not technically necessarily true, since the compiler could absolutely be doing implicit copy elision with the current code. But per my comments above I don't think any of these are cases where we need to intentionally worry about that (and haven't so far given the lack of explicit `move`s).

shell/platform/common/client_wrapper/include/flutter/basic_message_channel.h
cbracken: BinaryReply and BinaryMessageHandler are std::functions.
https://github.com/flutter/engine/blob/10c480310926d7e04a77ba8a04d321f9d225a545/shell/platform/common/client_wrapper/include/flutter/binary_messenger.h#L13-L24

Given that this is code that deals with async responses/message channels, are we introducing any lifetime issues by changing the third param to pass by reference rather than copy?

I'm actually sort of surprised that the compiler isn't shouting about a signature mismatch between the updated usage and definitions above given that it doesn't look like they were updated in this patch.

cbracken: Side note: If we were on C++23, which we're not, I suppose we'd have the option of [std::move_only_function](https://en.cppreference.com/w/cpp/utility/functional/move_only_function/move_only_function), which in theory restricts things a little more.

matanlurey: @stuartmorgan thoughts?

cbracken: To clarify, here's the scenario I'm thinking of:
1. You're in some block of code; let's say a function somewhere. You have some locals.
2. You construct a `BinaryReply` that captures (by reference) one or more of those locals. BinaryReply is a `std::function` so it can capture all it likes, unlike a function ptr.
3. You construct a `BinaryMessageHandler` using that `BinaryReply` which is, as of this change, being passed by reference, and you put that somewhere to be used later.
4. The scope in which you construct this thing ends and everything goes out of scope.

... time passes ...

6. The message handler invokes the BinaryReply and tries to make use of the captured local which is gone now.

stuartmorgan: From a technical standpoint I don't think this changes anything in your scenario. Arguably the by-value signature better communicates that capturing by reference in the fusion is a bad idea though?

cbracken: Ah yeah that's a good point. Those captures are going to be equally broken in a by-copy scenario. I suspect the only code that'll be broken by this is code that's already quite racy and unlikely to exist in practice. The BinaryReply itself would need to go out of scope between the time the BinaryMessageHandler is invoked with it and the time it's actually used, which seems pretty unlikely in practice.

cbracken (commented):
If we were on C++23 I'd prefer that these were just made std::move_only_function; but we're not and I'm not sure if/when that move is on the horizon (I suspect we're partially gated on what's available internally).

shell/platform/common/client_wrapper/include/flutter/basic_message_channel.h
stuartmorgan: I think we're saying slightly different things. My understanding is that there's no actual lifetime change here; AFAIK the by-value capture of the now-reference argument is still going to make a copy at that point.

PR #46060, opened 2023-09-19
[web] implement a workaround for JAWS for text nodes

yjbanov:
This implements a workaround for https://github.com/FreedomScientific/standards-support/issues/759 by moving the `aria-label` to an artificially added child span:

BEFORE:

```
<flt-semantics role="text" aria-label="Hello"></flt-semantics>
```

AFTER:

```
<flt-semantics role="text">
  <span aria-label="Hello"></span>
</flt-semantics>
```

The workaround is only used on Windows. Screen readers on other systems are happy with the original DOM structure. In fact, NVDA on Windows is also happy with the original, but we can't tell what screen reader is used on Windows, so we have to use the workaround for all of them.

### DIFF
--- a/lib/web_ui/lib/src/engine/semantics/label_and_value.dart
+++ b/lib/web_ui/lib/src/engine/semantics/label_and_value.dart
@@ -2,8 +2,28 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+import '../browser_detection.dart';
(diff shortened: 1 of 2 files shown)


mdebbar (approved):
LGTM!

yjbanov:
Still working on it.

yjbanov:
Closing. Will start a new PR.

PR #46061, opened 2023-09-19
Properly transfer objects between the main thread and web worker.

eyebrowsoffire:
We need to make sure to add objects to the transfer list when we send them across the ui thread/web worker boundary. Otherwise, they get copied, which is very expensive.

On my M1 MacBook Pro, I took measurements of scrolling in the material 3 demo. Before this change, the work on the web worker thread was taking about 25-40ms per frame. After the change, it's around 2ms.


yjbanov (approved):
lgtm

PR #46064, opened 2023-09-19
Made the warning about downgrading wide gamut happen at the correct time

gaaclarke:
fixes https://github.com/flutter/flutter/issues/135033

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/shell/platform/darwin/ios/framework/Source/FlutterView.mm
+++ b/shell/platform/darwin/ios/framework/Source/FlutterView.mm
@@ -48,6 +48,8 @@
     return NO;
   }
 
+  FML_DCHECK(self.screen);
+
   // This predicates the decision on the capabilities of the iOS device's
   // display.  This means external displays will not support wide gamut if the
   // device's display doesn't support it.  It practice that should be never.
@@ -68,10 +70,6 @@
   if (self) {
     _delegate = delegate;
     _isWideGamutEnabled = isWideGamutEnabled;
-    if (_isWideGamutEnabled && !self.isWideGamutSupported) {
-      FML_DLOG(WARNING) << "Rendering wide gamut colors is turned on but isn't "
-                           "supported, downgrading the color gamut to sRGB.";
-    }
     self.layer.opaque = opaque;
 
     // This line is necessary. CoreAnimation(or UIKit) may take this to do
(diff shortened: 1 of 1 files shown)


cyanglaz (approved):
LGTM

shell/platform/darwin/ios/framework/Source/FlutterView.mm
```diff
@@ -84,6 +82,16 @@ - (instancetype)initWithDelegate:(id<FlutterViewEngineDelegate>)delegate
   return self;
 }
 
+static void PrintWideGamutWarningOnce() {
+  static BOOL did_print = NO;
```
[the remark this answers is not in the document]
gaaclarke: I had that at first but then decided the table stakes are so low that threading doesn't need to happen.  Worst case scenario is that the print statement happens twice.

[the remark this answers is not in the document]
cyanglaz: make sense. LGTM!

shell/platform/darwin/ios/framework/Source/FlutterView.mm
```diff
@@ -97,7 +105,8 @@ - (void)layoutSubviews {
     layer.contentsScale = screenScale;
     layer.rasterizationScale = screenScale;
     layer.framebufferOnly = flutter::Settings::kSurfaceDataAccessible ? NO : YES;
```
jonahwilliams, line 107: drive by comment: shouldn't this be NO always? We set it to NO here: https://github.com/flutter/engine/blob/fdb33e4ee7a018badbd847d9f4cbf4ba3ca49efe/shell/platform/darwin/ios/ios_surface_metal_impeller.mm#L57

gaaclarke: I don't think so.  `YES` is the preferable setting since it has better performance.  I think I may have added this switch that allows you to turn it to NO so we can do wide gamut tests looking extended range pixels.  Its possible that NO is needed in that layer, but not on this one.  I don't know off the top of my head the relationship between these 2 layers.

jonahwilliams: This needs to be NO because of backdropfilters and platform views.

jonahwilliams: They're the same layer, so you're setting to YES and then its getting blown away. So just set it to NO here.

gaaclarke: I verified they are the same layer.  This seems messy if we have setup in 2 different places for the layer.  I guess this setting might be meaningful for the skia backend though?  That could explain why the setup is in 2 places.

gaaclarke: I'm not going to touch it since it could potentially mess up the skia backend and I think touching it would mean cleaning it up even further to unify the setup.

PR #46070, opened 2023-09-19
Implement JSObject instead of extending

srujzs:
JSObject will have a factory constructor to create an object literal, so you can't extend it as it will no longer have a generative constructor (@staticInterop types can't have generative constructors).

### DIFF
--- a/lib/web_ui/lib/src/engine/js_interop/js_loader.dart
+++ b/lib/web_ui/lib/src/engine/js_interop/js_loader.dart
@@ -73,7 +73,7 @@ abstract class FlutterEngineInitializer{
 @JS()
 @anonymous
 @staticInterop
-abstract class FlutterAppRunner extends JSObject {
+abstract class FlutterAppRunner implements JSObject {
   factory FlutterAppRunner({required RunAppFn runApp,}) => FlutterAppRunner._(
     runApp: (([RunAppFnParameters? args]) => futureToPromise(runApp(args))).toJS
   );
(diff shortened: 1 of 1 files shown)


eyebrowsoffire (approved):
LGTM!

mdebbar:
> JSObject will have a factory constructor to create an object literal

This would be really nice! We can finally replace `Object().toJSBox` with a proper `JSObject()` (in `canvaskit_api_test.dart` and `browser_detect_test.dart`).

srujzs:
> > JSObject will have a factory constructor to create an object literal
> 
> This would be really nice! We can finally replace `Object().toJSBox` with a proper `JSObject()` (in `canvaskit_api_test.dart` and `browser_detect_test.dart`).

Wow! I didn't really expect users to be using `toJSBox` for this. :) That API was more meant to export Dart objects instead of a way to create some arbitrary JS value. You can create an object literal a number of ways already  e.g. `js_util.newObject()` or using `@anonymous` factory constructors. Is there a reason these tests use `Object().toJSBox` instead e.g. brevity or avoiding another import? Nothing wrong with it, just seemed a bit strange.

eyebrowsoffire:
> > > JSObject will have a factory constructor to create an object literal
> > 
> > 
> > This would be really nice! We can finally replace `Object().toJSBox` with a proper `JSObject()` (in `canvaskit_api_test.dart` and `browser_detect_test.dart`).
> 
> Wow! I didn't really expect users to be using `toJSBox` for this. :) That API was more meant to export Dart objects instead of a way to create some arbitrary JS value. You can create an object literal a number of ways already e.g. `js_util.newObject()` or using `@anonymous` factory constructors. Is there a reason these tests use `Object().toJSBox` instead e.g. brevity or avoiding another import? Nothing wrong with it, just seemed a bit strange.

This was my change when the JS interop got more strict. Previously, we just passed `Object()` into a static interop declaration, which stopped working. So I just added a `toJSBox` to it to do the same thing we were doing before. It's just to set up a fake environment for a unit test, and the only thing that really mattered was that it was a non-falsey value. Seemed like the simplest way at the time, not requiring us to do any extra js interop declarations or import `js_util`.

PR #46072, opened 2023-09-19
Add description to assert in `history.dart`

kenzieschmoll:
I have hit this assertion error many times in tests and this message would be helpful for debugging.

### DIFF
--- a/lib/web_ui/lib/src/engine/navigation/history.dart
+++ b/lib/web_ui/lib/src/engine/navigation/history.dart
@@ -220,8 +220,8 @@ class MultiEntriesBrowserHistory extends BrowserHistory {
     assert(
       _hasSerialCount(currentState),
       currentState == null
-          ? 'unexpected null state'
-          : 'state is missing field \'serialCount\'',
+          ? 'unexpected null history state'
+          : "history state is missing field 'serialCount'",
     );
     final int backCount = _currentSerialCount;
     if (backCount > 0) {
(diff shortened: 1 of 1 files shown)


lib/web_ui/lib/src/engine/navigation/history.dart
```diff
@@ -217,7 +217,12 @@ class MultiEntriesBrowserHistory extends BrowserHistory {
     _isTornDown = true;
 
     // Restores the html browser history.
-    assert(_hasSerialCount(currentState));
+    assert(
+      _hasSerialCount(currentState),
+      currentState == null
+          ? 'unexpected null state'
+          : 'state is missing field \'serialCount\'',
```
mdebbar, line 224:
```suggestion
          ? 'unexpected null history state'
          : 'history state is missing field \'serialCount\'',
```

mdebbar (approved):
LGTM, thanks for the improvement!

Hixie:
test-exempt: assert message

...but i wouldn't shy away from testing this. I also wouldn't shy away from making the errors even more useful. You could give context (what is this history state error about?), you could give a solution (what should the developer do when they hit this?). See also https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo#error-messages-should-be-useful

kenzieschmoll:
>  You could give context (what is this history state error about?), you could give a solution (what should the developer do when they hit this?)

@mdebbar do you have suggestions on how we can make the error message more actionable? I have only hit this error in flutter web integration test environments, and even for the issue I was hitting, this error message didn't help me solve the problem. My problem was calling
```
html.window.history.replaceState(null, '', newUri.toString());
```
instead of 
```
html.window.history.replaceState(html.window.history.state, '', newUri.toString());
```
But I don't know whether this assert I hit is only applicable to my use case or to many potential causes.

PR #46073, opened 2023-09-19
[ios] fix asset url not found when loading app extension

cyanglaz:
In https://github.com/flutter/engine/commit/9446392af6a26a715c6db50a56b23d46c141d834, I refactored the assetsPath to use NSURL. It turns out that when the app bundle is not loaded (during launching app exgtension), the assetURL will return nil using the `URLForResource`, but the `pathForResource` successfully returns the raw path.

This PR reverts back to the raw path solution.

part of https://github.com/flutter/flutter/issues/124287

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/shell/platform/darwin/common/framework/Source/FlutterNSBundleUtils.mm
+++ b/shell/platform/darwin/common/framework/Source/FlutterNSBundleUtils.mm
@@ -59,6 +59,8 @@ NSString* FLTAssetPath(NSBundle* bundle) {
 
 NSString* FLTAssetsPathFromBundle(NSBundle* bundle) {
   NSString* flutterAssetsPath = FLTAssetPath(bundle);
+  // Use the raw path solution so that asset path can be returned from unloaded bundles.
+  // See https://github.com/flutter/engine/pull/46073
   NSString* assetsPath = [bundle pathForResource:flutterAssetsPath ofType:@""];
 
   if (assetsPath.length == 0) {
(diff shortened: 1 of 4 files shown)


shell/platform/darwin/common/framework/Source/FlutterNSBundleUtils.mm
```diff
@@ -54,15 +54,17 @@
...
+  return [bundle objectForInfoDictionaryKey:@"FLTAssetsPath"] ?: @"flutter_assets";
 }
 
-NSURL* FLTAssetsURLFromBundle(NSBundle* bundle) {
+NSString* FLTAssetsPathFromBundle(NSBundle* bundle) {
   NSString* flutterAssetsPath = FLTAssetPath(bundle);
-  NSURL* assets = [bundle URLForResource:flutterAssetsPath withExtension:nil];
+  // Use the raw path solution so that asset path can be returned from unloaded bundles.
+  // See https://github.com/flutter/engine/pull/46073
+  NSString* assetsPath = [bundle pathForResource:flutterAssetsPath ofType:@""];
```
stuartmorgan, line 64: Optional nit: Is there a reason you're not using `nil` instead of `@""` here and below, given that they do the same thing (according to the docs)? `nil` is more common as "no value" than an empty string.

stuartmorgan (approved):
LGTM

shell/platform/darwin/common/framework/Source/FlutterNSBundleUtils.mm
cyanglaz: I just reverted back to the original implementation, unsure why it was @"" instead of nil. I will change it to nil

PR #46084, opened 2023-09-20
[Impeller] Ensure that reused textures are cleared before getting sampled by backdrop textures

bdero:
Fix for https://github.com/flutter/flutter/issues/135053.

I'm going to try and optimize out this case later, since Wondrous is inadvertently paying for a backdrop filter that contributes nothing to the final image.

Wondrous with all fixes applied:

https://github.com/flutter/engine/assets/919017/37b49b77-b0ec-42d4-964a-c61b9a2cca8f

### DIFF
--- a/impeller/entity/entity_pass.cc
+++ b/impeller/entity/entity_pass.cc
@@ -537,6 +537,7 @@ EntityPass::EntityResult EntityPass::GetEntityForElement(
       // The subpass will need to read from the current pass texture when
       // rendering the backdrop, so if there's an active pass, end it prior to
       // rendering the subpass.
+      pass_context.GetRenderPass(pass_depth);
       pass_context.EndPass();
     }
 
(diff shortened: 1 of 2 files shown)


impeller/entity/entity_pass.cc
```diff
@@ -537,6 +537,7 @@ EntityPass::EntityResult EntityPass::GetEntityForElement(
       // The subpass will need to read from the current pass texture when
       // rendering the backdrop, so if there's an active pass, end it prior to
       // rendering the subpass.
+      pass_context.GetRenderPass(pass_depth);
```
jonahwilliams, line 540: I'm not sure I fully understand this case 🤔

bdero: If the very first thing we render in an EntityPass is a subpass that happens to have a backdrop filter, than that backdrop filter will end up rendering the raw, uncleared texture that came straight out of the texture cache. By calling `pass_context.GetRenderPass` here, we force at least one pass to get created with the correct config before that sampling happens, which will clear the texture.

Added this to a comment inline.

jonahwilliams: Oh, I see - so we're attaching the texture but we never created any rendering commands to record to it, so we never cleared the values. hmm. Slight flaw in the render target cache design

jonahwilliams (approved):
LGTM

PR #46085, opened 2023-09-20
[Impeller] Use BlackTransparent clear color when backdrop filters are present.

bdero:
Fix for https://github.com/flutter/flutter/issues/135053.

We already duck out of the Entity absorbing part of the optimization, but we also need to duck here when computing clear colors, otherwise we end up double-applying the effect of clearing entities at the beginning of a pass in some cases.

Wondrous with all fixes applied:

https://github.com/flutter/engine/assets/919017/37b49b77-b0ec-42d4-964a-c61b9a2cca8f


jonahwilliams (approved):
LGTM!

PR #46109, opened 2023-09-20
[web] - receive and set text editing scroll state, `line-height`, `letter-spacing`

htoor3:
Engine needs to receive scroll state from the framework and apply it to the text editing element.  Additionally, we need to set `line-height` and `letter-spacing` to better align the text input overlay.

Engine changes to support https://github.com/flutter/flutter/issues/135159

Framework changes are here: https://github.com/flutter/flutter/pull/135163

## Pre-launch Checklist

- [ ] I signed the [CLA].

### DIFF
--- a/lib/web_ui/lib/src/engine/text_editing/text_editing.dart
+++ b/lib/web_ui/lib/src/engine/text_editing/text_editing.dart
@@ -1048,7 +1048,7 @@ abstract class TextEditingStrategy {
 
   /// Set style to the native DOM element used for text editing.
   void updateElementStyle(EditableTextStyle style);
-  
+
   /// Set scroll state on the input or textarea element.
   void setScrollState(EditableTextScrollState scrollState);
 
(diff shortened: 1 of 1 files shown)


lib/web_ui/lib/src/engine/text_editing/text_editing.dart
```diff
@@ -2460,7 +2542,7 @@ class EditableTextGeometry {
   void applyToDomElement(DomHTMLElement domElement) {
     final String cssTransform = float64ListToCssTransform(globalTransform);
     domElement.style
-      ..width = '${width}px'
+      ..width = '${width - 2}px'
```
htoor3, line 2545: Our inputs don't take into account the scrollbar width, but the framework inputs do.   So we can sometimes end up running into an edge case where the framework text creates a line-break at the very end of a multiline input, but our inputs don't.  Slightly adjusting the width of our input to account for the scrollbar fixes this - but I'm wondering if there's a more reliable way to solve this than what I'm doing here. 

Would also need to cross-test this with different platform and browser combinations

PR #46117, opened 2023-09-20
Add link support in web accessibility

chunhtai:
fixes https://github.com/flutter/flutter/issues/134795

## Pre-launch Checklist

- [ ] I signed the [CLA].

### DIFF
--- a/lib/web_ui/lib/src/engine/semantics/link.dart
+++ b/lib/web_ui/lib/src/engine/semantics/link.dart
@@ -13,6 +13,7 @@ class Link extends PrimaryRoleManager {
   DomElement createElement() {
     final DomElement element = domDocument.createElement('a');
     // TODO(chunhtai): Fill in the real link once the framework sends entire uri.
+    // https://github.com/flutter/flutter/issues/102535.
     element.setAttribute('href', '#');
     element.style.display = 'block';
     return element;
--- a/lib/web_ui/lib/src/engine/semantics/link.dart
+++ b/lib/web_ui/lib/src/engine/semantics/link.dart
@@ -1,0 +1,20 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+import '../dom.dart';
+import '../semantics.dart';
+
+/// Provides accessibility for links.
+class Link extends PrimaryRoleManager {
+  Link(SemanticsObject semanticsObject) : super.withBasics(PrimaryRole.link, semanticsObject);
+
+  @override
+  DomElement createElement() {
+    final DomElement element = domDocument.createElement('a');
+    // TODO(chunhtai): Fill in the real link once the framework sends entire uri.
+    element.setAttribute('href', '#');
+    element.style.display = 'block';
+    return element;
+  }
+}
--- a/lib/web_ui/lib/src/engine/semantics/semantics.dart
+++ b/lib/web_ui/lib/src/engine/semantics/semantics.dart
@@ -441,7 +441,7 @@ abstract class PrimaryRoleManager {
   /// management intereferes with the widget's functionality.
   PrimaryRoleManager.blank(this.role, this.semanticsObject);
 
-  late final DomElement element = _initElement(createElement());
+  late final DomElement element = _initElement(createElement(), semanticsObject);
 
   /// The primary role identifier.
   final PrimaryRole role;
--- a/lib/web_ui/lib/src/engine/semantics/semantics.dart
+++ b/lib/web_ui/lib/src/engine/semantics/semantics.dart
@@ -24,6 +24,7 @@ import 'focusable.dart';
 import 'image.dart';
 import 'incrementable.dart';
 import 'label_and_value.dart';
+import 'link.dart';
 import 'live_region.dart';
 import 'platform_view.dart';
 import 'scrollable.dart';
@@ -382,6 +383,9 @@ enum PrimaryRole {
   ///
   /// Provides a label or a value.
   generic,
+
+  /// Contains a link.
+  link,
 }
 
 /// Identifies one of the secondary [RoleManager]s of a [PrimaryRoleManager].
@@ -437,6 +441,8 @@ abstract class PrimaryRoleManager {
   /// management intereferes with the widget's functionality.
   PrimaryRoleManager.blank(this.role, this.semanticsObject);
 
+  late final DomElement element = _createAndInitElement();
+
   /// The primary role identifier.
   final PrimaryRole role;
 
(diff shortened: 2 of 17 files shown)


lib/web_ui/lib/src/engine/semantics/link.dart
```diff
@@ -0,0 +1,105 @@
...
+/// Manages semantics objects that represents a link.
+///
+/// This role is implemented via a anchor HTML element.
+class Link extends PrimaryRoleManager {
+  Link(SemanticsObject semanticsObject) : super.blank(PrimaryRole.link, semanticsObject) {
+    _setupDomElement();
+  }
+
+  /// The element used for link, i.e. `<a>`.
+  late DomHTMLElement anchorElement;
```
chunhtai, line 20: There are still some more work needed to wire up the semantics action to this element, but I would like to know whether this is the way we want to pursuit.

An alternative is to Make semanticsObject.element to be a <a> directly, that way I can reuse all the RoleManager like Tappable or Focusable.

chunhtai (commented):
Hi @yjbanov can you take a look at this draft pr to see if this is the direction we want to go before I add more tests?

lib/web_ui/lib/src/engine/semantics/link.dart
```diff
@@ -0,0 +1,105 @@
...
+      //
+      // * [SemanticsObject.recomputePositionAndSize], which sets the position
+      //   and size of the parent `<flt-semantics>` element.
+      ..top = '0'
+      ..left = '0'
+      ..width = '${semanticsObject.rect!.width}px'
+      ..height = '${semanticsObject.rect!.height}px';
+    semanticsObject.element.append(anchorElement);
+    anchorElement.setAttribute('href', '#');
+    anchorElement.addEventListener('focus',
```
yjbanov, line 39: Can you reuse the `Focusable` role manager instead of implementing custom focus management?

lib/web_ui/lib/src/engine/semantics/link.dart
```diff
@@ -0,0 +1,105 @@
...
+    anchorElement.addEventListener('blur',
+        createDomEventListener((DomEvent event) {
+          if (semanticsObject.owner.gestureMode != GestureMode.browserGestures) {
+            return;
+          }
+
+          EnginePlatformDispatcher.instance.invokeOnSemanticsAction(
+              semanticsObject.id, ui.SemanticsAction.didLoseAccessibilityFocus, null);
+        }));
+    anchorElement.addEventListener('click',
```
yjbanov, line 57: Similarly here, can we reuse the `Tappable` role manager?

lib/web_ui/lib/src/engine/semantics/semantics.dart
```diff
@@ -1501,6 +1510,7 @@ class SemanticsObject {
       PrimaryRole.image => ImageRoleManager(this),
       PrimaryRole.platformView => PlatformViewRoleManager(this),
       PrimaryRole.generic => GenericRole(this),
+      PrimaryRole.link => Link(this),
```
yjbanov, line 1513: Let's keep the `generic` role last to underscore that it's the fallback role that we only use as a last resort.

lib/web_ui/lib/src/engine/semantics/link.dart
chunhtai: The focusable role manager will only work on SemanticsObject.element I think? If I want to do that, I will need to add a way to create an `<a>` as SemanticsObject.element instead of the `<flt-semantics-node> that it hardcoded to create. I am fine with that, but it seems to divert from what we were doing with text_field.dart

yjbanov: Ah, in that case you might want `AccessibilityFocusManager`.

chunhtai: I updated the code to use RoleManager instead, PTAL

chunhtai:
a friendly bump

lib/web_ui/lib/src/engine/semantics/semantics.dart
```diff
@@ -982,7 +958,51 @@ class SemanticsObject {
   final EngineSemanticsOwner owner;
 
   /// The DOM element used to convey semantics information to the browser.
-  final DomElement element = domDocument.createElement('flt-semantics');
+  late DomElement element;
+
+  void _updateElement(PrimaryRole role) {
+    switch (role) {
+      case PrimaryRole.link:
+        element = domDocument.createElement('a');
```
yjbanov, line 966: The `element` field is designed to be `final`, as it participates in the parent's DOM structure. Simply swapping the element field will result in a dangling element not attached to anywhere. Let's chat about what you are trying to do and find a better solution.

lib/web_ui/lib/src/engine/semantics/semantics.dart
```diff
@@ -1535,6 +1572,20 @@ class SemanticsObject {
       primaryRole = currentPrimaryRole;
       currentPrimaryRole.update();
     }
+
+    // Reparent element.
+    if (currentElement != element) {
+      final DomElement? container = _childContainerElement;
+      if (container != null) {
+        container.remove();
```
yjbanov, line 1580: You don't need to remove the container. The browser will do it automatically on the next line when you call `element.append`.

lib/web_ui/test/engine/semantics/semantics_test.dart
```diff
@@ -404,6 +412,45 @@ void _testEngineSemanticsOwner() {
...
+
+    tree = semantics().debugSemanticsTree!;
+    expect(tree.length, 2);
+    expect(tree[1]!.id, 1);
+    expect(tree[1]!.label, 'Hello');
+    expect(tree[1]!.element.tagName.toLowerCase(), 'a');
+    expectSemanticsTree('''
+<sem style="$rootSemanticStyle">
+  <sem-c>
+    <a aria-label="Hello" style="display: block;"></a>
```
yjbanov, line 446: Let's also test that the `role="..."` attribute was updated.

lib/web_ui/lib/src/engine/semantics/semantics.dart
```diff
@@ -437,6 +441,8 @@ abstract class PrimaryRoleManager {
   /// management intereferes with the widget's functionality.
   PrimaryRoleManager.blank(this.role, this.semanticsObject);
 
+  late final DomElement element = _createAndInitElement();
```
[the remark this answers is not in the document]
chunhtai: non-late final can only accept static function as its initialize. In this case, there isn't a way for sub class to inject their logic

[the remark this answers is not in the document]
yjbanov: Oh I see, you need to call `this.createElement` which not static. In that case `late` make sense. However, I'd still initialize it in the constructor. Otherwise, it gets initialized upon first read of the field, and since it's hard to predict when exactly that field is accessed we can get into race conditions, such as `_createAndInitElement` reading from `semanticsObject` that has changed its state after the constructor is finished. Can you move `element = _createAndInitElement();` into the constructor body?

lib/web_ui/lib/src/engine/semantics/link.dart
```diff
@@ -0,0 +1,20 @@
...
+import '../semantics.dart';
+
+/// Provides accessibility for links.
+class Link extends PrimaryRoleManager {
+  Link(SemanticsObject semanticsObject) : super.withBasics(PrimaryRole.link, semanticsObject);
+
+  @override
+  DomElement createElement() {
+    final DomElement element = domDocument.createElement('a');
+    // TODO(chunhtai): Fill in the real link once the framework sends entire uri.
```
yjbanov, line 15: Is there a Github issue this can point to?

lib/web_ui/lib/src/engine/semantics/semantics.dart
```diff
@@ -453,29 +459,83 @@ abstract class PrimaryRoleManager {
   @visibleForTesting
   List<Role> get debugSecondaryRoles => _secondaryRoleManagers?.map((RoleManager manager) => manager.role).toList() ?? const <Role>[];
 
+  @protected
+  DomElement createElement() => domDocument.createElement('flt-semantics');
+
+  DomElement _createAndInitElement() {
```
yjbanov, line 465: nit: actually, I'd still keep this `static` and pass all the information it needs to initialize the element (I think all it needs is the element created by `createElement` and `semanticsObject`). This way this method will not need access to `this`.

yjbanov (approved):
lgtm

PR #46122, opened 2023-09-20
Clang tidy-ify `DlColor` and friends.

matanlurey:
Closes https://github.com/flutter/flutter/issues/135057.

This is a fair bit more involved than previous changes, just due to the sheer number of implicit conversions.

Highlights:

- Made `public uint32_t argb` `private uint32_t argb_`, and added `argb()` instead.
- Added `ToSk(DlColor)` instead of using implicit conversions.

There were a bunch of places where I had to make a judgement call (particularly in tests) to keep the code a bit "messy", i.e. `DlColor(SK_RED)`, just to make the diff as small as possible and to prevent silly copy and paste bugs. I'd be open to filing a follow-up issue to reduce unnecessary wrapping.

### DIFF
--- a/display_list/benchmarking/dl_complexity_unittests.cc
+++ b/display_list/benchmarking/dl_complexity_unittests.cc
@@ -190,7 +190,7 @@ TEST(DisplayListComplexity, DrawShadow) {
   line_path.moveTo(SkPoint::Make(0, 0));
   line_path.lineTo(SkPoint::Make(10, 10));
   line_path.close();
-  builder_line.DrawShadow(line_path, SK_ColorRED, 10.0f, false, 1.0f);
+  builder_line.DrawShadow(line_path, DlColor(SK_ColorRED), 10.0f, false, 1.0f);
   auto display_list_line = builder_line.Build();
 
   DisplayListBuilder builder_quad;
@@ -198,7 +198,7 @@ TEST(DisplayListComplexity, DrawShadow) {
   quad_path.moveTo(SkPoint::Make(0, 0));
   quad_path.quadTo(SkPoint::Make(10, 10), SkPoint::Make(10, 20));
   quad_path.close();
-  builder_quad.DrawShadow(quad_path, SK_ColorRED, 10.0f, false, 1.0f);
+  builder_quad.DrawShadow(quad_path, DlColor(SK_ColorRED), 10.0f, false, 1.0f);
   auto display_list_quad = builder_quad.Build();
 
   DisplayListBuilder builder_conic;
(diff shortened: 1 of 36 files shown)


chinmaygarde (approved):
cc @flar

flar:
Ugh, auto conversion from uint32_t was a feature, not a bug... :(

display_list/benchmarking/dl_complexity_unittests.cc
```diff
@@ -190,31 +190,33 @@ TEST(DisplayListComplexity, DrawShadow) {
...
+  builder_line.DrawShadow(line_path, DlColor(SK_ColorRED), 10.0f, false, 1.0f);
   auto display_list_line = builder_line.Build();
 
   DisplayListBuilder builder_quad;
   SkPath quad_path;
   quad_path.moveTo(SkPoint::Make(0, 0));
   quad_path.quadTo(SkPoint::Make(10, 10), SkPoint::Make(10, 20));
   quad_path.close();
-  builder_quad.DrawShadow(quad_path, SK_ColorRED, 10.0f, false, 1.0f);
+  builder_quad.DrawShadow(quad_path, DlColor(SK_ColorRED), 10.0f, false, 1.0f);
```
flar, line 201: Meanwhile, you completely ignored all of the Dl color constants so this entire PR will have to be duplicated to use the actual Dl values that should be used here.

matanlurey: Yes, that was documented in the PR discussion as an intentional trade-off/choice. I'm open to fixing it, but didn't want that as a pre-req for enabling header checks.

flar: Oh, are you saying that you wanted to avoid `SK_ColorRed` => `DlColor::kBlue()` types of typos?

The easiest path to enabling header checks here would have been to exempt this one constructor from the explicit rule (since it was intentionally created as a converting constructor) and to keep the uint32_t converter in the class. Both of these were intentional since the conversion was pretty straightforward. The class, when you look at how it compiles is literally just a uint32_t being passed around - no allocations, just pass along a 32-bit value. While implicit conversions in general are a problem area, this was one case where it was considered appropriate.

flar: Actually, I need to double check if burying the field behind an accessor has kept the "just pass around a 32-bit value for efficiency" design of this class.

matanlurey: > Oh, are you saying that you wanted to avoid `SK_ColorRed` => `DlColor::kBlue()` types of typos?

Right, yes. I did about 20 of these PRs, so my goal was "make the code 1:1 where possible".

> The class, when you look at how it compiles is literally just a uint32_t being passed around

Fair. I don't feel strongly either way, but even looking at the codebase we sometimes were reading `.rgba` and sometimes didn't (implicit uint32_t conversion), so it appeared easy enough to just be consistent everywhere and explicitly read `.rgba()` (in practice it was a handful of changes, if it was 100s I might have decided otherwise).

PR #46127, opened 2023-09-20
[Impeller] removed global mutable variable for tessellation allocation function pointers

gaaclarke:
issue: https://github.com/flutter/flutter/issues/125749

This removes a global mutable variable, instead giving each tesellator its own struct.  I suspect the root cause of the linked  issue is memory corruption but this is good cleanup anyways.  If for some reason libTess2 is mutating that global variable at cleanup, that could be the source of this problem and this would fix the issue.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/impeller/tessellator/tessellator.cc
+++ b/impeller/tessellator/tessellator.cc
@@ -31,8 +31,7 @@ static const TESSalloc kAlloc = {
     0                                    /* =extraVertices */
 };
 
-Tessellator::Tessellator()
-    : c_tessellator_(nullptr, &DestroyTessellator) {
+Tessellator::Tessellator() : c_tessellator_(nullptr, &DestroyTessellator) {
   TESSalloc alloc = kAlloc;
   {
     // libTess2 copies the TESSalloc despite the non-const argument.
--- a/impeller/tessellator/tessellator.cc
+++ b/impeller/tessellator/tessellator.cc
@@ -21,7 +21,7 @@ static void HeapFree(void* userData, void* ptr) {
 }
 
 // Note: these units are "number of entities" for bucket size and not in KB.
-static TESSalloc alloc = {
+static const TESSalloc kAlloc = {
     HeapAlloc, HeapRealloc, HeapFree, 0, /* =userData */
     16,                                  /* =meshEdgeBucketSize */
     16,                                  /* =meshVertexBucketSize */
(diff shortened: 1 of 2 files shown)


jonahwilliams (approved):
LGTM

makes sense to me!

impeller/tessellator/tessellator.cc
```diff
@@ -32,7 +32,19 @@ static TESSalloc alloc = {
...
-    : c_tessellator_(::tessNewTess(&alloc), &DestroyTessellator) {}
+    : c_tess_alloc_(nullptr, &DestroyTessAlloc),
+      c_tessellator_(nullptr, &DestroyTessellator) {
+  {
+    CTessAlloc tess_alloc(new TESSalloc(), &DestroyTessAlloc);
+    c_tess_alloc_ = std::move(tess_alloc);
+  }
+  memcpy(c_tess_alloc_.get(), &kAlloc, sizeof(TESSalloc));
+  {
+    CTessellator tessellator(::tessNewTess(c_tess_alloc_.get()),
```
jason-simmons, line 43: `tessNewTess` copies the `TESSalloc` contents into the newly created tessellator and does not mutate the `alloc` argument:
https://flutter.googlesource.com/third_party/libtess2/+/refs/heads/master/Source/tess.h#84
https://flutter.googlesource.com/third_party/libtess2/+/refs/heads/master/Source/tess.c#610

If you want to clarify the lifetime of this structure, then it should be sufficient to declare the `TESSalloc` as a local variable here.

jason-simmons (changes requested):

impeller/tessellator/tessellator.cc
gaaclarke: Done.  The code as it was written is conforming to the api and docstrings since it takes a mutable pointer and has no mention of what it does to it.  libTess2 will never be updated so it should be safe to pass a pointer to something on the stack.

impeller/tessellator/tessellator.cc
```diff
@@ -31,8 +31,14 @@ static TESSalloc alloc = {
     0                                    /* =extraVertices */
 };
 
-Tessellator::Tessellator()
-    : c_tessellator_(::tessNewTess(&alloc), &DestroyTessellator) {}
+Tessellator::Tessellator() : c_tessellator_(nullptr, &DestroyTessellator) {
+  TESSalloc alloc = kAlloc;
+  {
```
jason-simmons, line 36: This nested scope can be removed

jason-simmons (approved):

impeller/tessellator/tessellator.cc
gaaclarke: I used those so someone doesn't accidentally use `tessellator` after the move.  That's not a compilation error, but is a clang tidy error.

PR #46130, opened 2023-09-21
[Impeller] Incorporate backdrop filters in subpass coverage.

bdero:
Resolves https://github.com/flutter/flutter/issues/134775.

Before:

image

After:

image

### DIFF
--- a/impeller/entity/entity_pass.cc
+++ b/impeller/entity/entity_pass.cc
@@ -124,10 +124,40 @@ std::optional<Rect> EntityPass::GetElementsCoverage(
 
       std::optional<Rect> unfiltered_coverage =
           GetSubpassCoverage(subpass, std::nullopt);
+
+      // If the current pass elements have any coverage so far and there's a
+      // backdrop filter, then incorporate the backdrop filter in the
+      // pre-filtered coverage of the subpass.
+      if (result.has_value() && subpass.backdrop_filter_proc_) {
+        std::shared_ptr<FilterContents> backdrop_filter =
+            subpass.backdrop_filter_proc_(FilterInput::Make(result.value()),
+                                          subpass.xformation_,
+                                          Entity::RenderingMode::kSubpass);
+        if (backdrop_filter) {
+          auto backdrop_coverage = backdrop_filter->GetCoverage({});
+          backdrop_coverage->origin += result->origin;
+          if (backdrop_coverage.has_value()) {
+            if (unfiltered_coverage.has_value()) {
+              unfiltered_coverage = coverage->Union(*backdrop_coverage);
+            } else {
+              unfiltered_coverage = backdrop_coverage;
+            }
+          }
+        }
+      }
+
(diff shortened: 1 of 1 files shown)


impeller/entity/entity_pass.cc
```diff
@@ -124,10 +124,40 @@ std::optional<Rect> EntityPass::GetElementsCoverage(
...
+
+      // If the current pass elements have any coverage so far and there's a
+      // backdrop filter, then incorporate the backdrop filter in the
+      // pre-filtered coverage of the subpass.
+      if (result.has_value() && subpass.backdrop_filter_proc_) {
+        std::shared_ptr<FilterContents> backdrop_filter =
+            subpass.backdrop_filter_proc_(FilterInput::Make(result.value()),
+                                          subpass.xformation_,
+                                          Entity::RenderingMode::kSubpass);
+        if (backdrop_filter) {
```
jonahwilliams, line 136: is subpass.backdrop_filter_proc_ and no backdrop filter response an error state? Should we add a validation log for that?

jonahwilliams (approved):
LGTM with question

impeller/entity/entity_pass.cc
bdero: Was thinking maybe we should gracefully support nullptr results for filters that do nothing, but on second thought I think that's confusing and filters should internally elide instead (as they already do)... Added a log.

impeller/entity/entity_pass.cc
```diff
@@ -124,10 +124,43 @@ std::optional<Rect> EntityPass::GetElementsCoverage(
...
+          auto backdrop_coverage = backdrop_filter->GetCoverage({});
+          backdrop_coverage->origin += result->origin;
```
zanderso, line 138: Should this line be under the `if` on the line below?

impeller/entity/entity_pass.cc
```diff
@@ -124,10 +124,43 @@ std::optional<Rect> EntityPass::GetElementsCoverage(
...
+          if (backdrop_coverage.has_value()) {
+            if (unfiltered_coverage.has_value()) {
+              unfiltered_coverage = coverage->Union(*backdrop_coverage);
```
zanderso, line 141: It looks like `coverage` could be undefined here.

zanderso (commented):
Trying to enable clang-tidy in this file, and hit a couple of flags: https://github.com/flutter/engine/pull/47316

impeller/entity/entity_pass.cc
bdero: yeah this could be placed just above `unfiltered_coverage = backdrop_coverage`

impeller/entity/entity_pass.cc
bdero: Ugh, yeah, if `coverage` is nullopt we should just `unfiltered_coverage = backdrop_coverage`

PR #46131, opened 2023-09-21
[Impeller] temp work around for cmd pool validation issues.

jonahwilliams:
I'm not able to reproduce most validation errors in the macrobenchmark app with this change. THe real fix seems to be to track this last cmd buffer but I'm struggling to find a way to do this.


matanlurey (approved):
Let's land this and see how the performance is impacted. It's not worse than what we have right now.

matanlurey:
@jonahwilliams Should this mark https://github.com/flutter/flutter/issues/135086 fixed?

PR #46167, opened 2023-09-21
[Impeller] fixes behavior for blurred rounded rect clear

gaaclarke:
fixes: https://github.com/flutter/flutter/issues/132849

I decided to mimic skia's blend function for clear only in the case for blurred rounded rects since that fixes the listed issue.  I'm not sure if there are are other cases where we'd want this behavior.  I've added golden tests for the other clear cases too to make sure we didn't break them.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/impeller/aiks/aiks_unittests.cc
+++ b/impeller/aiks/aiks_unittests.cc
@@ -3581,5 +3581,24 @@ TEST_P(AiksTest, VerticesGeometryUVPositionData) {
   ASSERT_TRUE(OpenPlaygroundHere(canvas.EndRecordingAsPicture()));
 }
 
+TEST_P(AiksTest, ClearBlendWithBlur) {
+  Canvas canvas;
+  Paint white;
+  white.color = Color::White();
+  canvas.DrawRect(Rect::MakeXYWH(0, 0, 600.0, 600.0), white);
+
+  Paint clear;
+  clear.color = Color::White();
+  clear.blend_mode = BlendMode::kClear;
+  clear.mask_blur_descriptor = Paint::MaskBlurDescriptor{
+      .style = FilterContents::BlurStyle::kNormal,
+      .sigma = Sigma(20),
+  };
+
+  canvas.DrawCircle(Point::MakeXY(300.0, 300.0), 200.0, clear);
+
+  ASSERT_TRUE(OpenPlaygroundHere(canvas.EndRecordingAsPicture()));
+}
+
 }  // namespace testing
 }  // namespace impeller
--- a/impeller/aiks/aiks_unittests.cc
+++ b/impeller/aiks/aiks_unittests.cc
@@ -3613,5 +3613,17 @@ TEST_P(AiksTest, ClearBlend) {
   ASSERT_TRUE(OpenPlaygroundHere(canvas.EndRecordingAsPicture()));
 }
 
+TEST_P(AiksTest, ClearBlendImage) {
+  Canvas canvas;
+  Paint paint;
+  auto dst = std::make_shared<Image>(CreateTextureForFixture("blend_mode_dst.png"));
+  auto src = std::make_shared<Image>(CreateTextureForFixture("blend_mode_src.png"));
+  canvas.DrawImage(dst, Point::MakeXY(100.0, 100.0), paint);
+  Paint clear;
+  clear.blend_mode = BlendMode::kClear;
+  canvas.DrawImage(src, Point::MakeXY(100.0, 100.0), clear);
+  ASSERT_TRUE(OpenPlaygroundHere(canvas.EndRecordingAsPicture()));
+}
+
 }  // namespace testing
 }  // namespace impeller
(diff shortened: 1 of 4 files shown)


impeller/aiks/aiks_unittests.cc
```diff
@@ -3581,5 +3581,51 @@ TEST_P(AiksTest, VerticesGeometryUVPositionData) {
...
+
+  Paint clear;
+  clear.blend_mode = BlendMode::kClear;
+
+  canvas.DrawCircle(Point::MakeXY(300.0, 300.0), 200.0, clear);
+
+  ASSERT_TRUE(OpenPlaygroundHere(canvas.EndRecordingAsPicture()));
+}
+
+TEST_P(AiksTest, ClearBlendImage) {
```
bdero, line 3616: Ah, there's actually already a golden for this, but it's kind of sneaky because the correct behavior is for it to render nothing, unlike all the other blend variants. :) `AiksTest.BlendModeClear`

It looks like both the existing golden and this new golden aren't quite rendering correctly with this change: https://api.flutter.dev/flutter/dart-ui/BlendMode.html#clear

impeller/entity/contents/content_context.cc
```diff
@@ -38,10 +38,19 @@ void ContentContextOptions::ApplyToPipelineDescriptor(
...
-      color0.dst_color_blend_factor = BlendFactor::kZero;
-      color0.src_alpha_blend_factor = BlendFactor::kZero;
-      color0.src_color_blend_factor = BlendFactor::kZero;
+      if (is_for_rrect_blur_clear) {
+        color0.alpha_blend_op = BlendOperation::kReverseSubtract;
+        color0.color_blend_op = BlendOperation::kReverseSubtract;
+        color0.dst_alpha_blend_factor = BlendFactor::kOne;
+        color0.dst_color_blend_factor = BlendFactor::kOne;
+        color0.src_alpha_blend_factor = BlendFactor::kDestinationColor;
+        color0.src_color_blend_factor = BlendFactor::kDestinationColor;
```
bdero, line 47: I think this will cause the clear to weaken as the paint color becomes transparent. Is that behavior correct?

impeller/aiks/aiks_unittests.cc
gaaclarke: Oh wow, the golden results are surprising, there is a lot to unpack:
- The new clear test `impeller_Play_AiksTest_ClearBlend_*` renders nothing correctly
- `impeller_Play_AiksTest_ClearBlendWithBlur_Vulkan` gives me different results locally with moltenvk
- The existing clear test `impeller_Play_AiksTest_BlendModeClear_Metal` rendered nothing incorrectly, but this PR seems to have broken that (I would have expected my change to only effect rrect blur). This is also surprising considering `impeller_Play_AiksTest_ClearBlend_*`
- `impeller_Play_AiksTest_ClearBlendImage_Vulkan` is also giving me different results locally

I'll see if I can get this sorted out.

## local impeller_Play_AiksTest_ClearBlendImage_Vulkan

Screenshot 2023-09-26 at 8 50 25 AM

## local impeller_Play_AiksTest_ClearBlendWithBlur_Vulkan

Screenshot 2023-09-26 at 8 50 09 AM

impeller/entity/contents/content_context.cc
gaaclarke: That's right, thats how you get the blurred results (that work locally):
Screenshot 2023-09-26 at 8 50 09 AM

impeller/aiks/aiks_unittests.cc
gaaclarke: I had been working with the vulkan tests, just verified that locally the metal tests give different/correct results.  I wonder if skiagold got messed up.  It is reporting https://github.com/flutter/engine/commit/98cec738eb342d39ad9a99fd7441bd9cec596f96 though.

gaaclarke: Filed an issue about the discrepancy between skiagold and local results: https://github.com/flutter/flutter/issues/135507

I thought maybe it was a moltenvk thing, but metal and vulkan are giving the same correct results locally which differ from skiagolds results.

gaaclarke: I removed `ClearBlendImage` since it's a duplicate.  We'll also see if pushing a new commit flushes out the skiagold issue.

gaaclarke:
Okay, pushing a commit flushed out the surprising results.  Skiagold did in fact get confused by potentially a rebase.  Now the remaining tests are broken because the the skiagold tests, the clear color is white instead of black.  We should fix that.  I'll file an issue.  In the meantime I'll switch the square in the tests from white to a different color so the results can be seen.

gaaclarke:
@bdero please take another look.  All the failures in skiagold were erroneous.

gaaclarke:
(Filed the issue about the clear color in the skiagold tests: https://github.com/flutter/flutter/issues/135511)

impeller/aiks/aiks_unittests.cc
bdero: Huh, weird. Yeah I couldn't spot anything that should cause the existing golden to fail yesterday. Thanks for digging in!

impeller/entity/contents/content_context.cc
bdero: Tried this out with wacky transparent colors in your repro and this seems to work great!

bdero (approved):
LGTM!

PR #46170, opened 2023-09-21
Make `./tools/clang_tidy/bin/main.dart` executable.

matanlurey:
LMK if there is another way you'd prefer this done.

I considered adding `./tools/clang_tidy` (shell script), but wasn't sure.

### DIFF
--- a/tools/clang_tidy/README.md
+++ b/tools/clang_tidy/README.md
@@ -4,7 +4,7 @@ A wrapper library and program that runs `clang_tidy` on the Flutter engine repo.
 
 ```shell
 # Assuming you are in the `flutter` root of the engine repo.
-dart ./tools/clang_tidy/bin/main.dart
+./tools/clang_tidy/bin/main.dart
 ```
 
 By default, the linter runs over _modified_[^1] files in the _latest_[^2] build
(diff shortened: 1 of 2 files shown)


tools/clang_tidy/README.md
```diff
@@ -4,7 +4,7 @@ A wrapper library and program that runs `clang_tidy` on the Flutter engine repo.
 
 ```shell
 # Assuming you are in the `flutter` root of the engine repo.
-dart ./tools/clang_tidy/bin/main.dart
+./tools/clang_tidy/bin/main.dart
```
zanderso, line 7: Maybe instead of adding a new script, or adding a shebang to main.dart, we could just update these docs to point at https://github.com/flutter/engine/blob/main/ci/clang_tidy.sh?

```suggestion
./ci/clang_tidy.sh
```

matanlurey: Yeah that's an option.

I don't love the discoverability of "the way to easily run clang_tidy is to run it from the CI folder", and it also has some weirdness, for example you won't be able to pass `--src-dir` because it's already provided by the CI script.

Maybe nothing is necessary tbh and I'll just revert this PR :)

zanderso: > Yeah that's an option.
> 
> I don't love the discoverability of "the way to easily run clang_tidy is to run it from the CI folder", and it also has some weirdness, for example you won't be able to pass `--src-dir` because it's already provided by the CI script.

Yeah, agreed it's not great. My thinking is that It could be an improvement to update the docs to reflect the current state of the tree, and then think about improving the discoverability separately.

PR #46171, opened 2023-09-21
[Impeller] Disable layer tree screenshots.

jonahwilliams:
The "rasterizer screenshot" VM service API uses the Skia software backend, even when running with impeller. At best, this leads to confusing half working screenshots that are missing images.

See https://github.com/flutter/flutter/issues/135052

We should disable this functionality when running with impeller. Perhaps this could be reworked to use picture toImage instead of the offscreen so that it uses the "real" backend regardless of the platform it is running on.

Fixes https://github.com/flutter/flutter/issues/135052


jonahwilliams:
Ahh, I can't actually turn on impeller in these tests 👿

jonahwilliams:
This is waiting on changes to allow testing.

chinmaygarde (approved):

PR #45310, opened 2023-08-31
[Windows] Update vsync on raster thread

loic-sharma:
## Background

If the Windows system compositor is enabled, the Windows embedder disables the swap interval so that presenting to a surface does not block until the v-blank. If the Windows system compositor is disabled (which is possible on Windows 7), the Windows embedder enables swap interval to prevent screen tearing.

Updating the swap interval requires making the GL surface current, which we currently do on the platform thread. However, the raster thread also needs the GL surface for rendering. Our current version of ANGLE allows making a GL surface current on multiple threads. However, the latest version of ANGLE errors if a GL context is made current on multiple threads. This is causing the ANGLE roll to fail ([example](https://ci.chromium.org/ui/p/flutter/builders/try/Windows%20Engine%20Drone/203788/overview)).

## Solution

There's two fixes:

1. The GL context is released once the swap interval is updated. This allows the platform thread to set the initial swap interval at start up, before the raster thread is started. This fixes the ANGLE roll.
3. When the system compositor changes, the Windows embedder now switches to the raster thread before updating the swap interval. This ensures that the GL surface is only made current on the raster thread once the raster thread has started. I updated unit tests to cover this scenario and tested this manually on a Windows 7 machine.

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [Flutter Style Guide] and the [C++, Objective-C, Java style guides].
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I added new tests to check the change I am making or feature I am adding, or Hixie said the PR is test-exempt. See [testing the engine] for instructions on writing and running engine tests.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I signed the [CLA].
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/flutter/wiki/Tree-hygiene#overview
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[C++, Objective-C, Java style guides]: https://github.com/flutter/engine/blob/main/CONTRIBUTING.md#style
[testing the engine]: https://github.com/flutter/flutter/wiki/Testing-the-engine
[...]

PR #46172, opened 2023-09-21
[Windows] Improve logic to update swap intervals

loic-sharma:
This relands https://github.com/flutter/engine/pull/45310 to unblock the ANGLE roll with the fix for https://github.com/flutter/flutter/issues/134262.

### Swap interval

If the Windows system compositor is enabled, the Windows embedder disables the swap interval so that presenting to a surface does not block until the v-blank. If the Windows system compositor is disabled (which is possible on Windows 7), the Windows embedder enables swap interval to prevent screen tearing.

### GL context threading

Our current version of ANGLE allows making a GL context current on multiple threads. However, the latest version of ANGLE errors if a GL context is made current on multiple threads. This is causing the ANGLE roll to fail ([example](https://ci.chromium.org/ui/p/flutter/builders/try/Windows%20Engine%20Drone/203788/overview)).

The Windows embedder has two GL context threading issues:

1. At startup, the platform thread creates and binds the GL context. This change ensures the GL context is released from the platform thread so that the raster thread can use the GL context for rendering.
2. When the system compositor updates, the GL context is bound to the platform thread to update the swap interval. This change ensures the swap interval update happens on the raster thread.

### Window resizing

Resizing the window recreates the GL surface and resets the swap interval.

The previous fix released the current GL context after updating the swap interval (this ensured the platform thread released the GL context at startup). This broke window resizing as it caused the engine to "lose" its GL context during rendering (see https://github.com/flutter/flutter/issues/134262). This reland releases the GL context only if on the startup case.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/shell/platform/windows/flutter_windows_view.cc
+++ b/shell/platform/windows/flutter_windows_view.cc
@@ -594,6 +594,12 @@ void FlutterWindowsView::CreateRenderSurface() {
     engine_->surface_manager()->CreateSurface(GetRenderTarget(), bounds.width,
                                               bounds.height, enable_vsync);
 
+    // The EGL context cannot be current on multiple threads.
+    // Creating the render surface runs on the platform thread and
+    // makes the EGL context current. Thus, the EGL context must be
+    // released so that the raster thread can use it for rendering.
+    engine_->surface_manager()->ClearContext();
+
     resize_target_width_ = bounds.width;
     resize_target_height_ = bounds.height;
   }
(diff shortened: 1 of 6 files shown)


shell/platform/windows/flutter_windows_view.cc
```diff
@@ -594,6 +594,12 @@ void FlutterWindowsView::CreateRenderSurface() {
     engine_->surface_manager()->CreateSurface(GetRenderTarget(), bounds.width,
                                               bounds.height, enable_vsync);
 
+    // The EGL context cannot be current on multiple threads.
+    // Creating the render surface runs on the platform thread and
+    // makes the EGL context current. Thus, the EGL context must be
+    // released so that the raster thread can use it for rendering.
+    engine_->surface_manager()->ClearContext();
```
loic-sharma, line 601: This is the fix.

Previously the GL context was released in `AngleSurfaceManager::SetVSyncEnabled`, which ensured the platform thread released the GL context at startup... but it also released the GL context in the middle of rendering if the window was resized.

Now, the GL context is only released on the startup case when the surface is created.

yaakovschectman: So this was the cause of the window resizing bug, and that should now be fixed, right?

yaakovschectman (approved):
LGTM

shell/platform/windows/flutter_windows_view.cc
loic-sharma: Yup

shell/platform/windows/flutter_windows_view_unittests.cc
```diff
@@ -1287,6 +1287,13 @@ TEST(FlutterWindowsViewTest, UpdatesVSyncOnDwmUpdates) {
   std::unique_ptr<MockAngleSurfaceManager> surface_manager =
       std::make_unique<MockAngleSurfaceManager>();
```
cbracken, line 1289: Is there a straightforward way to verify the release on the platform thread after surface creation, so that no one reverts that bit?

cbracken (approved):
lgtm with one suggestion -- thanks for getting to the bottom of this!

shell/platform/windows/flutter_windows_view_unittests.cc
loic-sharma: Added!

PR #46213, opened 2023-09-22
Multi-view pointer event

dkwingsmt:
This PR adds a new field `view_id` to embedder API's `FlutterPointerEvent`, allowing platforms to specify the source view of pointer events.

https://github.com/flutter/flutter/issues/112205

## Pre-launch Checklist

- [ ] I signed the [CLA].


loic-sharma (approved):
Please don't merge this yet as the embedder API isn't approved yet: [flutter.dev/go/multi-view-embedder-apis](https://flutter.dev/go/multi-view-embedder-apis)

(Feel free to reset my vote later when the design does receive approval)

loic-sharma (changes requested):
Please don't merge this yet as the embedder API isn't approved yet: [flutter.dev/go/multi-view-embedder-apis](https://flutter.dev/go/multi-view-embedder-apis)

(Feel free to reset my vote later when the design does receive approval)

lib/ui/window/pointer_data_packet_converter_unittests.cc
```diff
@@ -129,6 +131,7 @@ void CreateSimulatedTrackpadGestureData(PointerData& data,  // NOLINT
   data.pan_delta_y = 0.0;
   data.scale = scale;
   data.rotation = rotation;
+  data.view_id = kImplicitViewId;
```
loic-sharma, line 134: Nit: Consider using `0` directly instead. Ideally the engine shouldn't make any assumptions about the ID of the implicit view.

shell/platform/android/io/flutter/embedding/android/AndroidTouchProcessor.java
```diff
@@ -253,6 +258,10 @@ private void addPointerForIndex(
     if (pointerChange == -1) {
       return;
     }
+    // TODO(dkwingsmt): Use the correct source view ID once Android supports
+    // multiple views.
+    // https://github.com/flutter/flutter/issues/138167
```
loic-sharma, line 263: Looks like this should be: https://github.com/flutter/flutter/issues/134405

shell/platform/windows/flutter_windows_view.cc
```diff
@@ -332,6 +340,7 @@ void FlutterWindowsView::SendPointerMove(double x,
   FlutterPointerEvent event = {};
   event.x = x;
   event.y = y;
+  event.view_id = _viewIdForPointerEvent();
```
loic-sharma, line 343: Instead of setting `view_id` in all pointer methods, please set it in the common `SendPointerEventWithData` helper.

shell/platform/embedder/embedder.h
```diff
@@ -961,6 +961,8 @@ typedef struct {
   double scale;
   /// The rotation of the pan/zoom in radians, where 0.0 is the initial angle.
   double rotation;
+  /// The identifier of the view that received the pointer event.
+  int64_t view_id;
```
loic-sharma, line 965: Please introduce a type for the view identifiers. See this section in the design: https://docs.google.com/document/d/1kegGJ235y1778-vzyyv90MgSHmYQcROTupMuLtEDBq0/edit#heading=h.u7to6t120q4

dkwingsmt:
@loic-sharma @chinmaygarde @cbracken This PR should be ready to review. Can you take a look?

shell/platform/embedder/embedder.h
```diff
@@ -266,6 +266,15 @@ typedef enum {
 
 typedef struct _FlutterEngine* FLUTTER_API_SYMBOL(FlutterEngine);
 
+/// Identifier for views.
+///
+/// How view IDs are generated is decided by the embedding. The engine does not
+/// assume any specifics about the generation algorithm, including whether it
+/// utilizes the entire range of int64_t or permits the use of zero or negative
+/// IDs. The only requirement is that at any given moment, no two views share
+/// the same ID.
```
loic-sharma, line 275: What do you think of something like this?

```suggestion
/// Unique identifier for views.
///
///
/// View IDs are generated by the embedder and are
/// opaque to the engine. The engine does not interpret view IDs in any way.
```

shell/platform/embedder/embedder.cc
```diff
@@ -2326,6 +2326,7 @@ FlutterEngineResult FlutterEngineSendPointerEvent(
     pointer_data.pan_delta_y = 0.0;
     pointer_data.scale = SAFE_ACCESS(current, scale, 0.0);
     pointer_data.rotation = SAFE_ACCESS(current, rotation, 0.0);
+    pointer_data.view_id = SAFE_ACCESS(current, view_id, 0);
```
loic-sharma, line 2329: Can we use `kFlutterImplicitViewId` here? (FYI in the future the embedder will control the implicit view ID, so this will need to updated later)

```suggestion
    pointer_data.view_id = SAFE_ACCESS(current, view_id, kFlutterImplicitViewId);
```

shell/platform/darwin/ios/framework/Source/FlutterViewController.mm
```diff
@@ -58,6 +59,11 @@
 // change. Unfortunately unless you have Werror turned on, incompatible pointers as arguments are
 // just a warning.
 @interface FlutterViewController () <FlutterBinaryMessenger, UIScrollViewDelegate>
+// TODO(dkwingsmt): Make the view ID property public once the iOS shell
+// supports multiple views.
+// https://github.com/flutter/flutter/issues/138168
+@property(nonatomic, assign) int64_t viewId;
```
loic-sharma, line 65: [The runner API design](https://docs.google.com/document/d/1WkJuzifgJXyceaSoCCu8qTETYmYBOYKoz7Qaxtsore4/edit?tab=t.0#heading=h.jony3helpbsu) had the following property, which was Stuart's recommendation:

```suggestion
@property(nonatomic, readonly) int64_t viewIdentifier;
```

lib/ui/window/pointer_data_packet_converter_unittests.cc
```diff
@@ -129,6 +129,7 @@ void CreateSimulatedTrackpadGestureData(PointerData& data,  // NOLINT
   data.pan_delta_y = 0.0;
   data.scale = scale;
   data.rotation = rotation;
+  data.view_id = 0;
```
loic-sharma, line 132: Could we test the view ID packing/unpacking too?

shell/platform/android/io/flutter/embedding/android/AndroidTouchProcessor.java
```diff
@@ -80,8 +80,8 @@ public class AndroidTouchProcessor {
     int UNKNOWN = 4;
   }
 
-  // Must match the unpacking code in hooks.dart.
-  private static final int POINTER_DATA_FIELD_COUNT = 35;
+  // This value must match kPointerDataFieldCount in pointer_data.cc.
```
loic-sharma, line 83: Should this mention `AndroidTouchProcessor.java`?

shell/platform/embedder/embedder.cc
dkwingsmt: I'm kind of hesitant here. The benefit of using 0 as the default value is that we can flush the struct to be all 0s and be sure that it doesn't break even when new fields are added. But I also agree that the default value should likely be the implicit view. In this way, we should probably NOT allow embedders to customize the implicit view ID to be non-0.

shell/platform/darwin/ios/framework/Source/FlutterViewController.mm
dkwingsmt: Did he mention anywhere the reason for this? Does it only apply to Objective-C code or also the embedder API?

shell/platform/android/io/flutter/embedding/android/AndroidTouchProcessor.java
dkwingsmt: My idea is to have a centralized place to record the mirrored values instead of having all values to mention each other. So if you're to modify this value in `AndroidTouchProcessor`, you would realized "ok I have to change pointer_data.cc as well", then you go to `pointer_data.cc` and find other values to change accordingly.

loic-sharma: Gotcha. If we go with this approach, I would document this so that this pattern is clear:

1. I would update `pointer_data.cc` to declare itself as the centralized location that should be kept up-to-date with all other locations
2. I would update files that point to all other locations (like `pointer_data.h`) to point to `pointer_data.cc`
3. For comments that point to `pointer_data.cc`, I would add something like "See pointer_data.cc for the centralized list of locations that must be kept consistent".

shell/platform/darwin/ios/framework/Source/FlutterViewController.mm
loic-sharma: I think "fooIdentifier" is our preferred naming convention for Darwin platforms

I'm an Objective-C noob, so not sure about `readonly` versus `assign`, `assign` might be perfectly acceptable. I don't feel strongly about this one.

shell/platform/embedder/embedder.cc
loic-sharma: > The benefit of using 0 as the default value is that we can flush the struct to be all 0s and be sure that it doesn't break even when new fields are added

I don't follow. The embedder API - including the new multi-view APIs - is designed for forwards/backwards ABI compatibility. The `struct_size` member lets us determine how to handle the `view_id` without needing to fallback to `0`.

> In this way, we should probably NOT allow embedders to customize the implicit view ID to be non-0.

I don't understand this either, can you expand on this? The embedder API design doc had a section that allows the embedder to customize the implicit view ID: https://docs.google.com/document/d/1kegGJ235y1778-vzyyv90MgSHmYQcROTupMuLtEDBq0/edit#heading=h.1enoc2wkpbna

dkwingsmt: > The struct_size member lets us determine how to handle the view_id without needing to fallback to 0.

I'm talking about creating such a struct. For example, the following line from the GLFW example (somewhat simplified):

```cpp
void GLFWcursorPositionCallbackAtPhase(GLFWwindow* window,
                                       FlutterPointerPhase phase,
                                       double x,
                                       double y) {
  FlutterPointerEvent event = {};
  event.struct_size = sizeof(event);
  event.phase = phase;
  event.x = x * g_pixelRatio;
  event.y = y * g_pixelRatio;
  // event.view_id = ??? // Does not exist in old apps
  FlutterEngineSendPointerEvent(&event, 1);
}
```

This code is actually problematic. Old apps do not assign `event.view_id`. Therefore, once Flutter adds this new field `view_id` and this custom embedder is **recompiled** without editing, every pointer event Flutter received will have an uninitialized `view_id`, which will be a random number, because `FlutterPointerEvent` is a C struct and do not have default values. this PR.

The `struct_size` guard only works for linking existing compiled libraries.

Therefore, truly backward-compatible code needs to flush all bytes of C structs to 0.

```cpp
void GLFWcursorPositionCallbackAtPhase(GLFWwindow* window,
                                       FlutterPointerPhase phase,
                                       double x,
                                       double y) {
  FlutterPointerEvent event = {};
  memset(this, 0, sizeof(FlutterPointerEvent));  // Add this line
  event.struct_size = sizeof(event);
  event.phase = phase;
  event.x = x * g_pixelRatio;
  event.y = y * g_pixelRatio;
  FlutterEngineSendPointerEvent(&event, 1);
}
```

However, this is equivalent to setting `event.view_id = 0;`.

shell/platform/android/io/flutter/embedding/android/AndroidTouchProcessor.java
dkwingsmt: Done. Can you take a look to see if the words are clear enough?

shell/platform/embedder/embedder.cc
loic-sharma: There's two cases that are handled differently:

1. **Embedders in the engine repo** - These automatically use the latest embedder API whenever the API changes.
2. **Embedders outside the engine repo** - These must update their embedder API manually:
    1. [Generate a `flutter_embedder.h` file](https://medium.com/flutter/flutter-on-raspberry-pi-mostly-from-scratch-2824c5e7dcb1)
    2. Copy the `flutter_embedder.h` file into the embedder's project
    3. Update the embedder's logic to use the latest embedder API
    4. Recompile the embedder

See [Chinmay's minimal embedder example](https://github.com/chinmaygarde/flutter_from_scratch/blob/master/README.md).

The `struct_size` guard allows us to detect whether an "outside" embedder has updated their `flutter_embedder.h` file. If an embedder outside the engine repo sends a default value for the view ID, it means the embedder updated their `flutter_embedder.h` file without updating their embedder logic - this would be an error by the "outside" embedder.

dkwingsmt: If the convention is to copy a separate `flutter_embedder.h` then that's fine. But is it mentioned anywhere? Chinmay's example is just his own way. We might want to ask how Flutter defines compatibility here.

lib/ui/platform_dispatcher.dart
```diff
@@ -414,12 +414,10 @@ class PlatformDispatcher {
...
 
-  // If this value changes, update the encoding code in the following files:
-  //
-  //  * pointer_data.cc
-  //  * pointer.dart
-  //  * AndroidTouchProcessor.java
-  static const int _kPointerDataFieldCount = 35;
+  // This value must match kPointerDataFieldCount in pointer_data.cc. (The
+  // pointer_data.cc also listed out other locations that must be kept
+  // consistent.)
```
loic-sharma, line 419: Consider:

```suggestion
  // This value must match kPointerDataFieldCount in pointer_data.cc. (The
  // pointer_data.cc also lists other locations that must be kept consistent.)
```

I don't feel strongly, feel free to keep as is

lib/ui/window/pointer_data.cc
```diff
@@ -8,6 +8,17 @@
 
 namespace flutter {
 
+// The number of fields of PointerData.
+//
+// If kPointerDataFieldCount changes, update the corresponding values to:
+//
+//  * _kPointerDataFieldCount in platform_dispatcher.dart
+//  * POINTER_DATA_FIELD_COUNT in AndroidTouchProcessor.java
```
loic-sharma, line 16: Should this include `PointerData in pointer_data.h`?

dkwingsmt: `PointerData` points to `kPointerDataFieldCount`, not the other way around. You add a field in `PointerData`, you're pointed to `kPointerDataFieldCount`, and you update other fields.

I've added `pointer_data_packet_converter_unittests.cc` in. Thanks for pointing it out.

shell/platform/embedder/embedder.cc
loic-sharma: Here's where "outside" embedders put their copy of `flutter_embedder.h`:

* Sony: https://github.com/sony/flutter-embedded-linux/blob/master/src/flutter/shell/platform/embedder/embedder.h
* Samsung Tizen: https://github.com/flutter-tizen/embedder/blob/master/flutter/shell/platform/embedder/embedder.h
* Toyota: https://github.com/toyota-connected/ivi-homescreen/blob/main/third_party/flutter/shell/platform/embedder/embedder.h
* Go Flutter: https://github.com/go-flutter-desktop/go-flutter/blob/master/embedder/embedder.h
* Raspberry pi: https://github.com/ardera/flutter-pi/blob/master/third_party/flutter_embedder_header/include/flutter_embedder.h

dkwingsmt: I see. In that case we don't have to be limited to use 0 as the default value. And we don't have to use `disable_implicit_view`, since we can make it `enable_implicit_view` and make the default value (by `struct_size`) true.

loic-sharma: > At least I don't think it's not worth the "customizability" that I doubt if many people would use.

It feels a bit late to change this aspect of the design. This design has been out for over a month and has been reviewed by several stakeholders. We can still pivot if this is truly not worth the effort, but I'm not convinced that's the case.

loic-sharma: > And we don't have to use `disable_implicit_view`, since we can make it `enable_implicit_view` and make the default value (by `struct_size`) true.

We discussed this with Chinmay but decided to stick to `disable_implicit_view` regardless. The poor name does minimize the impact to a custom embedder that is updated incorrectly (assuming they zero initializes)

dkwingsmt: > It feels a bit late to change this aspect of the design. This design has been out for over a month and has been reviewed by several stakeholders. We can still pivot if this is truly not worth the effort, I'm not convinced that's the case (and it takes effort to review design updates!)

It's never too late to make improvement as long as it has not been submitted into `embedder.h`. Not to mention the fact that the implicit view ID and its customization are probably among last features of this project to add. If we worry about the minimizing the impact to a custom embedder that is updated incorrectly  for `disable_implicit_view`, then we probably want to make it similar for the view ID.

For this particular change though, I'll change it to `kFlutterImplicitViewId` since I agree that the default value should be the implicit view ID, no matter what that is.

lib/ui/window/pointer_data_packet_converter_unittests.cc
dkwingsmt: Done!

lib/ui/window/pointer_data.h
```diff
@@ -32,7 +28,13 @@ enum PointerButtonStylus : int64_t {
   kPointerButtonStylusSecondary = 1 << 2,
 };
 
-// This structure is unpacked by hooks.dart.
+// This structure is unpacked by platform_dispatcher.dart.
+//
+// If this struct changes, update:
+//  * kPointerDataFieldCount in pointer_data.cc. (The pointer_data.cc also
+//    listed out other locations that must be kept consistent.)
```
loic-sharma, line 35:
```suggestion
//    lists out other locations that must be kept consistent.)
```

loic-sharma (approved):
Engine, embedder API, and desktop embedder changes look good to me!

Please consider another review for Android and iOS changes.

chinmaygarde (approved):

PR #46214, opened 2023-09-22
[ios] Fix default assets url

cyanglaz:
When reverting the default asset url code, the old "flutter_assets" path was also copied in https://github.com/flutter/engine/pull/46073
This PR uses the correct URL. Although I'm unsure why the video player test passed before my change for the asset urls. 

I'm going to take another look at it but meanwhile this PR can unblock the roll once rolled into the framework.

fixes of https://github.com/flutter/flutter/issues/135323

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/shell/platform/darwin/common/framework/Source/FlutterNSBundleUtils.mm
+++ b/shell/platform/darwin/common/framework/Source/FlutterNSBundleUtils.mm
@@ -54,7 +54,7 @@ NSBundle* FLTFrameworkBundleWithIdentifier(NSString* flutterFrameworkBundleID) {
 }
 
 NSString* FLTAssetPath(NSBundle* bundle) {
-  return [bundle objectForInfoDictionaryKey:@"FLTAssetsPath"] ?: @"flutter_assets";
+  return [bundle objectForInfoDictionaryKey:@"FLTAssetsPath"] ?: kDefaultAssetPath;
 }
 
 NSString* FLTAssetsPathFromBundle(NSBundle* bundle) {
--- a/shell/platform/darwin/ios/framework/Source/FlutterDartProjectTest.mm
+++ b/shell/platform/darwin/ios/framework/Source/FlutterDartProjectTest.mm
@@ -102,7 +102,8 @@ FLUTTER_ASSERT_ARC
     id mockBundle = OCMClassMock([NSBundle class]);
     id mockMainBundle = OCMPartialMock([NSBundle mainBundle]);
     NSString* resultAssetsPath = @"path/to/foo/assets";
-    OCMStub([mockBundle pathForResource:@"Frameworks/App.framework/flutter_assets" ofType:@""]).andReturn(nil);
+    OCMStub([mockBundle pathForResource:@"Frameworks/App.framework/flutter_assets" ofType:@""])
+        .andReturn(nil);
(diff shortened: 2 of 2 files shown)


ditman (approved):
(RS)LGTM!

shell/platform/darwin/ios/framework/Source/FlutterDartProjectTest.mm
```diff
@@ -102,14 +102,30 @@ - (void)testFLTAssetsURLFromBundle {
...
+    XCTAssertEqualObjects(FLTAssetPath(mockBundle), @"foo/assets");
+  }
+  {
+    // No asset path in info.plist, use default value
+    id mockBundle = OCMClassMock([NSBundle class]);
+    OCMStub([mockBundle objectForInfoDictionaryKey:@"FLTAssetsPath"]).andReturn(nil);
+    XCTAssertEqualObjects(FLTAssetPath(mockBundle), kDefaultAssetPath);
+  }
+}
+
```
stuartmorgan, line 128: These tests are both of a private helper method; there's nothing about the structure here that would indicate that someone making a change like yours couldn't just change these tests if they broke.

The deeper problem is that we appear to have no tests of the public `lookupKeyForAsset:*` methods in this file. Could you add tests for each variant of that, with comments on the expectations saying that they are testing public API behaviors and anything that breaks the expectations is likely to break plugins?

stuartmorgan (changes requested):

cyanglaz:
@stuartmorgan Added the tests, PTAL

stuartmorgan (approved):
LGTM, thanks for the additional tests!

shell/platform/darwin/common/framework/Source/FlutterNSBundleUtils.mm
```diff
@@ -54,7 +54,7 @@
 }
 
 NSString* FLTAssetPath(NSBundle* bundle) {
-  return [bundle objectForInfoDictionaryKey:@"FLTAssetsPath"] ?: @"flutter_assets";
+  return [bundle objectForInfoDictionaryKey:@"FLTAssetsPath"] ?: kDefaultAssetPath;
```
cyanglaz, line 57: This is actually going to fail the g3 roll again. It turned out I was wrong about changing the API to using NSString fixing the issue. The root cause is when the full relative path is used, unloaded-bundle returns a nil path for assets. we have to use "flutter_assets" directly for unloaded-bundles. 

Because I reverted to @"flutter_assets" in https://github.com/flutter/engine/pull/46073, the roll was fixed.

The fix should be trying to load from kDefaultAssetPath first, if not successful, then try @"flutter_assets" for unloaded bundles. 

I will create a PR to fix this with an integration test. 

cc @stuartmorgan @XilaiZhang

XilaiZhang: looks like this was already rolled in by @CaseyHillers

cyanglaz: Interesting. Maybe it works for XCUITests but not for app extension? That's good to know, we need to fix it for app extensions anyway

PR #46217, opened 2023-09-22
Revert "[Impeller] fail if software backend is chosen and Impeller is enabled on iOS"

CaseyHillers:
Reverts flutter/engine#46124

See b/301660190. This is breaking a few Google iOS tests due to test bed configurations. We'll need more time to investigate fixing those tests before relanding.


dnfield:
As I mentioned in the internal bug, the test failing on this change is not actually testing what it thinks it is testing. Reverting this patch is equal to disabling the test, but will allow more tests to be created that are equally invalid.

It would be better to disable or fix the internal test.

zanderso (approved):

PR #46218, opened 2023-09-22
[Impeller] dont treat non-rects as rects.

jonahwilliams:
We were going down this branch for all closed paths, but we should only be taking it for closed paths that are also rectangles.

### DIFF
--- a/impeller/aiks/aiks_unittests.cc
+++ b/impeller/aiks/aiks_unittests.cc
@@ -1292,15 +1292,14 @@ TEST_P(AiksTest, CanDrawAnOpenPath) {
 TEST_P(AiksTest, CanDrawAnOpenPathThatIsntARect) {
   Canvas canvas;
 
-  // Starting at (50, 50), draw lines from:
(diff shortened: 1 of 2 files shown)


jason-simmons (approved):

impeller/aiks/aiks_unittests.cc
```diff
@@ -1289,6 +1289,28 @@ TEST_P(AiksTest, CanDrawAnOpenPath) {
   ASSERT_TRUE(OpenPlaygroundHere(canvas.EndRecordingAsPicture()));
 }
 
+TEST_P(AiksTest, CanDrawAnOpenPathThatIsntARect) {
```
zanderso, line 1292: Is this one of the playground tests that also makes a golden?

jonahwilliams: IIRC these are all playground tests, so we should see a golden delta pop up in a bit.

zanderso (approved):
lgtm

PR #46219, opened 2023-09-22
[Impeller] Dont blow away coverage hint on advanced blends.

jonahwilliams:
Advnaced blends are given a fairly accurate coverage hint based on the entity size. We were blowing this value away when applying the stencil coverage, resulting in oversized offscreen textures. This should improve the android advanced blend benchmark


impeller/entity/contents/content_context.cc
```diff
@@ -162,16 +162,19 @@ static std::unique_ptr<PipelineT> CreateDefaultPipeline(
 
 ContentContext::ContentContext(
     std::shared_ptr<Context> context,
-    std::shared_ptr<TypographerContext> typographer_context)
+    std::shared_ptr<TypographerContext> typographer_context,
+    std::shared_ptr<RenderTargetAllocator> render_target_allocator)
```
jonahwilliams, line 166: This change allows tests to provide their own RenderTargetAllocator implementation, which I use to introspect into texture allocation.

bdero (approved):
LGTM

PR #46232, opened 2023-09-23
In run_tests.py, separate Dart package unit tests from other tests

zanderso:
This PR creates a new test type for `run_tests.py` called `dart-host`. The tests remaining under the `dart` type are tests that run in `flutter_tester`. The `dart-host` tests run in the Dart CLI. This allows `run_tests.py` to be simplified a bit, and while doing this I spotted a couple of mistakes that were causing some tests not to run.

This PR also makes a couple of small style fixes to the build config json files, converting tabs to spaces, and moving some "script" fields before the "parameters" fields.

### DIFF
--- a/tools/pkg/engine_repo_tools/test/engine_repo_tools_test.dart
+++ b/tools/pkg/engine_repo_tools/test/engine_repo_tools_test.dart
@@ -202,20 +202,36 @@ void main() {
 
     try {
       // Create a valid engine.
-      io.Directory(p.join(emptyDir.path, 'src', 'flutter')).createSync(recursive: true);
-      io.Directory(p.join(emptyDir.path, 'src', 'out')).createSync(recursive: true);
+      final io.Directory srcDir = io.Directory(p.join(emptyDir.path, 'src'))
+        ..createSync(recursive: true);
+      final io.Directory flutterDir = io.Directory(p.join(srcDir.path, 'flutter'))
+        ..createSync(recursive: true);
+      final io.Directory outDir = io.Directory(p.join(srcDir.path, 'out'))
+        ..createSync(recursive: true);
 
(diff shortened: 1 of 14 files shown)


tools/pkg/engine_repo_tools/test/engine_repo_tools_test.dart
```diff
@@ -202,20 +202,36 @@ void main() {
...
       // Create two targets in out: host_debug and host_debug_unopt_arm64.
-      io.Directory(p.join(emptyDir.path, 'src', 'out', 'host_debug')).createSync(recursive: true);
-      io.Directory(p.join(emptyDir.path, 'src', 'out', 'host_debug_unopt_arm64')).createSync(recursive: true);
+      final io.Directory hostDebug = io.Directory(p.join(outDir.path, 'host_debug'))
+        ..createSync(recursive: true);
+      final io.Directory hostDebugUnoptArm64 = io.Directory(
+        p.join(outDir.path, 'host_debug_unopt_arm64'),
+      )..createSync(recursive: true);
+
+      final Engine engine = TestEngine.withPaths(
```
zanderso, line 219: The tests in this file weren't being run, and this test was failing on Linux. The test had been assuming that changing the modification time on a file would update the modification time on its parent directory, but this is generally not the case. To avoid thinking about that too hard, I just changed this test to use test instances. There's some follow-up thinking that needs to be done about how `latestOutput` works, but I'll leave that for a follow-up.

matanlurey: Yeah sounds good, those test outputs didn't exist when I originally authored, this is a better plat-agnostic approach.

matanlurey (approved):
Nice cleanup, and thanks for noticing the test wasn't running!

PR #46273, opened 2023-09-25
Add package:tar to DEPS to fix broken 3H configurations

bkonyi:
Filed https://github.com/flutter/flutter/issues/135463 to track adding support for auto-updating this dependency as part of the SDK roll.

### DIFF
--- a/DEPS
+++ b/DEPS
@@ -471,6 +471,9 @@ deps = {
   'src/third_party/dart/third_party/pkg/string_scanner':
    Var('dart_git') + '/string_scanner.git@da9142cf9809e7e1364144b8193ec60d87f0a4b8',
 
+  'src/third_party/dart/third_party/pkg/tar':
+   Var('dart_git') + '/external/github.com/simolus3/tar.git@3c68cba8e51c569428222b9185469249206172c6',
+
   'src/third_party/dart/third_party/pkg/term_glyph':
    Var('dart_git') + '/term_glyph.git@1b28285a7e818b8e87c4d2119d968c5b36d73c7a',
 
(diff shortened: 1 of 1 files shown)


aam (approved):

aam:
nit regarding  "to fix broken roll": it's not the roll that is broken(engine uses prebuilt dart sdk, so missing dependency in engine's DEPS is not noticeable), rather the hhh buildbot on go/darto (https://ci.chromium.org/ui/p/dart/builders/ci.sandbox/flutter-linux-host_debug_unopt/2855/overview example of the failure)

bkonyi:
> nit regarding "to fix broken roll": it's not the roll that is broken(engine uses prebuilt dart sdk, so missing dependency in engine's DEPS is not noticeable), rather the hhh buildbot on go/darto (https://ci.chromium.org/ui/p/dart/builders/ci.sandbox/flutter-linux-host_debug_unopt/2855/overview example of the failure)

Good point! I've updated the description.

bkonyi:
@aam I've updated the PR to instead just add the entry to the list of auto-updated Dart pkg DEPS. It looks like `create_updated_flutter_deps.py` does correctly handle this URL once the entry in the DEPS file is present (confirmed locally with a fake revision update).

a-siva:
> @aam I've updated the PR to instead just add the entry to the list of auto-updated Dart pkg DEPS. It looks like `create_updated_flutter_deps.py` does correctly handle this URL once the entry in the DEPS file is present (confirmed locally with a fake revision update).

Then why did the PR from @whesse (https://github.com/flutter/engine/pull/46190) not work ?

bkonyi:
I'm not sure, but I did verify this works locally. There's nothing wrong
with the script as far as I can tell.

aam:
> I'm not sure, but I did verify this works locally. There's nothing wrong with the script as far as I can tell.

I think as @whesse landed DEPS change to flutter/engine, dart sdk roll _without_ this tar dependency was incoming. So the create_updated_flutter_deps.py script that ran for that dart sdk roll(that ran after Bill's change in flutter/engine landed), removed tar dependency from the engine .

PR #46282, opened 2023-09-25
[Impeller] Fallback to no index buffer when tesselation count is large, split up nonZero contours.

jonahwilliams:
See example app in https://github.com/flutter/flutter/issues/135458

Because tessellation is generally nlogn best case (with allocation complexity as well), we can actually make it faster by breaking it into smaller pieces. For non zero fill modes, the contours can be tessellated individually without  losing fidelity.

In cases where this isn't possible, we need to check if we exceed the max uint16 index and fall back to no index buffer.

Fixes https://github.com/flutter/flutter/issues/135458

I Benchmarked signing my name.

### Before

 worst raster time 3+ seconds. Visual glitches

flutter_03

### After
worst raster time 30 ms, looks correct (though my signature is still wrong)

flutter_04

### DIFF
--- a/impeller/entity/geometry/fill_path_geometry.cc
+++ b/impeller/entity/geometry/fill_path_geometry.cc
@@ -3,6 +3,7 @@
 // found in the LICENSE file.
 
 #include "impeller/entity/geometry/fill_path_geometry.h"
+#include "impeller/core/formats.h"
 
 namespace impeller {
 
@@ -48,10 +49,15 @@ GeometryResult FillPathGeometry::GetPositionBuffer(
           size_t indices_count) {
         vertex_buffer.vertex_buffer = host_buffer.Emplace(
             vertices, vertices_count * sizeof(float), alignof(float));
-        vertex_buffer.index_buffer = host_buffer.Emplace(
-            indices, indices_count * sizeof(uint16_t), alignof(uint16_t));
-        vertex_buffer.vertex_count = indices_count;
-        vertex_buffer.index_type = IndexType::k16bit;
+        if (indices != nullptr) {
+          vertex_buffer.index_buffer = host_buffer.Emplace(
+              indices, indices_count * sizeof(uint16_t), alignof(uint16_t));
+          vertex_buffer.vertex_count = indices_count;
+          vertex_buffer.index_type = IndexType::k16bit;
+        } else {
+          vertex_buffer.vertex_count = vertices_count;
+          vertex_buffer.index_type = IndexType::kNone;
+        }
         return true;
       });
   if (tesselation_result != Tessellator::Result::kSuccess) {
--- a/impeller/entity/geometry/fill_path_geometry.cc
+++ b/impeller/entity/geometry/fill_path_geometry.cc
@@ -48,7 +48,7 @@ GeometryResult FillPathGeometry::GetPositionBuffer(
           const float* vertices, size_t vertices_count, const uint16_t* indices,
           size_t indices_count) {
         vertex_buffer.vertex_buffer = host_buffer.Emplace(
-            vertices, vertices_count * sizeof(float), alignof(float));
+            vertices, vertices_count * sizeof(float) * 2, alignof(float));
         if (indices != nullptr) {
           vertex_buffer.index_buffer = host_buffer.Emplace(
               indices, indices_count * sizeof(uint16_t), alignof(uint16_t));
@@ -56,7 +56,7 @@ GeometryResult FillPathGeometry::GetPositionBuffer(
           vertex_buffer.index_type = IndexType::k16bit;
         } else {
           vertex_buffer.index_buffer = {};
-          vertex_buffer.vertex_count = indices_count;
+          vertex_buffer.vertex_count = vertices_count;
           vertex_buffer.index_type = IndexType::kNone;
         }
         return true;
@@ -119,7 +119,7 @@ GeometryResult FillPathGeometry::GetPositionUVBuffer(
       [&vertex_builder, &texture_coverage, &effect_transform](
           const float* vertices, size_t vertices_count, const uint16_t* indices,
           size_t indices_count) {
-        for (auto i = 0u; i < vertices_count; i += 2) {
+        for (auto i = 0u; i < vertices_count * 2; i += 2) {
           VS::PerVertexData data;
           Point vtx = {vertices[i], vertices[i + 1]};
           data.position = vtx;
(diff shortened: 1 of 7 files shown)


jonahwilliams:
This still looks really messed up after just a bit of drawing

impeller/tessellator/tessellator.cc
```diff
@@ -78,51 +78,132 @@ Tessellator::Result Tessellator::Tessellate(
...
-    size_t start_point_index, end_point_index;
-    std::tie(start_point_index, end_point_index) =
-        polyline.GetContourPointBounds(contour_i);
-
-    ::tessAddContour(tessellator,  // the C tessellator
-                     kVertexSize,  //
-                     polyline.points.data() + start_point_index,  //
-                     sizeof(Point),                               //
-                     end_point_index - start_point_index          //
+  if (polyline.contours.size() > 30 && fill_type == FillType::kNonZero) {
```
zanderso, line 81: Consider pulling 30 out to a named constant, and as much as might be possible, explain why 30 was chosen.

impeller/tessellator/tessellator.cc
```diff
@@ -78,51 +78,132 @@ Tessellator::Result Tessellator::Tessellate(
...
+      for (int i = 0; i < elementItemCount; i++) {
+        indices[i] = static_cast<uint16_t>(elements[i]);
+      }
+      if (!callback(vertices, vertexItemCount, indices.data(),
+                    elementItemCount)) {
+        return Result::kInputError;
+      }
+    } else {
+      std::vector<Point> points;
+      std::vector<uint16_t> indices;
```
gaaclarke, line 188: This value is generated, but never used.

impeller/tessellator/tessellator.cc
gaaclarke: I'd also add a comment about not using indices since there is a lot of code to read through to understand that's the case.

impeller/tessellator/tessellator.cc
jonahwilliams: Fixed

impeller/tessellator/tessellator.cc
jonahwilliams: Added more comments here.

impeller/tessellator/tessellator.cc
```diff
@@ -78,51 +82,141 @@ Tessellator::Result Tessellator::Tessellate(
...
+                                    kPolygonSize,   // polygon size
+                                    kVertexSize,    // vertex size
+                                    nullptr  // normal (null is automatic)
+      );
+
+      if (result != 1) {
+        return Result::kTessellationError;
+      }
+
+      int vertexItemCount = tessGetVertexCount(tessellator) * kVertexSize;
```
gaaclarke, line 126:
```suggestion
      int vertex_item_count = tessGetVertexCount(tessellator) * kVertexSize;
```
These variable names don't conform to the c++ style guide.

impeller/tessellator/tessellator_unittests.cc
```diff
@@ -82,6 +83,52 @@ TEST(TessellatorTest, TessellatorBuilderReturnsCorrectResultStatus) {
     ASSERT_EQ(polyline.points.size(), 2u);
     ASSERT_EQ(result, Tessellator::Result::kInputError);
   }
+
+  // More than 30 contours, non-zero fill mode.
+  {
+    Tessellator t;
+    PathBuilder builder = {};
+    for (auto i = 0; i < 60; i++) {
```
gaaclarke, line 91: Instead of using a hardcoded 60, you can use the variable you created now to make this test a bit more robust to changes in that number.

impeller/entity/geometry/fill_path_geometry.cc
```diff
@@ -48,10 +49,16 @@ GeometryResult FillPathGeometry::GetPositionBuffer(
           size_t indices_count) {
```
gaaclarke, line 49: The variable name here belies the fact that it actually can represent 2 different things.

impeller/tessellator/tessellator.h
```diff
@@ -44,10 +44,17 @@ class Tessellator {
...
+  ///        The index buffer may not be populated, in which case [indices] will
+  ///        be nullptr. The [vertices_size] is the size of the vertices array
+  ///        and not the number of vertices, which is usually vertices_size / 2.
+  ///        In cases where the indices is nullptr, vertex_or_index_count will
+  ///        contain the count of indices.
   using BuilderCallback = std::function<bool(const float* vertices,
                                              size_t vertices_size,
                                              const uint16_t* indices,
-                                             size_t indices_size)>;
+                                             size_t vertex_or_index_count)>;
```
gaaclarke, line 57: Overloading this variable doesn't seem prudent.  Can't it just be indices_size and be zero if `indices == nullptr`?  We didn't need vertex count in the past

gaaclarke (changes requested):
Code looks good, awesome surprising performance results.  The only thing that should change is the overloaded parameter that can mean 2 different things.

impeller/tessellator/tessellator.h
jonahwilliams: The reason this is overloaded is a bit of an issue yes, I should just fix it. Really the problem is that these variables aren't counts, they are the value to multiply with sizeof(float)/sizeof(uint16_t) to compute offsets. But I just make them counts, then this problem goes away.

impeller/tessellator/tessellator.cc
jonahwilliams: Done

impeller/entity/geometry/fill_path_geometry.cc
jonahwilliams: Fixed

impeller/tessellator/tessellator.h
jonahwilliams: Done

impeller/tessellator/tessellator_unittests.cc
jonahwilliams: Done

impeller/tessellator/tessellator.cc
```diff
@@ -78,51 +78,141 @@ Tessellator::Result Tessellator::Tessellate(
...
-  for (int i = 0; i < elementItemCount; i++) {
-    indices[i] = static_cast<uint16_t>(elements[i]);
-  }
-  if (!callback(vertices, vertexItemCount, indices.data(), elementItemCount)) {
-    return Result::kInputError;
+    if (result != 1) {
+      return Result::kTessellationError;
+    }
+
+    int elementItemCount = tessGetElementCount(tessellator) * kPolygonSize;
```
gaaclarke, line 174:
```suggestion
    int element_item_count = tessGetElementCount(tessellator) * kPolygonSize;
```

impeller/tessellator/tessellator.cc
```diff
@@ -78,51 +78,141 @@ Tessellator::Result Tessellator::Tessellate(
...
+
+    // We default to using a 16bit index buffer, but in cases where we generate
+    // more tessellated data than this can contain we need to fall back to
+    // dropping the index buffer entirely. Instead code could instead switch to
+    // a uint32 index buffer, but this is done for simplicity with the other
+    // fast path above.
+    if (elementItemCount < 65535) {
```
gaaclarke, line 181:
```suggestion
    if (elementItemCount < USHRT_MAX) {
```

gaaclarke (approved):
lgtm!

PR #46286, opened 2023-09-26
[Impeller] Destroy all per-thread command pools tied to a context before deleting the context

jason-simmons:
Vulkan requires that all objects created using a device be destroyed before the device is destroyed.  The CommandPoolRecyclerVK maintains a thread-local map of each context's currently active command pool. Before the context deletes the device, it must force cleanup of all such command pools associated with that context.

This PR reinstates a mechanism that was used prior to the refactoring that introduced command pool recycling.


matanlurey (approved):
Thanks Jason!

PR #46304, opened 2023-09-26
Declare native wrapper classes in the GPU package as base classes

jason-simmons:
This prohibits other implementations of the class interface that can not act as native wrappers.

See https://github.com/flutter/flutter/issues/123756


bdero (approved):
LGTM

PR #46360, opened 2023-09-28
[Linux] Rename 'set_allow_channel_overflow' to 'set_warns_on_channel_…

bleroux:
## Description

This PR is a follow-up to https://github.com/flutter/engine/pull/44636 which introduces the `set_allow_channel_overflow` function. It renames this function to `set_warns_on_channel_overflow`.

The previous naming was inspired by the framework side implementation.
https://github.com/flutter/engine/blob/93e8901490e78c7ba7e319cce4470d9c6478c6dc/lib/ui/channel_buffers.dart#L574

During the review of the iOS/macOS implementation, https://github.com/flutter/engine/pull/44848#discussion_r1310463864, we agreed that the existing naming is confusing because it implies that overflow will be allowed, but this not the case this function is used to enable/disable error messages when a channel overflows. 

## Related Issue

Follow-up for https://github.com/flutter/flutter/issues/132386.

## Tests

Updates 2 tests.

### DIFF
--- a/shell/platform/linux/fl_binary_messenger.cc
+++ b/shell/platform/linux/fl_binary_messenger.cc
@@ -359,28 +359,28 @@ static void resize_channel(FlBinaryMessenger* messenger,
                                       nullptr);
 }
 
-// Called when a response is received for the allow channel overflow message.
-static void set_allow_channel_overflowl_response_cb(GObject* object,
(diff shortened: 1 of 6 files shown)


robert-ancell (approved):
LGTM

PR #46373, opened 2023-09-28
Don't respond to the `insertionPointColor` selector on iOS 17+

LongCatIsLooong:
Fixes https://github.com/flutter/flutter/issues/132548

## Pre-launch Checklist

- [ ] I signed the [CLA].

### DIFF
--- a/shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
+++ b/shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
@@ -943,7 +943,7 @@ static BOOL IsSelectionRectBoundaryCloserToPoint(CGPoint point,
       return NO;
     }
   }
-  return [FlutterTextInputView instancesRespondToSelector:selector];
+  return [super respondsToSelector:selector];
 }
 
 - (void)setTextInputClient:(int)client {
--- a/shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
+++ b/shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
@@ -902,10 +902,11 @@ static BOOL IsSelectionRectBoundaryCloserToPoint(CGPoint point,
 
 // Prevent UIKit from showing selection handles or highlights. This is needed
 // because Scribble interactions require the view to have it's actual frame on
-// the screen.
+// the screen. They're not needed on iOS 17 with the new
+// UITextSelectionDisplayInteraction API.
 //
-// These are not documented methods. On iOS 17, the insertion point color is also
-// used to paint the highlighted background of the selected IME candidate:
+// These are undocumented methods. On iOS 17, the insertion point color is also
+// used as the highlighted background of the selected IME candidate:
 // https://github.com/flutter/flutter/issues/132548
 // So the respondsToSelector method is overridden to return NO for this method
 // on iOS 17+.
--- a/shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
+++ b/shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
@@ -903,6 +903,12 @@ static BOOL IsSelectionRectBoundaryCloserToPoint(CGPoint point,
 // Prevent UIKit from showing selection handles or highlights. This is needed
 // because Scribble interactions require the view to have it's actual frame on
 // the screen.
+//
+// These are not documented methods. On iOS 17, the insertion point color is also
+// used to paint the highlighted background of the selected IME candidate:
+// https://github.com/flutter/flutter/issues/132548
+// So the respondsToSelector method is overridden to return NO for this method
+// on iOS 17+.
 - (UIColor*)insertionPointColor {
   return [UIColor clearColor];
 }
@@ -932,6 +938,7 @@ static BOOL IsSelectionRectBoundaryCloserToPoint(CGPoint point,
 
 - (BOOL)respondsToSelector:(SEL)selector {
   if (@available(iOS 17.0, *)) {
+    // See the comment on this method.
     if (selector == @selector(insertionPointColor)) {
       return NO;
     }
--- a/shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
+++ b/shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
@@ -930,6 +930,15 @@ static BOOL IsSelectionRectBoundaryCloserToPoint(CGPoint point,
   return _textInputPlugin.textInputDelegate;
 }
 
+- (BOOL)respondsToSelector:(SEL)selector {
+  if (@available(iOS 17.0, *)) {
+    if (selector == @selector(insertionPointColor)) {
+      return NO;
+    }
+  }
+  return [FlutterTextInputView instancesRespondToSelector:selector];
+}
+
 - (void)setTextInputClient:(int)client {
   _textInputClient = client;
   _hasPlaceholder = NO;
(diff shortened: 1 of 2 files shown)


shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
```diff
@@ -903,6 +903,12 @@ - (UITextContentType)textContentType {
 // Prevent UIKit from showing selection handles or highlights. This is needed
 // because Scribble interactions require the view to have it's actual frame on
 // the screen.
+//
+// These are not documened methods. On iOS 17, the insertion point color is also
+// used to paint the highlighted background of the selected IME candidate:
```
hellohuanlin, line 908: what else do we use this function for?

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
```diff
@@ -903,6 +903,12 @@ - (UITextContentType)textContentType {
 // Prevent UIKit from showing selection handles or highlights. This is needed
 // because Scribble interactions require the view to have it's actual frame on
 // the screen.
+//
+// These are not documened methods. On iOS 17, the insertion point color is also
```
stuartmorgan, line 907: Nit: documented

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
```diff
@@ -903,6 +903,12 @@ - (UITextContentType)textContentType {
...
+// used to paint the highlighted background of the selected IME candidate:
+// https://github.com/flutter/flutter/issues/132548
+// So the respondsToSelector method is overridden to return NO for this method
+// on iOS 17+.
```
stuartmorgan, line 911: I'm confused as to why we are both implementing it but claiming not to. Are we relying on an implementation detail of some system code calling this unconditionally and other code checking whether its implemented first, and relying on that distinction to make things work? If so, that seems incredibly fragile, and the sort of thing that's likely to cause another round of fire drills in a subsequent iOS version.

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
```diff
@@ -930,6 +936,16 @@ - (UIInputViewController*)inputViewController {
...
 }
 
+- (BOOL)respondsToSelector:(SEL)selector {
+  if (@available(iOS 17.0, *)) {
+    // See the comment on this method.
+    if (selector == @selector(insertionPointColor)) {
+      return NO;
+    }
+  }
+  return [FlutterTextInputView instancesRespondToSelector:selector];
```
stuartmorgan, line 946: Why isn't this just calling super?

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
LongCatIsLooong: The paragraph above says that it's used to make sure the system caret / selection handles aren't visible.

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
LongCatIsLooong: UIKit falls back to `UIColor.insertionPointColor` if the receiver doesn't respond to `insertionPointColor`. That color is unfortunately private. 

> some system code calling this unconditionally

The method is actually private (but it's in the codebase for a while and no complains from app store) and it's not part of any protocols, so UIKit has to check `respondsToSelector` first.

stuartmorgan: > The method is actually private (but it's in the codebase for a while and no complains from app store) and it's not part of any protocols, so UIKit has to check `respondsToSelector` first.

Then how does this change not break the "Prevent UIKit from showing selection handles or highlights" use case that the implementation is documented as being for?

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
LongCatIsLooong: According to the documentation `[super respondsToSelector:]` is equivalent to `[self respondsToSelector:]` https://developer.apple.com/documentation/objectivec/1418956-nsobject/1418583-respondstoselector#discussion, so i thought there was going to be an infinite loop. Let me try it out.

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
LongCatIsLooong: On iOS 17 there's new API for adding system UI for text selection/cursor/highlights: https://developer.apple.com/documentation/uikit/uitextselectiondisplayinteraction?language=objc . Without adding that interaction to the input view (and we don't add  [`UIITextInteraction`](https://developer.apple.com/documentation/uikit/uitextinteraction?language=objc) to the view unless the framework is currently sending editing state updates, as a hack to fix a different bug), UIKit doesn't seem to add selection handles or highlights automatically on iOS 17.

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
LongCatIsLooong: Well that worked. Changed to calling super.

stuartmorgan: The context there is important:

> You cannot test whether an object inherits a method from its superclass by sending respondsToSelector: to the object using the super keyword. This method will still be testing the object as a whole, not just the superclass’s implementation. Therefore, sending respondsToSelector: to super is equivalent to sending it to self.

For the purposes of testing where a method's implementation lives, it's equivalent. That doesn't mean it doesn't follow actual method dispatch rules.

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
stuartmorgan: I see; please update the comments on this method then to explain this, and that it's bypassed on iOS 17+ by a `respondsToSelector:` check.

LongCatIsLooong: Hi @fbcouch do you remember when is `insertionPointColor` called during scribble? I'd like to verify if this is still needed on iOS17.

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
```diff
@@ -930,6 +937,16 @@ - (UIInputViewController*)inputViewController {
   return _textInputPlugin.textInputDelegate;
 }
 
+- (BOOL)respondsToSelector:(SEL)selector {
+  if (@available(iOS 17.0, *)) {
+    // See the comment on this method.
+    if (selector == @selector(insertionPointColor)) {
+      return NO;
```
hellohuanlin, line 944: We need to be careful if this breaks other functionalities that uses `insertionPointColor`.

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
fbcouch: If I recall correctly, it was basically during any text interaction – since we changed to show the frame in it's proper place on the screen, iPadOS was able to draw it's own selection handles, etc

LongCatIsLooong:
I did some basic IME input testing and didn't notice any UI weirdness.

hellohuanlin (approved):

stuartmorgan (approved):
LGTM as long as we're reasonably confident that we've manually tested all of the cases that the code being bypassed was originally intended to fix.

PR #46381, opened 2023-09-28
Add initial support for 4x MSAA in OpenGLES backend.

matanlurey:
Initial support _towards_ https://github.com/flutter/flutter/issues/130045.

This updates the `PROC_TABLE`, and adds (and supports) `Type::kRenderBufferMultisampled`.

The type remains unused in the actual engine, and will be in a follow-up PR once we like this one!

/cc @chinmaygarde

### DIFF
--- a/impeller/renderer/backend/gles/gles.h
+++ b/impeller/renderer/backend/gles/gles.h
@@ -4,6 +4,8 @@
 
 #pragma once
 
+// IWYU pragma: begin_exports
 #include "GLES3/gl3.h"
 #define GL_GLEXT_PROTOTYPES
 #include "GLES2/gl2ext.h"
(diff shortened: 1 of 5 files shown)


impeller/renderer/backend/gles/gles.h
```diff
@@ -4,6 +4,8 @@
 
 #pragma once
 
+// IWYU pragma: begin_exports
```
matanlurey, line 7: This is neat, apparently you can tell IWYU how includes are being used:
https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/IWYUPragmas.md#iwyu-pragma-export

jonahwilliams (approved):
LGTM

PR #46389, opened 2023-09-28
[Impeller] flutter_tester --enable-impeller

dnfield:
This patch does the following:

- Updates `flutter_tester` to set up an Impeller rendering context and surface if `--enable-impeller` is set to true, using the Vulkan backend with Swiftshader.
- Updates `run_tests.py` to run all tests except the smoke test (that one really has no rendering impact whatsoever) with and without `--enable-impeller`.
- Updates a few tests to work that were trivial:
  - A couple tests needed updated goldens for very minor rendering differences. Filed https://github.com/flutter/flutter/issues/135684 to track using Skia gold for this instead.
  - Disabled SKP screenshotting if Impeller is enabled, and updated the test checking that to verify an error is thrown if an SKP is requested.
  - The Dart GPU based test now asserts that the gpu context is available if Impeller is enabled, and does not deadlock if run in a single threaded mode.
  - We were missing some trace events around `Canvas::SaveLayer` for Impeller as compared to Skia.
  - A couple other tests had strict checks about exception messages that are slightly different between Skia and Impeller.
- I've filed bugs for other tests that may require a little more work, and skipped them for now. For FragmentProgram on Vulkan I reused an existing bug.

This is part of my attempt to address https://github.com/flutter/flutter/issues/135693, although @chinmaygarde and I had slightly different ideas about how to do this.

The goals here are:

- Run the Dart unit tests we already have with Impeller enabled.
- Enable running more of the framework tests (including gold tests) with Impeller enabled.
- Run all of these tests via public `dart:ui` API rather than mucking around in C++ internals in the engine.


shell/testing/tester_main.cc
```diff
@@ -235,10 +297,52 @@ int RunTester(const flutter::Settings& settings,
...
+    context_settings.proc_address_callback =
+        reinterpret_cast<PFN_vkGetInstanceProcAddr>(&::vkGetInstanceProcAddr);
+    context_settings.shader_libraries_data = ShaderLibraryMappings();
+    context_settings.cache_directory = fml::paths::GetCachesDirectory();
+    context_settings.enable_validation = settings.enable_vulkan_validation;
+
+    impeller_context = impeller::ContextVK::Create(std::move(context_settings));
+    if (!impeller_context || !impeller_context->IsValid()) {
+      VALIDATION_LOG << "Could not create Vulkan context.";
+      return -1;
```
bdero, line 324:
```suggestion
      return EXIT_FAILURE;
```

shell/testing/tester_main.cc
```diff
@@ -29,6 +29,45 @@
...
+
+static std::vector<std::shared_ptr<fml::Mapping>> ShaderLibraryMappings() {
+  return {
+    std::make_shared<fml::NonOwnedMapping>(impeller_entity_shaders_vk_data,
+                                           impeller_entity_shaders_vk_length),
+        std::make_shared<fml::NonOwnedMapping>(
+            impeller_modern_shaders_vk_data, impeller_modern_shaders_vk_length),
+#if IMPELLER_ENABLE_3D
+        std::make_shared<fml::NonOwnedMapping>(
+            impeller_scene_shaders_vk_data, impeller_scene_shaders_vk_length),
```
bdero, line 55: Thank you for taking care of Scene. 🙏

bdero (approved):
LGTM, just nits!

shell/testing/tester_main.cc
```diff
@@ -235,10 +297,52 @@ int RunTester(const flutter::Settings& settings,
...
+    impeller::vk::HeadlessSurfaceCreateInfoEXT surface_create_info;
+    auto res = impeller_context->GetInstance().createHeadlessSurfaceEXT(
+        &surface_create_info,  // surface create info
+        nullptr,               // allocator
+        &vk_surface            // surface
+    );
+    if (res != impeller::vk::Result::eSuccess) {
+      VALIDATION_LOG << "Could not create surface for tester "
+                     << impeller::vk::to_string(res);
+      return -1;
```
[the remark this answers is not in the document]
dnfield: done here and elsewhere

testing/dart/canvas_test.dart
```diff
@@ -14,6 +14,8 @@ import 'package:vector_math/vector_math_64.dart';
 
 typedef CanvasCallback = void Function(Canvas canvas);
 
+bool get impellerEnabled => Platform.executableArguments.contains('--enable-impeller');
```
zanderso, line 17: Consider factoring this definition out into a `.dart` file under `testing/dart`, and importing it where needed.

testing/run_tests.py
```diff
@@ -850,17 +854,21 @@ def gather_dart_tests(build_dir, test_filter):
         logger.info(
             "Gathering dart test '%s' with observatory enabled", dart_test_file
         )
-        yield gather_dart_test(build_dir, dart_test_file, True, True)
-        yield gather_dart_test(build_dir, dart_test_file, False, True)
+        yield gather_dart_test(build_dir, dart_test_file, True, False, True)
```
zanderso, line 857: This probably should have been done earlier, but 4 unnamed boolean flags seems like too many. Maybe the simplest thing is just to always pass the arguments as named and always pass them all?

dnfield: Done

zanderso:
Many tests appear to be segfaulting. Maybe the tests that are failing under asan are a good starting point?

dnfield:
I'm not seeing segfaults anymore. I do see some flaky failures that seem unrelated to this patch, filed some bugs. Looks like this will be g2g now.

shell/testing/BUILD.gn
```diff
@@ -36,6 +46,26 @@ executable("testing") {
...
+      "//flutter/impeller",
+      "//third_party/swiftshader",
+    ]
+
+    if (is_win) {
+      libs += [ "$root_out_dir/vk_swiftshader.dll" ]
+    } else if (is_mac) {
+      libs += [ "$root_out_dir/libvk_swiftshader.dylib" ]
+    } else if (is_linux) {
+      libs += [ "$root_out_dir/libvk_swiftshader.so" ]
```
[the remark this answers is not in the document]
dnfield: I ended up moving this to `//flutter/vulkan/swiftshader_path.h`

PR #46391, opened 2023-09-28
[Impeller] simplify invert colors flag by supporting composed color filters.

jonahwilliams:
Fixes https://github.com/flutter/flutter/issues/135699

To avoid the ugly complicated code for conditionally applying both colorfilters, make the dl dispatcher handle merging these filters together in a defined order. The original bug is caused by not checking for invert colors in a CPU fast path.

### DIFF
--- a/impeller/aiks/aiks_unittests.cc
+++ b/impeller/aiks/aiks_unittests.cc
@@ -13,6 +13,7 @@
 #include "flutter/testing/testing.h"
 #include "impeller/aiks/aiks_playground.h"
 #include "impeller/aiks/canvas.h"
+#include "impeller/aiks/color_filter.h"
 #include "impeller/aiks/image.h"
 #include "impeller/aiks/image_filter.h"
 #include "impeller/aiks/paint_pass_delegate.h"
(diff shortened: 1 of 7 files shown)


impeller/display_list/dl_dispatcher.cc
```diff
@@ -484,12 +488,24 @@ static std::shared_ptr<ColorFilter> ToColorFilter(
...
 // |flutter::DlOpReceiver|
 void DlDispatcher::setInvertColors(bool invert) {
-  paint_.invert_colors = invert;
+  if (paint_.color_filter) {
+    auto invert_filter = ColorFilter::MakeMatrix(kColorInversion);
+    paint_.color_filter = std::make_shared<MergedColorFilter>(
+        invert_filter, std::move(paint_.color_filter));
+  } else {
+    paint_.color_filter = ColorFilter::MakeMatrix(kColorInversion);
+  }
```
jonahwilliams, line 508: This sucks but there is no good answer to:

1. Be resilient to dl dispatch order and
2. Keep paint as a struct with no private fields.

jonahwilliams:
Hmm, all of these colors should be the same....

jonahwilliams:
okay, changed the filters around...

impeller/aiks/aiks_unittests.cc
```diff
@@ -123,16 +124,47 @@ TEST_P(AiksTest, CanRenderImage) {
   ASSERT_TRUE(OpenPlaygroundHere(canvas.EndRecordingAsPicture()));
 }
 
-TEST_P(AiksTest, CanRenderInvertedImage) {
+constexpr ColorMatrix kColorInversion = {.array = {
```
dnfield, line 127: This is now defined in at least three places - it'd be nice if it could live in some common header somewhere.

dnfield (approved):
LGTM with nit

impeller/aiks/aiks_unittests.cc
jonahwilliams: We could pull this logic all the way up too painting.dart but then we wont be able to test it with impeller yet. For the purposes of this test though it doesn't really matter what the filter is, just that it composes.

impeller/display_list/dl_dispatcher.cc
```diff
@@ -484,12 +488,24 @@ static std::shared_ptr<ColorFilter> ToColorFilter(
 // |flutter::DlOpReceiver|
 void DlDispatcher::setColorFilter(const flutter::DlColorFilter* filter) {
   // Needs https://github.com/flutter/flutter/issues/95434
-  paint_.color_filter = ToColorFilter(filter);
+  if (paint_.color_filter) {
+    auto color_filter = ToColorFilter(filter);
+    paint_.color_filter =
+        ColorFilter::MakeComposed(paint_.color_filter, color_filter);
```
bdero, line 494: Does the paint state get cleared out between every draw call? I think this solution implicitly relies on `setColorFilter` and `setInvertColors` always being called exactly once per dispatch in this solution.

jonahwilliams: ugh yeah

jonahwilliams: I hate this

jonahwilliams: reverted!

impeller/display_list/dl_dispatcher.cc
bdero: To solve the composing problem, one possibility would be to add a "GetInverted()" to `ColorFilter` that returns a new composed filter with the color filter inversion appended.

bdero: (This is one of those cases where I would clone things liberally to prevent bugs -- the only reason these things are smart pointers is for dynamic dispatch, but like so many things in Impeller, these should really just be discriminated unions, but we currently have an aversion for those in C++)

jonahwilliams: Right, this is 2.5 which is "callers need to be aware of both fields" which I guess is just us, but not idea.

PR #46394, opened 2023-09-29
[Impeller] Add Context::GetTextureContents for readback.

jonahwilliams:
Fixes https://github.com/flutter/flutter/issues/133101

Adds an API for reading the base mip level in whatever format a texture happens to be in.


impeller/renderer/context.cc
```diff
@@ -16,4 +17,37 @@ bool Context::UpdateOffscreenLayerPixelFormat(PixelFormat format) {
   return false;
 }
 
+void Context::GetTextureContents(const std::shared_ptr<Texture>& texture,
```
jonahwilliams, line 20: We have something similar implemented in image_decode_impeller.cc, eventually we need to get rid of that and replace with our own logic to convert pixel formats. I'm not sure if the Skia implementation is unintentionally pulling in the software renderer.

jonahwilliams:
Maybe we can do something like an MPSImageConversion on iOS, but probably want to avoid really slow pixel conversions.

jonahwilliams:
Actually we can just do the conversion in the blit pass.

jonahwilliams:
Looks like Metal blit passes wont do format conversions so I'd need to do a render pass. Maybe a separate helper to convert texture format a to format b, then we can have platform specific implementations.

jonahwilliams:
Filled https://github.com/flutter/flutter/issues/135764

impeller/renderer/context.cc
```diff
@@ -16,4 +17,37 @@ bool Context::UpdateOffscreenLayerPixelFormat(PixelFormat format) {
...
+  DeviceBufferDescriptor desc;
+  desc.size = buffer_size;
+  desc.storage_mode = StorageMode::kHostVisible;
+  auto device_buffer = GetResourceAllocator()->CreateBuffer(desc);
+  if (!device_buffer) {
+    callback(nullptr);
+    return;
+  }
+
+  if (!blit_pass->AddCopy(texture, device_buffer)) {
```
bdero, line 36: Does a texture->device buffer blit actually force a transfer from the device to the _host_? If it does, I really think that behavior should be decoupled (not now, but later). But if it doesn't force a host transfer, I don't think this patch will work?

impeller/renderer/context.cc
bdero, line 20: As implemented, a more appropriate name would be `CopyToDeviceBuffer`.

impeller/renderer/context.h
```diff
@@ -155,6 +156,22 @@ class Context {
   ///
   virtual std::shared_ptr<CommandBuffer> CreateCommandBuffer() const = 0;
 
+  using ReadbackCallback =
+      std::function<void(std::shared_ptr<const DeviceBuffer> buffer)>;
```
bdero, line 160: Should this API emit a constant copy of the host byte data rather than a `DeviceBuffer`? I believe later we we can remove the unnecessary blit on all of the backends and even support onscreen textures on GLES.

impeller/renderer/context.cc
bdero: Also, no cap check needed here (like `SupportsBufferToTextureBlits`)?

impeller/renderer/context.h
jonahwilliams: > I believe later we we can remove the unnecessary blit on all of the backends and even support onscreen textures on GLES.

What?

impeller/renderer/context.cc
jonahwilliams: No, this is supported on all backends from my inspection. If it didn't work, then the test would be failing, no?

impeller/renderer/context.h
bdero: Sorry, I meant the device->device copy, as blitting to a `DeviceBuffer` implies.

jonahwilliams: Ahh yeah makes sense. I went back and forth a few times on what the API should be. This actually needs to end up as. a unique_ptr of something, so that I can use the Dart typed data APIs for readback (eventually). So If I change this to be something like std::unique_ptr<const uint8_t[]> or whatever, I don't know C++ , that should all work nicely right?

impeller/renderer/context.cc
bdero: For fixing up the blits (later), I think a good solution would be to have two different `AddCopy` overloads: One that _only_ copies to the host, and another that _only_ copies to a new device allocation. Copying to both should never be necessary.

```
  bool AddCopy(std::shared_ptr<Texture> source,
               std::vector<uint8_t> destination,
               std::optional<IRect> source_region = std::nullopt,
               IPoint destination_origin = {},
               std::string label = "");
```

impeller/renderer/context.h
bdero: Yeah, we could also just return an `std::vector<uint8_t>` and just use move semantics to avoid copies when it makes sense, especially since we're getting pretty good help from clang tidy in that department at this point.

jonahwilliams:
Moving to WIP while I iterate on some stuff

impeller/core/device_buffer.cc
```diff
@@ -69,4 +70,11 @@ const DeviceBufferDescriptor& DeviceBuffer::GetDeviceBufferDescriptor() const {
   return OnCopyHostBuffer(source, source_range, offset);
 }
 
+std::unique_ptr<const uint8_t[]> DeviceBuffer::CopyToUnqiueBuffer() const {
+  auto size = GetDeviceBufferDescriptor().size;
+  auto raw_data = static_cast<uint8_t*>(malloc(size));
+  memcpy(raw_data, OnGetContents(), size);
+  return std::unique_ptr<const uint8_t[]>(raw_data);
```
jonahwilliams, line 77: This is leaking right now because I need to give it a release proc

impeller/renderer/context.h
bdero: This kind of thing would jive well with a non-DeviceBuffer variant of `AddCopy` I mentioned [above](https://github.com/flutter/engine/pull/46394#discussion_r1341809385). Might as well eat an extra copy on the host for now and fix both the copies later. :)

impeller/core/device_buffer.cc
bdero: Rather than lugging around a device allocation, you could just eat an extra host copy now and we can repair this later when we [fix up the blits](https://github.com/flutter/engine/pull/46394#discussion_r1341864534).

impeller/renderer/context.cc
bdero: Made a follow-up issue about this here: https://github.com/flutter/flutter/issues/135783

jonahwilliams:
I think I might find a way to move this to texture.h too.

jonahwilliams:
actually it looks like only vulkan textures have a weak ptr to their context, gles and mtl textures do not. I could add more weak ptrs, but it would really only be in service of making the API slightly more ergonomic.

jonahwilliams:
The failing tests are entirely mock based. We should migrate them to dart:ui tests once flutter_tester supports impeller, but until then I'll leave the old readback approach.

zanderso:
@jonahwilliams What's the status of this PR?

jonahwilliams:
This is still a WIP

jonahwilliams:
Closing as I don't have time to work on this right now

PR #46422, opened 2023-09-29
[Impeller] Run DisplayList render tests.

flar:
Revisiting this project as it may help with [attribute] X [rendering op] matrix testing of Impeller. It's not an exhaustive combinatorial test of all combinations of rendering attributes, but it attempts each op with each attribute both default and non-default.

For Impeller it only tests that the rendering produced output (i.e wasn't a NOP due to missing implementation), that the output changed if the given attribute was supposed to change the output, and that it rendered within the computed bounds that the rest of the engine will be assuming.

These changes will not result in any automatic CI testing of Impeller as is since they are only run on the Metal backend which is not enabled by default. Running `out/host_[variant]/display_list_rendertests --enable-metal` manually on a Mac host, though, will run the tests through Impeller and there will be a lot of error output. Also, there is one known crash as reported in https://github.com/flutter/flutter/issues/135766

### DIFF
--- a/display_list/testing/dl_test_surface_metal.cc
+++ b/display_list/testing/dl_test_surface_metal.cc
@@ -3,6 +3,9 @@
 // found in the LICENSE file.
 
 #include "flutter/display_list/testing/dl_test_surface_metal.h"
+#include "flutter/impeller/display_list/dl_dispatcher.h"
+#include "flutter/impeller/display_list/dl_image_impeller.h"
+#include "flutter/impeller/typographer/backends/skia/typographer_context_skia.h"
 
 #include "third_party/skia/include/core/SkCanvas.h"
 #include "third_party/skia/include/core/SkSurface.h"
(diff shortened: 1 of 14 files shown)


display_list/testing/dl_test_surface_metal.cc
```diff
@@ -54,5 +57,71 @@ std::shared_ptr<DlSurfaceInstance> DlMetalSurfaceProvider::MakeOffscreenSurface(
...
+    screenshot_->WriteToPNG(path);
+  }
+
+ private:
+  std::unique_ptr<MetalScreenshot> screenshot_;
+  const uint32_t* addr_;
+  const uint32_t ints_per_row_;
+};
+
+sk_sp<DlPixelData> DlMetalSurfaceProvider::ImpellerSnapshot(
```
jonahwilliams, line 87: I don't think you need to use the metal screenshotter here. provided you have access to an impeller context, you should be able to read back the texture data without worrying about timing issues either.

See https://github.com/flutter/engine/blob/main/lib/ui/painting/image_encoding_impeller.cc#L108-L129 


I'm also adding a helper method to do this but its still WIP:  https://github.com/flutter/engine/blob/main/lib/ui/painting/image_encoding_impeller.cc#L108-L129

jonahwilliams: you'd need to use something like fml::WaitableEvent as a latch as the submission callback will be invoked later on a different thread

flar: Those point to the same code. When will the helper method be available? Is it worth using ScreenShotter until then as a temporary measure?

jonahwilliams: Oops! helper is https://github.com/flutter/engine/pull/46394 . Uh, won't be done today for sure. But you can use the code from image_encoding_impeller today and then this works on all backends.

flar: I looked into bringing over the changes there and there were quite a few dependencies I had to resolve. Would it be possible to just put this in as is using the ScreenShotter and then shift over to the new convenience mechanism when that lands?

Even though this doesn't get run by default right now, it does identify a number of Impeller bugs related to ImageFilter bounds and ignoring ImageFilters, so it would be useful to have this tool in the tree so that we can start fixing those problems. When the failing tests go to zero we can look at how to get it running under CI as a regression test. Many of the failures are dups (if we don't do something under one rendering op then we don't do it on many of them) so the number of problems that actually need fixing is less than the number of failing cases. I looked at the output and saw 3 or 4 problems that kept triggering over and over, but that's not an exhaustive search...

jonahwilliams: I think that is  fine. we should probably just refactor the screenshotter to have implementations for all platforms

flar:
I tagged 4 people for the next round of reviews for various reasons:

- Jonah already looked at it before
- Brandon might end up being the primary engineer for many of the bugs
- Matan because this might factor into our testing plans
- Chinmay because the others have a bunch of meetings on their plates... ;)

jonahwilliams:
Why do we specify the Impeller/Skia cases separately? Could we use the DL to specify both cases?

flar:
> Why do we specify the Impeller/Skia cases separately? Could we use the DL to specify both cases?

All tests are specified in 2 forms - SkCanvas calls and DlCanvas calls - we need both of these because one of the goals of the tests is to compare the results of rendering DLCanvas through Skia and rendering SkCanvas through Skia. That was one of the original goals of the tests. Basically it compares "rendering a thick line with a blur" through SkCanvas calls and "rendering a thick line with a blur" in DlCanvas calls and verifies that both render the same thing when using Skia as a backend.

To add Impeller, I'm not adding a new specification of the test code, I reuse the DL code, but I have to run it against Impeller. So, we have 2 specs for everything - one in SkCanvas and one in DlCanvas - and then we have 3 executors for them - one is SkCanvas rendering with Skia, another is DlCanvas rendering with Skia, and a 3rd is DlCanvas rendering with Impeller.

We could eliminate the SkCanvas rendering code at some point, but we'd have to shift to something like a golden image suite to do that. Essentially, for now, we are using a cross-check with Skia as a golden image suite for "rendering DL through Skia".

The new nascent Impeller support isn't comparing output to Skia or goldens, so it is mostly just verifying independently measurable things like "that should have caused a change" or "that went out of bounds". Even with those basic tests (when run manually) it is finding faults in Impeller.

jonahwilliams (approved):
LGTM

PR #46430, opened 2023-09-29
Add an AndroidManifest.xml flag to disable ImageReader backed Platform Views

johnmccutchan:
An internal customer is seeing flickering on some high FPS phones.

While we try and reproduce / gain more knowledge of this problem this PR adds a AndroidManifest.xml flag to disable their usage.

The boolean flag is:
```
"io.flutter.embedding.android.DisableImageReaderPlatformViews"
```

Adding the following to your manifest will disable their usage:

```
<meta-data
  android:name="io.flutter.embedding.android.DisableImageReaderPlatformViews"
  android:value="true" />
```

### DIFF
--- a/common/settings.h
+++ b/common/settings.h
@@ -218,6 +218,9 @@ struct Settings {
   bool enable_impeller = false;
 #endif
 
+  // Indicates if image reader backed platform views are disabled.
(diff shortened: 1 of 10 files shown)


jonahwilliams (approved):
RSLGTM

PR #46433, opened 2023-09-30
[Impeller] Implement a `MockGLES`, that provides trampolines for `ProcGLESTable`

matanlurey:
Unblocks testing https://github.com/flutter/flutter/issues/135715. See `README.md` for details.

I made one non-testing change to the proc_table itself, which is if a function call will fail, we will print out what is about to fail in the validation log. It was useful when debugging the test creation itself, and it's only enabled when GL call checks are enabled anyway.

@gaaclarke I originally implemented it with `FML_THREAD_LOCAL`, but figured doing a global lock essentially was the same thing, and would prevent parallel test runs from stepping on each-other in weird ways?

/cc @chinmaygarde for visibility.


impeller/renderer/backend/gles/proc_table_gles.h
```diff
@@ -75,6 +76,12 @@ struct GLProc {
   auto operator()(Args&&... args) const {
 #ifdef IMPELLER_DEBUG
     AutoErrorCheck error(error_fn, name);
+    // We check for the existence of extensions, and reset the function pointer
+    // but it's still called unconditionally below, and will segfault. This
+    // validation log will at least give us a hint as to what's going on.
+    if (!IsAvailable()) {
```
gaaclarke, line 82: If it's going to crash, it's probably best to crash here, no?

impeller/renderer/backend/gles/proc_table_gles.h
```diff
@@ -85,7 +92,6 @@ struct GLProc {
   constexpr bool IsAvailable() const { return function != nullptr; }
 
   void Reset() {
-    name = nullptr;
```
gaaclarke, line 88: Can we shift name to a string_view?  A raw pointer is unclear what the memory rules are.

impeller/renderer/backend/gles/test/mock_gles.cc
```diff
@@ -0,0 +1,121 @@
...
+
+// OpenGLES is not thread safe.
+//
+// This mutex is used to ensure that only one test is using the mock at a time.
+static std::mutex g_test_lock;
+
+static std::weak_ptr<MockGLES> g_mock_gles;
+
+// Has friend visibility into MockGLES to record calls.
+void recordCall(const char* name) {
```
gaaclarke, line 22:
```suggestion
void RecordGLCall(const char* name) {
```

It needs to be pascal case and `recordCall` is kind of a generic name to be placing in the testing namespace, so best to make it a bit more apropos.

impeller/renderer/backend/gles/test/mock_gles.cc
```diff
@@ -0,0 +1,121 @@
...
+  recordCall("PopDebugGroupKHR");
+}
+
+void glPushDebugGroupKHR(GLenum source,
+                         GLuint id,
+                         GLsizei length,
+                         const GLchar* message) {
+  recordCall("PushDebugGroupKHR");
+}
+
```
gaaclarke, line 87: nit: For these functions I found it nice to do a static assert that they match the same signature of the function they are replacing.

Check out https://github.com/flutter/engine/blob/675a7e456d49fa5f0081c3f1fbf5f5f00187c989/shell/platform/embedder/tests/embedder_vk_unittests.cc#L85

gaaclarke (approved):
Code lgtm, docs lgtm, I just have some notes.

impeller/renderer/backend/gles/proc_table_gles.h
matanlurey: Yeah that's a good point. Changed to a FML_CHECK.

impeller/renderer/backend/gles/proc_table_gles.h
matanlurey: I actually removed resetting the name, because it means we lose context on error messages.

Did you still want it changed?

matanlurey (commented):
Thanks for the review!

impeller/renderer/backend/gles/test/mock_gles.cc
```diff
@@ -0,0 +1,121 @@
...
+    return reinterpret_cast<void*>(&glGetIntegerv);
+  } else if (strcmp(name, "glGetError") == 0) {
+    return reinterpret_cast<void*>(&glGetError);
+  } else {
+    return reinterpret_cast<void*>(&glDoNothing);
+  }
+};
+
+MockGLES::MockGLES() : proc_table_(kMockResolver) {
+  g_test_lock.lock();
```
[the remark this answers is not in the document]
gaaclarke: You could use a static factory method with a `try_lock`.

impeller/renderer/backend/gles/proc_table_gles.h
gaaclarke: Up to you, string_view's are still probably preferable because they explicitly say they aren't taking ownership, while with c strings it's ambiguous.

gaaclarke:
The image is 487px wide and if the amplitude of x sin function is dropped to 450, the image will not blink.  Looks like it is winking out of existence at -487px.

impeller/renderer/backend/gles/test/mock_gles.cc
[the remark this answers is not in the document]
matanlurey: Great, works great!

impeller/renderer/backend/gles/proc_table_gles.h
matanlurey: Ack. I am going to file a follow-up issue to keep the scope small:

https://github.com/flutter/flutter/issues/135922

matanlurey:
Changed made. I'd like one more look from @gaaclarke and @jonahwilliams before submitting. Thanks!

gaaclarke (approved):

jonahwilliams (approved):
LGTM

PR #46434, opened 2023-09-30
[Impeller] Make some Open GL errors non-fatal, check in debug mode (not unopt).

matanlurey:
Closes https://github.com/flutter/flutter/issues/135767.

### DIFF
--- a/impeller/BUILD.gn
+++ b/impeller/BUILD.gn
@@ -38,10 +38,6 @@ config("impeller_public_config") {
     defines += [ "IMPELLER_TRACE_ALL_GL_CALLS" ]
   }
 
-  if (impeller_error_check_all_gl_calls) {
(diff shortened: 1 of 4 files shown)


chinmaygarde (approved):

jonahwilliams (approved):
LGTM

zanderso:
There are a few failures in the devicelab from this change. (In bringup so the tree is not closed.) Here is an example: https://ci.chromium.org/ui/p/flutter/builders/staging/Linux_pixel_7pro%20animated_blur_backdrop_filter_perf_opengles__timeline_summary/714/overview

jonahwilliams:
This is likely a misconfiguration in the texture to texture blit, the only places where glBlitFramebuffer is used. Thankfully we have a capabilities check for this and can disable it: https://github.com/flutter/engine/blob/main/impeller/renderer/capabilities.cc#L32C8-L32C37

PR #46443, opened 2023-09-30
[Impeller] disable texture to texture blit in GLES.

jonahwilliams:
These are failing with a GL_INVALID_OPERATION.


zanderso (approved):
rslgtm

PR #46230, opened 2023-09-23
[macOS] Synchronise modifiers from mouse events for RawKeyboard

knopp:
Fixes https://github.com/flutter/flutter/issues/135349

This has been done for FlutterEmbedderKeyResponder in https://github.com/flutter/engine/pull/37870, but has not been implemented for FlutterChannelKeyResponder, which results in RawKeyboard being out of sync with HardwareKeyboard.

## Pre-launch Checklist

- [x] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [x] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [x] I read and followed the [Flutter Style Guide] and the [C++, Objective-C, Java style guides].
- [x] I listed at least one issue that this PR fixes in the description above.
- [x] I added new tests to check the change I am making or feature I am adding, or the PR is [test-exempt]. See [testing the engine] for instructions on writing and running engine tests.
- [x] I updated/added relevant documentation (doc comments with `///`).
- [x] I signed the [CLA].
- [x] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/flutter/wiki/Tree-hygiene#overview
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[test-exempt]: https://github.com/flutter/flutter/wiki/Tree-hygiene#tests
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[C++, Objective-C, Java style guides]: https://github.com/flutter/engine/blob/main/CONTRIBUTING.md#style
[testing the engine]: https://github.com/flutter/flutter/wiki/Testing-the-engine
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/wiki/Chat

PR #46461, opened 2023-10-02
[macOS] Fix tests failing on Sonoma

knopp:
1. Using arbitrary struct passed as const reference  to `OCMStub` now fails in OCMock. Down the line this will result with `object_getClass` being called in [`OCMArg.m`](https://github.com/erikdoe/ocmock/blob/master/Source/OCMock/OCMArg.m#L129-L133) with the address of the reference, which is not a valid class instance. This seems to have worked pre-sonoma, but it seems like a weird thing to rely on.

2. `NSResponder` mock can not be set to view controller anymore. The controller will try to access an ivar of the `NSResponder`, but mocked responder does not have the ivar of original objects which will result on invalid selector being called on a `NSMutableArray` one of the ivar of mock objects. Solution for this is to inherit from `NSResponder` and forward calls to mocked object.

3. Adding `flutter::kModifierFlagShiftLeft` to a modifier flag containing `kCGEventFlagMaskShift`. The assertion was introduced in https://github.com/flutter/engine/pull/46230 but i missed the test failure because of the problems above.

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/shell/platform/darwin/macos/framework/Source/FlutterKeyboardManagerTest.mm
+++ b/shell/platform/darwin/macos/framework/Source/FlutterKeyboardManagerTest.mm
@@ -276,7 +276,9 @@ void clearEvents(std::vector<FlutterKeyEvent>& events) {
   OCMStub([viewDelegateMock onTextInputKeyEvent:[OCMArg any]])
       .andCall(self, @selector(handleTextInputKeyEvent:));
   OCMStub([viewDelegateMock getBinaryMessenger]).andReturn(_messengerMock);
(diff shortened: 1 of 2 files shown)


knopp:
Btw, Chromium seems to have patched the OCMArg problem: https://chromium-review.googlesource.com/c/chromium/src/+/3011651

cbracken (approved):

Thanks for the additional fix for the missing shift modifier.

cbracken:
Thanks again for the fix; this is additional justification for us to find time to dig into the non-hermetic test situation in the macOS embedder tests so we can re-enable the presubmits.

PR #46471, opened 2023-10-02
[ios] Link PlatformView back to semantics tree

cyanglaz:
The PlatformView does not have a semantics container when added to semantics tree, this PR gives it a semantics container to ensure accessibility traversal works.

This fixes https://github.com/flutter/flutter/issues/135504, which is a regression of https://github.com/flutter/engine/commit/b8332e3c4870512a941d685d7d375b1d2d62983c

Before https://github.com/flutter/engine/commit/b8332e3c4870512a941d685d7d375b1d2d62983c, the traversal works because the PlatformView is added to the accessibilityElements of the FlutterPlatFormViewSemanticsContainer, which implicitly made the FlutterPlatFormViewSemanticsContainer as the PlatformVIew's AccessibilityContainer. 

Now we use the PlatformVIew as the nativeAccessibility of the  FlutterPlatFormViewSemanticsContainer, we need to expilicitly set the container. 

This needs to be cherry-picked since the commit caused the regression was cherry-picked in https://github.com/flutter/engine/commit/b8332e3c4870512a941d685d7d375b1d2d62983c

## Pre-launch Checklist

- [x] I signed the [CLA].


shell/platform/darwin/ios/framework/Source/SemanticsObject.mm
```diff
@@ -882,12 +880,6 @@ - (id)nativeAccessibility {
   return _platformView;
 }
 
-#pragma mark - UIAccessibilityContainer overrides
-
-- (NSArray*)accessibilityElements {
-  return @[ _platformView ];
-}
```
cyanglaz, line 889: We implemented nativeAccessibility, this is not necessary.

shell/platform/darwin/ios/framework/Source/FlutterPlatformViews_Internal.h
```diff
@@ -234,6 +235,8 @@ class FlutterPlatformViewsController {
   // returns nil.
   UIView* GetPlatformViewByID(int64_t view_id);
 
+  FlutterTouchInterceptingView* GetFlutterTouchInterceptingViewByID(int64_t);
```
stuartmorgan, line 238: Needs a documentation comment.

shell/platform/darwin/ios/framework/Source/FlutterPlatformViews.mm
```diff
@@ -957,6 +962,7 @@ @implementation FlutterTouchInterceptingView {
   fml::scoped_nsobject<DelayingGestureRecognizer> _delayingRecognizer;
   FlutterPlatformViewGestureRecognizersBlockingPolicy _blockingPolicy;
   UIView* _embeddedView;
+  NSObject* _flutterAccessibilityContainer;
```
stuartmorgan, line 965: Needs a comment.

shell/platform/darwin/ios/framework/Source/FlutterPlatformViews.mm
```diff
@@ -1035,6 +1041,14 @@ - (void)touchesCancelled:(NSSet<UITouch*>*)touches withEvent:(UIEvent*)event {
 - (void)touchesEnded:(NSSet*)touches withEvent:(UIEvent*)event {
 }
 
+- (void)setFlutterAccessibilityContainer:(NSObject*)flutterAccessibilityContainer {
```
stuartmorgan, line 1044: Why is this being declared as an ivar and manual getter/setter methods, rather than declared as a property, with synthesized methods?

shell/platform/darwin/ios/framework/Source/FlutterPlatformViewsTest.mm
```diff
@@ -19,6 +19,10 @@
 static FlutterPlatformViewsTestMockPlatformView* gMockPlatformView = nil;
 const float kFloatCompareEpsilon = 0.001;
 
+@interface FlutterTouchInterceptingView (Test)
+- (id)accessibilityContainer;
```
stuartmorgan, line 23: Please make a `_Test.h` file instead of re-declaring methods directly in test files.

shell/platform/darwin/ios/framework/Source/SemanticsObject.h
```diff
@@ -19,6 +19,14 @@ constexpr float kScrollExtentMaxForInf = 1000;
 @class FlutterCustomAccessibilityAction;
 @class FlutterPlatformViewSemanticsContainer;
 
+@protocol FlutterSemanticsProtocol <NSObject>
+// Adding functionalities for any object to assign accessibilityContainer.
+// For example, this allows the PlatformView be able to link back to the accessibility tree.
```
stuartmorgan, line 24: I'm confused by what this property is actually for. I see where it's set, but I don't see anything reading it except in tests. Am I missing some non-test behavior that's influenced by this?

stuartmorgan (changes requested):

shell/platform/darwin/ios/framework/Source/FlutterPlatformViews.mm
cyanglaz: We don't need the getter, we only need the setter so the "hidden" getter, accessibilityContainer used by UIAccessibility can return the correct value.

shell/platform/darwin/ios/framework/Source/SemanticsObject.h
cyanglaz: It is read in `FlutterTouchInterceptingView`: accessibilityContainer, it is a UIAccessibilty property that is read by the UIKit.

This protocol was created so that we can call this method to `FlutterTouchInterceptingView`. I initially added the method to the `FlutterTouchInterceptingView` so that `FlutterTouchInterceptingView` can access the container in its `accessibiltiyContainer` method.  

Then I discovered that SemanticsObjectTests is ARC, so we couldn't not include `FlutterTouchInterceptingView` (which uses MRC features) in the header of `SemanticsObjectTests`

This does make the code more confusing.

cyanglaz:
@stuartmorgan I removed the protocol to reduce confusion, I also refactored the platform view related tests to a separate file + other review comments. PTAL

shell/platform/darwin/ios/framework/Source/FlutterTouchInterceptingView+Tests.h
stuartmorgan, line 1: `..._Test.h` seems to be the established convention in this directory.

shell/platform/darwin/ios/framework/Source/FlutterPlatformViewsTest.mm
stuartmorgan: This entire redeclaration block should be removed in favor of including the test header.

shell/platform/darwin/ios/framework/Source/FlutterPlatformViews_Internal.h
```diff
@@ -424,6 +432,10 @@ class FlutterPlatformViewsController {
 
 // Get embedded view
 - (UIView*)embeddedView;
+
+// The FlutterTouchInterceptingView assigns `flutterAccessibilityContainer` as its
```
stuartmorgan, line 436: Instance methods don't need to say that they are about the class they are on; this can just say "Sets `flutterAccessibilityContainer` as this view's `accessibilityContainer`."

shell/platform/darwin/ios/framework/Source/SemanticsObjectTest_mrc.mm
```diff
@@ -0,0 +1,89 @@
...
+  FlutterPlatformViewSemanticsContainer* container =
+      [[FlutterPlatformViewSemanticsContainer alloc] initWithBridge:bridge
+                                                                uid:1
+                                                       platformView:platformView];
+  XCTAssertEqualObjects(platformView.accessibilityContainer, container);
+  XCTAssertEqual(platformView.retainCount, 2u);
+  // Check if there's no more strong references to `platformView` after container and platformView
+  // are released.
+  [container release];
+  XCTAssertEqual(platformView.retainCount, 1u);
```
stuartmorgan, line 86: This is fragile to implementation details of the code using the platform view. It would help to at least wrap all the code between the declaration and here in an autorelease pool.

stuartmorgan (changes requested):

cyanglaz:
@stuartmorgan Updated, PTAL!

shell/platform/darwin/ios/framework/Source/FlutterPlatformViews_Internal.h
```diff
@@ -230,10 +231,17 @@ class FlutterPlatformViewsController {
...
+  // a `FlutterPlatformView` object associated with the view_id cannot be found, the method
   // returns nil.
   UIView* GetPlatformViewByID(int64_t view_id);
 
+  // Returns the `FlutterTouchInterceptingView` with the view_id.
+  //
+  // If the `FlutterPlatformViewsController` does not contain any `FlutterPlatformView` object or
+  // a `FlutterPlatformView` object associated with the view_id cannot be found, the method
+  // returns nil.
+  FlutterTouchInterceptingView* GetFlutterTouchInterceptingViewByID(int64_t);
```
stuartmorgan, line 243: Nit: I missed that this parameter doesn't have a name; it should be `int64_t view_id`.

stuartmorgan (approved):
LGTM with one last nit.

PR #46478, opened 2023-10-02
Restrict ImageReader backed Platform Views to Android 33+

johnmccutchan:
This is necessary to be able to access the fence associated with each frame that is produced. This API is not available below 33.


shell/platform/android/io/flutter/embedding/engine/renderer/FlutterRenderer.java
```diff
@@ -371,24 +376,36 @@ public void pushImage(Image image) {
...
     public Image acquireLatestImage() {
       Image r;
       synchronized (this) {
         r = this.image;
         this.image = null;
       }
+      if (r != null) {
+        try {
+          SyncFence fence = r.getFence();
+          boolean signaled = fence.awaitForever();
```
jonahwilliams, line 401: 😢 

Not necessary to do now, but we _might_ be able to get the Vulkan drive to do the waiting via something like https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkExternalFenceHandleTypeFlagBits.html

jonahwilliams (approved):
LGTM

shell/platform/android/io/flutter/embedding/engine/renderer/FlutterRenderer.java
johnmccutchan: I'm not sure it is possible to convert a SyncFence into a file descriptor (although it is possible to serialize into a Parceable and manually extracting the fd -- just unclear on how stable the parceable format is).

jonahwilliams: Yeah I'm not sure if that is exactly the right API, just that I thought there was something like that

PR #46489, opened 2023-10-03
[Impeller] Fix framebuffer blend UVs.

bdero:
Fixes b/303120488.

There's no need to use the `GetCoverageUVs` utility in this case. Correct usage here would require the geometry to match the coverage rectangle of the snapshot in pass space with no further transformations to the geometry.
We should just render the snapshot texture as-is with simple 0/1 UVs and the snapshot's transform.

Before:

https://github.com/flutter/engine/assets/919017/2908573e-b7cf-430f-9317-92c8aac6ade4

Screenshot 2023-10-02 at 9 57 10 PM

After:

https://github.com/flutter/engine/assets/919017/74901c9d-cb26-43c8-b963-a28bffbd5a44

Screenshot 2023-10-02 at 10 05 14 PM


jason-simmons (approved):

jonahwilliams (approved):
Lgtm

PR #46526, opened 2023-10-03
Use GdkEvent methods to access values, direct access is removed in GTK4.

robert-ancell:
This should have no effect on the current implementation, and will be one less thing needed when we migrate to GTK4.


cbracken (approved):
lgtm! Thanks!

stuartmorgan:

PR #46597, opened 2023-10-05
[Impeller] Track clip coverage per-pass when not collapsing.

bdero:
Resolves https://github.com/flutter/flutter/issues/135916.

Big credits to @gaaclarke for the repros and investigation around this.

Moves back to tracking clip coverage per-pass when not collapsing. This is especially important for SaveLayers with MatrixImageFilters, which can transform the rendered layer texture and thereby move elements inside or outside of existing parent clips.

This should have little to no performance impact, since we should already be generating correct minimal subpass textures based on correct subpass coverage computation.

Before:
Screenshot 2023-10-05 at 1 25 29 PM

https://github.com/flutter/engine/assets/919017/9284f055-bee1-40cd-8b7e-f478b00d01da

After:
Screenshot 2023-10-05 at 1 24 17 PM

https://github.com/flutter/engine/assets/919017/0fbd1f96-8920-472c-bb0e-4414187fc72d

### DIFF
--- a/impeller/aiks/aiks_unittests.cc
+++ b/impeller/aiks/aiks_unittests.cc
@@ -3639,7 +3639,26 @@ TEST_P(AiksTest, MatrixImageFilterDoesntCullWhenTranslatedFromOffscreen) {
       .image_filter = std::make_shared<MatrixImageFilter>(
           Matrix::MakeTranslation({300, 0}), SamplerDescriptor{}),
   });
-  canvas.DrawCircle({-300, 0}, 100, {.color = Color::White()});
+  canvas.DrawCircle({-300, 0}, 100, {.color = Color::Green()});
+  canvas.Restore();
+
+  ASSERT_TRUE(OpenPlaygroundHere(canvas.EndRecordingAsPicture()));
+}
+
+// Render a white circle at the top left corner of the screen.
+TEST_P(AiksTest,
+       MatrixImageFilterDoesntCullWhenScaledAndTranslatedFromOffscreen) {
+  Canvas canvas;
+  canvas.Scale(GetContentScale());
+  canvas.Translate({100, 100});
+  // Draw a circle in a SaveLayer at -300, but move it back on-screen with a
+  // +300 translation applied by a SaveLayer image filter.
+  canvas.SaveLayer({
+      .image_filter = std::make_shared<MatrixImageFilter>(
+          Matrix::MakeTranslation({300, 0}) * Matrix::MakeScale({2, 2, 2}),
+          SamplerDescriptor{}),
+  });
+  canvas.DrawCircle({-150, 0}, 50, {.color = Color::Green()});
   canvas.Restore();
 
   ASSERT_TRUE(OpenPlaygroundHere(canvas.EndRecordingAsPicture()));
--- a/impeller/aiks/aiks_unittests.cc
+++ b/impeller/aiks/aiks_unittests.cc
@@ -3619,15 +3619,27 @@ TEST_P(AiksTest, MatrixImageFilterMagnify) {
   canvas.Translate({600, -200});
   canvas.SaveLayer({
       .image_filter = std::make_shared<MatrixImageFilter>(
-          Matrix{
-              2, 0, 0, 0,  //
-              0, 2, 0, 0,  //
-              0, 0, 2, 0,  //
-              0, 0, 0, 1   //
-          },
-          SamplerDescriptor{}),
+          Matrix::MakeScale({2, 2, 2}), SamplerDescriptor{}),
   });
-  canvas.DrawImage(image, {0, 0}, Paint{.color = Color(1.0, 1.0, 1.0, 0.5)});
+  canvas.DrawImage(image, {0, 0},
+                   Paint{.color = Color::White().WithAlpha(0.5)});
+  canvas.Restore();
+
+  ASSERT_TRUE(OpenPlaygroundHere(canvas.EndRecordingAsPicture()));
+}
+
+// Render a white circle at the top left corner of the screen.
+TEST_P(AiksTest, MatrixImageFilterDoesntCullWhenTranslatedFromOffscreen) {
+  Canvas canvas;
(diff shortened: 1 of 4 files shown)


impeller/aiks/aiks_unittests.cc
```diff
@@ -3619,15 +3619,27 @@ TEST_P(AiksTest, MatrixImageFilterMagnify) {
...
+  });
+  canvas.DrawImage(image, {0, 0},
+                   Paint{.color = Color::White().WithAlpha(0.5)});
+  canvas.Restore();
+
+  ASSERT_TRUE(OpenPlaygroundHere(canvas.EndRecordingAsPicture()));
+}
+
+// Render a white circle at the top left corner of the screen.
+TEST_P(AiksTest, MatrixImageFilterDoesntCullWhenTranslatedFromOffscreen) {
```
gaaclarke, line 3632: Do we need a test that does a scale as well?  I know that translation only matrices are treated as special cases in the code.  (which in general seems risky)

impeller/entity/entity_pass.cc
```diff
@@ -198,12 +198,10 @@ std::optional<Rect> EntityPass::GetSubpassCoverage(
...
-  // If the filter graph transforms the basis of the subpass, then its space
-  // has deviated too much from the parent pass to safely intersect with the
-  // pass coverage limit.
-  coverage_limit =
-      (image_filter && !image_filter->IsTranslationOnly() ? std::nullopt
-                                                          : coverage_limit);
+  // If the subpass has an image filter, then its coverage space may deviate
+  // from the parent pass and make intersecting with the pass coverage limit
+  // unsafe.
+  coverage_limit = image_filter ? std::nullopt : coverage_limit;
```
gaaclarke, line 204: Is this just punting on the coverage test if there is an image_filter?

impeller/entity/entity_pass.cc
```diff
@@ -198,12 +198,10 @@ std::optional<Rect> EntityPass::GetSubpassCoverage(
   std::shared_ptr<FilterContents> image_filter =
       subpass.delegate_->WithImageFilter(Rect(), subpass.xformation_);
 
-  // If the filter graph transforms the basis of the subpass, then its space
-  // has deviated too much from the parent pass to safely intersect with the
-  // pass coverage limit.
-  coverage_limit =
-      (image_filter && !image_filter->IsTranslationOnly() ? std::nullopt
-                                                          : coverage_limit);
+  // If the subpass has an image filter, then its coverage space may deviate
```
gaaclarke, line 201: Can you add a docstring to `GetSubpassCoverage` please.

impeller/entity/entity_pass.cc
```diff
@@ -641,6 +639,14 @@ EntityPass::EntityResult EntityPass::GetEntityForElement(
     auto subpass_capture = capture.CreateChild("EntityPass");
     subpass_capture.AddRect("Coverage", *subpass_coverage, {.readonly = true});
 
+    // Start non-collapsed subpasses with a fresh clip coverage stack limited by
+    // the subpass coverage. This is important because image filters applied to
+    // save layers may transform the subpass texture after its rendered, causing
```
gaaclarke, line 644:
```suggestion
    // save layers may transform the subpass texture after it's rendered, causing
```

gaaclarke:
Oh man, you ran into https://github.com/flutter/flutter/issues/135511.   You are going to have to change the color of your circle so the golden test passes.

impeller/aiks/aiks_unittests.cc
bdero: Good idea, I added one that scales.

gaaclarke:
I double checked this against https://github.com/flutter/flutter/issues/126583 and it doesn't address this that issue.  That makes this the 2nd bug we've found and fixed in service of fixing that bug haha.

impeller/entity/entity_pass.cc
bdero: It just stops us from trying to use the screen area as a limit for coverage, since ImageFilters can wind up moving things around. We could make this more restrictive in the future by adding the ability to pull the transform from filter chains.

impeller/entity/entity_pass.cc
bdero: Beefed up the docstring.

impeller/entity/entity_pass.cc
bdero: Done. 👍

bdero:
Changed the color of the circle and added a scaling version. :)

impeller/entity/entity_pass.cc
gaaclarke: > pull the transform from filter chains.

Yea, that's what I was thinking of when writing that doc.

impeller/entity/entity_pass.h
```diff
@@ -142,7 +142,16 @@ class EntityPass {
   void SetEnableOffscreenCheckerboard(bool enabled);
 
   //----------------------------------------------------------------------------
-  /// @brief  Get the coverage of an unfiltered subpass.
+  /// @brief  Computes the coverage of a given subpass. This is used to
```
gaaclarke, line 145: Can you please include the coordinate space of the returned rect?  I think it should be relative to it's parent subpass in pixels?

gaaclarke (changes requested):
Code and tests LGTM.  I would like to see us document coordinate systems in our docstrings.  I found this code frustrating to work with because I didn't know what should be accounted for where.

impeller/entity/entity_pass.h
bdero: Done.

> I think it should be relative to it's parent subpass in pixels?

Actually, the output is in screen space. :)

After writing https://github.com/flutter/engine/pull/46524#issuecomment-1750194345, I also decided to do another iteration on the `GetCoverage` docstrings.

gaaclarke: Thank you! 🙏

gaaclarke (approved):

PR #46610, opened 2023-10-06
[Impeller] Use triangle strips to dedupe rectangle vertices.

bdero:
Remove a couple of duplicate verts in a bunch of situations.

Pedantic improvements: "Brandon, why are you like this" edition.

### DIFF
--- a/impeller/entity/contents/content_context.cc
+++ b/impeller/entity/contents/content_context.cc
@@ -192,6 +192,11 @@ ContentContext::ContentContext(
       .sample_count = SampleCount::kCount4,
       .color_attachment_pixel_format =
           context_->GetCapabilities()->GetDefaultColorFormat()};
+  auto options_trianglestrip = ContentContextOptions{
+      .sample_count = SampleCount::kCount4,
+      .primitive_type = PrimitiveType::kTriangleStrip,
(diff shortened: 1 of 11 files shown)


jonahwilliams (commented):
So if you change these from triangle, the default constructed PSO won't match anymore for these. Though if we're going to unconditionally use triangle strip, you could probably change the initialization code to specify that geometry instead?

bdero:
> Though if we're going to unconditionally use triangle strip, you could probably change the initialization code to specify that geometry instead?

Done.

I also got the ordering wrong on some of these. Let's see those new goldens.

jonahwilliams (commented):
You will probably need to update default_options L832 content_context.h

bdero:
I'm working on a solution for tracking different prototype defaults for different pipeline sets here: https://github.com/flutter/engine/pull/46751

bdero:
This should be good now. This is using the new solution for differing defaults and the goldens are looking good. :)

bdero:
The previous solution I tried to employ in this patch didn't work as intended because of the mismatching default lookup in content_context.h routines.

jonahwilliams (approved):
LGTM

PR #46642, opened 2023-10-07
[Impeller] Add IMPELLER_DEBUG around group.

jonahwilliams:
These tracing functions aren't free!


chinmaygarde (approved):
I mentioned this in a previous review. But how about just patching the proc table to make these no-ops when IMPELLER_DEBUG is unset? For instance, the way we do it for the framebuffer discard extensions.

jonahwilliams:
Done!

jonahwilliams:
actually wont work since I also ifdef out the cmd label. RIP

PR #46683, opened 2023-10-09
Switch to Chrome For Testing instead of Chromium

eyebrowsoffire:
This switches over to using Chrome for Testing instead of Chromium. This requires some changes from the recipes repo (https://flutter-review.googlesource.com/c/recipes/+/51482) in order to coordinate the change in filestructure on the mac versions.


lib/web_ui/dev/common.dart
```diff
@@ -33,6 +32,7 @@ abstract class PlatformBinding {
     }
     if (io.Platform.isMacOS) {
       if (environment.isMacosArm) {
+        print('MACOS ARM!');
```
yjbanov, line 35: Do we need this print?

lib/web_ui/dev/browser_lock.yaml
```diff
@@ -1,23 +1,7 @@
...
-  # The OS names here must match what recipe Python expression
-  # `self.m.platform.name.capitalize()` evaluates to. See:
-  #
-  # recipe_modules/web_util/api.py
-  Linux: 1181205
-  Mac: 1181205
-  Mac_Arm: 1181212
-  Win: 1181217
-  version: '117.0' # CIPD tag for the above Build IDs. Normally "ChromeMajorVersion.UploadAttempt". ;)
+  version: '118.0.5993.70'
```
yjbanov, line 4: \o/

yjbanov (approved):

lib/web_ui/dev/browser_roller.dart
```diff
@@ -294,9 +286,9 @@ class _BrowserRoller {
       directory: _rollDir,
       packageName: cipdPackageName,
       configFileName: 'cipd.chromedriver.${platform.name}.yaml',
-      description: 'Chromedriver for Chromium $majorVersion (build $chromeBuild) used for testing',
-      version: majorVersion,
-      buildId: chromeBuild,
+      description: 'Chromedriver for Chromium $version used for testing',
+      version: version,
+      buildId: version,
```
mdebbar, line 291: It looks like we can remove the `buildId` parameter now?

mdebbar (approved):

lib/web_ui/dev/common.dart
eyebrowsoffire: Doh

PR #46696, opened 2023-10-10
use TextFrame calls for Impeller in rendertests

flar:
Switch the impeller testing in rendertests to use TextFrame objects rather than TextBlob objects when drawing to an Impeller-managed destination.

Previously the DrawText tests would generate over 200 errors because none of the attribute variants would render anything at all. After this fix the number of failures in the DrawText calls is down to 1 which points out that wrapping a DrawTextFrame call in a nothing saveLayer somehow affects the glyph positioning (in a not very appealing manner).

### DIFF
--- a/display_list/testing/dl_rendering_unittests.cc
+++ b/display_list/testing/dl_rendering_unittests.cc
@@ -17,7 +17,7 @@
 #include "flutter/testing/display_list_testing.h"
 #include "flutter/testing/testing.h"
 #ifdef IMPELLER_SUPPORTS_RENDERING
-#include "impeller/typographer/backends/skia/text_frame_skia.h"  // nogncheck
+#include "flutter/impeller/typographer/backends/skia/text_frame_skia.h"
 #endif  // IMPELLER_SUPPORTS_RENDERING
 
 #include "third_party/skia/include/core/SkBBHFactory.h"
@@ -3637,7 +3637,9 @@ TEST_F(DisplayListRendering, DrawTextBlob) {
 #else
   sk_sp<SkTextBlob> blob =
       CanvasCompareTester::MakeTextBlob("Testing", kRenderHeight * 0.33f);
+#ifdef IMPELLER_SUPPORTS_RENDERING
   auto frame = impeller::MakeTextFrameFromTextBlobSkia(blob);
+#endif  // IMPELLER_SUPPORTS_RENDERING
   SkScalar render_y_1_3 = kRenderTop + kRenderHeight * 0.3;
   SkScalar render_y_2_3 = kRenderTop + kRenderHeight * 0.6;
   CanvasCompareTester::RenderAll(  //
@@ -3654,12 +3656,14 @@ TEST_F(DisplayListRendering, DrawTextBlob) {
             ctx.canvas->DrawTextBlob(blob, kRenderLeft, render_y_2_3, paint);
             ctx.canvas->DrawTextBlob(blob, kRenderLeft, kRenderBottom, paint);
           },
+#ifdef IMPELLER_SUPPORTS_RENDERING
           [=](const DlRenderContext& ctx) {
             auto paint = ctx.paint;
             ctx.canvas->DrawTextFrame(frame, kRenderLeft, render_y_1_3, paint);
             ctx.canvas->DrawTextFrame(frame, kRenderLeft, render_y_2_3, paint);
             ctx.canvas->DrawTextFrame(frame, kRenderLeft, kRenderBottom, paint);
           },
+#endif  // IMPELLER_SUPPORTS_RENDERING
           kDrawTextBlobFlags)
           .set_draw_text_blob(),
       // From examining the bounds differential for the "Default" case, the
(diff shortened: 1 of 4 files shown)


display_list/testing/dl_rendering_unittests.cc
```diff
@@ -16,6 +16,9 @@
 #include "flutter/fml/math.h"
 #include "flutter/testing/display_list_testing.h"
 #include "flutter/testing/testing.h"
+#ifdef IMPELLER_SUPPORTS_RENDERING
+#include "impeller/typographer/backends/skia/text_frame_skia.h"  // nogncheck
```
jonahwilliams, line 20: I think the test likely needs a dependency on this in the gn file for this to compiler correctly

jonahwilliams:
Stroked text is supported, however this is implemented by immediately converting the text blob to path data and drawing the paths. Is there a way we could use the existing code in paragraph_skia could be reused for this test? (See https://github.com/flutter/engine/blob/main/third_party/txt/src/skia/paragraph_skia.cc#L209 )

jonahwilliams (approved):
LGTM (but not to CI)

display_list/testing/dl_rendering_unittests.cc
flar: Doesn't the nogncheck comment negate that issue? I didn't write it that way, I copied it from ... ?performance overlay?

I didn't see any gn failures on Mac or Windows.

flar: I did forget to ifdef the code that used it, though...

jonahwilliams: nogncheck will only silence gn analysis warnings, so it can be used if the dependency is transitive through some other target but if its not compiled in at all then you'll still get the linking errors.

flar: it compiles fine on Mac...

flar: I removed the comment and fixed the gn error...

flar:
> Stroked text is supported, however this is implemented by immediately converting the text blob to path data and drawing the paths. Is there a way we could use the existing code in paragraph_skia could be reused for this test? (See https://github.com/flutter/engine/blob/main/third_party/txt/src/skia/paragraph_skia.cc#L209 )

That would take a lot of special casing. I'll save that for a future fix.

display_list/testing/dl_rendering_unittests.cc
jonahwilliams: Try adding "//flutter/impeller/typographer/backends/skia:typographer_skia_backend" to the display_list_rendertests gn target deps.

flar:
> > Stroked text is supported, however this is implemented by immediately converting the text blob to path data and drawing the paths. Is there a way we could use the existing code in paragraph_skia could be reused for this test? (See https://github.com/flutter/engine/blob/main/third_party/txt/src/skia/paragraph_skia.cc#L209 )
> 
> That would take a lot of special casing. I'll save that for a future fix.

Hmm, so we don't even get a TextFrame in that case so it should be covered by DrawPath tests. I can skip the Impeller tests if the stroking is set up.

(Also, technically the test in the paragraph code should be `!= Fill`...)

flar:
I've added some tests that exempt non-color and stroked text from the Impeller testing. Re-requesting a review...

display_list/testing/dl_rendering_unittests.cc
```diff
@@ -818,11 +816,34 @@ class TestParameters {
...
   bool uses_paint() const { return !flags_.ignores_paint(); }
   bool uses_gradient() const { return flags_.applies_shader(); }
 
+  bool impeller_compatible(const DlPaint& paint) const {
+    if (is_draw_text_blob()) {
+      // Non-color text is rendered as paths
+      if (paint.getColorSourcePtr() && !paint.getColorSourcePtr()->asColor()) {
+        return false;
+      }
+      // Non-filled text (stroke or stroke and fill) is rendered as paths
```
jonahwilliams, line 839: AFAIK stroke & fill is not exposed at the dart:ui layer, unless there is an indirect path to enabling it.

testing/display_list_testing.cc
```diff
@@ -287,6 +287,17 @@ static std::ostream& operator<<(std::ostream& os, const SkTextBlob* blob) {
   return os << "&SkTextBlob(ID: " << blob->uniqueID() << ", " << blob->bounds() << ")";
 }
 
+static std::ostream& operator<<(std::ostream& os,
```
jonahwilliams, line 290: Could be a utility in text_frame.cc

jonahwilliams (approved):
LGTM

display_list/testing/dl_rendering_unittests.cc
flar: Correct, but with respect to the enum being queried, `!= kFill` is correct. At some point we can eliminate the extra value, but this comparison will still be correct.

testing/display_list_testing.cc
flar: TextFrame might want to have a more fully featured streamer. This one just needs enough to hold it's place in the debugging output. The Text objects and Path just dump out some useful info here like bounds.

PR #46705, opened 2023-10-10
[Impeller] Don't cull readbacks outside the damage rect.

bdero:
Resolves https://github.com/flutter/flutter/issues/136058.

The damage rect properly expands to include the backdrop filter readback area to a certain extent, but it suddenly gets excluded once it no longer partially overlaps with the damage rect. This caused the illusion of EntityPass improperly culling the backdrop entity, leading me into a bit of a wild goose chase.

Before:

https://github.com/flutter/engine/assets/919017/94b8c077-0945-4a2c-96e0-27230d980c38

After:

https://github.com/flutter/engine/assets/919017/f1c78365-6e9b-46cb-9e69-33472d488831


knopp:
I think the idea of readback region that doesn't overlap filter drawn area hasn't occurred to me when implementing this. However the issue here is that readback regions are added unconditionally. That means that readback area of any backdrop filter, even if the filter is culled out, will be repainted. For blurred header / footer this would pretty much invalidate partial repaint.

I'm wondering if we could store redbacks as a map `filter_paint_bounds` -> `filter_readback_bounds`. And then if dirty area intersect `filter_paint_bounds`, add the entire `filter_readback_bounds`.

knopp:
I mean something like https://github.com/knopp/engine/commit/23147d73b031f6515f800a3a31d961d69e6c3477 might work, need to test it though.

knopp:
@bdero, https://github.com/bdero/flutter-engine/pull/1

gaaclarke (approved):
The code looks good.  The test is very difficult to follow.  This is a classic case of a test should assert one thing.  Also the number literals took some time to understand.  It would be easier to follow if they were broken down to how they are calculated.

It sounds like there are pathological cases where this is very inefficient as noted by knopp.  I think it probably makes sense to prefer slower but correct rendering and we can file an issue about the pathological cases.

knopp:
That test could have been split into multiple tests. @gaaclarke, I added PR for this PR that should fix the magnifier rendering without regressing in https://github.com/bdero/flutter-engine/pull/1.

gaaclarke:
> That test could have been split into multiple tests. @gaaclarke, I added PR for this PR that should fix the magnifier rendering without regressing in [bdero#1](https://github.com/bdero/flutter-engine/pull/1). If this gets merged like this I'll follow-up with a separate PR.

Awesome, I just gave it a look and added a few notes.  I'll let you and bdero decide how to best land it.  I really appreciate the work both of you put into this one.

knopp (approved):

bdero:
Thanks, @knopp!

PR #46724, opened 2023-10-10
Fix high FPS screen flicker of Platform Views when using ImageReaderPlatformViewRenderTarget

johnmccutchan:
The root bug (b/300627634) was that we were holding onto HardwareBuffers after the owning Image was closed. This CL refactors the C++ code to properly hold a reference to the Image until it is safe to dispose of.

This CL also refactors the Impeller GL and Skia GL code paths to share more code.

### DIFF
--- a/shell/platform/android/hardware_buffer_external_texture.cc
+++ b/shell/platform/android/hardware_buffer_external_texture.cc
@@ -3,6 +3,8 @@
 
 #include <android/hardware_buffer_jni.h>
 #include <android/sensor.h>
+
+#include "flutter/shell/platform/android/jni/platform_view_android_jni.h"
 #include "flutter/shell/platform/android/ndk_helpers.h"
 
 namespace flutter {
@@ -23,6 +25,7 @@ void HardwareBufferExternalTexture::Paint(PaintContext& context,
   if (state_ == AttachmentState::kDetached) {
     return;
   }
+  Attach(context);
   const bool should_process_frame =
       (!freeze && new_frame_ready_) || dl_image_ == nullptr;
   if (should_process_frame) {
@@ -39,7 +42,7 @@ void HardwareBufferExternalTexture::Paint(PaintContext& context,
         flutter::DlCanvas::SrcRectConstraint::kStrict  // enforce edges
     );
   } else {
-    FML_LOG(WARNING)
+    FML_LOG(ERROR)
         << "No DlImage available for HardwareBufferExternalTexture to paint.";
   }
 }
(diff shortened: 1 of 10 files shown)


shell/platform/android/hardware_buffer_external_texture.cc
```diff
@@ -57,53 +60,57 @@ void HardwareBufferExternalTexture::OnGrContextCreated() {
   state_ = AttachmentState::kUninitialized;
 }
 
-AHardwareBuffer* HardwareBufferExternalTexture::GetLatestHardwareBuffer() {
+// Implementing flutter::ContextListener.
+void HardwareBufferExternalTexture::OnGrContextDestroyed() {
+  FML_LOG(ERROR) << "HardwareBufferExternalTexture::OnGrContextDestroyed";
```
jonahwilliams, line 65: nit: debugging log?

shell/platform/android/hardware_buffer_external_texture_gl.cc
```diff
@@ -27,122 +28,171 @@
...
 namespace flutter {
 
+HardwareBufferExternalTextureGL::HardwareBufferExternalTextureGL(
+    int64_t id,
+    const fml::jni::ScopedJavaGlobalRef<jobject>& image_texture_entry,
+    const std::shared_ptr<PlatformViewAndroidJNI>& jni_facade)
+    : HardwareBufferExternalTexture(id, image_texture_entry, jni_facade) {}
+
+HardwareBufferExternalTextureGL::~HardwareBufferExternalTextureGL() {}
+
```
jonahwilliams, line 38: nit: can this be `= default`?

jonahwilliams (approved):
It would be nice to have a way to unit test this, especially if that is possible without mocking everything out. Barring that, if there were a way to simulate the conditions that gave fairly reliable dropped contents that we could run as an integration test...

LGTM

shell/platform/android/hardware_buffer_external_texture.cc
johnmccutchan: Done.

shell/platform/android/hardware_buffer_external_texture_gl.cc
johnmccutchan: Done.

johnmccutchan:
> It would be nice to have a way to unit test this, especially if that is possible without mocking everything out. Barring that, if there were a way to simulate the conditions that gave fairly reliable dropped contents that we could run as an integration test...
> 
> LGTM

I don't think we can write a test against the dropped frames, extremely difficult to reproduce as it is entirely timing dependent.  

Once the dust settles I plan on expanding golden image tests for Android platform views.

PR #46745, opened 2023-10-11
Remove support for `Paint.enableDithering=false` in `dart:ui`.

matanlurey:
Work towards https://github.com/flutter/flutter/issues/112498.

No behavioral changes to existing code, except `Paint.enableDithering = false` will no longer be an option.

### DIFF
--- a/lib/ui/painting.dart
+++ b/lib/ui/painting.dart
@@ -1091,9 +1091,8 @@ class Paint {
   /// Constructs an empty [Paint] object with all fields initialized to
   /// their defaults.
   Paint() {
-    if (enableDithering) {
-      _dither = true;
-    }
+    // TODO(matanlurey): Remove as part of https://github.com/flutter/flutter/issues/112498.
+    _enableDithering();
   }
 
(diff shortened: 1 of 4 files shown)


lib/ui/painting.dart
```diff
@@ -1091,9 +1091,8 @@ class Paint {
   /// Constructs an empty [Paint] object with all fields initialized to
   /// their defaults.
   Paint() {
-    if (enableDithering) {
-      _dither = true;
-    }
+    // TODO(matanlurey): Remove as part of https://github.com/flutter/flutter/issues/112498.
+    _enableDithering();
```
jonahwilliams, line 1095: Do we need to leave this in for now? Could we just rip it out entirely?

matanlurey: We could rip it out with a much larger change, but I'm trying to do it piecemeal (i.e. https://github.com/flutter/engine/pull/46746).

jonahwilliams (approved):
LGTM

PR #46751, opened 2023-10-11
[Impeller] Allow pipeline variant sets to have differing defaults.

bdero:
Track default prototype options along side pipeline variant sets, allowing differing defaults to be assigned during ContentContext construction.

### DIFF
--- a/impeller/entity/contents/content_context.h
+++ b/impeller/entity/contents/content_context.h
@@ -721,11 +721,59 @@ class ContentContext {
   std::shared_ptr<Context> context_;
   std::shared_ptr<LazyGlyphAtlas> lazy_glyph_atlas_;
 
-  template <class T>
-  using Variants = std::unordered_map<ContentContextOptions,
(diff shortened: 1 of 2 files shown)


impeller/entity/contents/content_context.h
```diff
@@ -722,10 +722,64 @@ class ContentContext {
   std::shared_ptr<LazyGlyphAtlas> lazy_glyph_atlas_;
 
   template <class T>
-  using Variants = std::unordered_map<ContentContextOptions,
-                                      std::unique_ptr<T>,
-                                      ContentContextOptions::Hash,
-                                      ContentContextOptions::Equal>;
+  using Variantss = std::unordered_map<ContentContextOptions,
```
jonahwilliams, line 725: 🐍

bdero: Whoops, removed.

jonahwilliams (approved):
Seems reasonable to me!

LGTM

PR #46758, opened 2023-10-11
[Impeller] Avoid invalid GL depth calls on macOS desktop.

bdero:
The macOS desktop driver doesn't support ES profiles, and `glClearDepthf` + `glDepthRangef` are throwing `GL_INVALID_OPERATION` and doing nothing. Noticed this while updating impeller-cmake past @matanlurey's recent changes to make `GL_INVALID_OPERATION` fatal. Impeller-cmake-example uses a depth buffer, whereas Impeller's 2D renderer currently does not.

Added a follow-up bug to make these calls work on macOS desktop: https://github.com/flutter/flutter/issues/136322.
Both Impeller Scene and the impeller-cmake-example currently just rely on the default state for clearing depth.


jonahwilliams (approved):
LGTM

PR #46771, opened 2023-10-11
Roll goldctl forward

zanderso:
For https://github.com/flutter/flutter/issues/135620

### DIFF
--- a/ci/builders/linux_web_engine.json
+++ b/ci/builders/linux_web_engine.json
@@ -351,7 +351,7 @@
       "test_dependencies": [
         {
           "dependency": "goldctl",
-          "version": "git_revision:3a77d0b12c697a840ca0c7705208e8622dc94603"
+          "version": "git_revision:dddc0623e63150cbbafdcb273d4048f329e1dd09"
         },
         {
           "dependency": "chrome_and_driver",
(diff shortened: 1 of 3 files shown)


kjlubick (approved):
cc @Piinks

vashworth (commented):
Ran it on macOS 13 and it appears to have fixed the issue! Thank you @zanderso!

https://chromium-swarm.appspot.com/task?id=6537fecff3544710

Should we also update the goldctl version here?
https://github.com/flutter/engine/blob/ec1d0d69fcf11aa64a271a1fe2f582df37f3f2e9/lib/web_ui/dev/generate_builder_json.dart#L152

zanderso:
> Ran it on macOS 13 and it appears to have fixed the issue! Thank you @zanderso!

Great!

> Should we also update the goldctl version here?

Yes =) Thanks. I had made that change locally and forgot to push it.

Piinks (approved):
LGTM! Thank you!
We should probably create a cadence for rolling this forward more regularly. We pinned it since sometimes goldctl would (rarely) have a bad build and then we would then be broken in the meantime. @godofredoc do we have other systems in place that roll dependencies forward like this?

PR #46786, opened 2023-10-11
Rename HardwareBufferExternalTexture* to ImageExternalTexture*

johnmccutchan:
Image is the more accurate name and we will eventually have a HardwareBufferExternalTexture in the future (see https://developer.android.com/reference/android/graphics/HardwareBufferRenderer available on Android >= 34).


jonahwilliams (approved):
RSLGTM

PR #46831, opened 2023-10-12
[fuchsia] Add fatal error for Vulkan failure

uysalere:
This CL changes vkWaitForFences to cause a FATAL failure as we can't recover from these errors.

Bug: b/297198565


jbauman42:
LGTM. This seems like a better way to handle the error, since recovering from device lost is very difficult.

chinmaygarde (approved):

PR #46840, opened 2023-10-12
[Impeller] implements a retry mechanism for dart:ui/Image.toByteData.

gaaclarke:
Design doc: [link](https://docs.google.com/document/d/1Uuiw3pdQxNFTA8OQuZ-kuvYg1NB42XgccQCZeqr4oII/edit#heading=h.hn6wreyrz6fm)
fixes: https://github.com/flutter/flutter/issues/135245

One slight deviation from the design doc is that I decided to make ContextMTL respond to changes to the SyncSwitch instead of having it observe the app state directly.  The benefits are:
1) This keeps that functionality in one location
1) It makes writing tests much easier
1) There's no need of conditional compilation between macos and ios
1) There is no need to add an objc class

## Pre-launch Checklist

- [x] I signed the [CLA].


impeller/renderer/backend/metal/context_mtl.mm
```diff
@@ -376,4 +381,28 @@ new ContextMTL(device, command_queue,
   return buffer;
 }
 
+void ContextMTL::StoreTaskForGPU(std::function<void()> task) {
+  tasks_awaiting_gpu_.emplace_back(std::move(task));
+  while (tasks_awaiting_gpu_.size() > kMaxTasksAwaitingGPU) {
```
jonahwilliams, line 386: Why do we need a maximum limit of tasks? If we do have a limit, do we need a mechanism to drop them gracefully?

For example, if there is a Future pending from toByteData, I'd think that dropping the task should cause an error to be thrown?

gaaclarke: `EncodeImageFailsWithoutGPUImpeller` tests this, it's handled gracefully since you have to check for the GPU in the callback and we never retry more than once.  Getting dropped is just like the behavior we had before we implemented the retry.

lib/ui/painting/image_encoding_impeller.cc
```diff
@@ -56,21 +56,58 @@ sk_sp<SkImage> ConvertBufferToSkImage(
...
+              }
+            };
+            if (retry_runner) {
+              retry_runner->PostTask(retry_task);
+            } else {
+              retry_task();
+            }
+          });
+    } else {
+      encode_task(status);
```
jonahwilliams, line 108: This code is a little hard to read. Are we dropping the status here in the event its not OK and the code isn't unavailable, or does that percolate up elsewhere?

gaaclarke: I added some comments.  Let me know if you think this is the right amount and if it is helpful.

impeller/renderer/backend/metal/context_mtl.mm
jonahwilliams: Oh, so when we exceed the size we execute the task, and then the task sees the GPU is still unavailable and errors? makes sense. I would document that a bit.

gaaclarke: Added that info do the Context::StoreTaskForGPU docstring.

impeller/renderer/context.h
```diff
@@ -52,6 +52,10 @@ class Context {
     kVulkan,
   };
 
+  /// The maximum number of tasks that should ever be stored for
+  /// `StoreTaskForGPU`.
+  static constexpr int32_t kMaxTasksAwaitingGPU = 10;
```
gaaclarke, line 57: Oh by the way.  I'm open to increasing this.  I just pulled this number out of nowhere but I knew I didn't want it to be completely unbounded.

jonahwilliams: I would leave a note that this was arbitrarily chosen so we're not afraid to increase it in the future.

gaaclarke: done!

jonahwilliams (approved):
LGTM

gaaclarke:
Turns out some tests were using nullptr for the syncswitch when it shouldn't.  I've updated the test.  Those tests are a pain to debug.

PR #46440, opened 2023-09-30
Remove the frontend server wrapper

zanderso:
Instead, this PR copies the frontend server from the Dart SDK (whether prebuilt or not) into the location expected by internal engine tests and artifact construction. This PR also consolidates the three GN templates that invoked the frontend server down to one.

Framework presubs: https://github.com/flutter/flutter/pull/135836

Related: https://github.com/flutter/flutter/issues/60007

PR #46842, opened 2023-10-12
Fix --no-prebuilt-dart-sdk build

zanderso:
Fix for previous commit https://github.com/flutter/engine/pull/46440

### DIFF
--- a/build/dart/rules.gni
+++ b/build/dart/rules.gni
@@ -36,10 +36,7 @@ template("flutter_frontend_server") {
 
   kernel_output = invoker.kernel_output
 
-  common_deps = [
-    "//flutter/flutter_frontend_server:frontend_server",
-    "//flutter/lib/snapshot:strong_platform",
-  ]
+  common_deps = [ "//flutter/lib/snapshot:strong_platform" ]
   if (defined(invoker.deps)) {
     common_deps += invoker.deps
   }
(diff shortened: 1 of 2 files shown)


aam (approved):

jason-simmons:
I had started looking at this, and IIUC the `packages` parameter in `dart_action` will add a `--packages` flag to the Dart command line (making it redundant with `packages_args`.

So I'd suggest:
* remove `packages_args`
* for the prebuilt Dart SDK case, add the packages file to `inputs` and the `--packages` flag to `args` if `package_config` is defined
* for the non-prebuilt case, set the `packages` parameter in `dart_action`

jason-simmons:
Also, after applying the fix I'm still seeing errors when configuring an Android build with `--no-prebuilt-dart-sdk`:

```
ERROR Unresolved dependencies.
//flutter/flutter_frontend_server:frontend_server(//build/toolchain/android:clang_arm)
  needs //flutter/build/dart:dart_sdk(//build/toolchain/android:clang_arm)
```

zanderso:
@jason-simmons Good catch. I've updated the PR.

jason-simmons:
With this patch I'm still seeing errors when building a 32-bit `android_debug_unopt` target on an x64 host.

When it builds the scenario app snapshot, the frontend server fails with:
```
output: VM initialization failed: JIT cannot simulate target architecture with different word size than host
```

zanderso:
It turns out that when not using the prebuilt Dart SDK, we were previously running the frontend server from source. I've updated this patch to go back to that behavior. Thanks, Jason!

zanderso:
Will merge this into the reland of https://github.com/flutter/engine/pull/46863

PR #46846, opened 2023-10-12
[Impeller] GPUTracer for Metal.

jonahwilliams:
Impelement GPU frame time tracing for metal. Unlike Vulkan, this can't use dummy cmd buffers as those don't seem to ever get an executed callback. Instead we decorate every submitted cmd buffer with tracing functionality, and min/max the timestamps to compute the range of computation time.


impeller/renderer/backend/metal/gpu_tracer_mtl.mm
```diff
@@ -0,0 +1,53 @@
...
+}
+
+void GPUTracerMTL::RecordCmdBuffer(id<MTLCommandBuffer> buffer) {
+#ifdef IMPELLER_DEBUG
+  Lock lock(trace_state_mutex_);
+  auto current_state = current_state_;
+  trace_states_[current_state].pending_buffers += 1;
+
+  if (@available(ios 10.3, tvos 10.2, macos 10.15, macCatalyst 13.0, *)) {
+    [buffer addCompletedHandler:^(id<MTLCommandBuffer> buffer) {
```
jonahwilliams, line 30: I think this has some issues with GPUTracerMTL being deleted (or retained), but I'm not quite sure how to solve that with object-c / c++ interop.

jonahwilliams: Fixed, thanks @gaaclarke !

jonahwilliams:
Oh, hmm, I wonder if this available check covers simulators. Need to build for sim locally and confirm

jonahwilliams:
This works on Simulators!

impeller/renderer/backend/metal/gpu_tracer_mtl.mm
```diff
@@ -0,0 +1,60 @@
...
+#include "impeller/renderer/backend/metal/formats_mtl.h"
+
+#include <memory>
+
+#include "impeller/renderer/backend/metal/gpu_tracer_mtl.h"
+
+namespace impeller {
+
+void GPUTracerMTL::MarkFrameEnd() {
+#ifdef IMPELLER_DEBUG
```
dnfield, line 17: Nit - rather than guarding each method, just don't even allocate this outside of debug?

impeller/renderer/backend/metal/gpu_tracer_mtl.mm
```diff
@@ -0,0 +1,60 @@
...
+          state.smallest_timestamp, static_cast<Scalar>(buffer.GPUStartTime));
+      state.largest_timestamp = std::max(
+          state.largest_timestamp, static_cast<Scalar>(buffer.GPUEndTime));
+
+      if (state.pending_buffers == 0) {
+        auto gpu_ms =
+            (state.largest_timestamp - state.smallest_timestamp) * 1000;
+        state.smallest_timestamp = std::numeric_limits<float>::max();
+        state.largest_timestamp = 0;
+        FML_TRACE_COUNTER("flutter", "GPUTracer",
```
dnfield, line 51: This seems like it should be a constant somewhere so that MTL and VK use the same one always

dnfield (approved):
Nits, LGTM

impeller/renderer/backend/metal/gpu_tracer_mtl.mm
jonahwilliams: Where would you want this? I can put it in core?

jonahwilliams: Im gonna punt on this - the trace event will probably change a few times

PR #46860, opened 2023-10-13
Avoid a deadlock in the flutter_tester process when deleting the Impeller Vulkan context during shutdown

jason-simmons:
The Impeller ContextVK contains a ConcurrentMessageLoop whose threads may invoke Dart timeline APIs.  The Dart APIs will create a thread-local object that will be deleted during thread shutdown.  Therefore, these threads should not outlive the engine/Shell and Dart VM.

Previously, RunTester held the ImpellerVulkanContextHolder on the stack, and its reference to the ContextVK would be dropped while exiting the function after the Shell is destructed.

This PR moves the contents of the tester's ImpellerVulkanContextHolder out of the instance on the stack and into a lambda owned by the Shell.

It also reenables the flutter_tester Impeller tests in the run_tests script.

### DIFF
--- a/shell/testing/tester_main.cc
+++ b/shell/testing/tester_main.cc
@@ -61,10 +61,57 @@ static std::vector<std::shared_ptr<fml::Mapping>> ShaderLibraryMappings() {
 }
 
 struct ImpellerVulkanContextHolder {
+  ImpellerVulkanContextHolder() = default;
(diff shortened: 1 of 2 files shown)


jonahwilliams (approved):
RSLGTM

zanderso (approved):

PR #46861, opened 2023-10-13
[Impeller] Move queue submission to worker thread.

jonahwilliams:
Queue submission can be moved to a worker thread, but it has to happen in-order with resepect to the original submit order and in order wrt present KHR. To solve this, create a a dedicated single thread worker pool that performs all submission.

Queue submission can be moved to a worker thread. Which is nice, because ARM seems to spend most of its time blocked on a mutex in submission:

image


jonahwilliams:
Oh wait, this technically isn't safe because it can happen out of order. it would need to be to a dedicated thread.

jonahwilliams:
I think we might need to do this anyway because with sufficient backpressure submitKHR can actually happen out of order on certain benchmarks.

PR #46887, opened 2023-10-13
[Impeller] fix clear color optimization for large subpasses.

jonahwilliams:
Fixes https://github.com/flutter/flutter/issues/136507 .

The cause of this bug is that we use different sizes to compute the clear color value when constructing the render target and when skipping entities:

https://github.com/flutter/engine/blob/main/impeller/entity/entity_pass.cc#L632
https://github.com/flutter/engine/blob/main/impeller/entity/entity_pass.cc#L731 and
https://github.com/flutter/engine/blob/main/impeller/entity/entity_pass.cc#L900C3-L900C3

Either the former should use the root pass size or the later should use the subpass size. This usually isn't an issue because if something covers the root pass size it generally always covers the subpass size. But during the page transition, the scaled subpass ends up slightly bigger than the root pass size and so these conditions are mismatched:

Subpass size: (1550, 3188) root pass size (1440, 3036)

I think subpass size is correct. If the subpass is larger than the parent, then checking the parent size will give incorrect result.

### DIFF
--- a/impeller/aiks/aiks_unittests.cc
+++ b/impeller/aiks/aiks_unittests.cc
@@ -3664,5 +3664,25 @@ TEST_P(AiksTest,
   ASSERT_TRUE(OpenPlaygroundHere(canvas.EndRecordingAsPicture()));
 }
 
+// This should be solid red, if you see a little red box this is broken.
+TEST_P(AiksTest, ClearColorOptimizationWhenSubpassIsBiggerThanParentPass) {
+  SetWindowSize({400, 400});
+  Canvas canvas;
+  canvas.Scale(GetContentScale());
+  canvas.DrawRect(Rect::MakeLTRB(200, 200, 300, 300), {.color = Color::Red()});
+  canvas.SaveLayer({
+      .image_filter = std::make_shared<MatrixImageFilter>(
+          Matrix::MakeScale({2, 2, 1}), SamplerDescriptor{}),
+  });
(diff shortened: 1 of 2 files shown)


gaaclarke (approved):
Is there any docstrings that should be updated to make this more clear that this is the correct math?

jonahwilliams:
I'm not really sure. I sort of eyeballed this after reading through the code for a bit because it didn't seem right that we were using the root pass size to compute the clear color.

jonahwilliams:
Looks like this is breaking CanRenderClippedLayers

jonahwilliams:
So with this change, we're now collapsing both colors into the save layer and using the CPU blends. I think the issue might just be that ColorDodge is incorrect, because If I change the advanced blend it renders as expected.

jonahwilliams:
ColorDodge on the CPU is computing a blended color of (1, 1, 1, 1). I think this is incorrect as it doesn't seem to match the GPU blend.

jonahwilliams:
I confirmed that there are otherwise no issues with the pass structures, other advanced blends show up correctly - the issue is entirely the result of Color::Blend.

jonahwilliams:
I'm tempted to just adjust the blend mode to a different advanced blend and file a bug for fixing color dodge. Thoughts @bdero / @gaaclarke ?

jonahwilliams:
I adjusted the test and filed a bug

PR #46891, opened 2023-10-13
[web] Support `flutterViewId` in platform view messages

mdebbar:
- Accept a new `flutterViewId` field in platform view messages.
- Keep transitory support for legacy platform view messages that don't contain `flutterViewId`.
- Default view factories set `width:100%` and `height:100%`.


lib/web_ui/lib/src/engine/platform_dispatcher.dart
```diff
@@ -585,15 +584,22 @@ class EnginePlatformDispatcher extends ui.PlatformDispatcher {
...
       case 'flutter/platform_views':
-        implicitView!.platformViewMessageHandler.handlePlatformViewCall(data, callback!);
+        final MethodCall(:String method, :dynamic arguments) = standardCodec.decodeMethodCall(data);
+        if (arguments is Map<dynamic, dynamic>) {
+          final int? flutterViewId = arguments.tryInt('flutterViewId');
+          if (flutterViewId != null) {
+            viewData[flutterViewId]!.platformViewMessageHandler.handlePlatformViewCall(method, arguments, callback!);
+            return;
+          }
+        }
```
ditman, line 601: The bit that extracts `flutterViewId` from `arguments` should probably be a standard method that we reuse (similar to the codec), so we end up with something like:

```dart
final MethodCall ... arguments = xxCodec.decodeMethodCall(data);
final int? flutterViewId = extractViewId(arguments);
if (flutterViewId != null) {
  // new stuff
} else {
  // legacy stuff
} 
```

I think that there are more platform dispatcher methods that are going to pass a `flutterViewId` parameter, like: `render` and `updateSemantics` (at least), so we could reuse the same method for all.

mdebbar: Done!

lib/web_ui/lib/src/engine/util.dart
```diff
@@ -627,6 +628,23 @@ extension JsonExtensions on Map<dynamic, dynamic> {
...
+int readFlutterViewId(Object? arguments) {
+  return tryFlutterViewId(arguments)!;
+}
+
+/// Extracts view ID from the [MethodCall.arguments] map.
+///
+/// Returns null if the view ID is not present or if [arguments] is not a map.
+int? tryFlutterViewId(Object? arguments) {
+  if (arguments is Map) {
+    return arguments.tryInt('flutterViewId');
```
ditman, line 643: I'm trying to figure out if this is going to be named `'viewId'` or something else before we land this :P (for the `render`, for example)

ditman: Yeah, the change is going to be a little bit of a pain, since the string needs to be kept in sync between the engine <-> framework. I think @goderbauer mentioned he'd use `viewId` for other messages. Should we rename this (and the fw version)?

lib/web_ui/lib/src/engine/platform_views/message_handler.dart
```diff
@@ -105,28 +103,28 @@ class PlatformViewMessageHandler {
...
   }
 
-  /// Handles legacy PlatformViewCalls that don't contain a `flutterViewId`.
+  /// Handles legacy PlatformViewCalls that don't contain a Flutter View ID.
   ///
   /// This is transitional code to support the old platform view channel. As
-  /// soon as the framework code is updated to send the `flutterViewId`, this
+  /// soon as the framework code is updated to send the Flutter View ID, this
   /// method can be removed.
   void handleLegacyPlatformViewCall(
```
ditman, line 129: I totally forgot about this method. Good repurposing though!

lib/web_ui/lib/src/engine/platform_dispatcher.dart
```diff
@@ -585,15 +584,21 @@ class EnginePlatformDispatcher extends ui.PlatformDispatcher {
...
             callback,
-            codec.encodeSuccessEnvelope(
-                _handleWebTestEnd2EndMessage(codec, data)));
+            jsonCodec.encodeSuccessEnvelope(
+                _handleWebTestEnd2EndMessage(jsonCodec, data)));
         return;
 
       case 'flutter/platform_views':
-        implicitView!.platformViewMessageHandler.handlePlatformViewCall(data, callback!);
+        final MethodCall(:String method, :dynamic arguments) = standardCodec.decodeMethodCall(data);
```
ditman, line 594: Did I tell you already how FANCY this looks?

lib/web_ui/lib/src/engine/platform_views/content_manager.dart
```diff
@@ -249,5 +249,9 @@ DomElement _defaultFactory(
 }) {
   params!;
   params as Map<Object?, Object?>;
-  return domDocument.createElement(params.readString('tagName'));
+  final DomElement element = domDocument.createElement(params.readString('tagName'));
+  element.style
+    ..width = '100%'
+    ..height = '100%';
```
ditman, line 255: I think you could still one-line-ize this:

```suggestion
  return domDocument.createElement(params.readString('tagName'))
    ..style.width = '100%'
    ..style.height = '100%';
```

(But I haven't tested my suggestion, I just remember writing similar code :P)

lib/web_ui/lib/src/engine/util.dart
```diff
@@ -627,6 +628,23 @@ extension JsonExtensions on Map<dynamic, dynamic> {
   }
 }
 
+/// Extracts view ID from the [MethodCall.arguments] map.
+///
+/// Throws if the view ID is not present or if [arguments] is not a map.
```
ditman, line 633: This is going to throw a very ugly error when it attempts to return `null` instead of `int` right? Should this throw a more specific exception when `tryViewId` returns `null`?

ditman (approved):
Thanks for taking the time to build this in a backwards-compatible way! LGTM other than minor comments.

lib/web_ui/lib/src/engine/platform_views/content_manager.dart
mdebbar: Works like a charm!

lib/web_ui/test/engine/platform_views/message_handler_test.dart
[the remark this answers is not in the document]
mdebbar: Not sure what you mean by `it should have been deleted earlier`. The `handleLegacyPlatformViewCall` method is a new one that I added in this PR to manage the transition.

lib/web_ui/lib/src/engine/platform_views/message_handler.dart
mdebbar: You must've been looking at an individual commit from the PR 🙂 This method is new.

PR #46895, opened 2023-10-13
[Impeller] Only allow Impeller in flutter_tester if vulkan is enabled.

dnfield:
Follows up on #46389 

That patch was too permissive in cases where a build system enables impeller but not vulkan. This change makes the build succeed in such systems.


dnfield:
Without this change, google's internal build is failing. /cc @XilaiZhang

zanderso (approved):

PR #46897, opened 2023-10-13
Oops, allow files in opted-in third_party directories.

matanlurey:
Closes https://github.com/flutter/flutter/issues/136551.

### DIFF
--- a/third_party/.gitignore
+++ b/third_party/.gitignore
@@ -1,6 +1,6 @@
 # Ignore everything by default, as these come from gclient/DEPS.
 # We'll explicitly include the folders we want to track.
-*
+/*
 
 # Include the .gitignore file itself and .clang-tidy.
 !.gitignore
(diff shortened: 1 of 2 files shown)


cyanglaz:
I tired this, didn't work, @jason-simmons found out that the below works. Unsure why
```
!txt/
!txt/**
```

matanlurey:
Ah, `/*` as an opt-out is better than `*`. Looks good now locally.

cyanglaz (approved):
LGTM
Verified locally it works

third_party/.gitignore
```diff
@@ -13,11 +13,10 @@
...
 !accessibility/
 !canvaskit/
-!ninja/
 !spring_animation/
 !test_shaders/
 !tonic/
 !txt/
 !web_locale_keymap/
-!web_test_fonts/
+!web_test_font/
```
jason-simmons, line 21: Looks like the removal of the `s` is a typo?

matanlurey: Yup! Nice catch, thanks.

jason-simmons (approved):

PR #46901, opened 2023-10-13
[web] - Fix `inputmode` on Android Firefox

htoor3:
The wrong keyboard type shows when users specify the phone input type in the framework. 

This is happening because:

* We usually set `inputmode` for mobile text editing strategies.
* We only use Android's text editing strategy for Blink browser engines (not firefox), so we fall through to the Firefox text editing strategy which I believe should only be used for desktop firefox.
* So we're ending up in a case where Android + Firefox doesn't set the inputmode and it pops open the default alphanumeric keyboard

Fixes https://github.com/flutter/flutter/issues/136351

## Pre-launch Checklist

- [x] I signed the [CLA].


lib/web_ui/lib/src/engine/text_editing/text_editing.dart
```diff
@@ -1936,7 +1936,7 @@ DefaultTextEditingStrategy createDefaultTextEditingStrategy(HybridTextEditing te
     strategy = IOSTextEditingStrategy(textEditing);
   } else if (browserEngine == BrowserEngine.webkit) {
     strategy = SafariDesktopTextEditingStrategy(textEditing);
-  } else if (browserEngine == BrowserEngine.blink &&
+  } else if ((browserEngine == BrowserEngine.blink || browserEngine == BrowserEngine.firefox) &&
```
htoor3, line 1939: Is there a contextual reason why we have to specify a browser engine as well as the OS when we're trying to select for a mobile text editing strategy? Or can we just always choose the android strategy when `OS == android` and always choose ios strategy when `OS == iOS`

mdebbar (approved):
LGTM!

PR #46904, opened 2023-10-13
[flutter roll cp][Impeller] Only allow Impeller in flutter_tester if vulkan is enabled…

XilaiZhang:
…. (#46895)

context: b/305246682

pr description:
Follows up on #46389 

That patch was too permissive in cases where a build system enables impeller but not vulkan. This change makes the build succeed in such systems.


fluttergithubbot (approved):

zanderso:
From PR review triage: Closing as presumed stale.

PR #46906, opened 2023-10-13
[web] Make HotRestartCacheHandler standalone

mdebbar:
- `EmbeddingStrategy` shouldn't own the creation of `HotRestartCacheHandler`.
- Simplify `HotRestartCacheHandler`'s JS-interop by using a `JSArray` directly instead of going through a Dart `List`.


lib/web_ui/lib/src/engine/view_embedder/embedding_strategy/embedding_strategy.dart
```diff
@@ -21,13 +18,7 @@ import 'full_page_embedding_strategy.dart';
...
 ///   initialization.
 abstract class EmbeddingStrategy {
-  EmbeddingStrategy() {
-    // Initialize code to handle hot-restart (debug only).
-    assert(() {
-      _hotRestartCache = HotRestartCacheHandler();
-      return true;
-    }());
-  }
+  EmbeddingStrategy();
```
ditman, line 21: Do we need an empty constructor here?

lib/web_ui/lib/src/engine/dom.dart
```diff
@@ -3591,3 +3591,10 @@ extension DomFinalizationRegistryExtension on DomFinalizationRegistry {
 /// Whether the current browser supports `FinalizationRegistry`.
 bool browserSupportsFinalizationRegistry =
     _finalizationRegistryConstructor != null;
+
+@JS()
+@staticInterop
+extension JSArrayExtension on JSArray {
```
ditman, line 3597: Move this closer to the definition of JSArray maybe?

lib/web_ui/lib/src/engine/view_embedder/embedding_strategy/custom_element_embedding_strategy.dart
```diff
@@ -4,6 +4,7 @@
 
 import 'package:ui/src/engine/dom.dart';
 
+import '../hot_restart_cache_handler.dart';
```
ditman, line 7:
```suggestion
import '../hot_restart_cache_handler.dart' show registerElementForCleanup;
```

That way it's more clear (to me) that you're importing that file specifically for the function you want from it!

lib/web_ui/lib/src/engine/view_embedder/embedding_strategy/full_page_embedding_strategy.dart
```diff
@@ -5,6 +5,7 @@
 import 'package:ui/src/engine/dom.dart';
 import 'package:ui/src/engine/util.dart' show setElementStyle;
 
+import '../hot_restart_cache_handler.dart';
```
ditman, line 8: ~~Unused import?~~ No, it's a free function!

```suggestion
import '../hot_restart_cache_handler.dart' show registerElementForCleanup;
```

lib/web_ui/lib/src/engine/view_embedder/hot_restart_cache_handler.dart
```diff
@@ -12,59 +13,61 @@ import '../dom.dart';
...
-    _hotRestartStore?.toObjectShallow as List<Object?>?;
+external JSArray? get _jsHotRestartStore;
 
 @JS('window.__flutterState')
-external set _hotRestartStore(JSArray? nodes);
-set hotRestartStore(List<Object?>? nodes) =>
-    _hotRestartStore = nodes?.toJSAnyShallow as JSArray?;
+external set _jsHotRestartStore(JSArray? nodes);
+
+final HotRestartCacheHandler? _hotRestartCache = () {
```
ditman, line 21: (Can you move the _hotRestartCache creation and the definition of registerElementForCleanup underneath the HotRestartCacheHandler class? I think it's more readable (Here's the class, and here's what we do with it, rather than the other way around))

ditman (approved):
Small nits that I think will help with readability of the code (or maybe it's because it's a little bit late and my brain is collapsing 🧠💨)

lib/web_ui/lib/src/engine/view_embedder/embedding_strategy/embedding_strategy.dart
mdebbar: Removing it causes the following error:
```
The class 'FullPageEmbeddingStrategy' can't extend 'EmbeddingStrategy' because 'EmbeddingStrategy' only has factory constructors (no generative constructors), and 'FullPageEmbeddingStrategy' has at least one generative constructor.
```
But I think we can make the subclasses `implements EmbeddingStrategy` instead of extending. Let me try that.

lib/web_ui/lib/src/engine/dom.dart
mdebbar: `JSArray` is defined in `dart:js_interop` :)

lib/web_ui/lib/src/engine/view_embedder/embedding_strategy/custom_element_embedding_strategy.dart
mdebbar: Done

lib/web_ui/lib/src/engine/view_embedder/embedding_strategy/embedding_strategy.dart
mdebbar: Done

lib/web_ui/lib/src/engine/view_embedder/hot_restart_cache_handler.dart
mdebbar: Done

PR #46911, opened 2023-10-13
Fix `Platform.script` for flutter_tester

dnfield:
This addresses the problem in https://github.com/flutter/flutter/issues/12847 which changed slightly over time.

Today, `Platform.script` does not give an empty `file` URI, it gives something like `file://path/to/package/main.dart` _regardless of how the file is actually named_.

After this change, it will give the absolute path to the file being run under test.

So before this change, the new test would have a URI like

```
file:///Users/dnfield/src/flutter/engine/src/main.dart
```

And now it has

```
file:///Users/dnfield/src/flutter/engine/src/out/host_debug_unopt_arm64/gen/platform_test.dart.dill
```

This is going to be helpful in generating relative paths from the test file.

### DIFF
--- a/shell/testing/tester_main.cc
+++ b/shell/testing/tester_main.cc
@@ -526,6 +526,7 @@ int main(int argc, char* argv[]) {
     // it as a positional argument instead.
     settings.application_kernel_asset = command_line.positional_args()[0];
   }
+  settings.advisory_script_uri = settings.application_kernel_asset;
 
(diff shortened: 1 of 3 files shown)


jonahwilliams (approved):
LGTM

zanderso (approved):

christopherfujino (approved):
cool

dnfield:
Not able to repro failure locally, try to rebase to see if it helps...

PR #46912, opened 2023-10-13
[Impeller] Adjust clip coverage operations to handle per-pass clip stacks

jason-simmons:
https://github.com/flutter/engine/pull/46597 creates a new clip coverage stack for each subpass.  This PR updates some clip coverage operations and assertions in EntityPass::OnRender to reflect that the current clip coverage stack only contains the range of depths within the current subpass.


bdero (approved):
LGTM, thanks!

PR #46915, opened 2023-10-14
[Impeller] Vulkan GPU tracer cleanups

jonahwilliams:
Cleanups that seem to reduce occurance of SIGBUS crash, which It turns out I did not in fact fix.

### DIFF
--- a/impeller/renderer/backend/vulkan/gpu_tracer_vk.cc
+++ b/impeller/renderer/backend/vulkan/gpu_tracer_vk.cc
@@ -100,23 +100,27 @@ void GPUTracerVK::RecordEndFrameTime() {
   // On completion of the second time stamp recording, we read back this value
   // and the previous value. The difference is approximately the frame time.
   const auto device_holder = strong_context->GetDeviceHolder();
-  if (!buffer->SubmitCommands([&, last_query,
-                               device_holder](CommandBuffer::Status status) {
+  auto submitted = buffer->SubmitCommands(
+      [&, last_query, device_holder](CommandBuffer::Status status) {
+        if (status != CommandBuffer::Status::kCompleted) {
+          return;
+        }
         uint64_t bits[2] = {0, 0};
         auto result = device_holder->GetDevice().getQueryPoolResults(
             query_pool_.get(), last_query, 2, sizeof(bits), &bits,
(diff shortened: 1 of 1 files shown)


impeller/renderer/backend/vulkan/gpu_tracer_vk.cc
```diff
@@ -102,10 +102,6 @@ void GPUTracerVK::RecordEndFrameTime() {
   const auto device_holder = strong_context->GetDeviceHolder();
   if (!buffer->SubmitCommands([&, last_query,
                                device_holder](CommandBuffer::Status status) {
-        auto strong_context = context_.lock();
```
jonahwilliams, line 105: We dont need this here, this might be causing shutdown issues since this may keep the context alive longer than it should

impeller/renderer/backend/vulkan/gpu_tracer_vk.cc
```diff
@@ -121,9 +117,7 @@ void GPUTracerVK::RecordEndFrameTime() {
               "FrameTimeMS", gpu_ms);
         }
       })) {
-    if (!buffer->SubmitCommands()) {
```
jonahwilliams, line 124: oops

impeller/renderer/backend/vulkan/gpu_tracer_vk.cc
```diff
@@ -121,9 +117,7 @@ void GPUTracerVK::RecordEndFrameTime() {
...
-      VALIDATION_LOG << "GPUTracerVK failed to record frame end time.";
-    }
+    VALIDATION_LOG << "GPUTracerVK failed to record frame end time.";
```
zanderso, line 120: optional minor readability nit: Since the body of the if follows a long closure argument to `SubmitCommands` it was hard for me to tell whether it was part of the closure, part of the if-body, or after the if-body. Maybe consider assigning the result of `SubmitCommands` to a bool variable, and then branch on the bool variable.

zanderso (approved):
lgtm w/ nit

impeller/renderer/backend/vulkan/gpu_tracer_vk.cc
jonahwilliams: Done

PR #46963, opened 2023-10-16
[Impeller] Rework Vulkan GPUTracker to decorate existing cmd buffers.

jonahwilliams:
Switches the GPU tracing implementation to work more like the metal version, where we decorate individual cmd buffers with start/end time queries and then min/max the final result. Hopefully this stabilizes the values on CI

### DIFF
--- a/impeller/renderer/backend/vulkan/command_encoder_vk.cc
+++ b/impeller/renderer/backend/vulkan/command_encoder_vk.cc
@@ -187,19 +187,19 @@ bool CommandEncoderVK::Submit(SubmitCallback callback) {
 
   auto status = command_buffer.end();
   if (status != vk::Result::eSuccess) {
-    gpu_tracer_->OnFenceComplete(end_frame);
+    gpu_tracer_->OnFenceComplete(end_frame, false);
     VALIDATION_LOG << "Failed to end command buffer: " << vk::to_string(status);
     return false;
   }
   std::shared_ptr<const DeviceHolder> strong_device = device_holder_.lock();
   if (!strong_device) {
-    gpu_tracer_->OnFenceComplete(end_frame);
+    gpu_tracer_->OnFenceComplete(end_frame, false);
     VALIDATION_LOG << "Device lost.";
     return false;
   }
   auto [fence_result, fence] = strong_device->GetDevice().createFenceUnique({});
   if (fence_result != vk::Result::eSuccess) {
-    gpu_tracer_->OnFenceComplete(end_frame);
+    gpu_tracer_->OnFenceComplete(end_frame, false);
     VALIDATION_LOG << "Failed to create fence: " << vk::to_string(fence_result);
     return false;
   }
@@ -209,7 +209,7 @@ bool CommandEncoderVK::Submit(SubmitCallback callback) {
   submit_info.setCommandBuffers(buffers);
   status = queue_->Submit(submit_info, *fence);
   if (status != vk::Result::eSuccess) {
-    gpu_tracer_->OnFenceComplete(end_frame);
+    gpu_tracer_->OnFenceComplete(end_frame, false);
     VALIDATION_LOG << "Failed to submit queue: " << vk::to_string(status);
     return false;
   }
@@ -222,7 +222,7 @@ bool CommandEncoderVK::Submit(SubmitCallback callback) {
       std::move(fence),
       [callback, tracked_objects = std::move(tracked_objects_), gpu_tracer,
        end_frame] {
-        gpu_tracer->OnFenceComplete(end_frame);
+        gpu_tracer->OnFenceComplete(end_frame, true);
         if (callback) {
           callback(true);
         }
--- a/impeller/renderer/backend/vulkan/command_encoder_vk.cc
+++ b/impeller/renderer/backend/vulkan/command_encoder_vk.cc
@@ -134,10 +134,8 @@ std::shared_ptr<CommandEncoderVK> CommandEncoderFactoryVK::Create() {
     context_vk.SetDebugName(tracked_objects->GetCommandBuffer(),
                             label_.value());
   }
-#ifdef IMPELLER_DEBUG
   context->GetGPUTracer()->RecordCmdBufferStart(
       tracked_objects->GetCommandBuffer());
-#endif  // IMPELLER_DEBUG
 
   return std::make_shared<CommandEncoderVK>(
       context_vk.GetDeviceHolder(), tracked_objects, queue,
@@ -185,10 +183,7 @@ bool CommandEncoderVK::Submit(SubmitCallback callback) {
 
   auto command_buffer = GetCommandBuffer();
 
-  size_t end_frame = 0u;
-#ifdef IMPELLER_DEBUG
-  end_frame = gpu_tracer_->RecordCmdBufferEnd(command_buffer);
-#endif  // IMPELLER_DEBUG
+  size_t end_frame = gpu_tracer_->RecordCmdBufferEnd(command_buffer);
 
   auto status = command_buffer.end();
   if (status != vk::Result::eSuccess) {
--- a/impeller/renderer/backend/vulkan/command_encoder_vk.cc
+++ b/impeller/renderer/backend/vulkan/command_encoder_vk.cc
@@ -183,7 +183,7 @@ bool CommandEncoderVK::Submit(SubmitCallback callback) {
 
   auto command_buffer = GetCommandBuffer();
 
-  size_t end_frame = gpu_tracer_->RecordCmdBufferEnd(command_buffer);
+  auto end_frame = gpu_tracer_->RecordCmdBufferEnd(command_buffer);
 
   auto status = command_buffer.end();
   if (status != vk::Result::eSuccess) {
@@ -222,7 +222,9 @@ bool CommandEncoderVK::Submit(SubmitCallback callback) {
       std::move(fence),
       [callback, tracked_objects = std::move(tracked_objects_), gpu_tracer,
        end_frame] {
-        gpu_tracer->OnFenceComplete(end_frame, true);
+        if (end_frame.has_value()) {
+          gpu_tracer->OnFenceComplete(end_frame, true);
+        }
         if (callback) {
           callback(true);
         }
--- a/impeller/renderer/backend/vulkan/command_encoder_vk.cc
+++ b/impeller/renderer/backend/vulkan/command_encoder_vk.cc
@@ -7,6 +7,7 @@
 #include "flutter/fml/closure.h"
 #include "impeller/renderer/backend/vulkan/context_vk.h"
 #include "impeller/renderer/backend/vulkan/fence_waiter_vk.h"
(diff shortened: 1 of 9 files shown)


jonahwilliams:
I even added ... well, its sort of a test.

impeller/renderer/backend/vulkan/command_encoder_vk.cc
```diff
@@ -178,18 +183,23 @@ bool CommandEncoderVK::Submit(SubmitCallback callback) {
 
   auto command_buffer = GetCommandBuffer();
 
+  size_t end_frame = gpu_tracer_->RecordCmdBufferEnd(command_buffer);
+
   auto status = command_buffer.end();
   if (status != vk::Result::eSuccess) {
+    gpu_tracer_->OnFenceComplete(end_frame);
```
dnfield, line 190: Here and below, should we somehow signal to the tracer that it's ending really fast because of a failure?

I'm assuming here we'd get a zero duration, do we want something like a -1 duration?

impeller/renderer/backend/vulkan/gpu_tracer_vk.cc
```diff
@@ -5,127 +5,153 @@
...
 
 #include <utility>
+#include "fml/logging.h"
 #include "fml/trace_event.h"
 #include "impeller/base/validation.h"
-#include "impeller/renderer/backend/vulkan/command_buffer_vk.h"
-#include "impeller/renderer/backend/vulkan/command_encoder_vk.h"
 #include "impeller/renderer/backend/vulkan/context_vk.h"
-#include "impeller/renderer/command_buffer.h"
+#include "vulkan/vulkan_enums.hpp"
```
dnfield, line 12: Is this right? Should this just be `vulkan/vulkan.hpp`?

jonahwilliams:
The previous implementation regressed raster time due to the addition of more cmd buffer submissions:

https://flutter-flutter-perf.skia.org/e/?keys=X9d806842a167e6fdac66ce7ec1922f27&requestType=0&selected=commit%3D37394%26name%3D%252Carch%253Dintel%252Cbranch%253Dmaster%252Cconfig%253Ddefault%252Cdevice_type%253DPixel_7_Pro%252Cdevice_version%253Dnone%252Chost_type%253Dlinux%252Csub_result%253Daverage_frame_rasterizer_time_millis%252Ctest%253Dnew_gallery_impeller__transition_perf%252C&xbaroffset=37394

We should make sure this fixes that.

impeller/renderer/backend/vulkan/gpu_tracer_vk.cc
jonahwilliams: I get "included header is not used"

impeller/renderer/backend/vulkan/command_encoder_vk.cc
jonahwilliams: Done. If we encounter an error we should not report for the frame since its probably an all bets are off situation.

impeller/renderer/backend/vulkan/gpu_tracer_vk.cc
```diff
@@ -5,127 +5,155 @@
...
-  }
-  query_pool_ = std::move(pool);
   // Disable tracing in release mode.
 #ifdef IMPELLER_DEBUG
   valid_ = true;
 #endif
 }
 
-void GPUTracerVK::RecordStartFrameTime() {
+bool GPUTracerVK::IsValid() const {
```
matanlurey, line 33: Just being pedantic, `IsValid` feels weird given it's totally valid for this instance to exist in an invalid state (i.e. either timestamp_period not being supported, or debug mode being disabled). 

Consider renaming this `IsEnabled()` and consider making it private since there isn't (AFAICT) a reason for another class to check this property, it's internal-only state.

impeller/renderer/backend/vulkan/gpu_tracer_vk.cc
```diff
@@ -5,127 +5,155 @@
...
+  return valid_;
+}
+
+void GPUTracerVK::MarkFrameStart() {
+  in_frame_ = true;
```
matanlurey, line 38: Is it worth adding a `FML_DCHECK` that a frame is never started when `in_frame_` already is true?

(Maybe that's fine given what you told me offline about multiple frames running concurrently - if so a 1-line comment documenting that would be nice to have)

impeller/renderer/backend/vulkan/gpu_tracer_vk.cc
```diff
@@ -5,127 +5,155 @@
...
-                                     query_pool_.get(), current_index_);
-
-  if (!buffer->SubmitCommands()) {
-    VALIDATION_LOG << "GPUTracerVK: Failed to record start time.";
-  }
+  Lock lock(trace_state_mutex_);
+  current_state_ = (current_state_ + 1) % 16;
+
+  auto& state = trace_states_[current_state_];
+  FML_DCHECK(state.pending_buffers == 0u);
```
matanlurey, line 49: nit: Can you add a `<< "Reason why bad"` sort of self-documenting why this should be 0?

It's confusing because the next line sets it to 0, why set it to 0 if it's already 0 :P

impeller/renderer/backend/vulkan/gpu_tracer_vk.cc
```diff
@@ -5,127 +5,155 @@
...
+    }
+    trace_states_[current_state_].query_pool = std::move(pool);
+    buffer.resetQueryPool(trace_states_[current_state_].query_pool.get(), 0,
+                          kPoolSize);
+  }
+
+  // We size the query pool to kPoolSize, but Flutter applications can create an
+  // unbounded amount of work per frame. If we encounter this, stop recording
+  // cmds.
+  if (state.current_index >= kPoolSize) {
```
matanlurey, line 84: Do you expect this to be common?

I wonder if it makes sense to at least log that recording did not happen.

impeller/renderer/backend/vulkan/gpu_tracer_vk.cc
```diff
@@ -5,127 +5,155 @@
...
+  state.pending_buffers = 0;
+  state.current_index = 0;
+  state.contains_failure = false;
+  in_frame_ = false;
 }
 
-void GPUTracerVK::RecordEndFrameTime() {
-  if (!valid_ || !started_frame_) {
+void GPUTracerVK::RecordCmdBufferStart(const vk::CommandBuffer& buffer) {
+  if (!valid_ || !in_frame_) {
```
matanlurey, line 58: Is the `!in_frame_` guard related to multiple concurrent frames?

If so can we add that as a 1-line comment. If not should it be DCHECK'd instead?

impeller/renderer/backend/vulkan/gpu_tracer_vk.cc
```diff
@@ -5,127 +5,155 @@
...
-    VALIDATION_LOG << "GPUTracerVK failed to record frame end time.";
+  buffer.writeTimestamp(vk::PipelineStageFlagBits::eTopOfPipe,
+                        trace_states_[current_state_].query_pool.get(),
+                        state.current_index);
+  state.pending_buffers += 1;
+  state.current_index += 1;
+}
+
+size_t GPUTracerVK::RecordCmdBufferEnd(const vk::CommandBuffer& buffer) {
+  if (!valid_ || !in_frame_) {
```
matanlurey, line 96: (Ditto above, though if no action is needed then disregard)

impeller/renderer/backend/vulkan/gpu_tracer_vk.cc
```diff
@@ -5,127 +5,155 @@
...
+  }
+  Lock lock(trace_state_mutex_);
+  GPUTraceState& state = trace_states_[frame_index];
+  if (state.pending_buffers == 0) {
+    return;
+  }
+  state.contains_failure = !success;
+  state.pending_buffers -= 1;
+
+  if (state.pending_buffers == 0 && !state.contains_failure) {
```
matanlurey, line 125: Explain in 1-line what you told me offline re: filtering out failures

impeller/renderer/backend/vulkan/test/mock_vulkan.cc
```diff
@@ -153,6 +157,7 @@ void vkGetPhysicalDeviceProperties(VkPhysicalDevice physicalDevice,
       static_cast<VkSampleCountFlags>(VK_SAMPLE_COUNT_1_BIT |
                                       VK_SAMPLE_COUNT_4_BIT);
   pProperties->limits.maxImageDimension2D = 4096;
+  pProperties->limits.timestampPeriod = 1;
```
matanlurey, line 160: (Just a comment)

We could make this configurable if you want to ensure timestampPeriod of 0 doesn't crash.

impeller/renderer/backend/vulkan/test/gpu_tracer_unittests.cc
matanlurey, line 1: As mentioned offline, I think given we're not sure this will stay landed (performance reasons, or otherwise), more tests aren't needed at this point as long as it "works on my machine". However I'd add a TODO for at least the following unit tests if this stays landed:

_These are just suggestions, feel free to tweak wording_

1. Success case where a frame is captured
2. Success case where MAX_FRAMES are captured
3. Success case where > MAX_FRAMES are captured and it wraps around (?)
4. Assert query pool is only initialized on the first frame
5. Failure case for unbounded amount of frames
6. Assert that contains_failure captures are dropped
7. Assert that your logic for longest frame WAI (i.e. multiple concurrent frames captures the right values)

Also should be asserting the final call to FML_TRACE_COUNTEr is actually recording the expected value.

matanlurey (approved):
Approved.

I'd file a follow-up issue for any comments/tests that are worth adding after this sticks.

impeller/renderer/backend/vulkan/gpu_tracer_vk.cc
jonahwilliams: This is more or less just for testing. Update to IsEnabled SGTM

jonahwilliams: Done

impeller/renderer/backend/vulkan/gpu_tracer_vk.cc
jonahwilliams: Done, and added more comments to the in_frame_ field

impeller/renderer/backend/vulkan/gpu_tracer_vk.cc
jonahwilliams: Done

impeller/renderer/backend/vulkan/gpu_tracer_vk.cc
jonahwilliams: No, I think this would be extremely rare but adapting the tracer to work with arbitrarily resizing query pools is much harder so punting for now.

jonahwilliams:
Fixed an issue where we could partially record the mipmap cmd buffers. We now record the raster thread id and only permit cmd buffers recorded from the raster thread (in additon to the thread boundary checks).

PR #46973, opened 2023-10-16
Include symbol maps in CanvasKit and Skwasm builds.

eyebrowsoffire:
This should help with https://github.com/flutter/flutter/issues/122834

Emitting DWARF data or source maps actually significantly changes the size of the build, since some binaryen optimizations must actually be skipped when producing either of those. So we can emit symbol maps for now, which don't affect the size of the actual wasm output. With these, we can at least manually deobfuscate stack traces.


yjbanov (approved):
lgtm

PR #46998, opened 2023-10-17
WIP - [Windows] Expose channel buffers 'resize' and 'overflow' control commands

bleroux:
## Description

This PR is a WIP PR to add two helper functions to invoke the 'resize' and 'overflow' commands exposed by the control channel.
See:

## Related Issue

Windows implementation for https://github.com/flutter/flutter/issues/132386

## Tests

To do

### DIFF
--- a/shell/platform/common/client_wrapper/basic_message_channel_unittests.cc
+++ b/shell/platform/common/client_wrapper/basic_message_channel_unittests.cc
@@ -28,6 +28,10 @@ class TestBinaryMessenger : public BinaryMessenger {
     last_message_handler_ = handler;
   }
 
+  void Resize(const std::string& channel, int64_t newSize) override {}
+
+  void SetWarnsOnOverflow(const std::string& channel, bool warns) override {}
+
   std::string last_message_handler_channel() {
     return last_message_handler_channel_;
   }
--- a/shell/platform/common/client_wrapper/binary_messenger_impl.h
+++ b/shell/platform/common/client_wrapper/binary_messenger_impl.h
@@ -36,6 +36,12 @@ class BinaryMessengerImpl : public BinaryMessenger {
   void SetMessageHandler(const std::string& channel,
                          BinaryMessageHandler handler) override;
 
+  // |flutter::BinaryMessenger|
+  void Resize(const std::string& channel, int64_t newSize) override;
+
+  // |flutter::BinaryMessenger|
+  void SetWarnsOnOverflow(const std::string& channel, bool warns) override;
+
  private:
   // Handle for interacting with the C API.
   FlutterDesktopMessengerRef messenger_;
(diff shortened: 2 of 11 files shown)


bleroux:
@stuartmorgan 
While working on the Windows implementation for https://github.com/flutter/flutter/issues/132386, I tried to figure out where to add the helpers functions.

On the other platforms there were somewhat obvious place to add this:
- iOS/macOS: https://github.com/flutter/engine/pull/44848
- Linux: https://github.com/flutter/engine/pull/44636
- Android: https://github.com/flutter/engine/pull/44434

In this WIP PR (to be completed), I choose to add them to the core BinaryMessenger implementation, I'm not sure it is the best place.
I'm very interested in your view on this.

stuartmorgan:
We try to keep the channel API as consistent as possible across the platforms, so I would expect these to be exposed on the same conceptual object(s) as on the other platforms.

bleroux:
> We try to keep the channel API as consistent as possible across the platforms, so I would expect these to be exposed on the same conceptual object(s) as on the other platforms.

I agree and i tried to find such a layer but it seems that for the Windows implementation it does not exist (probably because it was not required as the underlaying C++ API can be called directly).

For instance, when generating a plugin project for Windows, the channel and embedder implementations are directly exposed:

```cpp
#include <flutter/method_channel.h>
#include <flutter/plugin_registrar_windows.h>
#include <flutter/standard_method_codec.h>

...

void WindowsPlugin::RegisterWithRegistrar(
    flutter::PluginRegistrarWindows *registrar) {
  auto channel =
      std::make_unique<flutter::MethodChannel<flutter::EncodableValue>>(
          registrar->messenger(), "windows_plugin",
          &flutter::StandardMethodCodec::GetInstance());
...
```

On the other platforms, the conceptual object(s) where I exposed those helper functions were:
- Android: `BasicMessageChannel` and `MethodChannel`, Java implementation of the channel API.
- iOS/macOS: `FlutterBasicMessageChannel`, ObjectiveC++ implementation of the channel API.
- Linux: `FlBinaryMessenger`, GObject implementation of the binary messenger.

stuartmorgan:
> For instance, when generating a plugin project for Windows, the channel and embedder implementations are directly exposed

What do you mean by "directly exposed"?

> On the other platforms, the conceptual object(s) where I exposed those helper functions were:
> 
> * Android: `BasicMessageChannel` and `MethodChannel`, Java implementation of the channel API.
> * iOS/macOS: `FlutterBasicMessageChannel`, ObjectiveC++ implementation of the channel API.

How are these conceptually different from `flutter::MethodChannel`?

bleroux:
> What do you mean by "directly exposed"?

The way I understand the Android implementation is that it uses JNI to call the C++ implementation of the binary messenger (i have not look closely at the JNI calls so I might be wrong).

> How are these conceptually different from flutter::MethodChannel?

I saw `flutter::MethodChannel` as a lower level implementation, but you are right that, conceptually, there are the same.

Great, if I understand well, instead of implementing `Resize`and `SetWarnsOnOverflow`in `flutter::BinaryMessenger`, the path for this PR is to implement them in `flutter::MethodChannel` (and probably `flutter::BasicMessageChannel`). Is it the right path?

bleroux:
Closing this WIP PR and I will file a new PR based on https://github.com/flutter/engine/pull/46998#issuecomment-1768744782

PR #47038, opened 2023-10-18
[Impeller] Daisy chain final cmd buffer fence to latch.

jonahwilliams:
Ensure that all cmd buffers/encoders/pools are tracked via a fence waiter. Removes the semaphore usage from synchronizers, and instead waits on an fml::latch which is itself signaled by the fence waiter.

Allows us to enable cmd pool disposal and reset in a worker thread.


jonahwilliams:
This ended up not being necessary

PR #47053, opened 2023-10-18
Revert "Roll buildroot to pull in removal of //tools."

zanderso:
Reverts flutter/engine#47032

This is causing the Dart -> Engine roll to fail because one of the removed scripts is used by the autoroller.


jonahwilliams (approved):
LGTM

PR #47066, opened 2023-10-18
[Impeller] Skia gold for flutter_tester dart tests.

dnfield:
This removes skips for the golden tests in `//testing/dart/canvas_test.dart` and instead passes them up to Skia gold.

Adds a utility class for dealing with Skia gold from these tests, as well as the existing fuzzy identical image comparison for tests that just want to do in memory comparison of images generated from the same test.

Removes the old golden files that were in tree.

Part of https://github.com/flutter/flutter/issues/53784


dnfield:
(rather than trying to use Platform.script, I'm adding a dart define for a skia gold working directory)

dnfield:
There's a few things that need to be fixed still:

- Need to add a dimension for single and multi-threaded
- Need to differentiate based on PID or something for when multiple tests are run concurrently
- Need to actually get Gold to process these changes. Something still isn't entirely right there. There seems ot be a strange interaction between some of the async work that gets done for goldctl and litetest though.

.ci.yaml
```diff
@@ -293,6 +293,10 @@ targets:
       add_recipes_cq: "true"
       release_build: "true"
       config_name: linux_host_engine
+      dependencies: >-
+        [
+          {"dependency": "goldctl", "version": "git_revision:f808dcff91b221ae313e540c09d79696cd08b8de"}
```
Piinks, line 298: FYI, this is getting bumped in https://github.com/flutter/engine/pull/47391 to 720a542f6fe4f92922c3b8f0fdcc4d2ac6bb83cd

dnfield: Thanks!

dnfield: Should the Linux one also be the same hash?

dnfield: I looked at the PR and see that it should be.

Piinks: Thanks!

dnfield:
This should finally be ready for review.

The root cause of issues I was having were around using `Platform.script` in some shared web/VM code, and getting into the buildroot directory instead of the engine checkout directory. This was causing images to get sent to the wrong location for Skia gold and never actually processed meaningfully.

jonahwilliams (approved):
LGTM

testing/run_tests.py
```diff
@@ -1013,7 +1018,7 @@ def run_engine_tasks_in_parallel(tasks):
   # processes launched for the queue reader and thread wakeup reader).
   #
   # See: https://bugs.python.org/issue26903
-  max_processes = multiprocessing.cpu_count()
+  max_processes = 1  # multiprocessing.cpu_count()
```
zanderso, line 1021: Is this intentional?

testing/skia_gold_client/lib/skia_gold_client.dart
```diff
@@ -419,15 +447,26 @@ class SkiaGoldClient {
     return imageBytes;
   }
 
+  String _getEngineCheckoutPath() {
```
zanderso, line 450: Can you compute this using `this.workDirectory` and the `Engine.findWithin` constructor in https://github.com/flutter/engine/blob/main/tools/pkg/engine_repo_tools/lib/engine_repo_tools.dart#L87.

testing/run_tests.py
dnfield: No, I'll revert this. This was to try to isolate an issue on ci.

testing/skia_gold_client/lib/skia_gold_client.dart
dnfield: TIL, yes this should work nicely. Trying it.

zanderso (approved):

PR #47087, opened 2023-10-19
[Windows] Fix a11y node delegate lock lifetime

cbracken:
When taking a std::shared_pointer from a weak pointer node delegate, we need to hold the shared pointer to get the ownership guarantee on the underlying object until after its last use. Previously, we never assigned to a local, thus the lifetime of the returned shared_ptr (and thus the lock on the underlying object) was zero.

## Pre-launch Checklist

- [X] I added new tests to check the change I am making or feature I am adding, or the PR is [test-exempt]. See [testing the engine] for instructions on writing and running engine tests.
- [X] I signed the [CLA].

### DIFF
--- a/shell/platform/windows/flutter_platform_node_delegate_windows.cc
+++ b/shell/platform/windows/flutter_platform_node_delegate_windows.cc
@@ -62,9 +62,10 @@ gfx::NativeViewAccessible FlutterPlatformNodeDelegateWindows::HitTestSync(
   auto bridge = bridge_.lock();
   FML_DCHECK(bridge);
   for (const ui::AXNode* child : GetAXNode()->children()) {
-    std::shared_ptr<FlutterPlatformNodeDelegateWindows> win_delegate =
-        std::static_pointer_cast<FlutterPlatformNodeDelegateWindows>(
-            bridge->GetFlutterPlatformNodeDelegateFromID(child->id()).lock());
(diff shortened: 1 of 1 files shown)


shell/platform/windows/flutter_platform_node_delegate_windows.cc
```diff
@@ -62,9 +62,10 @@ gfx::NativeViewAccessible FlutterPlatformNodeDelegateWindows::HitTestSync(
   auto bridge = bridge_.lock();
   FML_DCHECK(bridge);
   for (const ui::AXNode* child : GetAXNode()->children()) {
-    std::shared_ptr<FlutterPlatformNodeDelegateWindows> win_delegate =
-        std::static_pointer_cast<FlutterPlatformNodeDelegateWindows>(
-            bridge->GetFlutterPlatformNodeDelegateFromID(child->id()).lock());
+    std::shared_ptr<FlutterPlatformNodeDelegate> delegate =
```
yaakovschectman, line 65: Is the existence of the `delegate` variable what makes the difference? And is it only used to cast to `win_delegate`? It looks like something the compiler might optimize away before the scope actually exits anyway, so are we sure this actually makes a difference?

gspencergoog:
@cbracken says we can just close this. (Desktop Triage)

PR #47062, opened 2023-10-18
Reland 2 (part 1): Enforce the rule of calling `FlutterView.Render`

dkwingsmt:
This PR relands part of https://github.com/flutter/engine/pull/45300, which was reverted in https://github.com/flutter/engine/pull/46919 due to performance regression. 

Due to how little and trivial production code the original PR touches, I really couldn't figure out the exact line that caused it except through experimentation, which requires changes to be officially landed on the main branch. After this PR lands, I'll immediately fire a performance test.

This PR contains the `Shell` refactor of the original PR. I made a slight change where the isolate snapshot is no longer returned through return value, but the parameter, in order to avoid the overhead of assigning.

It is intentional to not contain any unit tests or other changes of the original PR. They will be landed shortly after this PR.

Part of https://github.com/flutter/flutter/issues/136826.

## Pre-launch Checklist

- [ ] I read the [Contributor Guide] and followed the process outlined there for submitting PRs.
- [ ] I read the [Tree Hygiene] wiki page, which explains my responsibilities.
- [ ] I read and followed the [Flutter Style Guide] and the [C++, Objective-C, Java style guides].
- [ ] I listed at least one issue that this PR fixes in the description above.
- [ ] I added new tests to check the change I am making or feature I am adding, or the PR is [test-exempt]. See [testing the engine] for instructions on writing and running engine tests.
- [ ] I updated/added relevant documentation (doc comments with `///`).
- [ ] I signed the [CLA].
- [ ] All existing and new tests are passing.

If you need help, consider asking for advice on the #hackers-new channel on [Discord].

[Contributor Guide]: https://github.com/flutter/flutter/wiki/Tree-hygiene#overview
[Tree Hygiene]: https://github.com/flutter/flutter/wiki/Tree-hygiene
[test-exempt]: https://github.com/flutter/flutter/wiki/Tree-hygiene#tests
[Flutter Style Guide]: https://github.com/flutter/flutter/wiki/Style-guide-for-Flutter-repo
[C++, Objective-C, Java style guides]: https://github.com/flutter/engine/blob/main/CONTRIBUTING.md#style
[testing the engine]: https://github.com/flutter/flutter/wiki/Testing-the-engine
[CLA]: https://cla.developers.google.com/
[flutter/tests]: https://github.com/flutter/tests
[breaking change policy]: https://github.com/flutter/flutter/wiki/Tree-hygiene#handling-breaking-changes
[Discord]: https://github.com/flutter/flutter/wiki/Chat

PR #47095, opened 2023-10-19
Reland 2 (part 2): Enforce the rule of calling `FlutterView.Render`

dkwingsmt:
This PR relands part of https://github.com/flutter/engine/pull/45300, which was reverted in https://github.com/flutter/engine/pull/46919 due to performance regression.

Due to how little and trivial production code the original PR touches, I really couldn't figure out the exact line that caused it except through experimentation, which requires changes to be officially landed on the main branch. After this PR lands, I'll immediately fire a performance test.

This PR contains the render rule check performed by `PlatformDispatcher` of the original PR, the remaining changes to production code besides [the part 1](https://github.com/flutter/engine/pull/47062). Since part 1 shows no regression, the changes of this PR is highly likely to be the culprit. Therefore I made some changes: The rule enforcement is no longer performed in release mode, but only in debug mode. This will cause behavior deviation between builds, but since the developer should be able to notice violation in debug mode anyway, I think this design is acceptable.

It is intentional to not contain any unit tests or other changes of the original PR. They will be landed shortly after this PR.

Part of https://github.com/flutter/flutter/issues/136826.

## Pre-launch Checklist

- [ ] I signed the [CLA].


lib/ui/window.dart
```diff
@@ -353,7 +353,18 @@ class FlutterView {
...
+    // by _debugRenderedViews being null) are ignored. See _debugRenderedViews.
+    bool validRender = true;
+    assert(() {
+      validRender = platformDispatcher._debugRenderedViews?.add(this) ?? false;
+      return true;
+    }());
+    if (validRender) {
+      _render(scene as _NativeScene);
+    }
+  }
```
loic-sharma, line 367: Summarizing conversation with @dkwingsmt: we suspect that dropping invalid renders is what caused the performance regression. This reland only drops invalid renders in debug mode, we expect that the performance test will be unaffected now. @dkwingsmt will run a perf test once this lands to verify this.

Ideally our render rule enforcement would be stricter. Ideally invalid renders would trigger assertion errors on debug mode and would be dropped in release mode. However, this causes significant unit tests failures.

Original code which would dropped invalid render calls on release mode as well: https://github.com/dkwingsmt/engine/blob/e266da1854fc9e3e972e562e9c7d97e73eb738c6/lib/ui/window.dart#L356-L363

loic-sharma (approved):
LGTM

PR #47117, opened 2023-10-19
Fix async image loading issues in skwasm.

eyebrowsoffire:
This fixes https://github.com/flutter/flutter/issues/134045

There were a few different issues here:
* We need to do our own message passing for rendering pictures. The async methods provided by emscripten have their own queue that can drain synchronously, so basically it's not guaranteed to be FIFO with other messages sent to the web worker or main thread.
* When we drop frames, we should only drop intermediate frames, so that when the rendering flurry stops that the frame that is displayed is the last one that was actually requested.
* We need to reset the GL context after lazy image creation, otherwise skia's renderer gets into a bad state for that frame.

### DIFF
--- a/lib/web_ui/lib/src/engine/scene_view.dart
+++ b/lib/web_ui/lib/src/engine/scene_view.dart
@@ -16,17 +16,17 @@ abstract class PictureRenderer {
   FutureOr<DomImageBitmap> renderPicture(ScenePicture picture);
 }
 
-class SceneRender {
-  SceneRender(this.scene, this.completer) {
+class _SceneRender {
+  _SceneRender(this.scene, this._completer) {
     scene.beginRender();
   }
 
   final EngineScene scene;
-  final Completer<void> completer;
+  final Completer<void> _completer;
 
   void done() {
     scene.endRender();
-    completer.complete();
+    _completer.complete();
   }
 }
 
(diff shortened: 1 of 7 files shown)


lib/web_ui/lib/src/engine/scene_view.dart
```diff
@@ -16,6 +16,20 @@ abstract class PictureRenderer {
   FutureOr<DomImageBitmap> renderPicture(ScenePicture picture);
 }
 
+class SceneRender {
```
kevmoo, line 19: drive by comments

Does this class need to be public? Do the fields need to be public?

eyebrowsoffire: Yeah everything can be private except the scene itself. I'll update it.

lib/web_ui/test/engine/scene_view_test.dart
```diff
@@ -101,4 +106,24 @@ void testMain() {
 
     debugOverrideDevicePixelRatio(null);
   });
+
+  test('SceneView always renders most recent picture', () async {
```
harryterkelsen, line 110: nit: maybe add a test that shows that if you queue up 2 pictures (call them A and B) to be rendered while another picture is being rendered, that picture A is cancelled in favor of picture B

harryterkelsen (approved):
LGTM

lib/web_ui/test/engine/scene_view_test.dart
eyebrowsoffire: That's sort of what this test is supposed to actually be covering, but it doesn't actually check to see if anything was canceled. I'm going to just go ahead and modify this test to more explicitly check for the cancelation part of it.

PR #47121, opened 2023-10-19
[web] Attach pointer event listeners in the bubble phase

mdebbar:
There's no reason (that we know of) to listen to pointer events during the capture phase, so let's listen on the bubble phase.

I couldn't find a reason for listening to pointer events during the capture phase. So let's listen on the bubble phase which is more natural and allows platform views to `stopPropagation` in order to prevent Flutter from receiving an event.


lib/web_ui/test/engine/pointer_binding_test.dart
```diff
@@ -477,6 +477,50 @@ void testMain() {
 
   // ALL ADAPTERS
 
+  _testEach(
```
yjbanov, line 480: Let's leave a comment explaining why we test this. IIRC, we have two use-cases:

- The native text field.
- Platform views: there are cases when a platform view wants to handle an event and not propagate it to Flutter, but because Flutter listens during the capture phase, by the time the event reaches the platform view it's too late to stop propagation.

yjbanov (approved):
LGTM with one nit

lib/web_ui/test/engine/pointer_binding_test.dart
mdebbar: Done

PR #47131, opened 2023-10-19
[Impeller] Add support for VK_EXT_blend_operation_advanced.

jonahwilliams:
Advanced blends without framebuffer fetch.


jonahwilliams:
We need to do something with VK_ACCESS_COLOR_ATTACHMENT_READ_NONCOHERENT_BIT_EXT and memory barriers ... something!

jonahwilliams:
Actually to land this one I need to set up subpass self-dependency in a similar manner to the framebuffer ext version, so i'm going to land that one first instead of working on both simultaneously

jonahwilliams:
Skia self dependency logic:

https://github.com/google/skia/blob/b7ac9da3e5ac34baf5da70b3e37cdbfd898ed54b/src/gpu/ganesh/vk/GrVkRenderPass.cpp#L174-L198

PR #47133, opened 2023-10-19
Remove unused `fbo` parameter in GLES `SetAsFramebufferAttachment()`.

matanlurey:
Trivial change that removes an unused parameter.


jonahwilliams (approved):
LGTM

PR #47142, opened 2023-10-20
Add option to save Impeller failure images in rendertests

flar:
The rendering tests currently always save the impeller failure images into a temporary directory in /tmp which is out of the way and might accumulate over time.

The images are now only saved when `--save-impeller-failures` is specified on the command line and they are now saved into a local sub-directory with multiple runs saving into new sub-directories to keep the failure images from getting confused with each other over time. The new image directories are named `./impeller_failure_images/NNNN/*.png`


jonahwilliams (approved):
LGTM

PR #47158, opened 2023-10-20
[Windows] Expose channel buffers 'resize' and 'overflow' control commands exposed by the control channel.

bleroux:
## Description

This PR adds two helper functions to invoke the 'resize' and 'overflow' commands exposed by the control channel.
See:

Implementation based on the discussion from https://github.com/flutter/engine/pull/46998.

## Related Issue

Windows implementation for https://github.com/flutter/flutter/issues/132386

## Tests

Adds 4 tests.


shell/platform/common/client_wrapper/basic_message_channel_unittests.cc
```diff
@@ -20,23 +21,39 @@ class TestBinaryMessenger : public BinaryMessenger {
   void Send(const std::string& channel,
             const uint8_t* message,
             const size_t message_size,
-            BinaryReply reply) const override {}
+            BinaryReply reply) const override {
+    send_called_ = true;
+    int length = static_cast<int>(message_size);
+    last_message_size_ = length;
+    std::vector<uint8_t> last_message(length);
+    memcpy(&last_message[0], &message[0], length * sizeof(uint8_t));
```
loic-sharma, line 29: Can we do something like this?

```suggestion
    std::vector<uint8_t> last_message(message, message + length * sizeof(uint8_t));
```

shell/platform/common/client_wrapper/basic_message_channel_unittests.cc
```diff
@@ -20,23 +21,39 @@ class TestBinaryMessenger : public BinaryMessenger {
...
+    last_message_ = last_message;
```
loic-sharma, line 30: Can we do something like this?

```suggestion
    last_message_ = std::vector<uint8_t>(message, message + length * sizeof(uint8_t));
```

shell/platform/common/client_wrapper/include/flutter/method_channel.h
```diff
@@ -13,6 +13,11 @@
 #include "method_call.h"
 #include "method_codec.h"
 #include "method_result.h"
+#include "standard_method_codec.h"
+
+static constexpr char kControlChannelName[] = "dev.flutter/channel-buffers";
+static constexpr char kResizeMethod[] = "resize";
+static constexpr char kOverflowMethod[] = "overflow";
```
loic-sharma, line 20: Nit: please move these into the `flutter` namespace

shell/platform/common/client_wrapper/include/flutter/basic_message_channel.h
```diff
@@ -11,7 +11,8 @@
 
 #include "binary_messenger.h"
 #include "message_codec.h"
-
+#include "method_channel.h"
+#include "standard_method_codec.h"
```
loic-sharma, line 15:
```suggestion
#include "standard_method_codec.h"

```

shell/platform/common/client_wrapper/basic_message_channel_unittests.cc
```diff
@@ -86,4 +103,57 @@ TEST(BasicMessageChannelTest, Unregistration) {
...
+
+  // The expected content was created from the following Dart code:
+  //   MethodCall call = MethodCall('resize', ['flutter/test',3]);
+  //   StandardMethodCodec().encodeMethodCall(call).buffer.asUint8List();
+  const int expected_message_size = 29;
+
+  EXPECT_EQ(messenger.send_called(), true);
+  EXPECT_EQ(messenger.last_message_size(), expected_message_size);
+
+  int expected[expected_message_size] = {
```
loic-sharma, line 124: Could we use `StandardMessageCodec::DecodeMessage` to deserialize to an `EncodableValue`, and then assert the `EncodableValue` is expected? Debugging bytes can be painful

shell/platform/common/client_wrapper/basic_message_channel_unittests.cc
yaakovschectman: Agreed, this doesn't look preferred.

shell/platform/common/client_wrapper/include/flutter/basic_message_channel.h
bleroux: This import was meant to be removed.

shell/platform/common/client_wrapper/basic_message_channel_unittests.cc
bleroux: I agree that relying on EncodableValue would have been more readable.
The first reason I choosed the bytes approach is that this is the way it is tested on Dart side.
And because the Dart side relies on its own deserializing logic, it turned out that testing based on the bytes array was helpful to catch potential issues due to int conversion (see https://github.com/flutter/engine/pull/44848#discussion_r1315919306).

shell/platform/common/client_wrapper/basic_message_channel_unittests.cc
```diff
@@ -20,23 +21,38 @@ class TestBinaryMessenger : public BinaryMessenger {
...
+  std::vector<uint8_t> last_message() { return last_message_; }
+
+  int last_message_size() { return last_message_size_; }
+
  private:
+  mutable bool send_called_ = false;
   std::string last_message_handler_channel_;
   BinaryMessageHandler last_message_handler_;
+  mutable std::vector<uint8_t> last_message_;
+  mutable int last_message_size_;
```
stuartmorgan, line 55: ` = 0`

stuartmorgan: Although, why do we need this variable, rather than just checking the size of `last_message_`?

shell/platform/common/client_wrapper/basic_message_channel_unittests.cc
stuartmorgan: As in the linked case, we should have a comment in the code here explaining why it's tested this way (i.e., why it's critical to generate these exact bytes).

shell/platform/common/client_wrapper/basic_message_channel_unittests.cc
```diff
@@ -86,4 +102,57 @@ TEST(BasicMessageChannelTest, Unregistration) {
...
+
+  // The expected content was created from the following Dart code:
+  //   MethodCall call = MethodCall('overflow',['flutter/test', true]);
+  //   StandardMethodCodec().encodeMethodCall(call).buffer.asUint8List();
+  const int expected_message_size = 27;
+
+  EXPECT_EQ(messenger.send_called(), true);
+  EXPECT_EQ(messenger.last_message_size(), expected_message_size);
+
+  int expected[expected_message_size] = {
```
stuartmorgan, line 150: Same here.

shell/platform/common/client_wrapper/include/flutter/basic_message_channel.h
```diff
@@ -101,6 +102,41 @@ class BasicMessageChannel {
     messenger_->SetMessageHandler(name_, std::move(binary_handler));
   }
 
+  // Adjusts the number of messages that will get buffered when sending messages
+  // to channels that aren't fully set up yet. For example, the engine isn't
+  // running yet or the channel's message handler isn't set up on the Dart side
+  // yet.
+  //
+  // |new_size] is an int because the deserialization logic handles only 32 bits
```
stuartmorgan, line 110: `int` isn't required to be 32-bit, so if the goal is to constrain it via types it should be an `int32_t`. Although I would probably just remove this comment and truncate to `int32_t` internally; we could always change the implementation if anyone ever actually needed a bigger values in the future, so clients probably don't need to know this.

shell/platform/common/client_wrapper/include/flutter/basic_message_channel.h
```diff
@@ -101,6 +102,41 @@ class BasicMessageChannel {
...
+  // values, see
+  // https://github.com/flutter/engine/blob/93e8901490e78c7ba7e319cce4470d9c6478c6dc/lib/ui/channel_buffers.dart#L495.
```
stuartmorgan, line 112: This seems like a note intended for us, rather than client-facing, so it should be in the implementation code (e.g., on line 114).

shell/platform/common/client_wrapper/include/flutter/method_channel.h
```diff
@@ -13,9 +13,14 @@
 #include "method_call.h"
 #include "method_codec.h"
 #include "method_result.h"
+#include "standard_method_codec.h"
 
 namespace flutter {
 
+static constexpr char kControlChannelName[] = "dev.flutter/channel-buffers";
+static constexpr char kResizeMethod[] = "resize";
+static constexpr char kOverflowMethod[] = "overflow";
```
stuartmorgan, line 22: Why are these public? The channel serializations should be internal implementation details.

shell/platform/common/client_wrapper/include/flutter/method_channel.h
```diff
@@ -122,6 +127,41 @@ class MethodChannel {
...
+  // https://github.com/flutter/engine/blob/93e8901490e78c7ba7e319cce4470d9c6478c6dc/lib/ui/channel_buffers.dart#L495.
+  void Resize(int new_size) {
+    auto control_channel = std::make_unique<MethodChannel<EncodableValue>>(
+        messenger_, kControlChannelName, &StandardMethodCodec::GetInstance());
+
+    control_channel->InvokeMethod(
+        kResizeMethod, std::make_unique<EncodableValue>(EncodableList{
+                           EncodableValue(name_),
+                           EncodableValue(new_size),
+                       }));
```
stuartmorgan, line 146: Why are these implemented twice, rather than once in a shared location?

stuartmorgan (changes requested):

shell/platform/common/client_wrapper/include/flutter/method_channel.h
bleroux: I moved those methods to a new superclass (same for the public constants above).
Not sure about the name for this class, I came up with ResizableChannel but we might find a better name.

bleroux: Using a superclass was a dead end, I updated the PR using an anonymous namespace instead.

shell/platform/common/client_wrapper/include/flutter/method_channel.h
```diff
@@ -13,9 +13,20 @@
...
+namespace {
+
+void ResizeChannel(BinaryMessenger* messenger, std::string name, int new_size);
+
+void SetChannelWarnsOnOverflow(BinaryMessenger* messenger,
+                               std::string name,
+                               bool warns);
+
+}  // namespace
+
```
bleroux, line 29: @stuartmorgan I really need your help here.
Using an anonymous namespace compile nicely locally and I can run unit test ( `.\out\host_debug_unopt\client_wrapper_unittests.exe`).

On CI, there is a compilation error related to the this forward reference.
The main issue is that those functions implementations call `MethodChannel`, so there is a kind of cyclic declaration issue. I thought using this forward declaration is the way to solved this, but it leads to the following error or CI:

```
../../flutter/shell/platform/common/client_wrapper/include/flutter/method_channel.h:167:6: error: function 'ResizeChannel' is not needed and will not be emitted [-Werror,-Wunneeded-internal-declaration]
  167 | void ResizeChannel(BinaryMessenger* messenger, std::string name, int new_size) {
      |      ^~~~~~~~~~~~~
../../flutter/shell/platform/common/client_wrapper/include/flutter/method_channel.h:185:6: error: function 'SetChannelWarnsOnOverflow' is not needed and will not be emitted [-Werror,-Wunneeded-internal-declaration]
  185 | void SetChannelWarnsOnOverflow(BinaryMessenger* messenger,
      |      ^~~~~~~~~~~~~~~~~~~~~~~~~
```

If I remove the forward declaration, I still get no error locally but on CI there is a compilation error on lines 140 & 148 (the helpers functions can't be found, there are declared later (lines 167 & 185).

stuartmorgan: > Using an anonymous namespace compiles nicely locally

Using an anonymous namespace in a header file is pretty much going to give good results.

If the intent is to signal that people should not call these methods directly, you can make a namespace called, e.g., `internal` that they are declared in, and comment that they should not be called from outside the file. There's no way to strictly enforce that with templated code, unfortunately.

shell/platform/common/client_wrapper/include/flutter/method_channel.h
```diff
@@ -122,12 +133,70 @@ class MethodChannel {
...
+    SetChannelWarnsOnOverflow(messenger_, name_, warns);
+  }
+
  private:
   BinaryMessenger* messenger_;
   std::string name_;
   const MethodCodec<T>* codec_;
 };
 
+namespace {
```
stuartmorgan, line 157: Everything here should be in an implementation file since it's not templated.

stuartmorgan (changes requested):

shell/platform/common/client_wrapper/include/flutter/method_channel.h
bleroux: Moving this to an implementation file works nicely. 
Everything is green now, thanks!

jonahwilliams:
Friendly ping @stuartmorgan it looks like this needs another review.

stuartmorgan (changes requested):

shell/platform/common/client_wrapper/method_channel.cc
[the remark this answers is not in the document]
bleroux: Done

shell/platform/common/client_wrapper/include/flutter/method_channel.h
```diff
@@ -13,9 +13,21 @@
 #include "method_call.h"
 #include "method_codec.h"
 #include "method_result.h"
+#include "standard_method_codec.h"
 
 namespace flutter {
 
+namespace internal {
+// Internal helpers functions used by BasicMessageChannel and MethodChannel.
```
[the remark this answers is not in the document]
bleroux: Done

shell/platform/common/client_wrapper/include/flutter/method_channel.h
```diff
@@ -13,9 +13,21 @@
 #include "method_call.h"
 #include "method_codec.h"
 #include "method_result.h"
+#include "standard_method_codec.h"
```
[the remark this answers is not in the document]
bleroux: Sorry, this was a left over from a previous commit.

shell/platform/common/client_wrapper/core_implementations.cc
```diff
@@ -164,6 +166,51 @@ void ReplyManager::SendResponseData(const std::vector<uint8_t>* data) {
 
 }  // namespace internal
 
+// ========== method_channel.h ==========
+
+namespace internal {
+
+static constexpr char kControlChannelName[] = "dev.flutter/channel-buffers";
+static constexpr char kResizeMethod[] = "resize";
+static constexpr char kOverflowMethod[] = "overflow";
```
stuartmorgan, line 175: These three lines should be in an anonymous namespace (and thus don't need `static` to have internal linkage), rather than the `internal` namespace, since they don't need to be referenced outside of this file.

shell/platform/common/client_wrapper/core_implementations.cc
```diff
@@ -164,6 +166,51 @@ void ReplyManager::SendResponseData(const std::vector<uint8_t>* data) {
...
+
+// Adjusts the number of messages that will get buffered when sending messages
+// to channels that aren't fully set up yet. For example, the engine isn't
+// running yet or the channel's message handler isn't set up on the Dart side
+// yet.
```
stuartmorgan, line 180: Here and below: per the style guide, declaration comments belong at the declaration point (which is the header), not the definition point. That's true even for things that not intended for general use, like `internal`-namespaced functions or private class methods.

shell/platform/common/client_wrapper/include/flutter/basic_message_channel.h
```diff
@@ -14,6 +14,17 @@
 
 namespace flutter {
 
+namespace internal {
+// Internal helpers functions used by BasicMessageChannel and MethodChannel.
```
stuartmorgan, line 18: Nit: helper

stuartmorgan (changes requested):

stuartmorgan (approved):
LGTM, thanks!

PR #47169, opened 2023-10-20
[web] Remove workaround for safely removing slots on Safari

mdebbar:
Now that we want to support multi-view, we need to either:
1. Remove the workaround to safely remove slot elements in Safari, or
2. Fix it to support multi-view (can't use `flutterViewEmbedder.glassPaneShadow` anymore).

According to [this comment](https://github.com/flutter/flutter/issues/85816#issuecomment-1122638147), this is the right time to remove the workaround 😄 I tested on Desktop Safari and iOS Safari and wasn't able to reproduce the crash reported in https://github.com/flutter/flutter/issues/84832.

Fixes https://github.com/flutter/flutter/issues/85816


ditman (approved):
Good riddance! LGTM!

PR #47172, opened 2023-10-20
Lower the severity of a log message

johnmccutchan:
This log message can print on the first frame of a platform view and is not an error.


gspencergoog (approved):

PR #47183, opened 2023-10-20
[Impeller] Add FilterContents::GetSourceCoverage to enable filtered saveLayer clipping.

flar:
The new method allows a caller to ask for the coverage of the input pixels required to produce an output that spans a given area. This allows the code to clip a saveLayer that has an image filter applied to it.

### DIFF
--- a/impeller/aiks/aiks_unittests.cc
+++ b/impeller/aiks/aiks_unittests.cc
@@ -2763,28 +2763,56 @@ TEST_P(AiksTest, TranslucentSaveLayerWithColorAndImageFilterDrawsCorrectly) {
 TEST_P(AiksTest, ImageFilteredSaveLayerWithUnboundedContents) {
   Canvas canvas;
 
-  auto texture = CreateTextureForFixture("airplane.jpg");
-  auto blur_filter = ImageFilter::MakeBlur(Sigma{5.0}, Sigma{5.0},
+  auto blur_filter = ImageFilter::MakeBlur(Sigma{10.0}, Sigma{10.0},
                                            FilterContents::BlurStyle::kNormal,
-                                           Entity::TileMode::kClamp);
-  auto image_source = ColorSource::MakeImage(texture, Entity::TileMode::kRepeat,
-                                             Entity::TileMode::kRepeat, {}, {});
+                                           Entity::TileMode::kDecal);
+
+  auto DrawLine = [&canvas](const Point& p0, const Point& p1, const Paint& p) {
+    auto path = PathBuilder{}
+                    .AddLine(p0, p1)
+                    .SetConvexity(Convexity::kConvex)
+                    .TakePath();
+    Paint paint = p;
+    paint.style = Paint::Style::kStroke;
+    canvas.DrawPath(path, paint);
+  };
+  // Registration marks for the edge of the SaveLayer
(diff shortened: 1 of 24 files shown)


bdero:
Sorry for the delay! Doing a lot of catchup today. At a glance this looks really great and I'm not spotting anything unexpected...

Would you mind adding a quick golden to `aiks_unittests.cc` that reproduces the broken case?

flar:
One step closer, the culling of the saveLayer is fixed, but it now paints too much of the layer it creates to the screen. The layer size was expanded to collect source pixels for the filter, but the output needs to be clipped back to the SaveLayer bounds (if it had any).

Also, the test case clearly shows that the DrawPaint inside of the SaveLayer is not rendering to the whole texture.

flar:
I seem to have gone down a route with the new method to pass along the Matrix being used, but that matrix is already installed in the filter being traversed earlier in the [`GetSubpassCoverage`](https://github.com/flar/engine/blob/83a72a6fd52595b31bf2a154d3d5272d85500b59/impeller/entity/entity_pass.cc#L195) method. I could...

- Remove the installation of the Matrix when the filter is produced [at the top of the method](https://github.com/flar/engine/blob/83a72a6fd52595b31bf2a154d3d5272d85500b59/impeller/entity/entity_pass.cc#L198) since it causes a [deep rewrite](https://github.com/flar/engine/blob/83a72a6fd52595b31bf2a154d3d5272d85500b59/impeller/entity/contents/filters/filter_contents.cc#L145) of the filter tree and is redundant with the current method signature of GetSourceCoverage.
- Remove the Matrix argument from [`GetSourceCoverage`](https://github.com/flar/engine/blob/83a72a6fd52595b31bf2a154d3d5272d85500b59/impeller/entity/contents/filters/filter_contents.cc#L221) since it is already installed in the filter chain.
- ???

I lean towards methods that carry along the data they need rather than modifying the objects to contain all of the data and having to update that frequently as it is removes temporary state from the objects, but it looks like the design of Impeller is including some of this state in the objects here and there. Not sure how to untangle this particular case.

flar:
And I still need to look into those golden failures.

flar:
I'm suddenly concerned about the new optional<Rect> functions that I added to rect.h.

There seems to be some disagreement about what a nullopt means for the coverage code. Sometimes it means "there is no restriction" and sometimes it means "there is no content" which are opposite interpretations. As a result, the interpretation these methods place on what a nullopt argument means might conflict with a given piece of code.

I think we need to untangle that and come up with a single interpretation of nullopt for a rectangle. It should perhaps be "NaN", as in "there is no answer to your question", which can happen in a couple of places - but it should never equal "empty bounds" or "infinite bounds" - we have empty rectangles and (broken) infinite rectangles to handle those cases.

bdero:
Yeah I 100% agree w.r.t. the coverage conventions, added a bug around this suggesting a convention: https://github.com/flutter/flutter/issues/137306

flar:
This is ready for review now. All of the goldens look good to me, but I left them untriaged for now so that reviewers can review them and see if they agree.

bdero (approved):
LGTM! Really appreciate all the investigation you put into all the weird behaviors along the way. This is a super important fix.

PR #47184, opened 2023-10-20
[Impeller] Fix cmd pool reset from background thread.

jonahwilliams:
Make sure to clear the buffers before assigning to the recycler, as the dtor was racing with the raster thread usage.

### DIFF
--- a/impeller/renderer/backend/vulkan/command_pool_vk.cc
+++ b/impeller/renderer/backend/vulkan/command_pool_vk.cc
@@ -40,6 +40,7 @@ class BackgroundCommandPoolVK final {
     if (!recycler) {
       return;
     }
+    buffers_.clear();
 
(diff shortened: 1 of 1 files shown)


impeller/renderer/backend/vulkan/command_pool_vk.cc
```diff
@@ -75,10 +74,8 @@ CommandPoolVK::~CommandPoolVK() {
   auto reset_pool_when_dropped = BackgroundCommandPoolVK(
       std::move(pool_), std::move(collected_buffers_), recycler);
 
-  if (kResetOnBackgroundThread) {
```
matanlurey, line 78: Kill this static

matanlurey (approved):
LGTM.

PR #47208, opened 2023-10-23
[Impeller] Vk query cleanups.

jonahwilliams:
Use RAII to more accurately capture cmd buffer start/end, with additional protections for early frame ending. This also allows us to capture the execution time of the final cmd buffer, which should fix https://github.com/flutter/flutter/issues/136971


impeller/renderer/backend/vulkan/gpu_tracer_vk.cc
```diff
@@ -36,7 +38,7 @@ bool GPUTracerVK::IsEnabled() const {
 }
 
 void GPUTracerVK::MarkFrameStart() {
-  FML_DCHECK(!in_frame_);
+  FML_CHECK(!in_frame_);
```
jonahwilliams, line 41: this should be changed back to a DCHECK

impeller/renderer/backend/vulkan/gpu_tracer_vk.h
```diff
@@ -47,20 +38,34 @@ class GPUTracerVK {
   bool IsEnabled() const;
 
  private:
+  friend struct GPUProbe;
+
+  static const constexpr size_t kTraceStatesSize = 32u;
```
jonahwilliams, line 43: I increased the size of the number of pending queries because on very slow apps I observed that we would get backed up more than the previous 16.

jonahwilliams:
These tests are failing sometimes because I'm relying on the dtor to call this final query method, but my latch is invoked by the closure that runs first.

impeller/renderer/backend/vulkan/gpu_tracer_vk.h
```diff
@@ -81,4 +86,50 @@ class GPUTracerVK {
   bool enabled_ = false;
 };
 
+struct GPUProbe {
```
matanlurey, line 89: Given this behaves more class-like, I'd just make it a class
(https://google.github.io/styleguide/cppguide.html#Structs_vs._Classes)

matanlurey (approved):

impeller/renderer/backend/vulkan/gpu_tracer_vk.h
jonahwilliams: Good call. It did start out more struct like...

PR #47240, opened 2023-10-23
Move flutter.js into the engine.

eyebrowsoffire:
This will allow us to add tooling to do some bundling/minifying of `flutter.js`, which should make this more scalable/extensible long-term.

Also, this removes a few redundant build rules that produce artifacts that the flutter tool doesn't use anymore.


lib/web_ui/flutter_js/flutter.js
ditman, line 1: In order to integrate with more tooling in the future, would it be easier that this file lived in `engine/js/flutter_js` or similar? (I know little about GN).

((I guess that if needed it could be moved later, as long as its output ends up in the same directory?))

ditman (approved):
LGTM! 

Do you have a link to the framework side of the change, or are you waiting for this to land?

eyebrowsoffire:
> Do you have a link to the framework side of the change, or are you waiting for this to land?

Draft for consuming this is here: https://github.com/flutter/flutter/pull/137113

lib/web_ui/flutter_js/flutter.js
eyebrowsoffire: The files themselves can be anywhere in the source tree, so that doesn't really matter to gn. I'm not sure where you are suggesting this should go in the tree though, do you mean inside the actual `src/flutter/lib/web_ui/lib/src/engine` directory somewhere? I'd like to avoid having non-dart files in that part of the source tree, I think.

devoncarew:
> Also, this removes a few redundant build rules that produce artifacts that the flutter tool doesn't use anymore.

DartPad was using these :) https://github.com/dart-lang/dart-pad/issues/2718; specifically, `flutter_ddc_sdk_sound.dill`.

We using it with DDC to support modular compilation in dartpad - just compiling the user's source and using pre-build JS of the framework and related libraries (https://github.com/dart-lang/dart-pad/blob/main/pkgs/dart_services/tool/grind.dart#L247).

Did this have a maintenance cost? Can it be restored? Can or should we (dartpad) build this kernel file ourselves?

devoncarew:
The build failures are available from the failed jobs of this PR: https://github.com/dart-lang/dart-pad/pull/2719.

PR #47282, opened 2023-10-24
[Impeller] Switch from `glBlitFramebuffer` to implicit MSAA resolution.

matanlurey:
Closes https://github.com/flutter/flutter/issues/137093.

This widens supports from Open GLES 3.x to Open GLES 2.x, and uses [ARM GPU Best Practices](https://developer.arm.com/documentation/101897/0301/Fragment-shading/Multisampling-for-OpenGL-ES):

> Do not use `glBlitFramebuffer()` to implement a multisample resolve.

This PR:

- Removes usage of `glBlitFramebuffer`
- Adds the capability check, `SupportsImplicitResolvingMSAA`, which is `false` outside of GLES
- Does not discard color attachments resolved by `EXT_multisampled_render_to_texture` (done implicitly)

I spoke to @jonahwilliams about the changes to the HAL, who I believe also talked to @bdero about it. The short explantation is that, with the `EXT_multisampled_render_to_texture`, we can be more efficient by letting GLES perform multisampled rendering for us (no per-sample data is written out).

See also https://registry.khronos.org/OpenGL/extensions/EXT/EXT_multisampled_render_to_texture.txt for details.


jonahwilliams (approved):
LGTM!

impeller/renderer/render_target.cc
```diff
@@ -322,6 +327,17 @@ RenderTarget RenderTarget::CreateOffscreenMSAA(
...
 
+  if (context.GetCapabilities()->SupportsImplicitResolvingMSAA()) {
+    // If implicit MSAA is supported, then the resolve texture is not needed
+    // because the multisample texture is automatically resolved. We instead
+    // provide a view of the multisample texture as the resolve texture (because
+    // the HAL does expect a resolve texture).
+    //
+    // In practice, this is used for GLES 2.0 EXT_multisampled_render_to_texture
+    // https://registry.khronos.org/OpenGL/extensions/EXT/EXT_multisampled_render_to_texture.txt
+    color0.resolve_texture = color0_msaa_tex;
```
jonahwilliams, line 338: I think we should technically create some sort of "Wrapped" texture with the same descriptor but with sample count set to 1? We might need to expand on what wrapped means. Though if it doesn't cause any problems as is maybe we just leave it.

jonahwilliams:
Actually, what you already have will work for desktop GL, so maybe leave the existing code path if there is no implict resolve support?

matanlurey:
> Actually, what you already have will work for desktop GL, so maybe leave the existing code path if there is no implict resolve support?

My understanding is this is widely supported.

I'd prefer to go back and evaluate desktop OpenGLES when that's a priority (versus leaving in code that's going to go unused for a while).

matanlurey:
@chinmaygarde Friendly ping, I know I got Jonah's LGTM but I'm changing the HAL and want you review as well.

jonahwilliams:
Yeah thats fine, but just pointing out the code you added already works

jonahwilliams:
Actually from talking with @bdero the implicit multisample render to texture might just work on desktop too, so I say rip it out.

jonahwilliams:
Here is one issue I've found so far, though it exists in both versions and solving it seems out of scope: https://github.com/flutter/flutter/issues/137301

PR #47286, opened 2023-10-24
OpenGL <4.x does not support stencil-only formats, so delete the TODO

matanlurey:
Closes https://github.com/flutter/flutter/issues/137094.


jonahwilliams (approved):
LGTM

We might need a bug to make sure we do the right thing when we have both a depth and stencil attachment in GLES; avoid creating two separate DEPTH_STENCIL attachments and instead do a similar trick to map both attachments to the same texture.

PR #47291, opened 2023-10-25
[Impeller] hacky test for non-restore stencil.

jonahwilliams:
Testing an extremely hacky way of recording clip commands for replay.


impeller/renderer/command.h
```diff
@@ -176,6 +176,8 @@ struct Command : public ResourceBinder {
   ///
   size_t instance_count = 1u;
 
+  bool affects_stencil = false;
```
bdero, line 179: haha

PR #47306, opened 2023-10-25
Cleanup Dart package dependencies a bit

zanderso:
Removes package:quiver, and an older second copy of package:file


jonahwilliams (approved):
RSLGTM

PR #47315, opened 2023-10-25
Add `KeyEventDeviceType` to `KeyData`

gspencergoog:
## Description

Before we deprecate the `RawKeyEvent` code, it needs to provide parity in functionality. One thing that is missing is the `eventSource` field from the `RawKeyEventDataAndroid` class, which provides the device type for the event.

See https://developer.android.com/reference/android/view/InputDevice#SOURCE_KEYBOARD for an example.

This PR implements that support, and sets the source to `KeyEventDeviceType.keyboard` for platforms that don't provide this information. The main thing it does is add the enum `KeyEventDeviceType`, and a new field `KeyData.deviceType`.

## Related Issues
 - https://github.com/flutter/flutter/issues/136419

## Tests
 - Updated tests to also read/write/verify this property.


lib/ui/key.dart
```diff
@@ -14,7 +14,49 @@ enum KeyEventType {
...
+      repeat => 'Key Repeat',
+    };
+  }
+}
+
+/// The source device for the key event.
+///
+/// Not all platforms supply an accurate type.
+///
+/// Defaults to [keyboard].
```
dkwingsmt, line 32: I think default value should belong to a variable/field, not a type.

gspencergoog: Yes, good point.  I already removed this comment in other places, I just missed this one.

mossmana:
@gspencergoog Before I pull down the branch and test. Do you happen to know if this change will also help to address https://github.com/flutter/flutter/issues/120351 and similar issues?

gspencergoog:
@mossmana No, that has to do with touch keyboards, which don't produce key events.  This has to do with events generated by hardware keyboards.

mossmana:
> @mossmana No, that has to do with touch keyboards, which don't produce key events. This has to do with events generated by hardware keyboards.

Interesting, I didn't realize only hardware keyboards sent key events. Based on this comment https://github.com/flutter/flutter/issues/120351#issuecomment-1633390996, the Samsung soft keyboard does appear to send key events in some cases. Perhaps, the word key event is being overloaded.

gspencergoog:
Yes, some soft keyboards also synthesize hardware key events, sometimes only for specific keys, sometimes for all keys, but it's very dependent upon the soft keyboard and its settings.

lib/ui/key.dart
```diff
@@ -27,6 +67,7 @@ class KeyData {
     required this.logical,
     required this.character,
     required this.synthesized,
+    this.deviceType = KeyEventDeviceType.keyboard,
```
gspencergoog, line 70: In order to pass the smoke test for the framework, this needs to be optional.  Should we keep it that way, just to make migration easier? I'm leaning towards "yes", although all the other properties are required.

dkwingsmt: Yeah I think it's fine. At the end of the day it's `PlatformDispatcher` that creates this object, right?

gspencergoog: Yes, usually, although you could write your own test framework or something that used it.

dkwingsmt (approved):
LGTM

shell/platform/android/io/flutter/embedding/android/KeyEmbedderResponder.java
```diff
@@ -341,12 +342,33 @@ private boolean handleEventImpl(
...
+        break;
+    }
+
     output.timestamp = event.getEventTime();
     output.type = type;
     output.logicalKey = logicalKey;
     output.physicalKey = physicalKey;
     output.character = character;
     output.synthesized = false;
+    output.deviceType = KeyData.DeviceType.kKeyboard;
```
gmackall, line 371: From Android triage (specifically https://github.com/flutter/flutter/issues/151308):

Does this line override the above switch statement in all cases?

gspencergoog: Yeah, seems like it does.  That line should probably be removed.

PR #47317, opened 2023-10-25
Optimize overlays in CanvasKit

harryterkelsen:
Optimize overlays by detecting when Flutter-rendered content intersects with platform views. Overlay canvases are only needed when the content intersects, so by checking for this, we can optimize out most overlays.

## Pre-launch Checklist

- [x] I signed the [CLA].


lib/web_ui/lib/src/engine/canvaskit/overlay_scene_optimizer.dart
```diff
@@ -0,0 +1,200 @@
...
+///   * Platform views: being HTML content that is to be composited along with
+///     the Flutter content.
+class Rendering {
+  final List<RenderingEntity> entities = <RenderingEntity>[];
+
+  void add(RenderingEntity entity) {
+    entities.add(entity);
+  }
+
+  /// Returns [true] if this is equibalent to [other] for use in rendering.
```
eyebrowsoffire, line 30: typo: equivalent

lib/web_ui/lib/src/engine/canvaskit/overlay_scene_optimizer.dart
```diff
@@ -0,0 +1,200 @@
...
+
+// Computes the bounds of the platform view from its associated parameters.
+@visibleForTesting
+ui.Rect computePlatformViewBounds(EmbeddedViewParams params) {
+  ui.Rect currentClipBounds = ui.Rect.largest;
+
+  // The transforms are in logical pixels, but we want to compute physical
+  // pixel bounds, so transform by the device pixel ratio.
+  final double scale = EngineFlutterDisplay.instance.devicePixelRatio;
+  Matrix4 currentTransform = Matrix4.diagonal3Values(scale, scale, 1);
```
eyebrowsoffire, line 111: Correct me if I'm wrong on my thinking on this, but it seems like this should go at the end of the transformation chain rather than the start. Since, as your comment above states, the transforms are in logical pixels, it seems like multiplying by the scaling factor *first* will actually treat all these transforms as if they are in physical pixels. For example, if you had a 3:1 physical-to-logical pixel ratio, and one of your mutators is a transform that translates the view by 1 "unit", if you multiply by the scale factor first and then apply the mutator, it seems like the mutator would only actually translate the picture by 1 physical pixel or 1/3 of a logical pixel, but it probably intends to move it over 1 full logical pixel (or 3 physical pixels). I could be misunderstanding the situation though.

eyebrowsoffire (changes requested):
Mostly looks good, but I had a question about some of the matrix math when calculating the platform view bounds.

lib/web_ui/lib/src/engine/canvaskit/overlay_scene_optimizer.dart
harryterkelsen: Good catch. The way this was written was incorrect before. The transforms, clips, and platform view width and offset _are_ all in physical pixels. There is a transform that should come at the beginning of the mutator list which applies the devicePixelRatio skill. The bug before was that I was iterating over the mutators in the reverse order that I should have, so I applied the scaling myself, then un-applied it by dividing by the scale at the end. I fixed it properly by iterating over the mutators in the correct order now

harryterkelsen:
Thanks for the reviews! PTAL

lib/web_ui/lib/src/engine/canvaskit/embedded_views.dart
```diff
@@ -382,6 +383,35 @@ class HtmlViewEmbedder {
...
+      final CkPaint pictureBoundsPaint = CkPaint()
+        ..color = const ui.Color.fromARGB(100, 0, 0, 255);
+      for (final RenderingEntity entity in _activeRendering.entities) {
+        if (entity is RenderingPlatformView) {
+          if (entity.debugComputedBounds != null) {
+            boundsCanvas.drawRect(
+                entity.debugComputedBounds!, platformViewBoundsPaint);
+          }
+        } else if (entity is RenderingRenderCanvas) {
+          for (CkPicture picture in entity.pictures) {
```
eyebrowsoffire, line 405: The analyzer is complaining about this, I think this will fix it:

```suggestion
          for (final CkPicture picture in entity.pictures) {
```

eyebrowsoffire (approved):
LGTM! (Suggested a small fix for the analyzer)

zanderso:
SkiaPerf reported some benchmark improvements that look related to this PR: https://flutter-flutter-perf.skia.org/e/?keys=Xc9134affdf1a115fed5959303bc0d46f&requestType=0&selected=commit%3D39736%26name%3D%252Carch%253Dintel%252Cbranch%253Dmaster%252Cconfig%253Ddefault%252Cdevice_type%253Dnone%252Cdevice_version%253Dnone%252Chost_type%253Dlinux%252Csub_result%253Dbench_platform_view_infinite_scroll.canvaskit.totalUiFrame.average%252Ctest%253Dweb_benchmarks_canvaskit%252C&xbaroffset=39736

eyebrowsoffire:
Awesome! That is absolutely intended by this change. Excellent to see the perf improvements!

PR #47341, opened 2023-10-26
[Impeller] Fix the transform and geometry criteria for an optimization in TiledTextureContents

jason-simmons:
TiledTextureContents::RenderToSnapshot should only create a snapshot directly from the texture if the geometry is an axis-aligned rectangle.

Fixes https://github.com/flutter/flutter/issues/136504


bdero (approved):
LGTM!

PR #47347, opened 2023-10-26
[Impeller] Fix nullopt access and simplify coverage computation in GetSubpassCoverage.

bdero:
Follow up for comments in https://github.com/flutter/engine/pull/46130.

This case shouldn't actually be possible today, but we should be able to make this reasonably testable without goldens... added an issue to follow-up here: https://github.com/flutter/flutter/issues/137356
This branch noise will also melt away with: https://github.com/flutter/flutter/issues/137306

### DIFF
--- a/impeller/entity/entity_pass.cc
+++ b/impeller/entity/entity_pass.cc
@@ -2,8 +2,6 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
-// FLUTTER_NOLINT: https://github.com/flutter/flutter/issues/132129
-
 #include "impeller/entity/entity_pass.h"
 
 #include <memory>
@@ -97,19 +95,20 @@ size_t EntityPass::GetSubpassesDepth() const {
 
 std::optional<Rect> EntityPass::GetElementsCoverage(
     std::optional<Rect> coverage_limit) const {
-  std::optional<Rect> result;
+  std::optional<Rect> accumulated_coverage;
   for (const auto& element : elements_) {
-    std::optional<Rect> coverage;
+    std::optional<Rect> element_coverage;
 
     if (auto entity = std::get_if<Entity>(&element)) {
-      coverage = entity->GetCoverage();
+      element_coverage = entity->GetCoverage();
 
       // When the coverage limit is std::nullopt, that means there is no limit,
       // as opposed to empty coverage.
-      if (coverage.has_value() && coverage_limit.has_value()) {
+      if (element_coverage.has_value() && coverage_limit.has_value()) {
         const auto* filter = entity->GetContents()->AsFilter();
         if (!filter || filter->IsTranslationOnly()) {
-          coverage = coverage->Intersection(coverage_limit.value());
+          element_coverage =
+              element_coverage->Intersection(coverage_limit.value());
         }
       }
     } else if (auto subpass_ptr =
@@ -122,17 +121,18 @@ std::optional<Rect> EntityPass::GetElementsCoverage(
       // If the current pass elements have any coverage so far and there's a
       // backdrop filter, then incorporate the backdrop filter in the
       // pre-filtered coverage of the subpass.
-      if (result.has_value() && subpass.backdrop_filter_proc_) {
+      if (accumulated_coverage.has_value() && subpass.backdrop_filter_proc_) {
         std::shared_ptr<FilterContents> backdrop_filter =
-            subpass.backdrop_filter_proc_(FilterInput::Make(result.value()),
-                                          subpass.xformation_,
-                                          Entity::RenderingMode::kSubpass);
+            subpass.backdrop_filter_proc_(
+                FilterInput::Make(accumulated_coverage.value()),
+                subpass.xformation_, Entity::RenderingMode::kSubpass);
         if (backdrop_filter) {
           auto backdrop_coverage = backdrop_filter->GetCoverage({});
-          backdrop_coverage->origin += result->origin;
           if (backdrop_coverage.has_value()) {
+            backdrop_coverage->origin += accumulated_coverage->origin;
             if (unfiltered_coverage.has_value()) {
-              unfiltered_coverage = coverage->Union(*backdrop_coverage);
+              unfiltered_coverage =
+                  unfiltered_coverage->Union(*backdrop_coverage);
             } else {
               unfiltered_coverage = backdrop_coverage;
             }
@@ -161,29 +161,31 @@ std::optional<Rect> EntityPass::GetElementsCoverage(
       if (image_filter) {
         Entity subpass_entity;
         subpass_entity.SetTransformation(subpass.xformation_);
-        coverage = image_filter->GetCoverage(subpass_entity);
+        element_coverage = image_filter->GetCoverage(subpass_entity);
       } else {
-        coverage = unfiltered_coverage;
+        element_coverage = unfiltered_coverage;
       }
 
-      if (coverage.has_value() && coverage_limit.has_value() &&
+      if (element_coverage.has_value() && coverage_limit.has_value() &&
           (!image_filter || image_filter->IsTranslationOnly())) {
-        coverage = coverage->Intersection(coverage_limit.value());
+        element_coverage =
+            element_coverage->Intersection(coverage_limit.value());
       }
     } else {
       FML_UNREACHABLE();
     }
 
-    if (!result.has_value() && coverage.has_value()) {
-      result = coverage;
+    if (!accumulated_coverage.has_value() && element_coverage.has_value()) {
+      accumulated_coverage = element_coverage;
       continue;
     }
-    if (!coverage.has_value()) {
+    if (!element_coverage.has_value()) {
       continue;
     }
-    result = result->Union(coverage.value());
+    accumulated_coverage =
+        accumulated_coverage->Union(element_coverage.value());
   }
-  return result;
+  return accumulated_coverage;
 }
 
 std::optional<Rect> EntityPass::GetSubpassCoverage(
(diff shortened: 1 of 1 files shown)


impeller/entity/entity_pass.cc
```diff
@@ -129,11 +129,13 @@ std::optional<Rect> EntityPass::GetElementsCoverage(
                                           Entity::RenderingMode::kSubpass);
         if (backdrop_filter) {
           auto backdrop_coverage = backdrop_filter->GetCoverage({});
-          backdrop_coverage->origin += result->origin;
           if (backdrop_coverage.has_value()) {
             if (unfiltered_coverage.has_value()) {
-              unfiltered_coverage = coverage->Union(*backdrop_coverage);
+              unfiltered_coverage = coverage.has_value()
```
zanderso, line 134: I might be missing something, but I don't see how `coverage.has_value()` could ever be true here.

impeller/entity/entity_pass.cc
```diff
@@ -129,11 +129,13 @@ std::optional<Rect> EntityPass::GetElementsCoverage(
...
+                                        ? coverage->Union(*backdrop_coverage)
```
flar, line 135: I have new functions in https://github.com/flutter/engine/pull/47183 that perform nullopt aware Union and Intersection that might simplify this code.

impeller/entity/entity_pass.cc
flar: Also, this tests `backdrop_coverage.has_value()` after it already used the `->` operator on it...

flar: Is the bug here that the calculation should have been `unfiltered_coverage->Union(*backdrop_coverage)` all along?

impeller/entity/entity_pass.cc
```diff
@@ -122,17 +121,18 @@ std::optional<Rect> EntityPass::GetElementsCoverage(
       // If the current pass elements have any coverage so far and there's a
       // backdrop filter, then incorporate the backdrop filter in the
       // pre-filtered coverage of the subpass.
-      if (result.has_value() && subpass.backdrop_filter_proc_) {
+      if (accumulated_coverage.has_value() && subpass.backdrop_filter_proc_) {
```
flar, line 124: Why is this based on `accumulated_coverage`? If there is a backdrop filter on one of the elements, then the subpass coverage should be automatically expanded to the coverage_limit, shouldn't it?

impeller/entity/entity_pass.cc
```diff
@@ -122,17 +121,18 @@ std::optional<Rect> EntityPass::GetElementsCoverage(
...
-                                          subpass.xformation_,
-                                          Entity::RenderingMode::kSubpass);
+            subpass.backdrop_filter_proc_(
+                FilterInput::Make(accumulated_coverage.value()),
+                subpass.xformation_, Entity::RenderingMode::kSubpass);
         if (backdrop_filter) {
           auto backdrop_coverage = backdrop_filter->GetCoverage({});
-          backdrop_coverage->origin += result->origin;
           if (backdrop_coverage.has_value()) {
+            backdrop_coverage->origin += accumulated_coverage->origin;
```
flar, line 132: I'm not sure why this is done...? A BDF is independent of the prior content in the pass.

impeller/entity/entity_pass.cc
```diff
@@ -122,17 +121,18 @@ std::optional<Rect> EntityPass::GetElementsCoverage(
...
             if (unfiltered_coverage.has_value()) {
-              unfiltered_coverage = coverage->Union(*backdrop_coverage);
+              unfiltered_coverage =
+                  unfiltered_coverage->Union(*backdrop_coverage);
```
flar, line 135: This if/else can be replaced with `unfiltered_coverage = Union(backdrop_coverage.value(), unfiltered_coverage)`

impeller/entity/entity_pass.cc
```diff
@@ -161,29 +161,31 @@ std::optional<Rect> EntityPass::GetElementsCoverage(
...
+      accumulated_coverage = element_coverage;
       continue;
     }
-    if (!coverage.has_value()) {
+    if (!element_coverage.has_value()) {
       continue;
     }
-    result = result->Union(coverage.value());
+    accumulated_coverage =
+        accumulated_coverage->Union(element_coverage.value());
```
flar, line 186: All of these lines (178-187) => `accumulated_coverage = Union(accumulated_coverage, element_coverage)`

impeller/entity/entity_pass.cc
```diff
@@ -161,29 +161,31 @@ std::optional<Rect> EntityPass::GetElementsCoverage(
...
       } else {
-        coverage = unfiltered_coverage;
+        element_coverage = unfiltered_coverage;
       }
 
-      if (coverage.has_value() && coverage_limit.has_value() &&
+      if (element_coverage.has_value() && coverage_limit.has_value() &&
           (!image_filter || image_filter->IsTranslationOnly())) {
-        coverage = coverage->Intersection(coverage_limit.value());
+        element_coverage =
```
flar, line 171: My head is spinning a little here - why does image_filter disqualify this intersection?

flar (changes requested):

impeller/entity/entity_pass.cc
bdero: I think this is a bug and removed it... Let's see those goldens. The filter's coverage is computed above with an identity transform, but the placeholder rect given to the transform should already contain the offset.

impeller/entity/entity_pass.cc
bdero: Done.

impeller/entity/entity_pass.cc
bdero: Done.

impeller/entity/entity_pass.cc
bdero: We shouldn't need to guard against this anymore now that the coverage_limit incorporates image filters as we walk the subpasses (not that it was a correct fix to begin with). Removed it.

bdero:
Whoops, didn't push my changes until now 😅

zanderso (approved):
lgtm if clang-tidy is happy and @flar's comments are addressed.

impeller/entity/entity_pass.cc
bdero: Fixed, yeah this was supposed to union with unfiltered_coverage.

impeller/entity/entity_pass.cc
flar: Oh, I'm forgetting that BDF only operates on the current layer so it will depend on the accumulated coverage...

In the DiffContext partial repaint code BDF always accumulates the coverage_limit which is probably conservative for one that appears inside another layer (or one that appears with no previous "clear" command).

impeller/entity/entity_pass.cc
```diff
@@ -122,20 +122,16 @@ std::optional<Rect> EntityPass::GetElementsCoverage(
...
           auto backdrop_coverage = backdrop_filter->GetCoverage({});
-          backdrop_coverage->origin += result->origin;
           if (backdrop_coverage.has_value()) {
-            if (unfiltered_coverage.has_value()) {
-              unfiltered_coverage = coverage->Union(*backdrop_coverage);
-            } else {
-              unfiltered_coverage = backdrop_coverage;
-            }
+            unfiltered_coverage =
+                Union(backdrop_coverage.value(), unfiltered_coverage);
```
flar, line 134: This can be further simplified to `Union(unfiltered_coverage, backdrop_coverage)`

flar: (No need to test `backdrop_coverage.has_value()` around it as the `Union` overrides will do that for you.)

impeller/entity/entity_pass.cc
```diff
@@ -760,13 +746,17 @@ bool EntityPass::RenderElement(Entity& element_entity,
...
+      // else case and let std::nullopt get assigned as the coverage hint.
+      current_clip_coverage.has_value()) {
     // If the element already has a coverage hint (because its an advanced
     // blend), then we need to intersect the clip coverage hint with the
     // existing coverage hint.
     element_entity.GetContents()->SetCoverageHint(
-        current_clip_coverage->Intersection(
-            element_entity.GetContents()->GetCoverageHint().value()));
+        element_coverage_hint.value().Intersection(
+            current_clip_coverage.value()));
```
flar, line 759: Can all of this be replaced by

`element_entity.GetContents()->SetCoverageHint(Intersection(element_coverage_hint, current_clip_coverage)`?

flar (changes requested):

impeller/entity/entity_pass.cc
bdero: Done.

impeller/entity/entity_pass.cc
bdero: Sure can. Done.

bdero:
Appreciate you taking a look for spots we can simplify with the new utils. I think it removes quite a bit of cognitive overhead while reading through this part of the codebase.

impeller/entity/entity_pass.cc
```diff
@@ -97,19 +96,20 @@ size_t EntityPass::GetSubpassesDepth() const {
...
 
       // When the coverage limit is std::nullopt, that means there is no limit,
       // as opposed to empty coverage.
-      if (coverage.has_value() && coverage_limit.has_value()) {
+      if (element_coverage.has_value() && coverage_limit.has_value()) {
         const auto* filter = entity->GetContents()->AsFilter();
         if (!filter || filter->IsTranslationOnly()) {
-          coverage = coverage->Intersection(coverage_limit.value());
+          element_coverage =
+              element_coverage->Intersection(coverage_limit.value());
```
flar, line 112: This looks like a good place to use filter->GetSourceCoverage, maybe as a follow-on?

flar (approved):
Looks good. We should look at using GetSourceCoverage in there at some point.

impeller/entity/entity_pass.cc
bdero: Yes!

PR #47360, opened 2023-10-26
[web] fix clicks on merged semantic nodes (attempt #2)

yjbanov:
This relands https://github.com/flutter/engine/pull/43620 with a fix for nested tappable nodes. The first PR introduced this regression: https://github.com/flutter/flutter/issues/134842.

This PR includes the original PR and a fix for the regression. The fix is to call `stopPropagation` on the "click" event so that it is not handled by the ancestor if the child has already decided to send a `SemanticsAction.tap` to the framework. This ensures that there cannot be more than one `SemanticsAction.tap` sent to the framework.

Fixes https://github.com/flutter/flutter/issues/134842

### DIFF
--- a/lib/web_ui/lib/src/engine/pointer_binding.dart
+++ b/lib/web_ui/lib/src/engine/pointer_binding.dart
@@ -272,8 +272,7 @@ class ClickDebouncer {
       // recently and if the node is currently listening to event, forward to
       // the framework.
       if (isListening && _shouldSendClickEventToFramework(click)) {
-        EnginePlatformDispatcher.instance.invokeOnSemanticsAction(
-            semanticsNodeId, ui.SemanticsAction.tap, null);
+        _sendSemanticsTapToFramework(click, semanticsNodeId);
       }
       return;
     }
@@ -285,14 +284,27 @@ class ClickDebouncer {
       final DebounceState state = _state!;
       _state = null;
       state.timer.cancel();
-      EnginePlatformDispatcher.instance.invokeOnSemanticsAction(
-          semanticsNodeId, ui.SemanticsAction.tap, null);
+      _sendSemanticsTapToFramework(click, semanticsNodeId);
     } else {
       // The semantic node is not listening to taps. Flush the pointer events
       // for the framework to figure out what to do with them. It's possible
       // the framework is interested in gestures other than taps.
       _flush();
     }
+  }
+
+  void _sendSemanticsTapToFramework(DomEvent click, int semanticsNodeId) {
+    // Tappable nodes can be nested inside other tappable nodes. If a click
+    // lands on an inner element and is allowed to propagate, it will also
+    // land on the ancestor tappable, leading to both the descendant and the
+    // ancestor sending SemanticsAction.tap to the framework, creating a double
+    // tap/click, which is wrong. More details:
+    //
+    // https://github.com/flutter/flutter/issues/134842
+    click.stopPropagation();
+
+    EnginePlatformDispatcher.instance.invokeOnSemanticsAction(
+        semanticsNodeId, ui.SemanticsAction.tap, null);
   }
 
   void _startDebouncing(DomEvent event, List<ui.PointerData> data) {
--- a/lib/web_ui/lib/src/engine/pointer_binding.dart
+++ b/lib/web_ui/lib/src/engine/pointer_binding.dart
@@ -2,6 +2,7 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
+import 'dart:async';
 import 'dart:js_interop';
 import 'dart:math' as math;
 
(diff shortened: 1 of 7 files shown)


ditman (approved):
LGTM.

Added a small note in case the EventTarget generating the click/tap events may have multiple listeners (not just bubble up)

lib/web_ui/lib/src/engine/pointer_binding.dart
```diff
@@ -295,6 +293,20 @@ class ClickDebouncer {
...
 
+  void _sendSemanticsTapToFramework(DomEvent click, int semanticsNodeId) {
+    // Tappable nodes can be nested inside other tappable nodes. If a click
+    // lands on an inner element and is allowed to propagate, it will also
+    // land on the ancestor tappable, leading to both the descendant and the
+    // ancestor sending SemanticsAction.tap to the framework, creating a double
+    // tap/click, which is wrong. More details:
+    //
+    // https://github.com/flutter/flutter/issues/134842
+    click.stopPropagation();
```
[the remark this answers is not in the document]
yjbanov: Huh, TIL! Looks like it's a [standard API](https://developer.mozilla.org/en-US/docs/Web/API/Event/stopImmediatePropagation), not just jQuery. In this case we don't have multiple listeners on the same element. All the elements in question are created by the engine and we are attaching just one click listener (unless there's a bug, but that's a different story). So we only need to stop propagation to ancestor elements. Do you think we should still do something like this in this case?

lib/web_ui/lib/src/engine/pointer_binding.dart
```diff
@@ -156,20 +160,259 @@ class PointerBinding {
...
+
+  /// Notifies the debouncer of the browser-detected "click" DOM event.
+  ///
+  /// Forwards the event to the framework, unless it is deduplicated because
+  /// the corresponding pointer down/up events were recently flushed to the
+  /// framework already.
+  void onClick(DomEvent click, int semanticsNodeId, bool isListening) {
+    assert(click.type == 'click');
+
+    if (!isDebouncing) {
```
p-mazhnik, line 269: I wonder if it is possible that `isDebouncing` is `true` but `_state.target` is not equal to `click.target`

lib/web_ui/lib/src/engine/pointer_binding.dart
```diff
@@ -156,20 +160,259 @@ class PointerBinding {
...
+        // clicking, trackpad tapping and clicking, as well as the following
+        // screen readers: TalkBack on Android, VoiceOver on macOS, Narrator/
+        // NVDA/JAWS on Windows. 200ms seemed to hit the sweet spot by
+        // satisfying the following:
+        //   * It was short enough that delaying the `pointerdown` still allowed
+        //     drag gestures to begin reasonably soon (e.g. scrolling).
+        //   * It was long enough to register taps and clicks.
+        //   * It was successful at detecting taps generated by all tested
+        //     screen readers.
+        timer: Timer(const Duration(milliseconds: 200), _onTimerExpired),
```
p-mazhnik, line 334: This is not working for long taps (issue https://github.com/flutter/flutter/issues/147050). 

In long tap case, it is possible that `pointerup` event is fired after 200 milliseconds since `pointerdown`, and  as a result `isDebounce` will be `false` when `click` event is received, leading to duplication. 

In the previous solution (that is still exists in parallel), timer was added after `pointerup` (and others) as well, so it is probably should fix the issue
https://github.com/flutter/engine/blob/e5215e6854bc61cf8d5bae48715d73293bc1f91c/lib/web_ui/lib/src/engine/semantics/semantics.dart#L2232-L2250
https://github.com/flutter/engine/blob/e5215e6854bc61cf8d5bae48715d73293bc1f91c/lib/web_ui/lib/src/engine/semantics/semantics.dart#L2184-L2195

p-mazhnik (commented):
I revisited this PR after encountering issues with semantics on the web. Although it’s 8 months old, the code is still relevant, so I want to document here what I’ve found

lib/web_ui/lib/src/engine/pointer_binding.dart
```diff
@@ -156,20 +160,259 @@ class PointerBinding {
...
+        queue: <QueuedEvent>[(
+          event: event,
+          timeStamp: _BaseAdapter._eventTimeStampToDuration(event.timeStamp!),
+          data: data,
+        )],
+      );
+    } else {
+      // The event landed on an non-tappable target. Assume this won't lead to
+      // double clicks and forward the event to the framework.
+      _sendToFramework(event, data);
```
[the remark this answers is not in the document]
yjbanov: Yeah, I fixed one variant of this in https://github.com/flutter/engine/pull/53694. Just using the `event.target` is probably naive. It should start with `event.target` and then walk up to the nearest tappable ancestor.

lib/web_ui/lib/src/engine/pointer_binding.dart
yjbanov: This is correct, and it's on my radar to fix. This method was enough for the screen reader case. However, we're trying to make it so semantics can be always enabled, and therefore it must also work for all interactions without the screen reader. This code needs a revision.

lib/web_ui/lib/src/engine/pointer_binding.dart
yjbanov: Due to the problem you pointed out above (the sketchy target detection), it seems quite possible.

PR #47364, opened 2023-10-26
[Impeller] Enable GLES MSAA only if the multisampled_render_to_texture2 extension is available

jason-simmons:
The GLES MSAA implementation may need to call FramebufferTexture2DMultisampleEXT for the stencil attachment.

See https://github.com/flutter/flutter/issues/137252


jonahwilliams (approved):
LGTM

PR #47391, opened 2023-10-27
Bump goldctl

Piinks:
Related to https://github.com/flutter/flutter/issues/137177

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/.ci.yaml
+++ b/.ci.yaml
@@ -441,7 +441,7 @@ targets:
       config_name: mac_host_engine
       dependencies: >-
         [
-          {"dependency": "goldctl", "version": "git_revision:f808dcff91b221ae313e540c09d79696cd08b8de"}
+          {"dependency": "goldctl", "version": "git_revision:720a542f6fe4f92922c3b8f0fdcc4d2ac6bb83cd"}
         ]
       $flutter/osx_sdk : >-
         {
--- a/lib/web_ui/README.md
+++ b/lib/web_ui/README.md
@@ -206,7 +206,7 @@ After rolling Chrome and/or Firefox, also update the CI dependencies in
         [
           {"dependency": "chrome_and_driver", "version": "version:107.0"},
           {"dependency": "firefox", "version": "version:83.0"},
-          {"dependency": "goldctl", "version": "git_revision:3a77d0b12c697a840ca0c7705208e8622dc94603"}
+          {"dependency": "goldctl", "version": "git_revision:720a542f6fe4f92922c3b8f0fdcc4d2ac6bb83cd"}
         ]
 ```
 
--- a/lib/web_ui/dev/generate_builder_json.dart
+++ b/lib/web_ui/dev/generate_builder_json.dart
@@ -154,7 +154,7 @@ Iterable<dynamic> _getTestStepsForPlatform(
         'test_dependencies': <dynamic>[
           <String, dynamic>{
             'dependency': 'goldctl',
-            'version': 'git_revision:dddc0623e63150cbbafdcb273d4048f329e1dd09',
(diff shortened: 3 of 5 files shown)


lib/web_ui/dev/generate_builder_json.dart
```diff
@@ -154,7 +154,7 @@ Iterable<dynamic> _getTestStepsForPlatform(
         'test_dependencies': <dynamic>[
           <String, dynamic>{
             'dependency': 'goldctl',
-            'version': 'git_revision:dddc0623e63150cbbafdcb273d4048f329e1dd09',
+            'version': 'git_revision:720a542f6fe4f92922c3b8f0fdcc4d2ac6bb83cd',
```
yjbanov, line 157: @eyebrowsoffire can we get this value from `.ci.yaml` instead of hard-coding it here?

yjbanov (approved):

lib/web_ui/dev/generate_builder_json.dart
Piinks: Yeah I think it was better when we kept this version in one single source of truth in the recipes. Between here and the ci.yaml in the framework (updated in https://github.com/flutter/flutter/pull/137441), there were more than 7 different version strings. 😕 
Tracking all these down feels like it was better when there one one place this was updated.

eyebrowsoffire: Answering the specific question, no, there is not a way to specify these from `.ci.yaml` currently. Only the orchestrator CI step is actually specified in the `.ci.yaml` file. This generates the builder JSON file that specifies the `engine_v2` sub-builds, so we have to specify that here. I do understand the distaste for the deep layering of our CI processes, but it's just the way things are at the moment.

.ci.yaml
```diff
@@ -441,7 +441,7 @@ targets:
       config_name: mac_host_engine
       dependencies: >-
         [
-          {"dependency": "goldctl", "version": "git_revision:f808dcff91b221ae313e540c09d79696cd08b8de"}
```
eyebrowsoffire, line 444: Actually this one might not be necessary at all. This task is just the `engine_v2` orchestrator task (the parent task that spawns the sub-builds). It seems unlikely that any actual tests will run in the orchestrator that use skia gold.

lib/web_ui/README.md
```diff
@@ -206,7 +206,7 @@ After rolling Chrome and/or Firefox, also update the CI dependencies in
         [
           {"dependency": "chrome_and_driver", "version": "version:107.0"},
           {"dependency": "firefox", "version": "version:83.0"},
-          {"dependency": "goldctl", "version": "git_revision:3a77d0b12c697a840ca0c7705208e8622dc94603"}
```
eyebrowsoffire, line 209: Yeah and this readme is woefully out of date, from before the `engine_v2` migration. I should update it.

lib/web_ui/dev/generate_builder_json.dart
zanderso: My recommendation for the engine repo would be to pull the CIPD package with `gclient sync` by listing it in the DEPS file. Then at least for the engine repo, the hash would be in only one place. The tricky bit would be that the scripts that use goldctl would need to be updated to refer to the destination of the CIPD package in the engine tree.

Piinks: I think that sounds good. I am not at all familiar with the engine implementation of Gold, maybe @mdebbar can advise on changing it to use `gclient sync`? I'd like to land this for now since it is trying to resolve a high number of flakes we are seeing from the service.

PR #47405, opened 2023-10-27
[Impeller] started scaling the gaussian blur sigma to match skia output

gaaclarke:
fixes https://github.com/flutter/flutter/issues/127770

This will help performance as well since our kernels will be smaller.  The justification for scaling the sigma is that we need to account in the math the scaling down that we are doing of the input.

Here are the side-by-side comparisons with this math and skia, impeller is always on the right:

## sigma 10

blur_10

## sigma 75

blur_75

## sigma 164 (sigma in bug)
blur_164

## sigma 400
blur_400

## sigma 1000
blur_1000

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/impeller/entity/contents/filters/directional_gaussian_blur_filter_contents.cc
+++ b/impeller/entity/contents/filters/directional_gaussian_blur_filter_contents.cc
@@ -23,6 +23,23 @@
 
 namespace impeller {
 
+// This function was derived with polynomial regression when comparing the
+// results with Skia.  Changing the curve below should invalidate this.
+//
+// The following data points were used:
+//   0 | 1
+//  75 | 0.8
+// 150 | 0.5
+// 300 | 0.22
+// 400 | 0.2
+// 500 | 0.15
+Sigma ScaleSigma(Sigma sigma) {
+  // Limit the kernel size to 1000x1000 pixels, like Skia does.
+  Scalar clamped = std::min(sigma.sigma, 500.0f);
+  Scalar scalar = 1.02 - 3.89e-3 * clamped + 4.36e-06 * clamped * clamped;
+  return Sigma(clamped * scalar);
+}
+
 DirectionalGaussianBlurFilterContents::DirectionalGaussianBlurFilterContents() =
     default;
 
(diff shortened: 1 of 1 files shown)


gaaclarke:
I'm not sure what sort of test we should add.  This will probably be picked up in existing goldens.

impeller/entity/contents/filters/directional_gaussian_blur_filter_contents.cc
```diff
@@ -23,6 +23,23 @@
...
+//   0 | 1
+//  75 | 0.8
+// 150 | 0.5
+// 300 | 0.22
+// 400 | 0.2
+// 500 | 0.15
+Sigma ScaleSigma(Sigma sigma) {
+  // Limit the kernel size to 1000x1000 pixels, like Skia does.
+  Scalar clamped = std::min(sigma.sigma, 500.0f);
+  Scalar scalar = 1.02 - 3.89e-3 * clamped + 4.36e-06 * clamped * clamped;
```
jonahwilliams, line 39: TIL you can use scientific notation in C++

jonahwilliams (approved):
IIRC from reading the Skia code, it goes something like

- If the Sigma is <= 4, leave it alone.
- Otherwise scale the Sigma down to 4, and use that same scaling factor to adjust the texture size.

So ScaleSigma is related to the curve we use below to scale the texture size, as in the both need to approximately agree.

If this looks like an improvement overall, then LGTM, but we might end up adjusting the whole thing anyway

gaaclarke:
Here is the difference between skia and impeller with sigma 164 before this change, just to show this is an obvious improvement.

Screenshot 2023-10-27 at 3 58 30 PM

jonahwilliams:
Yeah that looks way better

bdero (approved):
Yes! Fix the look of your blur with this one line trick. :)

The scaledown purely _adds_ to the blurriness. Our blurs already look pretty good, so just remap it to match Skia as a holdover.

We could always adjust this later with a more closely fitted curve, but this parabola looks pretty close!

gaaclarke:
I looked through the goldens and they all seemed kosher.  Since there is existing coverage, I think I'm going to avoid adding another golden test.  `ClippedBlurFilterRendersCorrectlyInteractive` does have a large sigma.

PR #47411, opened 2023-10-28
[Impeller] Rect getters

flar:
This is a precursor change that will allow us to migrate all of the Impeller sources off of accessing the `origin` and `size` fields in preparation for improving the implementation of `Rect`.

No code has been migrated to use the new fields yet, and so this code should be completely benign as is, but there will be a number of follow-on PRs that all share this PR as a pre-requisite and so it needs to be landed independently of any of the PRs that will modify code currently in use.


bdero (approved):
lgtm!

PR #47414, opened 2023-10-28
[web] remove ulimit and add -i to felt

yjbanov:
The ulimit logic has been failing for me for months now and felt still ran fine. I think we don't need it any more.

Also add the `-i` option for incremental runs of `felt`. It causes felt to start faster because it doesn't run `pub get`. In the future, we can also run felt from the snapshot for even faster start-up.

### DIFF
--- a/lib/web_ui/dev/felt
+++ b/lib/web_ui/dev/felt
@@ -19,7 +19,7 @@ SDK_PREBUILTS_DIR="${FLUTTER_DIR}/prebuilts"
 # If set to 1 (via the -i option) will run felt incrementally.
 INCREMENTAL=0
 
-# All arguments passed to felt, expect -i, if any.
+# All arguments passed to felt, except -i, if any.
 ARGS=""
 for arg in "$@" ; do
   case "$arg" in
@@ -32,10 +32,6 @@ for arg in "$@" ; do
       ;;
   esac
 done
-
-echo $INCREMENTAL
-echo $ARGS
-exit 0
 
 if [ -z "${DART_SDK_DIR}" ]
 then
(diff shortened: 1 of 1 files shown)


ditman:
> In the future, we can also run felt from the snapshot for even faster start-up.

Didn't that feature get removed because then `felt` would break every time the dart SDK DEPS changed, right?

eyebrowsoffire (approved):
LGTM!

FWIW, I dislike the felt snapshot strategy unless we have proper change detection to invalidate the snapshot when necessary, which probably involves parsing a ninja dep file and hashing all the files, which seems like overkill. Or maybe we could flip the script somehow and use ninja to build the felt snapshot... and then felt would attempt to build itself on launch or something? Hmm... I don't know, it's tricky.

yjbanov:
Yes, it is an issue, but in the past snapshot runs were both the default _and_ broken. In this case I'm going for opt-in (i.e. "I know what I'm doing, and I always know to fallback to non-incremental build if things are going sideways).

lib/web_ui/dev/felt
```diff
@@ -14,6 +14,29 @@ ENGINE_SRC_DIR="$(dirname $(dirname $(dirname $(dirname ${FELT_DIR}))))"
 FLUTTER_DIR="${ENGINE_SRC_DIR}/flutter"
 SDK_PREBUILTS_DIR="${FLUTTER_DIR}/prebuilts"
 
+# If set to 0 (the default) will run felt non-incrementally. For example, it will
+# run pub get.
+# If set to 1 (via the -i option) will run felt incrementally.
+INCREMENTAL=0
+
+# All arguments passed to felt, expect -i, if any.
```
mdebbar, line 22: "except"?

lib/web_ui/dev/felt
```diff
@@ -14,6 +14,29 @@ ENGINE_SRC_DIR="$(dirname $(dirname $(dirname $(dirname ${FELT_DIR}))))"
...
+      ;;
+    *)
+      ARGS="${ARGS} ${arg}"
+      ;;
+  esac
+done
+
+echo $INCREMENTAL
+echo $ARGS
+exit 0
```
mdebbar, line 38: Debugging remnants?

mdebbar (approved):

PR #47430, opened 2023-10-28
[Impeller] Migrate unit tests to named Rect factories

flar:
This is a pre-requisite step to deleting the un-named Constructors for Rect and moving to named factories. The intention is both to improve code readability and so that the underlying storage information is not implied or exposed.

This PR only modifies the unit tests which will not affect actual running apps, but migrates >95% of the concrete Rect construction to the named factory methods. Use of these constructors in the rest of the code is very rare since most methods get their Rect instances from data they compute or are delivered from elsewhere. Those other instances will be eliminated more on a file by file basis to reduce the surface area of errors arising due to typos or porting mistakes.


bdero (approved):
LGTM

PR #47460, opened 2023-10-30
[web] Move scene DOM management to DomManager

mdebbar:
- Move scene insertion logic to `DomManager`.
- Add TODOs in `Renderer` subclasses (cc @harryterkelsen).


lib/web_ui/lib/src/engine/view_embedder/dom_manager.dart
```diff
@@ -63,4 +63,24 @@ class DomManager {
...
+
+  /// Inserts the [sceneElement] into the DOM.
+  ///
+  /// The [sceneElement] is inserted  as a child of the <flt-scene-host> element
+  /// inside the [renderingHost].
+  ///
+  /// If the [sceneElement] has already been inserted, this method does nothing
+  /// to avoid unnecessary DOM mutations. This is both faster and more correct,
+  /// because moving DOM nodes loses internal state, such as text selection.
+  void insertScene(DomElement sceneElement) {
```
ditman, line 79: I'm not sure `insert` is the right name (to me it means "add anywhere but keep the old stuff?).

However the only thing I can come up with is `replace` and I don't like it either. (This is private API and can be renamed later very easily so, let's let it gggo! ❄️)

ditman (approved):
This LGTM!

(QQ: Is the idea that at some point the `DomManager` will create all the DOM that is needed for Flutter, rather than taking in the flutter view embedder, hence the DEPRECATED names?)

mdebbar:
> (QQ: Is the idea that at some point the `DomManager` will create all the DOM that is needed for Flutter, rather than taking in the flutter view embedder, hence the DEPRECATED names?)

Yes. That's what I'm working towards right now. I need many other changes though, before I can move DOM creation to `DomManager`.

lib/web_ui/lib/src/engine/view_embedder/dom_manager.dart
mdebbar: Maybe `setScene`?

ditman: Yes!

lib/web_ui/lib/src/engine/canvaskit/renderer.dart
```diff
@@ -68,7 +68,9 @@ class CanvasKitRenderer implements Renderer {
     // added eagerly during initialization here and never touched, unless the
     // system is reset due to hot restart or in a test.
     _sceneHost = createDomElement('flt-scene');
-    embedder.addSceneToSceneHost(_sceneHost);
+    // TODO(harryterkelsen): Do this operation on the appropriate Flutter View.
+    final EngineFlutterView implicitView = EnginePlatformDispatcher.instance.implicitView!;
+    implicitView.dom.insertScene(_sceneHost!);
```
ditman, line 73:
```suggestion
    implicitView.dom.setScene(_sceneHost!);
```
?

harryterkelsen (approved):
LGTM

PR #47464, opened 2023-10-30
Reverts Dart to 3.3.0-53.0.dev

zanderso:
This reverts commit af01ff17284174dc645bbd4f77c857c7dfd3c07d.

Reverting for https://github.com/flutter/flutter/issues/137540


gaaclarke (approved):
Sounds good.  I would prefer a forward fix or a revert of only the analyzer, but I'm approving this if we decide this is the route we have to go.

PR #47487, opened 2023-10-30
[web] Refactor a11y announcements out of FlutterViewEmbedder

mdebbar:
- Remove a11y announcements from `FlutterViewEmbedder`.
- Simplify a11y announcements tests so they don't need `FlutterViewEmbedder` nor `DomManager`.
- Left a few a11y-multi-view TODOs (cc @yjbanov).


ditman:
The announcements host *might* be a full-blown global that all the views write to. I think one of the PRs that we got was to move the host "as lower down in the DOM as possible", so this might be something that we need to put near the end of the `body` tag.

(I am not an expert, but TL;DR: this might not need a per-view implementation!

mdebbar:
I would prefer not to change the scope of this PR if possible. This PR is about moving the ownership of the a11y announcement element from `FlutterViewEmbedder` to `DomManager` but keeping the element and its behaviors unchanged.

Making a11y announcements global (i.e. in the `<body>`) is orthogonal to what I'm trying to do in this PR.

That said, I'm working (in my head, nothing written yet) on a proposal to have a `<flutter-globals>` container that goes under `<body>` and contains all of our global stuff that doesn't make sense to put inside a view (e.g. a11y announcements, text rulers).

ditman (approved):
Ship it! I agree with having a `<flutter-globals>` area in the page, I don't see any way of escaping that.

PR #47488, opened 2023-10-30
[Impeller] null check drawable.

jonahwilliams:
Speculative fix for the first crash noted in https://github.com/flutter/flutter/issues/136628#issue-1944779964

nextDrawble can return null if the method times out.

Fixes https://github.com/flutter/flutter/issues/136525


shell/gpu/gpu_surface_metal_impeller.mm
```diff
@@ -99,6 +99,10 @@
 
   auto drawable = impeller::SurfaceMTL::GetMetalDrawableAndValidate(
```
jonahwilliams, line 100: This can be nullptr, see https://github.com/flutter/engine/blob/47465eb4fbe6e01285172e7ec358521907828c5e/impeller/renderer/backend/metal/surface_mtl.mm#L26-L39

zanderso:
Should this have a test, or does it need a test exemption?

jason-simmons (approved):

jonahwilliams:
Based on our discussion this week, this needs a test and I plan on adding one ... a bit later. :)

jonahwilliams:
I have added tests.

shell/gpu/BUILD.gn
```diff
@@ -88,3 +88,24 @@ source_set("gpu_surface_metal") {
...
+
+executable("gpu_surface_unittests") {
+  testonly = true
+  sources = []
+
+  if (is_mac) {
+    sources += [ "gpu_surface_metal_impeller_unittests.mm" ]
+  }
+
+  if (is_mac) {
```
zanderso, line 100: Consider combining back-to-back `if (is_mac)`

zanderso (approved):

shell/gpu/BUILD.gn
jonahwilliams: Done

matanlurey (approved):
L GEE TEE EM

jonahwilliams:
Tests are working for real!

jonahwilliams:
nvm still not working.

PR #47489, opened 2023-10-30
[web] Move styling from FlutterViewEmbedder to StyleManager

mdebbar:
- Move default style values from `FlutterViewEmbedder` to `StyleManager`.
- Move DOM tag names from `FlutterViewEmbedder` to `DomManager`.
- Deduplicate style sheet creation code and put it in `StyleManager`.

Part of Part of https://github.com/flutter/flutter/issues/134443


lib/web_ui/lib/src/engine/text/ruler.dart
```diff
@@ -24,22 +24,22 @@ String buildCssFontString({
   if (fontStyle != null) {
     result.write(fontStyle == ui.FontStyle.normal ? 'normal' : 'italic');
   } else {
-    result.write(FlutterViewEmbedder.defaultFontStyle);
+    result.write(StyleManager.defaultFontStyle);
   }
```
ditman, line 28: This is probably a left-over from the null-safety migration, but...

Can these if-elses be default values of the parameters of the `buildCssFontString` method?

So instead of having a `required ui.FontStyle?` (why make something required AND nullable?) do:

```dart
String buildCssFontString({
  required String fontFamily,
  ui.FontStyle fontStyle = StyleManager.defaultFontStyle,
  ui.FontWeight fontWeight = StyleManager.defaultFontWeight,
  double fontSize = StyleManager.defaultFontSize,
})
```

lib/web_ui/lib/src/engine/view_embedder/dom_manager.dart
```diff
@@ -4,8 +4,25 @@
 
 import 'package:ui/ui.dart' as ui;
 
+import '../display.dart';
 import '../dom.dart';
 import '../embedder.dart';
+import '../global_styles.dart';
+
+/// The tag name for the Flutter View root element.
+const String kFlutterViewTagName = 'flutter-view';
```
ditman, line 13: Should these be `static const` of `DomManager` (so they're consistent with the ones added in `StyleManager`?)

Also:

Drop the "The". Just "Facebook". It's cleaner.

Drop the `k`. Just `flutterViewTagName`. It's cleaner. 😝

lib/web_ui/lib/src/engine/view_embedder/dom_manager.dart
```diff
@@ -64,3 +81,60 @@ class DomManager {
...
+    semanticsHost.style
+      ..position = 'absolute'
+      ..transformOrigin = '0 0 0';
+    scaleSemanticsHost(semanticsHost);
+  }
+
+  /// The framework specifies semantics in physical pixels, but CSS uses
+  /// logical pixels. To compensate, an inverse scale is injected at the root
+  /// level.
+  static void scaleSemanticsHost(DomElement semanticsHost) {
```
ditman, line 135: Pass the `dpr` when calling this method, so we don't deeply access an unrelated object from here.

lib/web_ui/lib/src/engine/global_styles.dart
```diff
@@ -5,6 +5,7 @@
 import 'browser_detection.dart';
 import 'dom.dart';
 import 'text_editing/text_editing.dart';
+import 'view_embedder/dom_manager.dart';
 
 // Applies the required global CSS to an incoming [DomCSSStyleSheet] `sheet`.
 void applyGlobalCssRulesToSheet(
```
ditman, line 11: This method smells like should be a part of the `StyleManager` bag of methods.

(In fact, should this file just be renamed `style_manager.dart` and put the whole style manager here?)

lib/web_ui/lib/src/engine/text/canvas_paragraph.dart
```diff
@@ -5,9 +5,9 @@
 import 'package:ui/ui.dart' as ui;
 
 import '../dom.dart';
-import '../embedder.dart';
 import '../html/bitmap_canvas.dart';
 import '../profiler.dart';
+import '../view_embedder/dom_manager.dart';
```
ditman, line 10: (Yes, it's odd to `import dom_manager.dart` to end up using a class named `StyleManager` :P)

lib/web_ui/test/html/shaders/shader_mask_golden_test.dart
```diff
@@ -19,7 +19,7 @@ const bool debugTest = false;
 
 DomElement get sceneHost =>
     EnginePlatformDispatcher.instance.implicitView!.dom.renderingHost
-        .querySelector('flt-scene-host')!;
+        .querySelector(kSceneHostTagName)!;
```
ditman, line 22: (Note: If you notice we do `DomManager.someElement.querySelector('selector')` in production code, we probably need to cache the element in the `DomManager` as well, so we don't repeat selectors (even though the browser is probably optimizing the selector for us, so this might be a moot point))

ditman (approved):
Small comment to unify the global_styles.dart and the `StyleManager` into one big single, interrelated blob, but consider all my comments as severity "nitpick" at most!

lib/web_ui/test/html/shaders/shader_mask_golden_test.dart
mdebbar: Are you saying that we do **today** have usages of `DomManager.someElement.querySelector('selector')` in production code, and we should cache them?

Or are you saying **in the future** if we find usages of `DomManager.someElement.querySelector('selector')` in production code, then we should cache them?

(I couldn't find any production code usages in our code, so I'm going to assume you mean the latter).

lib/web_ui/lib/src/engine/text/ruler.dart
mdebbar: Nice suggestion! But I couldn't do exactly what you said because the parameter is `ui.FontStyle` but the default value is a `String` (for a good reason, I think). So I had to tweak it a bit, and it looks a lot better than it used to. See: https://github.com/flutter/engine/pull/47489/commits/c8c358dcf62e10a1148c25370fa90bf1fceecffc

lib/web_ui/lib/src/engine/view_embedder/dom_manager.dart
mdebbar: I like it!

lib/web_ui/lib/src/engine/view_embedder/dom_manager.dart
mdebbar: Done!

lib/web_ui/lib/src/engine/global_styles.dart
mdebbar: Done!

lib/web_ui/lib/src/engine/text/canvas_paragraph.dart
mdebbar: Yep!

lib/web_ui/test/html/shaders/shader_mask_golden_test.dart
ditman: > Or are you saying in the future if we find usages of DomManager.someElement.querySelector('selector') in production code, then we should cache them?

☝️ This is what I meant. Thanks for checking there were not weird usages already!!

lib/web_ui/lib/src/engine/text/ruler.dart
ditman: Ahhh very cool, I totally didn't notice they were different types! :P

lib/web_ui/lib/src/engine/view_embedder/style_manager.dart
```diff
@@ -20,7 +86,7 @@ void applyGlobalCssRulesToSheet(
 
   // Fixes #115216 by ensuring that our parameters only affect the flt-scene-host children.
   sheet.insertRule('''
-    $cssSelectorPrefix flt-scene-host {
+    $cssSelectorPrefix ${DomManager.sceneHostTagName} {
```
ditman, line 89: Fantastic!

ditman (approved):
I don't have much left to say here! A couple of minor questions about some small implementation details, but other than that: 🚀

lib/web_ui/lib/src/engine/text/paragraph.dart
```diff
@@ -738,12 +738,18 @@ class ParagraphPlaceholder {
...
+
+extension FontWeightExtension on ui.FontWeight {
+  /// Converts a [ui.FontWeight] value to its CSS equivalent.
+  String toCssString() {
+    return fontWeightIndexToCss(fontWeightIndex: index);
   }
-  return fontWeightIndexToCss(fontWeightIndex: fontWeight.index);
 }
 
 String fontWeightIndexToCss({int fontWeightIndex = 3}) {
```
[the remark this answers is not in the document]
mdebbar: This method is being used in `text_editing.dart` where the framework sends us the font weight index in a platform message.

lib/web_ui/lib/src/engine/text/paragraph.dart
```diff
@@ -738,12 +738,18 @@ class ParagraphPlaceholder {
   final ui.TextBaseline baseline;
 }
 
-/// Converts [fontWeight] to its CSS equivalent value.
-String? fontWeightToCss(ui.FontWeight? fontWeight) {
-  if (fontWeight == null) {
-    return null;
+extension FontStyleExtension on ui.FontStyle {
```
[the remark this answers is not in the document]
mdebbar: `FontStyle` is defined in the public part of the engine, so I don't want to pollute it with our private stuff.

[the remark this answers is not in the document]
ditman: Yeah, you're right, it might confuse analyzer if we modify the public api

lib/web_ui/lib/src/engine/text/paragraph.dart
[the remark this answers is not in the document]
ditman: Does this happen in many places? Shouldn't we convert the platform message to a proper `ui.FontWeight` with something like: `ui.FontWeight.values[index]`?

lib/web_ui/lib/src/engine/text/ruler.dart
```diff
@@ -21,27 +21,21 @@ String buildCssFontString({
...
-  }
+  // Font size.
+  final num cssFontSize = fontSize?.floor() ?? StyleManager.defaultFontSize;
+  result.write(cssFontSize);
   result.write('px ');
+
+  // Font family.
   result.write(canonicalizeFontFamily(fontFamily));
 
   return result.toString();
```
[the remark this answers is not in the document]
mdebbar: Yep, makes sense!

lib/web_ui/lib/src/engine/text/ruler.dart
```diff
@@ -18,33 +18,12 @@ String buildCssFontString({
   required double? fontSize,
   required String fontFamily,
 }) {
-  final StringBuffer result = StringBuffer();
+  final String cssFontStyle = fontStyle?.toCssString() ?? StyleManager.defaultFontStyle;
+  final String cssFontWeight = fontWeight?.toCssString() ?? StyleManager.defaultFontWeight;
+  final num cssFontSize = fontSize?.floor() ?? StyleManager.defaultFontSize;
```
ditman, line 23: It probably doesn't matter much, but:

```suggestion
  final int cssFontSize = fontSize?.floor() ?? StyleManager.defaultFontSize;
```

lib/web_ui/lib/src/engine/text/ruler.dart
```diff
@@ -18,33 +18,12 @@ String buildCssFontString({
...
-  if (fontSize != null) {
-    result.write(fontSize.floor());
-  } else {
-    result.write(FlutterViewEmbedder.defaultFontSize);
-  }
-  result.write('px ');
-  result.write(canonicalizeFontFamily(fontFamily));
-
-  return result.toString();
+  return '$cssFontStyle $cssFontWeight ${cssFontSize}px $cssFontFamily';
```
ditman, line 26: 🙏

lib/web_ui/lib/src/engine/view_embedder/style_manager.dart
```diff
@@ -2,11 +2,77 @@
...
+import '../browser_detection.dart';
+import '../dom.dart';
+import '../text_editing/text_editing.dart';
+import 'dom_manager.dart';
+
+/// Manages the CSS styles of the Flutter View.
+class StyleManager {
+  static const String defaultFontStyle = 'normal';
+  static const String defaultFontWeight = 'normal';
+  static const double defaultFontSize = 14;
```
ditman, line 16:
```suggestion
  static const int defaultFontSize = 14;
```

ditman (approved):
(Still leaving my approval, but for fontSize I think we should prefer `int` to `num` (or `double`?)

PR #47493, opened 2023-10-30
Update DisplayList tests to explicitly select the Roboto font

jason-simmons:
Skia is removing the API for constructing a default typeface (see https://issues.skia.org/issues/305780908)

Fixes https://github.com/flutter/flutter/issues/137565

### DIFF
--- a/display_list/BUILD.gn
+++ b/display_list/BUILD.gn
@@ -96,11 +96,12 @@ source_set("display_list") {
   }
 }
 
-if (enable_unittests) {
-  test_fixtures("display_list_fixtures") {
-    fixtures = []
-  }
+test_fixtures("display_list_fixtures") {
+  fixtures =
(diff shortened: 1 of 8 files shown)


jason-simmons:
(Requires https://github.com/flutter/engine/pull/47491)

@kjlubick

display_list/BUILD.gn
```diff
@@ -201,10 +202,6 @@ if (enable_unittests) {
   }
 }
 
-fixtures_location("display_list_benchmarks_fixtures") {
```
flar, line 204: Is this really unused?

flar (approved):
LGTM - other than my question about the previous benchmark fixtures...

display_list/BUILD.gn
jason-simmons: I did not see any references to fixtures in the DL benchmarks.  The benchmarks appear to be running as expected with this change.

kjlubick:
LGTM once the compilation issue is sorted out

PR #47494, opened 2023-10-30
Made clang tidy use arm64 if on an arm64 mac.

gaaclarke:
fixes https://github.com/flutter/flutter/issues/137260

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/ci/clang_tidy.sh
+++ b/ci/clang_tidy.sh
@@ -51,10 +51,7 @@ if command -v arch &> /dev/null && [[ $(arch) == "arm64" ]]; then
   CLANG_TIDY_PATH="buildtools/mac-arm64/clang/bin/clang-tidy"
 fi
 
-COMPILE_COMMANDS="$SRC_DIR/out/$BUILD_DIR/compile_commands.json"
-if [ ! -f "$COMPILE_COMMANDS" ]; then
-  (cd "$SRC_DIR"; ./flutter/tools/gn)
-fi
+(cd "$SRC_DIR"; ./flutter/tools/gn)
 
 echo "$(date +%T) Running clang_tidy"
 
--- a/ci/clang_tidy.sh
+++ b/ci/clang_tidy.sh
@@ -46,7 +46,14 @@ else
   fix_flag="--fix --lint-all"
 fi
 
-COMPILE_COMMANDS="$SRC_DIR/out/host_debug/compile_commands.json"
+# Determine wether to use x64 or arm64.
+if command -v arch &> /dev/null && [[ $(arch) == "arm64" ]]; then
+  CLANG_TIDY_PATH="buildtools/mac-arm64/clang/bin/clang-tidy"
+else
+  CLANG_TIDY_PATH="buildtools/mac-x64/clang/bin/clang-tidy"
+fi
+
+COMPILE_COMMANDS="$SRC_DIR/out/$BUILD_DIR/compile_commands.json"
 if [ ! -f "$COMPILE_COMMANDS" ]; then
   (cd "$SRC_DIR"; ./flutter/tools/gn)
 fi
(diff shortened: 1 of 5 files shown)


matanlurey (approved):
Seems reasonable to me, though it's unfortunate this is yet another divergence between the tool and the CI script to run the tool (sure, you could pass the path manually, but exactly nobody will remember to do that).

If we could have the _tool_ find the right binary, instead of relying on the CI script, I would prefer that, but this seems stricly better than the status quo.

gaaclarke:
Yea, fair point.  Plus the less code in bash the better.  I think people just run the bash script for the most point locally.  I don't think there is a reason to use the tool directly.

I'm interested to see what this does to our cpu usage numbers.  Maybe we can bring everything back to presubmit.

zanderso (commented):
I think an easier solution might be to put `--force-mac-arm64` on the right `gn` commands in the build config `.json` files.

tools/clang_tidy/test/clang_tidy_test.dart
```diff
@@ -214,6 +214,23 @@ Future<int> main(List<String> args) async {
     });
   });
 
+  test('clang-tidy specified', () async {
+    _withTempFile('shard-id-valid', (String path) {
+      final Options options = Options.fromCommandLine( <String>[
+          '--clang-tidy="foo/bar"',
+        ],);
+      expect(options.clangTidyPath, isNotNull);
+      expect(options.clangTidyPath, equals('foo/bar'));
```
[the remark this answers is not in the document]
gaaclarke: thanks!

gaaclarke:
> I think an easier solution might be to put `--force-mac-arm64` on the right `gn` commands in the build config `.json` files.
> 
> ETA: It's already there: https://github.com/flutter/engine/blob/main/ci/builders/mac_clang_tidy.json#L17, so at least on CI, this CL will be a no-op.

Ahh yea, a no-op on ci.  I commented in that other issue, we should get rid of `--force-mac-arm64`.  This PR will only make the execution faster locally.  Being able to override the clang-tidy executable is a sound design outside of this problem since deriving the `clang-tidy` location from `build_commands.json` is built on assumptions of the build environment.

So, considering this speeds up local execution, won't break when the ill-conceived `force-mac-arm64` is deleted and has value beyond the problem it addresses, I recommend still landing it.

ci/clang_tidy.sh
```diff
@@ -46,7 +46,12 @@ else
   fix_flag="--fix --lint-all"
 fi
 
-COMPILE_COMMANDS="$SRC_DIR/out/host_debug/compile_commands.json"
+# Determine wether to use x64 or arm64.
+if command -v arch &> /dev/null && [[ $(arch) == "arm64" ]]; then
+  CLANG_TIDY_PATH="buildtools/mac-arm64/clang/bin/clang-tidy"
+fi
+
+COMPILE_COMMANDS="$SRC_DIR/out/$BUILD_DIR/compile_commands.json"
```
zanderso, line 54: How is `$BUILD_DIR` defined?

gaaclarke: It doesn't.  This effectively makes gn get called every time which isn't an error.  In fact it might be desired.  It's quick to execute (1704ms) and means you'll get coverage from newly added files.  I'm going to delete the check.

zanderso: `gn` run time can vary a lot. Some runs on my M1 take >10s. Here's a run on CI taking >4s: https://ci.chromium.org/ui/p/flutter/builders/prod/Mac%20Production%20Engine%20Drone/194351/overview.

Sorry to nitpick, but would it be okay to just fix the arm vs. intel issue in this PR?

gaaclarke: Yea, no problem.  I brought back the check.  What is it doing in all that 10s?  In CI we should have just run `gn`... Oh I guess it won't match though because of the `--force-mac-arm64`.  How I loathe that flag.

gaaclarke: I actually for a moment considered making that check smarter.  It should be doing a timestamp check against the repository.  That will take a fair bit of work though so I left it for a future exercise.  The indiscriminate `gn` doesn't have that bug.

zanderso: The `clang_tidy` script depends on `engine_repo_tools`, which already has a way of finding the most recently generated `out/` directory: https://github.com/flutter/engine/blob/main/tools/pkg/engine_repo_tools/lib/engine_repo_tools.dart#L159, so that could be a good way to avoid adding similar logic in the shell script.

zanderso (approved):

PR #47495, opened 2023-10-30
[Impeller] OpenGLES MSAA Render Buffers (i.e. for stencils)

matanlurey:
Closes https://github.com/flutter/flutter/issues/137569.
Closes https://github.com/flutter/flutter/issues/137095.

As of the latest commit, it works on both the Moto G4 and Pixel 5.

### DIFF
--- a/impeller/renderer/backend/gles/texture_gles.cc
+++ b/impeller/renderer/backend/gles/texture_gles.cc
@@ -23,10 +23,9 @@ static TextureGLES::Type GetTextureTypeFromDescriptor(
   const auto render_target =
       static_cast<TextureUsageMask>(TextureUsage::kRenderTarget);
   const auto is_msaa = desc.sample_count == SampleCount::kCount4;
-  if (usage == render_target && !is_msaa) {
-    // TODO(matanlurey): MSAA render buffers?
-    // See https://github.com/flutter/flutter/issues/137095.
-    return TextureGLES::Type::kRenderBuffer;
+  if (usage == render_target) {
+    return is_msaa ? TextureGLES::Type::kRenderBufferMultisampled
+                   : TextureGLES::Type::kRenderBuffer;
   }
   return is_msaa ? TextureGLES::Type::kTextureMultisampled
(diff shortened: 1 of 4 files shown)


impeller/renderer/backend/gles/texture_gles.cc
```diff
@@ -23,10 +23,9 @@ static TextureGLES::Type GetTextureTypeFromDescriptor(
   const auto render_target =
       static_cast<TextureUsageMask>(TextureUsage::kRenderTarget);
   const auto is_msaa = desc.sample_count == SampleCount::kCount4;
-  if (usage == render_target && !is_msaa) {
-    // TODO(matanlurey): MSAA render buffers?
-    // See https://github.com/flutter/flutter/issues/137095.
-    return TextureGLES::Type::kRenderBuffer;
+  if (usage == render_target) {
+    return is_msaa ? TextureGLES::Type::kRenderBufferMultisampled
+                   : TextureGLES::Type::kRenderBuffer;
```
jonahwilliams, line 28: This should something like usage == render_target && desc.pixelformat == PixelFormat::S8

jonahwilliams: or maybe pixelformat != PixelFormat::whatever_the_gl_pixel_format_is

jonahwilliams: or maybe even better just pass in the expected attachment point and use that.

matanlurey:
Making sure to only use render buffers for stencils solves the artifact issue :)

impeller/renderer/backend/gles/texture_gles.cc
matanlurey: Unfortuantely the TextureGLES is created far before the concept of attachment points :-/

jonahwilliams: RIP. The formats for GLES are more or less constant, though if that changes this will be a fun landmine 😆

jonahwilliams (approved):
LGTM

PR #47496, opened 2023-10-30
[Impeller] Do not return a null Contents if a Paint's blur filter yields an empty coverage rect

jason-simmons:
Fixes https://github.com/flutter/flutter/issues/137484


bdero (approved):
LGTM, thanks!

PR #47529, opened 2023-10-31
[Impeller] Place Rect statics under the Rect template.

bdero:
`Rect::Intersection(a, b)` instead of `Intersection(a, b)`, etc.

Make them usable with other Rect variants, although `RoundOut` is not useful for integer rectangles, for example.


flar (approved):
LGTM

PR #47533, opened 2023-11-01
[Impeller] Include cstdint everywhere that uint32_t is used.

bdero:
Sprinkle cstdint everywhere that it's used. Unblocks the next impeller-cmake bump.

Ran into include issues during the build for the most recent impeller-cmake bump: https://github.com/bdero/impeller-cmake/pull/20

Going to add an optional libcxx STL build to help avoid stuff like this later (in the AAOS action items).

### DIFF
--- a/impeller/base/allocation.h
+++ b/impeller/base/allocation.h
@@ -4,6 +4,7 @@
 
 #pragma once
 
+#include <cstdint>
 #include <limits>
 #include <memory>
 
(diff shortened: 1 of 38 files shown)


jonahwilliams (approved):
LGTM

I checked with my local clangd configuration and it *doesn't* show an unused header warning, so this seems fine. though I suspect without a presubmit check the likelyhood that this will regress in < 1 week is close to 100%

PR #47566, opened 2023-11-01
[ios17][text_input]fix ios 17.0 keyboard freeze when switching languages (without relying on text affinity)

hellohuanlin:
After close examination of the UIKit's default string tokenizer, when querying the line enclosing the end of doc position **in forward direction**, we should return nil (regardless whether the position is forward or backward affinity). 

This aligns with the [API doc](https://developer.apple.com/documentation/uikit/uitextinputtokenizer/1614464-rangeenclosingposition?language=objc): 

> If the text position is at a text-unit boundary, it is considered enclosed only if the next position in the given direction is entirely enclosed.

Will cherry pick this soon. Otherwise it will be less and less important as users upgrade to iOS 17.1. 

### Why my previous workaround also works? 

It turns out my previous workaround PR https://github.com/flutter/engine/pull/46591 works only because our misuse of text affinity in our text input. Specifically, when adding text affinity support, we only added it to `FlutterTextPosition`, but not `FlutterTextRange`. So when getting the beginning/end position from the range, we assign arbitrary affinities. 

https://github.com/flutter/engine/pull/46591

## Pre-launch Checklist

- [x] I signed the [CLA].

### DIFF
--- a/shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
+++ b/shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
@@ -622,7 +622,9 @@ static BOOL IsSelectionRectBoundaryCloserToPoint(CGPoint point,
                           inDirection:(UITextDirection)direction {
   // TODO(hellohuanlin): remove iOS 17 check. The same logic should apply to older iOS version.
   if (@available(iOS 17.0, *)) {
-    // If querying the line from end of document in forward direction, report not found.
+    // According to the API doc if the text position is at a text-unit boundary, it is considered
+    // enclosed only if the next position in the given direction is entirely enclosed. Link:
+    // https://developer.apple.com/documentation/uikit/uitextinputtokenizer/1614464-rangeenclosingposition?language=objc
     FlutterTextPosition* flutterPosition = (FlutterTextPosition*)position;
     if (flutterPosition.index > _textInputView.text.length ||
         (flutterPosition.index == _textInputView.text.length &&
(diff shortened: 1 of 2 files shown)


hellohuanlin:
I need to do more testing on iOS 17.0 and 17.1 (and the newly released 17.1.1), around voice control, scribble and floating cursor selection. But so far this PR is looking good and should be ready for review

stuartmorgan:
> when querying the line enclosing the end of doc position **in forward direction**, we should return nil (regardless whether the position is forward or backward affinity). This is undocumented behavior of UIKit

Is it undocumented? The docs for the method we are implementing say:

>> If the text position is at a text-unit boundary, it is considered enclosed only if the next position in the given direction is entirely enclosed.

Isn't the case here that we are at a boundary, and the next position in the given direction doesn't exist (in which case the next position can't be entirely enclosed)?

hellohuanlin:
> Isn't the case here that we are at a boundary, and the next position in the given direction doesn't exist (in which case the next position can't be entirely enclosed)?

True. It sounds indeed like what we are doing here.

shell/platform/darwin/ios/framework/Source/FlutterTextInputPlugin.mm
```diff
@@ -618,7 +618,19 @@ - (UITextRange*)rangeEnclosingPosition:(UITextPosition*)position
   return result;
 }
 
-- (UITextRange*)lineEnclosingPosition:(UITextPosition*)position {
+- (UITextRange*)lineEnclosingPosition:(UITextPosition*)position
+                          inDirection:(UITextDirection)direction {
+  // TODO(hellohuanlin): remove iOS 17 check. The same logic should apply to older iOS version.
+  if (@available(iOS 17.0, *)) {
+    // If querying the line from end of document in forward direction, report not found.
```
stuartmorgan, line 625: This should link to and quote the docs I mentioned in my comment so that it's clear why.

stuartmorgan (approved):
LGTM with comment nit.

LongCatIsLooong (approved):
LGTM. Thanks for looking into this!

PR #47567, opened 2023-11-01
Add some important comments to some of the FilteContents methods

flar:
Fixes https://github.com/flutter/flutter/issues/137703

Also clarified and fixed indentation on docstring comment for the public method `FilterContents::GetSourceCoverage`

### DIFF
--- a/impeller/entity/contents/filters/filter_contents.h
+++ b/impeller/entity/contents/filters/filter_contents.h
@@ -133,9 +133,10 @@ class FilterContents : public Contents {
   const FilterContents* AsFilter() const override;
 
   /// @brief  Determines the coverage of source pixels that will be needed
-  ///         to produce results for the indicated output limit. This is a
-  ///         reverse bounds calculation computing a source coverage from
-  ///         an intended output coverage.
+  ///         to produce results for the indicated |output_limit| under the
+  ///         indicated |effect_transform|. This is essentially a reverse of
+  ///         the |GetCoverage| method computing a source coverage from
+  ///         an intended |output_limit| coverage.
   ///
   ///         The method computes a result such that if the filter is applied
   ///         to a set of pixels filling the computed source coverage, it
--- a/impeller/entity/contents/filters/filter_contents.h
+++ b/impeller/entity/contents/filters/filter_contents.h
@@ -133,23 +133,35 @@ class FilterContents : public Contents {
   const FilterContents* AsFilter() const override;
 
   /// @brief  Determines the coverage of source pixels that will be needed
-  ///         to produce results for the indicated |output_limit| under the
-  ///         indicated |effect_transform|. This is essentially a reverse of
+  ///         to produce results for the specified |output_limit| under the
+  ///         specified |effect_transform|. This is essentially a reverse of
   ///         the |GetCoverage| method computing a source coverage from
   ///         an intended |output_limit| coverage.
   ///
+  ///         Both the |output_limit| and the return value are in the
+  ///         transformed coordinate space, and so do not need to be
+  ///         transformed or inverse transformed by the |effect_transform|
+  ///         but individual parameters on the filter might be in the
+  ///         untransformed space and should be transformed by the
+  ///         |effect_transform| before applying them to the coverages.
+  ///
   ///         The method computes a result such that if the filter is applied
   ///         to a set of pixels filling the computed source coverage, it
-  ///         should produce an output that covers the entire indicated
+  ///         should produce an output that covers the entire specified
   ///         |output_limit|.
   ///
   ///         This is useful for subpass rendering scenarios where a filter
   ///         will be applied to the output of the subpass and we need to
   ///         determine how large of a render target to allocate in order
   ///         to collect all pixels that might affect the supplied output
-  ///         coverage limit. While we might clip the rendering of the subpass,
-  ///         we want to avoid clipping out any pixels that contribute to
-  ///         the output limit via the filtering operation.
+  ///         coverage limit. While we might end up clipping the rendering
+  ///         of the subpass to its destination, we want to avoid clipping
+  ///         out any pixels that contribute to the output limit via the
+  ///         filtering operation.
+  ///
+  /// @return The coverage bounds in the transformed space of any source pixel
+  ///         that may be needed to produce output for the indicated filter
+  ///         that covers the indicated |output_limit|.
   std::optional<Rect> GetSourceCoverage(const Matrix& effect_transform,
                                         const Rect& output_limit) const;
 
@@ -185,28 +197,18 @@ class FilterContents : public Contents {
   virtual void SetRenderingMode(Entity::RenderingMode rendering_mode);
 
  private:
-  /// @brief  Internal utility method for |GetLocalCoverage|.
-  ///
-  ///         |GetLocalCoverage| will intersect the filter's coverage as
-  ///         computed by this method with the oject's coverage hint.
-  ///         This method should compute the coverage of the associated
-  ///         inputs and then apply any modifications to that coverage
-  ///         representing the actions of the filtering operation, ignoring
-  ///         the coverage hint.
+  /// @brief  Internal utility method for |GetLocalCoverage| that computes
+  ///         the output coverage of this filter across the specified inputs,
+  ///         ignoring the coverage hint.
   virtual std::optional<Rect> GetFilterCoverage(
       const FilterInput::Vector& inputs,
       const Entity& entity,
       const Matrix& effect_transform) const;
 
-  /// @brief  Internal utility method for |GetSourceCoverage|.
-  ///
-  ///         The public method will use this method to compute the inverse
-  ///         coverage for the indicated |output_limit| and then go on to
-  ///         consult the inputs to further expand/refine the required
-  ///         source coverage.
-  ///         This method is only responsible for computing the inverse
-  ///         coverage for the filtering operation based on the indicated
-  ///         |output_coverage|, ignoring the inputs.
+  /// @brief  Internal utility method for |GetSourceCoverage| that computes
+  ///         the inverse effect of this transform on the specified output
+  ///         coverage, ignoring the inputs which will be accommodated by
+  ///         the caller.
   virtual std::optional<Rect> GetFilterSourceCoverage(
       const Matrix& effect_transform,
       const Rect& output_limit) const = 0;
@@ -220,6 +222,11 @@ class FilterContents : public Contents {
       const Rect& coverage,
       const std::optional<Rect>& coverage_hint) const = 0;
 
+  /// @brief  Internal utility method to compute the coverage of this
+  ///         filter across its internally specified inputs and subject
+  ///         to the coverage hint.
+  ///
+  ///         Uses |GetFilterCoverage|.
   std::optional<Rect> GetLocalCoverage(const Entity& local_entity) const;
 
   FilterInput::Vector inputs_;
--- a/impeller/entity/contents/filters/filter_contents.h
+++ b/impeller/entity/contents/filters/filter_contents.h
@@ -132,17 +132,23 @@ class FilterContents : public Contents {
   // |Contents|
   const FilterContents* AsFilter() const override;
 
-  /// @brief Determines the coverage of source pixels that will be needed
-  ///        to apply this filter under the given transform and produce
-  ///        results anywhere within the indicated coverage limit.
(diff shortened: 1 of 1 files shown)


impeller/entity/contents/filters/filter_contents.h
```diff
@@ -132,17 +132,23 @@ class FilterContents : public Contents {
   // |Contents|
   const FilterContents* AsFilter() const override;
 
-  /// @brief Determines the coverage of source pixels that will be needed
-  ///        to apply this filter under the given transform and produce
-  ///        results anywhere within the indicated coverage limit.
+  /// @brief  Determines the coverage of source pixels that will be needed
+  ///         to produce results for the indicated output limit. This is a
+  ///         reverse bounds calculation computing a source coverage from
+  ///         an intended output coverage.
```
gaaclarke, line 138: So, it's the inverse of `GetCoverage`?

flar: Correct. Is that confusing from what I wrote? Or are you suggesting a simplification?

flar: Oops, I lost the mention of the transform. Minor update coming...

flar: How about:

```
  /// @brief  Determines the coverage of source pixels that will be needed
  ///         to produce results for the indicated output coverage under the
  ///         indicated effect transform. This is essentially a reverse of
  ///         the |GetCoverage| method computing a source coverage from
  ///         an intended output coverage.
```

(I'm also renaming the `output_limit` parameter to `output_coverage` to match that text)

flar: Ugh, renaming the parameter and propagating it to all of the downstream methods would touch a bunch of files unnecessarily, so I'll change the comment to match instead.

impeller/entity/contents/filters/filter_contents.h
```diff
@@ -132,17 +132,23 @@ class FilterContents : public Contents {
...
+  ///         should produce an output that covers the entire indicated
+  ///         |output_limit|.
+  ///
+  ///         This is useful for subpass rendering scenarios where a filter
+  ///         will be applied to the output of the subpass and we need to
+  ///         determine how large of a render target to allocate in order
+  ///         to collect all pixels that might affect the supplied output
+  ///         coverage limit. While we might clip the rendering of the subpass,
+  ///         we want to avoid clipping out any pixels that contribute to
+  ///         the output limit via the filtering operation.
```
gaaclarke, line 152: This never defines what `effect_transform` is.  Looking at the implementation of some `GetFilterSourceCoverage` i know that only the basis is used.  That should be mentioned.

For `DirectionalGaussianBlurFilterContents::GetFilterSourceCoverage` shouldn't the result of `GetFilterSourceCoverage` be smaller than `output_limit` since the source will not have the blur radius?  It is actually returning a larger result.

The problem I ran into was this:
```c++
TEST(DirectionalGaussianBlurFilterContents, FilterSourceCoverage) {
  Scalar sigma_radius_1 = CalculateSigmaForBlurRadius(1.0);
  auto contents = std::make_unique<DirectionalGaussianBlurFilterContents>();
  contents->SetSigma(Sigma{sigma_radius_1});
  contents->SetDirection({1.0, 0.0});
  std::optional<Rect> coverage = contents->GetFilterSourceCoverage(
      /*effect_transform=*/Matrix::MakeScale({2.0, 2.0, 1.0}),
      /*output_limit=*/Rect::MakeLTRB(100, 100, 200, 200));
  ASSERT_EQ(coverage, ???);
}
```
Should the output_limit be `Rect::MakeLTRB(100 - 2, 100, 200 + 2, 200)` or `Rect::MakeLTRB(50 - 2, 50, 100 + 2, 100)` or `Rect::MakeLTRB(200 - 2, 200, 400 + 2, 400)`?  It's not clear how the `effect_transform` is related to the operation by reading the documentation.

impeller/entity/contents/filters/filter_contents.h
```diff
@@ -178,11 +184,28 @@ class FilterContents : public Contents {
   virtual void SetRenderingMode(Entity::RenderingMode rendering_mode);
 
  private:
+  /// @brief  Internal utility method for |GetLocalCoverage|.
```
gaaclarke, line 188: The brief should explain what it does, not how it is used.  It should be something like "Computes the filter's coverage sans coverage hint.

See also:
  - |GetLocalCoverage| - This combines this result with the coverage hint.
"

impeller/entity/contents/filters/filter_contents.h
```diff
@@ -178,11 +184,28 @@ class FilterContents : public Contents {
...
+  ///
+  ///         |GetLocalCoverage| will intersect the filter's coverage as
+  ///         computed by this method with the oject's coverage hint.
```
gaaclarke, line 191:
```suggestion
  ///         computed by this method with the object's coverage hint.
```

impeller/entity/contents/filters/filter_contents.h
```diff
@@ -178,11 +184,28 @@ class FilterContents : public Contents {
...
+  ///         This method should compute the coverage of the associated
+  ///         inputs and then apply any modifications to that coverage
+  ///         representing the actions of the filtering operation, ignoring
+  ///         the coverage hint.
   virtual std::optional<Rect> GetFilterCoverage(
       const FilterInput::Vector& inputs,
       const Entity& entity,
       const Matrix& effect_transform) const;
 
+  /// @brief  Internal utility method for |GetSourceCoverage|.
```
gaaclarke, line 201: This one is mostly good, but make the brief define what the function does briefly.  Also define effect_transform.  You'll probably just want to link to a definition somewhere else in this file.

What is the coordinate space of the `output_limit`?

impeller/entity/contents/filters/filter_contents.h
```diff
@@ -132,17 +132,24 @@ class FilterContents : public Contents {
   // |Contents|
   const FilterContents* AsFilter() const override;
 
-  /// @brief Determines the coverage of source pixels that will be needed
-  ///        to apply this filter under the given transform and produce
-  ///        results anywhere within the indicated coverage limit.
+  /// @brief  Determines the coverage of source pixels that will be needed
+  ///         to produce results for the indicated |output_limit| under the
```
gaaclarke, line 136: What coordinate system is `output_limit` in?

gaaclarke (commented):
As per our discussion in the chat.  If you want to format these differently.  I'm fine with that.  Here is an example of implicit brief/details that doesn't use extra whitespace.

```c++
/// This is brief.
/// This is a details, lots and lots of details.
/// Here's even more details.
void Foobar();
```

impeller/entity/contents/filters/filter_contents.h
gaaclarke: It was just a genuine question.  That doesn't match my understanding after having looked at `DirectionalGaussianBlurFilterContents`.

flar:
> As per our discussion in the chat. If you want to format these differently. I'm fine with that. Here is an example of implicit brief/details that doesn't use extra whitespace.
> 
> ```c++
> /// This is brief.
> /// This is a details, lots and lots of details.
> /// Here's even more details.
> void Foobar();
> ```

Consistency within files and modules is always a win, though. I was more questioning if we are sure we want to finalize on this style before we start going back and really filling out a lot of comments as we are likely to do in the near future. The chat showed widespread support for doing the style being used here.

impeller/entity/contents/filters/filter_contents.h
flar: > This never defines what `effect_transform` is. Looking at the implementation of some `GetFilterSourceCoverage` i know that only the basis is used. That should be mentioned.

Use of Basis is specific to a filter computing its best guess, not to the general operation. All this method specifies is that the filters should do a reverse computation, not how they should do it.

For the record, I don't think the use of basis on those filters is correct and I should file an issue about it ([done](https://github.com/flutter/flutter/issues/137723)). I used basis because the existing methods in the class already used basis and I wanted to be consistent with the current state of their implementations.

> For `DirectionalGaussianBlurFilterContents::GetFilterSourceCoverage` shouldn't the result of `GetFilterSourceCoverage` be smaller than `output_limit` since the source will not have the blur radius? It is actually returning a larger result.

No, it has to be larger. Input pixels are taken from a radius and so we need more pixels from the source than the size of the dest.

> The problem I ran into was this:
> 
> ```c++
> TEST(DirectionalGaussianBlurFilterContents, FilterSourceCoverage) {
>   Scalar sigma_radius_1 = CalculateSigmaForBlurRadius(1.0);
>   auto contents = std::make_unique<DirectionalGaussianBlurFilterContents>();
>   contents->SetSigma(Sigma{sigma_radius_1});
>   contents->SetDirection({1.0, 0.0});
>   std::optional<Rect> coverage = contents->GetFilterSourceCoverage(
>       /*effect_transform=*/Matrix::MakeScale({2.0, 2.0, 1.0}),
>       /*output_limit=*/Rect::MakeLTRB(100, 100, 200, 200));
>   ASSERT_EQ(coverage, ???);
> }
> ```
> 
> Should the output_limit be `Rect::MakeLTRB(100 - 2, 100, 200 + 2, 200)` or `Rect::MakeLTRB(50 - 2, 50, 100 + 2, 100)` or `Rect::MakeLTRB(200 - 2, 200, 400 + 2, 400)`? It's not clear how the `effect_transform` is related to the operation by reading the documentation.

I mentioned that it was relative to the transform in the most recent cut, but I'll be more specific. The supplied "output_limit" and the result should both be in the transformed space. The sigma is in the untransformed space, but the input and output coverages are both in the transformed space. So, you need to adjust the sigma to its correct value in the transformed space and then all 3 are in the same space.

So, here you would need to scale the radius by 2, but then apply it directly to the 100,100 values and return them relative to 100,100 - so expand by 2 in both left and right. Your first answer looks right to me, other than assuming that sigma is the value you should expand to. There is a calculation in the code that maps sigma to the "fringe" of the gaussian blur. That "fringe" calculation will be in the untransformed space, so compute sigma -> fringe, and then map fringe through the transform.

But since both the input and the output rectangles are in the transformed space of the `effect_transform`, neither of those gets run through, or inversed through, the transform.

impeller/entity/contents/filters/filter_contents.h
```diff
@@ -132,17 +132,23 @@ class FilterContents : public Contents {
   // |Contents|
   const FilterContents* AsFilter() const override;
 
-  /// @brief Determines the coverage of source pixels that will be needed
-  ///        to apply this filter under the given transform and produce
-  ///        results anywhere within the indicated coverage limit.
+  /// @brief  Determines the coverage of source pixels that will be needed
+  ///         to produce results for the indicated output limit. This is a
```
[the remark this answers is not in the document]
flar: I used the |output_limit| notation in my latest push, which is what I've used elsewhere for parameters. Is that correct?

impeller/entity/contents/filters/filter_contents.h
flar: `|GetSourceCoverage|` is a link to the full definition...

impeller/entity/contents/filters/filter_contents.h
flar: Perhaps you are confused by the non-intuitive fact that blur filters expand in both directions. Just as a given edge pixel can be the input for a fringe around the source and so it expands when you apply the filter, each pixel takes input from a range of pixels around it so you need more source pixels than output pixels in order to compute the values correctly.

So, the gaussian filter **is** computing a reverse bounds by _expanding_ it...

impeller/entity/contents/filters/filter_contents.h
flar: Rewritten...

impeller/entity/contents/filters/filter_contents.h
flar: I tried to detail the coordinate spaces in the latest version of the comment which has lots of changes, so I'm going to call this one resolved, though my new description may still have some confusion in it...

impeller/entity/contents/filters/filter_contents.h
flar: I basically condensed these comments (including their non-brief parts) into just a single brief sentence.

gaaclarke (approved):
This is much more helpful.  I still suspect they can be improved but I'm definitely not well informed enough to suggest anything.  Thanks!

PR #47573, opened 2023-11-01
Bundle flutter.js via esbuild

eyebrowsoffire:
* Roll esbuild as a CIPD package along with the browser roller. I renamed `browser_lock`/`browser_roller` etc to `package_lock` and `package_roller` since it will handle more than just browsers now.
* Download the esbuild CIPD package via DEPS.
* Have a build rule for running esbuild on some JavaScript or TypeScript
* Bundle and minify `flutter.js` using esbuild.
* Include in `flutter_web_sdk` the original `flutter.js` source, the minified `flutter.js`, and a sourcemap file to map between thnm.
* Also slightly changed the structure to put the `flutter.js` stuff in `flutter_web_sdk/flutter_js` instead of just at the root level of `flutter_web_sdk`. This should be fine because I haven't merged the change that has the flutter tool consume this yet.


eyebrowsoffire:
Btw, testing this in presubmit here: https://github.com/flutter/flutter/pull/137113

web_sdk/BUILD.gn
```diff
@@ -528,7 +520,16 @@ if (!is_fuchsia) {
...
-    sources += [ "$root_out_dir/flutter_web_sdk/flutter.js" ]
+    # flutter.js full sources
+    foreach(source, flutter_js_source_list) {
+      sources += [ "$root_out_dir/flutter_web_sdk/flutter_js/$source" ]
+    }
+
+    # flutter.js bundled source and sourcemap
+    sources += [
+      "$root_out_dir/flutter_web_sdk/flutter_js/flutter.js",
+      "$root_out_dir/flutter_web_sdk/flutter_js/flutter.js.map",
```
yjbanov, line 531: We should publish both optimized and raw `flutter.js` so that it is possible to troubleshoot issues in debug and profile builds. In fact, until the size of `flutter.js` is more than a blip, I wouldn't bother minifying, just strip the TypeScript type annotations. Alternatively, if we could roll `esbuild` all the way to the Flutter SDK (how big is the binary?), and then provide optimization/minification options in the Flutter Tool. This way they engine build of `flutter.js` can be distributed in original source.

eyebrowsoffire: We actually are publishing both with this PR now! We ship the minified version, the original source, and the sourcemaps that map between them.

kevmoo: Good question! about how big it is without minification?

eyebrowsoffire: Right now uncompressed it's 14kb before minification, 4kb after, so not terribly significant. The main reason to be bundling is to merge multiple files into one so that we don't have a complex waterfall with multiple roundtrips. Obviously right now there is just a single file, but I'm trying to prepare to modularize it into separate .ts files. Since we have sourcemaps and are shipping the original source, I don't really see a downside to actually minifying it since we're already bundling though.

yjbanov (approved):
LGTM if LGT @ditman

lib/web_ui/flutter_js/BUILD.gn
```diff
@@ -2,7 +2,25 @@
...
+
+copy("flutter_js_sources") {
+  sources = flutter_js_source_list
+  outputs =
+      [ "$root_out_dir/flutter_web_sdk/flutter_js/{{source_target_relative}}" ]
+}
+
+esbuild("flutter_js_bundle") {
+  entry_point = "$root_out_dir/flutter_web_sdk/flutter_js/src/flutter.js"
+  output_bundle = "$root_out_dir/flutter_web_sdk/flutter_js"
```
ditman, line 23: do we want this to be `$root_out_dir/flutter_web_sdk/flutter_js/dist` or similar? (this might be problematic for the sourcemaps maybe?)

build/esbuild/esbuild.gni
```diff
@@ -0,0 +1,38 @@
...
+    outputs = [
+      invoker.output_bundle,
+      invoker.output_bundle + ".map",
+    ]
+    absolute_output = rebase_path(invoker.output_bundle)
+    args = [
+      rebase_path(esbuild, root_build_dir),
+      "--bundle",
+      "--minify",
+      "--sourcemap",
```
ditman, line 33: Can these 3 args be passed in from the `invoker`? That way the `esbuild` action is more generic, and the operations that are performed on the source are defined close to where the actual flutter.js build target is defined?

lib/web_ui/flutter_js/BUILD.gn
```diff
@@ -2,7 +2,25 @@
...
+
```
ditman, line 24: (add the esbuild args here?)

ditman (approved):
Small comments, but let's go!

lib/web_ui/flutter_js/BUILD.gn
eyebrowsoffire: Yeah that will end up with the sourcemaps mapping to `../src/flutter.js`, which I don't think is going to work well when served via http. I'm not exactly sure how to around that issue with a `dist` folder to be honest.

PR #47576, opened 2023-11-01
[Impeller] Implements gaussian blur that scales down before applying the blur

gaaclarke:
A partial implementation of the new gaussian blur effect.  This should perform enough of the code to start getting some performance numbers.

Known outstanding problems:
  1) The edges of the blur are clipped.   I have notes on how I plan on expanding render space in the PR.
  1) Animating the sigma causes some "jumping around artifacts" resulting from the downsampling (maybe the discrete nature of texture pixel size?)
  1) Coverage hints are ignored.  I think depth tests will make that not much of an issue.
  1) We aren't ping ponging textures yet
  1) The snapshot's transform is ignored.

issue: https://github.com/flutter/flutter/issues/131580

## Pre-launch Checklist

- [x] I signed the [CLA].


impeller/entity/shaders/gaussian_blur/gaussian_blur.glsl
```diff
@@ -54,7 +55,7 @@ void main() {
   // just uses 0.5 as an optimization until the gaussian coefficients are
   // calculated and passed in from the cpu.
   for (float16_t i = -blur_info.blur_radius; i <= blur_info.blur_radius;
-       i += 2.0hf) {
+       i += blur_info.step_size) {
```
gaaclarke, line 58: This causes waffling that even more noticeable with the downscale approach.

impeller/entity/contents/filters/gaussian_blur_filter_contents_unittests.cc
```diff
@@ -0,0 +1,242 @@
...
+  auto filter_input = FilterInput::Make(texture);
+  Entity entity;
+  Quad uvs = GaussianBlurFilterContents::CalculateUVs(filter_input, entity,
+                                                      ISize(100, 100));
+  std::optional<Rect> uvs_bounds = Rect::MakePointBounds(uvs);
+  EXPECT_TRUE(uvs_bounds.has_value());
+  if (uvs_bounds.has_value()) {
+    EXPECT_TRUE(RectNear(uvs_bounds.value(), Rect::MakeXYWH(0, 0, 1, 1)));
+  }
+}
```
gaaclarke, line 239: If we used mock objects as we had originally I could test the uv values directly when rendering.  This is the best I can do with real objects.  cc @matanlurey @jonahwilliams

jonahwilliams: In theory you could instrument the creation of commands and render passes, and then read-back the real UVs. I might implement something like that in the future.

In the meantime I think its fair to say that based on the gaussian structure it would be too difficult to pull out the real UVs (by hooking into the cmd buffer) and that mocks may be appropriate. Factoring out the logic into a testable method is also a completely reasonable means to test IMO.

But @gaaclarke you're the one putting in the work here. If you don't feel like the static methods are sufficient then mock away.

gaaclarke: Yea, the issue I ran into is that I kept fixing one test and breaking another and golden image tests cannot be run locally so I have to do manual testing to make sure those 2 tests kept working.  Ideally I'd want to scale beyond checking 2 tests too.  I'll see if I get around to testing uvs directly via mocks.

impeller/entity/shaders/gaussian_blur/gaussian_blur.glsl
jonahwilliams: Interesting! I will dig into this afterwards. Its a nice win if we can get it working but is otherwise completely orthogonal to this change

gaaclarke: It's the issue I reported here: https://github.com/flutter/flutter/issues/138259

gaaclarke:
@jonahwilliams @bdero are you guys interested in doing a review for this now, not being 100% done?  I can keep working on it in this PR.  I would like to get the coverage expanding but it will make the review easier to do if that gets split out.  I think this is a good flag in the sand.

jonahwilliams:
Yes, I will happily take a look at this now (well, tomorrow 😄 )

jonahwilliams:
So I just tried this out. Wow 🚀 , its crazy fast compared to our current approach. I do see artifacts along the top and bottom edge. I wonder if these are coverage artifacts? I see that you've called that out as a current problem.

impeller/entity/contents/filters/filter_contents.cc
```diff
@@ -55,23 +56,38 @@ std::shared_ptr<FilterContents> FilterContents::MakeGaussianBlur(
...
-      /*secondary_sigma=*/sigma_x);
-  return y_blur;
+  constexpr bool use_new_filter =
+#ifdef IMPELLER_ENABLE_NEW_GAUSSIAN_FILTER
+      true;
+#else
+      false;
+#endif
+
+  // TODO(https://github.com/flutter/flutter/issues/131580): Remove once the new
```
jonahwilliams, line 66: Can you document (here or in the issue) which cases the blur doesn't handle now?

impeller/entity/contents/filters/filter_contents.cc
```diff
@@ -55,23 +56,38 @@ std::shared_ptr<FilterContents> FilterContents::MakeGaussianBlur(
...
+  // blur handles all cases.
+  if (use_new_filter) {
+    auto blur = std::make_shared<GaussianBlurFilterContents>(sigma_x.sigma);
+    blur->SetInputs({input});
+    return blur;
+  } else {
+    std::shared_ptr<FilterContents> x_blur = MakeDirectionalGaussianBlur(
```
jonahwilliams, line 73: nit: no else if there is an early return.

```suggestion
  if (use_new_filter) {
    auto blur = std::make_shared<GaussianBlurFilterContents>(sigma_x.sigma);
    blur->SetInputs({input});
    return blur;
  }
  
  std::shared_ptr<FilterContents> x_blur = MakeDirectionalGaussianBlur(
```

impeller/entity/contents/filters/gaussian_blur_filter_contents.cc
```diff
@@ -0,0 +1,273 @@
...
+using GaussianBlurFragmentShader = GaussianBlurPipeline::FragmentShader;
+
+namespace {
+SamplerDescriptor MakeSamplerDescriptor(MinMagFilter filter,
+                                        SamplerAddressMode address_mode) {
+  SamplerDescriptor sampler_desc;
+  sampler_desc.min_filter = filter;
+  sampler_desc.mag_filter = filter;
+  sampler_desc.width_address_mode = address_mode;
+  sampler_desc.width_address_mode = address_mode;
```
jonahwilliams, line 26: you set the width address mode twice here, does this need to be height address mode too? I wonder if this is causing the artifacts.

impeller/entity/contents/filters/gaussian_blur_filter_contents.cc
```diff
@@ -0,0 +1,273 @@
...
+      "Gaussian Blur Filter", subpass_size, subpass_callback);
+  return out_texture;
+}
+
+std::shared_ptr<Texture> MakeBlurSubpass(
+    const ContentContext& renderer,
+    std::shared_ptr<Texture> input_texture,
+    const SamplerDescriptor& sampler_descriptor,
+    const GaussianBlurFragmentShader::BlurInfo& blur_info) {
+  // TODO(gaaclarke): This doesn't render if the blur_info.sigma == 0.
```
jonahwilliams, line 90: Where do we check for sigma < kEhClose enough on the old blur? Presumibly we could detect that and just make the filter return the original contents, or "optimize" it out entirely

impeller/entity/contents/filters/gaussian_blur_filter_contents.cc
```diff
@@ -0,0 +1,273 @@
...
+/// |size| in integer sizes.
+Vector2 CalculateIntegerScale(Scalar scalar, ISize size) {
+  ISize new_size(size.width / scalar, size.height / scalar);
+  return Vector2(size.width / static_cast<Scalar>(new_size.width),
+                 size.height / static_cast<Scalar>(new_size.height));
+}
+
+/// Calculate how much to scale down the texture depending on the blur radius.
+/// This curve was taken from |DirectionalGaussianBlurFilterContents|.
+Scalar CalculateScale(Scalar radius) {
```
jonahwilliams, line 139: If I wanted to change this scaling, is this where I would do it?

As a follow up we may align to skia's curve, which scales sigma down to 4 proportionally.

jonahwilliams (commented):
Just some quick notes

impeller/entity/contents/filters/filter_contents.cc
gaaclarke: I have them listed in the issue above.  If we land this early I can copy them over to the issue.

impeller/entity/contents/filters/gaussian_blur_filter_contents.cc
jonahwilliams: I checked and this did not fix the artifacting issue. I suspect this is probably to do with coverage then

gaaclarke:
> So I just tried this out. Wow 🚀 , its crazy fast compared to our current approach. I do see artifacts along the top and bottom edge. I wonder if these are coverage artifacts? I see that you've called that out as a current problem.

We might want to tweak the scaling algorithm, its a bit jittery when it's animated.  So we have that lever to change to get animation quality versus performance.  I think the static results all look fine with this curve those.

Which artifacts?  You talking about the cropping of the "halo" around blurred items?  Thats the one I know about.

impeller/entity/contents/filters/gaussian_blur_filter_contents.cc
gaaclarke: Yep, oh interesting.  This was me copying the one from the other blur.  I think we'll have to revisit this since I think it's a bit too aggressive (because of the artifacts when animating the sigma).

impeller/entity/contents/filters/gaussian_blur_filter_contents.cc
gaaclarke: Good catch, that definitely would have shown up at some point.  The default value is nearest.  Maybe that was contributing to the worse waffling we were seeing with the 2 step?

impeller/entity/contents/filters/gaussian_blur_filter_contents.cc
gaaclarke: Good call.  I do have a todo to address this.  It didn't occur to me yet that the solution was so simple, done.

impeller/entity/contents/filters/filter_contents.cc
gaaclarke: done

jonahwilliams:
I see a dark line/flicker across the top and bottom of one of the blurs on wonderous, specifically the one that slides up on the timeline page.

gaaclarke:
> I see a dark line/flicker across the top and bottom of one of the blurs on wonderous, specifically the one that slides up on the timeline page.

Oh interesting.  I haven't seen that.  It probably isn't worth looking into until I've fixed the halo since that's going to change things around the edges.

gaaclarke:
Oh I'm seeing the glitch now, we should fix that.  Looking into it.

blurglitch

gaaclarke:
> I see a dark line/flicker across the top and bottom of one of the blurs on wonderous, specifically the one that slides up on the timeline page.

Fixed.

impeller/entity/contents/filters/gaussian_blur_filter_contents.h
```diff
@@ -0,0 +1,58 @@
...
+
+namespace impeller {
+
+/// Performs a bidirectional Gaussian blur.
+///
+/// This is accomplished by rendering multiple passes in multiple directions.
+/// Note: This will replace `DirectionalGaussianBlurFilterContents`.
+class GaussianBlurFilterContents final : public FilterContents {
+ public:
+  GaussianBlurFilterContents(Scalar sigma = 0.0f);
```
jonahwilliams, line 18: nit: explicit?

jonahwilliams (approved):
LGTM (with nits)!

impeller/entity/contents/filters/gaussian_blur_filter_contents.h
gaaclarke: done

impeller/entity/contents/filters/gaussian_blur_filter_contents.cc
```diff
@@ -0,0 +1,283 @@
...
+  return input_coverage.value().Expand(Point(blur_radii.x, blur_radii.y));
+}
+
+std::optional<Entity> GaussianBlurFilterContents::RenderFilter(
+    const FilterInput::Vector& inputs,
+    const ContentContext& renderer,
+    const Entity& entity,
+    const Matrix& effect_transform,
+    const Rect& coverage,
+    const std::optional<Rect>& coverage_hint) const {
```
bdero, line 185: I don't see the snapshot's transform being consumed anywhere, which will cause issues with some optimized snapshots.

The "destination rect" of `TextureContents` scales the texture when taking a snapshot to avoid an unnecessary render pass. I'd recommend absorbing the scale of the basis vectors in the first downsampling pass, and applying other aspects of the transform to the final returned entity.

So the texture of a snapshot may be in pass space, entity space, or something in-between.

image

bdero (approved):
Awesome progress! Left some comments.

I suspect there is a problem with the way we're scaling the sampling interval because I'm seeing extreme loss of high frequency info for low sigma values. But I haven't stared at it long enough to spot it yet and we can always fix it up later.

impeller/entity/contents/filters/gaussian_blur_filter_contents.cc
```diff
@@ -0,0 +1,283 @@
...
+  std::shared_ptr<Texture> pass2_out_texture = MakeBlurSubpass(
+      renderer, pass1_out_texture, input_snapshot->sampler_descriptor,
+      GaussianBlurFragmentShader::BlurInfo{
+          .blur_uv_offset = Point(0.0, pass1_pixel_size.height),
+          .blur_sigma = sigma_ / downsample.y,
+          .blur_radius = blur_radius / downsample.y,
+          .step_size = 1.0,
+      });
+
+  // TODO(gaaclarke): Make this pass reuse the texture from pass1.
```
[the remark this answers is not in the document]
gaaclarke: We should be able to do: 
`Downsample(src: input, dst: texture1) -> Horizontal Blur(src: texture1, dst: texture2) -> Vertical Blur(src: texture2, dst: texture1) -> texture1` 

Right now we are doing: 
`Downsample(src: input, dst: texture1) -> Horizontal Blur(src: texture, dst: texture2) -> Vertical Blur(src: texture2, dst: texture3) -> texture3`.  

I haven't looked into what plumbing we'd need to make this work yet, the ol' ping-pong maneuver.  Potentially it will require different synchronization but I suspect the way we are doing passes now it won't be needed.  I haven't looked into it yet.

impeller/entity/contents/filters/filter_contents.cc
```diff
@@ -55,6 +56,20 @@ std::shared_ptr<FilterContents> FilterContents::MakeGaussianBlur(
     Sigma sigma_y,
     BlurStyle blur_style,
     Entity::TileMode tile_mode) {
+  constexpr bool use_new_filter =
+#ifdef IMPELLER_ENABLE_NEW_GAUSSIAN_FILTER
```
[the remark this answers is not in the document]
gaaclarke: Just editing the code during development.  I haven't needed anything else yet.

impeller/entity/contents/filters/gaussian_blur_filter_contents.cc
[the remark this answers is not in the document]
jonahwilliams: FWIW I've thought about building this into the render target cache itself, but I don't have a good way to enforce that we don't create self dependencies. Perhaps if we deferred actual texture creation (from @chinmaygarde 's suggestion) and just used descriptors, it would be easier to apply these sorts of optimizations generally.

impeller/entity/contents/filters/gaussian_blur_filter_contents.cc
gaaclarke: What parameter are you editing to get that?  I did see that but I didn't have a case where it manifested.  I assumed that was getting absorbed elsewhere so the uv coords weren't effected.

impeller/geometry/rect.h
```diff
@@ -48,6 +48,11 @@ struct TRect {
     return TRect(0.0, 0.0, size.width, size.height);
   }
 
+  template <typename U>
+  constexpr static std::optional<TRect> MakePointBounds(const U& value) {
+    return MakePointBounds(value.begin(), value.end());
+  }
```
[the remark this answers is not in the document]
gaaclarke: done

impeller/entity/contents/filters/gaussian_blur_filter_contents.cc
bdero: It's the "Path XYWH" at the bottom

impeller/entity/contents/filters/gaussian_blur_filter_contents.cc
```diff
@@ -0,0 +1,283 @@
...
+  vtx_builder.AddVertices(vertices);
+  auto vtx_buffer = vtx_builder.CreateVertexBuffer(host_buffer);
+  cmd.BindVertices(vtx_buffer);
+}
+
+std::shared_ptr<Texture> MakeDownsampleSubpass(
+    const ContentContext& renderer,
+    std::shared_ptr<Texture> input_texture,
+    const SamplerDescriptor& sampler_descriptor,
+    const std::array<Point, 4>& uvs,
```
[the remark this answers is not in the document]
gaaclarke: done

impeller/entity/contents/filters/gaussian_blur_filter_contents.cc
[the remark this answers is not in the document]
bdero: Ah, yeah that makes sense. The first thing that popped into my mind was that you were planning to also sample from the first texture for something. 😆

gaaclarke:
> I suspect there is a problem with the way we're scaling the sampling interval because I'm seeing extreme loss of high frequency info for low sigma values.

@bdero I think that is related to our aggressive scaling function.  There is a conversation in this issue a couple places about changing it.

PR #47582, opened 2023-11-02
[Impeller] Move all remaining Rect construction to named factories

flar:
The named factories provide better readability and do not imply an internal storage format (which will be changing soon).


bdero (approved):
LGTM

PR #47586, opened 2023-11-02
[Impeller] Add support for external GL textures to TiledTextureContents.

bdero:
Resolves https://github.com/flutter/flutter/issues/135881.

OES_EGL_image_external only allows CLAMP_TO_EDGE address sampling modes, so emulate the other tiling modes.


jonahwilliams:
There are extensions we can check for for the remaining tile modes:

* https://registry.khronos.org/OpenGL/extensions/OES/OES_texture_border_clamp.txt
* https://registry.khronos.org/OpenGL/extensions/OES/OES_texture_mirrored_repeat.txt (Maybe?)

We also currently draw the texture with a repeat mode when we add it to the display list. Maybe we should change that to clamp to edge too

impeller/entity/contents/tiled_texture_contents.cc
```diff
@@ -153,34 +158,64 @@ bool TiledTextureContents::Render(const ContentContext& renderer,
...
+  if (is_external_texture) {
+    SamplerDescriptor sampler_desc;
+    // OES_EGL_image_external states that only CLAMP_TO_EDGE is valid, so we
+    // emulate all other tile modes here by remapping the texture coordinates.
+    sampler_desc.width_address_mode = SamplerAddressMode::kClampToEdge;
+    sampler_desc.height_address_mode = SamplerAddressMode::kClampToEdge;
+
+    // Also, external textures cannot be bound to color filters, so ignore this
+    // case for now.
+    FSExternal::BindSAMPLEREXTERNALOESTextureSampler(
```
jonahwilliams, line 200: I think we cannot actually hit this case with the API we have today (which doesn't allow adding color/image filters directly to external textures) - so perhaps a DCHECK is appropriate.

bdero:
> There are extensions we can check for for the remaining tile modes

Since we'll still sometimes need this emulation if supporting all the tiling modes, I think we should add faster paths in a less urgent follow-up.

jonahwilliams:
This PR isn't urgent, so take your time 😄

impeller/entity/contents/tiled_texture_contents.cc
bdero: > which doesn't allow adding color/image filters directly to external textures

Where is this disallowed? 🤔 There is no separate API for external textures in Aiks.

jonahwilliams: The optimization only applies to color filters that are on the paint object used in the draw call. External texture API does not accept a paint object so all filters need to be applied by a separate composited later

bdero: Oh I see, you mean how we expose it to plugins/framework. Added a dcheck. We can easily add support later for other customers if requested.

bdero:
Alright, this should be good to review, although I should probably try running the failing benchmark against it before landing.

impeller/renderer/testing/mocks.h
```diff
@@ -85,6 +88,19 @@ class MockBlitPass : public BlitPass {
               (override));
 };
 
+class MockRenderPass : public RenderPass {
```
jonahwilliams, line 91: nit: unused.

jonahwilliams (approved):
Overall LGTM, though of course please test on device too.

Also please file issues for the oes tile mode extensions.

bdero:
Alrighty, I ran the benchmark on device and it didn't crash. Hopefully the builder will stop failing once this rolls in (`Linux_android platform_views_scroll_perf_impeller__timeline_summary`).

PR #47592, opened 2023-11-02
[Impeller] Deprecate the exposed Rect fields

flar:
The exposed Rect fields prevent future optimizations which will depend on changing the internal storage format.


flar:
Note that I tried adding `[[deprecated]]` to these fields and it compiled and linted just fine on my Mac, but failed a lot of lint tasks on other platforms when I pushed it.

flar:
In the end these fields should be made private within a few days so the warning is just to prevent new accesses from entering the code base in the meantime.

bdero (approved):
LGTM

PR #47617, opened 2023-11-02
[Impeller] Use specialization constant for blur pipelines decal feature.

jonahwilliams:
Minor specialization constant cleanup


bdero (approved):
LGTM

